PR #133, opened 2015-02-06
Missed hint prs 219

wpride:


orangejenny:
This doesn't have all of the code. Compare QuestionWidget's addHelpPlaceholder with https://github.com/dimagi/opendatakit.collect/blob/master/src/org/odk/collect/android/widgets/QuestionWidget.java

Let's just get the 2.19 version right and it'll get merged in when 2.19 qa fixes get merged into master.

PR #137, opened 2015-02-07
Introduced developer flagged auto-refresh. 

ctsims:
Not ready for prime-time but  usable for testing


orangejenny:
Lovely, thanks for fitting this in, Clayton.

PR #142, opened 2015-02-16
Asynchronous Restore Requests

ctsims:
Extends the sync protocol to accept a 202 response from the server that is processed in full, and then instructs CommCare to wait for a set amount of time before requesting additional data.

Extends asynchronous tasks to be able to inform dialogs that they should enable/disable cancellation.

### DIFF
--- a/app/src/org/commcare/android/tasks/DataPullTask.java
+++ b/app/src/org/commcare/android/tasks/DataPullTask.java
@@ -79,17 +79,30 @@ public abstract class DataPullTask<R> extends CommCareTask<Void, Integer, Intege
     int mTotalItems = -1;
     long mSyncStartTime;
     
+    /** Time (in ms since epoch) when sync should be attempted again **/
+    private long mRetryAt;
+    
+    /** Time (in ms since epoch) when server requested an asynchronous wait **/
+    private long mRetryBegan;
+    
     private boolean wasKeyLoggedIn = false;
+    
+    /**
+     * Whether the task can be cancelled in its current state.
+     */
+    private boolean mIsCurrentlyCancellable = true;
     
     public static final int DATA_PULL_TASK_ID = 10;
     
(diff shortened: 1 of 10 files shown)


app/src/org/commcare/android/tasks/DataPullTask.java
```diff
@@ -219,196 +232,67 @@ public void reportProgress(int progress) {
...
-                        CommCareApplication._().logIn(CryptUtil.unWrapKey(ukr.getEncryptedKey(), password), ukr);
-                        wasKeyLoggedIn = true;
+                    //Check whether we're currently waiting for a retry timer to expire
+                    if(mRetryAt != -1 && !(System.currentTimeMillis() > mRetryAt)) {
+                        //if we are, wait for a short amount of time and then jump back to the beginning of the loop.
+                        //we have to poll this to be able to check for user cancel inputs.
+                        try {
+                            int secondsUntilSync = (int)(mRetryAt - System.currentTimeMillis());
+                            int secondsSinceStart = (int)(System.currentTimeMillis() - this.mRetryBegan);
+                            this.publishProgress(PROGRESS_SERVER_PROCESSING, secondsUntilSync, secondsSinceStart);
```
orangejenny: These are milliseconds, how about calling them millisUntilSync and millisSinceStart?

orangejenny:
One minor comment. Will merge once this PR is mergeable.

wpride:
@ctsims status of this? no longer mergable and jenny's comment

ctsims:
I need to reconcile this with some changes in the task, will try to get it
fixed up as soon as Simon's side is ready and we can test directly against
HQ.

-Clayton

phillipm:
can we close this?

ctsims:
Yeah, I'll do that. It's pointless until HQ will resource their side.

PR #143, opened 2015-02-18
PR fixes

orangejenny:
Fixes for https://github.com/dimagi/commcare-odk/pull/115

- "Continue" text wasn't correctly implemented
- Double-pane layout for extra large screens in landscape wasn't letting you swipe through tabbed case details. It's still a little weird - if you swipe and only touch the left side of the screen, it's possible that nothing will happen (no navigation because tabbed case details aren't in the right state, but tabbed case details don't change state because you didn't actually touch them) - but I don't think it's unacceptably weird.


orangejenny:
http://manage.dimagi.com/default.asp?157435

PR #149, opened 2015-02-20
Make swiping backwards on a list of forms take you to the previous screen

orangejenny:
http://manage.dimagi.com/default.asp?157945#878277

Slightly concerned that I'm opening up a can of worms with this, but I don't have a good argument against swiping being equivalent to "back" in as many places as possible.


wpride:
I like it a lot

PR #151, opened 2015-02-23
Fix build (graphing)

orangejenny:
Addendum to https://github.com/dimagi/commcare-odk/pull/150

### DIFF
--- a/app/project.properties
+++ b/app/project.properties
@@ -11,5 +11,5 @@
 split.density=false
 # Project target.
 target=Google Inc.:Google APIs:19
-android.library.reference.1=../libraries/mapballoons/android-mapviewballoons
-android.library.reference.3=../libraries/achartengine/achartengine
+android.library.reference.1=..\\libraries\\mapballoons\\android-mapviewballoons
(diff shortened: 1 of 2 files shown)


wpride:
@orangejenny thanks!

app/project.properties
```diff
@@ -11,5 +11,5 @@
 split.density=false
 # Project target.
 target=Google Inc.:Google APIs:19
-android.library.reference.1=../libraries/mapballoons/android-mapviewballoons
```
ctsims: was this not working prior? I'd really like to prevent us from needing these kinds of no-op syntax updates wherever possible (the 3->2 is fine, since that'll get updated automatically, but the mixed filepath separators seem like they're gonna require this file to get updated again int he future.

orangejenny: Something definitely needed to be updated, but it was probably the 3 => 2, I imagine I just have some eclipse checkbox checked somewhere that caused the separator change. I'm happy to change it, is there any reason to prefer one style? (i'm guessing the forward slashes, since they were already there)

orangejenny: will fixed this

PR #154, opened 2015-02-24
Allow swiping on all Android versions

orangejenny:
http://manage.dimagi.com/default.asp?157953

Not sure why I limited this to KitKat+ in the first place - looked over the original PR's changes and didn't find anything that's not in API level 14.


wpride:
@orangejenny confirmed working on a 2.X device?

PR #158, opened 2015-02-25
String Password Method Refactor

ctsims:
Refactored how we were manipulating password strings to be much more consistent


app/src/org/commcare/android/database/app/models/UserKeyRecord.java
```diff
@@ -121,13 +125,22 @@ public int getType() {
...
     
+    private static final int DEFAULT_SALT_LENGTH = 6;
+    
     public static String generatePwdHash(String pwd) {
+        return generatePwdHash(pwd, PropertyUtils.genGUID(DEFAULT_SALT_LENGTH).toLowerCase());
+    }
+    
+    public static String extractSalt(String pwdString) {
+        Matcher m = HASH_STRING_PATTERN.matcher(pwdString);
+        if(m.matches()) { return m.group(2);}
```
wpride: could use a comment either here or method - what is m.group(2)

ctsims:
@phillipm Changes look good

PR #161, opened 2015-02-26
Module filtering

orangejenny:
Adds a `relevant` attribute to `<menu>` elements in suite.xml


orangejenny:
cross-requested with https://github.com/dimagi/commcare/pull/50

orangejenny:
@ctsims addressed PR comments

PR #162, opened 2015-03-02
Makes the Barcode widget text editable

charlesfleche:
In some circumstances (low lights, partially damaged barcode) a barcode maybe be impossible to scan. This PR improve the usability of the Barcode widget by making its content editable by the user.

This PR simply replaces the TextView widget by an EditText.

### DIFF
--- a/app/src/org/odk/collect/android/widgets/BarcodeWidget.java
+++ b/app/src/org/odk/collect/android/widgets/BarcodeWidget.java
@@ -29,9 +29,9 @@ import android.view.Gravity;
 import android.view.View;
 import android.view.inputmethod.InputMethodManager;
 import android.widget.Button;
+import android.widget.EditText;
 import android.widget.LinearLayout;
 import android.widget.TableLayout;
-import android.widget.TextView;
 import android.widget.Toast;
 
 /**
@@ -41,7 +41,7 @@ import android.widget.Toast;
  */
 public class BarcodeWidget extends QuestionWidget implements IBinaryWidget {
     private Button mGetBarcodeButton;
-    private TextView mStringAnswer;
+    private EditText mStringAnswer;
     private boolean mWaitingForData;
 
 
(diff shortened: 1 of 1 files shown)


wpride:
Hmm... I'm not entirely convinced this is something we want in every use case - presumably many projects want a barcode to be acquired only directly with no room for human interaction/error. I'd think this should be added as a configurable appearance setting, similar to the way appearance="acquired" configures image questions

ctsims:
Agree that as a global change we wouldn't want people to be able to
manipulate the barcode scan.

Will: Barcode is a datatype, right? that means it should be pretty easy to
set this as an appearance, right?

charlesfleche:
The read-only aspect of the BarCode widget makes lot of sense, I agree. However, what should be the user experience / way out in case of unreadable / damaged BarCode ?

This happened to me quite a lot with low specs smartphones : scanning a barcode requires to be outdoor to maximize the light hitting the sensor. Indoor / night scanning wouldn't work at all. When this happens I [generate a barcode in a Linux terminal](http://linux.die.net/man/1/qrencode) and scan my laptop screen…

Could a setting like "Change Locale" do the trick ? By default Barcodes would be RO, but it would be possible to set them RW if required on per user / per handset basis ?

wpride:
I think you could accomplish this the equivalent of this by adding a String
question, either below the barcode question in a field list or possibly
toggled by a single select that asks if you want to use the barcode scanner
or not and updates the relevancy of the barcode/string questions
accordingly. You could configure this at the app level on a per user/device
basis via custom user data.

Would that be sufficient in the cases you're thinking of?

charlesfleche:
Thanks @wpride, this would work. However, shouldn't this Barcode issue resolved at an higher level ? I mean, at the moment, what do you guys usually do when you can not read barcodes on the field ?

wpride:
I don't have a great answer to that, likely better question for
commcare-users than for us devs, I've not heard of this being a blocking
issue for people before

charlesfleche:
Fair enough. I'll ask commcare-users. Thanks @wpride

wpride:
closing due to https://github.com/dimagi/commcare-odk/pull/283

PR #163, opened 2015-03-02
Handling dialer/sms app failures gracefully

dcluna:
Fix for callout failures, as described in [157429 - CommCare Doesn't Handle Failed Phone Dialer Intent Callout failure gracefully] (http://manage.dimagi.com/default.asp?157429#883355).

In case of failure (when there is no dialer/SMS app installed), the chooser displays an error toast instead of crashing the app.

### DIFF
--- a/app/assets/locales/messages_ccodk_default.txt
+++ b/app/assets/locales/messages_ccodk_default.txt
@@ -289,7 +289,8 @@ notification.sync.connections.action=You don't have any active data connections 
 
 #Add new localization strings here
 
-
+callout.failure.dialer=Dialer app not found
+callout.failure.sms=SMS app not found
 
 
 
--- a/app/assets/locales/messages_ccodk_default.txt
+++ b/app/assets/locales/messages_ccodk_default.txt
@@ -289,7 +289,7 @@ notification.sync.connections.action=You don't have any active data connections 
 
 #Add new localization strings here
 
-callout.failure.dialer=Dialer app not found
+callout.failure.dialer=Device is not currently configured to make telephone calls
 callout.failure.sms=SMS app not found
 
 
--- a/app/src/org/commcare/dalvik/activities/CallOutActivity.java
+++ b/app/src/org/commcare/dalvik/activities/CallOutActivity.java
@@ -12,6 +12,9 @@ import android.net.Uri;
 import android.os.Bundle;
 import android.telephony.PhoneStateListener;
 import android.telephony.TelephonyManager;
+import android.widget.Toast;
+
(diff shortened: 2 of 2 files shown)


app/src/org/commcare/dalvik/activities/CallOutActivity.java
```diff
@@ -107,16 +107,17 @@ public void onCancel(DialogInterface dialog) {
...
     
     private void dispatchAction(String action) {
+        // using createChooser to handle any errors gracefully
         if(Intent.ACTION_CALL.equals(action) ) {
             tManager.listen(listener, PhoneStateListener.LISTEN_CALL_STATE);
             
             Intent call = new Intent(Intent.ACTION_CALL);
             call.setData(Uri.parse("tel:" + number));
-            startActivity(call);
+            startActivity(Intent.createChooser(call, "Dialing " + number));
```
ctsims: Two comments

1) Does this create a chooser if there's only one dialer on the phone? That's probably confusing to end users who have to provide input that there's only one choice for.

2) We should make sure this is localized (pulled from translation files) if we are going to use this approach

app/src/org/commcare/dalvik/activities/CallOutActivity.java
```diff
@@ -107,16 +110,25 @@ public void onCancel(DialogInterface dialog) {
...
         if(Intent.ACTION_CALL.equals(action) ) {
             tManager.listen(listener, PhoneStateListener.LISTEN_CALL_STATE);
             
             Intent call = new Intent(Intent.ACTION_CALL);
             call.setData(Uri.parse("tel:" + number));
-            startActivity(call);
+            if(call.resolveActivity(getPackageManager()) != null){
+                startActivity(call);
+            } else {
+                Toast.makeText(getApplicationContext(), Localization.get("callout.failure.dialer"), Toast.LENGTH_SHORT).show();
```
ctsims: In general when inside of an Activity, you should use the current activity as the Context argument, so any details about it can be properly passed along to the next layer.

app/assets/locales/messages_ccodk_default.txt
```diff
@@ -289,7 +289,8 @@ notification.sync.connections.action=You don't have any active data connections
 
 #Add new localization strings here
 
-
+callout.failure.dialer=Dialer app not found
```
ctsims: The SMS one I think will make sense to a lot of users, but I think "Dialer app" might be a bit confusing for folks.

Can we make this something that end-users might be more familiar with, like

No calling app found. The current device is not currently configured to make telephone calls.

or something?

PR #165, opened 2015-03-03
Acra reports

wpride:


app/src/org/commcare/android/util/FileUtil.java
```diff
@@ -395,4 +396,20 @@ public static String getDumpDirectory(Context c){
                 return null;
             }
         }
+
+    public static Properties loadProperties(Context c) throws IOException {
```
ctsims: Hey,

out of curiosity, this will still result in our submission credentials being pretty cleanly available in the dissasembled APK, right? What is their best practice around protecting your credentials?

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -1,28 +1,5 @@
 package org.commcare.dalvik.activities;
 
-import org.commcare.android.database.global.models.ApplicationRecord;
```
ctsims: this whole file is no-ops

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -785,6 +786,7 @@ private void select() {
         // create intent for return and store path
         Intent i = new Intent(EntitySelectActivity.this.getIntent());
         i.putExtra(SessionFrame.STATE_DATUM_VAL, selectedIntent.getStringExtra(SessionFrame.STATE_DATUM_VAL));
+        ACRAUtil.addCustomData(ACRAUtil.CURRENT_CASE, selectedIntent.getStringExtra(SessionFrame.STATE_DATUM_VAL));
```
ctsims: I'm a bit mixed on whether it's OK for us to submit Case Id's to ACRA's stuff unless we're convinced we have a full chain of custody that's acceptable. We actually host our own server for htis data, right?

app/src/org/commcare/dalvik/application/CommCareApplication.java
```diff
@@ -64,67 +88,55 @@
...
+import java.util.Vector;
+
+import javax.crypto.SecretKey;
 
 /**
  * @author ctsims
  *
  */
+@ReportsCrashes(
+        formUri = "https://your/cloudant/report",
```
ctsims: what's up with the defaults? can we replace the ones that are find-and-fill with clear indicators that they're populated elsewhere?

app/src/org/commcare/dalvik/application/CommCareApplication.java
```diff
@@ -179,27 +192,44 @@ public void onCreate() {
...
         //StorageManager.registerStorage("fixture", this.getStorage("fixture", FormInstance.class));
-        
+
 //        Logger.registerLogger(new AndroidLogger(CommCareApplication._().getStorage(AndroidLogEntry.STORAGE_KEY, AndroidLogEntry.class)));
-//        
+//
 //        //Dump any logs we've been keeping track of in memory to storage
 //        pil.dumpToNewLogger();
-        
+
```
ctsims: tab changes? these are clean, right? IE: tabs -> spaces, not spces -> tabs, right?

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2762,6 +2763,10 @@ public void savingComplete(int saveStatus) {
      * @return status as determined in FormEntryController
      */
     public int saveAnswer(IAnswerData answer, FormIndex index, boolean evaluateConstraints) {
+
+        ACRAUtil.addCustomData(ACRAUtil.LAST_ANSWER_DATA, (answer == null ? "null" : answer.getDisplayText()));
```
ctsims: This is a super interesting idea. I like this a lot. For debugging it might be good to get both the display and data text in case the data has something interesting going on.

Also: Would be good to make sure every wrapping call like this is 100% execution safe. It's good to check the null thing here, but would like it to be the case that it's impossible for our logging calls to crash things if at all possible (not sure how cleanly we can wrap those contexts in java, though)

app/src/org/odk/collect/android/widgets/QuestionWidget.java
```diff
@@ -116,7 +117,7 @@ public void onClick(View v) {
         addQuestionText(p);
         addHelpPlaceholder(p);
         addHintText(p);
-
+        ACRAUtil.addCustomData(ACRAUtil.LAST_QUESTION_PROMPT, mPrompt.getLongText());
```
ctsims: as a heads up, question widgets can be created without being the last displayed prompt. The activity setting the view is really the only way we know.

app/src/org/commcare/android/util/FileUtil.java
wpride: http://stackoverflow.com/questions/14407884/how-can-i-secure-this-api-in-an-apk-file
http://stackoverflow.com/questions/14407884/how-can-i-secure-this-api-in-an-apk-file

Doesn't seem like there's really a great solution here. Setting up some server-side process that handles validating users and then using our API key might be the safest, not sure how much we want to put into this

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
wpride: Yeah, we host it on Cloudant. Maybe we should sit down at some point and go over everything?

app/src/org/commcare/android/util/FileUtil.java
wpride: Looks like some people use the NDK to obscure their keys:

http://stackoverflow.com/questions/27243896/c-c-code-in-order-to-store-private-keys-in-android-app

I think its a matter of how much we want to put into this

wpride:
@ctsims this is ready to be reviewed with hippaa information removed

PR #166, opened 2015-03-03
Refreshing home screen dropdown on locale change

dcluna:
Fix for the bug [159047](http://manage.dimagi.com/default.asp?159047), in which the home screen dropdown would not be refreshed automatically when settings changed.


wpride:
Not seeing where this function is coming from?

PR #169, opened 2015-03-05
"Awesome mode" callout crash

dcluna:
Fix for [159503](http://manage.dimagi.com/default.asp?159503#887208), which crashes in devices that split the screen in the dial activity ("awesome mode").

### DIFF
--- a/app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
+++ b/app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
@@ -8,6 +8,7 @@ import java.util.Vector;
 
 import org.commcare.android.adapters.EntityListAdapter;
 import org.commcare.android.framework.CommCareActivity;
+import org.commcare.android.logic.DetailCalloutListenerDefaultImpl;
 import org.commcare.android.models.AndroidSessionWrapper;
 import org.commcare.android.models.Entity;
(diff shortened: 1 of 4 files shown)


app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -44,6 +46,7 @@
 import android.content.Intent;
 import android.content.res.Configuration;
 import android.database.DataSetObserver;
+import android.net.Uri;
```
dcluna: Unused import, should be removed

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -788,6 +791,19 @@ private void select() {
         setResult(RESULT_OK, i);
         finish();
     }
+
+    // CommCare-159503: implementing DetailCalloutListener so it will not crash the app when requesting call/sms
```
orangejenny: This code is open source but FogBugz tickets are internal. Better to explain the bug than refer to a ticket.

czue: @orangejenny we do this a bunch on the HQ team, fyi. Agree is a bit odd, but definitely can be useful.

PR #186, opened 2015-03-16
Drop down search crash

phillipm:
When filtering over saved forms and changing the form status to list, the filtering method would crash because it tried to filter data before it was loaded. This commit  ensures that filtering only occurs after FormRecord text data has been loaded.

The biggest change is that IncompleteFormListAdapter extends FormRecordLoadListener and adds itself to the listeners to be called when FormRecordLoaderTask finishes loading data, in order to filterValues at the proper time.

Fixes [bug 161420](http://manage.dimagi.com/default.asp?161420)

### DIFF
--- a/app/src/org/commcare/android/adapters/IncompleteFormListAdapter.java
+++ b/app/src/org/commcare/android/adapters/IncompleteFormListAdapter.java
@@ -11,6 +11,7 @@ import java.util.Vector;
 import org.commcare.android.database.SqlStorage;
 import org.commcare.android.database.user.models.FormRecord;
 import org.commcare.android.tasks.FormRecordLoaderTask;
+import org.commcare.android.tasks.FormRecordLoadListener;
 import org.commcare.android.util.AndroidCommCarePlatform;
 import org.commcare.android.util.SessionUnavailableException;
 import org.commcare.android.view.IncompleteFormRecordView;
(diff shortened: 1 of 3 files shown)


ctsims:
Hey,

This looks pretty reasonable.

Two questions:
1) Did priority loading/filtering work before at all and this disables it in favor of the conservative loading technique? Is probably fine if priority filtering doesn't work, but priority loading (IE: getting the labels for the items on the screen first) is pretty valuable

2) Have you tried screen rotation with this? IE: Start the process, rotate the screen before the task finishes, make sure it keeps working (including filtering). a few things got moved into the activity state it looks like, (query string at least) and I'm not sure whether it'll move over correctly.

phillipm:
1) Priority loading still works, I think. I said it is "broken" because FormRecordListActivity.notifyPriorityLoaded is always called, even when a form record has not been registered as high-priority.

While IncompleteFormListAdapter.notifyLoaded is used to call filterValues at the correct time, I can't really see a use for implementing IncompleteFormListAdapter.notifyPriorityLoaded, unless we want to do incremental filtering of loaded forms.

I feel like there are some redundancies in the way form record info is loaded and displayed. I'll leave that refactor for another time, but when we do it we should also clean up the priority code. And of course take this criticism with a grain of salt: I don't yet fully understand the system and its constraints.


2) Good call! I just tried it out on my phone quickly and things seemed to work fine. Should I look into how screen rotation is effecting form lists in the code? Or is a superficial level pass good enough?

ctsims:
All sounds good. Incremental filtering of loaded forms seems out of scope
for sure.

I'm comfortable with a superficial test as long as you were able to test
the "start loading in one orientation, finish loading in the other"
behavior, which is where the majority of our async issues show up.

-Clayton

PR #187, opened 2015-03-18
Bugfix: Form Serialization Crash

ctsims:
Fixes an issue where when our caching optimization fails and it crashes the app

http://manage.dimagi.com/default.asp?161629#898010

### DIFF
--- a/app/src/org/odk/collect/android/tasks/FormLoaderTask.java
+++ b/app/src/org/odk/collect/android/tasks/FormLoaderTask.java
@@ -24,12 +24,15 @@ import java.io.IOException;
 
 import javax.crypto.spec.SecretKeySpec;
 
+import org.commcare.android.javarosa.AndroidLogger;
+import org.commcare.android.tasks.ExceptionReportTask;
 import org.javarosa.core.model.FormDef;
 import org.javarosa.core.model.instance.InstanceInitializationFactory;
 import org.javarosa.core.model.instance.TreeElement;
 import org.javarosa.core.model.instance.TreeReference;
 import org.javarosa.core.reference.ReferenceManager;
 import org.javarosa.core.reference.RootTranslator;
+import org.javarosa.core.services.Logger;
 import org.javarosa.core.util.externalizable.DeserializationException;
 import org.javarosa.form.api.FormEntryController;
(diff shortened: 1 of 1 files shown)


app/src/org/odk/collect/android/tasks/FormLoaderTask.java
```diff
@@ -329,28 +343,34 @@ public FormDef deserializeFormDef(File formDef) {
...
-                e.printStackTrace();
-            } catch (IOException e) {
-                e.printStackTrace();
+            } finally {
+                //make sure we clean up the stream
+                if(fos != null) {
+                    try {
+                        fos.close();
+                    } catch (IOException e) {
+                        // there was already an error, this one is just cruft
```
phillipm: Not sure I follow the logic of this try/catch
If we are already not catching IOException and letting it bubble up, why wrap fos.close() in a try? Wouldn't it suffice to just close the stream and let any exceptions propagate up the call stack?

ctsims: In this case we don't want to trump the exception getting thrown from fd.writeExternal. If, for instance, we threw a different type of exception (in the ticket it's a stack overflow), and then while cleaning up fos.close() errored out, we'll end up logging/reporting fos.close()'s exception path. 

It did just occur to me, however, that I shouldn't be closing dos above if we're closing fos in the finally cleanup, either way, I'll fix that

app/src/org/odk/collect/android/tasks/FormLoaderTask.java
```diff
@@ -329,28 +344,38 @@ public FormDef deserializeFormDef(File formDef) {
...
+            } finally {
+                //make sure we clean up the stream
+                if(outputStream != null) {
+                    try {
+                        outputStream.close();
+                    } catch (IOException e){
+                        // Swallow this. If we threw an exception from inside the 
+                        // try, this close exception will trump it on the return 
+                        // path, and we care a lot more about that exception
+                        // than this one.
```
ctsims: @phillipm Let me know if this clarifies it

phillipm:
@ctsims yeah, that makes sense. Thanks for the clarification!

PR #188, opened 2015-03-19
Refactoring how a session's form record is connected to a form instance

phillipm:
Moved logic for linking & updating Form Record to a form instance into InstanceProvider.

[Ticket 101852](http://manage.dimagi.com/default.asp?101852)

Currently undergoing QA testing


app/src/org/commcare/dalvik/odk/provider/InstanceProvider.java
```diff
@@ -365,10 +386,94 @@ public int update(Uri uri, ContentValues values, String where, String[] whereArg
...
+            return;
+        }
+
+        Cursor c = getContext().getContentResolver().query(instanceUri, null, null, null, null);
+        boolean complete = false;
+        try {
+            // register the instance uri and its status with the session
+            // NOTE: there is no safety checks preventing us from binding two
+            // different form instances to the form record found in the
+            // session. We just need to be careful to flush the session's form
```
phillipm: Should I be concerned about this NOTE?

app/src/org/commcare/dalvik/services/CommCareSessionService.java
```diff
@@ -272,20 +295,96 @@ public void run() {
...
+            pool.expire();
+            this.stopForeground(true);
+        }
+    }
+
+    /**
+     * Close user database, broadcast session logout, expire the cypher pool
+     */
+    // TODO: Remove this function after startLogout/finishLogout are finished
     public void logout() {
```
phillipm: Remove me

app/src/org/commcare/android/tasks/DataPullTask.java
```diff
@@ -133,7 +133,7 @@ public DataPullTask(String username, String password, String server, String keyP
     protected void onCancelled() {
         super.onCancelled();
         if(wasKeyLoggedIn) {
-            CommCareApplication._().logout();
+            CommCareApplication._().getSession().startLogout();
```
phillipm: I've replaced every instance of CommCareApplication._().logout() with the session startLogout.
startLogout eventually triggers finishLogout, which calls the former.
NOTE: finishLogout launches a CommCareHomeActivity intent at the end, prompting the user to log back in. I'm not sure that every instance of CommCareApplication._().logout() solicits this behavior, please check.

app/src/org/commcare/dalvik/services/CommCareSessionService.java
```diff
@@ -275,55 +295,126 @@ public void run() {
...
+            CommCareApplication._().logout();
+
             pool.expire();
             this.stopForeground(true);
+
+            // Re-direct to the home screen
+            Intent loginIntent = new Intent(getApplicationContext(),
+                    CommCareHomeActivity.class);
+            // XXX: Is this the best way to start the login view? It sort of
+            // just pops up even when CommCare isn't active... not ideal -- PLM
```
phillipm: See warning comment ^

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -229,10 +238,19 @@
     
     private static String mHeaderString;
     
+    // Was the form saved? Used to set activity return code and decide whether
+    // form save callbacks have saving work to do.
+    // XXX: hasSaved doesn't seem to reset to false if the form is editted after
+    // saving. -- PLM
```
phillipm: How important is this concern ^?

app/src/org/commcare/android/tasks/DataPullTask.java
ctsims: Leaving a note for myself to come back to this.

The datapull task "login" ops are more about seating a user DB than "logging in", but this might be equivalent. Will evaluate.

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -786,6 +627,140 @@ else if(resultCode == RESULT_OK) {
...
+
+        if (current == null) {
+            // somehow we lost the form record for the current session
+            // TODO: how should this be handled? -- PLM
+            Toast.makeText(this,
+                    "Error while trying to save the form!",
+                    Toast.LENGTH_LONG);
+            Logger.log(AndroidLogger.TYPE_ERROR_WORKFLOW,
+                    "Form Entry couldn't save because of corrupt state.");
+            clearSessionAndExit(currentState);
```
ctsims: We should return here, right? We don't wanna keep processing anything further?

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -786,6 +627,140 @@ else if(resultCode == RESULT_OK) {
...
+
+        // TODO: This should be the default unless we're in some "Uninit" or "incomplete" state
+        if (FormRecord.STATUS_COMPLETE.equals(current.getStatus()) ||
+                FormRecord.STATUS_SAVED.equals(current.getStatus()) ||
+                FormRecord.STATUS_UNSENT.equals(current.getStatus())) {
+            // Viewing an old form, so don't change the historical record
+            // regardless of the exit code
+            currentState.reset();
+            if (wasExternal) {
+                this.finish();
```
ctsims: I think this won't work quite right unless the parent call returns rather than proceeds with start next fetch

app/src/org/commcare/dalvik/odk/provider/InstanceProvider.java
ctsims: I don't think this should be affected.

This is basically just noting that the instance URI needs to uniquely represent a single form record, since we assume there's a one-to-one parity.

app/src/org/commcare/dalvik/odk/provider/InstanceProvider.java
```diff
@@ -365,10 +386,95 @@ public int update(Uri uri, ContentValues values, String where, String[] whereArg
...
+            current = currentState.commitRecordTransaction();
+        } catch (Exception e) {
+            // Something went wrong with all of the connections which should exist.
+            FormRecordCleanupTask.wipeRecord(getContext(), currentState);
+
+            // Notify the server of this problem (since we aren't going to crash)
+            ExceptionReportTask ert = new ExceptionReportTask();
+            ert.execute(e);
+
+            raiseFormEntryError("An error occurred: " + e.getMessage() + " and your data could not be saved.", currentState);
```
ctsims: What is the user facing outcome here if, for instance, the form is being saved as incomplete and an error occurs. Does the app let you finish filling out the form, but then it errors when you return because the session is gone?

app/src/org/odk/collect/android/activities/FormEntryActivity.java
ctsims: Initially I think it was only relevant on the first save, not 100% sure.

app/src/org/commcare/dalvik/services/CommCareSessionService.java
ctsims: Yeah, I'm also concerned about the new task backstack setup below.

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
phillipm: This code was just moved from the MODEL_RESULT case of onActivityResult, so I actually have no idea what to do here. 
Any concrete suggestions?

ctsims: Essentially there were two code paths out of the MODEL_RESULT case before (Listed below). easiest thing to do would be to have the new method have a return code that either returned out of the method, or fell through to the next fetch. Alternative would be to more deliberately encode the potential outcomes for all of these activity returns and follow that path

1) fallthrough (break, or simply continue executing) - These exits would jump into a session. This was the return path if, say, the user canceled form entry, or a new session frame was placed on the stack and the user should be directed to it. These paths would result in the "startNextFetch" method being called at the bottom of the activity result
2) exit (paths previously ending in 'return') - these exits would not launch a new activity. They would either launch the Home Activity, or (if "finish" was called) they would leave commcare entirely (if CommCare was launched in response to an external app)

app/src/org/commcare/android/models/AndroidSessionWrapper.java
```diff
@@ -132,45 +134,50 @@ public void setFormRecordId(int formRecordId) {
...
 
-
+        // update the form record to mirror the sessions instance uri and
+        // status.
         current = current.updateStatus(instanceUri, recordStatus);
-        
+
+        // save the updated form record
         try {
             FormRecord updated = FormRecordCleanupTask.getUpdatedRecord(CommCareApplication._(), platform, current, recordStatus);
```
phillipm: Double save crash occurs here from update() calling linkToSessionFormRecord()

app/src/org/commcare/dalvik/odk/provider/InstanceProvider.java
```diff
@@ -365,10 +385,91 @@ public int update(Uri uri, ContentValues values, String where, String[] whereArg
...
 
+        // If we've changed a particular form instance's status, and not
+        // created a new entry (hence count > 0 check), we need to mirror the
+        // change in that form's record.  NOTE: this conditional is crucial,
+        // since updating a form record in turn calls this update function, so
+        // we need to break the infinite loop by only updating the form record
+        // when the the status changes.
+        if (values.containsKey(InstanceColumns.STATUS) && count > 0) {
+            try {
+                linkToSessionFormRecord(uri);
```
phillipm: This needs to happen for some updating saves, but crashes on others.
Example of crash is when you try to save twice in a row while editing a form.
Example of when it is needed, when you open up a form to edit, complete it: the form record also needs to be updated to say it is complete.

app/src/org/commcare/dalvik/services/CommCareSessionService.java
```diff
@@ -275,55 +295,130 @@ public void run() {
...
+            Intent loginIntent = new Intent(getApplicationContext(),
+                    CommCareHomeActivity.class);
+            // TODO: instead of launching here, which will pop-up the login
+            // screen even if CommCare isn't in the foreground, we should
+            // broadcast an intent, which CommCareActivity can receive if in
+            // focus and dispatch the login activity. Will also need to extend
+            // CommCareActivity's onResume to check if we need to re-login when
+            // we bring CommCare back into the foreground, so that the user
+            // can't just continue doing work while logged out. -- PLM
+            loginIntent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
```
phillipm: Didn't fix, but sketched out a fix

app/src/org/commcare/dalvik/odk/provider/InstanceProvider.java
```diff
@@ -365,10 +387,140 @@ public int update(Uri uri, ContentValues values, String where, String[] whereArg
...
+     * entry in that table
+     * @param selection A selection criteria to apply when filtering rows. If
+     * null then all rows are included.
+     * @param selectionArgs You may include ?s in selection, which will be
+     * replaced by the values from selectionArgs, in order that they appear in
+     * the selection. The values will be bound as Strings.
+     * @return URI pointing to a row in the instance table, either the one passed
+     * in or built from a query using the method arguments.
+     */
+    private Uri getInstanceRowUri(Uri potentialUri, String selection, String[] selectionArgs) {
```
phillipm: Newly added to make sure we only try to link a form record to a uri pointing to a concrete instance.

PR #189, opened 2015-03-20
Developer flag for updating to latest app build

phillipm:
Adding a flag to the app's Developer Preferences that specifies that the app will "always request the newest version" of the app.

When performing the http callout to request the newest profile file if this mode is enabled, the app should append the apprpriate HTTP url query so HQ will know to serve the latest profile regardless of release status.

See [CommCare  Feature 157896](http://manage.dimagi.com/default.asp?157896) and [HQ Feature 159646](http://manage.dimagi.com/default.asp?159646)


app/src/org/commcare/android/tasks/ResourceEngineTask.java
```diff
@@ -150,111 +154,138 @@ protected ResourceEngineOutcomes doTaskBackground(String... profileRefs) {
...
+                // app (instead of latest release), add it to the query tags of
+                // the profile reference
+                if (DeveloperPreferences.isNewestAppVersionEnabled() &&
+                        (profileUrl != null) &&
+                        ("https".equals(profileUrl.getProtocol()) ||
+                         "http".equals(profileUrl.getProtocol()))) {
+                    if (profileUrl.getQuery() != null) {
+                        // If the profileRef url already have query strings
+                        // just add a new one to the end
+                        profileRef = profileRef + "&target=build";
```
ctsims: I'd probably write this using the URL tools instead (IE:android URI classes will let you add arguments programatically rather than as strings)

PR #200, opened 2015-03-27
Unit Test Shadows for Storage and tests for Case Index Removal

ctsims:
X-requested with: https://github.com/dimagi/commcare/pull/53

Added shadows for IO-related classes to test framework (Application and SqlStorage can be instantiated)

Wired in configs between robolectric and core project.

Configured Resources to be available at the classpath level for unit tests both in ANT and local runtime.

Made minor constructor changes for some DB related classes to let them not pull from the static context (nothing should be doing that anyway, ideally!!!)

Added tests and a fix for case transactions which remove indexes


unit-tests/src/org/commcare/android/shadows/SQLiteDatabaseNative.java
```diff
@@ -0,0 +1,305 @@
...
+@Implements(net.sqlcipher.database.SQLiteDatabase.class)
+public class SQLiteDatabaseNative {
+    private android.database.sqlite.SQLiteDatabase db;
+    
+    public void __constructor__(String path, char[] password, CursorFactory factory, int flags) {
+        db = android.database.sqlite.SQLiteDatabase.openDatabase(path, null, flags);
+    }
+    
+    public void __constructor__(String path, char[] password, CursorFactory factory, int flags, SQLiteDatabaseHook hook) {
+        db = android.database.sqlite.SQLiteDatabase.openDatabase(path, null, flags);
```
orangejenny: Why is there a constructor that takes a hook if it doesn't do anything with it?

orangejenny: Oh, nm, this is implementing SQLiteDatabase so presumably there's no choice.

unit-tests/src/org/commcare/android/tests/processing/ProcessingTest.java
```diff
@@ -0,0 +1,164 @@
...
+        assertEquals("Changed Type", "changed_unit_test", c2.getTypeId());
+    }
+
+
+    private TransactionParserFactory getFactory(final SQLiteDatabase db) {
+        return new TransactionParserFactory() {
+            
+            public TransactionParser getParser(String name, String namespace, KXmlParser parser) {
+                if(LedgerXmlParsers.STOCK_XML_NAMESPACE.equals(namespace)) {
+                    //return new LedgerXmlParsers(parser, CommCareApplication._().getUserStorage(Ledger.STORAGE_KEY, Ledger.class));
```
orangejenny: Deliberately commented out?

unit-tests/src/org/commcare/android/shadows/SQLiteDatabaseNative.java
ctsims: Yeah, it's a potential issue to check for in the future, but the incoming Hook here is the sqlcipher type, so we can't forward it. We'd have to make another wrapper like we did for the cursor, which is a super pain due to the way these SQL classes do a bunch of shit in their constructors.

unit-tests/src/org/commcare/android/tests/processing/ProcessingTest.java
ctsims: I'll clean this up. Agree that this isn't a good state to leave things in.

ctsims:
@orangejenny Fixed up the rest of the issues you mentioned

PR #212, opened 2015-04-01
Make ASyncEntity handle XPathSyntax exceptions

orangejenny:
evaluate now throws an XPathSyntaxException

### DIFF
--- a/app/src/org/commcare/android/models/AsyncEntity.java
+++ b/app/src/org/commcare/android/models/AsyncEntity.java
@@ -115,6 +115,9 @@ public class AsyncEntity extends Entity<TreeReference>{
                 } catch(XPathException xpe) {
                     xpe.printStackTrace();
                     data[i] = "<invalid xpath: " + xpe.getMessage() + ">";
+                } catch (XPathSyntaxException xse) {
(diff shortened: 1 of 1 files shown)


orangejenny:
cross-requested with https://github.com/dimagi/commcare/pull/58

app/src/org/commcare/android/models/AsyncEntity.java
```diff
@@ -115,6 +115,9 @@ public Object getField(int i) {
                 } catch(XPathException xpe) {
                     xpe.printStackTrace();
                     data[i] = "<invalid xpath: " + xpe.getMessage() + ">";
+                } catch (XPathSyntaxException xse) {
```
ctsims: Question: We don't generally handle these here because we try to parse all expressions inbound when we install the app, allowing people to catch syntax errors (which should be detectable before runtime) early. Is it impossible to make sure the graph expressions are compiled during install to throw their errors up there?

orangejenny: Good point, it'd be better to catch these errors earlier. Will do a separate PR.

PR #214, opened 2015-04-02
Graphing landscape mode crash - 163637

dcluna:
Quick fix for bug [163637](http://manage.dimagi.com/default.asp?163637#908481), in which the app would crash when in "awesome mode".

### DIFF
--- a/app/src/org/commcare/android/models/AndroidSessionWrapper.java
+++ b/app/src/org/commcare/android/models/AndroidSessionWrapper.java
@@ -240,6 +240,7 @@ public class AndroidSessionWrapper {
         //TODO: this has two components which can fail. be able to roll them back
 
         String form = getSession().getForm();
+        // COMMCARE-163637: in landscape mode, this returns null, and the FormRecord should not have any null fields
         form = form != null ? form : "";
 
         FormRecord r = new FormRecord("", FormRecord.STATUS_UNSTARTED, form, key.getEncoded(), null, new Date(0));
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/android/models/AndroidSessionWrapper.java
```diff
@@ -237,8 +238,13 @@ public void commitStub() throws StorageFullException {
         SecretKey key = CommCareApplication._().createNewSymetricKey();
         
         //TODO: this has two components which can fail. be able to roll them back
-        
-        FormRecord r = new FormRecord("", FormRecord.STATUS_UNSTARTED, getSession().getForm(), key.getEncoded(), null, new Date(0));
+
+        String form = getSession().getForm();
+        // COMMCARE-163637: in landscape mode, this returns null, and the FormRecord should not have any null fields
+        form = form != null ? form : "";
```
ctsims: Do we know why this is returning null in landscape mode? Seems like a big deal for this to have different behavior between the two view styles.

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -465,7 +465,12 @@ public void onClick(View v) {
                 
                 //csims@dimagi.com - Jan 24, 2012
                 //Since these are parceled across the content resolver, there's no guarantee of reference equality.
-                //We need to manually check value equality on the type 
+                //We need to manually check value equality on the type
+
+                if(uri == null) { // COMMCARE-163637: might be null if we're using the app in "awesome mode" - in which case we just get out
```
ctsims: I'm not super comfortable with this change. Why does this return null in awesome mode?

dcluna:
When in awesome mode, CommCare does not save any SessionFrame information, probably because it is ambiguous (in a split screen, which SessionFrame is currently active? I'd say both, but the current implementation only allows for one at a time).

orangejenny:
@ctsims @dcluna 

I think there's a higher-level problem here - this is the case list, it shouldn't be trying to get a form uri at all. Closing this PR and commenting on the ticket.

PR #216, opened 2015-04-06
Landscape detail fix

phillipm:
This PR removes EntityDetailActivity's (useless) "Continue" button that is displayed while in landscape mode, and that causes crashes when pressed.

Additional fixes:
 * A few comments.
 * When rotating from landscape to portrait with an entity selected, launch EntityDetailActivity with that selection.

[Bug 163637](http://manage.dimagi.com/default.asp?163637)


ctsims:
I'm a bit confused, does this always hide the button? It should still be there when we're in awesome mode and the select screen isn't a dead end

phillipm:
Could you give me an example of when an Entity Detail Activity view isn't a dead end?

PR #217, opened 2015-04-06
Don't show next button during view + landscape mode

phillipm:
Prevent EntityDetailActivity's "Continue" button from being displayed while in landscape mode

Also fix code that rotates from landscape to portrait with an entity selected and launches EntityDetailActivity with that selection.

[Bug 163637](http://manage.dimagi.com/default.asp?163637)

### DIFF
--- a/app/src/org/commcare/dalvik/activities/EntityDetailActivity.java
+++ b/app/src/org/commcare/dalvik/activities/EntityDetailActivity.java
@@ -59,8 +59,8 @@ public class EntityDetailActivity extends CommCareActivity implements DetailCall
     
     private int detailIndex;
 
-    // Is the detail screen for showing elements without moving on to form
-    // manipulation?
+    // Is the detail screen for showing entities, without option for moving
+    // forward on to form manipulation?
     private boolean mViewMode = false;
     
     @UiElement(value=R.id.entity_detail)
--- a/app/src/org/commcare/dalvik/activities/EntityDetailActivity.java
+++ b/app/src/org/commcare/dalvik/activities/EntityDetailActivity.java
@@ -49,7 +49,6 @@ public class EntityDetailActivity extends CommCareActivity implements DetailCall
     public static final String DETAIL_ID = "eda_detail_id";
     public static final String DETAIL_PERSISTENT_ID = "eda_persistent_id";
 
-    Entry prototype;
     Entity<TreeReference> entity;
     EntityDetailAdapter adapter;
     NodeEntityFactory factory;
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -175,7 +178,16 @@ public void onCreate(Bundle savedInstanceState) {
...
                 if(oldActivity != null) {
-                    if(oldActivity.selectedIntent != null) {
+                    Intent intent = this.getIntent();
+
+                    TreeReference selectedRef = SerializationUtil.deserializeFromIntent(intent,
+                            EntityDetailActivity.CONTEXT_REFERENCE, TreeReference.class);
+                    if (selectedRef != null) {
+                        Intent detailIntent = getDetailIntent(selectedRef, null);
+                        startOther = true;
+                        startActivityForResult(detailIntent, CONFIRM_SELECT);
```
phillipm: Has the bug where rotating from landscape to portrait while nothing is selected, but has been selected in the past, then that previous selection detail view is launched. Due to the fact that I don't know how to reset the CONTEXT_REFERENCE extra, I think...

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -175,7 +178,16 @@ public void onCreate(Bundle savedInstanceState) {
...
+                    } else if (oldActivity.selectedIntent != null) {
+                        // XXX: old code that didn't work -- PLM
```
phillipm: Not sure why the following code wasn't working. I couldn't inspect oldActivity.selectedIntent in the debugger, so maybe it was being cleaned up somehow...

app/src/org/commcare/dalvik/activities/EntityDetailActivity.java
```diff
@@ -43,11 +43,12 @@
     
     private CommCareSession session;
     private AndroidSessionWrapper asw;
-    public static final String IS_DEAD_END = "eda_ide";
+
```
phillipm: Got rid of unused IS_DEAD_END

app/src/org/commcare/dalvik/activities/EntityDetailActivity.java
```diff
@@ -81,6 +86,14 @@ public void onCreate(Bundle savedInstanceState) {
         
         Vector<Entry> entries = session.getEntriesForCommand(passedCommand == null ? session.getCommand() : passedCommand);
         prototype = entries.elementAt(0);
+
+        // (We shouldn't need the "" here, but we're avoiding making changes to
+        // commcare core for release issues)
+        if (entries.size() == 1 && (prototype.getXFormNamespace() == null || prototype.getXFormNamespace().equals(""))) {
```
ctsims: We do this calculation in two places, right? Seems like it's complex enough that 's dangerous to not centralize that it possible? I know that applies to a couple parts of the Awesome/normal display, but this in particular feels very fragile to recompute in two places.

phillipm:
Consolidated view mode detection into CommCareSession (cross-requested https://github.com/dimagi/commcare/pull/61)

PR #218, opened 2015-04-07
Module filter fix

ctsims:
Depends on https://github.com/dimagi/commcare/pull/62

Includes menu command ID with Evaluation Context request when filtering modules


ctsims:
Fix for: http://manage.dimagi.com/default.asp?162644

PR #222, opened 2015-04-09
Don't cache EvaluationContext

orangejenny:
Partial fix for http://manage.dimagi.com/default.asp?164631

Evaluation context was created using the first module with a relevancy condition, which could cause later modules with relevancy conditions to fail.

### DIFF
--- a/app/src/org/commcare/android/adapters/MenuAdapter.java
+++ b/app/src/org/commcare/android/adapters/MenuAdapter.java
@@ -62,9 +62,7 @@ public class MenuAdapter implements ListAdapter {
                 try {
                     XPathExpression relevance = m.getMenuRelevance();
                     if (m.getMenuRelevance() != null) {
-                        if (ec == null) {
-                            ec = asw.getEvaluationContext(m.getId());
-                        }
+                        ec = asw.getEvaluationContext(m.getId());
                         if (XPathFuncExpr.toBoolean(relevance.eval(ec)).booleanValue() == false) {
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/android/adapters/MenuAdapter.java
```diff
@@ -87,9 +85,7 @@ public MenuAdapter(Context context, CommCarePlatform platform, String menuID){
                         try {
                             XPathExpression mRelevantCondition = m.getCommandRelevance(m.indexOfCommand(command));
                             if(mRelevantCondition != null) {                            
-                                if (ec == null) {
-                                    ec = asw.getEvaluationContext();
-                                }
+                                ec = asw.getEvaluationContext();
```
ctsims: This caching should still be working, btw, since the EC here is global to the session

app/src/org/commcare/android/adapters/MenuAdapter.java
```diff
@@ -87,9 +85,7 @@ public MenuAdapter(Context context, CommCarePlatform platform, String menuID){
                         try {
```
ctsims: I think we need to move this try/catch up to the whole scope, btw, to catch the menu relevancy expressions too

orangejenny: Yeah, they were being caught but not displayed in any remotely helpful fashion. Fixed, should be good to go now.

PR #224, opened 2015-04-09
fix for external API login

wpride:


ctsims:
Is this ready for review? If so, can you mark it as such?

PR #228, opened 2015-04-15
another fix for  spannables in case list

wpride:


orangejenny:
hey, can you comment these like you did on the other PR? some well-meaning person is going to change them back

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -623,7 +626,18 @@ public boolean onOptionsItemSelected(MenuItem item) {
                 createSortMenu();
                 return true;
             case MENU_MAP:
+                try
+                {
+                    Class.forName("com.google.android.maps.MapActivity");
+                }
+                catch (Throwable e1) {
```
phillipm: Isn't it a good idea to be explicit about the exception being caught?
In this case "ClassNotFoundException"

phillipm: Maybe add a note saying why this try/catch is here (4.2.2 map crashing)

PR #230, opened 2015-04-16
Change login duration setting to be represent seconds not hours

phillipm:
To test apps we need to be able to set session expiration lengths to be on the order of seconds/minutes, so changed this app preference to be in terms of seconds instead of hours.

### DIFF
--- a/app/src/org/commcare/dalvik/preferences/CommCarePreferences.java
+++ b/app/src/org/commcare/dalvik/preferences/CommCarePreferences.java
@@ -238,7 +238,7 @@ public class CommCarePreferences extends PreferenceActivity implements OnSharedP
         SharedPreferences properties = CommCareApplication._().getCurrentApp().getAppPreferences();
 
         // default to 24 hours
-        return properties.getFloat(LOGIN_DURATION, 28800);
+        return properties.getInt(LOGIN_DURATION, 28800);
     }
 
     /*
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/dalvik/preferences/CommCarePreferences.java
```diff
@@ -86,7 +86,7 @@
 
     public final static String FUZZY_SEARCH = "cc-fuzzy-search-enabled";
 
-    public final static String LOGIN_DURATION = "cc-login-duration-hours";
+    public final static String LOGIN_DURATION = "cc-login-duration-seconds";
```
phillipm: This name change is okay, that is, nothing on HQ uses this yet, right?

app/src/org/commcare/dalvik/preferences/CommCarePreferences.java
```diff
@@ -230,10 +230,15 @@ public static boolean isFuzzySearchEnabled() {
...
+    /**
+     * @return How many seconds should a user session remain open before
+     *         expiring?
+     */
     public static int getLoginDuration() {
         SharedPreferences properties = CommCareApplication._().getCurrentApp().getAppPreferences();
 
-        return properties.getInt(LOGIN_DURATION, 24);
+        // default to 24 hours
+        return Integer.parseInt(properties.getString(LOGIN_DURATION, "28800"));
```
phillipm: Preferences set in HQ come in as strings, so parse into integers. This assumption is okay, right?

ctsims: The assumption is OK, but i'd prefer to wrap with a default if the parse fails

PR #231, opened 2015-04-17
Enable paste/copy for text entry in forms

phillipm:
Remove form entry options to clear an answer or delete a repeat entry. These options pop up when you tap and hold text entry boxes, which blocks availability of normal android paste/select functionality.

### DIFF
--- a/app/src/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/app/src/org/odk/collect/android/activities/FormEntryActivity.java
@@ -1335,7 +1335,7 @@ public class FormEntryActivity extends FragmentActivity implements AnimationList
     @Override
     public void onCreateContextMenu(ContextMenu menu, View v, ContextMenuInfo menuInfo) {
         super.onCreateContextMenu(menu, v, menuInfo);
-            menu.add(0, v.getId(), 0, StringUtils.getStringSpannableRobust(this, R.string.clear_answer));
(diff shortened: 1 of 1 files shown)


app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -1327,45 +1324,6 @@ private void clearAnswer(QuestionWidget qw) {
...
 
-
-    /*
-     * (non-Javadoc)
-     * @see android.app.Activity#onCreateContextMenu(android.view.ContextMenu, android.view.View, android.view.ContextMenu.ContextMenuInfo)
-     */
-    @Override
-    public void onCreateContextMenu(ContextMenu menu, View v, ContextMenuInfo menuInfo) {
-        super.onCreateContextMenu(menu, v, menuInfo);
-            menu.add(0, v.getId(), 0, StringUtils.getStringSpannableRobust(this, R.string.clear_answer));
```
phillipm: Leaving in R.string.clear_answer text entry, and all the other corresponding text entries.
They will be rendered unused, but might be useful in the future? Or should I just delete them?

ctsims: Hm, that's a tough question, generally. I'd say let's remove it. If we want
them in the future it will likely be to add this functionality back so we
can pull them out of version control.

phillipm:
Decided to leave in clear answer option for select-one/select-multiple question types. Opened a new PR for that: https://github.com/dimagi/commcare-odk/pull/233

PR #256, opened 2015-05-04
Restyling - new login screen

dcluna:
Redesigned login screen


app/res/drawable/login_edit_text.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
+    android:shape="rectangle">
+    <padding>8dp</padding>
+    <size android:height="48dp"></size>
```
ctsims: These are standard constants for us now, right? They should be either in a style or dimensions file, shoudln't they?

app/res/layout/screen_login.xml
```diff
@@ -1,48 +1,56 @@
...
         android:layout_alignParentTop="true"
         android:orientation="vertical" >
 
         <FrameLayout
             android:layout_width="fill_parent"
-            android:layout_height="0px" android:layout_weight=".2" android:id="@+id/screen_login_banner_pane">
+            android:layout_height="0px"
+            android:layout_weight=".2"
+            android:paddingTop="8dp"
+            android:paddingLeft="8dp"
```
ctsims: I think these 8dp's should also be in either a standard style or in dimens.xml

app/res/layout/screen_login.xml
```diff
@@ -1,48 +1,56 @@
...
         android:layout_height="wrap_content"
-        android:layout_above="@id/screen_login_bottom_pane"
         android:layout_alignParentTop="true"
         android:orientation="vertical" >
 
         <FrameLayout
             android:layout_width="fill_parent"
-            android:layout_height="0px" android:layout_weight=".2" android:id="@+id/screen_login_banner_pane">
+            android:layout_height="0px"
+            android:layout_weight=".2"
```
ctsims: do we think this is always the right weight? might be useful to think about a max_height or min_height depending on the screen size.

app/res/values/styles.xml
```diff
@@ -137,4 +137,33 @@
         <item name="android:textColor">@color/white</item>
     </style>
 
-</resources>
\ No newline at end of file
+    <style name="ListViewV2">
```
ctsims: Is there a better descriptor than V2 for these? Most of these before didn't actually have V1's, did they? I'm fine with just naming them, although context would also be helpful if they're only used in one specific place currently

app/src/org/commcare/dalvik/activities/LoginActivity.java
```diff
@@ -52,14 +55,14 @@
...
-    
+
     @UiElement(value=R.id.login_button, locale="login.button")
     Button login;
     
-    @UiElement(value=R.id.text_username, locale="login.username")
-    TextView userLabel;
-    @UiElement(value=R.id.text_password, locale="login.password")
-    TextView passLabel;
+//    @UiElement(value=R.id.text_username, locale="login.username")
```
ctsims: Can just delete these entirely.

app/res/values/styles.xml
dcluna: Maybe we should hold renaming (and doing other small code fixes) until we merge everything else? This style is used in a lot of places in the redesigned code, and I'd prefer if we avoid changing it for now.

app/src/org/commcare/dalvik/activities/LoginActivity.java
```diff
@@ -87,6 +110,7 @@
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         username.setInputType(InputType.TYPE_TEXT_FLAG_NO_SUGGESTIONS | InputType.TYPE_TEXT_VARIATION_VISIBLE_PASSWORD);
+        editTextColor = username.getCurrentTextColor();
```
wpride: recovering a color from a UI element seems like an odd pattern - are we basically using this to keep track of normal vs. red after bad login?

app/res/layout/screen_login.xml
```diff
@@ -51,60 +59,91 @@
...
+                    style="@style/LoginEditTextV2"
+                    android:hint="Password"
+                    android:textSize="@dimen/text_medium"
+                    />
+
+                <TextView
+                    android:id="@+id/screen_login_bad_password"
+                    android:layout_width="fill_parent"
+                    android:layout_height="wrap_content"
+                    android:text="We couldn't find a user with this password. Please try another!"
```
wpride: Is this localizable somewhere?

app/res/layout/screen_login.xml
```diff
@@ -51,60 +59,91 @@
...
+                <EditText
+                    android:id="@+id/edit_username"
+                    android:layout_width="fill_parent"
+                    android:layout_height="wrap_content"
+                    android:inputType="text"
+                    android:nextFocusDown="@+id/edit_password" android:contentDescription="Username"
+                    android:drawableLeft="@drawable/icon_user_neutral50"
+                    android:background="@drawable/login_edit_text"
+                    style="@style/LoginEditTextV2"
+                    android:hint="Username"
```
wpride: What's the method for localizing hints?

app/res/layout/screen_login.xml
```diff
@@ -51,60 +59,91 @@
...
 
-            <TextView
-                android:id="@+id/screen_login_bad_password"
-                android:layout_width="fill_parent"
-                android:layout_height="wrap_content"
-                android:text="Incorrect Username or Password" android:background="@drawable/bubble_danger" android:paddingTop="2dp" android:paddingBottom="2dp" android:textColor="@color/red" android:textStyle="bold" android:visibility="gone" android:layout_marginLeft="2dp" android:paddingLeft="5dp" android:paddingRight="5dp" android:layout_marginRight="2dp" android:gravity="center"/>
+                <TextView
+                    android:layout_width="wrap_content"
+                    android:layout_height="wrap_content"
+                    android:text="Welcome back! Please log in."
```
wpride: No id here - localizable?

app/src/org/commcare/dalvik/activities/LoginActivity.java
dcluna: Yes, is there a better way to do it?

app/src/org/commcare/dalvik/activities/LoginActivity.java
```diff
@@ -446,13 +474,25 @@ private void raiseMessage(NotificationMessage message, boolean showTop) {
...
+
         errorBox.setVisibility(View.VISIBLE);
         errorBox.setText(toastText);
         
         Toast.makeText(this,toastText, Toast.LENGTH_LONG).show();
     }
-    
-    
+
+    private void setLoginBoxesColor(int colorId, boolean resolve) {
```
phillipm: 'resolve' is an uninformative argument name and without method documentation, I have no idea what it is used for?

app/src/org/commcare/dalvik/activities/LoginActivity.java
wpride: The other pattern would be having some state Enum that tracks whether we're in error mode or not

app/res/layout/screen_login.xml
ctsims: @dcluna Agreed. This should have an Id to be localizable

app/res/layout/screen_login.xml
```diff
@@ -51,60 +59,91 @@
...
+                    android:textColor="@color/cc_neutral_color"
+                    android:layout_gravity="center"
+                    android:layout_margin="16dp"
```
ctsims: This should be in a common styleset, right? 16dp margins and medium text size for a textview seem like they are likely to be applied broadly?

app/res/layout/screen_login.xml
ctsims: Probably manual, but should definitely be done

app/src/org/commcare/dalvik/activities/LoginActivity.java
```diff
@@ -97,11 +121,14 @@ public void onCreate(Bundle savedInstanceState) {
...
             }
         }
+
+        final Activity loginAct = this;
         
         login.setOnClickListener(new OnClickListener() {
 
             public void onClick(View arg0) {
                 errorBox.setVisibility(View.GONE);
+                ViewUtil.hideVirtualKeyboard(loginAct);
```
ctsims: You can use the pattern

LoginActivity.this
to get the pointer to the outer class here, which will remove the need for the final declartion above

app/res/layout/screen_login.xml
ctsims: Note: Luna is doing this in the code below. Can close out this comment

app/src/org/commcare/dalvik/activities/LoginActivity.java
```diff
@@ -74,11 +73,35 @@
...
+            public void beforeTextChanged(CharSequence s, int start, int count, int after) {
+            }
+
+            @Override
+            public void onTextChanged(CharSequence s, int start, int before, int count) {
+            }
+
+            @Override
+            public void afterTextChanged(Editable s) {
+                setLoginBoxesColor(editTextColor, false);
```
ctsims: i'd pull all of these into a helper method in the activity called "setStyleDefault" or something similar to make it more clear what exactly each element is doing and make it triggerable from outside of this context as well.

app/src/org/commcare/dalvik/activities/LoginActivity.java
```diff
@@ -446,13 +474,30 @@ private void raiseMessage(NotificationMessage message, boolean showTop) {
...
     }
-    
-    
+
+    /**
+     * Sets the login boxes (user/pass) to the given color.
+     * @param colorIdOrCode Color ID of resource or color code
+     * @param isColorId Calls getResources.getColor if set, uses the given colorIdOrCode as the color code if not
+     */
+    private void setLoginBoxesColor(int colorIdOrCode, boolean isColorId) {
```
ctsims: This seems a bit over-complicated to me as a signature, honestly. It seems likely that it's worth the hit to the external method to just move the "getResources()" call to the caller and unify what the method does.

app/src/org/commcare/dalvik/activities/LoginActivity.java
dcluna: Added in latest commit

app/res/layout/screen_login.xml
dcluna: https://github.com/dimagi/commcare-odk/blob/restyling/login-screen-only-merge/app/src/org/commcare/dalvik/activities/LoginActivity.java#L67

app/res/layout/screen_login.xml
dcluna: It wraps the CommCare banner, so it's readjusted automatically

PR #257, opened 2015-05-04
First login automatically logs the user out

phillipm:
Logging in for the first time fails silently and requires a second login.
This PR fixes that.

### DIFF
--- a/app/src/org/commcare/dalvik/odk/provider/InstanceProvider.java
+++ b/app/src/org/commcare/dalvik/odk/provider/InstanceProvider.java
@@ -222,6 +222,7 @@ public class InstanceProvider extends ContentProvider {
 
         SQLiteDatabase db = mDbHelper.getWritableDatabase();
         long rowId = db.insert(INSTANCES_TABLE_NAME, null, values);
+        db.close();
 
         if (rowId > 0) {
             Uri instanceUri = ContentUris.withAppendedId(InstanceColumns.CONTENT_URI, rowId);
@@ -341,6 +342,7 @@ public class InstanceProvider extends ContentProvider {
             default:
                 throw new IllegalArgumentException("Unknown URI " + uri);
         }
+        db.close();
 
         getContext().getContentResolver().notifyChange(uri, null);
         return count;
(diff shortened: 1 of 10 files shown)


app/src/org/commcare/dalvik/odk/provider/InstanceProvider.java
```diff
@@ -222,6 +222,7 @@ public Uri insert(Uri uri, ContentValues initialValues) {
 
         SQLiteDatabase db = mDbHelper.getWritableDatabase();
         long rowId = db.insert(INSTANCES_TABLE_NAME, null, values);
+        db.close();
```
phillipm: Trying to get rid of a database error that occurs on first login. Didn't work, but these db.close() statements should probably stay anyways, right?

app/src/org/commcare/dalvik/services/CommCareSessionService.java
```diff
@@ -351,13 +354,15 @@ public void registerFormSaveCallback(FormSaveCallback callbackObj) {
...
-     * database. Performs CommCareApplication logout to unbind its connection
-     * to this object. Launches CommCareHomeActivity upon completion.
+     * Closes the key pool and user database. Performs CommCareApplication
+     * logout to unbind its connection to this object.
+     *
+     * @param sessionExpired should the user be redirected to the login screen
+     *                       upon closing this session?
      */
-    public void finishLogout() {
+    public void closeSession(boolean sessionExpired) {
```
phillipm: Fixing double login: Made closeSession take an argument to determine whether to dispatch to login screen or not

app/src/org/commcare/dalvik/odk/provider/InstanceProvider.java
ctsims: as long as we're not using any db stuff between getting the database handle and using it.

Does getWritableDatabase cache its result? If so this might cause issues

phillipm: The calls to db.close() I added occur after the local var 'db' (loaded by a getWritableDatabase call) is done being used, so we should be fine.

various implementations of getWritableDatabase aren't caching anything, but the user database is caching the result of a call to getWritableDatabase, which might be throwing the startup error, if that user db isn't eventually closed correctly.

PR #259, opened 2015-05-05
Change default session length from 8 to 24 hrs

phillipm:
Change the default app session length from 8 hrs (28800 seconds) to 24 hrs (86400 seconds)

### DIFF
--- a/app/src/org/commcare/dalvik/preferences/CommCarePreferences.java
+++ b/app/src/org/commcare/dalvik/preferences/CommCarePreferences.java
@@ -239,9 +239,9 @@ public class CommCarePreferences extends PreferenceActivity implements OnSharedP
 
         // default to 24 hours
         try {
-            return Integer.parseInt(properties.getString(LOGIN_DURATION, "28800"));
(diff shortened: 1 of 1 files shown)


snopoke:
@kaapstorm FYI

app/src/org/commcare/dalvik/preferences/CommCarePreferences.java
```diff
@@ -239,9 +239,9 @@ public static int getLoginDuration() {
 
         // default to 24 hours
         try {
-            return Integer.parseInt(properties.getString(LOGIN_DURATION, "28800"));
+            return Integer.parseInt(properties.getString(LOGIN_DURATION, "86400"));
```
ctsims: FWIW:

I prefer setting these constants with something like

24 * 60 * 60

instead of a concrete value, makes it a lot easier for someone reading it to know what the value is

kaapstorm: +1

phillipm: Good call!
https://github.com/dimagi/commcare-odk/pull/276

PR #262, opened 2015-05-05
Keyboard action fix 21

wpride:


wpride:
@ctsims another option is to just remove that onClick behavior in StringWidget that resets the current IMEoptions

mAnswer.setImeOptions(EditorInfo.IME_ACTION_UNSPECIFIED);

Is there a reason for this call to be in there?

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -572,12 +572,10 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
                 for (String key : callout.getResponses()) {
                     result = intent.getExtras().getString(key);
-                    if (result != null) {
-                        session.setDatum(key,result);
-                        if(!resultSet) {
-                            resultSet = true;
-                            this.searchbox.setText(result);
-                        }
+                    if (result != null && !resultSet) {
+                        resultSet = true;
```
phillipm: Is there a reason that this code made it into this PR?

wpride: only because I noticed that the previous datum code was in here (when it shouldn't have been)

ctsims: Will:

Please don't do this if at all possible, or if you do make sure it is noted in the PR description. We use the format to do things like collect notes for bugfixes, so including code for other areas without noting it makes pushes easy to miss.

PR #263, opened 2015-05-05
Restyling/install screen

dcluna:
Redesigned install screens.


app/src/org/commcare/android/fragments/SetupEnterURLFragment.java
```diff
@@ -0,0 +1,85 @@
...
+import android.widget.Spinner;
+import android.widget.TextView;
+
+import org.commcare.android.framework.ManagedUi;
+import org.commcare.android.framework.UiElement;
+import org.commcare.dalvik.R;
+import org.javarosa.core.services.locale.Localization;
+
+/**
+ * Created by dancluna on 3/17/15.
```
phillipm: Would be nice to include some documentation on what this class does.
Something simple like "Activity for inputting app installation URL; gives the user the ability to enter a bitly, .., or raw URL.

app/src/org/commcare/android/fragments/SetupEnterURLFragment.java
```diff
@@ -0,0 +1,85 @@
...
+            throw new ClassCastException(activity + " must implemement " + interfaceName);
+        } else {
+            listener = (URLInstaller) activity;
+        }
+    }
+
+    public String getURL(){
+        int selectedPrefix = prefixURLSpinner.getSelectedItemPosition();
+        String url = profileLocation.getText().toString();
+        if(url == null || url.length() == 0) return url;
```
phillipm: One-line conditional statements without braces are bug prone. Please wrap such occurrences in braces

app/src/org/commcare/android/fragments/SetupEnterURLFragment.java
```diff
@@ -0,0 +1,85 @@
...
+    public void onAttach(Activity activity) {
+        super.onAttach(activity);
+        if(!(activity instanceof URLInstaller)){
+            throw new ClassCastException(activity + " must implemement " + interfaceName);
+        } else {
+            listener = (URLInstaller) activity;
+        }
+    }
+
+    public String getURL(){
```
phillipm: Documentation telling us where the URL value is read from and how it is going to be formatted before being returned (like the method logic that automatically prefixing with "http://")

app/src/org/commcare/android/view/SquareButtonWithText.java
```diff
@@ -0,0 +1,104 @@
...
+
+    public void setColor(int backgroundColor) {
+        squareButton.setBackgroundResource(backgroundColor);
+    }
+
+    public void setTextColor(int textColor) { textView.setTextColor(getResources().getColor(textColor)); }
+
+    //endregion
+
+    public SquareButtonWithText(Context context) {
```
phillipm: Constructors should be the first methods to appear in a class.

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -57,8 +47,8 @@
  * @author ctsims
  *
  */
-@ManagedUi(R.layout.first_start_screen)
-public class CommCareSetupActivity extends CommCareActivity<CommCareSetupActivity> implements ResourceEngineListener{
+@ManagedUi(R.layout.first_start_screen_modern)
+public class CommCareSetupActivity extends CommCareActivity<CommCareSetupActivity> implements ResourceEngineListener, SetupEnterURLFragment.URLInstaller {
```
phillipm: This refactor is awesome! I was just trying to clean up this file myself and you breaking things up into fragments makes it so much better!

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -69,6 +59,14 @@
     public static final String KEY_AUTO = "is_auto_update";
     public static final String KEY_START_OVER = "start_over_uprgrade";
     public static final String KEY_LAST_INSTALL = "last_install_time";
+
+    @Override
+    public void OnURLChosen(String url) {
```
phillipm: Sandwiching a method in-between class fields is a bad place for a method. Please move this somewhere below the class constructor.

app/src/org/commcare/android/fragments/SetupEnterURLFragment.java
```diff
@@ -0,0 +1,85 @@
...
+import org.javarosa.core.services.locale.Localization;
+
+/**
+ * Created by dancluna on 3/17/15.
+ */
+@ManagedUi(R.layout.setup_enter_url)
+public class SetupEnterURLFragment extends Fragment {
+
+    public interface URLInstaller {
+        public void OnURLChosen(String url);
```
phillipm: Why is this method name capitalized?

phillipm: Please documentation to all interface method declarations; there is no code to give context as to what this method should be doing.

phillipm:
Noting that I made a review pass on the Java portions a few weeks.
Thanks for updating the PR based on my comments, looks good!

app/res/layout/setup_enter_url.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="utf-8"?>
+<RelativeLayout
+    android:layout_height="match_parent"
+    android:layout_width="match_parent"
+    xmlns:android="http://schemas.android.com/apk/res/android">
+
+        <TextView
+            android:id="@+id/TextView02"
```
wpride: bad name

app/src/org/commcare/android/fragments/SetupEnterURLFragment.java
```diff
@@ -0,0 +1,97 @@
...
+import org.commcare.android.framework.UiElement;
+import org.commcare.dalvik.R;
+import org.javarosa.core.services.locale.Localization;
+
+/**
+ * Fragment for inputting app installation URL, "returned" through the URLInstaller interface.
+ * Created by dancluna on 3/17/15.
+ */
+@ManagedUi(R.layout.setup_enter_url)
+public class SetupEnterURLFragment extends Fragment {
```
wpride: Fragments!

app/src/org/commcare/android/fragments/SetupEnterURLFragment.java
```diff
@@ -0,0 +1,97 @@
...
+        installButton.setText(Localization.get("install.button.start"));
+        prefixURLSpinner = (Spinner) view.findViewById(R.id.url_spinner);
+        profileLocation = (EditText) view.findViewById(R.id.edit_profile_location);
+        final TextView appProfile = (TextView) view.findViewById(R.id.TextView02);
+        appProfile.setText(Localization.get("install.appprofile"));
+
+        installButton.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(final View v) {
+                getFragmentManager().popBackStack();
```
wpride: what's this do?

app/src/org/commcare/android/fragments/SetupInstallFragment.java
```diff
@@ -0,0 +1,66 @@
...
+import android.widget.Toast;
+
+import org.commcare.android.framework.UiElement;
+import org.commcare.android.view.SquareButtonWithText;
+import org.commcare.dalvik.R;
+import org.commcare.dalvik.activities.CommCareSetupActivity;
+import org.javarosa.core.services.locale.Localization;
+
+/**
+ * Created by dancluna on 3/17/15.
```
wpride: class level comment

app/src/org/commcare/android/fragments/SetupKeepInstallFragment.java
```diff
@@ -0,0 +1,61 @@
...
+import android.view.ViewGroup;
+import android.widget.TextView;
+
+import org.commcare.android.framework.UiElement;
+import org.commcare.android.view.SquareButtonWithText;
+import org.commcare.dalvik.R;
+import org.javarosa.core.services.locale.Localization;
+
+/**
+ * Created by dancluna on 4/10/15.
```
wpride: class level comment

app/src/org/commcare/android/fragments/SetupKeepInstallFragment.java
```diff
@@ -0,0 +1,61 @@
...
+
+    public void setButtonCommands(final StartStopInstallCommands buttonCommands) {
+        this.buttonCommands = buttonCommands;
+    }
+
+    @Override
+    public View onCreateView(final LayoutInflater inflater, final ViewGroup container, final Bundle savedInstanceState) {
+        final View view = inflater.inflate(R.layout.setup_keepinstall, container, false);
+        btnStartInstall = (SquareButtonWithText) view.findViewById(R.id.btn_start_install);
+        btnStartInstall.setText(Localization.get("install.button.start"));
```
wpride: @ctsims not sure but wouldn't declaring these buttons as UIElements with localization ids handle this for you?

app/src/org/commcare/android/fragments/SetupKeepInstallFragment.java
```diff
@@ -0,0 +1,61 @@
...
+ * Created by dancluna on 4/10/15.
+ */
+public class SetupKeepInstallFragment extends Fragment {
+    @UiElement(R.id.btn_start_install)
+    SquareButtonWithText btnStartInstall;
+
+    @UiElement(R.id.btn_stop_install)
+    SquareButtonWithText btnStopInstall;
+
+    public interface StartStopInstallCommands {
```
wpride: So with this and the URLInstaller interface above we're trying to have an affordance to implement other types of installers? Is there a reason these two interfaces shouldn't be merged?

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
wpride: +1

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -190,200 +159,89 @@ public void onCreate(Bundle savedInstanceState) {
...
-                startActivity(i);
-                finish();
-            }
+        uiStateScreenTransition();
+    }
 
-        });
+    @Override
+    public void OnURLChosen(String url) {
+        Log.d("DEBUG-d", "SetupEnterURLFragment returned: " + url);
```
wpride: I think we want Logger.log here to reach the server

app/src/org/commcare/android/fragments/SetupKeepInstallFragment.java
dcluna: I think UIElement does not work with Fragments (yet)

app/src/org/commcare/android/fragments/SetupKeepInstallFragment.java
dcluna: The URLInstaller interface has a bad name: it actually is used only when the user chooses an URL through the first fragment ("scan app barcode" or "enter url"), while the StartStopInstallCommands interface is for the start/stop install screen.

app/res/layout/setup_enter_url.xml
ctsims: @dcluna agreed.

app/res/layout/setup_install.xml
```diff
@@ -0,0 +1,86 @@
...
+                android:layout_height="fill_parent"
+                android:orientation="vertical"
+                android:background="@color/cc_core_bg"
+    >
+    <LinearLayout
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:orientation="vertical">
+
+        <!--&lt;!&ndash; TODO: extract style definition &ndash;&gt;-->
```
ctsims: Is this hard to do in this state? It would be good to be getting as much of the style definition stuff out of here as possible.

app/res/layout/setup_install.xml
```diff
@@ -0,0 +1,86 @@
+<?xml version="1.0" encoding="utf-8"?>
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
+                xmlns:SquareButtonWithText="http://schemas.android.com/apk/res/org.commcare.dalvik"
+                android:id="@+id/screen_first_start_main"
+                android:layout_width="fill_parent"
+                android:layout_height="fill_parent"
+                android:orientation="vertical"
+                android:background="@color/cc_core_bg"
```
ctsims: We can probably set this in the Theme, btw, so we don't need to set it on each layout host.

app/res/layout/setup_install.xml
```diff
@@ -0,0 +1,86 @@
+<?xml version="1.0" encoding="utf-8"?>
```
ctsims: We can't use folders, so I'd like to make sure we're namespacing these components with prefixes like

fragment_setup_install

app/res/layout/setup_keepinstall.xml
```diff
@@ -0,0 +1,111 @@
...
+                </LinearLayout>
+                <LinearLayout
+                    android:layout_width="match_parent"
+                    android:layout_height="wrap_content">
+                    <ImageView
+                        android:layout_width="wrap_content"
+                        android:layout_height="wrap_content"
+                        android:src="@drawable/icon_wifi_neutral"
+                        android:layout_gravity="left"
+                        android:layout_marginLeft="32dp"
```
ctsims: Raw dimensions should be in a stlye or in dimens.xml

app/res/layout/setup_keepinstall.xml
```diff
@@ -0,0 +1,111 @@
...
+                    android:background="@color/cc_core_bg"
+        >
+        <LinearLayout
+            android:layout_width="match_parent"
+            android:layout_height="match_parent"
+            android:orientation="vertical">
+
+            <!--&lt;!&ndash; TODO: extract style definition &ndash;&gt;-->
+            <ImageView
+                android:id="@+id/screen_first_start_banner"
```
ctsims: It really seems like these two fragments are very similar, and that it's possible that there should be one host fragment (with the banner and stuff) and one internal fragment with the buttons, which would make this workflow a bit easier to extend. 

Might seem less relevant once the dimensions and stuff are pulled out a bit, though?

app/src/org/commcare/android/fragments/SetupEnterURLFragment.java
```diff
@@ -0,0 +1,97 @@
...
+                if(listener != null) listener.OnURLChosen(getURL());
```
ctsims: onAttach is always called before onCreateView, so you don't need to provide the listener null check here as long as the fragment will fail to attach unless its parent is of the right type.

app/src/org/commcare/android/fragments/SetupKeepInstallFragment.java
```diff
@@ -0,0 +1,61 @@
...
+        void onStartInstallClicked();
+        void onStopInstallClicked();
+    }
+
+    StartStopInstallCommands buttonCommands;
```
ctsims: this uses a different callback pattern than the other fragment (which treats its activity as the communication interface and configues the callback through that). Is there a reason they are done differently? Would be good for things to follow the same pattern if possible.

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
ctsims: +1

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
ctsims: I think this seems like normal command-line debugging. Log.d looks good to me.

app/res/layout/setup_keepinstall.xml
dcluna: The second fragment (start/stop install screen) has a few differences (header and footer text) that makes it easier to just have 2 different screens

app/src/org/commcare/android/fragments/SetupKeepInstallFragment.java
dcluna: Updated to use the same pattern (checking the parent Activity implemented it when attaching the fragment)

app/res/layout/setup_install.xml
dcluna: Done

app/res/layout/setup_install.xml
dcluna: Probably hard to reuse the style in this element, despite the comment

PR #266, opened 2015-05-06
Null pointer exception fix on form crash 

phillipm:


app/AndroidManifest.xml
```diff
@@ -2,7 +2,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="org.commcare.dalvik"
     android:versionCode="106"
-    android:versionName="2.22" >
```
wpride: probably don't want this change

app/res/values/strings.xml
```diff
@@ -15,7 +15,7 @@
 
     <integer-array name="commcare_version">
         <item>2</item>
-        <item>22</item>
```
wpride: same here

PR #268, opened 2015-05-06
Home activity shouldn't exit after returning from Validating Media

phillipm:


ctsims:
Is this after validating multimedia was successful?

The code path here currently is:

Home Activity -> Detect Multimedia validation needed -> Launch Validation Activity

If the validation is successful we return with "OK" and proceed.

If it _isn't_ the only thing the user can do is cancel, which exits the app (which is the desired behavior).

What issue is this addressing?

ctsims:
adding don't pull warning until question resolvede

PR #271, opened 2015-05-06
Revert "Removed old odk provider files"

ctsims:
Reverts dimagi/commcare-odk#269


ctsims:
@phillipm This broke the build so reverted it

PR #277, opened 2015-05-07
Form cancel should redirect to form selection

phillipm:
Erroring-out of form entry took the user back to the module selection screen, whereas it should take one back to the form selection screen. 

[ticket](http://manage.dimagi.com/default.asp?167803#935571)


ctsims:
Wait,

Phillip: Was the last thing the user did before entering the form select a
case?

The correct workflow should be that cancelling a form takes you back to the
last activity you were using before the form entry started...

phillipm:
@ctsims In the example that I tested, no cases were selected. I neglected to test escaping out of a form when a case has been selected.

PR #279, opened 2015-05-07
CommCareApplication refactor

phillipm:
- Made nested inner class static and make use of weak references to avoid memory leaks
 - Made lock synchronization objects final
 - Removed override methods that just called their supers
 - Made string comparison not use ==/!= !!
 - Removed a couple of functions that weren't used
 - Removed a method param that was unused in the body

Requires https://github.com/dimagi/commcare-odk/pull/282 to be merged first


app/src/org/commcare/dalvik/application/CommCareApplication.java
```diff
@@ -967,20 +974,6 @@ public CommCareSessionService getSession() throws SessionUnavailableException {
...
-        /*
-         * (non-Javadoc)
-         * @see android.os.Handler#handleMessage(android.os.Message)
-         */
-        @Override
-        public void handleMessage(Message m) {
-            NotificationMessage message = m.getData().getParcelable("message");
-            Toast.makeText(CommCareApplication.this,
-                    Localization.get("notification.for.details.wrapper", new String[]{message.getTitle()}),
-                    Toast.LENGTH_LONG).show();
```
phillipm: This was getting a memory leak lint warning due to the fact that an inner class maintains a reference to its parent class, hence if the Handler object remains around (which happens until messages linked to it are processed) and the context is no longer used, the context will not be garbage collected. 

I know Application objects live for the span of the app, so this is sort of pointless, but I figured I'd still refactor it.

app/src/org/commcare/dalvik/application/CommCareApplication.java
```diff
@@ -776,7 +782,7 @@ private boolean getPendingUpdateStatus() {
         boolean autoUpdateTriggered = preferences.getBoolean(CommCarePreferences.AUTO_TRIGGER_UPDATE, false);
 
         if (autoUpdateTriggered) {
-            preferences.edit().putBoolean(CommCarePreferences.AUTO_TRIGGER_UPDATE, false);
+            preferences.edit().putBoolean(CommCarePreferences.AUTO_TRIGGER_UPDATE, false).commit();
```
phillipm: Missing commit means this update was never saved.
Is adding this commit going to break something?

PR #281, opened 2015-05-08
Update activity_report_problem.xml

wpride:
re: http://manage.dimagi.com/default.asp?167865


app/res/layout/activity_report_problem.xml
```diff
@@ -34,7 +34,7 @@
         android:gravity="top"
         android:hint="@string/your_comment"
         android:inputType="textMultiLine"
-        android:lines="5"
+        android:maxLines="5"
```
phillipm: If you take out 'lines', then the text box can expand and cover up the button.
Changing it to maxLines makes it inflate up to only 5 lines tall

PR #282, opened 2015-05-08
Remove pref edits that were never committed

phillipm:
@wspride in commit f8928c70076b9b1a9dafb79eef17420815820647 , you added the following logic that I am now asking to be removed. That commit edited some preferences but never committed them by calling '.commit()'. This means that AUTO_TRIGGER_UPDATE was always set to false, making the statements removed in the PR no-ops.

I initially added '.commit()' back in but it broke things, so I decided to remove the code, since it is unclear what the intent of the logic was.


wpride:
@phillipm haha yikes... I wasn't always the coder I am today. Removed the (now unused) pref string too

PR #284, opened 2015-05-11
remove form entry long click

wpride:


wpride:
http://manage.dimagi.com/default.asp?123661

phillipm:
Probably also worth taking out the onContextItemSelected and createClearDialog methods because all they do is enable the clear answer functionality you remove in this PR. At least I think I'm correct in making this statement; worth a double check.

PR #286, opened 2015-05-12
Save on session expiration when constraints are violated

phillipm:
Saving on session expiration currently crashes the app if the form being edited violates constraints. This PR just aborts the save if constraints aren't met. We could also attempt to clear violating values and resave, should this be done?

[ticket](http://manage.dimagi.com/default.asp?168163#939208)

### DIFF
--- a/app/src/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/app/src/org/odk/collect/android/activities/FormEntryActivity.java
@@ -512,6 +512,7 @@ public class FormEntryActivity extends FragmentActivity implements AnimationList
         }
     }
 
+    @Override
     public void formSaveCallback() {
         // note that we have started saving the form
         savingFormOnKeySessionExpiration = true;
@@ -564,7 +565,7 @@ public class FormEntryActivity extends FragmentActivity implements AnimationList
                 if (!"done".equals(v.getTag())) {
                     FormEntryActivity.this.showNextView();
                 } else {
-                    FormEntryActivity.this.triggerUserFormComplete();
+                    saveDataToDisk(EXIT, true, getDefaultFormTitle(), false);
                 }
             }
         });
@@ -1587,8 +1588,10 @@ public class FormEntryActivity extends FragmentActivity implements AnimationList
                                         Toast.makeText(FormEntryActivity.this, StringUtils.getStringSpannableRobust(FormEntryActivity.this, R.string.save_as_error),
                                                 Toast.LENGTH_SHORT).show();
                                     } else {
-                                        saveDataToDisk(EXIT, instanceComplete.isChecked(), saveAs
-                                                .getText().toString(), false);
+                                        saveDataToDisk(EXIT,
+                                                       instanceComplete.isChecked(),
+                                                       saveAs.getText().toString(),
+                                                       false);
                                     }
                                 }
                             });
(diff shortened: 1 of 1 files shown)


app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2085,14 +2100,16 @@ public void onClick(DialogInterface dialog, int i) {
...
      * @param headless is this running as a GUI-less service
      *
      * @return Did the data save successfully?
      */
-    private boolean saveDataToDisk(boolean exit, boolean complete, String updatedSaveName, boolean headless) {
+    private boolean saveDataToDisk(boolean exit, boolean complete,
+                                   String updatedSaveName, boolean headless) {
         // save current answer
-        if (!saveAnswersForCurrentScreen(EVALUATE_CONSTRAINTS, complete)) {
+        if (!saveAnswersForCurrentScreen(EVALUATE_CONSTRAINTS, complete, headless)) {
```
phillipm: Session-expiration form saving will abort if the current screen has a value that violates the constraint. We could try to roll back that violating value and save, should we attempt this?

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -1309,56 +1309,71 @@ private boolean currentPromptIsQuestion() {
...
-                                                    (failOnRequired || saveStatus != FormEntryController.ANSWER_REQUIRED_BUT_EMPTY))) {
-                            createConstraintToast(index, mFormController.getQuestionPrompt(index) .getConstraintText(), saveStatus, success);
+                        int saveStatus = saveAnswer(answers.get(index),
+                                index, evaluateConstraints);
+                        if (evaluateConstraints &&
+                                ((saveStatus != FormEntryController.ANSWER_OK) &&
+                                        (failOnRequired ||
+                                                saveStatus != FormEntryController.ANSWER_REQUIRED_BUT_EMPTY))) {
+                            if (!headless) {
+                                createConstraintToast(index, mFormController.getQuestionPrompt(index).getConstraintText(), saveStatus, success);
```
phillipm: Don't show gui warnings if we are running in the background.

ctsims:
When we save forms as incomplete there is a flag that allows the data model to be serialized violating the overall constraints, although not the constraints of the current input.

I'm totally fine with zeroing questions which violate constraint conditions if that's easy enough to do from the widget ("required" constraints won't prevent saving, so it should be fine if they are empty)

phillipm:
Looks like we can call saveAnswersForCurrentScreen with the DO_NOT_EVALUATE_CONSTRAINTS arg, letting the current answer be saved even if it is invalid. I've pushed some code that enables this.

@ctsims ready to be looked over again.

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -564,7 +565,7 @@ public void onClick(View v) {
                 if (!"done".equals(v.getTag())) {
                     FormEntryActivity.this.showNextView();
                 } else {
-                    FormEntryActivity.this.triggerUserFormComplete();
+                    saveDataToDisk(EXIT, true, getDefaultFormTitle(), false);
```
ctsims: Hey,

So we used the old statement (triggerUserFormComplete) to denote the distinction between the form saving for an arbitrary reason with user input that means "I want the form to exit", which has some important distinctions, UI wise. I don't think they're relevant without some new UI stuff, but I think it would be good to maintain them

phillipm: @ctsims Okay, I've reverted back to including triggerUserFormComplete

PR #288, opened 2015-05-13
Display saved forms incrementally; only call load once.

phillipm:
The 'Saved Form Listing' view was trying to load saved forms more than once and only showing results once everything had loaded. 

This PR makes sure forms loading doesn't occur more than once per view entry and enables incremental results to be displayed.

Hopefully helpful for this [ticket](http://manage.dimagi.com/default.asp?167729)

### DIFF
--- a/app/src/org/commcare/android/adapters/IncompleteFormListAdapter.java
+++ b/app/src/org/commcare/android/adapters/IncompleteFormListAdapter.java
@@ -36,35 +36,49 @@ import android.widget.BaseAdapter;
  *
  */
 public class IncompleteFormListAdapter extends BaseAdapter implements FormRecordLoadListener {
+    private final Context context;
     
-    private AndroidCommCarePlatform platform;
-    private Context context;
+    private final List<DataSetObserver> observers;
     
-    List<DataSetObserver> observers;
-    
-    FormRecordFilter filter;
-    // loaded form records, before filtering occurs
+    private FormRecordFilter filter;
+
+    /**
+     * All loaded form records of a given status, with no filtering.
+     */
     private List<FormRecord> records;
-    // filtered form records
-    private List<FormRecord> current = new ArrayList<FormRecord>();
 
-    // Maps FormRecord ID to an array of text that will be shown to the user
-    // and query-able. Text should includes modified date, record title, & form
-    // name.
-    private Hashtable<Integer, String[]> searchCache;
+    /**
+     * Filtered form records. getView reads from this to display all the forms.
+     */
+    private final List<FormRecord> current = new ArrayList<FormRecord>();
 
-    // last query made, used to filter records
-    private String query;
+    /**
+     * Maps FormRecord ID to an array of text that will be shown to the user
+     * and query-able. Text should includes modified date, record title, & form
+     * name.
+     */
+    private final Hashtable<Integer, String[]> searchCache = new Hashtable<Integer, String[]>();
 
-    FormRecordLoaderTask loader;
+    /**
+     * The last query made, used to filter forms.
+     */
+    private String query = "";
 
-    // Maps form namespace (unique id for forms) to their form title
-    // (entry-point text). Needed because FormRecords don't have form title
-    // info, but do have the namespace.
-    Hashtable<String, Text> names;
+    /**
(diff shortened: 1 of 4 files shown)


app/src/org/commcare/dalvik/activities/FormRecordListActivity.java
```diff
@@ -176,6 +176,7 @@ public void onCreate(Bundle savedInstanceState) {
                 ArrayAdapter<String> spinneritems = new ArrayAdapter<String>(this, R.layout.form_filter_display, names);
                 filterSelect.setAdapter(spinneritems);
                 spinneritems.setDropDownViewResource(R.layout.form_filter_item);
+                filterSelect.setSelection(0, false);
```
phillipm: Prevents onItemSelected from firing on spinner initialization, which calls resetRecords unnecessarily, causing a complete form record reload.

app/src/org/commcare/android/tasks/FormRecordLoaderTask.java
```diff
@@ -238,7 +238,7 @@ protected void onPostExecute(Integer result) {
      * @see android.os.AsyncTask#onProgressUpdate(Progress[])
      */
     @Override
-    protected void onProgressUpdate(Pair<Integer, ArrayList<String>>... values) {
+    protected void onProgressUpdate(Pair<FormRecord, ArrayList<String>>... values) {
```
phillipm: Changed Integer (FormRecord ID) argument to the actual FormRecord so that IncompleteFormListAdapter.notifyPriorityLoaded can incrementally add the record to its list of current forms (if it satisfies the current query).

app/src/org/commcare/android/adapters/IncompleteFormListAdapter.java
```diff
@@ -88,25 +102,38 @@ public IncompleteFormListAdapter(Context context, AndroidCommCarePlatform platfo
...
-     * Unused since priority loading logic is currently too messy/broken.
+     * Add a newly-loaded form to the current list if it satisfies the current
+     * query.
      */
-    public void notifyPriorityLoaded(Integer first, boolean contains) {
-        return;
+    @Override
+    public void notifyPriorityLoaded(FormRecord record, boolean isLoaded) {
+        if (isLoaded && satisfiesQuery(record)) {
+            current.add(record);
```
phillipm: Allows for showing results incrementally as they load.

app/src/org/commcare/dalvik/activities/FormRecordListActivity.java
```diff
@@ -177,17 +173,18 @@ public void onCreate(Bundle savedInstanceState) {
...
-                        adapter.setFormFilter(FormRecordFilter.values()[index]);
-                        adapter.resetRecords();
-                        adapter.notifyDataSetChanged();
-                        
+                        // NOTE: This gets called every time a spinner gets
+                        // set-up and also every time spinner state is restored
+                        // on scree-rotation. Hence we defer onCreate record
+                        // loading until this gets triggered automatically.
+                        adapter.setFilterAndResetRecords(FormRecordFilter.values()[index]);
+
```
phillipm: onItemSelected gets called every time this activity is setup/rotated. Use this fact to launch record loading. It might get called multiple times if restoring a previous selection, so I've included logic in the method making sure record reloading only occurs when the filter actually changes.

phillipm:
Tested this PRs changes on a 4.2.2 tablet with 300 forms on it.
Form loading is _slightly_ more responsive, and loading results are shown incrementally.

app/src/org/commcare/android/adapters/IncompleteFormListAdapter.java
```diff
@@ -306,25 +334,41 @@ private void filterValues() {
 
         current.clear();
 
-        if (query == null || query.equals("")) {
+        if ("".equals(query)) {
```
wpride: you've changed the truth value here, just double checking that that was intentional

phillipm: It was intentional: I made sure that 'query' was initialized to the empty string, so it should never be set to null, unless Editable.toString ever returns null, which I don't suspect it should.

PR #290, opened 2015-05-13
Catch missing fixture

orangejenny:
http://manage.dimagi.com/default.asp?168200


app/src/org/commcare/android/framework/CommCareActivity.java
```diff
@@ -701,6 +703,43 @@ public Integer getProgress() {
         }
         return null;
     }
+
+    protected void createErrorDialog(String errorMsg, boolean shouldExit) {
```
orangejenny: This protected method + public static method pattern isn't lovely, but FormEntryActivity and EntitySelectActivity don't have a common ancestor, and it beats code duplication.

orangejenny:
oops - 2.21, not master

PR #291, opened 2015-05-13
Catch missing fixture

orangejenny:
http://manage.dimagi.com/default.asp?168200

### DIFF
--- a/app/src/org/commcare/android/framework/CommCareActivity.java
+++ b/app/src/org/commcare/android/framework/CommCareActivity.java
@@ -8,9 +8,11 @@ import org.commcare.android.tasks.templates.CommCareTask;
 import org.commcare.android.tasks.templates.CommCareTaskConnector;
 import org.commcare.android.util.MarkupUtil;
 import org.commcare.android.util.SessionUnavailableException;
+import org.commcare.android.util.StringUtils;
 import org.commcare.dalvik.application.CommCareApplication;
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/android/framework/CommCareActivity.java
```diff
@@ -701,6 +703,43 @@ public Integer getProgress() {
         }
         return null;
     }
+
+    protected void createErrorDialog(String errorMsg, boolean shouldExit) {
```
orangejenny: This protected method + public static method pattern isn't lovely, but FormEntryActivity and EntitySelectActivity don't have a common ancestor, and it beats code duplication.

app/src/org/commcare/android/framework/CommCareActivity.java
```diff
@@ -701,6 +703,43 @@ public Integer getProgress() {
...
+        createErrorDialog(this, errorMsg, shouldExit);
+    }
+    
+    /**
+     * Pop up a semi-friendly error dialog rather than crashing outright.
+     * @param activity Activity to which to attach the dialog.
+     * @param errorMsg
+     * @param shouldExit If true, cancel activity when user exits dialog.
+     */
+    public static void createErrorDialog(final Activity activity, String errorMsg, final boolean shouldExit) {
```
phillipm: Yay! Well formatted and no lint warnings!

PR #295, opened 2015-05-15
ManageAsyncTasks; enabling task cancellation on session expiration

phillipm:
Was experiencing a crash in EntityLoaderTask, where if the database closed when it was running, it would crash commcare. Fix this by canceling the task when the db is closed.

 - Create ManageAsyncTask to keep track of running tasks.
 - ManageAsyncTask.cancelTasks calls cancel on all the running tasks; call on session expiration
 - have CommCareTask extend it ManageAsyncTask


ctsims:
I like this idea in general. The Static set of Managed Tasks really worries me though.

Maybe the CommCareActivity task management system should be initialized with the current session service and register/manage callbacks through there? It's fairly robust WRT its hooks for things coming into/out of context.

PR #297, opened 2015-05-15
Increase form save logout timeout to 90 secs

phillipm:
When the session expires and a form tries to save, there is a timeout period, which when reached the form saving is aborted. Increase this timeout to 90 seconds because who knows how long it might take to save some huge form on a slow phone.


ctsims:
30 seconds is honestly a bit low still, for forms with really complex logic.

Can we make this like 90 seconds? That should work with even super advanced clincial forms

phillipm:
@ctsims Increased to 90 secs

PR #299, opened 2015-05-15
Wrap getSession calls in try/catch

phillipm:
Many places call getSession (the CommCareSessionService one, not the one with unfortunately the same name found in AndroidSessionWrapper). These calls usually occur when a task, that accesses the user database, is getting ready to spin up. 

This PR wraps almost all calls to getSession in a try/catch to handle when the session has expired and is no longer available. In most cases such failed calls to getSession will result in tasks not getting launched.

### DIFF
--- a/app/src/org/commcare/android/tasks/LogSubmissionTask.java
+++ b/app/src/org/commcare/android/tasks/LogSubmissionTask.java
@@ -215,7 +215,7 @@ public class LogSubmissionTask extends AsyncTask<Void, Long, LogSubmitOutcomes> 
             return false;
         }
 
-        if (user.getUserType().equals(User.TYPE_DEMO)) {
+        if (User.TYPE_DEMO.equals(user.getUserType())) {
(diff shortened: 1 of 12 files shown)


app/src/org/commcare/android/tasks/LogSubmissionTask.java
```diff
@@ -207,8 +207,15 @@ private boolean submit(DeviceReportRecord slr, int index) {
...
-        if(user.getUserType().equals(User.TYPE_DEMO)) {
+        User user = null;
+        try {
+            user = CommCareApplication._().getSession().getLoggedInUser();
+        } catch (SessionUnavailableException e) {
+            // Session probably expired, so we will continue submitting logs,
+            // pretending to be a demo user.
+        }
+
+        if (user == null || user.getUserType().equals(User.TYPE_DEMO)) {
```
phillipm: Is it valid to submit logs as demo user if the session isn't found? Or should this just abort if the session has expired?

PR #301, opened 2015-05-18
Add a lock for session expiration/form send synchronization

phillipm:
Wrap ProcessAndSendTask worker in session logout lock so that when it is running the session will block on closing.

This PR only half-solves the solution: I didn't wrap DataPullTask in the same lock because it is mixed up with login code and calls closeSession, which would cause a deadlock (I think).

Good enough for now...

I've commented the new lines, since re-indentation created a huge (mostly meaningless) diff

### DIFF
--- a/app/src/org/commcare/android/tasks/ProcessAndSendTask.java
+++ b/app/src/org/commcare/android/tasks/ProcessAndSendTask.java
@@ -21,6 +21,7 @@ import org.commcare.android.util.FormUploadUtil;
 import org.commcare.android.util.SessionUnavailableException;
 import org.commcare.dalvik.activities.LoginActivity;
 import org.commcare.dalvik.application.CommCareApplication;
+import org.commcare.dalvik.services.CommCareSessionService;
 import org.commcare.suite.model.Profile;
 import org.commcare.util.CommCarePlatform;
 import org.javarosa.xml.util.InvalidStructureException;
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/android/tasks/ProcessAndSendTask.java
```diff
@@ -111,219 +112,222 @@ public ProcessAndSendTask(Context c, String url, int sendTaskId, boolean inSyncM
...
-        for(int i = 0; i < records.length ; ++i ) {
-            //Assume failure
-            results[i] = FormUploadUtil.FAILURE;
-        }
-        //The first thing we need to do is make sure everything is processed,
-        //we can't actually proceed before that.
-        for(int i = 0 ; i < records.length ; ++i) {
-            FormRecord record = records[i];
+        // Don't allow session to expire while we are syncing
+        synchronized (CommCareSessionService.LOGOUT_LOCK) {
```
phillipm: Added this lock

app/src/org/commcare/dalvik/services/CommCareSessionService.java
```diff
@@ -370,61 +377,63 @@ public void unregisterFormSaveCallback() {
...
-            if (!isActive()) {
-                // Since both the FormSaveCallback callback and the maintenance
-                // timer might call this, only run if it hasn't been called
-                // before.
-                return;
-            }
-            key = null;
-            String msg = "Logging out service login";
+        synchronized (lock) {
+            synchronized (CommCareSessionService.LOGOUT_LOCK) {
```
phillipm: Added this sync

app/src/org/commcare/dalvik/services/CommCareSessionService.java
```diff
@@ -66,6 +66,13 @@
     private Date sessionExpireDate;
     
     private final Object lock = new Object();
+
+    /**
+     * Lock that must be obtained in order to close down the session.
+     * NOTE PLM: Worth considering eventually removing this and refactoring
+     * processes to be more easily interrupted/aborted.
+     */
+    public static final Object LOGOUT_LOCK = new Object();
```
phillipm: Added this

phillipm:
Not sure why but syncing/auto-logout hang when this PRs code is present.

PR #302, opened 2015-05-18
Restyling/Case search (with new action bar)

dcluna:
Redesigned case search and action bar


app/src/org/commcare/android/util/AndroidUtil.java
```diff
@@ -71,4 +79,51 @@ public static void initializeStaticHandlers() {
             return a;
         }
     }
+
+    public static void setClickListenersForEverything(Activity activity, ViewGroup v) {
```
phillipm: Heads up that this is also in the form entry PR

app/src/org/commcare/android/view/EntityDetailView.java
```diff
@@ -134,6 +138,27 @@ public EntityDetailView(Context context, CommCareSession session, Detail d, Enti
...
+                R.attr.entity_detail_odd_row_color,
+                R.attr.entity_detail_even_row_color
+        };
+        Resources.Theme theme = context.getTheme();
+        for (int i = 0; i < colorAttr.length; i++) {
+            TypedValue typedValue = new TypedValue();
+            theme.resolveAttribute(colorAttr[i], typedValue, true);
+            int color = typedValue.data;
+            if(i == 0) { oddRowColor = color; }
+            else { evenRowColor = color; }
```
phillipm: Yay, thanks for wrapping your conditional bodies!

phillipm:
The Java portions of this PR look good to me.

app/res/layout/component_entity_detail_item.xml
```diff
@@ -3,19 +3,23 @@
...
         android:layout_height="fill_parent"
-        android:background="@drawable/blue_tabbed_box"
-        android:gravity="top|left|center_vertical"
+        android:gravity="left|center_vertical"
         android:text="Text"
-        android:textAppearance="@style/EntityItemText"
+        android:textStyle="bold"
         android:typeface="normal" android:layout_weight="4" android:visibility="visible"/>
+    <!--android:background="@drawable/blue_tabbed_box"-->
+    <!--android:textAppearance="@style/EntityItemText"-->
```
ctsims: Is there a reason we didn't just update this style? It's still used somewhere else? 

Would be good to use a different one instead of custom

app/res/layout/component_entity_detail_item.xml
```diff
@@ -30,9 +34,10 @@
             android:layout_gravity="right|center_vertical"
             android:gravity="left|center_vertical"
             android:padding="15dp"
```
ctsims: this should be 16dp now, right? To make our padding/margin use consistent?

app/res/layout/entity_select_layout.xml
```diff
@@ -3,13 +3,25 @@
...
   android:layout_height="wrap_content" android:layout_width="fill_parent">
 
-
+<TextView
+    android:id="@+id/no_search_results"
+    android:layout_width="fill_parent"
+    android:layout_height="wrap_content"
+    android:layout_below="@+id/universal_frame_tile"
+    android:paddingLeft="16dp"
+    android:paddingRight="16dp"
```
ctsims: This should either be a style or at least

android:padding: FULL_PADDING_TEXT

or something, right?

app/res/menu/activity_report_problem.xml
```diff
@@ -1,4 +1,13 @@
 <menu xmlns:android="http://schemas.android.com/apk/res/android">
+    <item android:id="@+id/search_action_bar"
```
ctsims: Is this in the right place?

app/res/values-land/dimens.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <dimen name="divider_left_margin">112dp</dimen>
```
ctsims: This is an awfully specific somewhat random value. What is this used for?

app/res/values/attrs.xml
```diff
@@ -3,4 +3,9 @@
         <attr name="ratio_width" format="float" />
         <attr name="ratio_height" format="float" />
     </declare-styleable>
-</resources>
\ No newline at end of file
+    <attr name="entity_detail_even_row_color" format="reference"></attr>
```
ctsims: Nice. This seems like a good way to do this.

app/res/values/colors.xml
```diff
@@ -113,4 +113,4 @@
     <color name="cc_light_warm_accent_bg">#F8ECBD</color>
     <color name="cc_light_warm_accent_text">#685300</color>
 
-</resources>
\ No newline at end of file
+</resources>
```
ctsims: Might cause a lot of conflicts to have this changed. Is it easy to revert?

app/res/xml/entity_select_search.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<searchable xmlns:android="http://schemas.android.com/apk/res/android"
+            android:label="@string/app_name"
+            android:hint="Search for case">
```
ctsims: Is this localizable? Also: We should avoid using the word 'case' on search stuff as the word has no meaning to our end users.

app/src/org/commcare/android/framework/BreadcrumbBarFragment.java
```diff
@@ -82,12 +82,12 @@ public void onCreate(Bundle savedInstanceState) {
      * reference to the newly created Activity after each configuration change.
      */
     @Override
-    public void onAttach(Activity activity) {
+    public void onAttach(final Activity activity) {
```
ctsims: this makes things super hard to read

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -681,19 +696,59 @@ public boolean onCreateOptionsMenu(Menu menu) {
...
+                final int[] searchViewStyle = AndroidUtil.getThemeColorIDs(this, new int[]{R.attr.searchbox_action_bar_color});
+                final int id = searchView.getContext()
+                        .getResources()
+                        .getIdentifier("android:id/search_src_text", null, null);
+                final TextView textView = (TextView) searchView.findViewById(id);
+                textView.setTextColor(searchViewStyle[0]);
+            }
+
+            final View bottomSearchWidget = findViewById(R.id.searchfooter);
+            if(bottomSearchWidget != null) bottomSearchWidget.setVisibility(View.GONE);
```
ctsims: Don't we want to not-do-this if the build version is < honeycomb?

Also: I think we shouldn't check for null here. If the bottom search view isn't there because we changed the file, we should delete this code.

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -681,19 +696,59 @@ public boolean onCreateOptionsMenu(Menu menu) {
...
+        }
+
         return true;
     }
+
+    @SuppressWarnings("NewApi")
+    private CharSequence getSearchText(){
+        // not checking for build version because searchview will be null if not supported
+        if(searchView != null) return searchView.getQuery();
```
ctsims: What about < honeycomb when we have to keep using the old search bar?

ctsims:
Added a bunch of smaller things, but two major points we need to clear up before we can pull

1) The huge "Everything is final" refactor doesn't really seem like it should be in here. It makes the PR very hard to review and I don't think it's a desirable change for now

2) Pre-Honeycomb it seems like we still need to rely on the old search bar, but it doesn't look like that's enabled here.

app/res/values-land/dimens.xml
dcluna: This sets the left margin for the new divider in the case list screen

app/res/menu/activity_report_problem.xml
dcluna: Yes, it's in res/menu/, the report problem activity layout is under res/layout/

app/res/xml/entity_select_search.xml
dcluna: This element does not support an 'id' element

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
dcluna: If the build version is < honeycomb, we're inflating the new search bar and removing the bottom search bar.

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
dcluna: If the build version is < honeycomb, searchView will be null, thus falling back to the previous code (returning the bottom search bar's text).

app/res/layout/component_entity_detail_item.xml
dcluna: Yes, it's still used in a few places

app/res/values/colors.xml
dcluna: Reverted

PR #305, opened 2015-05-20
Make session only expire if syncing isn't occurring

phillipm:
Make ProcessAndSendTask and DataPullTask background tasks acquire a session logout lock so that the session can't close down when they are running.

Eventually when the task finishes then the session will be able to grab the lock and close down.


app/src/org/commcare/dalvik/services/CommCareSessionService.java
```diff
@@ -52,26 +53,34 @@
...
     private static final long MAINTENANCE_PERIOD = 1000;
-    // session length in MS
+
+    /**
+     * Session length in MS
+     */
     private static long sessionLength = 1000 * 60 * 60 * 24;
-    
+
+    public static final ReentrantLock logout_lock = new ReentrantLock();
```
wpride: what's the reason for the naming here?

PR #306, opened 2015-05-22
Check array size before removing questions and dividers

phillipm:
QA is experiencing a strange crash when session expiration tries to save a form that has questions that violate constraints.

I was unable to reproduce the bug, but worked off of the stack trace, found at http://manage.dimagi.com/default.asp?168163#945195

### DIFF
--- a/app/src/org/odk/collect/android/views/ODKView.java
+++ b/app/src/org/odk/collect/android/views/ODKView.java
@@ -197,11 +197,17 @@ public class ODKView extends ScrollView implements OnLongClickListener, WidgetCh
     }
     
     public void removeQuestionFromIndex(int i){
-        mView.removeView((View) widgets.get(i));
         int dividerIndex = Math.max(i - 1, 0);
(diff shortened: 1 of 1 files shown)


app/src/org/odk/collect/android/views/ODKView.java
```diff
@@ -197,11 +197,17 @@ public ODKView(Context context, FormEntryPrompt[] questionPrompts, FormEntryCapt
...
+
+        if (dividerIndex < dividers.size()) {
+            mView.removeView(dividers.get(dividerIndex));
+            dividers.remove(dividerIndex);
+        }
+
+        if (i < widgets.size()) {
+            mView.removeView((View) widgets.get(i));
+            widgets.remove(i);
+        }
```
phillipm: I assume that changing the order of operations here doesn't break anything.

wpride: OOC what situation is causing this to be called with an index that doesn't exist? Is there another bug further upstream?

PR #309, opened 2015-05-22
Proguard configuration to reduce jar size

dcluna:
Configuring proguard to reduce the app's jar size.


app/proguard.cfg
```diff
@@ -1,4 +1,4 @@
--optimizationpasses 5
+-optimizationpasses 15
```
wpride: this isn't going to make complilation take like three times longer is it? how long is each "pass"?

dcluna: Proguard [stops](http://proguard.sourceforge.net/manual/usage.html#optimizationoptions) when it cannot optimize anymore, so in practice you shouldn't see much difference from a 5-pass run.

ctsims:
Hey,

Two things:

1) This is going to make our stack traces unusable, right? Do we have a
plan for that

2) What's the impact ont he final jar size? If it's not _very_ substantial
I don't think it's gonna outweigh the first concern here.

-Clayton

PR #311, opened 2015-05-25
Remove TransactionParser's name & namespace fields

phillipm:
Since TransactionParser and its extensions didn't use the following, remove them:
 -  'parses' method
 - 'name' and 'namespace' fields

Change TransactionParserFactory's getParser to not take in a name or namespace field. These were rarely used, and in the cases that they were, the name/namespace values were obtained from the parser argument, and hence were redundant.

Cross requested with https://github.com/dimagi/commcare/pull/84

### DIFF
--- a/app/src/org/commcare/android/tasks/FormRecordCleanupTask.java
+++ b/app/src/org/commcare/android/tasks/FormRecordCleanupTask.java
@@ -68,8 +68,7 @@ public abstract class FormRecordCleanupTask<R> extends CommCareTask<Void, Intege
         this.platform = platform;
         this.taskId = taskId;
     }
-    
-    
+
     /*
      * (non-Javadoc)
      * @see org.commcare.android.tasks.templates.CommCareTask#doTaskBackground(java.lang.Object[])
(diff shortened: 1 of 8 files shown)


app/src/org/commcare/android/tasks/FormRecordCleanupTask.java
```diff
@@ -174,14 +173,17 @@ public static FormRecord getUpdatedRecord(Context context, CommCarePlatform plat
...
+                String name = parser.getName();
+
+                if (name == null) {
+                    return null;
+                }
 
-            public TransactionParser getParser(String name, String namespace, KXmlParser parser) {
-                if(name == null) { return null;}
-                if("case".equals(name)) {
+                if ("case".equals(name)) {
```
phillipm: All the other 'name' checks use equalsIgnoreCase. Should we change the 'equals' to that in this check?

app/src/org/commcare/xml/CommCareTransactionParserFactory.java
```diff
@@ -21,9 +21,22 @@
...
+ *
+ * It should be the central point of processing for transactions (eliminating
+ * the use of the old datamodel based processors) and should be used in any
+ * situation where a transaction is expected to be present.
  *
+ * It is expected to behave more or less as a black box, in that it directly
+ * creates/modifies the data models on the system, rather than producing them
+ * for another layer or processing.
+ *
+ * @author ctsims
```
phillipm: Copied this comment over from commcare repo's application/src/org/commcare/util/CommCareTransactionParserFactory.java.

PR #313, opened 2015-05-28
Make labels copyable

phillipm:
Make label text selectable (for copying) during form entry.

[ticket](http://manage.dimagi.com/default.asp?169335)

### DIFF
--- a/app/src/org/odk/collect/android/widgets/TriggerWidget.java
+++ b/app/src/org/odk/collect/android/widgets/TriggerWidget.java
@@ -159,7 +159,7 @@ public class TriggerWidget extends QuestionWidget {
     @Override
     protected void addQuestionText(final FormEntryPrompt p) {
         super.addQuestionText(p);
-        // Enabl users to copy form display outputs.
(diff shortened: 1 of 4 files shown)


app/src/org/odk/collect/android/widgets/TriggerWidget.java
```diff
@@ -30,52 +16,55 @@
...
+     */
     private TextView mStringAnswer;
-    private boolean mInteractive = true; 
-    private static String mOK = "OK";
 
-    private FormEntryPrompt mPrompt;
+    /**
+     * Shows a checkbox when set.
+     */
+    private boolean mInteractive = true;
```
phillipm: Does anyone ever use interactive mode? It is enabled when the appearance attribute of label in the formbuilder is set to something that isn't 'minimal'

wpride: yes, people do use it - it  gives you that little check box that some people like but I don't think we fully consider a question type. I think its a weird legacy thing from odk.collect mostly

PR #314, opened 2015-05-28
Support pixel-based case list column widths

orangejenny:
http://manage.dimagi.com/default.asp?169125


orangejenny:
https://github.com/dimagi/commcare/pull/86
https://github.com/dimagi/javarosa/pull/99

PR #315, opened 2015-05-29
startFormEntry refactor

phillipm:
Noticed that a (most likely) costly incomplete form search could often be ruled out by checking if the session descriptor contained "CASE_ID" in it first.

Figured I'd do some style refactoring while I was changing that.


app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -918,47 +917,93 @@ else if(needed == SessionFrame.STATE_COMMAND_ID) {
...
-            formEntry(platform.getFormContentUri(record.getFormNamespace()), record, CommCareActivity.getTitle(this, null));
-            
-        } catch (StorageFullException e) {
-            throw new RuntimeException(e);
+        FormRecord record = state.getFormRecord();
+
+        if (platform == null &&
+                CommCareApplication._().getCurrentApp() != null) {
+            platform = CommCareApplication._().getCommCarePlatform();
         }
```
phillipm: Removed this try/catch because StorageFullException extends RuntimeException, so why catch and re-raise?
I can add it back in if you see importance to this.

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -918,47 +917,93 @@ else if(needed == SessionFrame.STATE_COMMAND_ID) {
...
-            if(state.getFormRecordId() == -1) {
-                    
-                //If form management isn't enabled we can't have these old forms around anyway
-                if(CommCarePreferences.isIncompleteFormsEnabled()) {
-                    //First, see if we've already started this form before
-                    SessionStateDescriptor existing = state.searchForDuplicates();
-                    
-                    //I'm not proud of the second clause, here. Basically, only ask if we should continue entry if the
-                    //saved state actually involved selecting some data.
-                    if(existing != null && existing.getSessionDescriptor().contains(SessionFrame.STATE_DATUM_VAL)) {
```
phillipm: searchForDuplicates checks for SessionStateDescriptors that match the current one, so if the current one doesn't contain 'STATE_DATUM_VAL', then why even bother running searchForDuplicates, right?

Based on this assumption I've re-ordered the code to only run searchForDuplicates if the current SessionStateDescriptor contains 'STATE_DATUM_VAL'

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -986,51 +1031,6 @@ public String getActivityTitle() {
     protected boolean isTopNavEnabled() {
         return false;
     }
-
-    private void formEntry(Uri formUri, FormRecord r) throws SessionUnavailableException{
-        formEntry(formUri, r, null);
-    }
-    
-    private void formEntry(Uri formUri, FormRecord r, String headerTitle) throws SessionUnavailableException{
-        Logger.log(AndroidLogger.TYPE_FORM_ENTRY, "Form Entry Starting|" + r.getFormNamespace());
```
phillipm: Removed this logger line since it seems like a not very useful thing to log.. Is that cool?
What is the protocol behind logging?

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -918,47 +917,93 @@ else if(needed == SessionFrame.STATE_COMMAND_ID) {
...
+                record, CommCareActivity.getTitle(this, null));
     }
-    
+
+    private void formEntry(Uri formUri, FormRecord r) throws SessionUnavailableException{
+        formEntry(formUri, r, null);
+    }
+
+    private void formEntry(Uri formUri, FormRecord r, String headerTitle) throws SessionUnavailableException {
+        //TODO: This is... just terrible. Specify where external instance data
```
phillipm: Moved this method next to startFormEntry and ran the code re-formatter on it.

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -1045,12 +1048,7 @@ private void startFormEntry(AndroidSessionWrapper state) throws SessionUnavailab
                 platform = CommCareApplication._().getCommCarePlatform();
             }
 
-            //TODO: May need to pass session over manually
             formEntry(platform.getFormContentUri(record.getFormNamespace()), record, CommCareActivity.getTitle(this, null));
-
-        } catch (StorageFullException e) {
-            throw new RuntimeException(e);
-        }
```
phillipm: Removed this try/catch because StorageFullException extends RuntimeException, so why catch and re-raise?
I can add it back in if you see importance to this.

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -1013,31 +1013,34 @@ public void onClick(DialogInterface dialog, int i) {
...
     private void startFormEntry(AndroidSessionWrapper state) throws SessionUnavailableException {
-        try {
-            //If this is a new record (never saved before), which currently all should be 
             if (state.getFormRecordId() == -1) {
-
-                //If form management isn't enabled we can't have these old forms around anyway
-                if (CommCarePreferences.isIncompleteFormsEnabled()) {
-                    //First, see if we've already started this form before
+                if (CommCarePreferences.isIncompleteFormsEnabled() &&
+                        state.getSessionStateDescriptor().getSessionDescriptor().contains(SessionFrame.STATE_DATUM_VAL)) {
```
phillipm: searchForDuplicates checks for SessionStateDescriptors that match the current one, so if the current one doesn't contain 'STATE_DATUM_VAL', then why even bother running searchForDuplicates, right?

Based on this assumption I've re-ordered the code to only run searchForDuplicates if the current SessionStateDescriptor contains 'STATE_DATUM_VAL'

phillipm: Also another very important assumption:
searchForDuplicates returns a SessionStateDescriptor and then we check if it contains STATE_DATUM_VAL. The refactored code checks if the current state's SessionStateDescriptor contains STATE_DATUM_VAL.

This shouldn't matter because the current ssd and the one returned from searchForDuplicates should have the same descriptor hash, and hence both contain STATE_DATUM_VAL, right?

phillipm:
@ctsims Can you review this one? Thanks

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
ctsims: Agreed with this logic. 

Should it go here or should it go in the searchForDuplicates code, though?

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
ctsims: Storagefull used to not be runtime. Please delete these whenever you see them.

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
phillipm: @ctsims I agree and have moved the check for a case in the descriptor into the helper method. Ready to continue review.

PR #316, opened 2015-05-29
FormRecordCleanupTask refactor

phillipm:
Unify form record updating and writing logic. Make sure that the SessionStateDescriptor is written to storage in the same method as the FormRecord is written, instead of segmenting them across methods. 

Otherwise mostly consolidating code.


app/src/org/commcare/android/tasks/FormRecordCleanupTask.java
```diff
@@ -77,184 +72,150 @@ public FormRecordCleanupTask(Context context, CommCarePlatform platform, int tas
...
+            XmlPullParserException, UnfullfilledRequirementsException {
+
+        Pair<FormRecord, String> recordUpdates =
+            reparseRecord(context, oldRecord, FormRecord.STATUS_SAVED);
+        FormRecord updated = recordUpdates.first;
+        String caseId = recordUpdates.second;
+
+        if (caseId != null &&
+                FormRecord.STATUS_UNINDEXED.equals(oldRecord.getStatus()) &&
+                !FormRecord.STATUS_UNINDEXED.equals(newStatus)) {
```
phillipm: Added an additional check to make sure that we don't do indexing if the new record status is STATUS_UNINDEXED.

phillipm: Actually realized newStatus is either SAVED or the same as oldRecord.getStatus, so in a subsequent refactor I take it out completely.

app/src/org/commcare/android/tasks/FormRecordCleanupTask.java
```diff
@@ -263,38 +234,31 @@ public void commit(String[] meta) throws IOException, SessionUnavailableExceptio
...
-            asw.setFormRecordId(parsed.getID());
-            
-            SqlStorage<SessionStateDescriptor> ssdStorage = CommCareApplication._().getUserStorage(SessionStateDescriptor.class);
-            
-            //Also bad: this is not synchronous with the parsed record write
-            try {
-                ssdStorage.write(asw.getSessionStateDescriptor());
-            } catch (StorageFullException e) {
-            }
-        }
```
phillipm: Move SessionStateDescriptor storage write out of this method and into its caller, who also saves the FormRecord to storage. These writes should happen as a transaction, so this makes it easier to see that they are linked

app/src/org/commcare/android/tasks/FormRecordCleanupTask.java
```diff
@@ -76,181 +72,156 @@ public FormRecordCleanupTask(Context context, CommCarePlatform platform, int tas
...
+                        oldRecord.getFormNamespace(), caseId);
+            asw.setFormRecordId(updated.getID());
+
+            SqlStorage<SessionStateDescriptor> ssdStorage =
+                CommCareApplication._().getUserStorage(SessionStateDescriptor.class);
+
+            try {
+                ssdStorage.write(asw.getSessionStateDescriptor());
+            } catch (StorageFullException e) {
+                // TODO PLM should we abort completely here?
```
phillipm: How should we handle this error? Abort completely to ensure that the FormRecord isn't written to storage without a corresponding SessionStateDescriptor?

phillipm:
@ctsims This PR deals with Form Record write/update logic, so you should probably be the one to look at it.

phillipm:
@ctsims any chance you can look at this soon?

ctsims:
reviewing now

app/src/org/commcare/android/tasks/FormRecordCleanupTask.java
ctsims: for now we can actually dump the catch, this is now a runtime exception we really don't expect on android.

PR #319, opened 2015-06-01
Replace println's with Log.d

phillipm:
For finer grain control over logging output, replace all System.out.println statements with Log.d


dcluna:
When initializing TAG variables with the same name as the enclosing class, I prefer to use this form:
```java
public static final String TAG = MyClass.class.getSimpleName();
```
So the TAGs won't be misleading after a rename or any other refactoring.

phillipm:
@dcluna Didn't know about that, good call. Updated the PR, should be good to re-review/merge.

dcluna:
@phillipm We need to replicate this on master-oldui now

PR #320, opened 2015-06-01
AndroidSessionWrapper: Make string equality tests use .equals

phillipm:
This is untested, but pretty sure it does the right thing.

### DIFF
--- a/app/src/org/commcare/android/models/AndroidSessionWrapper.java
+++ b/app/src/org/commcare/android/models/AndroidSessionWrapper.java
@@ -275,14 +275,15 @@ public class AndroidSessionWrapper {
         Hashtable<String, Entry> entries = platform.getMenuMap();
         for(StackFrameStep step : session.getFrame().getSteps()) {
             String val = null; 
-            if(step.getType() == SessionFrame.STATE_COMMAND_ID) {
+            if(SessionFrame.STATE_COMMAND_ID.equals(step.getType())) {
                 //Menu or form. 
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/android/models/AndroidSessionWrapper.java
```diff
@@ -299,9 +288,12 @@ public String getTitle() {
...
         //investigate this.
-        
-        //Walk backwards until we find something with a long detail
-        while(session.getFrame().getSteps().size() > 0 && (session.getNeededData() != SessionFrame.STATE_DATUM_VAL || session.getNeededDatum().getLongDetail() == null)) {
+
+        // Walk backwards until we find something with a long detail
+        while (session.getFrame().getSteps().size() > 0 &&
+                (!SessionFrame.STATE_DATUM_VAL.equals(session.getNeededData()) ||
+                        !(session.getNeededDatum() != null &&
+                                session.getNeededDatum().getLongDetail() != null))) {
```
phillipm: Added additional null check to make sure that session.getNeededDatum() isn't null before calling .getLongDetail() on it.

ctsims: Is there a specific reason? If session.getNeededData() is returning a datum but getNeededDatum() is null that feels like a pretty serious issue. I might suggest failing inside getNeededData structurally to make the nature of the error more clear rather than bubbling up to here.

PR #329, opened 2015-06-12
Adding missing code in FormRecordListActivity (incomplete forms search)

dcluna:
Also centering the logo in connection_diagnostic screen.

### DIFF
--- a/app/src/org/commcare/dalvik/activities/FormRecordListActivity.java
+++ b/app/src/org/commcare/dalvik/activities/FormRecordListActivity.java
@@ -30,6 +30,7 @@ import android.util.Pair;
 import android.view.ContextMenu;
 import android.view.ContextMenu.ContextMenuInfo;
 import android.view.Menu;
+import android.view.MenuInflater;
 import android.view.MenuItem;
 import android.view.View;
 import android.widget.AdapterView;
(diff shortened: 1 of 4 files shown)


dcluna:
Better to isolate the action bar logic in CommCareActivity itself (the same logic is used in other class), to avoid code duplication

app/src/org/commcare/dalvik/activities/FormRecordListActivity.java
```diff
@@ -93,7 +96,8 @@
     private ImageButton barcodeButton;
     private Spinner filterSelect;
     private ListView listView;
-    
+    private SearchView searchView;
```
phillipm: The linter says that this variable can be made local to the onCreateOptionsMenu method.

phillipm:
Looks good, I'll merge this after you make "searchView" a method-level variable instead of a class-level variable.

app/src/org/commcare/dalvik/activities/FormRecordListActivity.java
dcluna: After the last commit, this linter warning was removed

phillipm:
Thanks for addressing that and nice unification of repeated code!
I tidied up a few style things and merged.

PR #332, opened 2015-06-15
Update IntentWidget.java

wpride:


wpride:
http://manage.dimagi.com/default.asp?171026

fixes bug causing crash for intentwidget question types

ctsims:
@wpride I don't understand what your fogbugz link is supposed to be pointing to. Nothing in there mentions intent callouts, is that where the bug came from?

wpride:
updated with correct ticket @ctsims

app/src/org/odk/collect/android/widgets/IntentWidget.java
```diff
@@ -84,7 +84,6 @@ public void makeTextView(FormEntryPrompt prompt) {
         }
 
         // finish complex layout
-        addView(launchIntentButton);
```
ctsims: If the bug is that this is null at startup, that also applies to the above condition (s != null, results in a direct dereference), that should also be cleaned up before this is accepted.

dcluna:
Apparently this also solves http://manage.dimagi.com/default.asp?171260

wpride:
@ctsims @dcluna this is ready for review

dcluna:
See also https://github.com/dimagi/commcare-odk/pull/338

phillipm:
This is fixes a 2.22 bug, did these commits actually make it into the commcare_2.22 branch?

PR #333, opened 2015-06-15
Version bump to 2.23

dcluna:


app/AndroidManifest.xml
```diff
@@ -2,7 +2,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="org.commcare.dalvik"
     android:versionCode="106"
-    android:versionName="2.22" >
+    android:versionName="2.23" >
```
ctsims: can we make this, like, 2.23.N so we can spot it from the field?

PR #341, opened 2015-06-22
fix image resizing

wpride:
Relates to this ticket:

http://manage.dimagi.com/default.asp?171248

Basically our old algorithm would not resize UP a very small image unless the getMinimumHeight() method of the parent view suggested it. This seems to go against the feature definition of full/half/width resizing which suggests that the image will be stretched to take up that amount of the screen.

I think this change does the correct thing (just use the max width/height of the screen for calculation baselines). However, the image resizing is tricky (figuring our how much screen real estate you actually have esp. given nav bar, keyboard, variable widget sizes, etc) so I don't know if there's a completely "correct" way to do this. Happy to hear comment.

### DIFF
--- a/app/src/org/odk/collect/android/views/ResizingImageView.java
+++ b/app/src/org/odk/collect/android/views/ResizingImageView.java
@@ -1,11 +1,5 @@
 package org.odk.collect.android.views;
 
-import java.io.File;
-
-import org.commcare.dalvik.R;
-import org.javarosa.core.reference.InvalidReferenceException;
-import org.javarosa.core.reference.ReferenceManager;
-
 import android.annotation.SuppressLint;
 import android.content.ActivityNotFoundException;
 import android.content.Context;
@@ -20,6 +14,12 @@ import android.view.MotionEvent;
 import android.view.ScaleGestureDetector;
 import android.widget.ImageView;
 import android.widget.Toast;
+
+import org.commcare.dalvik.R;
+import org.javarosa.core.reference.InvalidReferenceException;
+import org.javarosa.core.reference.ReferenceManager;
+
+import java.io.File;
 
 /**
  * @author wspride
@@ -104,6 +104,7 @@ public class ResizingImageView extends ImageView {
          */
         @Override
         public boolean onDoubleTap(MotionEvent e) {
+            getSuggestedMinimumHeight();
             setFullScreen();
             return true;
         }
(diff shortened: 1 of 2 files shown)


phillipm:
@wpride you sure you included the fix in this PR? The diff is one line and just changes a error message string.

wpride:
@phillipm updated. sorry about that. forcing myself to exclusively use command line git resulting in some missteps

app/src/org/odk/collect/android/views/ResizingImageView.java
```diff
@@ -157,41 +158,22 @@ public void setFullScreen(){
...
 
-                        Pair<Integer,Integer> mPair = new Pair<Integer,Integer>(new Double(width * scaleFactor).intValue(), new Double(height * scaleFactor).intValue());
+        int maxHeight = hMode == MeasureSpec.AT_MOST
+                ? Math.min(MeasureSpec.getSize(heightMeasureSpec), mMaxHeight)
+                    : mMaxHeight;
 
+        Pair<Integer,Integer> mPair = new Pair<Integer,Integer>(new Double(maxWidth * scaleFactor).intValue(), new Double(maxHeight * scaleFactor).intValue());
 
-                        return mPair;
+        return mPair;
```
phillipm: It is cleaner to return an expression instead of assigning it a variable and then immediately returning that variable.

app/src/org/odk/collect/android/views/ResizingImageView.java
```diff
@@ -157,41 +158,22 @@ public void setFullScreen(){
...
-                            height = (int) (width / ratio);
-                        }
+                    : mMaxWidth;
 
-                        Pair<Integer,Integer> mPair = new Pair<Integer,Integer>(new Double(width * scaleFactor).intValue(), new Double(height * scaleFactor).intValue());
+        int maxHeight = hMode == MeasureSpec.AT_MOST
+                ? Math.min(MeasureSpec.getSize(heightMeasureSpec), mMaxHeight)
+                    : mMaxHeight;
 
+        Pair<Integer,Integer> mPair = new Pair<Integer,Integer>(new Double(maxWidth * scaleFactor).intValue(), new Double(maxHeight * scaleFactor).intValue());
```
phillipm: The convention of prefixing variable names with 'm' denotes (class-level) member variable. In this case, mPair is a method level var, so it probably shouldn't have the 'm'.

app/src/org/odk/collect/android/views/ResizingImageView.java
```diff
@@ -157,41 +158,22 @@ public void setFullScreen(){
         }
     }
 
-    public Pair<Integer,Integer> getWidthHeight(int widthMeasureSpec, int heightMeasureSpec, Drawable drawable, double scaleFactor){
+    public Pair<Integer,Integer> getWidthHeight(int widthMeasureSpec, int heightMeasureSpec, double scaleFactor){
```
phillipm: The method parameter 'scaleFactor' overshadows the class-level 'scaleFactor' field. This is confusing.

app/src/org/odk/collect/android/views/ResizingImageView.java
```diff
@@ -157,41 +158,22 @@ public void setFullScreen(){
...
-                        int height = (int) (width / ratio);
-
-                        height = Math.min(Math.max(height, getSuggestedMinimumHeight()), maxHeight);
-                        width = (int) (height * ratio);
-
-                        if (width > maxWidth) {
-                            width = maxWidth;
-                            height = (int) (width / ratio);
-                        }
+                    : mMaxWidth;
```
phillipm: I find ternary expressions quite unreadable, I think something like this is cleaner: https://gist.github.com/phillipm/51ab801dd079f4d961c0

app/src/org/odk/collect/android/views/ResizingImageView.java
```diff
@@ -157,41 +158,22 @@ public void setFullScreen(){
         }
     }
 
-    public Pair<Integer,Integer> getWidthHeight(int widthMeasureSpec, int heightMeasureSpec, Drawable drawable, double scaleFactor){
+    public Pair<Integer,Integer> getWidthHeight(int widthMeasureSpec, int heightMeasureSpec, double scaleFactor){
 
         int wMode = MeasureSpec.getMode(widthMeasureSpec);
         int hMode = MeasureSpec.getMode(heightMeasureSpec);
```
phillipm: wMode and hMode variables are only read once, so you should be fine not creating the variables at all, but just using the getMode expressions in the conditionals below.

app/src/org/odk/collect/android/views/ResizingImageView.java
phillipm: This method can be private

phillipm:
@wpride done reviewing

PR #343, opened 2015-06-22
Cleaning up CommCareActivity

phillipm:
Refactor that mostly removes unused code from CommCareActivity.

Other changes:
 - simplification of conditional logic
 - rename method to not start with capital letter
 - use StringBuilder inside of loop instead of String concatenation


app/src/org/commcare/android/framework/CommCareActivity.java
```diff
@@ -236,7 +187,7 @@ private void loadFields() {
                             }
                         }
                         
-                        if(element.locale() != "") {
+                        if(!"".equals(element.locale())) {
```
phillipm: might trigger more than old code, but old code used the java language in a sketchy way.

PR #348, opened 2015-06-24
Ensure audio playback never occurs in parallel

phillipm:
Fix for issue 1) in [this ticket](http://manage.dimagi.com/default.asp?158453#894411).

The issue was that form questions with multiple audio buttons allowed playback of all buttons at once, instead of stopping one when another was clicked.

To fix this bug it was sort of necessary to do a complete refactor of the controller that handles audio playback.

Background context:
In that past there were two instances of this audio controller, one attached to the CommCareActivity, that processed audio playback when listing cases for selection. The other was an anonymous instance built in the AudioButton. There was some hard to follow logic that toggled between these, and the second instance only made use of some of the interface exposed by the controller.

Instead of having an audio controller interface implemented by the CommCareActivity, this PR introduces a singleton audio controller object. This is since only once piece of audio should be playing at a time, so we only need one instance floating around that handles this.
Activities can register media and control playback through this controller, and importantly, the invariant that only one piece of audio be played at once can be enforced.


app/src/org/odk/collect/android/views/ODKView.java
```diff
@@ -453,21 +398,11 @@ public void setOnFocusChangeListener(OnFocusChangeListener l) {
...
     @Override
     public boolean onLongClick(View v) {
         return false;
     }
     
-
-    /*
-     * (non-Javadoc)
-     * @see android.view.View#cancelLongPress()
-     */
```
phillipm: I'm going to say that these non-Javadoc comments are bad form.
They add clutter, can get out of date, and contain information that is easily looked up. A simple Override annotation and good method names should suffice.

app/src/org/odk/collect/android/views/media/AudioButton.java
```diff
@@ -27,32 +27,26 @@
...
     private String URI;
     private MediaState currentState;
-    private final AudioController controller;
     private Object residingViewId;
 
+    /**
+     * Used by inflater.
+     */
     public AudioButton(Context context, AttributeSet attrs) {
         super(context, attrs);
```
phillipm: What sort of logic needs to be in constructors that are used by the media inflater?

app/src/org/odk/collect/android/views/media/AudioButton.java
```diff
@@ -22,40 +21,61 @@
...
 
+    /**
+     * The id of the ListAdapter view that contains this button
+     */
+    private ViewId residingViewId;
+
+    /**
+     * Used by media inflater.
+     */
     public AudioButton(Context context, AttributeSet attrs) {
```
phillipm: @dcluna Do you know what sort of logic needs to be in constructors that are used by the media inflater?

dcluna: Only the bits that enable preview in Android Studio

phillipm: I took it out, because it looked like it wasn't being called, and the app crashed on login.

phillipm:
@amstone326 Hey, would be awesome if you could review this PR, since you did a fair amount of the original coding. Thanks!

amstone326:
This looks good to me, and about 100x cleaner than how it was being done previously. Still new at the PR review thing so my lack of comments doesn't mean a lot, but this probably only needs a quick pass by someone else to be ready to merge.

phillipm:
@ctsims do you want to do a quick pass of this too?

PR #349, opened 2015-06-24
On login, check if sync button is null before invoking it.

phillipm:
I was getting a null pointer exception sometimes upon login. This makes sure the syncButton isn't null before having it invoke methods.

Any ideas as to why it is null sometimes?

### DIFF
--- a/app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
+++ b/app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
@@ -1507,13 +1507,14 @@ public class CommCareHomeActivity extends CommCareActivity<CommCareHomeActivity>
         if (syncTextKey == null) {
             syncTextKey = isDemoUser() ? "home.sync.demo" : "home.sync";
         }
-        if (syncDetails.second[0] > 0) {
-            Spannable syncIndicator = (this.localize("home.sync.indicator", new String[]{String.valueOf(syncDetails.second[0]), Localization.get(syncTextKey)}));
-            syncButton.setNotificationText(syncIndicator);
-            adapter.notifyDataSetChanged();
-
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -1507,13 +1507,14 @@ private void setSyncText(Pair<Long, int[]> syncDetails, String syncTextKey) {
...
-            syncButton.setNotificationText(syncIndicator);
-            adapter.notifyDataSetChanged();
-
-        } else {
-            if (syncButton != null) syncButton.setText(this.localize(syncTextKey));
+        if (syncButton != null) {
+            if (syncDetails.second[0] > 0) {
+                Spannable syncIndicator = (this.localize("home.sync.indicator", new String[]{String.valueOf(syncDetails.second[0]), Localization.get(syncTextKey)}));
+                syncButton.setNotificationText(syncIndicator);
+                adapter.notifyDataSetChanged();
```
phillipm: Its okay that notifyDataSetChanged isn't called when syncButton is null, right?

dcluna: Yea, but syncButton should not be null when calling this...

phillipm: When calling the setSyncText method? If so I will move this check to the caller of setSyncText

dcluna: Exactly, it only makes sense to set the syncButton text if it has been created (i.e., is visible)

phillipm:
@dcluna updated

PR #350, opened 2015-06-26
Allow uploading of all audio, image, and video filetypes to HQ

phillipm:
There was a bug where the sync process wasn't sending certain case audio attachments up to HQ. The problem was that only a handful of hard coded filetypes were accepted as valid.

This PR allows for all files that have an audio, video, or image mimetype to be uploaded as case attachments to HQ.

Note: currently if a case is uploaded to HQ that expects an attachment but the phone didn't include that attachment, the only info HQ returns is a 500 error.

Fix for [this ticket](http://manage.dimagi.com/default.asp?173209)

Also includes:
 - made AudioWidget stop adding double backslashes to filepaths
 - removed unneeded platform parameter from SendTask constructor
 - Refactored FormUploadUtil to use helper methods
 - make CommCare not crash if you select an audio file in AudioWidget that doesn't have an extension (even though this should never really happen, good to be safe)

### DIFF
--- a/app/src/org/commcare/android/util/FormUploadUtil.java
+++ b/app/src/org/commcare/android/util/FormUploadUtil.java
@@ -11,7 +11,6 @@ import javax.crypto.Cipher;
 import javax.crypto.NoSuchPaddingException;
 import javax.crypto.spec.SecretKeySpec;
 
-import org.apache.http.Header;
 import org.apache.http.HttpResponse;
 import org.apache.http.client.ClientProtocolException;
 import org.apache.http.entity.mime.MultipartEntity;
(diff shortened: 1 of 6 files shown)


app/src/org/commcare/android/util/FormUploadUtil.java
```diff
@@ -63,255 +70,308 @@ public static Cipher getDecryptCipher(SecretKeySpec key) {
...
             } else if (f.getName().endsWith(".3gp")) {
                 fb = new FileBody(f, "video/3gpp");
                 if (fb.getContentLength() <= MAX_BYTES) {
                     entity.addPart(f.getName(), fb);
-                    Log.i(t, "added video file " + f.getName());
+                    Log.i(TAG, "added video file " + f.getName());
+                } else {
+                    Log.i(TAG, "file " + f.getName() + " is too big");
+                }
+            } else if (supported || isAudioVisualMimeType(f)) {
```
phillipm: The actual bug fix. Adds queued up files that have correct mimetype to the multipart entity to be submitted to HQ.

phillipm:
We want to merge this into the old and new ui branches of commcare_2.22, since it is a QA bug.

PR #368, opened 2015-06-30
don't allow pixel-based case list column widths

orangejenny:
Old UI version of https://github.com/dimagi/commcare-odk/pull/352


phillipm:
@orangejenny Is that the correct merge target? I would expect oldui/commcare_2.22

orangejenny:
@phillipm yep, you're right

PR #370, opened 2015-06-30
Restore query in case search

dcluna:
Restoring query string in search bar for case search screens.

### DIFF
--- a/app/src/org/commcare/android/framework/CommCareActivity.java
+++ b/app/src/org/commcare/android/framework/CommCareActivity.java
@@ -4,6 +4,7 @@ import android.app.Activity;
 import android.app.AlertDialog;
 import android.content.Context;
 import android.content.DialogInterface;
+import android.content.SharedPreferences;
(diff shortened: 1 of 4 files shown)


app/src/org/commcare/android/framework/CommCareActivity.java
```diff
@@ -117,7 +125,15 @@ protected void onCreate(Bundle savedInstanceState) {
         
         mGestureDetector = new GestureDetector(this, this);
     }
-    
+
+    protected void restoreLastQueryString(String key) {
+        SharedPreferences settings = getSharedPreferences(CommCarePreferences.ACTIONBAR_PREFS, 0);
+        lastQueryString = settings.getString(key, null);
+        if (BuildConfig.DEBUG) {
```
phillipm: I'm missing an import for BuildConfig. Which import is the correct one? dalvik, mapviewballoons, chartdemo?

phillipm:
Looks good other than the missing import

PR #377, opened 2015-07-01
CommCareSetupActivity Refactor

phillipm:
General linting & FindBugs changes. Removed lots of unused fields, code, and all the non-Javadoc override comments. 

**Includes commits from https://github.com/dimagi/commcare-odk/pull/376**, so review that first


app/src/org/commcare/android/tasks/ResourceEngineTask.java
```diff
@@ -230,10 +228,6 @@ protected ResourceEngineOutcomes doTaskBackground(String... profileRefs) {
                 platform.upgrade(global, temporary, recovery);
 
                 // And see where we ended up to see whether an upgrade actually occurred
-            } else if (partialMode) {
```
phillipm: partialMode was always false, so let's just remove this, especially since the 'else' body below has pretty much the same behaviour

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -550,65 +495,31 @@ public void done(boolean requireRefresh) {
...
-                this.canRetry = canRetry;
-            } else {
-                
-                this.displayMessage= message.getDetails();
-                this.canRetry = canRetry;
-                
-                String fullErrorMessage = message.getDetails();
-                
-                if(alwaysNotify){
-                    fullErrorMessage = fullErrorMessage + message.getAction();
```
phillipm: All of this logic was useless because the vars being assigned are never read elsewhere

PR #387, opened 2015-07-02
CommCarePlatform and CommCareApp refactor

phillipm:
Misc cleanup of platform and app classes, plus a few other small changes.

Mostly dead code removal and linting.

### DIFF
--- a/app/src/org/commcare/android/util/AndroidCommCarePlatform.java
+++ b/app/src/org/commcare/android/util/AndroidCommCarePlatform.java
@@ -14,34 +14,25 @@ import org.commcare.util.CommCarePlatform;
 import org.javarosa.core.model.instance.FormInstance;
 import org.javarosa.core.services.storage.IStorageUtilityIndexed;
 
-import android.content.Context;
 import android.net.Uri;
-import android.os.Bundle;
 
 /**
  * @author ctsims
  */
 public class AndroidCommCarePlatform extends CommCarePlatform {
 
-    public static final String ENTITY_NONE = "NONE";
-    public static final String STATE_REFERRAL_TYPE = "REFERRAL_TYPE";
-
-    private Hashtable<String, String> xmlnstable;
-    private Context c;
+    private final Hashtable<String, String> xmlnstable;
     private ResourceTable global;
     private ResourceTable upgrade;
     private ResourceTable recovery;
 
     private Profile profile;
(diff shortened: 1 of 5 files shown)


app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -716,9 +716,6 @@ else if(resultCode == RESULT_OK) {
                     break;
                 } else if(resultCode == RESULT_OK) {
                     currentState.getSession().setDatum(currentState.getSession().getNeededDatum().getDataId(), intent.getStringExtra(SessionFrame.STATE_DATUM_VAL));
-                    if(intent.hasExtra(CallOutActivity.CALL_DURATION)) {
-                        platform.setCallDuration(intent.getLongExtra(CallOutActivity.CALL_DURATION, 0));
-                    }
```
dcluna: What was this used for before?

phillipm: Oh, for some reason I thought this was unused, but now double checking, it isn't! Thanks for asking, I'll add it back

app/src/org/commcare/android/util/AndroidCommCarePlatform.java
```diff
@@ -96,34 +87,10 @@ public void setProfile(Profile p) {
...
-        }
-    }
-
-    public void setCallDuration(long callDuration) {
-        this.callDuration = callDuration;
-    }
-
-    public long getCallDuration() {
-        return callDuration;
-    }
```
phillipm: This also needs to be added back in.

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
phillipm: @dcluna Actually, that line calls AndroidCommCarePlatform.setCallDuration, which sets a class field that is never accessed. That is why I took all this out. The functionality seems to have been moved to CalloutActivity.java. Please double check this (that AndroidCommCarePlatform.getCallDuration is never called), but I think what this PR removes is never used.

phillipm: According to @ctsims, all the call duration logic can be removed, so we shouldn't have to worry to hard about taking out something important

PR #388, opened 2015-07-02
remove URL prep

wpride:


wpride:
fixes bug that was breaking installs from local jr: file system

dcluna:
This should also go to 2.22, right?

ctsims:
Quite likely, yeah, I'm comfortable pulling over as-is

PR #389, opened 2015-07-02
markdown for groups, repeats, lists

wpride:
Allows labels in repeats, field lists, groups to have markdown formatting

http://manage.dimagi.com/default.asp?174085


wpride:
@emord

PR #399, opened 2015-07-06
Check multimedia file extension during form entry

phillipm:
Don't let user select multimedia files of invalid type during form entry.

Currently, when the user chooses a non-supported file for the audio/video/image widget during form entry, no warning is shown. Then the form fails to be uploaded to the server because of it.

This PR shows an error message and clears the user's file choice if they try to use a file that isn't supported for audio/video/image widget.

Fix for [this ticket](http://manage.dimagi.com/default.asp?173209)
oldui version of this: https://github.com/dimagi/commcare-odk/pull/400

### DIFF
--- a/app/res/values/strings.xml
+++ b/app/res/values/strings.xml
@@ -107,6 +107,7 @@ Copyright (c) 2012 readyState Software Ltd.</string>
     <string name="altitude">Altitude</string>
     <string name="app_name">ODK Collect</string>
     <string name="attachment_oversized">Warning: attached large file of size %s kb. This could cause problems with submitting to the server depending on your network conditions.</string>
+    <string name="attachment_invalid">Invalid filetype choosen. Please select a valid multimedia file.</string>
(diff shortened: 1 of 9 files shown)


app/res/values/strings.xml
```diff
@@ -107,6 +107,7 @@ Copyright (c) 2012 readyState Software Ltd.</string>
     <string name="altitude">Altitude</string>
     <string name="app_name">ODK Collect</string>
     <string name="attachment_oversized">Warning: attached large file of size %s kb. This could cause problems with submitting to the server depending on your network conditions.</string>
+    <string name="attachment_invalid">Invalid filetype choosen. Please select a valid multimedia file.</string>
```
phillipm: What is the process for getting this translated into different languages?

dcluna: add this string to messages_ccodk_default.txt, call Localization.get() on the message key

PR #401, opened 2015-07-07
New tabs

dcluna:
New tabs, styled just like the ones in the Play Store app.


app/res/layout/tabbed_detail_header_modern.xml
```diff
@@ -0,0 +1,37 @@
...
+        android:minWidth="48dp"
+        android:layout_centerVertical="true"
+        android:src="@drawable/icon_chevron_left_brand"/>
+    <TextView
+        android:id="@+id/header_title"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_centerInParent="true"
+        android:background="@color/transparent"
+        android:text="Tab title"
```
orangejenny: This will never be displayed, will it? Should it be here at all?

app/src/org/commcare/android/framework/EntityDetailFragment.java
```diff
@@ -33,12 +34,14 @@
...
     public static final String DETAIL_INDEX = "edf_detail_index";
     public static final String CHILD_REFERENCE = "edf_detail_reference";
-    
+
     private AndroidSessionWrapper asw;
     private NodeEntityFactory factory;
     private EntityDetailAdapter adapter;
     private EntityDetailAdapter.EntityDetailViewModifier modifier;
 
+    private boolean tabbedDetailHeader = true;
```
orangejenny: Looks like this isn't used.

app/src/org/commcare/android/view/TabbedDetailView.java
```diff
@@ -159,7 +166,7 @@ public void refresh(Detail detail, TreeReference reference, int index, boolean h
                 hasDetailCalloutListener, new DefaultEDVModifier()
         );
         mViewPager.setAdapter(mEntityDetailPagerAdapter);
-        markSelectedTab(0);
+        if(!useNewTabStyle) markSelectedTab(0);
```
orangejenny: Why is this line conditional on `useNewTabStyle`, but not the other places that the first tab is selected by default?

app/src/org/commcare/android/framework/EntityDetailFragment.java
```diff
@@ -33,12 +34,14 @@
     public static final String DETAIL_ID = "edf_detail_id";
     public static final String DETAIL_INDEX = "edf_detail_index";
     public static final String CHILD_REFERENCE = "edf_detail_reference";
-    
+
     private AndroidSessionWrapper asw;
     private NodeEntityFactory factory;
```
phillipm: This can be made local to onCreateView

app/src/org/commcare/android/framework/EntityDetailFragment.java
```diff
@@ -20,6 +20,7 @@
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
+import android.widget.ImageView;
```
phillipm: Unused import (along with AndroidUtil and TextView)

app/src/org/commcare/android/view/TabbedDetailView.java
```diff
@@ -37,8 +38,10 @@
     private EntityDetailPagerAdapter mEntityDetailPagerAdapter;
     private ViewPager mViewPager;
     private View mViewPagerWrapper;
-    
+
     private int mAlternateId = -1;
+
+    private boolean useNewTabStyle = true;
```
phillipm: Might be good to make this USE_NEW_TAB_STYLE and final, so that it is clear that it is a hand set flag.

app/src/org/commcare/android/view/TabbedDetailView.java
```diff
@@ -11,6 +11,7 @@
 import android.support.v4.app.FragmentActivity;
 import android.support.v4.view.ViewPager;
 import android.util.AttributeSet;
+import android.util.Log;
```
phillipm: Unused import

phillipm:
Done reviewing. Looks good after comments are addressed.

To consider in a future PR: It looks like TabbedDetailView.DefaultEDVModifier doesn't have an implementation of Parcelable.CREATOR, which _might_ be important.

app/src/org/commcare/android/view/TabbedDetailView.java
dcluna: removed the useNewTabStyle var, it makes no sense at all (we'll always want to use the new style, right?)

dcluna:
I don't think the Creator field is important right now, because the only variable we're referencing is retrieved in a static context (get the colors from the current theme).

app/src/org/commcare/android/view/TabbedDetailView.java
```diff
@@ -102,54 +101,7 @@ public void onPageScrollStateChanged(int arg0) { }
     /*
      * Populate view with content from given Detail.
      */
-    public void setDetail(Detail detail) {
-        Detail[] details = detail.getDetails();
-
-        LinearLayout.LayoutParams pagerLayout = new LinearLayout.LayoutParams(LinearLayout.LayoutParams.MATCH_PARENT, LinearLayout.LayoutParams.WRAP_CONTENT);
```
phillipm: Just want to verify that you meant to remove the bulk of this method in your latest code review commit.

dcluna: Yes, we're doing this in the fragment now

PR #402, opened 2015-07-08
Fix case list dimensions

orangejenny:
Revert https://github.com/dimagi/commcare-odk/pull/352, which was itself a revert, and fix the actual bug.

@ctsims

### DIFF
--- a/app/src/org/commcare/android/view/EntityView.java
+++ b/app/src/org/commcare/android/view/EntityView.java
@@ -69,14 +69,12 @@ public class EntityView extends LinearLayout {
         mIsAsynchronous = e instanceof AsyncEntity;
         this.searchTerms = searchTerms;
         this.tts = tts;
-        this.setWeightSum(1);
         this.controller = controller;
         this.renderedGraphsCache = new Hashtable<Integer, Hashtable<Integer, View>>();
         this.rowId = rowId;
         this.views = new View[e.getNumFields()];
         this.forms = d.getTemplateForms();
         this.mHints = d.getTemplateSizeHints();
-        float[] weights = calculateDetailWeights(d.getTemplateSizeHints());
         
         for (int i = 0; i < views.length; ++i) {
             if (mHints[i] == null || !mHints[i].startsWith("0")) {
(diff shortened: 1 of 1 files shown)


orangejenny:
don't merge just yet - making some changes to make calculateDetailWeights more robust

orangejenny:
put changes in master instead of 2.22; this is good to merge

phillipm:
I'm having trouble following what is going on here.
Are you saying this PR is good to merge into 2.22 and we should also merge https://github.com/dimagi/commcare-odk/pull/412 into master? 
If so, we should take care merging 2.22 back into master, these changes might be at odds with each other.

orangejenny:
@phillipm Yes, this PR can go into 2.22 and 412 can go into master. I think the merge will go smoothly based on the specific lines that have changed, but yeah, I'll double-check when the 2.22 merge happens.

phillipm:
@orangejenny Cool, thanks!

ctsims:
@orangejenny Is there a oldui version of this as well?

orangejenny:
@ctsims yes: https://github.com/dimagi/commcare-odk/pull/403

app/src/org/commcare/android/view/EntityView.java
```diff
@@ -530,5 +491,8 @@ protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
                 views[i].setLayoutParams(params);
             }
         }
-    }*/
+        
+        // Use children's new widths
+        super.onMeasure(widthMeasureSpec, heightMeasureSpec);
```
phillipm: Can you explain why there are two calls to super.onMeasure in this method?
Could you remove the first one or the second one or are they both necessary?

orangejenny: The first `super.onMeasure` is so that `getMeasuredWidth` returns the correct width of the row. The function then sets all of the children's desired widths, and the second `super.onMeasure` is where the children's new dimensions actually get set.

^^ This was a "take stabs in the dark until something works" kind of bug, followed by looking at https://developer.android.com/guide/topics/ui/how-android-draws.html to rationalize *why* it works.

phillipm: I'm curious, do you think it would work to replace the second super call with setMeasuredDimension?

ctsims: @phillipm I doubt it in this case.

This setup is relying on the layout engine to still be performing sizing on the height of all of the associated views, so we can't set anything manually without doing those calculations on our own.

The first call to onMeasure results in this view being clear on its height and width, after which it sets the layoutparams for its children to force them to a certain width. After it does that, this view needs both the children and itself to have heights recalculated, which you can only trigger by asking for a recursive measurement again.

phillipm: @ctsims you're right, I just tried it and it failed horribly!
I added some in-code comments describing what you and @orangejenny elaborated here.

orangejenny: > I added some in-code comments describing what you and @orangejenny elaborated here.

Nice, thanks.

PR #403, opened 2015-07-08
Fix case list dimensions

orangejenny:
Old UI version of https://github.com/dimagi/commcare-odk/pull/402


orangejenny:
don't merge just yet - making some changes to make calculateDetailWeights more robust

orangejenny:
put changes in master instead of 2.22; this is good to merge

PR #410, opened 2015-07-08
Fixing autocomplete issues with 'Raw' prefix

dcluna:
Autofilling the default location when user chooses 'Raw' prefix and automatically handling URLs without a :// prefix.

This complements #388.

### DIFF
--- a/app/src/org/commcare/android/fragments/SetupEnterURLFragment.java
+++ b/app/src/org/commcare/android/fragments/SetupEnterURLFragment.java
@@ -7,6 +7,7 @@ import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
 import android.support.v4.app.Fragment;
+import android.widget.AdapterView;
 import android.widget.Button;
 import android.widget.EditText;
 import android.widget.Spinner;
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/android/fragments/SetupEnterURLFragment.java
```diff
@@ -51,6 +55,21 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
         installButton = (Button) view.findViewById(R.id.start_install);
         installButton.setText(Localization.get("install.button.start"));
         prefixURLSpinner = (Spinner) view.findViewById(R.id.url_spinner);
+        prefixURLSpinner.setOnItemSelectedListener(new AdapterView.OnItemSelectedListener() {
+            @Override
+            public void onItemSelected(AdapterView<?> parent, View view, int position, long id) {
+                String selectedText = profileLocation.getText().toString();
+                if((position == prefixURLSpinner.getCount() -1) &&
+                        (selectedText == null || selectedText.length() == 0)) {
```
orangejenny: Relying on position seems more brittle than, say, looking for the text "raw" (though that also isn't perfect). But the assertion below is a reasonable way to catch unintended consequences of changing the dropdown.

PR #412, opened 2015-07-09
Better handle edge cases in case list widths

orangejenny:
Make calculateDetailWidths better handle cases where the requested column widths don't fit well into the available screen width.

Putting in master rather than 2.22 because HQ shouldn't actually be generating problem layouts unless you're using very large icons in case details. Which is possible, since previously HQ would scale any large icons down to 13%, but pretty edge-casey.

### DIFF
--- a/app/src/org/commcare/android/view/EntityView.java
+++ b/app/src/org/commcare/android/view/EntityView.java
@@ -441,60 +441,69 @@ public class EntityView extends LinearLayout {
         }
     }
 
+    /**
+     * Determine width of each child view, based on mHints, the suite's size hints.
+     * mHints contains a width hint for each child view, each one of
+     * - A string like "50%", requesting the field take up 50% of the row
+     * - A string like "200", requesting the field take up 200 pixels
+     * - Null, not specifying a width for the field
+     * This function will parcel out requested widths and divide remaining space among unspecified columns.
+     * @param fullSize Width, in pixels, of the containing row.
+     * @return Array of integers, each corresponding to a child view,
+     * representing the desired width, in pixels, of that view.
+     */
     private int[] calculateDetailWidths(int fullSize) {
         // Convert any percentages to pixels. Percentage columns are treated as percentage of the entire screen width.
-        int[] hints = new int[mHints.length];
+        int[] widths = new int[mHints.length];
         for (int i = 0; i < mHints.length; i++) {
             if (mHints[i] == null) {
-                hints[i] = -1;
+                widths[i] = -1;
             } else if (mHints[i].contains("%")) {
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/android/view/EntityView.java
```diff
@@ -454,25 +454,47 @@ private static void removeSpans(Spannable raw) {
...
+            else {
+                indeterminateColumns++;
             }
         }
-        
-        // Set column widths
-        int defaultWidth = sharedBetween == 0 ? 0 : fullSize / sharedBetween;
-        for(int i = 0; i < hints.length; ++i) {
-            widths[i] = hints[i] == -1 ? defaultWidth : hints[i];
+        if (
```
phillipm: Not a fan of the atypical line breaks in this if's test expression

app/src/org/commcare/android/view/EntityView.java
```diff
@@ -454,25 +454,47 @@ private static void removeSpans(Spannable raw) {
...
+            fullSize < claimedSpace + indeterminateColumns
+            || fullSize > claimedSpace && indeterminateColumns == 0
```
phillipm: Please use parenthesis so that we don't have to think about order of operations for || and &&

app/src/org/commcare/android/view/EntityView.java
```diff
@@ -442,7 +442,7 @@ private static void removeSpans(Spannable raw) {
     }
 
     private int[] calculateDetailWidths(int fullSize) {
```
phillipm: Even though this is private it would be nice to have method documentation.

After reading this method's code I'm still curious about: 
 - what the entries in the returned array represent?
 - what indeterminate columns are?
 - where hints come from and what type of hint parsing is going on to calculate measurements?

app/src/org/commcare/android/view/EntityView.java
```diff
@@ -442,7 +442,7 @@ private static void removeSpans(Spannable raw) {
     }
 
     private int[] calculateDetailWidths(int fullSize) {
-        // Convert any percentages to pixels
+        // Convert any percentages to pixels. Percentage columns are treated as percentage of the entire screen width.
         int[] hints = new int[mHints.length];
```
phillipm: Is there a better name for this variable? maybe columnWidths?

orangejenny:
@phillipm Cleaned up. Added the 2.22 bugfix (the second super.onMeasure call) during testing, figured it's not hurting anything to leave it in there.

phillipm:
Looks great, the calculateDetailWidths doc helps a lot!

PR #413, opened 2015-07-13
Fixing str_version issue

dcluna:
Fixing overlap in version string (bug [17006](http://manage.dimagi.com/default.asp?176006)).

### DIFF
--- a/app/res/layout-land/mainnew.xml
+++ b/app/res/layout-land/mainnew.xml
@@ -44,7 +44,8 @@
                     android:layout_height="wrap_content"
                     android:drawableLeft="@drawable/pencil"
                     android:text="Start" />
-
+
+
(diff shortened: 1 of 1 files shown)


app/res/layout-land/mainnew.xml
```diff
@@ -78,15 +79,19 @@
                             android:text="Saved" />
                     </LinearLayout>
                 </LinearLayout>
-
+                <!-- empty view with weight 1 to ensure str_version is bottom-aligned -->
+                <View
```
wpride: Odd workaround - could you find an SO post or anything on why this is necessary?

dcluna: http://stackoverflow.com/a/5694449/1536138

PR #414, opened 2015-07-13
Fix app crash in LoginActivity & persistently log any task crashes

phillipm:
Don't try to switch on a null value, which occurs when the DataPullTask launched by LoginActivity errors out. Noticed this when looking at this crash from the play store crash reports: 
```
java.lang.NullPointerException
at org.commcare.dalvik.activities.LoginActivity$3.deliverResult(LoginActivity.java:181)
at org.commcare.dalvik.activities.LoginActivity$3.deliverResult(LoginActivity.java:173)
at org.commcare.android.tasks.templates.CommCareTask.onPostExecute(CommCareTask.java:83)
at android.os.AsyncTask.finish(AsyncTask.java:417)
at android.os.AsyncTask.access$300(AsyncTask.java:127)
at android.os.AsyncTask$InternalHandler.handleMessage(AsyncTask.java:429)
at android.os.Looper.loop(Looper.java:130)
at android.app.ActivityThread.main(ActivityThread.java:3821)
at java.lang.reflect.Method.invokeNative(Native Method)
at java.lang.reflect.Method.invoke(Method.java:507)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:839)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:597)
at dalvik.system.NativeStart.main(Native Method)
```

Also log task crashes persistently.

### DIFF
--- a/app/src/org/commcare/android/tasks/DataPullTask.java
+++ b/app/src/org/commcare/android/tasks/DataPullTask.java
@@ -70,8 +70,6 @@ import android.util.Log;
  * @author ctsims
  */
 public abstract class DataPullTask<R> extends CommCareTask<Void, Integer, Integer, R> implements CommCareOTARestoreListener {
-    private static final String TAG = DataPullTask.class.getSimpleName();
-
     String server;
     String keyProvider;
     String username;
@@ -116,7 +114,6 @@ public abstract class DataPullTask<R> extends CommCareTask<Void, Integer, Intege
     private static final boolean DEBUG_LOAD_FROM_LOCAL = false;
     private InputStream mDebugStream;
 
-    
     public DataPullTask(String username, String password, String server, String keyProvider, Context c) {
         this.server = server;
         this.keyProvider = keyProvider;
@@ -124,6 +121,8 @@ public abstract class DataPullTask<R> extends CommCareTask<Void, Integer, Intege
         this.password = password;
         this.c = c;
         this.taskId = DATA_PULL_TASK_ID;
+
+        TAG = DataPullTask.class.getSimpleName();
     }
 
     /* (non-Javadoc)
--- a/app/src/org/commcare/android/tasks/templates/CommCareTask.java
+++ b/app/src/org/commcare/android/tasks/templates/CommCareTask.java
@@ -2,23 +2,25 @@ package org.commcare.android.tasks.templates;
 
 import android.os.AsyncTask;
 
+import org.javarosa.core.services.Logger;
+
 /**
  * @author ctsims
- *
  */
 public abstract class CommCareTask<A, B, C, R> extends ManagedAsyncTask<A, B, C> {
-    
+    protected static String TAG;
+
     public static final int GENERIC_TASK_ID = 32;
     
     private Object connectorLock = new Object();
     private CommCareTaskConnector<R> connector;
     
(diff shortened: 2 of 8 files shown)


app/src/org/commcare/android/tasks/templates/CommCareTask.java
```diff
@@ -30,8 +32,12 @@ protected final C doInBackground(A... params) {
         try {
             return doTaskBackground(params);
         } catch(Exception e) {
+            Logger.log(TAG, e.getMessage());
```
phillipm: Log task crashes.
Also, wrapping an entire method in a generic 'Exception' catch is poor form, so we should fix that one at some point.

app/src/org/commcare/android/tasks/templates/CommCareTask.java
```diff
@@ -152,14 +158,10 @@ protected void onProgressUpdate(B... values) {
             //something from connecting in the mean time
             if(connector != null) { return connector; }
             
-            //Otherwise, cancel if possible
             if(this.getStatus() == AsyncTask.Status.RUNNING) {
                 this.cancel(false);
             }
-            //Mark this as a lost orphan
-            this.isLostOrphan = true;
```
phillipm: dead code

app/src/org/commcare/dalvik/activities/LoginActivity.java
```diff
@@ -226,13 +226,14 @@ private void startOta() {
...
-                     * (non-Javadoc)
-                     * @see org.commcare.android.tasks.templates.CommCareTask#deliverResult(java.lang.Object, java.lang.Object)
-                     */
                     @Override
                     protected void deliverResult( LoginActivity receiver, Integer result) {
+                        if (result == null) {
+                            // The task crashed unexpectedly
+                            receiver.raiseLoginMessage(StockMessages.Restore_Unknown, true);
+                            return;
+                        }
```
phillipm: Check for null result before trying to switch on it, which causes a NPE

app/src/org/commcare/android/tasks/DataPullTask.java
```diff
@@ -124,6 +121,8 @@ public DataPullTask(String username, String password, String server, String keyP
         this.password = password;
         this.c = c;
         this.taskId = DATA_PULL_TASK_ID;
+
+        TAG = DataPullTask.class.getSimpleName();
```
dcluna: Why this over using the old static final pattern?

phillipm: The error logging occurs in the parent class, but we want the TAG to be set to the child class's name so in most cases the extending class sets the TAG in its constructor. There are some cases where the extending class doesn't provide its own constructor (see https://github.com/dimagi/commcare-odk/blob/master/app/src/org/commcare/android/tasks/MultimediaInflaterTask.java#L17), so I was lazy and let the parent constructor initialize TAG to the general "CommCareTask". This pattern doesn't allow for the use of 'final'.

If you would like I can add explicit constructors to all the abstract extensions of CommCareTask so that the TAG field is set correctly.

dcluna: No need, just curious. Thanks!

PR #420, opened 2015-07-16
Don't show task progress dialog when activity is in bg

phillipm:
When you log in (initially) a data pull task is spun up to download a bunch of data. If the device goes to sleep while this is running then the task crashes due to a fragment state loss error. This causes the app to log the user out.

The solution is to not try to show a dialog when the corresponding activity has been stopped. This PR introduces flags that keep track of the activities lifecycle state to prevent this.

Also fixed user error notification when login fails, such that it doesn't also show a bad network error.

Fix for http://manage.dimagi.com/default.asp?175212


app/src/org/commcare/dalvik/activities/LoginActivity.java
```diff
@@ -244,9 +244,8 @@ protected void deliverResult( LoginActivity receiver, Integer result) {
                         case DataPullTask.DOWNLOAD_SUCCESS:
                             if(!tryLocalLogin(true)) {
                                 receiver.raiseLoginMessage(StockMessages.Auth_CredentialMismatch, true);
-                            } else {
-                                break;
                             }
+                            break;
```
phillipm: Pretty sure we always want to break here, not just when the local login succeeds. Otherwise when the login fails it will show two error messages: bad auth credentials and bad network connect.

ctsims: Indeed. Sheel actually reported this bug recently.

phillipm: @ctsims This doesn't fix Sheel's bug, but I'll look into it.

app/src/org/commcare/android/framework/CommCareActivity.java
```diff
@@ -73,10 +78,30 @@
     public static final String KEY_LAST_QUERY_STRING = "LAST_QUERY_STRING";
     protected String lastQueryString;
 
+    /**
+     * Activity has been put in the background. This is used to prevent dialogs
+     * from being shown while activity isn't active.
+     */
+    private boolean activityStopped;
```
amstone326: being picky, but should change this to activityPaused now that it's in onPause

phillipm: good call, updated

PR #425, opened 2015-07-16
Install/upgrade related comments & refactor

phillipm:
Comments for methods that deal with resource installation

 - Add missing Override annotations
 - Change an if/else into a switch statement
 - Lots of comments
 - Flatten identical 'catch' bodies


app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -371,15 +375,12 @@ public void stopBlockingForTask(int id) {
     }
 
     /**
-     * @param startOverUpgrade what determines whether
-     *                         CommCarePlatform.stageUpgradeTable() reuses the
-     *                         last version of the upgrade table, or starts
-     *                         over
+     * @param startOverUpgrade If set CommCarePlatform.stageUpgradeTable()
+     *                         will reuse the last version of the upgrade
+     *                         table, or clear it and start over
```
amstone326: This is actually the reverse, should say something like: "If true, CommCarePlatform.stageUpgradeTable() will clear the last version of the upgrade table and start over. Otherwise, will reuse the last version of the upgrade table.

PR #427, opened 2015-07-17
Adding on click effects to SquareButtonWithText

dcluna:
To give feedback to users when they click the buttons in the home screen, SquareButtonWithText now shows a bluish color when pressed.


ctsims:
Luna: thanks for getting to this!

How did you settle on the clicking behavior? Is there a material design
guideline for this? I ask because since these have a "floating" affordance
I would have assumed that the click would have (also?) done something like
shift the button down closer to its shadow to simulate a downward press.

-Clayton

PR #432, opened 2015-07-21
Provide a Better Error if Phone Time Doesn't Match Server

phillipm:
Fix error handling for when user tries to log in when their device has an incorrect date.

Other small refactors too

In response to this ticket: http://manage.dimagi.com/default.asp?175197


app/src/org/commcare/android/tasks/ManageKeyRecordTask.java
```diff
@@ -325,16 +320,17 @@ protected boolean processSuccesfulRequest() {
...
-        UserKeyRecord current = getCurrentValidRecord(app, username, password, !HttpCalloutNeeded() || (calloutFailed && !HttpCalloutRequired()));
+        boolean acceptExpiredCredentials = !HttpCalloutNeeded() || (calloutFailed && !HttpCalloutRequired());
+        UserKeyRecord current = getCurrentValidRecord(app, username, password, acceptExpiredCredentials);
         
-        if(current == null)  {
-            //TODO: What is this failure mode
+        if (current == null)  {
+            // existance of username and validity of password have already
+            // been checked, so this means the user record isn't valid.
+            return HttpCalloutOutcomes.BadCertificate;
```
phillipm: The fix

phillipm:
on talking to @ctsims, I need to: 
 * call getCurrentValidRecord with acceptExpired set to true
 * return HttpCalloutOutcomes.UnknownError if current == null, which shouldn't be a state we can get into
 * write a  note as to why we want to acceptExpired UserKeyRecords: We want to eventually enforce expiration, but can't do it until we proactively refresh key records that are going to expire in 3 or so months, otherwise field workers outside of internet service for longer than that will not be able to log in

phillipm:
@ctsims This is ready to review again

PR #433, opened 2015-07-21
Acra reports 2

wpride:
@ctsims this is ready to be merged, ACRA backend all set


app/src/org/commcare/android/util/FileUtil.java
```diff
@@ -395,4 +396,20 @@ public static String getDumpDirectory(Context c){
                 return null;
             }
         }
+
+    public static Properties loadProperties(Context c) throws IOException {
```
phillipm: Would it be more appropriate to have this live in ACRAUtil?

app/src/org/commcare/android/util/FileUtil.java
```diff
@@ -395,4 +396,20 @@ public static String getDumpDirectory(Context c){
...
+
+    public static Properties loadProperties(Context c) throws IOException {
+        String[] fileList = { "local.properties" };
+        Properties prop = new Properties();
+        for (int i = fileList.length - 1; i >= 0; i--) {
+            String file = fileList[i];
+            try {
+                InputStream fileStream = c.getAssets().open(file);
+                prop.load(fileStream);
+                fileStream.close();
```
phillipm: the fileStream.close() should be in a 'finally' block, right?

app/src/org/commcare/android/util/FileUtil.java
```diff
@@ -395,4 +396,20 @@ public static String getDumpDirectory(Context c){
                 return null;
             }
         }
+
+    public static Properties loadProperties(Context c) throws IOException {
+        String[] fileList = { "local.properties" };
+        Properties prop = new Properties();
+        for (int i = fileList.length - 1; i >= 0; i--) {
```
phillipm: Is there a reason this is a for-loop even though there is only one file in the array of files?

phillipm: Is there a particular reason you iterating backwards through the list?

phillipm: Whenever possible use a for each loop instead of manually looping through the indices of an array.

app/src/org/commcare/dalvik/activities/ReportProblemActivity.java
```diff
@@ -40,21 +40,48 @@ public void onClick(View v) {
...
+    }
+
+    public static String getVersion(){
         try {
-            username = CommCareApplication._().getSession().getLoggedInUser().getUsername();
-        } catch (SessionUnavailableException e) {
-            username = "unknown_username";
+            return CommCareApplication._().getCurrentVersionString();
+        } catch(NullPointerException e){
+            return "Version not set.";
```
phillipm: Not convinced that getCurrentVersionString() throws a NullPointerException.

app/src/org/commcare/dalvik/activities/ReportProblemActivity.java
```diff
@@ -40,21 +40,48 @@ public void onClick(View v) {
...
+            return "Domain not set.";
+        }
+    }
+
+    public static String getPostURL(){
+        try{
+            SharedPreferences prefs = CommCareApplication._().getCurrentApp().getAppPreferences();
+            return prefs.getString(HttpRequestGenerator.USER_DOMAIN_SUFFIX, "not found");
+        } catch(NullPointerException e){
+            return "PostURL not set.";
```
phillipm: Would be good to be more explicit about error handling.
Maybe:
CommCareApp currentApp = CommCareApplication._().getCurrentApp();
if (currentApp == null) {
    return "No installed apps";
}
return currentApp.getAppPreferences().getString(HttpRequestGenerator.USER_DOMAIN_SUFFIX, "PostUrl not found");

app/src/org/commcare/dalvik/activities/ReportProblemActivity.java
```diff
@@ -40,21 +40,48 @@ public void onClick(View v) {
...
-        } catch (SessionUnavailableException e) {
-            username = "unknown_username";
+            return CommCareApplication._().getCurrentVersionString();
+        } catch(NullPointerException e){
+            return "Version not set.";
         }
-        String version = CommCareApplication._().getCurrentVersionString();
+    }
+
+    public static String buildMessage(String userInput){
```
phillipm: This method can be private

app/src/org/commcare/android/util/ACRAUtil.java
```diff
@@ -0,0 +1,21 @@
...
+ * Created by wpride1 on 3/3/15.
+ */
+public class ACRAUtil {
+
+    public static final String POST_URL = "post_url";
+    public static final String VERSION = "version";
+    public static final String DOMAIN = "domain";
+    public static final String USERNAME = "username";
+
+    public static void addCustomData(String key, String value){
```
phillipm: Add comment like:
Add debugging data to the acra log. The latest value is kept for each key

app/src/org/commcare/android/util/ACRAUtil.java
```diff
@@ -0,0 +1,21 @@
+package org.commcare.android.util;
+
+import org.acra.ACRA;
+import org.acra.ErrorReporter;
+
+/**
+ * Created by wpride1 on 3/3/15.
```
phillipm: Replace comment template with a real comment, maybe something like:
/** 
* Specifies the custom data collected by ACRA error reporting.
*
* @author wpride
*/

app/src/org/commcare/dalvik/application/CommCareApplication.java
```diff
@@ -198,6 +210,22 @@ public void onCreate() {
...
         resourceState = initializeAppResources();
+
+        ACRA.init(this);
+        this.initACRA();
+    }
+
+    public void initACRA(){
+        try {
+            Properties properties = FileUtil.loadProperties(this);
+            ACRAConfiguration mAcraConfig = ACRA.getConfig();
```
phillipm: Don't prefix non-member variables with a 'm'

app/src/org/commcare/dalvik/application/CommCareApplication.java
```diff
@@ -198,6 +210,22 @@ public void onCreate() {
 
         //The fallback in case the db isn't installed 
         resourceState = initializeAppResources();
+
+        ACRA.init(this);
+        this.initACRA();
+    }
+
+    public void initACRA(){
```
phillipm: This can be static and live in ACRAUtils. (Just have it take a Context argument)

app/src/org/commcare/android/util/FileUtil.java
wpride: at the moment possibly it would but I think its more generically useful if we need to use other keys or local properties in other areas in the future

app/src/org/commcare/dalvik/activities/ReportProblemActivity.java
wpride: this is actually catching the case where CommCareApplication()._() is not instantiated yet IE crash before you install

wpride:
@phillipm thanks, updated

app/src/org/commcare/dalvik/activities/ReportProblemActivity.java
phillipm: The CommCareApplication should always be the first thing that runs when the phone app starts, right? CommCareApplication._() shouldn't return null unless you call it before CommCareApplication.onCreate() finishes, which shouldn't be the case here.

app/src/org/commcare/android/util/ACRAUtil.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public static final String POST_URL = "post_url";
+    public static final String VERSION = "version";
+    public static final String DOMAIN = "domain";
+    public static final String USERNAME = "username";
+
+    /**
+     * Add debugging value to the ACRA report bundle. Only most recent value stored for each key.
+     * @param key
+     * @param value
```
phillipm: Please don't include 'param' entries if they don't have any info included.

app/src/org/commcare/android/util/FileUtil.java
phillipm: Don't we already have a lot of property-like loading, such as android preferences? We can always generalize property loading in the future, but since we aren't sure that we will ever need it, I think it is better to be very specific.

I would recommend changing the method name to 'loadAcraProperties', moving it to ACRAUtils, and changing the name of the properties file to  'acra.properties'.

Feel free to disagree and leave this here, but that is my position on this.

app/src/org/commcare/dalvik/activities/ReportProblemActivity.java
wpride: Yeah actually I think you're right on that one

However the reporting was failing when I tested crashes at certain points during/before install, so I threw this error handling (though  can't remember the exact cause now). I'm actually now leaning towards making these all catch just a generic Exception so that we never fail to report due to crashes in generating the per-user info

wpride:
@phillipm actually realized it'd be safer to add the keys as part of the ant build rather than as a resource so local.properties discussion is moot

phillipm:
Sweet, thanks for making all those fixes!

PR #434, opened 2015-07-21
CommCare crashes when you mistype bit.ly url during installation

phillipm:
If you mistype a bit.ly url such that it redirects to a url that doesn't point to a profile file, CommCare crashes in a nondescript manner.

This was because SetupActivity was returning a OK result code even when it errored-out. Fix is to return CANCELED code and make the home activity re-launch the install screen on aborted installs.

Fix for http://manage.dimagi.com/default.asp?176536


ctsims:
I think this also may have gotten fixed in 2.22. Should test that before we
merge.

phillipm:
Yup, it was

phillipm:
@ctsims even though I couldn't reproduce the bug on 2.22, I think this PR is a valid refactor because failed installations were returning RESULT_OK to the home activity, which isn't good form.

amstone326:
Agreed with that. I ran into this issue in multiple apps and was going to introduce a different (probably less correct) way of indicating that an install had failed to the origin activity. So this would be good.

PR #439, opened 2015-07-22
Reorganize user session handling & remove login popup on session expiration

phillipm:
User session expiration currently pops up a the CommCare login screen, which is annoying if the user is in another app when the session expires.

Refactor how session expiration triggers the opening of the login screen. This is unfortunately more complicated than one might wish, here is what is up:
 * Make everything that extends FragmentActivity actually extend SessionAwareFragmentActivity
 * Make everything that extends PreferenceActivity actually extend SessionAwarePreferenceActivity
 * For classes that should be session aware but extend random activities, such as FormHierarchyActiviy extending ListActivity, add calls to  handleOrListenForSessionExpiration/unregisterSessionExpirationReceiver in onResume/onPause.
 * on expiration, broadcast an intent to activities registered with SessionActivityRegistration code, or those that extend SessionAware*
 * During SessionAware*.onResume:
   * register a BroadcastReceiver to catch the aforementioned intent and redirect to the login screen.
   * check if the session has expired while the fragment wasn't active. This is needed because the registered BroadcastReceiver won't receive anything while SessionAwareFragmentActivity is in the background.
 * During SessionAware*.onPause, unregister the BroadcastReceiver

I decided not to make GeoPoint related activities or the signature collecting activity be session aware, since when they return to the FormEntryActivity, the user will be logged out by the fact that form entry is session aware. Can you think of any activities I'm missing?

Also contains the following refactors:
* Rename a bunch of session management related methods to hopefully be clearer.
* Moves the main logout functionality from CommCareSessionService to CommCareApplication, which let's us avoid a lot of unnecessary error handling!
* Distinguish between log-outs that are for maintenance (releaseUserResourcesAndServices) and those that are triggered by the user or session expiration (closeUserSession). This is needed because normal log-outs should cancel any running async tasks, while maintenance logouts, which are called from tasks such as DataPullTask and ManageKeyRecordTask, shouldn't cancel running async tasks. Or else they would cancel themselves!


app/src/org/commcare/android/framework/SessionAwareFragmentActivity.java
```diff
@@ -0,0 +1,76 @@
...
+import android.support.v4.app.FragmentActivity;
+
+import org.commcare.dalvik.activities.LoginActivity;
+
+/**
+ * Manage redirection to login screen when session expiration occurs.
+ *
+ * @author Phillip Mates (pmates@dimagi.com)
+ */
+public abstract class SessionAwareFragmentActivity extends FragmentActivity {
```
phillipm: @ctsims I created this fragment activity that checks to see if the session has expired and the user needs to redirect to the login screen. 

I've made CommCareActivity extend it. Can you think of any other classes that extend FragmentActivity that would need to extend it?

phillipm:
@amstone326 Here is the refactor that stops the login screen from popping up when the user is logged out. Feel free to take a look at it, especially since I renamed some functions that you probably use in MultipleApps. I'm assigning it to @ctsims though because the PR effects a large part of the app.

app/src/org/commcare/android/framework/SessionAwareFragmentActivity.java
```diff
@@ -0,0 +1,25 @@
...
+
+import android.os.Bundle;
+import android.support.v4.app.FragmentActivity;
+
+/**
+ * Manage redirection to login screen when session expiration occurs.
+ *
+ * @author Phillip Mates (pmates@dimagi.com)
+ */
+public abstract class SessionAwareFragmentActivity extends FragmentActivity {
```
phillipm: Is there some fancy thing I can do to avoid duplicating the code here in SessionAwarePreferenceActivity as well?

PR #442, opened 2015-07-23
Migrate ODK to Gradle

dcluna:
Migrating ODK to Gradle.

Depends on #109.


build.gradle
```diff
@@ -0,0 +1,144 @@
...
+        minSdkVersion 9
+        targetSdkVersion 21
+    }
+
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_7
+        targetCompatibility JavaVersion.VERSION_1_7
+    }
+
+    def sourceLocations = ['app/src'
```
ctsims: Shouldn't the location of ../commcare be parameterized somehow/somewhere?

gradlew.bat
```diff
@@ -0,0 +1,90 @@
+@if "%DEBUG%" == "" @echo off
```
ctsims: should these files be here?

settings.gradle
```diff
@@ -0,0 +1,3 @@
+include ':libraries:mapballoons'
+include ':javarosa'
+project(':javarosa').projectDir = new File('../javarosa/core')
```
ctsims: CommCare's project dir should also be set up like this, right?

gradlew.bat
dcluna: Yes, it's the gradle wrapper that runs with Android Studio

settings.gradle
dcluna: Only if it has a build.gradle (currently it doesn't)

PR #444, opened 2015-07-24
Gradle working with jenkins.... in a real jenky way.

phillipm:
Configure gradle build such that javarosa and commcare projects are only depended upon during debug builds (assembleDebug or buildDebug). For release builds, the build script expects javarosa and commare jars in app/libs.

With this, Jenkins can build the project, in a sadly jenky way! It manually copies a settings.gradle file that points to javarosa-core-library/core and commcare-mobile in the jenkins machine's workspace. This is needed because I can't figure out how to remove source level dependencies completely (if those project dirs aren't present, gradle complains even if we are doing a assembleRelease)

__Whoever merges this PR needs to change the branch of the [commcare-odk](http://jenkins.dimagi.com/job/commcare-odk/) build back to */master__

Removed the gradle wrapper related files, because I don't think those should be in the repo.

Upgrade to Google API 22

Future work: The build setup doesn't inject build numbers or file.key.properties


build.gradle
```diff
@@ -62,21 +65,6 @@ project(':libraries:mapballoons') {
...
-    buildToolsVersion rootProject.ext.buildToolsVersion
-  }
-}
-
-project(':commcare') {
-  android{
-    compileSdkVersion rootProject.ext.compileSdkVersion
-    buildToolsVersion rootProject.ext.buildToolsVersion
-  }
-}
```
phillipm: These aren't needed because they don't have android code in them

build.gradle
```diff
@@ -39,20 +39,23 @@ repositories {
...
+    // release build type expects javarosa and commcare jars to be in app/libs
+    debugCompile project(':javarosa')
+    debugCompile project(':commcare')
 }
 
 project.ext {
-    compileSdkVersion = "Google Inc.:Google APIs:21"
-    buildToolsVersion = "21.1.2"
+    compileSdkVersion = "Google Inc.:Google APIs:22"
+    buildToolsVersion = "22.0.1"
```
phillipm: Bump to API 22s

PR #446, opened 2015-07-24
Switching HeaderFooterGridView for StaggeredGridView

dcluna:
To avoid the bug in which the 'Logged In:' text gets cut off.

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -426,7 +426,7 @@
         android:name="android.permission.USE_CREDENTIALS" />
 
     <uses-sdk
-        android:minSdkVersion="8"
+        android:minSdkVersion="10"
         android:targetSdkVersion="14" >
     </uses-sdk>
 
--- a/build.gradle
+++ b/build.gradle
@@ -66,11 +66,6 @@ project(':libraries:mapballoons') {
   }
 }
 
-ext {
-  compileSdkVersion = "Google Inc.:Google APIs:21"
-  buildToolsVersion = "21.1.2"
-}
-
 android {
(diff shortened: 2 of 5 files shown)


app/AndroidManifest.xml
```diff
@@ -426,7 +426,7 @@
         android:name="android.permission.USE_CREDENTIALS" />
 
     <uses-sdk
-        android:minSdkVersion="8"
+        android:minSdkVersion="10"
```
dcluna: this is required by staggeredgridview

build.gradle
```diff
@@ -65,6 +66,11 @@ project(':libraries:mapballoons') {
   }
 }
 
+ext {
+  compileSdkVersion = "Google Inc.:Google APIs:21"
+  buildToolsVersion = "21.1.2"
```
phillipm: This needs to be 22

build.gradle
```diff
@@ -86,8 +92,8 @@ android {
     }
 
     defaultConfig {
-        minSdkVersion 9
-        targetSdkVersion 22
+        minSdkVersion 10
+        targetSdkVersion 21
```
phillipm: 22

app/AndroidManifest.xml
ctsims: We can't change this, unfortunately, our app needs to be able to target SDK 8, and last time we changed this a few projects got blocked.

dcluna: Then we cannot merge this PR. I'll search for other alternatives.

dcluna:
Closed because we need to support sdk version 8

PR #449, opened 2015-07-24
Allow for Jenkins to sign apk builds via gradle

phillipm:
Read in signing and other sensitive properties from ~/.gradle/gradle.properties on the build server.

If that file isn't present, like on local machines, use default values specified in build.gradle

Also remove the GridViewWithHeaderAndFooter git submodule, because gradle is downloading the jar


phillipm:
@wpride You should probably be the one to review this since it modifies some build property setup that you put into place for the old build system.

PR #455, opened 2015-07-28
Gradle: Set jni dir instead of using native lib task

phillipm:
After pull https://github.com/dimagi/commcare-odk/pull/453 I was experiencing some build issues.

The first error I got was multiple dex files error. I fixed that by disabling incremental dexing.
The second error I got was stlport_shared.so missing, fixed that by setting the jniLibs dirs.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -156,7 +156,6 @@ android {
             buildConfigField "String", "MAPS_API_KEY", "\"${project.ext.MAPS_API_KEY}\""
             signingConfig signingConfigs.release
             java.srcDirs = sourceLocations + ['release']
-            // minifyEnabled true
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'app/proguard.cfg'
           }
         }
--- a/build.gradle
+++ b/build.gradle
@@ -100,7 +100,6 @@ android {
 
     dexOptions {
         preDexLibraries = true
-        incremental true
     }
 
     defaultConfig {
--- a/build.gradle
+++ b/build.gradle
@@ -20,19 +20,6 @@ buildscript {
 
 apply plugin: 'com.android.application'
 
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -134,6 +119,7 @@ android {
 
     sourceSets {
       main {
+        jniLibs.srcDirs = ['app/libs']
```
phillipm: Setting this removes the need for nativeLibsToJar task

build.gradle
```diff
@@ -161,6 +147,7 @@ android {
             buildConfigField "String", "ACRA_PASSWORD", "\"${project.ext.ACRA_PASSWORD}\""
             buildConfigField "String", "ACRA_USER", "\"${project.ext.ACRA_USER}\""
             buildConfigField "String", "MAPS_API_KEY", "\"${project.ext.MAPS_API_KEY}\""
+            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'app/proguard.cfg'
```
phillipm: I think this is needed here, in the debug build too, but I'm not positive sure.

build.gradle
```diff
@@ -169,7 +156,6 @@ android {
             buildConfigField "String", "MAPS_API_KEY", "\"${project.ext.MAPS_API_KEY}\""
             signingConfig signingConfigs.release
             java.srcDirs = sourceLocations + ['release']
-            minifyEnabled true
```
phillipm: Do we want this or not.

build.gradle
```diff
@@ -115,7 +100,6 @@ android {
 
     dexOptions {
         preDexLibraries = true
-        incremental true
```
phillipm: This will slow down builds, but I was getting multiple dex errors during building when it was enabled

PR #456, opened 2015-07-28
Fixing install screen button issue

dcluna:
Non-stretched buttons in the install screens.

### DIFF
--- a/app/src/org/commcare/android/view/SquareButton.java
+++ b/app/src/org/commcare/android/view/SquareButton.java
@@ -24,17 +24,19 @@ public class SquareButton extends ImageButton {
     protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
 
         //Get canvas width and height
-        int w = MeasureSpec.getSize(widthMeasureSpec);
-        int h = MeasureSpec.getSize(heightMeasureSpec);
+        int width = MeasureSpec.getSize(widthMeasureSpec);
+        int height = MeasureSpec.getSize(heightMeasureSpec);
 
-        if(w == 0 || h == 0) { // this can happen in the home screen gridView, and then we'll have an invisible view
-            w = Math.max(w, h);
+        if (width == 0 || height == 0) {
+            /*
+             * If this happens (for instance, the home screen GridView tries to force all its child views to have a height of 0),
+             * we avoid making this button invisible by picking the greatest of the two dimensions.
+             */
+            width = Math.max(width, height);
         } else {
-            w = Math.min(w, h);
+            width = Math.min(width, height);
         }
 
-        h = w;
-
-        setMeasuredDimension(w,w);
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/android/view/SquareButton.java
```diff
@@ -27,7 +27,12 @@ protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
...
         int h = MeasureSpec.getSize(heightMeasureSpec);
 
-        w = Math.max(w, h);
+        if(w == 0 || h == 0) { // this can happen in the home screen gridView, and then we'll have an invisible view
+            w = Math.max(w, h);
+        } else {
+            w = Math.min(w, h);
+        }
+
         h = w;
```
phillipm: This assignment seems useless

app/src/org/commcare/android/view/SquareButton.java
```diff
@@ -27,7 +27,12 @@ protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
         int w = MeasureSpec.getSize(widthMeasureSpec);
         int h = MeasureSpec.getSize(heightMeasureSpec);
```
phillipm: Are these the canvas size or the button size?

app/src/org/commcare/android/view/SquareButton.java
```diff
@@ -27,7 +27,12 @@ protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
...
 
-        w = Math.max(w, h);
+        if(w == 0 || h == 0) { // this can happen in the home screen gridView, and then we'll have an invisible view
```
phillipm: I'm sort of irrationally picky about this, but can you put comments on their own lines, and inside of the conditional body that they correspond to?

app/src/org/commcare/android/view/SquareButton.java
dcluna: It is

app/src/org/commcare/android/view/SquareButton.java
phillipm: You can avoid the comment by naming 'w' and 'h' to 'canvasWidth' and 'canvasHeight'

dcluna: Button size

app/src/org/commcare/android/view/SquareButton.java
phillipm: Please put a space in between 'if' and the paren. 'if' is not a function, but rather a keyword, so we shouldn't be using functional call syntax on it.

phillipm: This comment is pretty confusing. When do we have an invisible view? Are we avoiding an invisible view?

PR #462, opened 2015-07-29
Enter url screen fixes

dcluna:
Makes the 'Enter URL' textbox single-line and correctly shows the 'Enter URL' screen when resuming.


app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -244,7 +246,19 @@ private void uiStateScreenTransition() {
                 break;
 
             default:
```
phillipm: I don't like that you are re-instating the url chooser fragment in the default case, given that UiState.advanced is the dedicated UiState for being in the url chooser fragment. Couldn't you just set that when the chooser is launched and use it to properly restore?

phillipm: And better yet, rename UiState.advanced to UiState.choosing_url or something, because 'advanced' is a horrible name.

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -244,7 +246,19 @@ private void uiStateScreenTransition() {
...
+                if(lastIndex > -1) {
+                    fragment = fgmts.get(lastIndex);
+                    if (BuildConfig.DEBUG) {
+                        Log.v(TAG, "Last fragment: " + fragment);
+                    }
+                }
+                if(!(fragment instanceof SetupEnterURLFragment)){
+                    fragment = installFragment;
+                } else {
+                    // we restore the state of the SetupEnterURLFragment if it is our last fragment
```
phillipm: Please avoid empty conditional bodies. You can change the comment to not need them:
```
if(!(fragment instanceof SetupEnterURLFragment)){
    // last fragment wasn't url entry, so default to the installation method chooser
    fragment = installFragment;
}
```
My comment in the example above is a little over the top, you should really be able to use var names that self-document what is going on.

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -244,7 +246,19 @@ private void uiStateScreenTransition() {
...
+                }
```
phillipm: There is a fair amount of logic going on here. Could you move it into a helper method with a good name so that it is easier to understand what the 'default' case is achieving?
Maybe:
```
fragment = restoreInstallSetupFragment()
...
private Fragment restoreInstallSetupFragment() { ...
```

phillipm:
Something else I noticed was the fact that uiStateScreenTransition was called at the end of onCreate and the beginning of onStart. Pretty sure you can take out the onCreate call to uiStateScreenTransition, since it will always be preceded by the onStart call.

app/src/org/commcare/android/fragments/SetupInstallFragment.java
```diff
@@ -54,6 +55,10 @@ public void onClick(View v) {
             @Override
             public void onClick(View v) {
                 SetupEnterURLFragment enterUrl = new SetupEnterURLFragment();
+                Activity currentActivity = getActivity();
+                if (currentActivity instanceof CommCareSetupActivity) {
+                    ((CommCareSetupActivity)currentActivity).setUiState(CommCareSetupActivity.UiState.IN_URL_ENTRY);
+                }
```
phillipm: Not the cleanest fix, but this can be fixed in future refactors, which I will do when for the background auto-update feature I'm working on.

phillipm:
@dcluna I think this PR fixes a few of the installation bugs we've been experiencing. I cleaned it up so that we can merge it.

PR #463, opened 2015-07-29
Update messages_ccodk_default.txt

orangejenny:
Pretty sure this doesn't belong...


phillipm:
For http://manage.dimagi.com/default.asp?177329

PR #464, opened 2015-07-29
Refactoring StateFragment

phillipm:
Cleaning up the StateFragment class.

Easiest reviewed commit by commit:
 - rename wakelock related methods to hopefully be more intuitive
 - remove comments that should be obvious by the code they are near
 - re-order methods to flow better
 - run Android Studio code formatter
 - move all class fields to top of class
 - break out test expression into helper method
 - make class fields private


app/src/org/commcare/android/framework/StateFragment.java
```diff
@@ -10,95 +10,94 @@
...
 
-    private WakeLock wakelock;
-
-    public synchronized void wakelock(int lockLevel) {
-        if(wakelock != null) {
-            if(wakelock.isHeld()) {
+    public synchronized void acquireWakeLock(int lockLevel) {
+        if (wakelock != null) {
+            if (wakelock.isHeld()) {
                 wakelock.release();
```
amstone326: Could this just call releaseWakeLock()?

app/src/org/commcare/android/framework/StateFragment.java
```diff
@@ -10,95 +10,94 @@
...
+
+        PowerManager pm = (PowerManager)boundActivity.getSystemService(Context.POWER_SERVICE);
         wakelock = pm.newWakeLock(lockLevel, "CommCareLock");
         wakelock.acquire();
     }
 
-    public synchronized void unlock() {
-        if(wakelock != null) {
+    public synchronized void releaseWakeLock() {
+        if (wakelock != null) {
```
amstone326: Should this also check for wakelock.isHeld(), since you are doing that before you call release() in acquireWakeLock?

app/src/org/commcare/android/framework/StateFragment.java
phillipm: In this case ```wakelock``` is an object defined in a library (android.os.PowerManager.WakeLock), so we can't change the method name.

phillipm: Oh, misunderstood that. Updated the PR.

PR #468, opened 2015-07-29
Square button simulating ImageButton effects

dcluna:
Simulating stock Android's ImageButton effects.


phillipm:
This looks fine for SquareButtonWithText but looks bad for SquareButtonWithNotification due to the shadowed backdrop. I think that should be addressed before merging this PR. Or I can merge it if you plan on addressing it later.

I personally like the old blue highlight color instead of this new orange one, but that is personal preference...

dcluna:
I also prefer the old blue color, maybe we can drop this PR? Not sure it adds much besides consistency with the previous style.

PR #469, opened 2015-07-30
Newui/show CommCare version as Snackbar

dcluna:
Shows the CommCare version as a Snackbar element with an optional button to copy the version to the clipboard.

CommCareActivity now extends AppCompatActivity, and this may cause some unexpected bugs. If this is not acceptable, I'll try another approach.

### DIFF
--- a/app/assets/locales/messages_ccodk_default.txt
+++ b/app/assets/locales/messages_ccodk_default.txt
@@ -70,6 +70,8 @@ home.menu.formdump=Manage SD
 home.menu.wifi.direct=Wifi Direct
 home.menu.connection.diagnostic=Connection Test
 home.menu.saved.forms=Saved Forms
+home.menu.version=CommCare Version
+home.version.copy=Copy
 
 install.version.mismatch=The application requires CommCare version ${0}. You are running ${1}.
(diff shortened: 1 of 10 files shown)


phillipm:
I get this error on a 2.3.7 device:
```
DEVICE SHELL COMMAND: pm install -r "/data/local/tmp/org.commcare.dalvik"
pkg: /data/local/tmp/org.commcare.dalvik
Failure [INSTALL_FAILED_MISSING_SHARED_LIBRARY]
```

Install works fine on a 5.1 device

dcluna:
The missing shared library error happens in a physical device or an emulator?

phillipm:
A physical device, though it is running cyanogenmod, so it is entirely possible it is missing some crucial google library. Works fine on a 4.4.2 tablet.

phillipm:
Just checked and it has the same problem on a gingerbread emulator too.

phillipm:
Actually, I'm also getting this error for older devices on master too. It is probably because my emulator wasn't using Google APIs and my old device is cyanogenmod, which probably has similar issues. Did you get it running on a gingerbread emulator?

phillipm:
I think this might introduce more bugs than it is worth. For instance, the app title isn't shown on the Login screen and background colors seem grey in places that they used to be white or black.

PR #470, opened 2015-07-31
Using compatibility library to avoid crashes in pre-HoneycombM2 devices

dcluna:
GridEntityView was extending the stock GridLayout, thus crashing in old devices. This PR fixes that.

### DIFF
--- a/app/src/org/commcare/android/view/GridEntityView.java
+++ b/app/src/org/commcare/android/view/GridEntityView.java
@@ -3,19 +3,6 @@
  */
 package org.commcare.android.view;
 
-import org.commcare.android.models.AsyncEntity;
-import org.commcare.android.models.Entity;
-import org.commcare.android.util.CachingAsyncImageLoader;
-import org.commcare.android.util.MarkupUtil;
-import org.commcare.dalvik.R;
-import org.commcare.suite.model.Detail;
-import org.commcare.util.GridCoordinate;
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/android/view/GridEntityView.java
```diff
@@ -142,11 +142,18 @@ public GridEntityView(Context context, Detail detail, Entity entity, String[] se
...
 		//setup all the various dimensions we need
 		Point size = new Point();
-		((Activity)context).getWindowManager().getDefaultDisplay().getSize(size);
-		
-		screenWidth = size.x;
-		screenHeight = size.y;
-		
+		Display display = ((Activity)context).getWindowManager().getDefaultDisplay();
+		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.HONEYCOMB_MR2)
+		{
```
amstone326: shouldn't be on separate line

app/src/org/commcare/android/view/GridEntityView.java
```diff
@@ -142,11 +142,18 @@ public GridEntityView(Context context, Detail detail, Entity entity, String[] se
 		
 		//setup all the various dimensions we need
 		Point size = new Point();
```
amstone326: Since this variable is only used within the if() clause, should move its declaration there for clarity.

dcluna:
Should be good to go now

PR #472, opened 2015-07-31
sms installs initial commit

wpride:


wpride:
@phillipm @czue re: email thread. Assumes the valid profile link will be between squiggly brackets

czue:
well that was fast.

wpride:
@gcapalbo

wpride:
Release Note: Enabled installation workflow from SMS messages

wpride:
@krsdimagi

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -550,22 +556,103 @@ protected void deliverError(CommCareSetupActivity receiver,
...
+     * and if found and verified return the discovered install link
+     * @return the verified install link, null if none found
+     * @throws SignatureException if we discovered a valid-looking message but could not verify it
+     */
+    public String scanSMSLinks() throws SignatureException{
+        String installLink = null;
+        final Uri SMS_INBOX = Uri.parse("content://sms/inbox");
+        Cursor cursor = getContentResolver().query(SMS_INBOX, null, null, null, "date desc limit 1");
+        if(cursor.moveToFirst()) {
+            Log.v("Body", cursor.getString(cursor.getColumnIndex("body")));
```
phillipm: Probably shouldn't log the body of every text.

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -550,22 +556,103 @@ protected void deliverError(CommCareSetupActivity receiver,
...
+            String textMessageBody = cursor.getString(cursor.getColumnIndex("body"));
+            if(textMessageBody.contains("ccapp:") && textMessageBody.contains("signature:")) {
+                installLink = parseAndVerifySMS(textMessageBody);
+            }
+        }
+        cursor.close();
```
phillipm: You'll want to use the try/finally or managed resource pattern to ensure that `cursor.close()` gets called.

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -550,22 +556,103 @@ protected void deliverError(CommCareSetupActivity receiver,
...
+
+
+    // http://stackoverflow.com/questions/11301046/search-sms-inbox
+
+    /**
+     * Scan the SMS inbox, looking for messages that meet the expected install format,
+     * and if found and verified return the discovered install link
+     * @return the verified install link, null if none found
+     * @throws SignatureException if we discovered a valid-looking message but could not verify it
+     */
```
phillipm: Nice!

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -550,22 +556,103 @@ protected void deliverError(CommCareSetupActivity receiver,
...
+            }
+        } catch(SignatureException e){
+            // possibly we want to do something more severe here? could be malicious
+            e.printStackTrace();
+            Toast.makeText(this, Localization.get("menu.sms.not.found"), Toast.LENGTH_LONG).show();
+        }
+    }
+
+
+    // http://stackoverflow.com/questions/11301046/search-sms-inbox
```
phillipm: Can you move this into the body of the method it is related to? That will ensure that it stays 'linked' to the code that uses it.

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -215,13 +220,14 @@ protected void onCreate(Bundle savedInstanceState) {
...
 
-        Log.v("UiState","Current vars: " +
+        Log.v("UiState", "Current vars: " +
                         "UIState is: " + this.uiState + " " +
                         "incomingRef is: " + incomingRef + " " +
                         "startAllowed is: " + startAllowed + " "
         );
 
         uiStateScreenTransition();
+        performSMSInstall(false);
```
phillipm: Are you sure you want this here?

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -550,22 +556,103 @@ protected void deliverError(CommCareSetupActivity receiver,
     public boolean onCreateOptionsMenu(Menu menu) {
         super.onCreateOptionsMenu(menu);
         menu.add(0, MODE_ARCHIVE, 0, Localization.get("menu.archive")).setIcon(android.R.drawable.ic_menu_upload);
+        menu.add(0, MODE_SMS, 1, Localization.get("menu.sms")).setIcon(android.R.drawable.ic_menu_upload);
```
phillipm: The icon that is set is the same as the `archive` menu entry found above. This is only relevant to Android 2 devices, but it is worth choosing something other than `ic_menu_upload`.

Maybe `stat_notify_chat` 
image
or something else from http://www.darshancomputing.com/android/1.5-drawables.html

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -550,22 +556,103 @@ protected void deliverError(CommCareSetupActivity receiver,
...
+                incomingRef = profileLink;
+                if (installAutomatically) {
+                    startResourceInstall();
+                } else {
+                    uiState = UiState.READY_TO_INSTALL;
+                    Toast.makeText(this, Localization.get("menu.sms.ready"), Toast.LENGTH_LONG).show();
+                }
+            } else {
+                // only notify if this was manually triggered, since most people won't use this
+                if(installAutomatically) {
```
phillipm: This instance of `else { if (...) { ... ` can be collapsed to `else if (...) { ...`

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -550,22 +556,103 @@ protected void deliverError(CommCareSetupActivity receiver,
...
+                    Toast.makeText(this, Localization.get("menu.sms.not.found"), Toast.LENGTH_LONG).show();
+                }
+            }
+        } catch(SignatureException e){
+            // possibly we want to do something more severe here? could be malicious
+            e.printStackTrace();
+            Toast.makeText(this, Localization.get("menu.sms.not.found"), Toast.LENGTH_LONG).show();
+        }
```
phillipm: Try to localize your try/catch code to wrap the minimal amount of code that needed. In this case, the only statement that throws `SignatureException` is `scanSMSLinks`, so the try body should only include that statement.

This is important because it will help the reader more easily understand what could fail and when.

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -550,22 +556,103 @@ protected void deliverError(CommCareSetupActivity receiver,
...
+        // signature: and that the end of the signature is the end of the text content
+        String downloadLink = text.substring(text.indexOf("ccapp:") + 7, text.indexOf(","));
+        String signature = text.substring(text.indexOf("signature:") + 11);
+        if(verifySMS(signature, downloadLink)){
+            return downloadLink;
+        }
+        throw new SignatureException();
+    }
+
+    public boolean verifySMS(String signature, String message){
```
phillipm: This method, and all the others you have added here can be made private.

app/src/org/commcare/android/util/SigningUtil.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    private static PublicKey getPublicKey(String key) throws Exception{
+        byte[] derPublicKey = Base64.decode(key);
+        X509EncodedKeySpec spec = new X509EncodedKeySpec(derPublicKey);
+        KeyFactory kf = KeyFactory.getInstance("RSA");
+        return kf.generatePublic(spec);
+    }
+
+    private static boolean verify(PublicKey publicKey, String messageString, String signature) throws SignatureException {
```
phillipm: Can you give this a different name? Overloading a method when one implementation is private and the other is public confuses me.

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
wpride: I think so - the goal of this is to basically make install as seamless as possible and since we have no ability to configure per domain at this point I think we basically have to have this here if we want automatic installs

app/src/org/commcare/android/util/SigningUtil.java
```diff
@@ -0,0 +1,57 @@
...
+     * @param publicKeyString the known public key of CCHQ
+     * @param message the message content
+     * @param messageSignature the signature generated by HQ with its private key and the message content
+     * @return whether or not the message was verified to be sent with HQ's private key
+     */
+    public static boolean verify(String publicKeyString, String message, String messageSignature) {
+        try {
+            PublicKey publicKey = getPublicKey(publicKeyString);
+            boolean verified = verify(publicKey, message, messageSignature);
+            return verified;
```
phillipm: `verified` is redundant, could be `return verify(publicKey, message, messageSignature);`

app/src/org/commcare/android/util/SigningUtil.java
```diff
@@ -0,0 +1,57 @@
...
+    private static boolean verify(PublicKey publicKey, String messageString, String signature) throws SignatureException {
+        try {
+            Signature sign = Signature.getInstance("SHA256withRSA/PSS", new BouncyCastleProvider());
+            byte[] signature_binary = Base64.decode(signature);
+            byte[] message = messageString.getBytes();
+            sign.initVerify(publicKey);
+            sign.update(message);
+            return sign.verify(signature_binary);
+        } catch (Exception ex) {
+            throw new SignatureException(ex);
```
phillipm: You throw this and then immediately swallow it in the caller method. You could remove this catch/raise SignatureException code and just let any exceptions thrown here be caught by the caller.

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -550,22 +556,103 @@ protected void deliverError(CommCareSetupActivity receiver,
...
+            e.printStackTrace();
+            Toast.makeText(this, Localization.get("menu.sms.not.found"), Toast.LENGTH_LONG).show();
+        }
+    }
+
+
+    // http://stackoverflow.com/questions/11301046/search-sms-inbox
+
+    /**
+     * Scan the SMS inbox, looking for messages that meet the expected install format,
```
phillipm: Important to note that this only looks at the latest sms. I would change the method name to reflect this as well.

phillipm:
I didn't look closely at any of the crypto code, since I'm not well versed with crypto.
Looks good otherwise; should be good after my comments are considered/addressed.

wpride:
@phillipm thanks for the fast and solid review

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -550,22 +556,108 @@ protected void deliverError(CommCareSetupActivity receiver,
...
+     *
+     * @param text the parsed out text message in the expected link/signature format
+     * @return the download link if the message was valid and verified, null otherwise
+     * @throws SignatureException if we discovered a valid-looking message but could not verifyMessageSignatureHelper it
+     */
+    private String parseAndVerifySMS(String text) throws SignatureException {
+        // parse out the app link and signature. We assume there is a space after ccapp: and
+        // signature: and that the end of the signature is the end of the text content
+        String downloadLink = text.substring(text.indexOf("ccapp:") + 7, text.indexOf(","));
+        String signature = text.substring(text.indexOf("signature:") + 11);
```
gcapalbo: This does match the format I shared in the example, but now that I think about it, there's a couple weaknesses in that 1) it exposes another link to the user which might confuse them and make them click on the wrong link, and 2) the user might think it's junk and delete it which would cause problems too.  What do you think about changing the format to be:

`[COMMCARE SMS - DO NOT DELETE] <base 64 encoded message>`, and once you decode `<base 64 encoded message>`, you get: `ccapp: <applink> signature: <signature>`, with `<applink>` being the same app link and now `<signature>` is the signature in binary format.  Also note this gets rid of the comma which I think makes things cleaner so that there's just one space as padding.

app/src/org/commcare/android/logic/GlobalConstants.java
```diff
@@ -50,4 +50,9 @@
     //All of the app state is contained in these values
     public static final String STATE_USER_KEY = "COMMCARE_USER";
     public static final String STATE_USER_LOGIN = "USER_LOGIN";
+
+    public static final String CCHQ_PUBLIC_KEY = "MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQDHiuy2ULV4pobkuQN2TEjmR1tn" +
+            "HJ+F335hm/lVdaFQzvBmeq64MUMbumheVLDJaSUiAVzqSHDKJWH01ZQRowqBYjwo" +
+            "ycVSQSeO2glc6XZZ+CJudAPXe8iFWLQp3kBBnBmVcBXCOQFO7aLgQMv4nqKZsLW0" +
+            "HaAJkjpnc165Os+aYwIDAQAB";
```
gcapalbo: Is there a way to make the public key be configurable somehow? I'm just thinking if there are commcare deployments besides our main ones, someone would have to fork the commcare code, update the public key, and recompile it in order to make this feature work.  Would be nice if there was an easier way to configure it.

app/src/org/commcare/android/net/HttpRequestGenerator.java
```diff
@@ -259,7 +259,7 @@ public static boolean isValidRedirect(URL url, URL newUrl) {
             return true;
         }
 
-        //if it is, verify that we're on the same server.
+        //if it is, verifyMessageSignatureHelper that we're on the same server.
```
gcapalbo: was this just a find and replace typo?

app/src/org/commcare/android/util/SigningUtil.java
```diff
@@ -0,0 +1,54 @@
...
+     * @param messageSignature the signature generated by HQ with its private key and the message content
+     * @return whether or not the message was verified to be sent with HQ's private key
+     */
+    public static boolean verifyMessageSignatureHelper(String publicKeyString, String message, String messageSignature) {
+        try {
+            PublicKey publicKey = getPublicKey(publicKeyString);
+            return verifyMessageSignature(publicKey, message, messageSignature);
+        } catch (Exception e) {
+            // a bunch of exceptions can be thrown from the crypto methods. I mostly think we just
+            // care that we couldn't verifyMessageSignatureHelper it
```
gcapalbo: another find and replace typo?

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -550,22 +556,108 @@ protected void deliverError(CommCareSetupActivity receiver,
...
+     * @param installAutomatically install automatically if reference is found
+     */
+    private void performSMSInstall(boolean installAutomatically){
+        String profileLink = null;
+        try {
+            profileLink = this.scanSMSLinks();
+        } catch(SignatureException e){
+            // possibly we want to do something more severe here? could be malicious
+            e.printStackTrace();
+            Toast.makeText(this, Localization.get("menu.sms.not.found"), Toast.LENGTH_LONG).show();
```
gcapalbo: if this is being displayed while catching a `SignatureException`, then should the message be `menu.sms.not.verified`?

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -550,22 +556,108 @@ protected void deliverError(CommCareSetupActivity receiver,
...
+            // or move to READY_TO_INSTALL state
+            Log.v("install", "Performing SMS install with link : " + incomingRef);
+            incomingRef = profileLink;
+            if (installAutomatically) {
+                startResourceInstall();
+            } else {
+                uiState = UiState.READY_TO_INSTALL;
+                Toast.makeText(this, Localization.get("menu.sms.ready"), Toast.LENGTH_LONG).show();
+            }
+        } else if(installAutomatically) {
```
gcapalbo: should this be `!installAutomatically`?

gcapalbo:
What happens if the sms with the app install link is not found - does it just install odk, display the sms not found message, and that's it? For now, I've made it optional in the server backend to send an app link (in case someone wants to register without selecting an app right away), but I think the plan is to make it required to select an app in the ui when initiating the workflow anyway. Just curious what would happen if someone were to not send an app link.

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -550,22 +556,108 @@ protected void deliverError(CommCareSetupActivity receiver,
...
+        if (profileLink != null) {
+            // we found a valid profile link, either start install automatically
+            // or move to READY_TO_INSTALL state
+            Log.v("install", "Performing SMS install with link : " + incomingRef);
+            incomingRef = profileLink;
+            if (installAutomatically) {
+                startResourceInstall();
+            } else {
+                uiState = UiState.READY_TO_INSTALL;
+                Toast.makeText(this, Localization.get("menu.sms.ready"), Toast.LENGTH_LONG).show();
```
gcapalbo: Once the install completes, is there a way to display a message to the user notifying them that it's now ok to delete the encoded SMS?

app/src/org/commcare/android/logic/GlobalConstants.java
wpride: Can't really think of a way to do this that wouldn't require the application to be recompiled... @ctsims any thoughts?

wpride:
@gcapalbo there are two workflows: the SMS install triggered automatically at the CommCare install screen, and the one triggered manually from the menu.

If automatic: at the moment if no SMS with a valid install link is found then no notification will be displayed at all - the behavior would be the same as that of a normal user
If manual: if no SMS with a valid install link is found, a notification will be shown stating this

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
wpride: We could add this, but would be a bit annoying since ATM after we read the text we just enter the generic install workflow with the bit.ly link. @krsdimagi what do you think? is this important?

wpride:
@gcapalbo @krsdimagi updated PR, I think good to go pending some UX decisions

krsdimagi:
@gcapalbo @wpride I think the automatic workflow is probably the way to go (at least at this point).  We are going to be doing some UX review after the first round of messages goes out later this week, and if it breaks down too often then we may look at users taking more steps.  However, my gut feeling right now is that if a manual workflow breaks down a notification won't really help much.  Also, the idea here would be to always send an app link.

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
krsdimagi: @gcapalbo @wpride commented below but I'll add it inline here - we shouldn't worry about this (especially if it's annoying) because presumably the user will be in CommCare at this point.  If they delete the text before completing the workflow, they probably aren't super interested anyway.

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -550,22 +556,109 @@ protected void deliverError(CommCareSetupActivity receiver,
...
+            // possibly we want to do something more severe here? could be malicious
+            e.printStackTrace();
+            Toast.makeText(this, Localization.get("menu.sms.not.verified"), Toast.LENGTH_LONG).show();
+        }
+        if (profileLink != null) {
+            // we found a valid profile link, either start install automatically
+            // or move to READY_TO_INSTALL state
+            Log.v("install", "Performing SMS install with link : " + incomingRef);
+            incomingRef = profileLink;
+            if (installTriggeredManually) {
```
gcapalbo: I may be misunderstanding the flow, but should this be `!installTriggeredManually` now?

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
gcapalbo: @wpride @krsdimagi @sheelio What do you think about this point? Basically deciding on what the format of the encoded applink SMS should be, either:

`ccapp: <applink>, signature: <base 64 encoded signature>` (current format with potentially confusing plaintext link), or:
`[COMMCARE SMS - DO NOT DELETE] <base 64 encoded message>`

See above for details.

krsdimagi: @gcapalbo I like the second option - definitely seems a bit less confusing and is probably better from a UX standpoint.

sheelio: @gcapalbo 

I like the second option better - can we set it to be [COMMCARE APP - DO NOT DELETE] <base 64 encoded message>

Do we also send a link to the play store CommCare install to the user?

gcapalbo: @krsdimagi @sheelio Thanks for the feedback, I'll update the server side.  @wpride see above for details on the new message format.  I can send you a sample message to test with.

Should we somehow indicate to the user that it's ok to delete this message after some time? They only need to keep it on the phone until the install completes. Maybe something like `[KEEP THIS SMS UNTIL COMMCARE IS INSTALLED]`?

@sheelio we display the play store link right after the user registers online in the web interface

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
wpride: My thinking is that if they trigger the SMS scan themselves, its fine to then start the install automatically. However, if the scan was triggered without user input, then I'd prefer to give them the prompt and then the option to install or not (envisioning a scenario, possibly with multiple apps, where they have a text message but do not want to install)

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
wpride: Personally I don't think its that important to let them know they can delete - plenty of services/apps use the SMS inbox to post notifications without any intention of deleting

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
gcapalbo: got it, makes sense

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
krsdimagi: @gcapalbo I agree with @wpride I don't think we have to let them know when to delete.  The workflow is going to be complicated enough as it is, so I think that additional information would just be frustrating.

gcapalbo: ok, thanks for the feedback everyone - @wpride I'll send you the sample sms with the new format

wpride:
@gcapalbo

app/src/org/commcare/android/util/SigningUtil.java
```diff
@@ -0,0 +1,157 @@
...
+ *
+ * [COMMCARE APP - DO NOT DELETE] Y2NhcHA6IGh0dHA6Ly9iaXQubHkvMU5JMUl6MyBzaWduYXR1cmU6IEvECygFUhiUH
+ * 3TRjC0lClQrpLR7lG//IpDYpRH7ComtZRjTirteXmPyM9fRgbPZ9K6jG9zEms9WQj55Uo7jTujKNYThjU8rJJmWLouJBr/Yn
+ * WobEupwzn6DP2FavPF1YLPbp0ZctOfymW3m4j3VZ0lR2dMOjmInMSBiInqICKid
+ *
+ * WHERE BASE64-Encoded-Message is decoded into:
+ *
+ * ccapp: <profile link> signature: <binary signature>
+ *
+ * And we can then verify that the profile link was in fact encrpyted (using SHA256withRSA) by
```
gcapalbo: (technically - signed, not encrypted)

app/src/org/commcare/android/util/SigningUtil.java
```diff
@@ -0,0 +1,157 @@
...
+        byte[] message = messageString.getBytes();
+        sign.initVerify(publicKey);
+        sign.update(message);
+        return sign.verify(signature);
+    }
+
+    private static String decodeEncodedSMS(String text) throws  SignatureException{
+        String decodedMessage = null;
+        try {
+            decodedMessage = new String(Base64.decode(text), "UTF-8");
```
gcapalbo: Sorry if I didn't make this completely clear on skype, but we shouldn't be calling `String()` on the entire message because there's a good chance that at some point we're going to generate a signature that can't be decoded with UTF-8 and it will raise an unnecessary exception.

I think the easiest way to go about this is to just inspect the raw bytes (iterate over each byte) and look for the third space. Everything after the third space is the binary signature, and everything before it can be cast to a `String` since it's just ascii.

wpride:
@gcapalbo thanks

app/src/org/commcare/android/util/SigningUtil.java
```diff
@@ -0,0 +1,181 @@
...
+     * bytes
+     *
+     * @param textMessage
+     * @return the binary representation of the signtature
+     * @throws Exception if any errors are encountered
+     */
+    public static byte[] getSignatureBytes(String textMessage) throws Exception {
+        byte[] messageBytes = getBytesFromSMS(textMessage);
+        String decodedMessageString = parseAndDecodeSMS(textMessage);
+        int lastSpaceIndex = getSignatureStartIndex(messageBytes);
```
gcapalbo: if `lastSpaceIndex == -1`, we shouldn't keep processing

app/src/org/commcare/android/util/SigningUtil.java
```diff
@@ -0,0 +1,181 @@
...
+     * directly because the conversion from Base64 can have a non-1:1 correspondence with the actual
+     * bytes
+     *
+     * @param textMessage
+     * @return the binary representation of the signtature
+     * @throws Exception if any errors are encountered
+     */
+    public static byte[] getSignatureBytes(String textMessage) throws Exception {
+        byte[] messageBytes = getBytesFromSMS(textMessage);
+        String decodedMessageString = parseAndDecodeSMS(textMessage);
```
gcapalbo: seems like we can remove this line now right? because it's still going to call `String()` on the whole byte array which can cause the issue

app/src/org/commcare/android/util/SigningUtil.java
```diff
@@ -0,0 +1,181 @@
...
+     * @param text the parsed out text message in the expected link/signature format
+     * @return the download link if the message was valid and verified, null otherwise
+     * @throws SignatureException if we discovered a valid-looking message but could not verifyMessageSignatureHelper it
+     */
+    public static String parseAndVerifySMS(String text) throws SignatureException {
+        // parse out the app link and signature. We assume there is a space after ccapp: and
+        // signature: and that the end of the signature is the end of the text content
+        try {
+            byte[] signatureBytes = getSignatureBytes(text);
+            String decodedMessage = parseAndDecodeSMS(text);
```
gcapalbo: The `parseAndDecodeSMS` method still calls `decodeEncodedSMS` which calls `String()` on the whole byte array which will cause the issue I mentioned earlier.  It looks like you just have to make `decodeEncodedSMS` call `String()` on only the the `ccapp: ... signature:` part of the message (i.e., excluding the binary signature) and then this should be good.

app/src/org/commcare/android/util/SigningUtil.java
```diff
@@ -0,0 +1,181 @@
...
+import java.security.spec.X509EncodedKeySpec;
+
+/**
+ * A set of helper methods for verifying whether a message was genuinely sent from HQ. Currently we
+ * expcect the SMS in the format [COMMCARE APP - DO NOT DELETE] {BASE64-Encoded-Message} EG
+ *
+ * [COMMCARE APP - DO NOT DELETE] Y2NhcHA6IGh0dHA6Ly9iaXQubHkvMU5JMUl6MyBzaWduYXR1cmU6IEvECygFUhiUH
+ * 3TRjC0lClQrpLR7lG//IpDYpRH7ComtZRjTirteXmPyM9fRgbPZ9K6jG9zEms9WQj55Uo7jTujKNYThjU8rJJmWLouJBr/Yn
+ * WobEupwzn6DP2FavPF1YLPbp0ZctOfymW3m4j3VZ0lR2dMOjmInMSBiInqICKid
+ *
```
gcapalbo: Seems like we could also write a quick test by calling `SigningUtil.parseAndVerifySMS`, passing it this example

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -550,22 +557,93 @@ protected void deliverError(CommCareSetupActivity receiver,
...
+    private String scanSMSLinks() throws SignatureException, IOException{
+        // http://stackoverflow.com/questions/11301046/search-sms-inbox
+        final Uri SMS_INBOX = Uri.parse("content://sms/inbox");
+        Cursor cursor = getContentResolver().query(SMS_INBOX, null, null, null, "date desc");
+        try {
+            if (cursor.moveToFirst()) { // must check the result to prevent exception
+                while (!cursor.isAfterLast()) {
+                    String textMessageBody = cursor.getString(cursor.getColumnIndex("body"));
+                    if (textMessageBody.contains(GlobalConstants.SMS_INSTALL_KEY_STRING)) {
+                        String installLink = SigningUtil.retrieveParseAndVerifyURL(textMessageBody);
```
gcapalbo: `retrieveParseAndVerifyURL` expects only the url, so you should parse out the url here with substring (the text message is `[commcare app - do not delete] <url>`

app/src/org/commcare/android/util/SigningUtil.java
```diff
@@ -0,0 +1,220 @@
...
+import java.security.SignatureException;
+import java.security.spec.X509EncodedKeySpec;
+
+/**
+ * A set of helper methods for verifying whether a message was genuinely sent from HQ. Currently we
+ * expcect the SMS in the format [COMMCARE APP - DO NOT DELETE] {BASE64-Encoded-Message} EG
+ *
+ * [COMMCARE APP - DO NOT DELETE] Y2NhcHA6IGh0dHA6Ly9iaXQubHkvMU5JMUl6MyBzaWduYXR1cmU6IEvECygFUhiUH
+ * 3TRjC0lClQrpLR7lG//IpDYpRH7ComtZRjTirteXmPyM9fRgbPZ9K6jG9zEms9WQj55Uo7jTujKNYThjU8rJJmWLouJBr/Yn
+ * WobEupwzn6DP2FavPF1YLPbp0ZctOfymW3m4j3VZ0lR2dMOjmInMSBiInqICKid
```
gcapalbo: we should update this comment since now the format of the sms is:
`[commcare app - do not delete] <url>`

once you follow the url, you get the base64-encoded message which is decoded the same way described here below

app/src/org/commcare/android/util/SigningUtil.java
```diff
@@ -0,0 +1,220 @@
...
+    // given a text message, parse out the cruft and return the raw String
+    private static String parseAndDecodeSMS(String newMessage) throws Exception {
+        String parsed = parseMessage(newMessage);
+        int lastSpaceIndex = getSignatureStartIndex(getBytesFromSMS(newMessage));
+        String messageString = parsed.substring(0, lastSpaceIndex);
+        return decodeEncodedSMS(messageString);
+    }
+
+    // given a text message, parse out the cruft and return
+    private static String parseMessage(String newMessage){
```
gcapalbo: this method is no longer needed since now the url only returns the base64 encoded message. we should remove any calls made to this method

app/src/org/commcare/android/util/SigningUtil.java
```diff
@@ -0,0 +1,220 @@
...
+
+    // given a text message, parse out the cruft and return the raw Base64 bytes
+    public static byte[] getBytesFromSMS(String newMessage) throws Exception{
+        return Base64.decode(parseMessage(newMessage));
+    }
+    // given a text message, parse out the cruft and return the raw String
+    private static String parseAndDecodeSMS(String newMessage) throws Exception {
+        String parsed = parseMessage(newMessage);
+        int lastSpaceIndex = getSignatureStartIndex(getBytesFromSMS(newMessage));
+        String messageString = parsed.substring(0, lastSpaceIndex);
```
gcapalbo: this isn't going to work - you're passing the base64-encoded message here as newMessage

gcapalbo: i think you'll have to base64 decode `newMessage`, get the message bytes, then get the signature start index, convert to a new byte array which excludes the signature, then cast the bytes to String

app/src/org/commcare/android/util/SigningUtil.java
```diff
@@ -0,0 +1,224 @@
...
+ * And we can then verify that the profile link was in fact signed (using SHA256withRSA) by
+ * the CommCareHQ private key
+ *
+ * Created by wpride1 on 9/11/15.
+ */
+public class SigningUtil {
+
+    public static String retrieveParseVerifyMessage(String url) throws SignatureException, IOException {
+        String markedMessagePayload = convertUrlToPayload(url);
+        String messagePayload = trimMessagePayload(markedMessagePayload);
```
gcapalbo: you'll actually want to call `trimMessagePayload` first, since the String that's being passed as `url` to this method right now is actually:

`[commcare app - do not delete] <url>`

gcapalbo:
@wpride The refactor looks great, much cleaner and easier to read now.  Just one comment left and this should be all set

gcapalbo:
All the code looks good now, should be good to merge!  Just one last thing - I see the comment at the top of SigningUtil.java doesn't reflect the new format of the sms. Can you update it?

app/src/org/commcare/android/logic/GlobalConstants.java
ctsims: We can put it in the gradle settings to make it easier for people to swap out.

Also: I'd refer to it as "TRUSTED_SOURCE_PRIVATE_KEY" or something rather than CCHQ (generally best to avoid ever referring to our server specifically)

wpride:
@gcapalbo @krsdimagi I've confirmed this build is working with Giovanni's SMS. Should be good to go

krsdimagi:
@gcapalbo @wpride Thanks.  Just an FYI, as it turns out the client decided to try the initial send Monday (something happened today and they didn't want to start the workflow).

app/src/org/commcare/android/logic/GlobalConstants.java
```diff
@@ -50,4 +52,8 @@
     //All of the app state is contained in these values
     public static final String STATE_USER_KEY = "COMMCARE_USER";
     public static final String STATE_USER_LOGIN = "USER_LOGIN";
+
+    public static final String TRUSTED_SOURCE_PRIVATE_KEY = BuildConfig.TRUSTED_SOURCE_PRIVATE_KEY;
```
gcapalbo: we should rename this to `TRUSTED_SOURCE_PUBLIC_KEY` (since it's hq's public key we're defining here) to avoid confusion

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -720,5 +772,40 @@ public void setUiState(UiState newState) {
         uiState = newState;
     }
 
-    //endregion
+
+    @Override
+    public void downloadLinkReceived(String url) {
+        if (url != null)
```
gcapalbo: don't you need curly braces here?

gcapalbo:
@wpride ok, this looks good to merge - is the build failure real or does that always happen?

wpride:
Hmm, this build isn't failing but looks like whatever this test job is running is - however looks like that's the case on master too. Think its not related to this code

wpride:
confirmed with @phillipm this is an issue with the tests

PR #474, opened 2015-07-31
Small fixes to the login screen

dcluna:
Noticed that the login screen was behaving in a weird way in devices with small screens (such as my 2.3.7 emulated phone), so this PR fixes that.

Changes:
Added scrolling capabilities to the login screen.
Ensuring the 'Log In' box has the minimum padding top and bottom.


phillipm:
This fixes the login button size on old devices, nice!

PR #478, opened 2015-08-04
fix barcode widget

wpride:
Re: http://manage.dimagi.com/default.asp?178126


amstone326:
Release Note: Fix bug creating two "Get Data" buttons in the Barcode Widget

PR #479, opened 2015-08-04
Fixing invisible progressbar in old devices

dcluna:
In old Android versions, the progress bar in form entry does not appear. This PR fixes that.


app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -919,6 +921,8 @@ public void updateNavigationCues(View view) {
             progressBar.setProgress(details.completedQuestions);
         }
 
+        progressBar.getProgressDrawable().setBounds(bounds);  //Set the bounds to the saved value
```
phillipm: Can you elaborate how moving this method call down fixes the bug?

dcluna: http://stackoverflow.com/a/17231649/1536138

PR #483, opened 2015-08-05
Fixing crash in older Android versions

dcluna:
Checking if build version is Honeycomb or greater before calling refreshActionBar().


app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -1093,7 +1093,10 @@ protected void onResume() {
         if (CommCareApplication._().getCurrentApp() != null) {
             platform = CommCareApplication._().getCommCarePlatform();
         }
-        refreshActionBar();
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.HONEYCOMB)
+        {
```
phillipm: ```{``` shouldn't be on its own line

PR #487, opened 2015-08-05
Barcode scan working in incomplete forms

dcluna:
Barcode scan wasn't working in the incomplete forms screen. This PR fixes that and refactors the callout implementation in a different class + interface.


app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -205,90 +199,58 @@ public void onCreate(Bundle savedInstanceState) {
...
-                    Log.i("SCAN","Using default barcode scan");
-                    Intent i = new Intent("com.google.zxing.client.android.SCAN");
-                    try {
-                        startActivityForResult(i, BARCODE_FETCH);
-                    } catch (ActivityNotFoundException anfe) {
-                        Toast.makeText(EntitySelectActivity.this,
-                                "No barcode reader available! You can install one " +
-                                "from the android market.",
-                                Toast.LENGTH_LONG).show();
+        barcodeScanOnClickListener = BarcodeScanListenerDefaultImpl.makeCalloutOnClickListener(
```
dcluna: Extracted all this logic to makeCalloutOnClickListener

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -534,100 +495,102 @@ public void onItemClick(AdapterView<?> listView, View view, int position, long i
...
             if (mNoDetailMode) {
                 returnWithResult(i);
-            } else  {
+            } else {
                 startActivityForResult(i, CONFIRM_SELECT);
             }
         }
     }
 
+    public void onBarcodeFetch(String result, Intent intent) {
```
dcluna: New interface for barcode scan

app/src/org/commcare/android/logic/BarcodeScanListenerDefaultImpl.java
```diff
@@ -0,0 +1,121 @@
...
+                    act.startActivityForResult(i, CALLOUT);
+                } catch (ActivityNotFoundException anfe) {
+                    Toast.makeText(act, "No application found for action: " + actionName, Toast.LENGTH_LONG).show();
+                }
+            }
+        };
+    }
+
+    public static View.OnClickListener makeCalloutOnClickListener(final Activity act, Callout callout, CalloutActionSetup calloutActionSetup) {
+        if (callout == null) {
```
phillipm: Whenever you have a function that checks the nullity of its arguments and performs completely separate logic based on that check, it is a sign that you should break it up into two separate functions.

In this case you should have two methods, something like: ```barcodeIntentLauncher(Activity activity)``` and ```makeCalloutWithActionStep(Activity activity, Callout callout, CalloutActionStep calloutActionStep)```.
Then ```EntitySelectActivity``` can call  ```makeCalloutWithActionStep``` and ```FormRecordList``` can call ```barcodeIntentLauncher```

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
phillipm: This needs an override annotation, as well as all the other implementations of the BarcodeScanListener interface

app/src/org/commcare/android/logic/BarcodeScanListenerDefaultImpl.java
```diff
@@ -0,0 +1,121 @@
...
+    public interface CalloutAction {
+        void callout();
+    }
+
+    public static final int BARCODE_FETCH = 1;
+    public static final int CALLOUT = 3;
+
+    public static void onBarcodeResult(BarcodeScanListener barcodeScanListener, int requestCode, int resultCode, Intent intent) {
+        if (BuildConfig.DEBUG) {
+            if (!(requestCode == BARCODE_FETCH)) {
```
phillipm: This can never be true because every call of onBarcodeResult is preceded by a check that resultCode == BARCODE_FETCH. My vote is to remove this because it adds clutter to code when being read.

phillipm: Also very strange to only throw an exception when using a debug build. Who knows how that exception might be caught. If you really want to exit, I would consider asserting false or something that is guaranteed to crash the app.

app/src/org/commcare/android/logic/BarcodeScanListenerDefaultImpl.java
```diff
@@ -0,0 +1,121 @@
...
+                throw new IllegalArgumentException("requestCode should've been BARCODE_FETCH!");
+            }
+        }
+        if (resultCode == Activity.RESULT_OK) {
+            String result = intent.getStringExtra("SCAN_RESULT");
```
phillipm: Consider replacing every occurrence of "SCAN_RESULT" with a constant. Makes it easier to adapt our code if 3rd party apps decide to change.

app/src/org/commcare/android/logic/BarcodeScanListenerDefaultImpl.java
```diff
@@ -0,0 +1,121 @@
...
+
+        return new CalloutAction() {
+            @Override
+            public void callout() {
+                Log.i("SCAN", "Using barcode scan with action: " + actionName);
+                Intent i = new Intent(actionName);
+
+                for (Map.Entry<String, String> keyValue : extras.entrySet()) {
+                    i.putExtra(keyValue.getKey(), keyValue.getValue());
+                }
```
phillipm: If you make the intent final and move its setup code out of the anonymous CalloutAction instance, then you don't need to create the final ```actionName``` and ```extras``` vars.

app/src/org/commcare/android/logic/BarcodeScanListenerDefaultImpl.java
```diff
@@ -0,0 +1,121 @@
...
+        void onCalloutResult(String result, Intent intent);
+    }
+
+    public interface CalloutActionSetup {
+        void onImageFound(CalloutData calloutData);
+    }
+
+    public interface CalloutAction {
+        void callout();
+    }
```
phillipm: This interface can be made private, which might help make it clearer that it is used solely because Java doesn't do higher order functions so well.

app/src/org/commcare/android/logic/BarcodeScanListenerDefaultImpl.java
```diff
@@ -0,0 +1,121 @@
...
+            act.startActivityForResult(i, BARCODE_FETCH);
+        } catch (ActivityNotFoundException anfe) {
+            Toast.makeText(act,
+                    "No barcode reader available! You can install one " +
+                            "from the android market.",
+                    Toast.LENGTH_LONG).show();
+        }
+    }
+
+    public static CalloutAction makeCalloutAction(final Activity act, Callout callout, CalloutActionSetup calloutActionSetup) {
```
phillipm: Should be private, especially if you make CalloutAction private

app/src/org/commcare/android/logic/BarcodeScanListenerDefaultImpl.java
```diff
@@ -0,0 +1,121 @@
...
+
+/**
+ * Created by dancluna on 8/5/15.
+ */
+public final class BarcodeScanListenerDefaultImpl {
+    public interface BarcodeScanListener {
+        void onBarcodeFetch(String result, Intent intent);
+
+        void onCalloutResult(String result, Intent intent);
+    }
```
phillipm: I like the idea of moving this interface into its own file. This will make all the uses of it much cleaner, because it won't have to be prefixed by BarcodeScanListenerDefaultImpl

app/src/org/commcare/android/logic/BarcodeScanListenerDefaultImpl.java
```diff
@@ -0,0 +1,121 @@
...
+
+    public interface CalloutActionSetup {
+        void onImageFound(CalloutData calloutData);
+    }
```
phillipm: Consider moving into callout related files. Such as org.commcare.suite.model.Callout or org.commcare.suite.model.CalloutData

app/src/org/commcare/android/logic/BarcodeScanListenerDefaultImpl.java
phillipm: On second thought, this interface probably isn't needed. Couldn't you remove its need if ```makeCalloutAction``` returned an ```View.OnClickListener``` instance?

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -534,100 +496,102 @@ public void onItemClick(AdapterView<?> listView, View view, int position, long i
...
-                Callout callout = shortSelect.getCallout();
-                for (String key : callout.getResponses()) {
-                    result = intent.getExtras().getString(key);
-                    if (result != null && !resultSet) {
-                        resultSet = true;
-                        this.searchbox.setText(result);
-                        break;
+        switch (requestCode) {
+            case BarcodeScanListenerDefaultImpl.BARCODE_FETCH:
+                BarcodeScanListenerDefaultImpl.onBarcodeResult(this, requestCode, resultCode, intent);
```
phillipm: A little confused why handling a return from barcode scanning needs to be routed through onBarcodeResult, which in turn calls onBarcodeFetch defined above. Why not just call onBarcodeFetch directly?

If you are looking to avoid duplicating the following code, just have local implementations of onBarcodeFetch call a util method that does this processing:
```
            String result = intent.getStringExtra("SCAN_RESULT");
```

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -534,100 +496,102 @@ public void onItemClick(AdapterView<?> listView, View view, int position, long i
...
+        boolean resultSet = false;
+        if (result != null) {
+            setSearchText(result);
+            resultSet = true;
+        }
+        Callout callout = shortSelect.getCallout();
+        for (String key : callout.getResponses()) {
+            result = intent.getExtras().getString(key);
+            if (result != null && !resultSet) {
+                resultSet = true;
```
phillipm: Unused variable assignment

app/src/org/commcare/dalvik/activities/FormRecordListActivity.java
```diff
@@ -210,13 +239,38 @@ protected void onStop() {
...
+    public void onBarcodeFetch(String result, Intent intent) {
+        setSearchText(result);
+    }
+
+
+    public void onCalloutResult(String result, Intent intent) {
+        if (BuildConfig.DEBUG) {
+            throw new IllegalArgumentException("Callout not implemented!");
+        }
+    }
```
phillipm: Try to avoid implementing an interface if you aren't going to actually implement one of its methods. It is often a sign that the interface abstractions being used aren't necessary or need more refining.

app/src/org/commcare/android/logic/BarcodeScanListenerDefaultImpl.java
dcluna: This is an assertion, only in a different style because AS complains that "assertions are not reliable" when I try to use an actual

```java
assert requestCode == BARCODE_FETCH
```

dcluna: Wrapped the assertion into //region blocks, to improve visibility (folded by default in AS)

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
dcluna: If we leave this extra layer of indirection, we can centralize error processing and reduce LoC. I'm planning on a bigger refactor for barcode-related stuff, so let's just leave it there for now.

PR #492, opened 2015-08-06
Localize() in CommCareActivity should work with random Strings

dcluna:
CommCareActivity's localize method should now work with Localization keys and random Strings.

Fixing bug [177776](http://manage.dimagi.com/default.asp?177776#990571).


ctsims:
Hm. Not sure I 100% agree with this. This basically turns the behavior of a missing locale key from "Fail fast and let the user know their app is missing something" to "display something potentially confusing and don't let the user know anything was wrong"

The recovery mode is special in that it doesn't know whether keys will be available or not. I think the fixes we make should be specific to that use case.

dcluna:
I'll update the PR to make this behavior specific to this activity.

dcluna:
Ready for review

PR #499, opened 2015-08-07
allow multiple intent responses with the same key

dannyroberts:
to load one response value into multiple nodes

Check out the [Files changed](https://github.com/dimagi/commcare-odk/pull/499/files?w=1) in Ignore Whitespace mode

Some context, this PR is for http://manage.dimagi.com/default.asp?176497, where we want to have multiple `<response>`s with the same `@key` within one `<intent>`. In other words, we want to take the one variable from the callout response, and load it into more than one node in our form. For example:
```xml
<odkx:intent xmlns:odkx="http://opendatakit.org/xforms" id="aadhar_uid" class="org.commcarehq.aadharuid.barcode.LAUNCH">
	<response key="rawString" ref="/data/aadhar/rawString" />
	<response key="uid" ref="/data/aadhar/uid" />
	<response key="uid" ref="/data/aadhar_num" />
	<response key="name" ref="/data/aadhar/name" />
	<response key="name" ref="/data/name" />
	<response key="gender" ref="/data/aadhar/gender" />
	<response key="gender" ref="/data/gender" />
	<response key="yob" ref="/data/aadhar/yob" />
	<response key="co" ref="/data/aadhar/co" />
	<response key="house" ref="/data/aadhar/house" />
	<response key="street" ref="/data/aadhar/street" />
	<response key="lm" ref="/data/aadhar/lm" />
	<response key="loc" ref="/data/aadhar/loc" />
	<response key="vtc" ref="/data/aadhar/vtc" />
	<response key="po" ref="/data/aadhar/po" />
	<response key="dist" ref="/data/aadhar/dist" />
	<response key="subdist" ref="/data/aadhar/subdist" />
	<response key="state" ref="/data/aadhar/state" />
	<response key="state" ref="/data/state" />
	<response key="pc" ref="/data/aadhar/pc" />
	<response key="dob" ref="/data/aadhar/dob" />
	<response key="dob" ref="/data/dob" />
</odkx:intent>
```
Here the `name` gets loaded both into a hidden variable at `/data/aadhar/name` and into the Name question at `/data/name`.


dannyroberts:
@ctsims @phillipm

app/src/org/odk/collect/android/jr/extensions/IntentCallout.java
```diff
@@ -126,70 +129,70 @@ public boolean processResponse(Intent intent, ODKView currentView, FormInstance
                     key = "";
                 }
 
-                //Figure out where it's going
-                TreeReference ref = responses.get(key);
+                for (TreeReference ref : responses.get(key)) {
```
phillipm: If you don't mind, it would be sweet to lift all the logic in this loop into a helper method. It pretty burly at the moment.

dannyroberts: Ok I can do that

app/src/org/odk/collect/android/jr/extensions/IntentCallout.java
```diff
@@ -129,70 +132,73 @@ public boolean processResponse(Intent intent, ODKView currentView, FormInstance
...
+            File src = new File(responseValue);
+            if (!src.exists()) {
+                //TODO: How hard should we be failing here?
+                Log.w(TAG, "ODK received a link to a file at " + src.toString() + " to be included in the form, but it was not present on the phone!");
+                //Wipe out any reference that exists
+                form.setValue(null, ref);
+                return;
+            }
+
+            File newFile = new File(destination, src.getName());
```
phillipm: ```destination``` is now out of scope and hence won't build

dannyroberts: Good catch just noticed this myself

dannyroberts:
I've now built and tested on my phone, and it has the desired effect

PR #500, opened 2015-08-07
Limit pan & zoom to full-screen graphs

orangejenny:
http://manage.dimagi.com/default.asp?178653

### DIFF
--- a/app/src/org/commcare/android/view/GraphView.java
+++ b/app/src/org/commcare/android/view/GraphView.java
@@ -104,7 +104,7 @@ public class GraphView {
     public Intent getIntent(GraphData data) throws InvalidStateException {
         render(data);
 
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/android/view/GraphView.java
```diff
@@ -103,7 +103,8 @@ private void render(GraphData data) throws InvalidStateException {
         
     public Intent getIntent(GraphData data) throws InvalidStateException {
         render(data);
-        
+
+        setPanAndZoom(Boolean.valueOf(mData.getConfiguration("zoom", "false")).equals(Boolean.TRUE));
```
phillipm: Should be able to remove the ```.equals(Boolean.TRUE)``` part, right?

orangejenny: indeed. fixed!

PR #506, opened 2015-08-07
evaluate "intent/@ref"s in the context of the current node

dannyroberts:
@ctsims


dannyroberts:
Still need to test in a repeat where this is relevant, but wanted to get general feedback on approach

dannyroberts:
@phillipm because you're my new commcare friend

czue:
@dannyroberts first mobile PR? :balloon: :balloon:

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -623,8 +624,24 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
-    
+
+    private IntentWidget tryToGetIntentWidget() {
+        IntentWidget bestMatch = null;
+
+        //Ugh, copied from the odkview mostly, that's stupid
+        for(QuestionWidget q : ((ODKView)mCurrentView).getWidgets()) {
+            //Figure out if we have a pending intent widget
+            if (q instanceof IntentWidget) {
+                if(((IntentWidget) q).isWaitingForBinaryData() || bestMatch == null) {
```
ctsims: Hey,

I think this whole function can be replaced with the calls below? Looks like you're still relying on the old system here before jumping into the new logic

dannyroberts: I see, you mean reimplement this function as

```return mFormController.getPendingCalloutFormIndex().whateverTheGetCurrentWidgetFunctionIsCalled()```

right? I wasn't certain that was possible, but your hinting that it is gives me courage so I'll give it a shot.

Comment (ctsims) [inline app/src/org/odk/collect/android/activities/FormEntryActivity.java]: roughly this should work as: q.getFormIndex.equals(getPendingFormIndex)

but yeah, specifically is concerning because the result of
tryToGetIntentWidget is used to control whether your new code is called at
all, which doesn't seem right

**Comment by dannyroberts** [comment]
This is ready to go, from my side. This PR allows you to use relative references in intent responses. It doesn't fix a separate bug, in which loading intent responses into questions on the select screen doesn't work.

**Comment by dannyroberts** [comment]
@ctsims feel free to merge now, or if you feel like there's more to the story than I'm understanding, feel free to wait until tomorrow

**Comment by dannyroberts** [comment]
Decided to fix "intents loading into questions on the same screen" bug in this PR (previous commit)

File: app/src/org/odk/collect/android/activities/FormEntryActivity.java [hunk 2]
```diff
@@ -611,8 +612,24 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
                 refreshCurrentView(false);
                 break;
         }
+        mFormController.setPendingCalloutFormIndex(null);
```
ctsims: This needs to also execute after line 482 where the intent is processed for cancellation.

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -611,8 +612,24 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
-    
+
+    private IntentWidget getPendingIntentWidget() {
+        IntentWidget bestMatch = null;
+
+        //Ugh, copied from the odkview mostly, that's stupid
+        for(QuestionWidget q : ((ODKView)mCurrentView).getWidgets()) {
+            //Figure out if we have a pending intent widget
+            if (q instanceof IntentWidget) {
+                if(((IntentWidget) q).getFormId().equals(mFormController.getPendingCalloutFormIndex())) {
```
ctsims: It would be good to generalize this process into the intent callout setup above and make it usable by Image Capture, etc. 

Don't have to do in this PR, but maybe a TODO: or a comment would be a helpful hint to someone in the future that this is mostly done and just needs to get wired up

app/src/org/odk/collect/android/jr/extensions/IntentCallout.java
```diff
@@ -106,15 +104,27 @@ public Intent generate(EvaluationContext ec) {
         return i;
     }
 
-    public boolean processResponse(Intent intent, ODKView currentView, FormInstance instance, File destination) {
+    private void setNodeValue(TreeReference reference, String stringValue) {
+        if (stringValue != null) {
+            EvaluationContext evaluationContext = new EvaluationContext(form.getEvaluationContext(), reference);
```
ctsims: Can we make a note that this code is almost identical to setValueAction's setter and the code below and note that no one should copy this process without refactoring it to unify that code?

app/src/org/odk/collect/android/logic/PendingCalloutInterface.java
```diff
@@ -0,0 +1,11 @@
+package org.odk.collect.android.logic;
+
+import org.javarosa.core.model.FormIndex;
+
+/**
```
ctsims: Class comment?

app/src/org/odk/collect/android/widgets/BarcodeWidget.java
```diff
@@ -44,15 +45,16 @@
 
     private TextView mStringAnswer;
 
-    public BarcodeWidget(Context context, FormEntryPrompt prompt, Intent i, IntentCallout ic) {
-        super(context, prompt, i, ic, FormEntryActivity.BARCODE_CAPTURE);
+    public BarcodeWidget(Context context, FormEntryPrompt prompt, Intent i, IntentCallout ic,
```
ctsims: can you add a TODO to just mention that we aren't using this yet so whoever implements the Pending Callouts for images and stuff knows to include this widget  as well?

dannyroberts:
Ok, I think this is good to go now.

PR #507, opened 2015-08-10
Auto-update checks in background

phillipm:
Enable auto-update checks and downloads to occur in the background.
Add a new update manager activity that allows users to check the status of updates, cancel downloading updates, and apply downloaded updates.

Cross-requested with https://github.com/dimagi/commcare/pull/117

[Spec found here](https://docs.google.com/document/d/1oSd6EF0pDgNui8tDbCY1eETx4KlHGDPt9m1L-jya9WI/edit?usp=sharing)

List of TODOs:
 - Fix rectangle button text/image layout
 - Check for UpdateTask cancellation in the middle of large file downloads

__Release Notes__:
Perform auto-update checks for new app versions in the background.


app/src/org/commcare/dalvik/activities/LoginActivity.java
```diff
@@ -150,6 +154,11 @@ protected void onCreate(Bundle savedInstanceState) {
             public void onClick(View arg0) {
                 errorBox.setVisibility(View.GONE);
                 ViewUtil.hideVirtualKeyboard(LoginActivity.this);
+
+                if (ResourceInstallUtils.isUpdateInstallReady()) {
+                    installPendingUpdate();
+                }
```
phillipm: This is an awkward place for installing a downloaded update, because if you give incorrect creds but there is a pending update, it will still install it. I'm going to punt on this for now because the LoginActivity is so convoluted that it is hard for me to know where a good place to put this is. I would like to refactor LoginActivity, at which point I will move this.

app/res/layout-land/update_activity.xml
```diff
@@ -0,0 +1,117 @@
...
+                        android:layout_height="wrap_content"
+                        android:orientation="vertical"
+            >
+
+            <LinearLayout
+                android:layout_width="match_parent"
+                android:layout_height="match_parent"
+                android:gravity="center_horizontal"
+                android:orientation="vertical"
+                >
```
amstone326: this is ridiculously picky but I hate the floating arrow thing in layout files, can they just go at the end of the line above

app/src/org/commcare/android/resource/AndroidResourceManager.java
```diff
@@ -0,0 +1,267 @@
...
+
+            return AppInstallStatus.UpdateStaged;
+        }
+    }
+
+    /**
+     * Load the latest profile into the upgrade table. Clears the upgrade table
+     * if it's partially populated with an out-of-date version.
+     */
+    private void instantiateLatestProfile()
```
amstone326: would be good to rename to something that makes it clear that it's for the upgrade table

app/src/org/commcare/android/resource/AndroidResourceManager.java
```diff
@@ -0,0 +1,267 @@
...
+    /**
+     * Download the latest profile; if it is new, download and stage the entire
+     * update.
+     *
+     * @param profileRef Reference that resolves to the profile file used to
+     *                   seed the update
+     * @return UpdateStaged upon update download, UpToDate if no new update,
+     * otherwise an error status.
+     */
+    public AppInstallStatus checkAndPrepareUpgradeResources(String profileRef) {
```
amstone326: Could make this section a little clearer if this method only did the "check" part -- would need to add another value to AppInstallStatus called like "UpdateNeeded" that would be returned at the end of the try block instead of calling prepareUpgradeResources(). And then in stageUpdate(), you would check for the return value of the call to checkForUpgrade(), and call prepareUpgradeResources() iff it returned AppInstallStatus.UpdateNeeded.

app/src/org/commcare/android/resource/AndroidResourceManager.java
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    /**
+     * Clear update table, log failure with update stats,
+     * and, if appropriate, schedule a update retry
+     *
+     * @param exception Log error with update statistics
+     * @param context   Used for showing pinned notification of update task retry
+     */
+    public void processUpdateFailure(Exception exception, Context context) {
```
amstone326: unused

app/src/org/commcare/android/resource/AndroidResourceManager.java
```diff
@@ -0,0 +1,267 @@
...
+
+        upgradeTable.clear();
+
+        retryUpdateOrGiveUp(context);
+    }
+
+    public void processUpdateFailure(AppInstallStatus result, Context ctx) {
+        updateStats.registerUpdateException(new Exception(result.toString()));
+
+        boolean reusePartialTable =
```
amstone326: Not sure if this is just carry-over from what was here in another form before, but how did you decide on the conditions for this?

app/src/org/commcare/android/resource/AndroidResourceManager.java
```diff
@@ -0,0 +1,267 @@
...
+
+            upgradeTable.clear();
+        } else {
+            Log.w(TAG, "Retrying auto-update");
+            UpdateStatPersistence.saveStatsPersistently(app, updateStats);
+            scheduleUpdateTask(ctx);
+        }
+    }
+
+    private void scheduleUpdateTask(final Context ctx) {
```
amstone326: might be worth naming or commenting this in a way that makes it clear that it's being used exclusively in the case of a retry after some sort of failure

app/src/org/commcare/android/resource/ResourceInstallUtils.java
```diff
@@ -0,0 +1,208 @@
...
+ * @author Phillip Mates (pmates@dimagi.com)
+ */
+public class ResourceInstallUtils {
+    private static final String DEFAULT_APP_SERVER_KEY = "default_app_server";
+
+    /**
+     * @return Is the current app's designated upgrade table staged and ready
+     * for installation?
+     */
+    public static boolean isUpdateInstallReady() {
```
amstone326: slightly confusing name, maybe isUpdateReadyToInstall() ?

app/src/org/commcare/android/tasks/ResourceEngineTask.java
```diff
@@ -149,16 +61,9 @@ public ResourceEngineTask(boolean upgradeMode, CommCareApp app,
...
         String profileRef = profileRefs[0];
-        AndroidCommCarePlatform platform = app.getCommCarePlatform();
-        SharedPreferences prefs = app.getAppPreferences();
-
-
-        // Make sure we record that an attempt was started.
-        Editor editor = prefs.edit();
-        editor.putLong(CommCarePreferences.LAST_UPDATE_ATTEMPT, new Date().getTime());
-        editor.commit();
+        ResourceInstallUtils.recordUpdateAttemptTime(app);
```
amstone326: Why does this count as an update attempt? Isn't ResourceEngineTask only for normal installations?

app/src/org/commcare/android/resource/analytics/InstallAttempts.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.Vector;
+
+/**
+ * Stores install attempt data for a given resource in the update table.
+ *
+ * @param <A> Type of data stored at each install attempt. Must be
+ *            instantiated with a class that implements Serializeable.
+ * @author Phillip Mates (pmates@dimagi.com)
+ */
+class InstallAttempts<A> implements Serializable {
```
amstone326: What is the purpose of using generics in this class? It looks like it's only getting used with type String right now, do you anticipate needing/wanting to use it with other types in the future?

app/src/org/commcare/android/resource/analytics/InstallAttempts.java
```diff
@@ -0,0 +1,69 @@
...
+
+    public void addFailure(A failureData) {
+        failures.add(new FailureEvent<>(failureData));
+    }
+
+    public void registerSuccesfulInstall() {
+        wasSuccessful = true;
+    }
+
+    public String toString() {
```
amstone326: worth adding an @Override annotation here

app/src/org/commcare/android/resource/analytics/UpdateStatPersistence.java
```diff
@@ -0,0 +1,69 @@
...
+import org.commcare.dalvik.application.CommCareApp;
+
+import java.io.IOException;
+
+/**
+ * Saves and loads stats associated with an app update job to shared preferences.
+ *
+ * @author Phillip Mates (pmates@dimagi.com)
+ */
+public class UpdateStatPersistence {
```
amstone326: In a very uncharacteristic move, I might argue that this extra class adds unnecessary complexity, given that there are already so many different classes surrounding creating and logging the update statistics. These 3 methods might do fine as just static methods of the UpdateStats class.

app/src/org/commcare/android/resource/analytics/UpdateStats.java
```diff
@@ -0,0 +1,129 @@
...
+        InstallAttempts<String> attempts = resourceInstallStats.get(resourceName);
+        if (attempts == null) {
+            attempts = new InstallAttempts<>(resourceName);
+            resourceInstallStats.put(resourceName, attempts);
+        }
+        String stackTrace = getStackTraceString(errorMsg);
+        attempts.addFailure(stackTrace);
+    }
+
+    private String getStackTraceString(Exception e) {
```
amstone326: should be static

app/src/org/commcare/android/tasks/InstallStagedUpdateTask.java
```diff
@@ -0,0 +1,50 @@
...
+    protected AppInstallStatus doTaskBackground(Void... params) {
+        CommCareApp app = CommCareApplication._().getCurrentApp();
+        app.setupSandbox();
+
+        AndroidCommCarePlatform platform = app.getCommCarePlatform();
+        AndroidResourceManager resourceManager =
+            new AndroidResourceManager(platform);
+
+        if (!resourceManager.isUpgradeTableStaged()) {
+            resourceManager.recordUpdateInstallFailure(AppInstallStatus.UnknownFailure);
```
amstone326: Depends on what we're actually using this stuff for, but could be worth making an AppInstall Status that expresses "we thought we had a staged upgrade table and then didn't when we tried to install it", instead of just using UnknownFailure

app/src/org/commcare/android/tasks/ResourceEngineTask.java
```diff
@@ -149,16 +61,9 @@ public ResourceEngineTask(boolean upgradeMode, CommCareApp app,
     }
 
     @Override
-    protected ResourceEngineOutcomes doTaskBackground(String... profileRefs) {
+    protected AppInstallStatus doTaskBackground(String... profileRefs) {
```
amstone326: Could clean up this method even more by pulling the exception handling code for each separate exception out into methods.

app/src/org/commcare/android/tasks/UpdateTask.java
```diff
@@ -0,0 +1,266 @@
...
+        if (pinnedNotificationProgress != null) {
+            pinnedNotificationProgress.handleTaskUpdate(values);
+        }
+        if (taskListener != null) {
+            taskListener.handleTaskUpdate(values);
+        }
+    }
+
+    @Override
+    protected void onPostExecute(AppInstallStatus result) {
```
amstone326: Might want to clear the UpdateStats here like we talked about

app/src/org/commcare/android/resource/AndroidResourceManager.java
phillipm: Good call, though I think I'm going to leave this for a later date because the error handling code in this method is quite substantial and would be shared between the check and prepare methods. This sort of thing will be super easy to abstract once anonymous functions (via Java 8) are supported. Let me know if you can think of a clean way to avoid not duplicating large try/catch blocks like found here.

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -201,8 +203,8 @@ protected void returnToLogin() {
     }
 
     protected void attemptSync() {
-        if (isNetworkNotConnected()) {
-            if (isAirplaneModeOn()) {
+        if (ConnectivityStatus.isNetworkNotConnected(CommCareHomeActivity.this)) {
```
amstone326: CommCareActivity already has an isNetworkNotConnected() method, which calls out to a method in AndroidUtil. Would be nice to use that here, and then maybe move the isAirplaneModeOn() method into AndroidUtil and call it in a similar way, so that we don't need the extra util class. We definitely already have both of these functionalities scattered across a few different places in the code right now. Would be good to try to keep it unified going forward, and eventually do a real pass at eliminating duplicates.

app/src/org/commcare/android/resource/AndroidResourceManager.java
phillipm: Not really sure. I did deviate from the old way of determining this: https://github.com/dimagi/commcare-odk/blob/master/app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java#L482-L522.
Probably a good thing to go over with @ctsims

app/src/org/commcare/android/resource/analytics/InstallAttempts.java
phillipm: Left this generic in case we want to keep track of different types of information. This is a result of not spec'ing out this part of the feature very well. Should get narrowed soon, when I revisit this part of the feater

app/src/org/commcare/dalvik/activities/UpdateActivity.java
```diff
@@ -0,0 +1,256 @@
...
+                break;
+            case FINISHED:
+                uiController.errorUiState();
+                break;
+            default:
+                uiController.errorUiState();
+        }
+    }
+
+    private void pendingUpdateOrIdle() {
```
amstone326: seems like this method could be in the ui controller too

app/src/org/commcare/android/tasks/ResourceEngineTask.java
phillipm: Holdover from the old code, which counted the app install as the first update attempt when scheduling auto-updates.

app/src/org/commcare/dalvik/activities/UpdateActivity.java
phillipm: Wanted to have all the decision logic be in one place, mainly the activity. I should probably rename the `UpdateUiController` to `UpdateUiState` to signify that it is just used to enter different UI states, but doesn't actually contain any decision logic.

amstone326:
Finished a pretty thorough review, although I may have faded slightly in attentiveness toward the end. All the core changes look super solid, most of my comments were re-naming/clarification suggestions and the like. Will leave for @ctsims to do a pass and merge.

app/src/org/commcare/android/resource/AndroidResourceManager.java
amstone326: Ah, I hadn't noticed that. I think a relatively decent solution would be to move the bodies of the catch blocks out into their own methods, and then duplicating the catch blocks in 2 places is much cleaner, because you just call the same methods from both places. Still means a bit of duplication, but much less. Might be worth it for the clarification of not doing 2 things in 1 method, but your call on which is the lesser evil.

app/src/org/commcare/dalvik/activities/UpdateActivity.java
amstone326: Ah ok yea, definitely a good distinction to make.

app/src/org/commcare/android/database/app/AppDatabaseUpgrader.java
```diff
@@ -40,6 +41,9 @@ public void upgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
                 oldVersion = 5;
             }
         }
+        if (oldVersion == 5) {
+            upgradeFiveSix(db);
```
ctsims: This should set oldVersion to 6 after it finishes

app/src/org/commcare/android/database/app/DatabaseAppOpenHelper.java
```diff
@@ -67,19 +73,24 @@ public void onCreate(SQLiteDatabase database) {
...
             database.setTransactionSuccessful();
-
         } finally {
             database.endTransaction();
         }
     }
 
+    public static String tableIndexQuery(String tableName, String indexName) {
+        return "CREATE INDEX " + indexName + " ON " +
+                tableName + "( " + Resource.META_INDEX_PARENT_GUID + " )";
```
ctsims: META_INDEX_PARENT_GUID is the name of the column to be indexed here? would it make more sense for this helper method to get an index as an argument as well? Seems like an odd level of deduplication?

app/src/org/commcare/android/resource/AndroidResourceManager.java
```diff
@@ -0,0 +1,258 @@
...
+     * and, if appropriate, schedule a update retry
+     *
+     * @param result update attempt result
+     * @param ctx    Used for showing pinned notification of update task retry
+     */
+    public void processUpdateFailure(AppInstallStatus result, Context ctx) {
+        updateStats.registerUpdateException(new Exception(result.toString()));
+
+        boolean reusePartialTable =
+                (result == AppInstallStatus.UnknownFailure ||
```
ctsims: One thing you may find helpful here is that enumerations can have properties set on them. You could set the AppInstallStatus enumeration to have "can reuse in future" value that can be applied to new enumeration outcomes, which would be easier to maintain than have to track back to here.

app/src/org/commcare/android/resource/AndroidResourceManager.java
```diff
@@ -0,0 +1,258 @@
...
+                try {
+                    UpdateTask updateTask = UpdateTask.getNewInstance();
+                    updateTask.startPinnedNotification(ctx);
+                    updateTask.execute(ref);
+                } catch (IllegalStateException e) {
+                    // The user may have started the update process in the meantime
+                    Log.w(TAG, "Trying trigger an auto-update retry when it is already running");
+                }
+            }
+        }, UPDATE_RETRY_DELAY_IN_MS);
```
ctsims: Do we want this value to back off at all?

app/src/org/commcare/android/resource/AndroidResourceManager.java
```diff
@@ -0,0 +1,258 @@
...
+            @Override
+            public void run() {
+                String ref = ResourceInstallUtils.getDefaultProfileRef();
+                try {
+                    UpdateTask updateTask = UpdateTask.getNewInstance();
+                    updateTask.startPinnedNotification(ctx);
+                    updateTask.execute(ref);
+                } catch (IllegalStateException e) {
+                    // The user may have started the update process in the meantime
+                    Log.w(TAG, "Trying trigger an auto-update retry when it is already running");
```
ctsims: hm, wondering whether if there is some stateful stuff here that something should be locking, or will that always come through on the execute?

app/src/org/commcare/android/resource/analytics/UpdateStats.java
```diff
@@ -0,0 +1,184 @@
...
+                return new UpdateStats();
+            }
+        } else {
+            return new UpdateStats();
+        }
+    }
+
+    private static Object deserialize(String s) throws IOException,
+            ClassNotFoundException {
+        byte[] data = Base64.decode(s, Base64.DEFAULT);
```
ctsims: Assuming we're treating it as a loss if this changes over time and the serialization breaks?

app/src/org/commcare/android/tasks/InstallStagedUpdateTask.java
```diff
@@ -0,0 +1,50 @@
...
+            return AppInstallStatus.UnknownFailure;
+        }
+
+        try {
+            resourceManager.upgrade();
+        } catch (UnresolvedResourceException e) {
+            resourceManager.recordUpdateInstallFailure(e);
+            return AppInstallStatus.UnknownFailure;
```
ctsims: Is this an unknown error?

app/src/org/commcare/android/tasks/UpdateTask.java
```diff
@@ -0,0 +1,266 @@
...
+        }
+
+        if (pinnedNotificationProgress != null) {
+            pinnedNotificationProgress.handleTaskCompletion(result);
+        }
+        if (taskListener != null) {
+            taskListener.handleTaskCompletion(result);
+        }
+
+        synchronized (lock) {
```
ctsims: Am wondering whether it makes sense for this lock to actually start at the beginning of this method, since it's not like something can usefully jump back onto this Task once the onPostExecute phase has begun?

app/src/org/commcare/dalvik/activities/LoginActivity.java
```diff
@@ -638,4 +651,41 @@ public void onItemSelected(AdapterView<?> parent, View view, int position, long
...
+     * Block the user with a dialog while downloaded update is installed.
+     */
+    private void installPendingUpdate() {
+        InstallStagedUpdateTask<LoginActivity> task =
+                new InstallStagedUpdateTask<LoginActivity>(TASK_UPGRADE_INSTALL) {
+                    @Override
+                    protected void deliverResult(LoginActivity receiver,
+                                                 AppInstallStatus result) {
+                        if (result == AppInstallStatus.Installed) {
+                            Toast.makeText(LoginActivity.this,
```
ctsims: This should be bound to receiver, not this.

app/src/org/commcare/dalvik/activities/LoginActivity.java
```diff
@@ -638,4 +651,41 @@ public void onItemSelected(AdapterView<?> parent, View view, int position, long
...
+                    protected void deliverUpdate(LoginActivity receiver,
+                                                 int[]... update) {
+                    }
+
+                    @Override
+                    protected void deliverError(LoginActivity receiver,
+                                                Exception e) {
+                        e.printStackTrace();
+                        Log.e(TAG, "update installation on login failed: " + e.getMessage());
+                        Toast.makeText(LoginActivity.this,
```
ctsims: again, receiver, not this.

app/src/org/commcare/dalvik/activities/SingleAppManagerActivity.java
```diff
@@ -123,7 +123,11 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
         switch (requestCode) {
             case CommCareHomeActivity.UPGRADE_APP:
                 if (resultCode == RESULT_CANCELED) {
-                    Toast.makeText(this, R.string.update_canceled, Toast.LENGTH_LONG).show();
+                    UpdateTask task = UpdateTask.getRunningInstance();
```
ctsims: This is a bit awkward, timing wise, right? The instance could easily have started running again in a different thread by the time we're back here.

app/src/org/commcare/dalvik/activities/SingleAppManagerActivity.java
```diff
@@ -123,7 +123,11 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
+                    if (task != null) {
+                        Toast.makeText(this, R.string.update_canceled, Toast.LENGTH_LONG).show();
+                        task.cancel(true);
```
ctsims: similar to comment above, I think the right modulo here is acutally

UpdateTask.cancelRunningInstanceIfAny()

which is locked internally and returns a true/false for whether there was an existing running task

app/src/org/commcare/dalvik/activities/UpdateActivity.java
```diff
@@ -0,0 +1,256 @@
...
+        uiState.idleUiState();
+    }
+
+    protected void startUpdateCheck() {
+        try {
+            updateTask = UpdateTask.getNewInstance();
+            updateTask.startPinnedNotification(this);
+            updateTask.registerTaskListener(this);
+        } catch (IllegalStateException e) {
+            enterErrorState("There is already an existing update task instance.");
```
ctsims: Should this maybe just connect the running instance to this activity?

app/src/org/commcare/dalvik/activities/UpdateActivity.java
```diff
@@ -0,0 +1,256 @@
...
+     */
+    protected void lauchUpdateInstallTask() {
+        InstallStagedUpdateTask<UpdateActivity> task =
+                new InstallStagedUpdateTask<UpdateActivity>(DIALOG_UPGRADE_INSTALL) {
+
+                    @Override
+                    protected void deliverResult(UpdateActivity receiver,
+                                                 AppInstallStatus result) {
+                        if (result == AppInstallStatus.Installed) {
+                            uiState.updateInstalledUiState();
```
ctsims: receiver.uiState

app/src/org/commcare/dalvik/activities/UpdateActivity.java
```diff
@@ -0,0 +1,256 @@
...
+                        } else {
+                            uiState.errorUiState();
```
ctsims: receiver.uiState

app/src/org/commcare/dalvik/activities/UpdateActivity.java
```diff
@@ -0,0 +1,256 @@
...
+
+                    @Override
+                    protected void deliverUpdate(UpdateActivity receiver,
+                                                 int[]... update) {
+                    }
+
+                    @Override
+                    protected void deliverError(UpdateActivity receiver,
+                                                Exception e) {
+                        uiState.errorUiState();
```
ctsims: receiver.uiState

app/src/org/commcare/dalvik/activities/UpdateActivity.java
```diff
@@ -0,0 +1,256 @@
...
+                    }
+                };
+        task.connect(this);
+        task.execute();
+    }
+
+    @Override
+    public CustomProgressDialog generateProgressDialog(int taskId) {
```
ctsims: Was the general CommCareActivity dialog not able to handle this?

ctsims:
reviewed and commented

app/src/org/commcare/android/database/app/AppDatabaseUpgrader.java
phillipm: @ctsims `oldVersion` is not referenced after its assignment in the cases above. We should use comments if it just used to notify readers of the code. We could also choose to log the versions.

app/src/org/commcare/android/database/app/DatabaseAppOpenHelper.java
phillipm: Sounds good: I renamed `tableIndexQuery` to `indexOnTableWithPGUIDCommand` and also created the helper function `indexOnTableCommand` to take in column args.

app/src/org/commcare/dalvik/activities/UpdateActivity.java
phillipm: Not sure I follow; this is using the CommCareActivity dialog framework, right?

app/src/org/commcare/android/resource/AndroidResourceManager.java
phillipm: Don't really follow. This exception occurs if a retry is scheduled, user tries to manually update, and then the delayed retry is fired during this manual update. If this occurs, retry attempts will stop being scheduled.

app/src/org/commcare/android/resource/analytics/UpdateStats.java
phillipm: Sounds okay to me for now.

app/src/org/commcare/dalvik/activities/SingleAppManagerActivity.java
phillipm: @amstone326 requires users to logout of their commcare session before updating. Not sure if that is necessary anymore, but that saves us from having to worry about multiple instances of UpdateTask floating around, which sounds great because I designed it as a singleton task.

phillipm:
@ctsims thanks for the review. Updated the PR with most of the feedback and asked questions on the comments I didn't understand.

app/src/org/commcare/android/database/app/AppDatabaseUpgrader.java
ctsims: @phillipm What do you mean? It's referenced below each time as it falls through the if conditions (they aren't if/elses)

app/src/org/commcare/android/resource/AndroidResourceManager.java
ctsims: Ah, I see now. I didn't realize it was firing from the newInstance call.

app/src/org/commcare/dalvik/activities/UpdateActivity.java
ctsims: Ah, yeah, you're right, sorry.

ctsims:
@phillipm Good to go once API's are in.

PR #515, opened 2015-08-12
Use Etsy's StaggeredGridView on devices with API level >= 10

dcluna:
This fixes bug [178554](http://manage.dimagi.com/default.asp?178554#993157) and improves the home screen UI.


phillipm:
Tested on an API 10 device and looks good.
@dcluna Did you test this on an API 8 emulator?

ctsims:
tested on API8 and it works, but the buttons move around and aren't in a
consistent order.

ctsims:
I imagine that should be replicable locally by just changing the flag, btw,
so it uses the old staggered view on newer devices.

PR #516, opened 2015-08-12
Stop mobile UCR crash

orangejenny:
http://manage.dimagi.com/default.asp?167204

Not sure if it's legitimate to assume that if value wasn't provided, it's because you want to start the session over. One the other hand, "start over" is a better assumption than "crash".

@ctsims


ctsims:
Hm, not 100% sure I agree. I think i'd rather we failfast when unexpected/undefined behavior occurs rather than try to hide it.

I think Aliza's working on something that might fix this, as a heads up.

orangejenny:
@ctsims 

> I think Aliza's working on something that might fix this, as a heads up.

Do you mean this - https://github.com/dimagi/commcare/pull/132 - or something else? (Aliza & I chatted, that seemed like the most related thing she's working on)

PR #517, opened 2015-08-12
Unconditional exit method to CommCareHomeActivity

dcluna:
Solving bug [178853](http://manage.dimagi.com/default.asp?178853#993131).

### DIFF
--- a/app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
+++ b/app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
@@ -63,7 +63,6 @@ import org.commcare.android.util.StorageUtils;
 import org.commcare.android.view.HorizontalMediaView;
 import org.commcare.android.view.SquareButtonWithNotification;
 import org.commcare.android.view.ViewUtil;
-import org.commcare.dalvik.BuildConfig;
 import org.commcare.dalvik.R;
 import org.commcare.dalvik.application.AndroidShortcuts;
 import org.commcare.dalvik.application.CommCareApp;
@@ -1163,7 +1162,6 @@ public class CommCareHomeActivity extends CommCareActivity<CommCareHomeActivity>
 
     /**
      * Helper method to leave CommCare unconditionally
-     * @param context
      */
     public static void exitCommCare(Context context) {
         Intent intent = new Intent(context, CommCareHomeActivity.class);
--- a/app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
+++ b/app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
@@ -4,6 +4,7 @@ import android.annotation.SuppressLint;
 import android.app.Activity;
 import android.app.AlertDialog;
 import android.app.Dialog;
+import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.SharedPreferences;
@@ -62,6 +63,7 @@ import org.commcare.android.util.StorageUtils;
 import org.commcare.android.view.HorizontalMediaView;
 import org.commcare.android.view.SquareButtonWithNotification;
 import org.commcare.android.view.ViewUtil;
+import org.commcare.dalvik.BuildConfig;
 import org.commcare.dalvik.R;
 import org.commcare.dalvik.application.AndroidShortcuts;
 import org.commcare.dalvik.application.CommCareApp;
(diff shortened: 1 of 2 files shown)


phillipm:
I'm confused by the description of this bug. If you are on the login screen and press back, why would you expect to go to the home screen?

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -1153,6 +1161,16 @@ private void dispatchHomeScreen() {
         }
     }
 
+    /**
+     * Helper method to leave CommCare unconditionally
+     * @param context
```
phillipm: Please remove javadoc method argument annotations that don't contain any content

dcluna:
The problem is that clicking 'back' on the login screen keeps you there, instead of going to the device home screen.

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
dcluna: Removed

amstone326:
Re: "The problem is that clicking 'back' on the login screen keeps you there, instead of going to the device home screen (leaving ODK)." -- Are we sure that is always the desired behavior? This seems like a very hard-coded approach unless we're 100% sure that should always be the case.

dcluna:
This is the default behavior on almost every other Android app I know of -
clicking 'back' when in the main screen leaves the app. Is there a special
reason we would want otherwise in ODK?

amstone326:
Maybe not, but it still seems really hardcoded-- Would it be better to figure out why this isn't what's happening? Presumably that means that there is something else on the backstack? Or is there another reason?

dcluna:
When leaving LoginActivity, we get returned to CommCareHomeActivity,
specifically to the dispatchHomeScreen() method. At the line
https://github.com/dimagi/commcare-odk/blob/a37ff30abcc2e07b6c5c634e656bd61cb1e0f6cc/app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java#L1119,
we get an SessionUnavailableException (
https://github.com/dimagi/commcare-odk/blob/a37ff30abcc2e07b6c5c634e656bd61cb1e0f6cc/app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java#L1138),
which makes us go to the login page again. But even if this exception
didn't happen, I assume the check for an active session would return false,
thus sending us to the login activity again.

tl,dr: An unintentional loop in the code, fixed by this PR with a "get out
unconditionally" flag.

phillipm:
Ah, I see, totally misunderstood that.

I noticed the following bug: 
 - Install an app that has the session expire after 30 secs or so (try http://bit.ly/1CRUJkY)
 - Let the session expire, which takes you to the login screen.
 - Press back, which will close the app, good
 - Press the "CommCare Login Expired" pinned notification, which pops up the login screen.
 - Login in, which exits the app :( 

Wasn't able to reproduce on the master branch

ctsims:
@phillipm @dcluna @amstone326 

Hey, sorry, just caught up on these: This seems like a bizarrely hard exit to me. The onyl reason we end up on DispatchHomeScreen is because we aren't catching/returning anywhere directly in the onActivityResult. Is there a reason why returning from Login to Home with "Cancelled" set doesn't finish instead of this?

ctsims:
Specifically 

case LOGIN_USER:
                if(resultCode == RESULT_CANCELED) {
                    //quit somehow.
                    this.finish();
                    return;
                } 

this is what I was thinking of. Is the issue that we never get here because onActivityResult happens _after_ on resume or something?

ctsims:
@phillipm @dcluna 

I see. The issue is that launching login started happening from "SessionActivityRegistraiton" instead of the current activity.

This is really sketchy in my mind. I'm worried this might result in us having more than one back stack which is really, really troubling.

Phillip/Luna it might be worth going over what made this the right approach, I think there might be a simpler mechanism and it worries me that we may be adding a lot of unneeded complexity.

phillipm:
@ctsims GitHub auto merged this when https://github.com/dimagi/commcare-odk/pull/522 was merged. https://github.com/dimagi/commcare-odk/pull/522 doesn't use tasks, so we shouldn't worry about multiple backstacks, rather it uses ```FLAG_ACTIVITY_CLEAR_TOP``` to ensure we always go back to the first CommCareHomeActivity when redirecting to the login activity.

Let me know if there are lingering concerns after looking at 522.

ctsims:
I see, so the hard launch of CommCareHomeActivity only happens when
returning to login from the external location.

Does that, like, start a new CommCareSession or anything, or does it resume
the current one?

-Clayton

phillipm:
@ctsims As far as I can tell, logging in should create a new CommCareSession no matter how the login was triggered.

PR #522, opened 2015-08-13
Exit App when back button pressed in login activity

phillipm:
Pressing the back button while on the login screen should close down the app. 

This is a different approach to https://github.com/dimagi/commcare-odk/pull/517 which addresses http://manage.dimagi.com/default.asp?178853#993131

Also fixes bug where demo user warning isn't shown on demo login and check unsent forms task isn't started on normal login


app/src/org/commcare/android/framework/SessionActivityRegistration.java
```diff
@@ -65,10 +66,15 @@ public static void registerSessionExpiration() {
...
-        i.putExtra(LoginActivity.REDIRECT_TO_HOMESCREEN, true);
+    private static void letHomeScreenRedirectToLogin(Context context) {
+        Intent i = new Intent(context.getApplicationContext(), CommCareHomeActivity.class);
+        i.setFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
         context.startActivity(i);
     }
+
+    public static void returnToLogin(Activity activity) {
+        Intent i = new Intent(activity.getApplicationContext(), LoginActivity.class);
+        i.setFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
```
phillipm: ```FLAG_ACTIVITY_CLEAR_TOP``` clears the backstack of activities down to the earliest instance of CommCareHomeActivity, which shouldn't have anything else below it on the backstack because it is the first activity launched.

app/src/org/commcare/android/framework/SessionActivityRegistration.java
```diff
@@ -65,10 +66,15 @@ public static void registerSessionExpiration() {
         }
     }
 
-    public static void returnToLogin(Context context) {
-        Intent i = new Intent(context.getApplicationContext(), LoginActivity.class);
-        i.setFlags(Intent.FLAG_ACTIVITY_REORDER_TO_FRONT);
-        i.putExtra(LoginActivity.REDIRECT_TO_HOMESCREEN, true);
+    private static void letHomeScreenRedirectToLogin(Context context) {
+        Intent i = new Intent(context.getApplicationContext(), CommCareHomeActivity.class);
+        i.setFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
```
phillipm: ```FLAG_ACTIVITY_CLEAR_TOP``` clears the backstack of activities down to the earliest instance of CommCareHomeActivity, which shouldn't have anything else below it on the backstack because it is the first activity launched.

app/src/org/commcare/android/framework/SessionActivityRegistration.java
```diff
@@ -65,10 +66,15 @@ public static void registerSessionExpiration() {
...
         context.startActivity(i);
     }
+
+    public static void returnToLogin(Activity activity) {
+        Intent i = new Intent(activity.getApplicationContext(), LoginActivity.class);
+        i.setFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
+        activity.startActivityForResult(i, CommCareHomeActivity.LOGIN_USER);
```
phillipm: Allows the launching activity to inspect the result of the ```LoginActivity```.
In the case of ```CommCareHomeActivity```, it will call ```finish``` if the result is cancel, and otherwise sync and show demo user warnings
see https://github.com/dimagi/commcare-odk/blob/master/app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java#L573-L595

app/src/org/commcare/dalvik/activities/LoginActivity.java
```diff
@@ -571,12 +556,21 @@ public CustomProgressDialog generateProgressDialog(int taskId) {
         }
         return dialog;
     }
-    
+
     @Override
     public boolean isBackEnabled() {
-        return false;
+        return true;
```
ctsims: All this does is make the "<" chevron appear next to the Activity name, btw, which I don't think was anticipated.

app/src/org/commcare/dalvik/activities/LoginActivity.java
```diff
@@ -571,12 +556,21 @@ public CustomProgressDialog generateProgressDialog(int taskId) {
...
     @Override
     public boolean isBackEnabled() {
-        return false;
+        return true;
     }
-    
+
+    @Override
+    public void onBackPressed() {
+        super.onBackPressed();
```
ctsims: This is the behavior for the "Back" button on the action bar, which isn't what we want, right? I don't think it makes sense for that to ever take you out of CommCare.

amstone326: I don't think that's true? http://developer.android.com/reference/android/app/Activity.html#onBackPressed()

ctsims: Oh, you're totally right. 

I'm confused then, since this behavior is ~identical to the stock behavior when back is pressed, I think?

phillipm: @ctsims oh, you're right, it is identical. I've removed it in https://github.com/dimagi/commcare-odk/pull/531

PR #523, opened 2015-08-14
Enable install requests from external applications with multiple apps

amstone326:
If entry into CommCareSetupActivity came from an external application (such as a bit.ly url entered into a browser), do not redirect in onResume.

### DIFF
--- a/app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
+++ b/app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
@@ -110,9 +110,15 @@ public class CommCareSetupActivity extends CommCareActivity<CommCareSetupActivit
     private CommCareApp ccApp;
 
     /**
-     * Indicates whether this activity was launched from the AppManagerActivity
+     * Indicates that this activity was launched from the AppManagerActivity
      */
-    private boolean mFromManager;
+    private boolean fromManager;
+
+    /**
+     * Indicates that this activity was launched from an outside application (such as a bit.ly
+     * url entered in a browser)
+     */
+    private boolean fromExternal;
 
     /**
      * Whether this needs to be interactive (if it's automatic, we want to skip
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -140,7 +146,7 @@
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         CommCareSetupActivity oldActivity = (CommCareSetupActivity)this.getDestroyedActivityState();
-        this.mFromManager = this.getIntent().
+        this.fromManager = this.getIntent().
```
phillipm: I like this diff :)

ctsims: This change seems odd to me. Are we not doing the 'm' thing anymore?

amstone326: I changed it because no other instance variables in this class use that syntax. My general practice with that has been to just be consistent with whatever is already being done in the class.

phillipm: I wasn't aware that was a style convention we followed.
My reasoning against the 'm' is that modern IDEs highlight class fields, so there isn't a reason to also label them with an 'm'. Plus we don't seem very good at making sure that variables prefixed by 'm' are actually class fields...

Not prefixing with 'm' does go against Google's style guide for Android platform dev, but I'm fine with that, since we don't work in that codebase.

ctsims: I don't have super strong preferences here, although I do like the
mNotation more than most types because more than just clarifying meaning in
a class, it also makes it hard to, for instance, confuse a variable
assignment with a class member assignment (and I really dislike seeing
"this" everywhere). We don't do it super consistently because we started
not-doing-it and drifted to doing it more often.

-Clayton

PR #528, opened 2015-08-14
Better error handling for unresolved xpath with auto gps enabled

amstone326:


phillipm:
Oooh, this is a fix for one of the Play store crashes, right? Nice!

PR #531, opened 2015-08-17
Remove unneeded back button related overrides

phillipm:
https://github.com/dimagi/commcare-odk/pull/522 included some changes that were essentially no-ops. This PR removes those

Also reverts the PendingIntent flag change that I don't think was correct.


app/src/org/commcare/dalvik/services/CommCareSessionService.java
```diff
@@ -589,12 +586,4 @@ private String getTickerText(int current, int total) {
...
         sessionLength = CommCarePreferences.getLoginDuration() * 1000;
     }
-
-    public boolean isMultimediaVerified(){
-        return multimediaIsVerified;
-    }
-    
-    public void setMultiMediaVerified(boolean toggle){
-        multimediaIsVerified = toggle;
-    }
```
phillipm: Noticed these are unused

app/src/org/commcare/dalvik/services/CommCareSessionService.java
```diff
@@ -215,7 +212,7 @@ private void showLoggedOutNotification() {
         // The PendingIntent to launch our activity if the user selects this notification
         Intent i = new Intent(this, CommCareHomeActivity.class);
         
-        PendingIntent contentIntent = PendingIntent.getActivity(this, 0, i, PendingIntent.FLAG_ONE_SHOT);
+        PendingIntent contentIntent = PendingIntent.getActivity(this, 0, i, PendingIntent.FLAG_CANCEL_CURRENT);
```
phillipm: This shouldn't be FLAG_ONE_SHOT because you may want to click the "Session expired" pinned notification multiple times and have it launch the login activity. I think FLAG_UPDATE_CURRENT works https://developer.android.com/reference/android/app/PendingIntent.html#FLAG_UPDATE_CURRENT

app/src/org/commcare/dalvik/activities/LoginActivity.java
```diff
@@ -557,20 +557,6 @@ public CustomProgressDialog generateProgressDialog(int taskId) {
...
-    }
-
-    @Override
-    public void onBackPressed() {
-        super.onBackPressed();
-        Intent i = new Intent();
-        setResult(RESULT_CANCELED, i);
-
-        finish();
-    }
```
phillipm: This is the default behavior of onBackPressed, so we don't need to override it.

app/src/org/commcare/dalvik/activities/LoginActivity.java
```diff
@@ -557,20 +557,6 @@ public CustomProgressDialog generateProgressDialog(int taskId) {
         return dialog;
     }
 
-    @Override
-    public boolean isBackEnabled() {
-        return true;
-    }
```
phillipm: This adds a "<" to the action bar, so isn't needed for the login screen.

ctsims:
ready to pull once this isn't conflicting

phillipm:
@ctsims merged master into it, should be good now.

PR #534, opened 2015-08-18
Fix image resizing bug

wpride:
Fix for http://manage.dimagi.com/default.asp?178867


wpride:
apparently adjustViewBounds() in a recent API to allow images to upscale to the maximum available space (what we did manually before), so we need to not do this in the default case (non-scale) case

phillipm:
Just tested this on https://www.commcarehq.org/a/ccqa/apps/view/9e0457117893a5337ea7d9e0d70e9256/releases/ and got 
screen
which seems like it is still broken?

wpride:
Did you use 89 (most recent) or 87 (most recent starred)?

phillipm:
Pretty sure it was 87. Is it working for you?

wpride:
Try 89 - on 87 this is set to full resize so is expected behavior

phillipm:
Got it. Looks good!

wpride:
Fix bug with image resizing for form entry media

PR #535, opened 2015-08-18
Refactor setting computed datum

amstone326:
-move logic for handling and setting a computed datum entirely into CommCareSession, passing the necessary evaluation context
-get evaluation context in the proper way

Cross: https://github.com/dimagi/commcare/pull/140

### DIFF
--- a/app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
+++ b/app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
@@ -43,7 +43,6 @@ import org.commcare.android.database.user.models.SessionStateDescriptor;
 import org.commcare.android.database.user.models.User;
 import org.commcare.android.framework.BreadcrumbBarFragment;
 import org.commcare.android.framework.CommCareActivity;
-import org.commcare.android.framework.SessionActivityRegistration;
 import org.commcare.android.framework.SessionAwareCommCareActivity;
 import org.commcare.android.javarosa.AndroidLogger;
 import org.commcare.android.logic.GlobalConstants;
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -880,7 +874,7 @@ private void startNextFetch() {
     // extremely long method
 
     private void readyToProceed(final CommCareSession session) {
```
ctsims: hm, so both of these receive a session as an argument that we're now ignoring, should they recieve the session wrapper instead?

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -918,28 +912,13 @@ private void handleGetDatum(CommCareSession session) {
...
-            session.setDatum("", "awful");
-        } else {
-            try {
-                session.setDatum(datum.getDataId(), XPathFuncExpr.toString(form.eval(ec)));
-            } catch (XPathException e) {
-                displayException(e);
-                return;
-            }
+            session.setComputedDatum(ec);
+            startNextFetch();
```
ctsims: this doesn't throw an XPath exception, does it? can be moved out of the try

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -918,28 +912,13 @@ private void handleGetDatum(CommCareSession session) {
...
+        } catch (XPathException e) {
+            displayException(e);
```
ctsims: does this exit the app?

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
amstone326: Not sure what you mean. readyToProceed still uses the session argument. And what do you mean by both?

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
amstone326: No-- when the dialog opened by displayException() is closed, it just calls finish()

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
ctsims: It seems bad that we're receiving a session arg, but also creating our own session reference internally from the static pointer. 

Would prefer to pass an AndroidSessionWrapper into each method instead, and use that reference rather thn create a new one.

PR #539, opened 2015-08-19
Put GraphicalActivity back in manifest

orangejenny:
http://manage.dimagi.com/default.asp?179320

@dcluna this was removed in https://github.com/dimagi/commcare-odk/pull/453 - why?


dcluna:
It was removed because Android Studio issued a warning saying that this was an 'Invalid section', so I assumed it was a no-op. Sorry for any breaking changes this might have caused.

orangejenny:
@dcluna Huh, that sounds annoying. I don't see warnings in my AS at the moment, but let's see if any pop up for anyone once this is merged...

PR #541, opened 2015-08-19
Updated pinned notifications to Android 5 standards

phillipm:
Pinned notifications weren't showing up on devices targeting API 21/22 because the standards changed to require transparency in the notification images.

Updated notification icons to have transparent backgrounds to comply with Android 5 standards. https://developer.android.com/about/versions/android-5.0-changes.html#BehaviorNotifications

Tested on API 10 and API 22 devices


ctsims:
Looks great! 

Question: Is it clear whether any older devices are gonna look odd with this change?

phillipm:
@ctsims On the API 10 device the icon is a little hard to see because it is a white logo on a silver notification background. Is it worth doing an api version check and drawing a different background on lower APIs?

Just tested on a Android 4 tablet and it looks fine.

ctsims:
should be as easy as moving the old ones to drawable-v10, right?

phillipm:
@ctsims oh, right, good call! I've changed it such that only Android 4 devices and above get the new icons.

PR #544, opened 2015-08-20
trim barcode result

wpride:


wpride:
http://manage.dimagi.com/default.asp?179642

app/src/org/commcare/android/logic/BarcodeScanListenerDefaultImpl.java
```diff
@@ -31,7 +31,7 @@ public static void onBarcodeResult(BarcodeScanListener barcodeScanListener, int
         //endregion
 
         if (resultCode == Activity.RESULT_OK) {
-            String result = intent.getStringExtra(SCAN_RESULT);
```
ctsims: QQ: Do we know these are non-null at this step?

ctsims: @wpride bump

ctsims:
@wpride Also, you need to tag your PR's as ready for review if you think they are good to pull

app/src/org/commcare/android/logic/BarcodeScanListenerDefaultImpl.java
wpride: Hmm... I _think_ we do because the barcode scan has returned status "RESULT_OK" which would be odd to do for a null response when you're explicity expecting a result. I suppose this is more contractual than hard-code though... maybe I'll add some null checking for good measure

app/src/org/commcare/android/logic/BarcodeScanListenerDefaultImpl.java
```diff
@@ -32,6 +32,10 @@ public static void onBarcodeResult(BarcodeScanListener barcodeScanListener, int
 
         if (resultCode == Activity.RESULT_OK) {
             String result = intent.getStringExtra(SCAN_RESULT);
+            // I guess technically a (bad) application could return a null result with status OK
+            if (!(result == null)){
```
phillipm: It is clearer for the reader to see ```foo != null```, please use that here and below.

PR #545, opened 2015-08-21
Make table storage names constants

phillipm:
Create constants for "RECOVERY_RESOURCE_TABLE", "GLOBAL_RESOURCE_TABLE", "UPGRADE_RESOURCE_TABLE" which were found in various parts of the code

### DIFF
--- a/app/src/org/commcare/android/database/app/AppDatabaseUpgrader.java
+++ b/app/src/org/commcare/android/database/app/AppDatabaseUpgrader.java
@@ -51,7 +51,7 @@ public class AppDatabaseUpgrader {
     private boolean upgradeTwoThree(SQLiteDatabase db) {
         db.beginTransaction();
         try {
-            TableBuilder builder = new TableBuilder(CommCareApp.RECOVERY_STORAGE_TABLE);
+            TableBuilder builder = new TableBuilder(CommCareApp.RECOVERY_STORAGE_TABLE_KEY);
             builder.addData(new Resource());
             db.execSQL(builder.getTableCreateString());
             db.setTransactionSuccessful();
@@ -64,7 +64,7 @@ public class AppDatabaseUpgrader {
     private boolean upgradeOneTwo(SQLiteDatabase db, int oldVersion, int newVersion) {
         db.beginTransaction();
         try {
-            TableBuilder builder = new TableBuilder(CommCareApp.RECOVERY_STORAGE_TABLE);
+            TableBuilder builder = new TableBuilder(CommCareApp.RECOVERY_STORAGE_TABLE_KEY);
             builder.addData(new Resource());
             db.execSQL(builder.getTableCreateString());
             db.setTransactionSuccessful();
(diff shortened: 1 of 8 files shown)


app/src/org/commcare/android/database/app/AppDatabaseUpgrader.java
```diff
@@ -76,9 +77,13 @@ private boolean upgradeOneTwo(SQLiteDatabase db, int oldVersion, int newVersion)
...
-            db.execSQL("CREATE INDEX global_index_id ON GLOBAL_RESOURCE_TABLE ( " + Resource.META_INDEX_PARENT_GUID + " )");
-            db.execSQL("CREATE INDEX upgrade_index_id ON UPGRADE_RESOURCE_TABLE ( " + Resource.META_INDEX_PARENT_GUID + " )");
-            db.execSQL("CREATE INDEX recovery_index_id ON RECOVERY_RESOURCE_TABLE ( " + Resource.META_INDEX_PARENT_GUID + " )");
+            db.execSQL("CREATE INDEX global_index_id ON " + 
+                    CommCareApp.GLOBAL_STORAGE_TABLE_KEY +
+                    " ( " + Resource.META_INDEX_PARENT_GUID + " )");
+            db.execSQL("CREATE INDEX upgrade_index_id ON " + CommCareApp.UPGRADE_STORAGE_TABLE_KEY +
+                    " ( " + Resource.META_INDEX_PARENT_GUID + " )");
+            db.execSQL("CREATE INDEX recovery_index_id ON " + CommCareApp.RECOVERY_STORAGE_TABLE_KEY +
+                    " ( " + Resource.META_INDEX_PARENT_GUID + " )");
```
phillipm: Is it sketchy for the database migration code to make use of a constant that _could possibly_ change instead of a hard-coded string?

app/src/org/commcare/dalvik/application/CommCareApp.java
```diff
@@ -56,7 +56,11 @@
     private static Stylizer mStylizer;
 
     private int resourceState;
-    
+
+    public static final String GLOBAL_STORAGE_TABLE_KEY = "GLOBAL_RESOURCE_TABLE";
+    public static final String UPGRADE_STORAGE_TABLE_KEY = "UPGRADE_RESOURCE_TABLE";
+    public static final String RECOVERY_STORAGE_TABLE_KEY = "RECOVERY_RESOURCE_TABLE";
+
```
phillipm: Is this the correct place for these constants?

app/src/org/commcare/android/database/app/AppDatabaseUpgrader.java
ctsims: Yes, I'd definitely hard-code any names during migration. Hm, that should actually be something we're doing for the other fields as well, good catch

app/src/org/commcare/dalvik/application/CommCareApp.java
ctsims: I think they should go wherever the tables are retrieved from storage, so this seems right to me.

ctsims:
?

phillipm:
@ctsims I realized that after hard-coding the database migration strings there weren't really any other references to "RECOVERY_RESOURCE_TABLE", "GLOBAL_RESOURCE_TABLE", or "UPGRADE_RESOURCE_TABLE". Hence it didn't really warrant creating constants.

Plus I didn't really want to mess with the migration code because I'm afraid of breaking something with it.

PR #546, opened 2015-08-21
fix barcode widget

wpride:


wpride:
Fix bug creating two "Get Data" buttons in the Barcode Widget

PR #549, opened 2015-08-21
Image capture resizing

amstone326:
Cross: https://github.com/dimagi/javarosa/pull/148


app/src/org/odk/collect/android/jr/extensions/ImageRestrictionExtensionParser.java
```diff
@@ -0,0 +1,28 @@
...
+        super(elementName);
+    }
+
+    @Override
+    public QuestionDataExtension parse(Element e) {
+        String s = e.getAttributeValue(XFormParser.NAMESPACE_JAVAROSA,
+                "imageDimensionScaledMax");
+        if (s != null) {
+            // Parse off the "px" and cast to int
+            int maxDimens = Integer.parseInt(s.substring(0, s.length() - 2));
```
wpride: What will happen if the pixel limit is bad (either short or not an int)? it looks like we'll hard crash - maybe we should just log and not set a pixel limit?

Also: I think I think in other places (like the case list) we allow 100px and just 100 - maybe we should do that here too? (IE just remove the "px" if it exists)

amstone326: Yea all true, will add more error checking for sure. This is far from done so maybe not worth taking the time to review yet, was just pushing my first pass.

wpride: Yeah noticed the "Don't Pull" after looking it over a bit, will chill

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -932,33 +932,33 @@ private void handleCompute(AndroidSessionWrapper asw) {
      * @param state Needed for FormRecord manipulations
      */
     private void startFormEntry(AndroidSessionWrapper state) {
```
amstone326: The only change made to this file is just fixing over-indentation

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -614,34 +526,233 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
+     *
+     * @param originalImage the image file returned by the image capture or chooser intent
+     * @param isImage if false, indicates that the image is from a signature capture, so should
+     *                not attempt to scale
+     * @return a version of the image that is in its final location but has NOT been scaled, for
+     * use in displaying the captured/chosen image on the device screen when this question widget
+     * is in view (since the purpose of scaling is to limit the size of images sent to HQ, but we
+     * still want best possible quality on the device)
+     */
+    private File moveAndScaleImage(File originalImage, boolean isImage) {
```
wpride: isImage should maybe just be shouldScale ?

app/src/org/odk/collect/android/widgets/QuestionWidget.java
```diff
@@ -187,6 +189,11 @@ public FormEntryPrompt getPrompt() {
 
     public abstract void setOnLongClickListener(OnLongClickListener l);
 
+    public void applyExtension(QuestionDataExtension extension) {
```
phillipm: Needs an override annotation

phillipm:
Nice refactors and the code you added is really clean :+1:

app/src/org/odk/collect/android/tasks/FormLoaderTask.java
```diff
@@ -133,7 +136,7 @@ protected FECWrapper doInBackground(Uri... form) {
                 fis = new FileInputStream(formXml);
                 XFormParser.registerHandler("intent", new IntentExtensionParser());
                 XFormParser.registerStructuredAction("pollsensor", new PollSensorExtensionParser());
-                fd = XFormUtils.getFormFromInputStream(fis);
+                fd = XFormUtils.getFormFromInputStream(fis, new ImageRestrictionExtensionParser());
```
phillipm: I'm curious why you didn't follow the existing pattern of registering the extension parser statically, like ```registerStrucuturedAction```. Both approaches seem fine, but it feels weird to mix them because they require different init procedures, like having to pass the extension parser into ```getFormFromInputStream``` instead of calling a static registration method.

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -614,34 +526,233 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
-            if (q instanceof IntentWidget) {
-                if(((IntentWidget) q).getFormId().equals(mFormController.getPendingCalloutFormIndex())) {
-                    bestMatch = (IntentWidget)q;
+    /**
+     * Attempts to scale down an image file based on the max dimension in the current ImageWidget.
+     * If a max dimension is set, and at least one of the dimensions of the original image exceeds
+     * that maximum, then make the larger side's dimension equal to the max dimension, and scale
+     * down the smaller side such that the original aspect ratio is maintained
+     */
+    private boolean scaleImage(File originalImage, String finalFilePath) {
```
phillipm: This file is huge. Is it possible to extract the image scaling methods you're adding here into another file?

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -614,34 +526,233 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
+
+    // In order for this method to successfully return the "current" widget, the subclass of
+    // QuestionWidget that you wish to use it for must take in a PendingCalloutInterface in its
+    // constructor (passed from WidgetFactory) and call setPendingCalloutFormIndex() when it
+    // launches its intent (See IntentWidget or ImageWidget for examples)
+    private QuestionWidget getPendingWidget() {
+        QuestionWidget bestMatch = null;
+        for (QuestionWidget q : ((ODKView)mCurrentView).getWidgets()) {
+            if (q.getFormId().equals(mFormController.getPendingCalloutFormIndex())) {
+                bestMatch = q;
```
phillipm: Do you ever expect more than one widget to have a pending callout at a time? Seems like that should never happen. If you don't, I would change this method to return the first match.

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -614,34 +526,233 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
+    }
+
+    // In order for this method to successfully return the "current" widget, the subclass of
+    // QuestionWidget that you wish to use it for must take in a PendingCalloutInterface in its
+    // constructor (passed from WidgetFactory) and call setPendingCalloutFormIndex() when it
+    // launches its intent (See IntentWidget or ImageWidget for examples)
+    private QuestionWidget getPendingWidget() {
+        QuestionWidget bestMatch = null;
+        for (QuestionWidget q : ((ODKView)mCurrentView).getWidgets()) {
+            if (q.getFormId().equals(mFormController.getPendingCalloutFormIndex())) {
```
phillipm: if ```mFormController.getPendingCalloutFormIndex``` returns ```null``` then you don't need to loop over all the question widgets. Would it be worth first checking if ```mFormController.getPendingCalloutFormIndex() ==  null```?

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -614,34 +526,233 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
+        }
+        saveAnswersForCurrentScreen(DO_NOT_EVALUATE_CONSTRAINTS);
+        refreshCurrentView();
+    }
+
+    // In order for this method to successfully return the "current" widget, the subclass of
+    // QuestionWidget that you wish to use it for must take in a PendingCalloutInterface in its
+    // constructor (passed from WidgetFactory) and call setPendingCalloutFormIndex() when it
+    // launches its intent (See IntentWidget or ImageWidget for examples)
+    private QuestionWidget getPendingWidget() {
```
phillipm: Try to avoid comments that explain code and constraints in the code. If the code changes, then you also have to update the comments.

Do you think a comment like this captures the same gist?:
"Search the view's widgets for a widget that has registered a pending callout with the form controller"

phillipm:
Good to merge, in my opinion, after considering my ```getPendingWidget``` comments

app/src/org/odk/collect/android/tasks/FormLoaderTask.java
amstone326: Did this explicitly actually, after talking to @ctsims  about it. Basically the static pattern is a really terrible one, and actually causes unwanted behavior for this feature (having 2 extension parsers registered). I really should have refactored the others that are currently static to be non-static as well, but decided to leave it for another day separately.

app/src/org/commcare/android/resource/installers/XFormAndroidInstaller.java
```diff
@@ -76,8 +77,10 @@ protected int customInstall(Resource r, Reference local, boolean upgrade) throws
 
         FormDef formDef;
         try {
-            formDef = new XFormParser(new InputStreamReader(local.getStream(), "UTF-8")).parse();
-        } catch (XFormParseException xfpe) {
+            XFormParser parser = new XFormParser(new InputStreamReader(local.getStream(), "UTF-8"));
```
ctsims: If possible it might be good to, like, unify in CommCareAndroid "this is where you get an XForm Parser from" so we don't have too keep the registraitons and extensions spread out among multiple places

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
ctsims: thanks for the note! Is helpful

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -614,16 +521,168 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
-    private IntentWidget getPendingIntentWidget() {
-        IntentWidget bestMatch = null;
+    /**
+     * Performs any necessary relocating and scaling of an image coming from either a
+     * SignatureWidget or ImageWidget (capture or choose)
+     *
+     * @param originalImage the image file returned by the image capture or chooser intent
+     * @param shouldScale if false, indicates that the image is from a signature capture, so should
+     *                not attempt to scale
+     * @return a version of the image that is in its final location but has NOT been scaled, for
```
ctsims: I think this return statement is fairly confusing. Can remove info if it makes it less abstract.

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -614,16 +521,168 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
+            // to the unmodified version
+            String rawDirPath = instanceFolder + "/raw";
+            File rawDir = new File(rawDirPath);
+            if (!rawDir.exists()) {
+                rawDir.mkdir();
+            }
+            File rawImageFile = new File (rawDirPath + "/" + imageFilename);
+            if (!originalImage.renameTo(rawImageFile)) {
+                fileToReturn = originalImage;
+                Log.e(TAG, "Failed to rename " + originalImage.getAbsolutePath());
```
ctsims: This pattern is followed across this method, but the outcomes are confusing to me.

What happens if the renames/modifications don't work? Is there a reason we're handling the if/else's if we can't complete the requested action?

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -614,16 +521,168 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
+
+        // get gp of chosen file
+        Uri selectedImage = intent.getData();
+        File originalImage = new File(FileUtils.getPath(this, selectedImage));
+
+        if (originalImage.exists()) {
+            File unscaledFinalImage = moveAndScaleImage(originalImage, true);
+
+            // Add the new image to the Media content provider so that the viewing is fast in Android 2.0+
+            ContentValues values = new ContentValues(6);
```
ctsims: This is identical to the above method and should be in a helper method.

app/src/org/odk/collect/android/activities/FormEntryActivity.java
ctsims: danny recently refactored this code, so it's possible we should check with him on whether htis is all resolved currently

app/src/org/odk/collect/android/jr/extensions/ImageRestrictionExtension.java
```diff
@@ -0,0 +1,44 @@
...
+import java.io.DataOutputStream;
+import java.io.IOException;
+
+/**
+ * Represents an extension to the "upload" question type, for setting a maximum
+ * dimension for the size of an image capture
+ *
+ * @author amstone
+ */
+public class ImageRestrictionExtension implements QuestionDataExtension {
```
ctsims: I think we can likely move this straight into /javarosa/ actually, since j2me could also in theory be extended with this info.

app/src/org/odk/collect/android/jr/extensions/ImageRestrictionExtensionParser.java
```diff
@@ -0,0 +1,37 @@
...
+                "imageDimensionScaledMax");
+        if (s != null) {
+            if (s.endsWith("px")) {
+                s = s.substring(0, s.length() - 2);
+            }
+            try {
+                int maxDimens = Integer.parseInt(s);
+                return new ImageRestrictionExtension(maxDimens);
+            } catch (NumberFormatException e) {
+                return null;
```
ctsims: Would be good for this to throw an XForm parse warning that says that we got an unexpected input. Otherwise debugging this will be tough.

app/src/org/odk/collect/android/utilities/FileUtils.java
```diff
@@ -199,6 +199,53 @@ public static String getMd5Hash(File file) {
...
+            if (width > height) {
+                bitmap = Bitmap.createScaledBitmap(bitmap, largerDimen, smallerDimen, false);
+            } else {
+                bitmap = Bitmap.createScaledBitmap(bitmap, smallerDimen, largerDimen, false);
+            }
+            // Write this scaled bitmap to the final file location
+            FileOutputStream out = null;
+            try {
+                out = new FileOutputStream(finalFilePath);
+                bitmap.compress(Bitmap.CompressFormat.JPEG, 100, out);
```
ctsims: The input file may not have been a JPG. We may need to check the file's input type when we get here, identify whether we can write the appropriate output (JPEG, PNG, etc) and if not throw a warning that says we werne't able to scale down an image of type FOO.

app/src/org/commcare/android/resource/installers/XFormAndroidInstaller.java
amstone326: Fair. Although there are also a lot of places where an XFormParser is instantiated but we're not registering the extensions with it. Doesn't really change what you said but made me realize that I'm not sure what all of those other places are doing and why they don't need the extensions?

app/src/org/odk/collect/android/activities/FormEntryActivity.java
amstone326: I actually talked with him in the process of doing this part because I saw that he had just changed it. This method is the result of combining his refactor (which introduced the PendingCalloutInterface) with what I needed, and is really the same as what he had originally besides making it return a generic QuestionWidget, so it can be used in multiple places instead of just for an IntentWidget.

app/src/org/commcare/android/resource/installers/XFormAndroidInstaller.java
```diff
@@ -76,8 +77,8 @@ protected int customInstall(Resource r, Reference local, boolean upgrade) throws
 
         FormDef formDef;
         try {
-            formDef = new XFormParser(new InputStreamReader(local.getStream(), "UTF-8")).parse();
-        } catch (XFormParseException xfpe) {
+            formDef = XFormExtensionUtils.getFormFromInputStream(local.getStream());
```
ctsims: :D Nice. The name is kind of awkward, though. Why not just XFormUtil.getFormFrom...

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -614,34 +522,177 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
+     *                not attempt to scale
+     * @return the image file that should be displayed on the device screen when this question
+     * widget is in view
+     */
+    private File moveAndScaleImage(File originalImage, boolean shouldScale) throws IOException {
+        // We want to save our final image file in the instance folder for this form, so that it
+        // gets sent to HQ with the form
+        String instanceFolder =
+                mInstancePath.substring(0, mInstancePath.lastIndexOf("/") + 1);
+        String imageFilename = System.currentTimeMillis() + ".jpg";
```
ctsims: Oh weird, have we always done it this way? I'd think we'd just try to retain the original extension?

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -614,34 +522,177 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
+        File fileToReturn;
+        if (!savedScaledImage) {
+            // If we didn't create a scaled image and save it to the final path, then relocate the
+            // original image from the temp filepath to our final path
+            File finalFile = new File(finalFilePath);
+            if (!originalImage.renameTo(finalFile)) {
+                throw new IOException("Failed to rename " + originalImage.getAbsolutePath() +
+                        " to " + finalFile.getAbsolutePath());
+            } else {
+                fileToReturn = finalFile;
```
ctsims: If there's nothing blocking it, it would reduce the complexity of this code to return here and below directly, rather than letting htem both fall through to the bottom case.

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -614,34 +522,177 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
+         * video
+         */
+
+        // The intent is empty, but we know we saved the image to the temp file
+        File originalImage = ImageWidget.TEMP_FILE_FOR_IMAGE_CAPTURE;
+        try {
+            File unscaledFinalImage = moveAndScaleImage(originalImage, isImage);
+            saveImageWidgetAnswer(unscaledFinalImage);
+        } catch (IOException e) {
+            showCustomToast(Localization.get("image.capture.not.saved"), Toast.LENGTH_LONG);
```
ctsims: We set some text in the e above, we should likely log or throw that error so we know what he issue was on a console, rather than losing the info about what was in 'e'

app/src/org/odk/collect/android/jr/extensions/ImageRestrictionExtension.java
```diff
@@ -0,0 +1,44 @@
+package org.odk.collect.android.jr.extensions;
```
ctsims: This got moved, but is still in this PR.

app/src/org/odk/collect/android/jr/extensions/ImageRestrictionExtensionParser.java
```diff
@@ -0,0 +1,38 @@
+package org.odk.collect.android.jr.extensions;
```
ctsims: Same as above, this is in JR now

app/src/org/odk/collect/android/utilities/FileUtils.java
```diff
@@ -199,6 +200,74 @@ public static String getMd5Hash(File file) {
...
+     * dimension equal to the max dimension, and scale down the smaller side such that the
+     * original aspect ratio is maintained
+     */
+    public static boolean scaleImage(File originalImage, String finalFilePath, int maxDimen) {
+        boolean scaledImage = false;
+        String extension = getExtension(originalImage.getAbsolutePath());
+        ImageWidget.ImageType type = ImageWidget.ImageType.fromExtension(extension);
+        if (type == null) {
+            // The selected image is not of a type that can be decoded to or from a bitmap
+            Log.i(t, "Could not scale image " + originalImage.getAbsolutePath() + " due to incompatible extension");
```
ctsims: If this happens, do we just use the original, or does it fail to let you use the image at all (The former is preferable)

app/src/org/commcare/android/resource/installers/XFormAndroidInstaller.java
amstone326: I didn't name it that because we already have an XFormUtil class somewhere else, but I guess that's ok?

app/src/org/odk/collect/android/activities/FormEntryActivity.java
amstone326: Oh wow yea good catch, that was there to begin with. I'll change it though.

amstone326: Actually, there are places in the comments where it talks specifically about jpg format. Are we sure there's no reason that we seem to have been insisting on that?

app/src/org/odk/collect/android/utilities/FileUtils.java
amstone326: Yep we use the original, because that's what happens when scaleImage returns false.

PR #551, opened 2015-08-25
Small LoginActivity refactor

phillipm:
Linting & remove need for ```LoginBoxesStatus``` enum

### DIFF
--- a/app/src/org/commcare/dalvik/activities/LoginActivity.java
+++ b/app/src/org/commcare/dalvik/activities/LoginActivity.java
@@ -1,6 +1,5 @@
 package org.commcare.dalvik.activities;
 
-import android.content.Context;
 import android.content.Intent;
 import android.content.SharedPreferences;
 import android.graphics.Bitmap;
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/dalvik/activities/LoginActivity.java
```diff
@@ -66,45 +64,45 @@
...
-    TextView welcomeMessage;
+    private TextView welcomeMessage;
     
-    public static final int TASK_KEY_EXCHANGE = 1;
+    private static final int TASK_KEY_EXCHANGE = 1;
     
-    SqlStorage<UserKeyRecord> storage;
-    private ArrayList<String> appIdDropdownList = new ArrayList<>();
+    private SqlStorage<UserKeyRecord> storage;
+    private final ArrayList<String> appIdDropdownList = new ArrayList<>();
```
amstone326: Why was this made final?

amstone326: (merging this because it's fine for it to be final, but just curious what made you choose to change this in particular, when there are a lot of fields all over the place that could be final and aren't)

phillipm: The linter will detect when an argument is only assigned once and suggest that you make it final. It is pretty common to see list-like instances with ```final``` qualifiers. I think it enables compiler optimizations but also tells the reader that they don't have to worry about the field ever being set to point to another object.

PR #552, opened 2015-08-25
Consolidate and improve delete method in FileUtil

amstone326:
-There were 2 methods in FileUtil.java doing the same thing (deleting a file or directory), and one was slightly wrong because it was not recursive. Consolidate into one method that is recursive and returns a boolean indicating success.
-Fix over-indentation of most of the methods in FileUtil

### DIFF
--- a/app/src/org/commcare/android/util/FileUtil.java
+++ b/app/src/org/commcare/android/util/FileUtil.java
@@ -44,19 +44,25 @@ public class FileUtil {
             }
             return made;
     }
-    
-    public static boolean deleteFile(File f) {
-        if(!f.exists()) { return true; }
-        if(!f.isDirectory()) {
-            return f.delete();
-        } else {
-            for(File child : f.listFiles()) {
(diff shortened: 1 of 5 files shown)


app/src/org/commcare/android/util/FileUtil.java
```diff
@@ -44,19 +44,25 @@ public static boolean createFolder(String path) {
...
+        }
+        if (f.isDirectory()) {
+            for (File child : f.listFiles()) {
+                if (!deleteFileOrDir(child)) {
                     return false;
                 }
             }
-            return f.delete();
         }
+        Log.i(LOG_TOKEN, "Attempting to delete file: " + f.getAbsolutePath());
```
ctsims: this seems like it's going to be incredibly verbose. Can we nix tihs?

amstone326: Oh sure, was just maintaining the logging that had existed in the other method.

PR #554, opened 2015-08-25
Subentity detail

orangejenny:
Roundabout fix for http://manage.dimagi.com/default.asp?167204

Support details that specify a nodeset. This nodeset gets expanded in the context of the case, and the detail displays a row for each node in the expanded set. Useful for mobile UCR, where the main detail object is the report and we then want to display detail with a line for each row in the report.


orangejenny:
dependent on https://github.com/dimagi/commcare/pull/150

app/src/org/commcare/android/adapters/EntitySubnodeListAdapter.java
```diff
@@ -0,0 +1,111 @@
...
+
+import java.util.List;
+
+/**
+ * Created by jschweers on 8/24/2015.
+ * <p/>
+ * Adapter for taking a nodeset, contextualizing it against an entity,
+ * and then displaying one item for each node in the resulting set.
+ */
+public class EntitySubnodeListAdapter implements ListAdapter {
```
orangejenny: Better naming suggestions would be welcome.

app/src/org/commcare/android/adapters/EntitySubnodeListAdapter.java
```diff
@@ -0,0 +1,111 @@
...
+        EntityView view = (EntityView) convertView;
+        Entity entity = this.entities.get(position);
+        if (view == null) {
+            view = new EntityView(context, detail, entity, null, null, position, false);
+        } else {
+            view.refreshViewsForNewEntity(entity, false, position);
+        }
+        int color = this.rowColors[position % 2];
+        if (color != -1) {
+            view.setBackgroundColor(color);
```
orangejenny: The striping in the regular entity detail activity is several times more complicated than this...not sure why. Tested that this works fine as you scroll and views get reused.

orangejenny:
Updated screenshot:
screenshot_2015-08-26-10-31-18

app/res/layout/entity_detail_list.xml
```diff
@@ -4,9 +4,18 @@
               android:layout_height="match_parent"
               android:orientation="vertical"
     >
+
+    <LinearLayout
+        android:id="@+id/entity_detail_header"
+        android:layout_width="fill_parent"
+        android:layout_height="wrap_content"
+        android:layout_above="@+id/screen_entity_detail_list"
+        android:orientation="horizontal">
```
ctsims: The default visibility here should be "gone", right?

ctsims:
This looks pretty good to me. One question: The EntityDetailFragment is already fairly limited in scope, and it seems really confusing to wire it together in the same class. Is it possible to separate the concerns, such that there's one fragment for contextualized entity views, and one for flat entity views?

app/res/layout/entity_detail_list.xml
orangejenny: Visibility gets set in the code (whether VISIBLE or GONE), doe sit matter if there's a default value here?

ctsims: It matters a bit due to the way the stuff gets displayed in views, but possibly not enough that we should worry about it

orangejenny: will fix it, just hadn't known it made a difference

orangejenny:
@ctsims good idea. carved out an EntitySubnodeDetailFragment subclass.

app/src/org/commcare/android/view/EntityView.java
```diff
@@ -86,35 +87,42 @@ public EntityView(Context context, Detail d, Entity e, TextToSpeech tts,
...
         super(context);
         this.context = context;
         this.views = new View[headerText.length];
         this.mHints = d.getHeaderSizeHints();
         String[] headerForms = d.getHeaderForms();
-        
-        for (int i = 0 ; i < views.length ; ++i) {
+
+        int[] colors = AndroidUtil.getThemeColorIDs(context, new int[]{R.attr.entity_view_header_background_color, R.attr.entity_view_header_text_color});
+        Log.i("DEBUG-i", "Background color is: " + colors[0] + ", text color is: " + colors[1]);
```
ctsims: hanging debug

PR #565, opened 2015-08-27
account for padding

wpride:


wpride:
http://manage.dimagi.com/default.asp?179732

PR #566, opened 2015-08-27
Forms provider fix for multiple apps

amstone326:


app/src/org/commcare/android/database/global/GlobalDatabaseUpgrader.java
```diff
@@ -50,6 +54,15 @@ private boolean upgradeOneTwo(SQLiteDatabase db, int oldVersion, int newVersion)
     }
 
     private boolean upgradeTwoThree(SQLiteDatabase db) {
```
amstone326: @ctsims It's worth noting that since upgradeTwoThree() was already merged into master (although NOT released yet), anyone who has the build from master installed on their device (i.e. us, and anyone trying out multiple apps or using it for some other special reason) will have to completely uninstall CommCare for this to work. If we don't like that, I could change the dbVersion again to 4, and put this stuff in upgradeThreeFour(). Let me know what you think.

app/src/org/commcare/android/database/global/GlobalDatabaseUpgrader.java
```diff
@@ -78,4 +91,40 @@ private boolean upgradeTwoThree(SQLiteDatabase db) {
...
+     * performs the necessary one-time migration from a global db that still exists on a device
+     * to the new per-app system
+     */
+    private boolean upgradeProviderDb(SQLiteDatabase db, ProviderUtils.ProviderType type) {
+        File oldDbFile = CommCareApplication._().getDatabasePath(type.getOldDbName());
+        if (oldDbFile.exists()) {
+            File newDbFile = CommCareApplication._().getDatabasePath(
+                    ProviderUtils.getProviderDbName(type, getInstalledAppRecord(c, db).getApplicationId()));
+            if (!oldDbFile.renameTo(newDbFile)) {
+                // Big problem, should potentially crash here ?
```
amstone326: Should this fail harder in some way?

app/src/org/commcare/android/database/global/GlobalDatabaseUpgrader.java
```diff
@@ -78,4 +91,40 @@ private boolean upgradeTwoThree(SQLiteDatabase db) {
...
+     * installed at once, we need to have one forms db and one instances db per app. This method
+     * performs the necessary one-time migration from a global db that still exists on a device
+     * to the new per-app system
+     */
+    private boolean upgradeProviderDb(SQLiteDatabase db, ProviderUtils.ProviderType type) {
+        File oldDbFile = CommCareApplication._().getDatabasePath(type.getOldDbName());
+        if (oldDbFile.exists()) {
+            File newDbFile = CommCareApplication._().getDatabasePath(
+                    ProviderUtils.getProviderDbName(type, getInstalledAppRecord(c, db).getApplicationId()));
+            if (!oldDbFile.renameTo(newDbFile)) {
```
amstone326: Should this cause some sort of harder failure than just returning false?

app/src/org/commcare/dalvik/odk/provider/FormsProvider.java
```diff
@@ -106,9 +119,12 @@ public boolean onCreate() {
     }
     
     public void init() {
-        //this is terrible, we need to be binding to the cc service, etc. Temporary code for testing
-        if(mDbHelper == null) {
-            mDbHelper = new DatabaseHelper(CommCareApplication._(), DATABASE_NAME);
+        // this is terrible, we need to be binding to the cc service, etc. Temporary code for testing
```
amstone326: ...This doesn't seem like it ended up being temporary code. Has this turned out to be acceptable, or should we be changing it?

app/src/org/commcare/android/database/global/GlobalDatabaseUpgrader.java
ctsims: Hey, 

In my mind it only depends on whether you could do them independently.

I think if you could do the upgrade on a v3 (IE: Not-released, but the v3 that was pushed), it seems like it isn't really any more work to move the upgrade to v4 and I'm fine with having multiple-bumps in a cycle.

If you are depending on the pre-v3 state to do the operations, though, it makes sense to keep it here.

amstone326: Ok cool then I'll split it out into 2 upgrades, there's no dependency that I can think of.

app/src/org/commcare/android/database/global/GlobalDatabaseUpgrader.java
ctsims: It seems like it likely should. If we can't rename the file the app won't really work. 

You could do

"Start sqldb transaction for app record.
try to move form provider
try to move instance provider
if both succeed, commit transaciton
if one fails, revert the transaction and restart commcare."

And have the app try again when it restarts.

app/src/org/commcare/android/database/global/GlobalDatabaseUpgrader.java
```diff
@@ -77,4 +90,40 @@ private boolean upgradeTwoThree(SQLiteDatabase db) {
...
+    }
+
+    private static ApplicationRecord getInstalledAppRecord(Context c, SQLiteDatabase db) {
+        SqlStorage<Persistable> storage = new SqlStorage<Persistable>(
+                ApplicationRecord.STORAGE_KEY,
+                ApplicationRecord.class,
+                new ConcreteDbHelper(c, db));
+        for (Persistable p : storage) {
+            ApplicationRecord r = (ApplicationRecord) p;
+            if (r.getStatus() == ApplicationRecord.STATUS_INSTALLED) {
```
ctsims: If there's more than one installed record we likely should either deliberately crash or print a message saying "You need to wipe and restart CommCare" if it's easy enough to do so.

app/src/org/commcare/dalvik/odk/provider/FormsProvider.java
ctsims: I think we can wipe the comment, is stable for now.

app/src/org/commcare/dalvik/odk/provider/FormsProvider.java
```diff
@@ -106,9 +118,12 @@ public boolean onCreate() {
...
+        if (mDbHelper == null || mDbHelper.getAppId() != ProviderUtils.getSeatedOrInstallingAppId()) {
```
ctsims: It worries me a bit that we have the seated or installed notion, because it concerns me that we may end up thinking we have both a seated and an "installing" app at the same time. 

Is there some sort of consistency checking to ensure that we don't have a seated app when we set the installing app, and that we don't have an installing app when we seat?

app/src/org/commcare/dalvik/odk/provider/InstanceProvider.java
```diff
@@ -97,13 +111,13 @@ public boolean onCreate() {
...
     void init() {
-        //this is terrible, we need to be binding to the cc service, etc. Temporary code for testing
-        if(mDbHelper == null) {
-            mDbHelper = new DatabaseHelper(CommCareApplication._(), DATABASE_NAME);
+        // this is terrible, we need to be binding to the cc service, etc. Temporary code for testing
+        if (mDbHelper == null || mDbHelper.getAppId() != ProviderUtils.getSeatedOrInstallingAppId()) {
+            String dbName = ProviderUtils.getProviderDbName(
+                    ProviderUtils.ProviderType.INSTANCES,
+                    ProviderUtils.getSeatedOrInstallingAppId());
+            mDbHelper = new DatabaseHelper(CommCareApplication._(), dbName);
```
ctsims: I'm a bit confused. Why don't we pass this the DB name + the app_id and have it seat against that, rather than reading the app ID from the DB name?

app/src/org/commcare/android/database/global/GlobalDatabaseUpgrader.java
amstone326: That makes sense essentially, although I'm not seeing a clear mechanism to revert/rollback in SqlLiteDatabase. Would it maybe make sense to just tell them "You need to wipe your data and restart commcare" like you suggested below? There's a small chance that just restarting commcare will fix whatever weirdness is going on causing it to fail, but if it doesn't then they're going to just be stuck in this loop of restarting and re-trying.

app/src/org/commcare/dalvik/odk/provider/InstanceProvider.java
amstone326: No idea, stupid choice.

app/src/org/commcare/android/database/MigrationException.java
```diff
@@ -0,0 +1,28 @@
+package org.commcare.android.database;
+
+/**
+ * Exception thrown when an error is encountered during global db upgrades
+ *
+ * @author amstone
+ */
+public class MigrationException extends RuntimeException {
+
+    public static final String DEFINITE_FAILURE_MESSAGE = "CommCare was unable to to migrate your " +
```
ctsims: Since these aren't semantically meaningful, I would move the messages themselves to the Android Strings resource so they can be localized by the Android system if we needed

app/src/org/commcare/android/database/global/GlobalDatabaseUpgrader.java
ctsims: The mechanism in SqlLiteDatabase is to open a transaction and then close it without calling "setTransactionSuccesful". There should be an example in this file. You'll just want to call these methods from inside the upgradeAppRecords class after you've found one-installed-app.

PR #568, opened 2015-08-28
Configure zoom placement & tweak aesthetics

orangejenny:
More from http://manage.dimagi.com/default.asp?179685

Allow zoom controls to be placed in different corners of graph instead of always bottom right, and make it a little less obtrusive. No way to do this without fiddling with AChartEngine. Increasingly considering switching chart libraries.

Most changes are AS reformatting AChartEngine code.

Before:
screenshot_2015-08-28-11-10-40

After:
screenshot_2015-08-28-10-59-48

### DIFF
--- a/app/src/org/commcare/android/view/GraphView.java
+++ b/app/src/org/commcare/android/view/GraphView.java
@@ -13,6 +13,7 @@ import org.achartengine.chart.PointStyle;
 import org.achartengine.model.TimeSeries;
 import org.achartengine.model.XYMultipleSeriesDataset;
 import org.achartengine.model.XYSeries;
+import org.achartengine.renderer.DefaultRenderer;
 import org.achartengine.renderer.XYMultipleSeriesRenderer;
 import org.achartengine.renderer.XYSeriesRenderer;
 import org.commcare.android.models.RangeXYValueSeries;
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/android/view/GraphView.java
```diff
@@ -127,6 +128,24 @@ private void setPanAndZoom(boolean allow) {
...
         mRenderer.setZoomEnabled(allow, allow);
         mRenderer.setZoomButtonsVisible(allow);
+        if (allow) {
+            DefaultRenderer.Location loc = stringToLocation(mData.getConfiguration("zoom-location", "bottom-right"));
+            mRenderer.setZoomLocation(loc);
+        }
+    }
+
+    private DefaultRenderer.Location stringToLocation(String str) {
+        // Java, why don't you have a concise way to define a hashmap?
```
phillipm: :)

phillipm: As of Java 7 you can match strings using switch statements. Doing so would clean things up a little bit.

phillipm:
Please try to split reformatting changes into separate commits so that it is easier to the functional changes.

app/src/org/commcare/android/view/GraphView.java
orangejenny: had forgotten that, good point.

PR #569, opened 2015-08-28
Universal content-aware drop shadow

ctsims:
Adds a component drop shadow to the "static" frame of views on the screen, compatible all OS versions.


app/src/org/commcare/android/fragments/SetupEnterURLFragment.java
```diff
@@ -90,6 +91,21 @@ public void onClick(final View v) {
...
+
+        if(activity != null) {
+            if (activity.getCurrentFocus() != null) {
+                InputMethodManager imm = (InputMethodManager) activity.getSystemService(Context.INPUT_METHOD_SERVICE);
+                imm.hideSoftInputFromWindow( activity.getCurrentFocus().getWindowToken(), 0);
+            }
+
+        }
+
+        super.onPause();
```
phillipm: ```super.onPause()``` should always be the first statement when overriding ```onPause```

PR #573, opened 2015-08-28
change or clear the last username in refreshForNewApp()

amstone326:


phillipm:
@amstone326 Just realized that the password isn't cleared when you switch apps. That might be nice

PR #575, opened 2015-08-31
Enable back icon & action on full-screen graphs

orangejenny:
Another part of http://manage.dimagi.com/default.asp?179685 - enabling the "back" icon in the action bar.

Code is stolen from CommCareActivity; I don't think a cleaner solution is worth the trouble right now.

### DIFF
--- a/libraries/achartengine/achartengine/src/org/achartengine/GraphicalActivity.java
+++ b/libraries/achartengine/achartengine/src/org/achartengine/GraphicalActivity.java
@@ -1,12 +1,12 @@
 /**
  * Copyright (C) 2009 - 2013 SC 4ViewSoft SRL
- *  
+ * <p/>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *  
- *      http://www.apache.org/licenses/LICENSE-2.0
- *  
+ * <p/>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p/>
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


libraries/achartengine/achartengine/src/org/achartengine/GraphicalActivity.java
```diff
@@ -1,12 +1,12 @@
...
+ * <p/>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *  
- *      http://www.apache.org/licenses/LICENSE-2.0
- *  
+ * <p/>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p/>
```
phillipm: Why are these html tags here?
In my opinion, javadocs should not have html tags. We read them at the source level, and pretty much never compile them to html, so adding hard to read html tags is not useful.

libraries/achartengine/achartengine/src/org/achartengine/GraphicalActivity.java
```diff
@@ -15,34 +15,59 @@
...
-  /** The chart to be drawn. */
-  private AbstractChart mChart;
+    /**
+     * The encapsulated graphical view.
+     */
+    private GraphicalView mView;
+    /**
+     * The chart to be drawn.
+     */
+    private AbstractChart mChart;
```
phillipm: Given the type and name of these fields, the comments don't provide any additional info.

libraries/achartengine/achartengine/src/org/achartengine/GraphicalActivity.java
```diff
@@ -15,34 +15,59 @@
...
-    String title = extras.getString(ChartFactory.TITLE);
-    if (title == null) {
-      requestWindowFeature(Window.FEATURE_NO_TITLE);
-    } else if (title.length() > 0) {
-      setTitle(title);
+    @Override
+    public boolean onOptionsItemSelected(MenuItem item) {
+        switch (item.getItemId()) {
+            case android.R.id.home:
+                System.out.println("[jls] this was unexpected");
```
phillipm: Please remove this line

libraries/achartengine/achartengine/src/org/achartengine/GraphicalActivity.java
```diff
@@ -15,34 +15,59 @@
...
-    mView = new GraphicalView(this, mChart);
-    String title = extras.getString(ChartFactory.TITLE);
-    if (title == null) {
-      requestWindowFeature(Window.FEATURE_NO_TITLE);
-    } else if (title.length() > 0) {
-      setTitle(title);
+    @Override
+    public boolean onOptionsItemSelected(MenuItem item) {
+        switch (item.getItemId()) {
+            case android.R.id.home:
```
phillipm: I prefer if/else statements in the place of switch statements that only have one or two cases. This is because switch statements have much trickery semantics: easy to fall through to the next case, easy to forget a case, default cases are often awkward or completely left out, etc.

libraries/achartengine/achartengine/src/org/achartengine/GraphicalActivity.java
```diff
@@ -15,34 +15,59 @@
...
 import android.os.Bundle;
+import android.view.MenuItem;
 import android.view.Window;
 
+import org.achartengine.chart.AbstractChart;
+
 /**
  * An activity that encapsulates a graphical view of the chart.
  */
 public class GraphicalActivity extends Activity {
```
ctsims: Is there a reason this doesn't just extend CommCareActivity? Would make the back behavior and such much easier to encapsulate.

orangejenny: Seemed awkward to make library code depend on cc-odk.

libraries/achartengine/achartengine/src/org/achartengine/GraphicalActivity.java
orangejenny: Agreed, but I don't see the point of reworking the comments in a third-party library.

libraries/achartengine/achartengine/src/org/achartengine/GraphicalActivity.java
orangejenny: reasonable. updated.

libraries/achartengine/achartengine/src/org/achartengine/GraphicalActivity.java
orangejenny: AS default formatter disagrees, apparently. But I think that makes sense. Dropped the paragraph tags.

PR #577, opened 2015-08-31
Add @Override annotation to interface implementation

phillipm:
Java will statically ensure that you actually implement the methods in an interface when you use the ```implements``` keyword. Regardless, override annotations are important for code readability and to ensure that you are actually overriding something when you think you are.


dannyroberts:
Thanks!

PR #579, opened 2015-09-01
check null

wpride:


wpride:
Fix bug causing crashes when some external application returned with invalid responses

PR #584, opened 2015-09-02
Fix for Gridview Overscroll

ctsims:
Moved Etsy Gridview to be a submodule dependency, due to the published library being fundamentally broken in landscape and applied a fix to the overscroll. 

Also updated gradle config to include the gradlew as recommended by gradle best practices

### DIFF
--- a/gradlew.bat
+++ b/gradlew.bat
@@ -1,0 +1,90 @@
+@if "%DEBUG%" == "" @echo off
+@rem ##########################################################################
+@rem
+@rem  Gradle startup script for Windows
+@rem
+@rem ##########################################################################
+
+@rem Set local scope for the variables with windows NT shell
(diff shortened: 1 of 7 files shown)


gradlew.bat
```diff
@@ -0,0 +1,90 @@
+@if "%DEBUG%" == "" @echo off
```
phillipm: Do we need this or the gradlew file?

ctsims: Hey,

I checked and the gradle and android best practices say to include the
gradle wrapper at the top level of the project files.

It's ~impossible to test or debug gradle stuff without running --info and
--debug which are only available at the terminal level, and gradlew is the
recommended way to be doing CLI gradle.

-Clayton

PR #585, opened 2015-09-02
fix timezone

wpride:


ctsims:
Hey,

Can you articulate in this ticket both what the original and new behavior is and/or make that clear in the code signature?

wpride:
@ctsims I wanted to wait until I had more information from that ticket to confirm I'm actually solving the correct problem, but the problem is best described in the SO in the code:

http://stackoverflow.com/questions/5451152/how-to-handle-jodatime-illegal-instant-due-to-time-zone-offset-transition/5451245#5451245

Essentially, we're trying to convert a moment in time that does not exist to UTC. The common example is let's say country Foo goes to daylight savings time  at 2am so that at 1:59:59 on March 31 we immediately jump to 3:00:00. Then if we try to use time 2:00:00 Mar 31 in this timezone, that time does not exist and so cannot be converted to a UTC time.

We resolve this by getting the timezone offset for the device' time zone and applying this to the date time we create from the picker (which otherwise just takes some default time value as we cast the Date to a DateTime)

phillipm:
Fixed in https://github.com/dimagi/commcare-odk/pull/602

PR #599, opened 2015-09-04
Refactor SquareButton classes to allow for rectangular buttons too

phillipm:
Views look awkward if you have 3 square buttons, 2 on one line, and the 3rd dangling on the next.

This PR refactors the custom buttons that Luna setup to allow for creation of rectangular buttons with custom images, text, and colors. Should help the 3rd button in the view described look better.

Consider reviewing commit-by-commit

Currently isn't used by any views, but will be used in the background auto update feature https://github.com/dimagi/commcare-odk/pull/507/

Has a bug where the text appears on the left of the image, instead of the right.

screen


amstone326:
Ah awesome, really good idea!

app/src/org/commcare/android/view/RectangleButtonWithText.java
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    public RectangleButtonWithText(Context context, AttributeSet attrs, int defStyleAttr) {
+        super(context, attrs, defStyleAttr);
+    }
+
+    @Override
+    int getLayoutResourceId() {
+        return R.layout.rectangle_button_text;
+    }
```
phillipm: I don't really like how `RectangleButtonWithText` and `SquareButtonWithText` implement/extend `CustomButtonWithText` by implementing `getLayoutResourceId`. A bit constrained by Android's inflation framework, but let me know if you can think of a better way to achieve this.

app/src/org/commcare/android/view/CustomButtonWithText.java
```diff
@@ -0,0 +1,126 @@
...
+    abstract int getLayoutResourceId();
+
+    protected void setUI(int backgroundColor, Drawable backgroundImg, String text, int colorButtonText) {
+        setColor(backgroundColor);
+        setImage(backgroundImg);
+        setText(text);
+        setTextColor(colorButtonText);
+    }
+
+    public void setText(String text) {
```
amstone326: It looks like all of these setters can be made private

phillipm: Oh, good call! This one actually needs to be public because it is used in `SetupKeepInstallFragment`, but a bunch of other fields/methods here needed to be private

app/src/org/commcare/android/view/SquareButtonWithNotification.java
```diff
@@ -13,53 +13,56 @@
 import org.commcare.dalvik.R;
 
 /**
+ * Square button with subtext and notification text
+ *
  * @author Daniel Luna (dcluna@dimagi.com)
  */
 public class SquareButtonWithNotification extends RelativeLayout {
```
amstone326: I didn't look super closely at this, and I'm not sure how much real value it would add besides just neatness, but is there a reason that SquareButtonWithNotification can't also extend CustomButtonWithText?

app/src/org/commcare/android/view/RectangleButtonWithText.java
amstone326: This seems fine to me. Given the way you have to use these classes explicitly in the layout files, not sure what a better option would be.

amstone326:
Looks good to me besides one high-level question I had. Should probably have one more set of eyes on it since I wasn't familiar with how the new ui button stuff was working to begin with.

app/src/org/commcare/android/view/SquareButtonWithNotification.java
phillipm: I think it is possible but I can't really figure out how to do it. There is some sorta funky stuff going on with nesting styleables (https://github.com/dimagi/commcare-odk/blob/refactor/text-image-button-restructure/app/res/values/attrs.xml#L33-L42) that would have to be ironed out. I'm going to differ this to a later date.

PR #602, opened 2015-09-08
Alternate fix for date issues which stays within Jodatime

ctsims:
Alternate fix for

http://manage.dimagi.com/default.asp?180595

which stays within the jodatime framework, but works with dates that are on edges (since it doesn't force setting a specific time)


ctsims:
Release Note:

Fix for issues with specific boundary dates on some timezones crashing form entry.

ctsims:
Old UI Version mirrored: https://github.com/dimagi/commcare-odk/pull/603

PR #606, opened 2015-09-09
Show the user why the app is crashing

phillipm:
In some instances it is important to show the user why CommCare is crashing. For instance, if a string localization fails, the user should know that the app is crashing because of it. In this case we won't report the crash to the Play Store or ACRA because the problem can and should be fixed by the user. The crash will still be reported to HQ

Also adding a folder to keep track of svgs that are used to generate drawable resources, in case we want to tweak them.

screen

screen

screen

__Release Notes:__
Show reason for app forced closure when the issue is addressable by the user.


app/src/org/commcare/android/util/CommCareExceptionHandler.java
```diff
@@ -1,26 +1,56 @@
...
+
+        if (warnUserBeforeExit(ex)) {
+            System.exit(0);
+        } else {
+            parent.uncaughtException(thread, ex);
+        }
+    }
+
+    private boolean warnUserBeforeExit(Throwable ex) {
+        if (ex instanceof NoLocalizedTextException) {
```
phillipm: Can anyone think of any other Exceptions that the user should be warned about before exiting?

app/src/org/commcare/android/util/CommCareExceptionHandler.java
```diff
@@ -1,26 +1,56 @@
...
     public void uncaughtException(Thread thread, Throwable ex) {
         ExceptionReportTask task = new ExceptionReportTask();
         task.execute(ex);
-        parent.uncaughtException(thread, ex);
+
+        if (warnUserBeforeExit(ex)) {
+            System.exit(0);
+        } else {
+            parent.uncaughtException(thread, ex);
+        }
```
phillipm: Note: if the user is shown a warning they will not be given the option to report a problem. Do you agree with this design decision?

orangejenny:
This is a cool idea, I like it. We sometimes make in-app, CHW-facing messaging simpler, basically, "Sorry we crashed, please show this to your supervisor" and then put the HQ-user-facing message inside the notification. What would you think of using that pattern here?

phillipm:
@orangejenny Thanks! What do you mean by "HQ-user-facing message inside the notification"? Like have a way to expand the view to show the messy crash message?

wpride:
@phillipm Really cool idea Phillip. Just want to confirm that this won't interfere with the ACRA/Play Store error reporting?

app/src/org/commcare/dalvik/activities/CrashWarningActivity.java
```diff
@@ -0,0 +1,79 @@
...
+        closeButton.setText(Localization.get("crash.warning.button"));
+        closeButton.setOnClickListener(new View.OnClickListener() {
+            public void onClick(View v) {
+                setResult(RESULT_OK);
+                finish();
+            }
+        });
+
+        Intent intent = getIntent();
+        setWarningText(intent);
```
wpride: Will these methods ever be performed separately/one and not the other? If not maybe makes sense to combine to one method?

app/src/org/commcare/android/util/CommCareExceptionHandler.java
wpride: possibly org.javarosa.xpath.XPathTypeMismatchException for when an XPath is bad

app/res/layout/activity_crash_warning.xml
```diff
@@ -0,0 +1,67 @@
...
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="fill_parent"
+    android:layout_height="fill_parent">
+
+    <RelativeLayout
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:paddingBottom="@dimen/content_start"
+        android:paddingLeft="@dimen/login_box_margins"
```
amstone326: I think we're trying to move to following the Android style guidelines rule that all padding/spacing should be multiples of 8dp. We should definitely do a pass at some point to remove existing dimensions in our dimens.xml file that violate that rule, so that we stop trying to use them.

app/src/org/commcare/dalvik/activities/CrashWarningActivity.java
```diff
@@ -0,0 +1,136 @@
...
+        showErrorButton = (Button)findViewById(R.id.ShowError);
+        showErrorButton.setText(Localization.get("crash.show.error.button"));
+        showErrorButton.setOnClickListener(new View.OnClickListener() {
+            public void onClick(View v) {
+                toggleErrorMessageVisibility();
+            }
+        });
+    }
+
+    private void restartCommCare() {
```
amstone326: This idea of restarting CommCare is something we do in multiple places in the code, and seem to do it differently in each one. Other places off the top of my head: SingleAppManagerActivity, UnrecoverableErrorActivity. Should we try to unify this, or is there a reason that they need to be slightly different in each place? Probably need @ctsims to weigh in on this mostly.

amstone326:
@phillipm I realized that maybe the reason the UI for this looked a little weird to me is because the error text appears between the 2 buttons. I wonder if it would look cleaner if you switched the buttons, and then had the error text appear below them both?

orangejenny:
> I wonder if it would look cleaner if you switched the buttons, and then had the error text appear below them both?

Thoughts on getting rid of the show/hide button altogether? Looks like it just toggles the red text, but eventually they have to restart, right (i just skimmed the code, am not reviewing this closely)? So it looks like you have two equally-weighted options, but you don't really.

phillipm:
@orangejenny don't we want to hide the error message from the end user as to not confuse them? Can you think of a better interface? Maybe a different button type for the show/hide button?

orangejenny:
So it's hidden by default (seriously have not looked closely at the code)? Yeah, a more subtle button for the toggle would be better. Putting it in a notification would be significantly better, though, because then it can potentially stay on the phone long enough to be shown to a supervisor.

app/res/layout/activity_crash_warning.xml
ctsims: Definitely! all of our padding should conform to using one of the preset
dimensions from the styles/dimens whenever possible.

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -246,7 +246,7 @@ private void configUi() {
 
         startButton = adapter.getButton(R.layout.home_start_button);
         if (startButton != null) {
-            startButton.setText(Localization.get("home.start"));
+            startButton.setText(Localization.get("home.start.asdf"));
```
wpride: @phillipm crashing

PR #607, opened 2015-09-09
fix scrollbar overlap in case tiles

wpride:


wpride:
http://manage.dimagi.com/default.asp?181173

ctsims:
Will:

Is there a way we can make these calculations not dependent on global screen size in general? I don't super love custom casing things like this and it makes a lot of assumptions about screen elements

wpride:
@ctsims maybe we should offline but confused what the alternative would be?

ctsims:
I think the alternative would be either grabbing the parent view size in
the onmeasure phase or making the layout dynamic. Likely the former. We can
chat about how to structure that.

-Clayton

wpride:
@ctsims I've looked around and can't find an obvious solution here, can you suggest something?

orangejenny:
looks all right to me, @ctsims is your comment blocking?

ctsims:
Pulling for now. will address the long-term implications later.

PR #608, opened 2015-09-09
fix getBitmapScaledToDisplay() to never cause OutOfMemory error

amstone326:
Fix for http://manage.dimagi.com/default.asp?181327#1009226

### DIFF
--- a/app/src/org/odk/collect/android/utilities/FileUtils.java
+++ b/app/src/org/odk/collect/android/utilities/FileUtils.java
@@ -267,33 +267,41 @@ public class FileUtils {
         return scaledImage;
     }
 
-
-
     public static Bitmap getBitmapScaledToDisplay(File f, int screenHeight, int screenWidth) {
-        // Determine image size of f
+        // Determine dimensions of original image
         BitmapFactory.Options o = new BitmapFactory.Options();
         o.inJustDecodeBounds = true;
         BitmapFactory.decodeFile(f.getAbsolutePath(), o);
+        int imageHeight = o.outHeight;
+        int imageWidth = o.outWidth;
 
-        int heightScale = o.outHeight / screenHeight;
-        int widthScale = o.outWidth / screenWidth;
-
-        // Powers of 2 work faster, sometimes, according to the doc.
-        // We're just doing closest size that still fills the screen.
+        // Get a scale-down factor -- Powers of 2 work faster according to the docs, but we're
+        // just doing closest size that still fills the screen
+        int heightScale = Math.round((float)imageHeight / screenHeight);
+        int widthScale = Math.round((float)imageWidth / screenWidth);
         int scale = Math.max(widthScale, heightScale);
-
-        // get bitmap with scale ( < 1 is the same as 1)
-        BitmapFactory.Options options = new BitmapFactory.Options();
-        options.inSampleSize = scale;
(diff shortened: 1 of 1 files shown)


app/src/org/odk/collect/android/utilities/FileUtils.java
```diff
@@ -267,34 +267,42 @@ public static boolean scaleImage(File originalImage, String finalFilePath, int m
...
-        Bitmap b = BitmapFactory.decodeFile(f.getAbsolutePath(), options);
-        if (b != null) {
-        Log.i(t,
-            "Screen is " + screenHeight + "x" + screenWidth + ".  Image has been scaled down by "
-                    + scale + " to " + b.getHeight() + "x" + b.getWidth());
+        try {
+            return BitmapFactory.decodeFile(f.getAbsolutePath(), options);
+        } catch (OutOfMemoryError e) {
+            scale++;
+            return performSafeScaleDown(f, scale);
```
phillipm: How many times do you suspect it will need to recursively call itself?  Is there any way that this could loop forever? Would it be worth aborting after some number of attempts or a given scale size?

amstone326: My understanding is that it's pretty much impossible that it would-- the reason this bug came up is that a lot of large media files are being loaded into memory at the same time (since it's in a question list), and the combined memory required was too much. But it seems like you should always be able to scale down the files enough to work, unless you were like really trying to load an unreasonable number of really really huge files at once.

amstone326: That being said, it would be really easy to add a limit, I'm just not sure what the right number would be. @wpride do you have any thoughts?

wpride:
@amstone326 @phillipm I don't imagine this would recurse more than once or twice, the image would have to be huge.

app/src/org/odk/collect/android/utilities/FileUtils.java
```diff
@@ -267,34 +267,42 @@ public static boolean scaleImage(File originalImage, String finalFilePath, int m
...
+    private static Bitmap performSafeScaleDown(File f, int scale) {
         BitmapFactory.Options options = new BitmapFactory.Options();
         options.inSampleSize = scale;
-        Bitmap b = BitmapFactory.decodeFile(f.getAbsolutePath(), options);
-        if (b != null) {
-        Log.i(t,
-            "Screen is " + screenHeight + "x" + screenWidth + ".  Image has been scaled down by "
-                    + scale + " to " + b.getHeight() + "x" + b.getWidth());
+        try {
+            return BitmapFactory.decodeFile(f.getAbsolutePath(), options);
```
ctsims: There should likely be a recursion counter here that only lets us scale down so far to prevent issues with the code diving here forever (unlikely, but we don't really control whether the error gets thrown, so is dangerous to assume it won't)

amstone326: @ctsims We just had this conversation above, and Will and I thought that could never really happen, but I can definitely add. Not sure what the right level of recursion to stop at is though.

ctsims: I don't it matters too much, I'd just pick a limit like 5 so we can't end
up recursing forever.

@wpride - it's as much to handle malicious requests as anything. I don't
like unbounded recursive calls which you don't have control over what's
triggering the recursion.

-Clayton

PR #610, opened 2015-09-09
Parameters incorrect for getTimeChartIntent

orangejenny:
R&D feature + probably only affecting TDH = don't care about replicating this on the old ui


orangejenny:
http://manage.dimagi.com/default.asp?181403

PR #611, opened 2015-09-09
re-add some logic

wpride:


wpride:
http://manage.dimagi.com/default.asp?181324

ctsims:
Will:

Can you explain what this PR is doing? What the bug was specifically and what the fix is? I can't quite tell from the code or ticket.

wpride:
@ctsims Sure - Basically, the bug was that we were returning the maximum height available as the height of the image, rather than the actual height of the scaled drawable. This logic changes it to use the correct scaled height, and also to ensure that we don't go below the minimum suggested size or above the max available size

PR #613, opened 2015-09-09
Fix for persistent tiles showing dropdown icon

ctsims:
Fix for bug where persistent tiles would show affordance for dropdown when there was no inline detail

### DIFF
--- a/app/src/org/commcare/android/framework/BreadcrumbBarFragment.java
+++ b/app/src/org/commcare/android/framework/BreadcrumbBarFragment.java
@@ -235,6 +235,13 @@ public class BreadcrumbBarFragment extends Fragment {
 
         final ImageButton infoButton = ((ImageButton)holder.findViewById(R.id.com_tile_holder_btn_open));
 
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/android/framework/BreadcrumbBarFragment.java
```diff
@@ -235,6 +235,13 @@ private View findAndLoadCaseTile(final Activity activity) {
 
         final ImageButton infoButton = ((ImageButton)holder.findViewById(R.id.com_tile_holder_btn_open));
 
+        Detail contextDetail = CommCareApplication._().getCurrentSessionWrapper().getSession().
+                getDetail(inlineDetail);
+
+        if(contextDetail == null) {
```
orangejenny: wouldn't checking `inlineDetail == null` do the same thing?

ctsims: Oh, right! Good point, this totally doens't seem right. Will revise

ctsims:
@orangejenny fixed!

PR #615, opened 2015-09-09
limit recursion depth

amstone326:


app/src/org/odk/collect/android/utilities/FileUtils.java
```diff
@@ -217,21 +217,24 @@ public static Bitmap getBitmapScaledToDisplay(File f, int screenHeight, int scre
...
+            return null;
+        }
         BitmapFactory.Options options = new BitmapFactory.Options();
         options.inSampleSize = scale;
         try {
             return BitmapFactory.decodeFile(f.getAbsolutePath(), options);
         } catch (OutOfMemoryError e) {
-            scale++;
-            return performSafeScaleDown(f, scale);
+            return performSafeScaleDown(f, ++scale, ++depth);
```
phillipm: nitpick: try to avoid using `var++` if you aren't then going to use the var being updated in multiple places. In this case `var + 1` suffices and tells the reader that you don't care about `var` being incremented anywhere else.

Also, I always get confused by the difference of `var++` and `++var`. The semantics of `var + 1` don't make me have to think at all :)

amstone326: Not sure what you mean, what about this case in particular makes that undesirable? I switched to using that because now that both variables needed to be incremented, it's much more succinct.

ctsims: Aliza: ++scale is basically saying:

scale = scale + 1;

in this case, we're not using scale again, we're passing in scale +1 as an arg, so it's potentially confusing

PR #619, opened 2015-09-10
Get Robolectric tests working again

phillipm:
Get robolectric tests working with gradle and robolectric 3.

You should be able to run them using `gradle testDebug` from CLI or Android Studio, but I'm having trouble with that.
You are better off creating a JUnit build task in Android Studio that looks like:
2015-09-11-173218_1920x1080_scrot


wpride:
@phillipm Is there a Readme or other documentation for running these?

PR #620, opened 2015-09-10
check whether we're writing too big a blob

wpride:


wpride:
cross https://github.com/dimagi/commcare/pull/162

wpride:
Fix a bug where we were weren't able to write a fixture > 1MB in size to the SQLite DB but would fail silently until the point that the user tried to use this fixture. Instead, fail during restore and give output

Not terribly happy with this fix, but think its a reasonable holdover until we can actually fix the underlying issue (we should be storing fixtures >1mb in files and saving the reference in the DB). 

The strangest part is that Android seems to think everything has gone fine with the DB write until the moment you try to read it (IE there wasn't any exception at write time). So possible solutions were trying to read the object immediately after writing and erroring if that failed, or just testing the blob length (as I did) and reporting a failure if that blob exceeded the size.

ctsims:
Should be targetting commcare_2.23

app/src/org/commcare/android/database/DbHelper.java
```diff
@@ -103,6 +104,12 @@ public ContentValues getContentValues(Externalizable e) {
                 values.put(TableBuilder.scrubName(key), value);
             }
         }
+
+        if(blob.length > 1000000){
```
ctsims: This should be 1024 ^ 2, right?

app/src/org/commcare/android/database/DbHelper.java
```diff
@@ -77,7 +78,7 @@ public DbHelper(Context c) {
         return new Pair<String, String[]>(ret, arguments);
     }
     
-    public ContentValues getContentValues(Externalizable e) {
+    public ContentValues getContentValues(Externalizable e) throws StorageFullException {
```
ctsims: I'm not super clear on StorageFull here being the exception we throw, as we likely want to differentiate between a record being too large and the storage filling up.

phillipm:
:+1:  @ctsims want to take a quick pass at it?

app/src/org/commcare/android/database/DbHelper.java
```diff
@@ -103,6 +103,10 @@ public ContentValues getContentValues(Externalizable e) {
                 values.put(TableBuilder.scrubName(key), value);
             }
         }
+
+        if(blob.length > Math.pow(1024, 2)){
```
ctsims: Quick question: Do we get the 1MB issue when we put the record in the values cursor or write it? Feels dangerous to hard code the limit instead of catching the failing write and then throwing the RecordTooLargeException  to make it explicit to the code what is failing.

wpride: I can't remember the exact path it used, but it was thrown in such an obscure way that it was very difficult to catch the actual exception at an actionable point. Considering we'll want to refactor this to use the file system at some point I didn't want to sink too much into this fix

PR #621, opened 2015-09-11
Translate 'Done' on view-only case list

orangejenny:
http://manage.dimagi.com/default.asp?179685

reviewing without whitespace will hide the AS changes


ctsims:
ooof. Sorry about that one.

PR #625, opened 2015-09-14
fix typo

wpride:


phillipm:
I pushed this fix directly to master (is that considered poor form?) so I'm going to close this out.

PR #626, opened 2015-09-14
Fix case tile density

wpride:


wpride:
this line was totally borking the density algorithm, we ended up dividing our row height calculation by some huge number so that it was effectively 0 and images weren't resizing properly

ctsims:
Hey,

I'm really confused as to what this does. 

I replaced the old density switch here 

https://github.com/dimagi/commcare-odk/commit/d02555dcd6b4e0a4504e7a2f5cbe3834e9ea3504#diff-e6a35e7c7ebe472ea0051b8cc938acc0L132

with a calculation, resulting in the density multiplier being the same for the stated screen densities, but also growing compatibly with new densities.

What does this do? This looks to me like it'll result in the density multiplier being between 1 and, like, 1.12, which would result in the original behavior on my phone where there was a bunch of empty space.

wpride:
extraDensity is something in the range of like ~100 - ~150 (in what I've seen)

So

densityRowMultiplier = 1 + extraDensity;

Was taking some value like ~125

so then:

rowHeight = screenHeight/(NUMBER_CROWS_PER_SCREEN_WIDE*densityRowMultiplier);

Resulted in rowHeight being basically 0 because the denominator was so big

So when we used this for ImageView calculations, we were basically getting 0 available space and making the images the minimum available size

In reality we want to scale the rows something like 1-1.5 times their original size (which I think we'll do with the change)

wpride:
@ctsims also updated

PR #627, opened 2015-09-14
fix density algo

wpride:


wpride:
same as https://github.com/dimagi/commcare-odk/pull/626

wpride:
@ctsims updated

phillipm:
Shall we close this because it didn't make it into 2.23?

PR #637, opened 2015-09-18
Home activity UI refactor

amstone326:


app/src/org/commcare/dalvik/activities/HomeActivityUIController.java
```diff
@@ -0,0 +1,391 @@
...
+        if (logoutButton != null) {
+            logoutButton.setText(Localization.get("home.logout"));
+            logoutButton.setNotificationText(activity.getActivityTitle());
+            adapter.notifyDataSetChanged();
+        }
+        adapter.setOnClickListenerForButton(R.layout.home_disconnect_button, getLogoutButtonListener());
+    }
+
+    private void setupViewOldFormsButton() {
+        SquareButtonWithNotification viewOldForms = adapter.getButton(R.layout.home_savedforms_button);
```
wpride: Why is this button declared in the method but others declared at the class level?

app/src/org/commcare/dalvik/activities/HomeActivityUIController.java
```diff
@@ -0,0 +1,391 @@
...
+    }
+
+    // endregion
+
+
+    // region - text setters for all buttons
+
+    private void setSyncButtonText(Pair<Long, int[]> syncDetails, String syncTextKey) {
+        if (syncTextKey == null) {
+            syncTextKey = activity.isDemoUser() ? "home.sync.demo" : "home.sync";
```
wpride: I think @phillipm hates this pattern

app/src/org/commcare/dalvik/activities/HomeActivityUIController.java
```diff
@@ -0,0 +1,391 @@
...
+            }
+        };
+    }
+
+    // endregion
+
+
+    // region - text setters for all buttons
+
+    private void setSyncButtonText(Pair<Long, int[]> syncDetails, String syncTextKey) {
```
wpride: not this PR's job but that return type of getSyncDisplayParameters is just awful: "@return A pair comprised of last sync time and an array with unsent and
     * incomplete form counts."

app/src/org/commcare/dalvik/activities/HomeActivityUIController.java
```diff
@@ -0,0 +1,391 @@
...
+        if (syncTextKey == null) {
+            syncTextKey = activity.isDemoUser() ? "home.sync.demo" : "home.sync";
+        }
+        if (syncDetails.second[0] > 0) {
```
wpride: Hard to know what this means because the syncDetails param is so obfuscated, comment might be good

app/src/org/commcare/dalvik/activities/HomeActivityUIController.java
```diff
@@ -0,0 +1,391 @@
...
+    private HomeScreenAdapter adapter;
+    private GridViewWithHeaderAndFooter gridView;
+    private StaggeredGridView newGridView;
+    private View mTopBanner;
+
+    private String syncKey = "home.sync";
+    private String lastMessageKey = "home.sync.message.last";
+    private String homeMessageKey = "home.start";
+    private String logoutMessageKey = "home.logout";
+    private Pair<Long, int[]> syncDetails;
```
wpride: this syncDetails value seems like the weirdest remaining thing here because it has a really weird class type and is the only non-UI statefulness in this class. Seems like the statefulness is unavoidable, maybe we should split out time since last sync/number unsent/number incomplete into separate variables? dunno

app/src/org/commcare/dalvik/activities/HomeActivityUIController.java
phillipm: Haha, I do have a perhaps illogical hatred of ternary operators! Not gonna worrying about it here because it is copied, not new, code :)

app/src/org/commcare/dalvik/activities/HomeActivityUIController.java
phillipm: yeah, the logic surrounding syncDetails is pretty strange; remember seeing a lot of bugs around it a while back. Will be nice to get cleaned up sometime!

app/src/org/commcare/dalvik/activities/HomeActivityUIController.java
amstone326: Because it's the only button that isn't referenced by other methods. But I can change it for consistency purposes.

app/src/org/commcare/dalvik/activities/HomeActivityUIController.java
amstone326: Ok as a short-term solution for at least making this file cleaner, I'm just pulling out the relevant info from sync details into clear class fields, and referencing those: https://github.com/dimagi/commcare-odk/commit/a55635796c5448d16efc0aefd81b9f65742a348c

app/src/org/commcare/dalvik/activities/HomeActivityUIController.java
```diff
@@ -0,0 +1,404 @@
...
+    private void setupButtons() {
+        refreshDataFromSyncDetails();
+        setupStartButton();
+        setupIncompleteFormsButton();
+        setupLogoutButton();
+        setupViewOldFormsButton();
+        setupSyncButton();
+    }
+
+    // region - setup methods for all buttons
```
phillipm: Can we take out `region` comments wherever we see them? I'm not a fan. Good method naming and organization should do that job.

app/src/org/commcare/dalvik/activities/HomeActivityUIController.java
```diff
@@ -0,0 +1,404 @@
...
+
+    public HomeActivityUIController(CommCareHomeActivity activity) {
+        this.activity = activity;
+    }
+
+    public View getTopBanner() {
+        return mTopBanner;
+    }
+
+    public void setupUI() {
```
phillipm: Might be nice to make this private and call it from inside of the `HomeActivityUIController` constructor.

app/src/org/commcare/dalvik/activities/HomeActivityUIController.java
phillipm: Unless you use regions; did't mean to rain on your party

app/src/org/commcare/dalvik/activities/HomeActivityUIController.java
```diff
@@ -0,0 +1,404 @@
...
+
+    // region - all helper methods used by refreshView()
+
+    private void refreshVersionText() {
+        TextView version = (TextView)activity.findViewById(R.id.str_version);
+        if (version == null) {
+            return;
+        }
+        version.setText(CommCareApplication._().getCurrentVersionString());
+    }
```
phillipm: I don't think the version string is actually shown on the home screen. Do you think this is a ui element that is misconfigured to not actually show up?

app/src/org/commcare/dalvik/activities/HomeActivityUIController.java
```diff
@@ -0,0 +1,404 @@
...
+
+    private final static String UNSENT_FORM_NUMBER_KEY = "unsent-number-limit";
+    private final static String UNSENT_FORM_TIME_KEY = "unsent-time-limit";
+
+    private final CommCareHomeActivity activity;
+    private SquareButtonWithNotification startButton;
+    private SquareButtonWithNotification logoutButton;
+    private SquareButtonWithNotification viewIncompleteFormsButton;
+    private SquareButtonWithNotification syncButton;
+    private SquareButtonWithNotification viewOldFormsButton;
```
phillipm: The use of "Old" is a little confusing, what about `viewSavedFormsButton`? If you agree it would be worth changing the method names associated with this field as well

app/src/org/commcare/dalvik/activities/HomeActivityUIController.java
```diff
@@ -0,0 +1,404 @@
...
+        } else {
+            syncKey = "home.sync";
+            lastMessageKey = "home.sync.message.last";
+            homeMessageKey = "home.start";
+            logoutMessageKey = "home.logout";
+        }
+    }
+
+    private void refreshHomeAndLogoutButtons() {
+        if (startButton != null) {
```
phillipm: I feel like we should be able to take out a lot of these nullity checks. Let's keep this in mind for a future refactor.

phillipm:
This PR is great!! Going to make building upon the home screen so much easier!
Made a few comments but happy to merge whenever you feel ready.

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -153,25 +133,24 @@
...
-    private GridViewWithHeaderAndFooter gridView;
-    private StaggeredGridView newGridView;
-    private View mTopBanner;
-
+    private HomeActivityUIController uiController;
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+        finishIfNotRoot();
```
ctsims: this used to bail after this call if it wasn't the root, does returning v. not returning from onCreate() in that case affect anything?

app/src/org/commcare/dalvik/activities/HomeActivityUIController.java
ctsims: Yeah, can kill this, we were waffling a bit on whether to add the version string back, but I think it's not gonna happen.

app/src/org/commcare/dalvik/activities/HomeActivityUIController.java
```diff
@@ -0,0 +1,404 @@
...
+    }
+
+    protected void displayMessage(String message, boolean bad, boolean suppressToast) {
+        if (!suppressToast) {
+            Toast.makeText(activity, message, Toast.LENGTH_LONG).show();
+        }
+        adapter.setNotificationTextForButton(R.layout.home_sync_button, message);
+    }
+
+    private boolean syncNotOk() {
```
ctsims: I think this name could be a bit less concerning. It's not that the sync state isn't ok, it's just that the user should be notified that it's been a while.

app/src/org/commcare/dalvik/activities/HomeActivityUIController.java
amstone326: Nah you're right, removing them.

app/src/org/commcare/dalvik/activities/HomeActivityUIController.java
amstone326: Adding soft asserts.

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
amstone326: fixed.

amstone326: good catch

PR #640, opened 2015-09-21
FormHierarchyActivity fixes: entry background color, back button in nested entry

phillipm:
In the form hierarchy view for an incomplete form instance, which shows a forms questions in a list view, the back button redirects to editing the incomplete form if you are in a nested entry. The back button should really traverse back up the form hierarchy. E.g. if you enter a repeat entry, the form hierarchy drills down into that repeat entry, at which point the back button should return you to the top level hierarchy view.

Also, correctly set the background color of form question entries in this view.

Before:
screen

After:
screen

Includes linting and code reformatting. I've marked the actually bug fixes with github comments.

### DIFF
--- a/app/src/org/odk/collect/android/activities/FormHierarchyActivity.java
+++ b/app/src/org/odk/collect/android/activities/FormHierarchyActivity.java
@@ -3,8 +3,6 @@ package org.odk.collect.android.activities;
 import android.app.ListActivity;
 import android.graphics.Color;
 import android.os.Bundle;
-import android.util.Log;
-import android.view.KeyEvent;
 import android.view.View;
 import android.view.View.OnClickListener;
 import android.widget.Button;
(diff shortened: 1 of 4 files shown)


app/src/org/odk/collect/android/activities/FormHierarchyActivity.java
```diff
@@ -443,13 +424,12 @@ protected void onListItemClick(ListView l, View v, int position, long id) {
...
     @Override
-    public boolean onKeyDown(int keyCode, KeyEvent event) {
-        switch (keyCode) {
-            case KeyEvent.KEYCODE_BACK:
-                FormEntryActivity.mFormController.jumpToIndex(mStartIndex);
+    public void onBackPressed() {
+        if (FormEntryActivity.mFormController.getFormIndex().isTerminal()) {
+            super.onBackPressed();
+        } else {
+            goUpLevel();
```
phillipm: Fixes back button behaviour

app/src/org/odk/collect/android/views/HierarchyElementView.java
```diff
@@ -1,44 +1,25 @@
...
+    private final ImageView mIcon;
 
     public HierarchyElementView(Context context, HierarchyElement it) {
         super(context);
 
-        RelativeLayout layout = (RelativeLayout) inflate(context, R.layout.hierarchy_element_view, null);
+        RelativeLayout layout = (RelativeLayout)inflate(context, R.layout.hierarchy_element_view, null);
 
-        layout.setBackgroundColor(it.getColor());
+        setBackgroundColor(it.getColor());
```
phillipm: Fixes background color of form question entries

ctsims:
@phillipm Can you also set this view to have the commcareactivity "back is enabled" flag set? That'll enable the "< 💮 CommCare" behavior on the action bar

phillipm:
@ctsims Updated; FormHierarchyActivity now has a action bar back button at every level

PR #642, opened 2015-09-21
Case autoselect

amstone326:
All of the real work for case list auto-select stuff is in commcare-- this PR just contains refactors and reorganizations surrounding that:
-Factored session navigation logic out of CommCareHomeActivity and into its own Android-independent module in the commcare repo
-CommCareHomeActivity now implements an interface for responding to information from that session navigation module
-Pulled some small functionality that was tied up in EntitySelectActivity logic out into static methods for use elsewhere

cross: https://github.com/dimagi/commcare/pull/164


app/src/org/commcare/android/tasks/EntityLoaderTask.java
```diff
@@ -102,21 +102,21 @@ protected void onPostExecute(Pair<List<Entity<TreeReference>>, List<TreeReferenc
     protected Pair<List<Entity<TreeReference>>, List<TreeReference>> doInBackground(TreeReference... nodeset) {
```
amstone326: just fixed indentation in this file

PR #644, opened 2015-09-22
Install a CommCare app in Robolectric tests

phillipm:
Allow tests to install apps, create test users, and then start a session with that user and app.

The extracted ccz app I've included is super basic and not actually useful to run tests over. I'll remove it once people start using other apps in tests.

Includes some largish refactors:
 - modularlized response cache handling in DataPullTask to allow for injecting testing implementations that pull data from local sources. The testing version ```DebugDataPullResponse``` isn't currently used, but should be helpful in the future.  
 - restructure and move the class that creates a demo user and user key record. went a little overboard on this one.

__The test is failing on jenkins because I need to get bouncycastle setup correctly on the build server__


unit-tests/resources/commcare-apps/flipper/profile.ccpr
```diff
@@ -0,0 +1,106 @@
...
+         update="https://www.commcarehq.org/a/flipper/apps/download/05938664cd52392fb39a3567d8d8902a/media_profile.ccpr?latest=true"
+         requiredMajor="2"
+         requiredMinor="22"
+         uniqueid="36c0bdd028d14a52cbff95bb1bfd0962"
+         name="Untitled Application"
+         descriptor="Profile File">
+    <property key="BackupMode" value="file_mode"/>
+    <property key="backup-url" value="file:///E:/CommCare.Backup"/>
+    <property key="restore-url" value="file:///E:/CommCare.Backup"/>
+    <property key="ota-restore-url" value="https://www.commcarehq.org/a/flipper/phone/restore/"/>
```
ctsims: Can we minimize these app resources a bit?

If you look at the MockApp app resources they are stripped super bare, which I'd prefer we do for these setups. Using locale files and keeping around a bunch of config properties seems like it's likely to cause confusion and not be clear on what the tests are doing.

unit-tests/src/org/commcare/android/tests/initialization/AppInitializationTest.java
```diff
@@ -0,0 +1,62 @@
...
+ * @author Phillip Mates (pmates@dimagi.com).
+ */
+@Config(application = org.commcare.dalvik.application.CommCareApplication.class,
+        constants = BuildConfig.class)
+@RunWith(CommCareTestRunner.class)
+public class AppInitializationTest {
+
+    @Before
+    public void setup() {
+        Robolectric.getBackgroundThreadScheduler().pause();
```
ctsims: dd you test without thse?

unit-tests/src/org/commcare/android/tests/initialization/AppInitializationTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Tests that use the ability to install a CommCare app and login as a test
+ * user.
+ *
+ * @author Phillip Mates (pmates@dimagi.com).
+ */
+@Config(application = org.commcare.dalvik.application.CommCareApplication.class,
+        constants = BuildConfig.class)
+@RunWith(CommCareTestRunner.class)
```
ctsims: FWIW: I think we can add the static initialization stuff to this class if we wanted, which would make the test setups much easier.

unit-tests/src/org/commcare/android/tests/initialization/AppInitializationTest.java
phillipm: oops, just took them out.

unit-tests/resources/commcare-apps/flipper/profile.ccpr
phillipm: I took out the media_suite stuff, but the string translations are needed otherwise that app won't start. This is really a poor example for a test app because it doesn't flex any of the platforms capabilities. My hope is to remove it completely after we add some real test apps.

unit-tests/src/org/commcare/android/tests/initialization/AppInitializationTest.java
phillipm: Not sure I follow. Like add the app installation to the `CommCareTestRunner`?

phillipm:
@ctsims addressed your comments, good to review again

unit-tests/src/org/commcare/android/tests/initialization/AppInitializationTest.java
ctsims: I was thinking the prototype factory and stuff, but can refactor when it's needed (IE: when we need to expand the test set) rather than pre-emptively.

PR #649, opened 2015-09-25
Log entity detail views

orangejenny:
http://manage.dimagi.com/default.asp?182839


app/src/org/commcare/dalvik/preferences/CommCarePreferences.java
```diff
@@ -85,10 +82,14 @@
     public final static String YES = "yes";
     public final static String NO = "no";
 
+    public final static String TRUE = "True";
```
orangejenny: Added true/false so I could use a checkbox on HQ instead of a yes/no dropdown.

orangejenny:
Related to, but independent of, https://github.com/dimagi/commcare-hq/pull/8361

orangejenny:
Build failure doesn't look related

`java.lang.RuntimeException: java.security.NoSuchAlgorithmException: PBEWITHSHA-256AND256BITAES-CBC-BC SecretKeyFactory not available`

PR #650, opened 2015-09-25
Soft assert when showing view at end of form

phillipm:
View creation should never happen for an end of form event. Throw a soft assert in and move the user back one step if this occurs.


czue:
cool you guys have soft asserts!

PR #652, opened 2015-09-25
Fix button press background color change

phillipm:
Fix for bug introduced in https://github.com/dimagi/commcare-odk/pull/599, in which I broke the functionality where the background color of CustomButtonWithText instances changes when you press them.


phillipm:

PR #653, opened 2015-09-27
Advance sms message cursor when looking for app install link

phillipm:
Fixes infinite loop that is entered in CommCareSetupActivity while trying to install an app. The loop was caused by the sms message query cursor not being advanced.

Also only looks 100 messages back in search of a sms message with a valid commcare install link. I added this 100 message limit in case searching through all the incoming texts on a phone takes a long time. It doesn't on my phone because it is new and doesn't have many texts, but we should test this on an older phone that has thousands of texts. If we do want to search through all texts, it should be done on a background thread that is non-UI blocking.

### DIFF
--- a/app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
+++ b/app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
@@ -89,6 +89,11 @@ public class CommCareSetupActivity extends CommCareActivity<CommCareSetupActivit
     private static final String KEY_LAST_INSTALL = "last_install_time";
 
     /**
+     * How many sms messages to scan over looking for commcare install link
+     */
+    private static final int SMS_CHECK_COUNT = 100;
+
+    /**
      * UI configuration states.
      */
     public enum UiState {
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -579,32 +584,33 @@ private void performSMSInstall(boolean installTriggeredManually){
...
-                        mTask.execute(textMessageBody);
-                        break;
-                    }
+            while (cursor.moveToNext() && messageIterationCount <= SMS_CHECK_COUNT) {
+                String textMessageBody = cursor.getString(cursor.getColumnIndex("body"));
+                messageIterationCount++;
+                if (textMessageBody.contains(GlobalConstants.SMS_INSTALL_KEY_STRING)) {
+                    RetrieveParseVerifyMessageTask mTask =
+                            new RetrieveParseVerifyMessageTask(this, installTriggeredManually);
+                    mTask.execute(textMessageBody);
```
ctsims: @wpride 

This should probably be wrapped in 

        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.HONEYCOMB) {
            mProcess.executeOnExecutor(AsyncTask.THREAD_POOL_EXECUTOR, records);
        } else {
            mProcess.execute(records);
        }


by the way, also shouldn't this task be connected to this activity using the task framework.

PR #654, opened 2015-09-28
Sms installs

wpride:


app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -600,17 +601,64 @@ private void scanSMSLinks(boolean installTriggeredManually){
...
                     RetrieveParseVerifyMessageTask mTask =
-                            new RetrieveParseVerifyMessageTask(this, installTriggeredManually);
-                    mTask.execute(textMessageBody);
+                            new RetrieveParseVerifyMessageTask<CommCareSetupActivity>(this, installTriggeredManually) {
+
+                                @Override
+                                protected void deliverResult(CommCareSetupActivity receiver, String result) {
+                                    if (installTriggeredManually) {
+                                        if (result != null) {
+                                            incomingRef = result;
```
ctsims: All of this logic needs to be executed against the receiver, not the current class instance.

It's good to wrap these calls like

deliverResult() {
    receiver.doAction(result)
}

to make sure that happens reliably

phillipm:
Can you also move the sms iteration into the background task? Currently the UI thread is block by looping through texts and launching the parse/verify task. This probably doesn't block UI for very long, but still worth doing; especially since you are already launching a task.

ctsims:
Unfortunately that'll be hard because the user's first workflow is affected
by the presence of the SMS.

If this is a huge hit on devices with, say, tons and tons of users who are
sending texts, maybe we'll need to introduce some other limitations like
only searching back to a certain period?

-Clayton

wpride:
@krsdimagi - any thoughts here, specifically WRT how far in the past we should reasonably be expecting an install text to be? I'm assuming we usually expect the install to be either pretty immediate after the SMS is received or not happen at all

krsdimagi:
@wspride I think it's a pretty safe assumption that if someone has service to receive an SMS they can also then do the install, so it makes sense that it would occur quickly.  However, having a window to allow this to occur would be nice - maybe like 10 texts or something (or even killing after 24 hours?).  What is the LOE on changing this in the future if we think that another number might turn out to be better from a UX perspective?

wpride:
@krsdimagi Would be very easy to change in the future (is an easy change itself)

PR #655, opened 2015-09-28
Singleton android apps

phillipm:
- Install more than one version of CommCare on your android device.
 - Configure build to enable providing an app ccz in the apk and automatically installing it on initial startup.

try it out: 
 - plug in your device
 - from the command line run `gradle installStandaloneDebug -Pcc_domain=corpora -Pcc_app_id=cc3315e9bdb5719c7b57d6ee5b19e797`
 - open the `Klänge by CommCare` app on your device
 - notice that the ccqa coverage app is installed. 

_Requires changing jenkins jobs_ that call `gradle assembleRelease` to specify `Standard` the build variant: `gradle assembleStandardRelease` 

I've left in the custom images and strings to demonstrate what resources you need to place where to override them


app/AndroidManifest.xml
```diff
@@ -110,26 +110,26 @@
 
         <provider
             android:name=".odk.provider.FormsProvider"
-            android:authorities="org.commcare.android.provider.odk.forms" />
+            android:authorities="${applicationId}.provider.odk.forms" />
         <provider
             android:name=".odk.provider.InstanceProvider"
-            android:authorities="org.commcare.android.provider.odk.instances" />
+            android:authorities="${applicationId}.provider.odk.instances" />
```
phillipm: Default `applicationId` is `org.commcare.dalvik`, so this will change the overall  `android:authorities="org.commcare.android.provider.odk.instances"` to
`android:authorities="org.commcare.dalvik.provider.odk.instances"`

Not really sure the effects of this change.

ctsims:
@phillipm Let me know when you have a bit to talk through this.

amstone326:
@phillipm Before this gets merged, you should create a confluence page with basic instructions on how to do this for any new app

phillipm:
@amstone326 good call: https://confluence.dimagi.com/display/MD/Gradle+build+tips+and+tricks

app/src/org/commcare/dalvik/provider/CaseDataAPI.java
```diff
@@ -13,8 +15,7 @@
  *
  */
 public class CaseDataAPI {
-    
-    public static final String AUTHORITY = "org.commcare.dalvik.case";
+    public static final String AUTHORITY = BuildConfig.CC_AUTHORITY + ".case";
```
amstone326: BuildConfig.CC_AUTHORITY is set to ${applicationId}.provider, which in the normal CommCare case would be org.commcare.dalvik.provider, right? So that would make this not equivalent to what was there before.

app/src/org/commcare/dalvik/provider/FixtureDataAPI.java
```diff
@@ -11,8 +13,7 @@
  * @author wspride
  */
 public class FixtureDataAPI {
-    
-    public static final String AUTHORITY = "org.commcare.dalvik.fixture";
+    public static final String AUTHORITY = BuildConfig.CC_AUTHORITY + ".fixture";
```
amstone326: Same issue here as above.

app/src/org/commcare/dalvik/provider/CaseDataAPI.java
phillipm: You're right! That would have really broken some shit!

amstone326:
@phillipm Do you think all of the app-specific content (the icons and banner and app name string) should stay in when the PR gets merged? It feels a little odd to have random content like that in our app, but they do serve as really useful placeholders to indicate what should get replaced by someone who wants to use this for another app. So not sure what I think.

build.gradle
```diff
@@ -202,5 +241,43 @@ android {
               systemProperty 'robolectric.logging', 'stdout'
         }
     }
+}
 
+/**
+ * Download and unpack commcare app associated with 'cc_app_id' into assets
+ * folder
+ */
+task downloadCCApp(type: Exec)
```
amstone326: This isn't getting used anymore, right? Do you still want to leave it in?

amstone326:
@phillipm Left a few comments, good to merge after they're addressed

build.gradle
phillipm: Think this is a (necessary) declaration. `downloadCCApp` is still being used to trigger the acquisition of the ccz from HQ via a `wget` call

phillipm:
They add some cruft but I don't see a better solution then to leave them in

build.gradle
amstone326: Oh sorry I was thinking about having changed how this worked for parto, but that's just on that branch

PR #657, opened 2015-09-28
Update text to properly call .CCZ files instead of Zip

wpride:


wpride:
Re: Jennifer's comments on global field-dev

PR #662, opened 2015-09-30
Don't crash unregistering a receiver that wasn't previously registered

phillipm:
Fix crash being seen on ACRA where, in onPause, unregistering the session expiration receiver that wasn't previously registered causes force close. Not sure how this could ever happen, because it should always be registered in onResume, but might as well not fail hard when this happens.

https://commcare-acralyzer.cloudant.com/acralyzer/_design/acralyzer/index.html#/reports-browser/commcare-odk/bug/524bbfea0dcd4c2ba3c645f0bb97e485

### DIFF
--- a/app/src/org/commcare/android/framework/SessionActivityRegistration.java
+++ b/app/src/org/commcare/android/framework/SessionActivityRegistration.java
@@ -41,13 +41,14 @@ public class SessionActivityRegistration {
      * methods of activities that are session sensitive.
      */
     public static void handleOrListenForSessionExpiration(Activity activity) {
+        activity.registerReceiver(userSessionExpiredReceiver, expirationFilter);
+
         synchronized (registrationLock) {
             if (unredirectedSessionExpiration) {
                 unredirectedSessionExpiration = false;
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/android/framework/SessionActivityRegistration.java
```diff
@@ -52,7 +56,12 @@ public static void handleOrListenForSessionExpiration(Activity activity) {
      * this method in onPause methods of activities that are session sensitive.
      */
     public static void unregisterSessionExpirationReceiver(Activity activity) {
-        activity.unregisterReceiver(userSessionExpiredReceiver);
+        try {
+            activity.unregisterReceiver(userSessionExpiredReceiver);
```
ctsims: Is there any way to do this without  catching the exception? I'm guessing 'no', but generally am a bit squicked by try/catch module rather than checking explicitly

phillipm: Sadly, not that I could find.

PR #664, opened 2015-09-30
Add export attribute to providers

wpride:
Apparently our content providers are broken because at some point this attribute stopped being "true" by default. I can't find a lot of great information about when/what caused this

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -350,7 +350,7 @@
             android:theme="@style/Theme.Transparent" >
         </activity>
         <activity android:name=".activities.KeyAccessRequestActivity"
-            android:exported="true">>
+            android:exported="true">
             <intent-filter>
                 <action android:name="org.commcare.dalvik.action.CommCareKeyAccessRequest" />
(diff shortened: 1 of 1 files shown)


wpride:
Apparently due to switching target SDK to 22 in this release: 

http://developer.android.com/guide/topics/manifest/provider-element.html

"The default value is "true" for applications that set either android:minSdkVersion or android:targetSdkVersion to "16" or lower. For applications that set either of these attributes to "17" or higher, the default is "false”."

app/AndroidManifest.xml
```diff
@@ -345,10 +349,10 @@
             android:name=".activities.UnrecoverableErrorActivity"
             android:theme="@style/Theme.Transparent" >
         </activity>
-        <activity android:name=".activities.KeyAccessRequestActivity" >
+        <activity android:name=".activities.KeyAccessRequestActivity"
+            android:exported="true">>
```
phillipm: `>>` typo

wpride:
@phillipm fixed

PR #671, opened 2015-10-02
Add missing empty constructor for IntentCallout serialization

phillipm:
Add missing empty constructor for IntentCallout serialization because it was causing form deserialization to fail.


ctsims:
@phillipm 

Is there an easy test we can add to make sure IntentCallout serialization is working? There's a test file for formdef serializations

PR #672, opened 2015-10-05
Stacked bar graphs

orangejenny:
Played around with this a few weeks ago. Turns out TDH actually wants it released, ideally for 2.24.

Has some formatting changes mixed it, but `?w=1` cleans them up.


orangejenny:
ticket: http://manage.dimagi.com/default.asp?183614

PR #673, opened 2015-10-05
Scale down images in the case list 

amstone326:
Fix for: http://manage.dimagi.com/default.asp?183486

Images embedded in the case list were not being scaled down from their original size at all, causing OOM errors very quickly if the original images were at all large.

### DIFF
--- a/app/src/org/commcare/android/view/EntityView.java
+++ b/app/src/org/commcare/android/view/EntityView.java
@@ -30,10 +30,12 @@ import org.javarosa.core.reference.InvalidReferenceException;
 import org.javarosa.core.reference.ReferenceManager;
 import org.javarosa.core.services.Logger;
 import org.javarosa.core.services.locale.Localization;
+import org.odk.collect.android.utilities.FileUtils;
 import org.odk.collect.android.views.media.AudioButton;
 import org.odk.collect.android.views.media.ViewId;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.util.Hashtable;
 import java.util.Vector;
 
@@ -54,6 +56,7 @@ public class EntityView extends LinearLayout {
     public static final String FORM_IMAGE = "image";
     public static final String FORM_GRAPH = "graph";
     public static final String FORM_CALLLOUT = "callout";
+    private static final int CASE_LIST_IMAGE_MAX_DIMEN = 150;
 
     private boolean mFuzzySearchEnabled = true;
     private boolean mIsAsynchronous = false;
@@ -254,18 +257,23 @@ public class EntityView extends LinearLayout {
     }
 
 
-    /*
-    * Updates the ImageView layout that is passed in, based on the
-    * new id and source
-    */
+    /**
+     * Updates the ImageView layout that is passed in, based on the new id and source
+     */
     public void setupImageLayout(View layout, final String source) {
         ImageView iv = (ImageView) layout;
         Bitmap b;
         if (!source.equals("")) {
             try {
-                b = BitmapFactory.decodeStream(ReferenceManager._().DeriveReference(source).getStream());
+                InputStream imageStream = ReferenceManager._().DeriveReference(source).getStream();
+                b = FileUtils.getScaledBitmap(imageStream, CASE_LIST_IMAGE_MAX_DIMEN);
                 if (b == null) {
-                    //Input stream could not be used to derive bitmap, so showing error-indicating image
+                    // Means we didn't need to scale down the image, so just decode it normally
+                    b = BitmapFactory.decodeStream(imageStream);
+                }
+                if (b == null) {
+                    // Means the input stream could not be used to derive bitmap, so showing
+                    // error-indicating image
                     iv.setImageDrawable(getResources().getDrawable(R.drawable.ic_menu_archive));
                 } else {
                     iv.setImageBitmap(b);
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/android/view/EntityView.java
```diff
@@ -54,6 +56,7 @@
     public static final String FORM_IMAGE = "image";
     public static final String FORM_GRAPH = "graph";
     public static final String FORM_CALLLOUT = "callout";
+    private static final int CASE_LIST_IMAGE_MAX_DIMEN = 150;
```
amstone326: I chose this as a number that just seemed reasonable and wasn't causing crashes with a bunch of cases. I definitely don't think it should be any bigger because the case list images only every show up as very small anyway, but possible that it should be smaller? If anyone wants to weight in, feel free.

orangejenny: How about using the actual width of the column instead?

orangejenny:
Hey, I think this needs to handle pixel density. This ticket has some background on a similar issue: http://manage.dimagi.com/default.asp?181590 (ctrl+F "painfully")

ctsims:
yeah, we have a slightly more coordinated strategy on this that we're trying to get done, so we should have a conversation about it before we pullt his.

amstone326:
@orangejenny, @ctsims ready to be reviewed again. Jenny- Clayton agreed that the solution you proposed is the right strategy here, so that's what I implemented.

app/src/org/commcare/android/view/EntityView.java
```diff
@@ -253,19 +260,46 @@ public void onClick(View v) {
         }
     }
 
+    private void addLayoutToRedrawQueue(View layout, String source) {
```
orangejenny: not convinced this function is adding value, given that `imagesViewsToRedraw` is a nice descriptive name, but I don't feel that strongly about it

app/src/org/commcare/android/view/EntityView.java
```diff
@@ -505,8 +539,12 @@ protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
             }
         }
 
-        // Re-calculate the view's measurements based on the percentage
-        // adjustments above
+        // Re-calculate the view's measurements based on the percentage adjustments above
         super.onMeasure(widthMeasureSpec, heightMeasureSpec);
+
+        onMeasureCalled = true;
+        if (imageViewsToRedraw.size() > 0) {
```
orangejenny: the `if` is redundant

app/src/org/odk/collect/android/utilities/FileUtils.java
```diff
@@ -230,30 +228,18 @@ public static boolean scaleImage(File originalImage, String finalFilePath, int m
...
             FileOutputStream out = null;
             try {
                 out = new FileOutputStream(finalFilePath);
-                bitmap.compress(type.getCompressFormat(), 100, out);
-                scaledImage = true;
+                scaledBitmap.compress(type.getCompressFormat(), 100, out);
+                return true;
             } catch (Exception e) {
                 e.printStackTrace();
+                return false;
```
orangejenny: java question: will the `finally` block still get executed if there's a return statement in the `catch`?

app/src/org/commcare/android/view/EntityView.java
```diff
@@ -505,8 +539,12 @@ protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
...
         }
 
-        // Re-calculate the view's measurements based on the percentage
-        // adjustments above
+        // Re-calculate the view's measurements based on the percentage adjustments above
         super.onMeasure(widthMeasureSpec, heightMeasureSpec);
+
+        onMeasureCalled = true;
+        if (imageViewsToRedraw.size() > 0) {
+            redrawImageLayoutsInQueue();
```
orangejenny: does this need to be called every time `onMeasure` executes, or only the first time? i don't think it makes a difference to the code, just curious for my own understanding.

app/src/org/odk/collect/android/utilities/FileUtils.java
```diff
@@ -214,14 +214,12 @@ private static String last(String[] strings) {
...
-     * Attempts to scale down an image file based on the max dimension given. If at least one of
-     * the dimensions of the original image exceeds that maximum, then make the larger side's
-     * dimension equal to the max dimension, and scale down the smaller side such that the
-     * original aspect ratio is maintained
+     * @return whether or not originalImage was scaled down according to maxDimen, and saved to
+     * the location given by finalFilePath
      */
-    public static boolean scaleImage(File originalImage, String finalFilePath, int maxDimen) {
-        boolean scaledImage = false;
+    public static boolean scaleAndSaveImage(File originalImage, String finalFilePath, int maxDimen) {
```
orangejenny: good rename

orangejenny:
looks reasonable to me, though I'd appreciate another set of eyes ( @wpride ?)

app/src/org/odk/collect/android/utilities/FileUtils.java
wpride: finally will always be called unless there's a system exit afaik

wpride:
Looks good to me too. @amstone326 the CachingAsyncImageLoader class might be of interest to you here

app/src/org/commcare/android/view/EntityView.java
amstone326: No, it doesn't- it just doesn't make a material difference in code execution because if the queue is empty then nothing happens (essentially the same reason you noted that the if statement is unnecessary).

PR #674, opened 2015-10-05
Drop chevron from read-only entity detail (horizontal view)

orangejenny:
In https://github.com/dimagi/commcare-odk/pull/645 I forgot to also make the change for the horizontal entity select+detail screen.


orangejenny:
ticket (for both PRs): http://manage.dimagi.com/default.asp?183610

PR #675, opened 2015-10-05
Saved forms are being purged at incorrect times

phillipm:
Saved forms were being purged on login whenever the `cc-days-form-retain` property in the profile.ccpr is set to anything that isn't -1.

It was due to wonky date processing, perhaps because form records are using java's string representation for dates. Java's string representation for dates unfortunately uses text time zone abbreviations, whose non-uniqueness can cause problems. We should eventually construct a migration to move away from that date representation.

This PR introduces tests to check form purging logic. This includes loading saved form instances from xml payload files. So if you ever want to use that in your tests, look here.

I also did some other refactoring because I thought it would be needed to load saved forms in via tests. Turns out it wasn't needed; I'll comment those areas so you can ignore them.

http://manage.dimagi.com/default.asp?183576


app/src/org/commcare/android/framework/CommCareActivity.java
```diff
@@ -383,15 +382,6 @@ void attemptDismissDialog() {
     }
 
     /**
-     * Handle an error in task execution.
-     */
-    protected void taskError(Exception e) {
-        //TODO: For forms with good error reporting, integrate that
-        Toast.makeText(this, Localization.get("activity.task.error.generic", new String[]{e.getMessage()}), Toast.LENGTH_LONG).show();
-        Logger.log(AndroidLogger.TYPE_ERROR_WORKFLOW, e.getMessage());
-    }
```
phillipm: Used by logic which is now in `ArchivedFormRemoteRestore`, which is only used by developers to load saved forms onto a phone manually. Taking it out because these errors are also logged elsewhere.

app/src/org/commcare/android/storage/FormSaveUtil.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * @author Phillip Mates (pmates@dimagi.com).
+ */
+public class FormSaveUtil {
+
+    /**
+     * @param context used to query form instances through android's content provider framework
+     * @return A mapping from an installed form's namespace to its install path.
+     */
+    public static Hashtable<String, String> getNamespaceToFilePathMap(Context context) {
```
phillipm: Broke out this logic into a helper function from `DataPullTask` so that it can also be used in tests to load saved form instances.

app/src/org/commcare/dalvik/activities/FormRecordListActivity.java
```diff
@@ -502,105 +497,15 @@ public boolean onPrepareOptionsMenu(Menu menu) {
...
-                    @Override
-                    protected void deliverError(FormRecordListActivity receiver, Exception e) {
-                        receiver.taskError(e);
-                    }
-                };
-                pull.connect(this);
-                pull.execute();
+                ArchivedFormRemoteRestore<FormRecordListActivity> archiveRemoteForms =
+                        new ArchivedFormRemoteRestore<>(this, this, platform);
+                archiveRemoteForms.pullArchivedFormsFromServer(source);
```
phillipm: Launches task that is only used by developers to saved load forms manually onto phones.

app/src/org/commcare/xml/AndroidCaseXmlParser.java
```diff
@@ -46,7 +46,8 @@
     EntityStorageCache mEntityCache;
     CaseIndexTable mCaseIndexTable;
     
-    public AndroidCaseXmlParser(KXmlParser parser, IStorageUtilityIndexed storage, EntityStorageCache entityCache, CaseIndexTable indexTable) {
+    public AndroidCaseXmlParser(KXmlParser parser, IStorageUtilityIndexed storage,
```
phillipm: Changes in this file are only line length related

app/src/org/odk/collect/android/listeners/FormDownloaderListener.java
```diff
@@ -1,26 +0,0 @@
...
-
-package org.odk.collect.android.listeners;
-
-import java.util.HashMap;
-
-
-/**
- * @author Carl Hartung (carlhartung@gmail.com)
- */
-public interface FormDownloaderListener {
```
phillipm: Unused

app/src/org/odk/collect/android/logic/ArchivedFormManagement.java
```diff
@@ -0,0 +1,102 @@
...
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Vector;
+
+/**
+ * Determines when saved forms should be purged.
+ *
+ * @author Phillip Mates (pmates@dimagi.com).
+ */
+public class ArchivedFormManagement {
```
phillipm: The fix: reworking of the purge logic lifted from CommCareActivity. Now uses Joda date time for bulk of the computations (delegating through java's date parsing for the time being; we need to run a migration to remove this parse dependency)

wpride:
Looks good to me

app/src/org/odk/collect/android/logic/ArchivedFormRemoteRestore.java
```diff
@@ -0,0 +1,131 @@
...
+        DataPullTask<A> pull = new DataPullTask<A>(u.getUsername(), u.getCachedPwd(), remoteUrl, taskReceiver) {
+            @Override
+            protected void deliverResult(A receiver, Integer status) {
+                switch (status) {
+                    case DataPullTask.DOWNLOAD_SUCCESS:
+                        FormRecordCleanupTask<A> task = new FormRecordCleanupTask<A>(taskReceiver, platform, CLEANUP_ID) {
+
+                            @Override
+                            protected void deliverResult(A receiver, Integer result) {
+                                if (receiver instanceof FormRecordListActivity) {
```
amstone326: A little confused- do you mean to take this out before this gets merged?

app/src/org/odk/collect/android/logic/ArchivedFormRemoteRestore.java
```diff
@@ -0,0 +1,131 @@
...
+import org.javarosa.core.services.locale.Localization;
+
+/**
+ * Only used for developer debugging.
+ *
+ * Load saved form instances manually from a xml payload.
+ *
+ * @author Phillip Mates (pmates@dimagi.com).
+ */
+public class ArchivedFormRemoteRestore<A extends CommCareActivity> {
```
amstone326: I'm guessing I'm missing something nuanced here, but this class seems overly complex. Why does it need to use generics for the Activity type, since there's only one Activity that ever uses it? Worth nothing that if that isn't in fact necessary, you could simplify the class into just a single static method that lives in ArchivedFormManagement.java, which would be nice for having all of the archived form stuff in one place.

app/src/org/odk/collect/android/logic/ArchivedFormManagement.java
```diff
@@ -0,0 +1,102 @@
...
+    /**
+     * Read the validity range for archived forms out of app preferences.
+     *
+     * @return how long archived forms should kept on the phone. -1 if saved
+     * forms should be kept indefinitely
+     */
+    public static int getArchivedFormsValidityInDays(CommCareApp app) {
+        int daysForReview = -1;
+        String daysToPurge =
+                app.getAppPreferences().getString("cc-days-form-retain", "-1");
```
amstone326: I generally put keys like this into a static class-level variable, although I know this is only being used once so not nearly as important. Just looked funny to me but might just be force of my own habit, so up to you.

app/src/org/odk/collect/android/logic/ArchivedFormRemoteRestore.java
```diff
@@ -0,0 +1,131 @@
...
+
+    public ArchivedFormRemoteRestore(final CommCareTaskConnector<A> taskConnector,
+                                     final A taskReceiver,
+                                     final CommCarePlatform platform) {
+        this.taskConnector = taskConnector;
+        this.taskReceiver = taskReceiver;
+        this.platform = platform;
+    }
+
+    public void pullArchivedFormsFromServer(String remoteUrl) {
```
amstone326: Really like the pattern of putting the execution of tasks like this into their own class! There are a few other big places in the code where we should definitely do something similar

unit-tests/resources/commcare-apps/archive_form_tests/modules-0/forms-0.xml
```diff
@@ -0,0 +1,25 @@
+<h:html xmlns:h="http://www.w3.org/1999/xhtml" xmlns:orx="http://openrosa.org/jr/xforms" xmlns="http://www.w3.org/2002/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:jr="http://openrosa.org/javarosa" xmlns:vellum="http://commcarehq.org/xforms/vellum">
```
amstone326: I know @ctsims is in favor of making these test resources as lean as possible. I'm personally very bad at determining what exactly can and can't be removed, but this looks like it probably has a bunch of stuff in it that isn't needed

amstone326:
@phillipm Left a couple of questions, good to merge once they're addressed!

app/src/org/odk/collect/android/logic/ArchivedFormRemoteRestore.java
phillipm: I over generalized here because I thought I was going to use this task from within a test, as well as where it is being used in the app code. I figured I'd leave it in this state in case we do ever need it in a test (probably super unlikely...). I can remove the generics if you think that is the best way to leave things.

I don't want to mix it with `ArchivedFormManagement` because that code is actually used in production, whereas this is a debug feature that manually has to be enabled by the dev (ugly, but useful at times)

app/src/org/odk/collect/android/logic/ArchivedFormManagement.java
phillipm: Good call; fixed

app/src/org/odk/collect/android/logic/ArchivedFormRemoteRestore.java
phillipm: Glad you like it. It does force us to make certain methods in the receiving activity public. I think this should be fine if we are careful in the way we design the activities.

app/src/org/odk/collect/android/logic/ArchivedFormRemoteRestore.java
amstone326: Ah ok, makes sense on leaving it in it's own class then, but I think I'd advocate removing the generics. Significantly overcomplicating a class for a potential test that is likely not to get written doesn't seem like the right choice. Good to merge once that's changed, and possibly leaning out the test resource file I commented on? Sorry I didn't offer any useful comments on what to take out, but don't want to mislead you- Clayton could definitely help with that.

phillipm: @amstone326 removed the generics, should be good to go.
Realized that this should be occurring in a background task (Kriti was complaining about recent slowness). I'll get that going, but it should go into a separate PR

PR #680, opened 2015-10-07
Enlarge graph points

orangejenny:
http://manage.dimagi.com/default.asp?183803

`w=1`

### DIFF
--- a/app/src/org/commcare/android/view/GraphView.java
+++ b/app/src/org/commcare/android/view/GraphView.java
@@ -49,7 +49,7 @@ public class GraphView {
 
     public GraphView(Context context, String title) {
         mContext = context;
-        mTextSize = (int) context.getResources().getDimension(R.dimen.text_large);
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/android/view/GraphView.java
```diff
@@ -405,6 +405,7 @@ private void configureSeries(SeriesData s, XYSeriesRenderer currentRenderer) {
             currentRenderer.setPointStyle(style);
             currentRenderer.setFillPoints(true);
             currentRenderer.setPointStrokeWidth(2);
+            mRenderer.setPointSize(6);
```
phillipm: I wonder if it would be a good idea to try to scale this relative to the view's resolution.

orangejenny: That is an interesting idea. Not going to do it right now, will keep in mind when migrating graph libraries.

PR #684, opened 2015-10-08
Use the latest support libs

phillipm:
Linting build.gradle :)

This will require you to launch your sdk manager and downloaded the corresponding updates, but it shouldn't be a big download.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,7 +15,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.3.0'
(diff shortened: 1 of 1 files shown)


ctsims:
@phillipm  - Do we need to update the SDK manager on the build server as well?

phillipm:
oh, right. I'll do that now

phillipm:
@ctsims should be good to go

ctsims:
@phillipm Hey,

this broke on jenkins so I had to revert:

http://jenkins.dimagi.com/job/commcare-odk/407739/

PR #686, opened 2015-10-12
Perform purging of invalid archived forms in backgroud

phillipm:
Users were reporting lag upon login that was probably associated with slow archived form purging. Do this in the background. Also block users from accessing the saved forms list while the purging task is running. 

Includes some linting/small refactors to the task connector framework even though I ended up not using task connectors. Instead I chose to use the framework I created in the background update PR (https://github.com/dimagi/commcare-odk/pull/507) because they both have the same pattern of launching singleton tasks in one place that need to be connected to specific activities when they are launched. I will generalize this framework when background update PR gets merged.

Can't figure out how to get the created `CustomProgressDialog` to not cancel when the back button is pressed. Ideas?

### DIFF
--- a/app/assets/locales/messages_ccodk_default.txt
+++ b/app/assets/locales/messages_ccodk_default.txt
@@ -153,6 +153,9 @@ form.record.unsent=Unsent
 form.record.gone=[Form Removed on Current Version]
 form.record.gone.message=The form that was used to enter this data is no longer on the current version of the app, and can't be viewed.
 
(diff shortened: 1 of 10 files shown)


ctsims:
@amstone326  @phillipm Let's treat the customdialog issue as a separate one.

Aliza: It looks like even when the custom dialog isn't cancellable it goes away when you hit 'Back'. Can you look into that? Seems likely to cause issues, especially if we are unifying to the new common framework

PR #687, opened 2015-10-13
Show user error message for XPathTypeMismatchException app crashes

phillipm:
Helped Kriti debug an app today by simply providing her with the `XPathTypeMismatchException` error message `Location /data/ql_dep_drug_history/dep_other_detail was not found`.

She said it would be really helpful if the app just told her that, so I decided to enable it for the https://github.com/dimagi/commcare-odk/pull/606 feature I implemented.

NOTE: this prevents the crash from being reported on ACRA. I don't think there is any way to report to ACRA due to threading issues. Figure it is much more valuable for the user to see these crashes than for us to see them. I did make sure the exception log submission task is called when crashes are caught this way.

### DIFF
--- a/app/src/org/commcare/android/util/CommCareExceptionHandler.java
+++ b/app/src/org/commcare/android/util/CommCareExceptionHandler.java
@@ -2,6 +2,7 @@ package org.commcare.android.util;
 
 import android.content.Context;
 import android.content.Intent;
+import android.os.Bundle;
 
 import org.commcare.android.tasks.ExceptionReportTask;
 import org.commcare.dalvik.activities.CrashWarningActivity;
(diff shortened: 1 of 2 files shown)


ctsims:
Hey, @phillipm. This is super helpful to have, although it duplicates some funcitonality in the Form Entry activity that does more or less the same thing (shows the user an error with their form), that I think we're just missing because we aren't wrapping the right things with a try/catch. 

We generally expect to catch these in showNextView in the catch that runs CommCareActivity.createErrorDialog

phillipm:
As per @ctsims advice, going to try to accomplish this by wrapping offending expressions: see https://github.com/dimagi/commcare-odk/pull/713

PR #689, opened 2015-10-13
Fix for unnecessary spacing in custom case lists on new UI

amstone326:
IMPORTANT NOTE: For the material diff, look at the 1st commit only. The 2nd commit is just reformatting (somehow this file got missed out on our conversion from tabs to spaces) but it makes basically the whole file have a diff, so you can't see the meaningful changes from looking at the overall diff.

Explanation: For both REC and Parto (the 2 main apps using case tiles for their case lists), on the new UI there is currently an awkward amount of extra spacing between lines of content in each case tile. Since the spacing of the tile contents are dependent on the tile sizes, which are in turn dependent upon how many tiles the device tries to show per screen, the solution was to increase the default number of grid tiles that are shown per screen (from 5 to 7). Screenshots included below for how this changes the look of the case lists in both Parto and REC.

Also includes some linting to GridEntityView (mostly renaming of variables and removal of unnecessary class-level variables).

Got the OK from ICDS and REC that this looks ok for them.

### DIFF
--- a/app/src/org/commcare/android/view/GridEntityView.java
+++ b/app/src/org/commcare/android/view/GridEntityView.java
@@ -1,5 +1,5 @@
 /**
- * 
+ *
  */
 package org.commcare.android.view;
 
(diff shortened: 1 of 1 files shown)


amstone326:
REC case list before changes:
screenshot_2015-10-14-02-28-50

REC case list after changes:
screenshot_2015-10-14-02-33-00

amstone326:
Parto case list before changes:
screenshot_2015-10-14-02-36-53

Parto case list after changes:
screenshot_2015-10-14-02-34-34

amstone326:
@ctsims -- Got ok from all relevant projects, so good to go.

PR #690, opened 2015-10-13
Fixes for Parto persistent case tile

amstone326:
Fixes 2 related issues with the persistent case tile in the Parto app:

1)
-Problem: The app uses 3 colors for its case tiles, depending on the value a specific case property: either red, yellow, or transparent. The transparent option was causing problems in persistent tile form, because we have a default background color for persistent tiles of dark blue, so the transparent background was allowing that to show through, rather than appearing white as it does in the case list. Maryam and Joy (understandably) want the case tile to look consistent between the case list view and the detail view.
-Solution: Add an explicit "no-background" field in the custom xml for those case tiles in the Parto app, and then check for that value in the same way that "red-background" and "yellow-background" are being checked for, and set a white background in that case.

2)
-Problem: Because our default background color for persistent case tiles is dark blue, we set a white text color for those tiles. However, this looks odd against the red background, and is obviously illegible against the white and yellow backgrounds.
-Solution: Set a flag that indicates if an app is using a custom background color, and if it is, don't actually execute the contents of setTextColor() or setTitleTextColor()

..I realize both of these solutions are a bit hacky, but the whole framework for these case tiles is a little hacky and very fragile, so I wasn't really comfortable digging deeper to really refactor how things are done, given the timeframe we're aiming for.

### DIFF
--- a/app/src/org/commcare/android/view/TabbedDetailView.java
+++ b/app/src/org/commcare/android/view/TabbedDetailView.java
@@ -2,7 +2,6 @@ package org.commcare.android.view;
 
 import android.annotation.SuppressLint;
 import android.content.Context;
-import android.content.res.Resources;
 import android.content.res.TypedArray;
 import android.support.v4.app.FragmentActivity;
 import android.support.v4.view.ViewPager;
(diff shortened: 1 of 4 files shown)


app/src/org/commcare/android/view/TabbedDetailView.java
```diff
@@ -53,7 +52,6 @@ public TabbedDetailView(Context context, AttributeSet attrs) {
 
     private void loadViewConfig(Context context, AttributeSet attrs) {
         TypedArray typedArray = context.obtainStyledAttributes(attrs, R.styleable.TabbedDetailView);
-        Resources.Theme theme = context.getTheme();
```
amstone326: unused

app/src/org/commcare/dalvik/activities/EntityDetailActivity.java
```diff
@@ -46,7 +45,6 @@
     public static final String DETAIL_PERSISTENT_ID = "eda_persistent_id";
 
     Entity<TreeReference> entity;
-    EntityDetailAdapter adapter;
```
amstone326: unused

PR #692, opened 2015-10-13
Localize the login welcome message

phillipm:
Login screen was always showing "Welcome back! Please log in." even if the user wasn't using an English locale. Expose the string to the translation framework.

Also format XML and remove duplicate login button variable

http://manage.dimagi.com/default.asp?184140

### DIFF
--- a/app/assets/locales/messages_ccodk_default.txt
+++ b/app/assets/locales/messages_ccodk_default.txt
@@ -119,10 +119,11 @@ login.attempt.badcred=Username or password are incorrect. Please try again.
 
 login.username=Username
 login.password=Password
-login.welcome.message=Welcome back! Please log in.
 login.button=Log In
 login.sync=Synchronize with server
 login.bad.password=We couldn't find a user with this password. Please try another!
+login.welcome.single=Welcome back! Please log in.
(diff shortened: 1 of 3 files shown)


app/assets/locales/messages_ccodk_default.txt
```diff
@@ -119,6 +119,7 @@ login.attempt.badcred=Username or password are incorrect. Please try again.
 
 login.username=Username
 login.password=Password
+login.weclome.message=Welcome back! Please log in.
```
orangejenny: `weclome` => `welcome`

orangejenny:
looks good except typo

orangejenny:
Hey, I re-read this and realized there's already a `login.welcome.single` and `login.welcome.multiple`, which are both in HQ's translation's file, so 
a) the login button should probably use one of those and
b) that ticket should be passed to HQ's interrupt team to figure out why those strings aren't showing up in the download. My understanding is that everything in https://github.com/dimagi/commcare-translations/blob/master/messages_en-2.txt should be in the Excel download, but it looks like a bunch are missing  (login.welcome.*, seating.app, printjob.* ...)

PR #694, opened 2015-10-14
Clear the search bar when we launch from the home activity

wpride:
Fix for http://manage.dimagi.com/default.asp?183986

We were persisting the previous search in Entity/Form List activities both when re-launched from the home activity and when returning from form entry. This fix makes us only persist the search in the latter case


phillipm:
I just tried this on my phone and I don't think it works as expected.
When you back out of a case's form selection screen back to the case detail/selection screen, the detail/selection activity is (re)launched via the home screen, so it always has the clear flag you've created.
That said, I don't have any ideas of a way to tackle this.

phillipm:
(I was assuming this was ready for review; please mark your PRs so we can be sure)

wpride:
@ctsims @phillipm you both bring up good edge cases and I'm also not sure the right way to do this anymore. Is it safe to say that if we hit "Back" from EntitySelectActivty we *always* want to clear the search? If so the cleanest way to do this might be just to Override the back command of that activity to clear the cache

ctsims:
"Is it safe to say that if we hit "Back" from EntitySelectActivty we
*always* want to clear the search?"

Assuming it's a big lift to make the search specific to each module type,
so I think the clearest expectation would be that we only auto-search again
if the user has moved backward onto the search screen or if the session has
restarted/a new session has started. Pressing back from the search screen
would trigger that as long as the session restarting also did it.

-Clayton

wpride:
Let's go to voice on this at the standup tomorrow, still not clear to me what the desired behavior is

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -279,12 +279,10 @@ public void onClick(View v) {
                 adapter.registerDataSetObserver(this.mListStateObserver);
             }
         }
-        //cts: disabling for non-demo purposes
-        //tts = new TextToSpeech(this, this);
         restoreLastQueryString(TAG + "-" + KEY_LAST_QUERY_STRING);
 
         if (!isUsingActionBar()) {
-            searchbox.setText(lastQueryString);
+            searchbox.setText(getLastQueryString());
```
ctsims: This will still trigger the search behavior even if it is null, right? I think we likely need to simply not set the text if the last query is null

phillipm:
Addressed by https://github.com/dimagi/commcare-odk/pull/795

PR #695, opened 2015-10-14
We should scale the max image size before calculations

wpride:
Fix for http://manage.dimagi.com/default.asp?179788

Basically with half image resizing we were previously calculating the "Ideal" image size (based on the drawable and screen size), THEN rescaling to half that width/height. What we actually want is to say that only half the screen width/height is available to begin with and calculate the ideal image size from there.

### DIFF
--- a/app/src/org/odk/collect/android/views/ResizingImageView.java
+++ b/app/src/org/odk/collect/android/views/ResizingImageView.java
@@ -140,14 +140,14 @@ public class ResizingImageView extends ImageView {
     }
 
     private Pair<Integer,Integer> getWidthHeight(int widthMeasureSpec, int heightMeasureSpec, double imageScaleFactor) {
-        int maxWidth = mMaxWidth;
-        int maxHeight = mMaxHeight;
+        double maxWidth = mMaxWidth;
+        double maxHeight = mMaxHeight;
 
         if (MeasureSpec.getMode(widthMeasureSpec) == MeasureSpec.AT_MOST) {
-            maxWidth = Math.min(MeasureSpec.getSize(widthMeasureSpec), mMaxWidth);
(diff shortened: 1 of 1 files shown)


app/src/org/odk/collect/android/views/ResizingImageView.java
```diff
@@ -156,8 +156,8 @@ public void setFullScreen(){
         float dHeight = dipToPixels(getContext(), drawable.getIntrinsicHeight());
         float ratio = (dWidth) / dHeight;
 
-        int width = (int) Math.min(Math.max(dWidth, getSuggestedMinimumWidth()), maxWidth);
-        int height = (int) (width / ratio);
+        double width = (int) Math.min(Math.max(dWidth, getSuggestedMinimumWidth()), maxWidth);
```
amstone326: Something seems off here, this is of type double but you're casting to an int?

app/src/org/odk/collect/android/views/ResizingImageView.java
```diff
@@ -167,8 +167,7 @@ public void setFullScreen(){
             height = (int) (width / ratio);
         }
 
-        return new Pair<Integer, Integer>(new Double(width * imageScaleFactor).intValue(),
-                new Double(height * imageScaleFactor).intValue());
+        return new Pair<Integer, Integer>(new Double(width).intValue(), new Double(height).intValue());
```
amstone326: Hey, a little confused on what the ultimate goal of all the type conversions going on here is, but seems like between this line and the line above, something could definitely be simplified.

amstone326:
@wpride Left a question, and this should probably get a release note before merging.

wpride:
@amstone326 Should be clearer now. Basically we will end up with ints no matter what (since screen pixels is an int) but need to preserve some precision in the beginning for some of the scaling operations.

wpride:
Release note: Fix "Half Screen" Image resizing to scale to half the screen size, not half the image size

PR #697, opened 2015-10-14
Sms installs

wpride:


wpride:
@phillipm I had the same fix yesterday but I guess I forgot to push... yours is cleaner though

PR #698, opened 2015-10-15
Enabled the grid view menus when requested by the suite file

ctsims:
Depends on

https://github.com/dimagi/commcare/pull/177

### DIFF
--- a/app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
+++ b/app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
@@ -208,12 +208,24 @@ public class CommCareHomeActivity
 
     protected void enterRootModule() {
         Intent i;
-        if (DeveloperPreferences.isGridMenuEnabled()) {
+        if (useGridMenu("root")) {
             i = new Intent(getApplicationContext(), MenuGrid.class);
         } else {
             i = new Intent(getApplicationContext(), MenuList.class);
         }
         startActivityForResult(i, GET_COMMAND);
+    }
+
+    private boolean useGridMenu(String menuId) {
+        if(menuId == null) {
+            menuId = org.commcare.suite.model.Menu.ROOT_MENU_ID;
+        }
+        if(DeveloperPreferences.isGridMenuEnabled()) { return true; }
+        String commonDisplayStyle = platform.getMenuDisplayStyle(menuId);
+        if("grid".equals(commonDisplayStyle)) {
+            return true;
+        }
+        return false;
     }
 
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -208,7 +208,7 @@ private void goToFormArchive(boolean incomplete, FormRecord record) {
 
     protected void enterRootModule() {
         Intent i;
-        if (DeveloperPreferences.isGridMenuEnabled()) {
+        if (useGridMenu("root")) {
```
phillipm: `"root"` -> `Menu.ROOT_MENU_ID`?

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -216,6 +216,18 @@ protected void enterRootModule() {
...
+        if(menuId == null) {
+            menuId = org.commcare.suite.model.Menu.ROOT_MENU_ID;
+        }
+        if(DeveloperPreferences.isGridMenuEnabled()) { return true; }
+        String commonDisplayStyle = platform.getMenuDisplayStyle(menuId);
+        if("grid".equals(commonDisplayStyle)) {
+            return true;
+        }
+        return false;
+    }
```
phillipm: replace conditional with
`return "grid".equals(commonDisplayStyle);`
then replace `"grid"` with a constant defined in the appropriate menu/style related class

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -216,6 +216,18 @@ protected void enterRootModule() {
         startActivityForResult(i, GET_COMMAND);
     }
 
+    private boolean useGridMenu(String menuId) {
+        if(menuId == null) {
+            menuId = org.commcare.suite.model.Menu.ROOT_MENU_ID;
+        }
+        if(DeveloperPreferences.isGridMenuEnabled()) { return true; }
```
phillipm: `return true;` should be on its own line

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
ctsims: Refactored for static conditional. I left the static MENU_STYLE_GRID in this file because the logic for what styles are available/used is platform-dependent and no other object has the right level of context (It would be weird to put it in /commcare/ because commcare doesn't actually support the grid).

PR #699, opened 2015-10-15
Fix for case tiles when there are computed datums

ctsims:
Fixes issues with persisted case tiles not showing


czue:
did you just make this PR between sentences as the DSI field/dev session?

#beast

app/src/org/commcare/android/framework/BreadcrumbBarFragment.java
```diff
@@ -300,7 +300,11 @@ public void run() {
             StackFrameStep step = v.elementAt(i);
 
             if(SessionFrame.STATE_DATUM_VAL.equals(step.getType())) {
-                stepToFrame = step;
+                //Only add steps which have a tile.
+                SessionDatum d = asw.getSession().findDatumDefinition(step.getId());
+                if(d.getPersistentDetail() != null) {
```
phillipm: `d` can be null, might want to add a null check in there

PR #702, opened 2015-10-16
Improve inflation of display images

amstone326:
Summary: Technique for predictable inflation of user-provided display images, applied across CommCare wherever it is applicable (Qualification: Not actually everywhere yet, there are probably places I missed. @ctsims, Would like to do a pass with you to determine if there is anywhere else in the app that should be using this). 

Details: The key method is getBitmapScaledForNativeDensity() in MediaUtil, which implements the following essential logic (I'm simplifying slightly in words, the code makes the nuances clear):
-Users can set a per-app target density, which indicates the screen density they are designing their images for
-We then apply a scaling technique to up- or down-scale images if they are being shown on a screen whose density is different from the target density
-We also have the option of providing bounding dimensions for the inflation -- if provided, the image will never be scaled up to larger than those dimens. This is useful for not wasting memory on inflating/keeping around large bitmaps in instances in which we know that an image is going into a container of a certain size, which is frequent (images in the case list, icons in GridMediaView, etc.). If explicit bounding dimens are not provided, the screen's width and height are used.

Other Important Notes:
-This is currently disabled by default, and can be enabled in Developer Settings. You can then also set the desired target density in the Developer Settings. Did this to make it easy for ICDS and other projects to play around with different target densities combined with different original image sizes at will.
-I have set it up such that this strategy and ResizingImageView cannot be applied in tandem -- If this strategy is turned off, ResizingImageView will still be applied, assuming the app has it turned on. If this strategy is enabled, ResizingImageView will not get used no matter what.

One less important note: There are currently 4 calls to inflateDisplayImage() where we are NOT providing explicit bounding dimens (and thus defaulting to the screen dimens). In 2 of them I think this is fine/correct because there doesn't seem to be any clear bounding box. But in the other 2, I think there probably is a clear boundary, but I couldn't figure out how exactly to get at them from the code available. Those 2 places are:
1. MenuAdapter.getView(), which is where icons for menu lists are inflated (always going to be small)
2. EntityDetailAdapter.setParams(), which is where images in a basic case detail are inflated (always going to be ~somewhat small I think?)
Both of these seem like they could benefit from restricting the size of the image inflation based on container. If anyone is doing a really thorough review of this and has thoughts on where to get those values in these 2 instances, let me know!


app/src/org/commcare/android/util/MediaUtil.java
```diff
@@ -1,82 +1,260 @@
...
+        return customDpiScaleFactor * proportionalAdjustmentFactor;
     }
 
+
     /**
-     * Pass in a string representing either a GeoPont or an address and get back a valid
+     * Pass in a string representing either a GeoPoint or an address and get back a valid
      * GeoURI that can be passed as an intent argument 
      */
     public static String getGeoIntentURI(String rawInput){
```
amstone326: This feels like the wrong place for this method, but didn't have ideas for a better util class to move it to. If you have any thoughts, let me know.

app/src/org/odk/collect/android/utilities/FileUtils.java
```diff
@@ -308,46 +282,6 @@ private static Bitmap getScaledBitmap(Bitmap originalBitmap, int maxDimen,
         }
     }
 
-
-    public static Bitmap getBitmapScaledToDisplay(File f, int screenHeight, int screenWidth) {
```
amstone326: moved these to MediaUtil

amstone326:
One other note: There are currently 4 calls to inflateDisplayImage() where we are NOT providing explicit bounding dimens (and thus defaulting to the screen dimens). In 2 of them I think this is fine/correct because there doesn't seem to be any clear bounding box. But in the other 2, I think there probably is a clear boundary, but I couldn't figure out how exactly to get at them from the code available. Those 2 places are:
1. MenuAdapter.getView(), which is where icons for menu lists are inflated (always going to be small)
2. EntityDetailAdapter.setParams(), which is where images in a basic case detail are inflated (always going to be ~somewhat small I think?)

Both of these seem like they could benefit from restricting the size of the image inflation based on container. If anyone is doing a really thorough review of this and has thoughts on where to get those values in these 2 instances, let me know!

app/src/org/commcare/android/util/MediaUtil.java
orangejenny: There's a GeoUtils class in ./app/src/org/odk/collect/android/utilities, maybe there?

(I haven't read this PR, am saying this solely based on the method name)

amstone326:
Discovered a bug while writing tests, so going to fix that and finish the tests and then will mark as ready again.

app/src/org/commcare/android/util/MediaUtil.java
amstone326: Sweet, thanks!

app/src/org/commcare/android/util/MediaUtil.java
```diff
@@ -1,82 +1,274 @@
...
+     */
+    public static Bitmap inflateDisplayImage(Context context, String jrUri,
+                                             int boundingWidth, int boundingHeight) {
+        if (jrUri == null || jrUri.equals("")) {
             return null;
         }
-        
-        BitmapFactory.Options options = new BitmapFactory.Options();
+        try {
+            String imageFilename = ReferenceManager._().DeriveReference(jrUri).getLocalURI();
```
ctsims: This used to work with jr refs that didn't point to files, we should probably have a fallback for what to do here when this doesn't point to a file

app/src/org/commcare/android/util/MediaUtil.java
```diff
@@ -1,82 +1,274 @@
...
+                                                   int boundingHeight, int boundingWidth) {
+
+        Pair<Integer, Integer> dimensImposedByContainer = getProportionalDimensForContainer(
+                originalHeight, originalWidth, boundingHeight, boundingWidth);
+        int newWidth = Math.min(dimensImposedByContainer.first, calcWidth);
+        int newHeight = Math.min(dimensImposedByContainer.second, calcHeight);
+
+        try {
+            BitmapFactory.Options o = new BitmapFactory.Options();
+            o.inScaled = false;
```
ctsims: if we know the file boundaries here, we should be able to scale down using this, which will also improve our performance when loading these images.

app/src/org/commcare/android/util/MediaUtil.java
```diff
@@ -1,82 +1,274 @@
...
+     * exceeding either boundingWidth or boundingHeight
+     */
+    private static Bitmap getBitmapScaledDownExact(String imageFilepath,
+                                                   int originalHeight, int originalWidth,
+                                                   int calcHeight, int calcWidth,
+                                                   int boundingHeight, int boundingWidth) {
+
+        Pair<Integer, Integer> dimensImposedByContainer = getProportionalDimensForContainer(
+                originalHeight, originalWidth, boundingHeight, boundingWidth);
+        int newWidth = Math.min(dimensImposedByContainer.first, calcWidth);
```
ctsims: I'm confused as to how we can choose these numbers one at a time. If we used dimensImposedByContainer.first, and calcHeight, we'd end up with an image in the wrong aspect ratio, right?

app/src/org/commcare/android/util/MediaUtil.java
```diff
@@ -1,82 +1,274 @@
...
+    }
+
+    /**
+     * @return A bitmap representation of the given image file, scaled up as close as possible to
+     * desiredWidth and desiredHeight, without exceeding either boundingHeight or boundingWidth
+     */
+    private static Bitmap attemptBoundedScaleUp(String imageFilepath, int desiredHeight, int desiredWidth,
+                                         int boundingHeight, int boundingWidth) {
+        if (boundingHeight < desiredHeight || boundingWidth < desiredWidth) {
+            float heightScale = ((float)boundingHeight) / desiredHeight;
```
ctsims: This pattern comes up a few times. Am wondering whether it would improve clarity to have it be a submethod

app/src/org/commcare/android/util/MediaUtil.java
amstone326: Since the values of both calcWidth/calcHeight, and dimensImposedByContainer.first/second preserve the aspect ratio, it's not possible that you would end up with dimens from different pairs getting mixed.

app/src/org/commcare/android/util/MediaUtil.java
amstone326: you mean just dividing 2 ints to get a float?

app/src/org/commcare/android/util/MediaUtil.java
```diff
@@ -1,82 +1,274 @@
...
+        return customDpiScaleFactor * proportionalAdjustmentFactor;
+    }
+
+
+    /**
+     * @return A bitmap representation of the given image file, potentially adjusted from the
+     * image's original size such that its width is no larger than containerWidth, and its height
+     * is no larger than containerHeight
+     */
+    private static Bitmap getBitmapScaledToContainer(String imageFilepath, int containerHeight,
```
ctsims: This only scales down in the BitmapFactory scale proportions, right? That seems a bit problematic. Feels like we should do one of:

1) Scale down to the _last_ scale factor that is still larger than the container (IE: the bounding box is 100x100, the image is 480x480, scale x 4 gives us 120x120, which then in an imageview is displayed as 100x100. If we scaled down x5 (the current behavior, since round(4.8) -> 5) we get 60x60, which then wouldn't fit in the container

app/src/org/commcare/android/view/EntityView.java
```diff
@@ -277,44 +273,24 @@ private void redrawImageLayoutsInQueue() {
      */
     public void setupImageLayout(View layout, final String source) {
         ImageView iv = (ImageView) layout;
-        Bitmap b;
-        if (!source.equals("")) {
```
ctsims: Wow, amazing how much cleaner this is now, nice!

app/src/org/commcare/android/view/GridEntityView.java
```diff
@@ -361,12 +366,13 @@ private View getView(Context context, String multimediaType, String horzAlign, S
                         break;
                 }
                 retVal.setPadding(CELL_PADDING_HORIZONTAL, CELL_PADDING_VERTICAL, CELL_PADDING_HORIZONTAL, CELL_PADDING_VERTICAL);
-                // image loading is handled asyncronously by the TCImageLoader class to allow smooth scrolling
                 if (rowData != null && !rowData.equals("")) {
                     if (mImageLoader != null) {
-                        mImageLoader.display(rowData, ((ImageView) retVal), R.drawable.info_bubble);
+                        mImageLoader.display(rowData, ((ImageView) retVal), R.drawable.info_bubble,
```
ctsims: I don't think we wanna scale this, if possible. The drawable bitmaps are super efficient at getting themselves in/out of memory, and this'll result in there being a new copy of it

app/src/org/odk/collect/android/views/media/MediaLayout.java
```diff
@@ -204,39 +203,26 @@ else if (qrCodeContent != null ) {
...
-
-
-                        int screenWidth = display.getWidth();
-                        int screenHeight = display.getHeight();
-                        b =
-                            FileUtils
-                                    .getBitmapScaledToDisplay(imageFile, screenHeight, screenWidth);
+                        b = MediaUtil.inflateDisplayImage(getContext(), imageURI, maxBounds[0],
+                                maxBounds[1]);
                     } catch (OutOfMemoryError e) {
```
ctsims: this isn't possible anymore, right? Since the OOM is now caught inside of the Utility? Would clean up the code a bit to get rid of it.

ctsims:
@amstone326 This is looking good after hte changes outlined

app/src/org/commcare/android/util/MediaUtil.java
```diff
@@ -1,82 +1,297 @@
...
+        int newWidth, newHeight;
+        if (targetHeight == -1 || targetWidth == -1) {
+            newWidth = dimensImposedByContainer.first;
+            newHeight = dimensImposedByContainer.second;
+        } else {
+            newWidth = Math.min(dimensImposedByContainer.first, targetWidth);
+            newHeight = Math.min(dimensImposedByContainer.second, targetHeight);
+        }
+
+        int approximateScaleFactor = originalWidth / newWidth;
```
ctsims: Is this always > 1? If not it'll make it scale to 0, right?

amstone326: @ctsims Added a check

app/src/org/commcare/android/util/MediaUtil.java
```diff
@@ -1,82 +1,297 @@
...
+        try {
+            BitmapFactory.Options o = new BitmapFactory.Options();
+            o.inSampleSize = approximateScaleFactor;
+            // Decode the bitmap with the largest integer scale down factor that will not make it
+            // smaller than the final desired size
+            Bitmap originalBitmap = BitmapFactory.decodeFile(imageFilepath, o);
+            // From that, generate a bitmap scaled to the exact right dimensions
+            return Bitmap.createScaledBitmap(originalBitmap, newWidth, newHeight, false);
```
ctsims: So when we aren't scaling for density reasons (where the exact final dimension isn't required to be specific), I'm worried about the runtime implications of this scaling. If we load an image and don't density-adjust it and it's 20% bigger than it needs to be it seems like this is gonna really slow things down, right? I can't imagine it's a fast operation. 

(Just trying to consider implications for people who don't have the new feature enabled)

PR #706, opened 2015-10-17
Don't use deprecated Fragment.onAttach(Activity)

phillipm:
- Rename `PreferenceActivity` to `FormEntryPreferences`
 - Update fragments to not use deprecated `onAttach(Activity activity)`.

I was going to  update `CommCarePreferences`, `DeveloperPreferences`, and `PreferencesActivity` to stop extending the deprecated `PreferenceActivity`, and instead use `PreferenceFragment`. Turns out `PreferenceFragment` is only API 11 and greater, so I'll leave this fix for another time.

### DIFF
--- a/app/src/org/commcare/android/fragments/SetupEnterURLFragment.java
+++ b/app/src/org/commcare/android/fragments/SetupEnterURLFragment.java
@@ -8,7 +8,6 @@ import android.util.Log;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
-import android.view.WindowManager;
 import android.view.inputmethod.InputMethodManager;
 import android.widget.AdapterView;
 import android.widget.Button;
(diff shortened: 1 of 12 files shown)


app/src/org/commcare/android/fragments/SetupEnterURLFragment.java
```diff
@@ -133,14 +133,17 @@ public void onResume() {
...
-        activity.getWindow().setSoftInputMode(WindowManager.LayoutParams.SOFT_INPUT_STATE_ALWAYS_VISIBLE);
+    public void onAttach(Context context) {
+        super.onAttach(context);
 
-        super.onAttach(activity);
-        if (!(activity instanceof URLInstaller)) {
-            throw new ClassCastException(activity + " must implemement " + interfaceName);
+        if (context instanceof Activity) {
+            ((Activity)context).getWindow().setSoftInputMode(WindowManager.LayoutParams.SOFT_INPUT_STATE_ALWAYS_VISIBLE);
+        }
```
phillipm: Sorta feels strange having an `instanceof Activity` check. Might be possible to set this input mode somewhere else, or by other means, in a more natural way?

ctsims: Yeah, agree it's kind of unsatisfying. Probably good tech debt to cover later?

PR #708, opened 2015-10-19
Case list icon resize

wpride:
So adding this property will scale the image to take up the entire space available to the image (the red in the second picture) without messing with the aspect ratio. I *think* this is more often to be what we want, especially if HQ will continue adding the "13%" default space to icons since this will end up being tiny on small screens.

Long term, I think better HQ case tile support is the correct-er solution here

@orangejenny 

bigger

screen shot 2015-10-19 at 5 37 27 pm

### DIFF
--- a/app/res/layout/entity_item_image.xml
+++ b/app/res/layout/entity_item_image.xml
@@ -1,7 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <ImageView xmlns:android="http://schemas.android.com/apk/res/android"
     android:layout_width="match_parent"
-    android:layout_height="match_parent" style="@style/EntityItemTextView" android:textAppearance="@style/EntityItemText">
+    android:layout_height="match_parent" style="@style/EntityItemTextView"
+    android:textAppearance="@style/EntityItemText"
+    android:scaleType="centerCrop">
     
 
 </ImageView>
--- a/app/src/org/commcare/android/view/EntityView.java
+++ b/app/src/org/commcare/android/view/EntityView.java
@@ -282,6 +282,7 @@ public class EntityView extends LinearLayout {
             try {
                 if (onMeasureCalled) {
                     int columnWidthInPixels = layout.getLayoutParams().width;
+                    System.out.println("columnWidthInPixels "  + columnWidthInPixels);
                     b = FileUtils.getScaledBitmap(
                             ReferenceManager._().DeriveReference(source).getStream(),
                             columnWidthInPixels, true);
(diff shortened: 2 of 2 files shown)


amsagoff:
:+1:

amstone326:
Hey @wpride -- I'm doing a bunch of work right now to make the way we inflate images standardized across CommCare, and predictable for users in terms of what size their images will be, so it seems at odds with that goal to be adding custom resizing rules in specific places. (It also looks to me personally like the after image here has the icons too big, but that's something more specific we can debate). Either way I'm gonna throw a don't pull tag on this for now until we talk tonight and can sync about what the right strategy for this is.

wpride:
this can be pulled now that @amstone326 confirmed its unrelated to her refactor

app/res/layout/entity_item_image.xml
```diff
@@ -1,7 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <ImageView xmlns:android="http://schemas.android.com/apk/res/android"
     android:layout_width="match_parent"
-    android:layout_height="match_parent" style="@style/EntityItemTextView" android:textAppearance="@style/EntityItemText">
+    android:layout_height="match_parent" style="@style/EntityItemTextView"
+    android:textAppearance="@style/EntityItemText"
+    android:scaleType="centerCrop">
```
ctsims: @wpride The behavior here is "make the image fill the entire space", right? That's not correct and isn't what we want to be doing here. People won't necessarily expect that their icons scale up to fill the entire view, and it will distort smaller images that don't fit in the space.

ctsims:
@orangejenny @wpride: I don't agree with the strategy here. We don't generally distort peoples images by forcing them to scale to fill a space entirely, and I don't think we should.

wpride:
@ctsims @amsagoff this feels like something we should discuss in roadmap... I can see what you're saying (people might want a smaller image than max) but the current situation feels pretty arbitrary, especially given how pixel density can futz with things. In general it feels like people want icons to take up some consistent portion of screen size across devices and layouts - I'd prefer a system where the icon scales to max and they can adjust the width allotment personally

ctsims:
Can discuss in roadmap. Aliza's recent PR fixes the ambiguity and makes the
image pop-in entirely deterministic. If people want things to take up the
full space that way it is easy for them to do so, but especially with flags
and stuff people often want some amount of padding/whitespace and it looks
really disproportionate for it to take up the full column height when
nothing else in the row does.

amstone326:
@wpride, @ctsims Agreed we should discuss in the same room. I'm a little unclear on what the original behavior of case list image sizing was (before this PR), I know there was something about HQ automatically resizing to 13% of the screen? Seems like that might be relevant / something we want to change. But agreed with Clayton that this isn't the behavior we want. In general the direction we want to be moving with images is to do less artificial resizing, and base everything on the original image size, assuming that the users intended it to be the size they made it.

PR #711, opened 2015-10-19
Sd logging

wpride:
Just some added logging for SD Card Dump and Wifi Direct so we can track/verify if we're losing files anywhere

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -57,6 +57,8 @@
     <uses-feature
         android:name="android.hardware.location.network"
         android:required="false" />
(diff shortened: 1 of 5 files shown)


app/AndroidManifest.xml
```diff
@@ -57,6 +57,8 @@
     <uses-feature
         android:name="android.hardware.location.network"
         android:required="false" />
+    <uses-feature android:name="android.hardware.usb.host"
+        android:required="false"/>
```
phillipm: Can you elaborate why this is needed?

wpride:
@phillipm is not, was from another PR, my b

PR #712, opened 2015-10-20
Sms install fix

wpride:
In response to http://manage.dimagi.com/default.asp?184633. Basically if we never found a text and started the install task we never notified the user on the outcome.

### DIFF
--- a/app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
+++ b/app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
@@ -608,10 +608,12 @@ public class CommCareSetupActivity extends CommCareActivity<CommCareSetupActivit
         }
         int messageIterationCount = 0;
         try {
+            boolean attemptedInstall = false;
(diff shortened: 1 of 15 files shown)


app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -664,6 +666,11 @@ protected void deliverError(CommCareSetupActivity receiver, Exception e) {
                     break;
                 }
             }
+            // attemptedInstall will only be true if we found no texts with the SMS_INSTALL_KEY_STRING tag
+            // if we found one, notification will be handle by the task receiver
+            if(!attemptedInstall) {
+                Toast.makeText(this, Localization.get("menu.sms.not.found"), Toast.LENGTH_LONG).show();
+            }
```
phillipm: Probably only want to show this if `installTriggeredManually` is `true`. 
So `if (!attemptedInstall && installTriggeredManually)`

PR #714, opened 2015-10-21
Optional Report Error Button on Home Menu

ctsims:
Potential fix for the awkwardness of only having 3 buttons
device-2015-10-21-122536


phillipm:
Probably worth running by @amsagoff and @biyeun.
In my opinion, we don't want our app to be crashing enough to warrant a large "report an issue" button on the home screen. Maybe we could change the logout button to be a rectangle at the bottom of the screen similar to https://github.com/dimagi/commcare-odk/pull/599

ctsims:
Hey,

It's basically just there temporarily for Matt et. All to demo the app. The
real end-users don't care about the 3-button layout, really.

-Clayton

phillipm:
Sounds good. They are going to be on a master build, right? So we could merge this after 2.24 is cut, and then have time to come up with a more robust solution.

ctsims:
Yeah, that seems true.

-Clayton

biyeun:
OMG. Did you make that button @phillipm? :heart_eyes:

ctsims:
@biyeun That's a clayton original thank you very much :p

amsagoff:
So just to clarify, this is *not* a product feature? @ctsims ?

ctsims:
Yeah, it's hidden and mostly for people to demo with.

I still think it'd be nice for 2.24 but no strong feelings.

biyeun:
@ctsims are you a secret designer o_O

wpride:
@ctsims good to merge after conflicts resolved

phillipm:
Bump
Should we close this or try to get it into 2.25?

phillipm:
We should use a vector image for this, to prevent high memory usage from drawing the button, just like the other home screen buttons. I can take care of that, just don't merge in the meantime

PR #715, opened 2015-10-21
Android Support Library Update

ctsims:
Update to Android support 23.1.0 to fix issue with tabs not displaying. The 23.0.0/1 versions have a bug with padding that breaks text display.


phillipm:
https://github.com/dimagi/commcare-odk/pull/707

PR #717, opened 2015-10-21
Propogate Subnode Variables to Children

ctsims:
When using subnode entity screens, this lets the parent screen define variables in the original context that can then be read by the subnode detail screen when it is displaying data. Useful if the two have unrelated nodesets

### DIFF
--- a/app/src/org/commcare/android/framework/EntitySubnodeDetailFragment.java
+++ b/app/src/org/commcare/android/framework/EntitySubnodeDetailFragment.java
@@ -71,6 +71,14 @@ public class EntitySubnodeDetailFragment extends EntityDetailFragment implements
         return rootView;
     }
 
+    /**
+     * Creates an evaluation context which is preloaded with all of the variables and context from
+     * the parent detail definition.
+     *
+     * @param childReference The qualified reference for the nodeset in the parent detail
+     * @return An evaluation context ready to be used as the base of the subnode detail, including
+     * any variable definitions included by the parent.
+     */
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/android/framework/EntitySubnodeDetailFragment.java
```diff
@@ -69,6 +71,17 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
         return rootView;
     }
 
+    private EvaluationContext prepareEvaluationContext(TreeReference childReference) {
+        EvaluationContext sessionContext = asw.getEvaluationContext();
+        EvaluationContext parentDetailContext = new EvaluationContext(sessionContext, childReference);
+        getParentDetail().populateEvaluationContextVariables(parentDetailContext);
```
phillipm: This can through an `XPathTypeMismatchException`, right? Is there a good way to catch this and show it to the user when we aren't in something that extends `CommCareActivity`?

app/src/org/commcare/android/framework/EntitySubnodeDetailFragment.java
```diff
@@ -69,6 +71,17 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
         return rootView;
     }
 
+    private EvaluationContext prepareEvaluationContext(TreeReference childReference) {
```
phillipm: I don't have the background in this area of the code to understand the PR's description, but I imagine adding a comment here that says essentially the same thing would be helpful for readers that do.

PR #720, opened 2015-10-21
Surface refactoring of entity detail and select activities

phillipm:
Note that I started taking out the text to speech stuff, that seems to be unused


ctsims:
Hey,

Now that we have developer feature flags it could be nice to just feature-flag the TTS stuff. I think the India team might be interested in piloting.

phillipm:
recreated PR, leaving in TTS code: https://github.com/dimagi/commcare-odk/pull/778

PR #725, opened 2015-10-22
Add task to refresh test results

phillipm:


phillipm:
Reverted this. In retrospect not sure if it is needed due to the android plugin for gradle. Will look into it further

PR #726, opened 2015-10-22
Don't try to do fragment transaction when paused

phillipm:
Fragment transactions throw `java.lang.IllegalStateException: Can not perform this action after onSaveInstanceState` exceptions if you try to perform them when the activity is paused. This PR prevents that from happening with the CommCareSetupActivity.

[Addresses this ACRA crash report](https://commcare-acralyzer.cloudant.com/acralyzer/_design/acralyzer/index.html#/reports-browser/commcare-odk/bug/1ea83dfd44b3bfaa79e788a9bae25560)

### DIFF
--- a/app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
+++ b/app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
@@ -193,7 +193,6 @@ public class CommCareSetupActivity extends CommCareActivity<CommCareSetupActivit
                         "startAllowed is: " + startAllowed + " "
         );
 
-        uiStateScreenTransition();
         performSMSInstall(false);
     }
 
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -283,12 +290,6 @@ private Fragment restoreInstallSetupFragment() {
         return fragment;
     }
 
-    @Override
-    protected void onStart() {
-        super.onStart();
-
-        uiStateScreenTransition();
```
phillipm: Moved this call to `onResume` which should pretty much be the same thing

phillipm:
Going to target 2.24: https://github.com/dimagi/commcare-odk/pull/740

PR #730, opened 2015-10-23
Fix where we were never dismissing the alert dialog

wpride:


wpride:
http://manage.dimagi.com/default.asp?184706#1033818

phillipm:
We'll be doing weekly merges of 2.24 back into master, so there isn't a need to create a master PR for 2.24 changes

PR #733, opened 2015-10-23
Save form load error and reshow dialog on orientation change

phillipm:
Prevent app from crashing when form load error dialog is shown and the app is rotated.

Fixed a bug where the `CommCareHomeActivity.platform` field wasn't being setup correctly between orientation changes. The solution was to not have a `platform` class field, but rather lookup the platform every time we need it. Not sure why `platform` was a class field in the first place, so worth double checking I didn't break anything/make anything slower. @ctsims, thoughts?

http://manage.dimagi.com/default.asp?184630

### DIFF
--- a/app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
+++ b/app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
@@ -154,8 +154,6 @@ public class CommCareHomeActivity
 
     private int mDeveloperModeClicks = 0;
 
-    private AndroidCommCarePlatform platform;
-
     private HomeActivityUIController uiController;
     private SessionNavigator sessionNavigator;
 
@@ -226,6 +224,8 @@ public class CommCareHomeActivity
         if(DeveloperPreferences.isGridMenuEnabled()) {
             return true;
         }
+
+        AndroidCommCarePlatform platform = CommCareApplication._().getCommCarePlatform();
         String commonDisplayStyle = platform.getMenuDisplayStyle(menuId);
         return MENU_STYLE_GRID.equals(commonDisplayStyle);
     }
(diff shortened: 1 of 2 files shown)


ctsims:
@phillipm This looks fine to me, although I thought part of the point of fragment dialogs was that they automatically re-instantiated themselves if the screen rotated, so it's surprising to me you're needing to pop it back up. Are we just missing a hook that properly puts the fragments back in?

phillipm:
@ctsims I think we only use fragment dialogs for the `CustomProgressDialog` and no where else. This approach sounds reasonable: http://stackoverflow.com/a/15729764. Might even need to implement it to fix http://manage.dimagi.com/default.asp?184706#BugEvent.1033877

amstone326:
@phillipm- It looks like this PR is doing something very custom for this one instance to re-show the alert dialog after orientation. Given that you started a generic fix for making alert dialogs persist on orientation across the board, doesn't it make sense to wait for that and scrap this?

amstone326:
(adding don't pull label until we can chat about this really quickly)

phillipm:
@amstone326 I agree. In fact, I've already migrated this fix to use a persistent dialog in https://github.com/dimagi/commcare-odk/pull/734. That is a bigger refactor though so I figured we should have it target 2.25. I'm happy to make a note of this quick fix and remove it in master after we do our first merge of the 2.24 branch into master.

amstone326:
@phillipm Got it, makes sense. I was actually thinking we would aim for the bigger rotation fix to be added to 2.24 since it's a pretty major bug with the new dialogs. Does that seem unrealistic?

phillipm:
Decided to get the persistent dialog fix into 2.24

PR #734, opened 2015-10-23
Make alert dialog pop-ups persist across orientation change

phillipm:
Continues the work of https://github.com/dimagi/commcare-odk/pull/735 but applying it to all instances of `AlertDialogFactory`

### DIFF
--- a/app/src/org/commcare/dalvik/dialogs/AlertDialogFactory.java
+++ b/app/src/org/commcare/dalvik/dialogs/AlertDialogFactory.java
@@ -1,7 +1,7 @@
 package org.commcare.dalvik.dialogs;
 
-import android.app.Activity;
 import android.app.AlertDialog;
+import android.content.Context;
 import android.content.DialogInterface;
 import android.view.LayoutInflater;
 import android.view.View;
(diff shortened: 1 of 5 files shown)


app/src/org/commcare/dalvik/dialogs/AlertDialogFactory.java
```diff
@@ -81,12 +82,11 @@ public void onClick(DialogInterface dialog, int which) {
     }
 
     public void showDialog() {
-        dialog.setView(this.view);
```
amstone326: @phillipm There's a reason this call needs to be here -- there are methods that can be called on an instance of AlertDialogFactory in between its construction and calling showDialog() that update the view.

app/src/org/commcare/dalvik/dialogs/AlertDialogFragment.java
```diff
@@ -0,0 +1,72 @@
...
+            factory.setNeutralButton(args.getString(NEUTRAL_MESSAGE_KEY),
+                    listener);
+        }
+
+        return factory.getDialog();
+    }
+
+    @Override
+    public void onDestroyView() {
+        // Ohh, you know, just a 5 year old Android bug ol' G hasn't fixed yet
```
amstone326: This makes me feel better about our backlog

phillipm:
@amstone326 has a more comprehensive approach that targets 2.24 https://github.com/dimagi/commcare-odk/pull/738

PR #735, opened 2015-10-23
Make "Continue Form" alert dialog pop-ups persist across orientation change

phillipm:
The "Continue Form" alert dialog shouldn't disappear when the screen orientation changes.

image

http://manage.dimagi.com/default.asp?184706#BugEvent.1033877

### DIFF
--- a/app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
+++ b/app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
@@ -53,6 +53,7 @@ import org.commcare.dalvik.application.AndroidShortcuts;
 import org.commcare.dalvik.application.CommCareApp;
 import org.commcare.dalvik.application.CommCareApplication;
 import org.commcare.dalvik.dialogs.AlertDialogFactory;
+import org.commcare.dalvik.dialogs.AlertDialogFragment;
 import org.commcare.dalvik.dialogs.CustomProgressDialog;
 import org.commcare.dalvik.odk.provider.FormsProviderAPI;
 import org.commcare.dalvik.odk.provider.InstanceProviderAPI;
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -1030,8 +1032,11 @@ protected void deliverError(CommCareHomeActivity receiver, Exception e) {
     }
 
     @Override
-    protected void onResume() {
-        super.onResume();
+    protected void onResumeFragments() {
```
phillipm: Supposedly `FragmentActivity` should override `onResumeFragments` not `onResume` whenever possible.

app/src/org/commcare/dalvik/dialogs/AlertDialogFactory.java
```diff
@@ -30,6 +30,7 @@ public AlertDialogFactory(Activity context, String title, String msg) {
         messageView.setText(msg);
 
         this.dialog = builder.create();
+        dialog.setView(view);
```
phillipm: @amstone326 is there a reason this was being called in `showDialog` and not here? Seems like it should happen here in case someone tries to pull `getDialog().show()` instead of `showDialog()`.

amstone326: @phillipm Oops just made a comment about this on the other PR before I saw this. Yes, it needs to be here because there are methods that can be called on an AlertDialogFactory in between it's construction and calling showDialog() that change the view. Seems like the solution to the issue you're talking about is to call setView() in both getDialog() and showDialog()

phillipm:
Fix encapsulated in https://github.com/dimagi/commcare-odk/pull/738

PR #736, opened 2015-10-23
Make alert dialog scroll long messages

orangejenny:
http://manage.dimagi.com/default.asp?185212

Put the buttons inside the scrollable area because I had issues with them disappearing when the ScrollView only contained the TextView. Open to looking at it more if anyone thinks that's important, but i don't think it's a big deal.


amstone326:
@orangejenny Made a direct change to the branch so that the buttons are outside the scrollview. Good to merge this as long as you're happy with the change.

PR #738, opened 2015-10-26
Make alert dialogs persist on rotation V.2

amstone326:
This PR makes it so that all alert dialogs spawned by any CommCareActivity will persist on orientation change. Alert Dialogs spawned by non-CC activities will still disappear on rotation, but this seems ok for now-- none of those instances are things that will break the app in any way or cause confusion. 

This is a partial (but not complete) solution to http://manage.dimagi.com/default.asp?184630. @phillipm is going to make a PR to address the rest of that ticket.

### DIFF
--- a/app/src/org/commcare/android/framework/CommCareActivity.java
+++ b/app/src/org/commcare/android/framework/CommCareActivity.java
@@ -249,14 +249,7 @@ public abstract class CommCareActivity<R> extends FragmentActivity
     @Override
     @TargetApi(11)
     protected void onResumeFragments() {
-        showPendingDialog();
-    }
-
-    private void showPendingDialog() {
-        if (dialogToShowOnResume != null) {
-            dialogToShowOnResume.show(getSupportFragmentManager(), KEY_ALERT_DIALOG_FRAG);
-            dialogToShowOnResume = null;
-        }
+        showPendingAlertDialog();
     }
 
     protected View getBannerHost() {
--- a/app/src/org/commcare/android/framework/CommCareActivity.java
+++ b/app/src/org/commcare/android/framework/CommCareActivity.java
@@ -247,8 +247,9 @@ public abstract class CommCareActivity<R> extends FragmentActivity
     }
 
     @Override
-    @TargetApi(11)
     protected void onResumeFragments() {
+        super.onResumeFragments();
+
         showPendingAlertDialog();
     }
 
@@ -525,7 +526,7 @@ public abstract class CommCareActivity<R> extends FragmentActivity
      * @param activity   Activity to which to attach the dialog.
      * @param shouldExit If true, cancel activity when user exits dialog.
      */
-    public static void createErrorDialog(final Activity activity, String errorMsg,
+    public static void createErrorDialog(final CommCareActivity activity, String errorMsg,
                                          final boolean shouldExit) {
         String title = StringUtils.getStringRobust(activity, org.commcare.dalvik.R.string.error_occured);
         AlertDialogFactory factory = new AlertDialogFactory(activity, title, errorMsg);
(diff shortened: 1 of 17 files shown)


amstone326:
@phillipm This is ready for review

app/src/org/commcare/android/framework/CommCareActivity.java
```diff
@@ -243,6 +246,12 @@ protected void onResume() {
         AudioController.INSTANCE.playPreviousAudio();
     }
 
+    @Override
+    @TargetApi(11)
```
phillipm: Don't think this annotation is needed

app/src/org/commcare/android/framework/CommCareActivity.java
```diff
@@ -243,6 +246,12 @@ protected void onResume() {
         AudioController.INSTANCE.playPreviousAudio();
     }
 
+    @Override
+    @TargetApi(11)
+    protected void onResumeFragments() {
+        showPendingAlertDialog();
```
phillipm: Missing call to `super` implementation

app/src/org/commcare/android/framework/CommCareActivity.java
```diff
@@ -528,14 +545,18 @@ public void onClick(DialogInterface dialog, int i) {
         };
         CharSequence buttonDisplayText = StringUtils.getStringSpannableRobust(activity, org.commcare.dalvik.R.string.ok);
         factory.setPositiveButton(buttonDisplayText, buttonListener);
-        factory.showDialog();
+        if (activity instanceof CommCareActivity) {
+            ((CommCareActivity)activity).showAlertDialog(factory);
```
phillipm: `CommCareActivity` should probably be `DialogController`

app/src/org/commcare/android/framework/CommCareActivity.java
```diff
@@ -528,14 +545,18 @@ public void onClick(DialogInterface dialog, int i) {
         };
         CharSequence buttonDisplayText = StringUtils.getStringSpannableRobust(activity, org.commcare.dalvik.R.string.ok);
         factory.setPositiveButton(buttonDisplayText, buttonListener);
-        factory.showDialog();
+        if (activity instanceof CommCareActivity) {
+            ((CommCareActivity)activity).showAlertDialog(factory);
+        } else {
+            factory.showDialog();
```
phillipm: If you change `FormNavigationUI.activity` to be a `CommCareActivity`, not `Activity`, then there won't be a need to include this conditional

app/src/org/odk/collect/android/utilities/GeoUtils.java
```diff
@@ -79,7 +80,11 @@ public static void showNoGpsDialog(Activity activity,
             factory.makeCancelable();
             factory.setOnCancelListener(onCancel);
         }
-        factory.showDialog();
+        if (activity instanceof CommCareActivity) {
+            ((CommCareActivity)activity).showAlertDialog(factory);
+        } else {
+            factory.showDialog();
```
phillipm: Add a note that in this case the dialog isn't persisted between orientation changes

app/src/org/odk/collect/android/utilities/GeoUtils.java
```diff
@@ -79,7 +80,11 @@ public static void showNoGpsDialog(Activity activity,
             factory.makeCancelable();
             factory.setOnCancelListener(onCancel);
         }
-        factory.showDialog();
+        if (activity instanceof CommCareActivity) {
```
phillipm: Should probably be `DialogController` here too, then we can implement the appropriate methods in `GeoPointActivity` and dialog persistence will work.

PR #743, opened 2015-10-26
Don't try to get session before cc app is installed

phillipm:
Pressing back on the install screen caused a crash because the onResultActivity code path was trying to load the commcare session before an app was installed.

http://manage.dimagi.com/default.asp?185291


amstone326:
@phillipm Yikes sorry, this is definitely a change I made, thought I was just simplifying duplicate code.

PR #744, opened 2015-10-26
Use default title if refreshing breadcrumb title while updating app

phillipm:
If you rotate the screen while applying an app update then the breadcrumb bar crashes while trying to localize an installed resource because it isn't available. Wrap the offending line in a try/catch that returns a default title when this happens.

Isn't ideal, but is a minimal change and I suspect hunting down a robust solution would take me hours.


amstone326:
@phillipm Is there a FB ticket to go with this or did you just find the issue?

phillipm:
@amstone326 There is no ticket, I found it while fixing another FB ticket.

PR #745, opened 2015-10-26
Make alert dialogs cancelable

phillipm:
Pressing back on an alert dialog that has a cancel listener wasn't doing anything.

This fixes that is very unfortunate way.

### DIFF
--- a/app/src/org/commcare/android/framework/CommCareActivity.java
+++ b/app/src/org/commcare/android/framework/CommCareActivity.java
@@ -529,22 +529,34 @@ public abstract class CommCareActivity<R> extends FragmentActivity
     public static void createErrorDialog(final CommCareActivity activity, String errorMsg,
                                          final boolean shouldExit) {
         String title = StringUtils.getStringRobust(activity, org.commcare.dalvik.R.string.error_occured);
+
         AlertDialogFactory factory = new AlertDialogFactory(activity, title, errorMsg);
         factory.setIcon(android.R.drawable.ic_dialog_info);
-        DialogInterface.OnClickListener buttonListener = new DialogInterface.OnClickListener() {
-            @Override
-            public void onClick(DialogInterface dialog, int i) {
-                switch (i) {
-                    case DialogInterface.BUTTON_POSITIVE:
+
+        DialogInterface.OnCancelListener cancelListener =
+                new DialogInterface.OnCancelListener() {
+                    @Override
+                    public void onCancel(DialogInterface dialog) {
                         if (shouldExit) {
                             activity.setResult(RESULT_CANCELED);
                             activity.finish();
                         }
-                        break;
-                }
-            }
-        };
-        CharSequence buttonDisplayText = StringUtils.getStringSpannableRobust(activity, org.commcare.dalvik.R.string.ok);
+                    }
+                };
+        factory.setOnCancelListener(cancelListener);
+
+        CharSequence buttonDisplayText =
+                StringUtils.getStringSpannableRobust(activity, org.commcare.dalvik.R.string.ok);
+        DialogInterface.OnClickListener buttonListener =
+                new DialogInterface.OnClickListener() {
+                    @Override
+                    public void onClick(DialogInterface dialog, int i) {
+                        if (shouldExit) {
+                            activity.setResult(RESULT_CANCELED);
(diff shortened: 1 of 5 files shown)


app/src/org/commcare/dalvik/dialogs/AlertDialogFragment.java
```diff
@@ -32,6 +34,11 @@ public void onCreate(Bundle savedInstanceState) {
     }
 
     @Override
+    public void onCancel(DialogInterface dialog) {
+        factory.performCancel(dialog);
```
phillipm: :( I really tried to not do this by calling things like `dialog.cancel()`, but I couldn't find a way that worked. I think `AlertDialogFragment` and `AlertDialogFactory` or too closely coupled. We should try and find a way to make them less dependent

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -1837,11 +1838,13 @@ public void handleFormLoadCompletion(FormController fc) {
      * Call when the user provides input that they want to quit the form
      */
     private void triggerUserQuitInput() {
-        //If we're just reviewing a read only form, don't worry about saving
-        //or what not, just quit
-        if(mFormController.isFormReadOnly()) {
-            //It's possible we just want to "finish" here, but
-            //I don't really wanna break any c compatibility
+        if(!formHasLoaded()) {
+            finish();
```
phillipm: Making sure you can back out of a form entry without triggering an NPE if the form didn't load succesfully

app/src/org/commcare/android/framework/CommCareActivity.java
```diff
@@ -529,22 +529,34 @@ protected void createErrorDialog(String errorMsg, boolean shouldExit) {
...
                                          final boolean shouldExit) {
         String title = StringUtils.getStringRobust(activity, org.commcare.dalvik.R.string.error_occured);
+
         AlertDialogFactory factory = new AlertDialogFactory(activity, title, errorMsg);
         factory.setIcon(android.R.drawable.ic_dialog_info);
-        DialogInterface.OnClickListener buttonListener = new DialogInterface.OnClickListener() {
-            @Override
-            public void onClick(DialogInterface dialog, int i) {
-                switch (i) {
-                    case DialogInterface.BUTTON_POSITIVE:
```
phillipm: This switch isn't needed, right? This listener is always assigned to a single button, so we don't need to check the button type

app/src/org/commcare/dalvik/dialogs/AlertDialogFactory.java
```diff
@@ -19,6 +19,8 @@
 
     private final AlertDialog dialog;
     private final View view;
+    private DialogInterface.OnCancelListener cancelListener;
```
phillipm: Wish I could find a way to do this that didn't involve saving the cancel listener and manually calling it.

app/src/org/commcare/dalvik/dialogs/AlertDialogFragment.java
amstone326: I think this is probably fine and correct, I'm not sure what else you would do besides overriding the onCancel() here.

app/src/org/odk/collect/android/activities/FormEntryActivity.java
amstone326: @phillipm Going to merge this PR, but would be good to get a quick explanation of what this is doing-- seems unrelated to the rest of the PR.

phillipm: Yeah it is mostly unrelated; trying to make form entry erroring more robust. I found that if you dismissed the form load error dialog (which shouldn't be possible possible anymore) on form entry without finishing back to the previous activity, CommCare would crash when you tried to leave Form Entry because it assumed there was a form when running the exit code.

app/src/org/commcare/dalvik/dialogs/AlertDialogFragment.java
phillipm: Yeah, I guess its okay, just feels like an unnecessary level of indirection. The framework should be doing this work for us; I just wasn't able to find out how to get it to do that.

PR #749, opened 2015-10-27
Create a standard choice dialog

amstone326:
Creates a standardized choice dialog, which is being used by:
-FormEntryActivity.createQuitDialog() (the dialog that pops up when you press 'back' inside a form)
-FormEntryActivity.createRepeatDialog()
-FormEntryActivity.createLanguageDialog()
-CallOutActivity.showChoiceDialog()
-EntitySelectActivity.createSortMenu()
-AndroidShortcuts.buildMenuList()
-DrawActivity.createQuitDrawDialog()

For the exit form dialog, looking for input on the following:
-The right icon to use for each option
-The right text to use for each option
-The right order to display the options in
I'm guessing that I didn't get any of these exactly right on my own, so expecting to change them based on feedback. @biyeun, would be good to get your input.


amstone326:
Exit form dialog when incomplete forms is turned on:

screenshot_2015-10-28-11-35-18

amstone326:
Exit form dialog when incomplete forms is turned off:

screenshot_2015-10-28-11-37-30

amstone326:
Repeat dialog (essentially the same as before, just refactored to use the new PaneledChoiceDialog class, looks a little cleaner):

screenshot_2015-10-28-11-38-08

biyeun:
those icons look pixelated (the blue ones)---you you need or have access to the vector versions?

amstone326:
@biyeun If you're talking about the ones for the repeat dialog (<, + and >), those icons weren't touched or changed in this PR, I'm using the ones that were already being used. And I don't have access to another version of them.

amstone326:
As part of this PR I did a pass on a couple of the remaining dialogs that were not included in the first round of dialog refactors to make them look a little more consistent with the new look and feel, although they are not actually using either AlertDialogFactory or PaneledChoiceDialog:
-QuestionWidget.fireHelpText()
-DialogCreationHelpers.createAboutCommCareDialog()

This leaves only 2 dialogs in the app that still clearly have the old look, both of which I don't think can/should be changed:
-CommCareHomeActivity.showDemoModeWarning() -- This has a very different look and feel even from what the old alert dialogs looked like, achieved by deliberately using a different theme. I'm going on the assumption that this was intentional to make it very noticeable, and therefore shouldn't be changed.
-SpinnerMultiWidget - The view of this widget includes an Alert Dialog with multi-choice items -- don't think this should be changed since it is within a question widget.

amstone326:
@ctsims- Made the text changes you wanted, and made the icon choices configurable in dev options. Do you want to review and pull this, and then start doing the hallway testing, or just do the testing on a separate branch and wait to pull this until we've made a decision on the icons? 

@everyone- Either way, 99.9% of the code here is unrelated to the icon choices for that one dialog, so can still be reviewed.

app/src/org/commcare/dalvik/application/AndroidShortcuts.java
```diff
@@ -32,23 +34,32 @@ protected void onCreate(Bundle bundle) {
...
-                if("root".equals(m.getRoot())) {
+    private void buildMenuList() {
+        final PaneledChoiceDialog dialog = new PaneledChoiceDialog(this, "Select CommCare Shortcut");
+        dialog.setChoiceItems(getChoiceItemList(dialog));
+        dialog.setOnCancelListener(new OnCancelListener() {
+            public void onCancel(DialogInterface dialog) {
+                AndroidShortcuts sc = AndroidShortcuts.this;
+                sc.setResult(RESULT_CANCELED);
+                sc.finish();
+                return;
```
phillipm: Return statement not needed

app/src/org/commcare/dalvik/dialogs/DialogChoiceItem.java
```diff
@@ -0,0 +1,23 @@
...
+
+    public static final boolean ICON_TO_LEFT = true;
+    public static final boolean ICON_ON_TOP = false;
+
+    public DialogChoiceItem(String text, int iconResId, View.OnClickListener listener) {
+        this.text = text;
+        this.iconResId = iconResId;
+        this.listener = listener;
+    }
+}
```
phillipm: Super clean :+1:

app/src/org/commcare/dalvik/dialogs/DialogCreationHelpers.java
```diff
@@ -20,8 +21,11 @@ public static AlertDialog buildAboutCommCareDialog(Activity activity) {
 
         LayoutInflater li = LayoutInflater.from(activity);
         View view = li.inflate(R.layout.about_commcare_dialog, null);
-        TextView aboutText = (TextView)view.findViewById(R.id.about_commcare_text);
 
+        TextView titleView = (TextView) view.findViewById(R.id.dialog_title).findViewById(R.id.dialog_title_text);
+        titleView.setText("About CommCare");
```
phillipm: Should this text be available for localization?

app/src/org/commcare/dalvik/dialogs/PaneledChoiceDialog.java
```diff
@@ -0,0 +1,111 @@
...
+                    DialogChoiceItem.ICON_ON_TOP);
+        }
+    }
+
+    public static void populateChoicePanel(Context context, Button choicePanel,
+                                                 DialogChoiceItem item, boolean iconToLeft) {
+        choicePanel.setText(item.text);
+        choicePanel.setOnClickListener(item.listener);
+        if (item.iconResId != -1) {
+            Drawable icon = context.getResources().getDrawable(item.iconResId);
```
phillipm: `getDrawable` is deprecated, use `ContextCompat.getDrawable(context, R.drawable.***)`

app/src/org/commcare/dalvik/dialogs/PaneledChoiceDialog.java
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * Created by amstone326 on 10/27/15.
+ */
+public class PaneledChoiceDialog {
+
+    private View view;
+    private Context context;
+    private AlertDialog dialog;
+    private boolean usingThreePanelView;
```
phillipm: These can all be final

app/src/org/commcare/dalvik/dialogs/PaneledChoiceDialog.java
```diff
@@ -0,0 +1,111 @@
...
+
+    public static boolean HORIZONTAL_THREE_PANEL = true;
+
+    public PaneledChoiceDialog(Context context, String title) {
+        this(context, title, false);
+    }
+
+    public PaneledChoiceDialog(Context context, String title, boolean useSpecialThreePanelView) {
```
phillipm: Instead of having constructors with boolean flags that dictate instance behavior, some people advocate for having static factory methods to enable naming the instance creation method. In general, static factory methods are helpful in a [few different ways](http://www.informit.com/articles/article.aspx?p=1216151).

phillipm: If you aren't into making a static factory like `buildThreeChoiceDialog`, it might be even better to just create a subclass that appropriately overrides the constructor and `setChoiceItems` method such that you don't need the `usingThreePanelView` or `HORIZONTAL_THREE_PANEL` fields

app/src/org/commcare/dalvik/application/AndroidShortcuts.java
amstone326: @phillipm I think sometimes a return is actually needed after calling finish(). I'm not at all versed in when that is the case, but this is code that I just moved from one place to another so I think I'm going to leave it the way it was.

app/src/org/commcare/dalvik/dialogs/DialogCreationHelpers.java
amstone326: Probably. Again, this is just code I moved but I'll add it now.

app/src/org/commcare/dalvik/dialogs/PaneledChoiceDialog.java
amstone326: Good call, that's much cleaner.

amstone326:
Release Note: New look and feel for several dialogs.

app/src/org/commcare/dalvik/dialogs/HorizontalPaneledChoiceDialog.java
```diff
@@ -0,0 +1,38 @@
...
+ * A specialized version of PaneledChoiceDialog that shows exactly 3 choice items (rather than N
+ * items) in a horizontal (rather than vertical) orientation
+ *
+ * @author amstone
+ */
+public class HorizontalPaneledChoiceDialog extends PaneledChoiceDialog {
+
+    public HorizontalPaneledChoiceDialog(Context context, String title) {
+        super(context, title);
+        view = LayoutInflater.from(context).inflate(R.layout.choice_dialog_three_panel, null);
```
amstone326: @phillipm Seems pretty clean to me honestly, will go with that for now. Thanks for all the feedback.

amstone326:
@ctsims Did hallway testing yesterday, and consensus was pretty straightforward to choose from the results of that:
-Blue forward arrow for "Stay in Form" (the backward arrow was confusing to almost everyone because they interpreted that as "continue going backward"; Amelia nixed the green color because it indicates elsewhere being done with a form and submitting)
-Red trash can for "Delete and Exit" (using an x confused every person I tried because they interpreted that as 'dismiss this dialog and go back to where I was')
-Orange saved form icon for "Save and Exit" (this one didn't have an alternative icon but people seemed happy with it)

The only lingering concern is that Amelia pointed out that a trashcan may not be a universally-recognized symbol. But she also said that she couldn't think of any better alternatives and is happy for us to try this out and see how it goes.

I think you and Phillip have both already done passes on the actual code here, so this is ready to merged, assuming you're ok with the outcome.

ctsims:
@amstone326 Will do another quick review just in case. Could we get a screenshot in the PR, possibly?

I have two concerns about the trashcan
1) I agree that its meaning is extremely contextual. Folks in Bihar aren't going to get any meaning from a trashcan, and the "Trash" metaphor doesn't have much exposure tech-wise. It's also not super clear what it is, so it might be hard to direct people to.

It's good to know about people's "X" intuition, though, that's good evidence we shouldn't be using that either. The trash seems like the clear winner unless we can find something else. The only other suggestion I would have would be that it might make sense to standardize on Google's Material Design trashcan from their glyph set: https://github.com/google/material-design-icons since if there is any exposure to the glyph for most people I would expect it to be from this set (from other things on the phone).

amstone326:
@ctsims Screenshot:
screenshot_20151124-165922

I'm honestly not sure that https://www.google.com/design/icons/#ic_delete looks different enough from what we have to make it worth the switch. The benefit of leaving the current icon is that all 3 icons came from the same place, so they were sized together. The sizing options for the material design icons are slightly off of how the current icons are sized, so we would probably need to replace all 3. It also doesn't look like the material design site allows you to download in any colors. Do you think it still makes sense?

amstone326:
@ctsims Can we merge this so that it gets into 2.25? I can always change icons post-codefreeze since that doesn't affect QA.

PR #751, opened 2015-10-27
Catch xpath mismatch exception in FormHierarchyBuilder

phillipm:
Show the following text error instead of crashing when the form hierarchy view encounters a xpath type mismatch error:

screen

http://manage.dimagi.com/default.asp?115051


orangejenny:
Nice, much better than crashing.

Visual comment: can you change the text color, probably to white? The error is tough to read.

Would also be good to add something like `error_background` to colors.xml instead of using red directly. (Are we doing this kind of error handling generally? This is the same kind of use case as when there are exceptions in xpath in case list/details, where we just display text like `<invalid xpath: XPathMismatchExcepetion: blah blah blah`. Would be good for those to match eventually.)

phillipm:
@orangejenny haha, I thought about this and decided I was too lazy. Thanks for calling me out on it! Updated to hopefully use a more readable text color .

orangejenny:
looks much better, thanks!

ctsims:
Hey,

One thing on this: I'm not 100% sure I agree is uniformly better than
crashing. The thing is that our end-users don't have the ability to act on
those errors, and generally don't get that it's something that can be
fixed, so if we swallow those errors (until we have a way to bubble them up
to the app builders in a clear/concise way), it's somewhat equivalent to
never fixing them, whereas if the app crashes, their supervisor/app builder
_has_ to fix the problem.

It's a bit of a failfast v. fail silently issue. It might be better to let
people keep using the app, but it also might be worse for an app which has
a broken critical piece of functionality to fail to keep looking like its
working.

-Clayton

phillipm:
@ctsims I sorta agree, but we need to be consistent. Currently, when a user navigates to a question with an error in it, they are shown an error dialog and form entry closes rather than crashing the app. With this PR I was hoping to get something along those lines: show a message rather than crash CommCare. If you would like I can change the PR to show the same style of pop-up error message, which blocks all form hierarchy functionality, I can do that. Note that while blocking form hierarchy functionality makes the supervisor more likely to fix the problem, it also prevents the anyone from seeing directly which question is raising the exception.

phillipm:
Preventing the app from crashing means we don't see the exception on ACRA, which is good for us. It is also less jarring to the user, who can continue to use other aspects of the app.

ctsims:
Caught up with Phillip, fine with this solution for now and we're going to
plan on moving forward as soon as possible with getting the "App Error"
logs in a good state to be consumed by HQ.

-Clayton

app/src/org/odk/collect/android/views/HierarchyElementView.java
```diff
@@ -19,12 +19,14 @@ public HierarchyElementView(Context context, HierarchyElement it) {
...
         RelativeLayout layout = (RelativeLayout)inflate(context, R.layout.hierarchy_element_view, null);
 
-        setBackgroundColor(it.getColor());
+        setBackgroundColor(it.getBgColor());
 
         mPrimaryTextView = ((TextView)layout.findViewById(R.id.hev_primary_text));
         mPrimaryTextView.setText(it.getPrimaryText());
+        mPrimaryTextView.setTextColor(it.getTextColor());
         mSecondaryTextView = ((TextView)layout.findViewById(R.id.hev_secondary_text));
         mSecondaryTextView.setText(it.getSecondaryText());
```
wpride: Shouldn't this (and above) use setPrimaryText() and setTextcolor() respectively?

app/src/org/odk/collect/android/adapters/HierarchyListAdapter.java
```diff
@@ -41,10 +41,11 @@ public View getView(int position, View convertView, ViewGroup parent) {
             hev = new HierarchyElementView(mContext, mItems.get(position));
         } else {
             hev = (HierarchyElementView)convertView;
-            hev.setPrimaryText(mItems.get(position).getPrimaryText());
-            hev.setSecondaryText(mItems.get(position).getSecondaryText());
-            hev.setIcon(mItems.get(position).getIcon());
-            hev.setColor(mItems.get(position).getColor());
+            HierarchyElement entry = mItems.get(position);
```
wpride: Might be nice to add a function used by this and the HierarchyElementView that accepts a HierarchyElement and then parse out the colors/texts (since this is a little repeated in HierarchyElementview)

phillipm:
@wpride Good suggestions, I've updated the PR (also includes https://github.com/dimagi/commcare-odk/pull/779 otherwise it wouldn't build locally)

PR #752, opened 2015-10-27
Add high res launcher icons

phillipm:
Was getting annoyed by the grainy launcher icon on higher res screens.

Before:
old

After:
screen

Notice that the new icons, generated by [Android Asset Studio](https://romannurik.github.io/AndroidAssetStudio/icons-launcher.html), are slightly smaller, think this is probably okay. They are the same size as Instagram's icon :)


amstone326:
Should this be going into 2.24? Don't know how much we care about being conservative with what we sneak in after code-freeze, but this seems borderline.

amstone326:
(Happy to merge it if that's not something we care about enforcing, definitely a nice add).

ctsims:
Hey,

This isn't the new icon, is it? I thought we had one with the other blue?

PR #753, opened 2015-10-27
Stop groups at end of form from preventing user from reaching end   

phillipm:
Save form and exit if we somehow get into the end of form state, which is not supposed to happen anymore. For example, this form, which is technically invalid because it has an unbounded repeat group inside of a question list, gets into that state:
https://www.commcarehq.org/a/flipper/apps/view/36c0bdd028d14a52cbff95bb1bfd0962/modules-0/forms-2/


amstone326:
Why are we keeping the EVENT_END_OF_FORM state in the code at all if it's not supposed to happen anymore?

phillipm:
I think EVENT_END_OF_FORM is still pretty integral to form manipulations, there is just no longer a view for it in the form entry activity

PR #755, opened 2015-10-28
Set FormRecord class field properly

amstone326:
The value of lastModified was not actually getting set in the constructor if the if value passed in was null. Not sure what the implications of this are or if it could have been causing any bugs -- if it could be, should  probably just put in 2.24.


phillipm:
Nice catch! Did you find this via lint warnings or did it come up while using the app?

amstone326:
@phillipm Just saw it while looking at the file

PR #757, opened 2015-10-28
Propagate parent detail variables to all child details

orangejenny:
Extend https://github.com/dimagi/commcare-odk/pull/717 to apply to all child details, not only those that iterate over subnodes.

@ctsims


orangejenny:
moving to master

PR #758, opened 2015-10-28
Remove deprecated size call in banner customization

phillipm:
Remove deprecated calls to `Display.getWidth()` and `Display.getHeight()` in custom banner display code.

### DIFF
--- a/app/src/org/commcare/android/framework/CommCareActivity.java
+++ b/app/src/org/commcare/android/framework/CommCareActivity.java
@@ -8,13 +8,13 @@ import android.content.SharedPreferences;
 import android.graphics.Bitmap;
 import android.os.Build;
 import android.os.Bundle;
+import android.support.annotation.NonNull;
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/android/framework/CommCareActivity.java
```diff
@@ -70,8 +69,6 @@
     private static final String KEY_PROGRESS_DIALOG_FRAG = "progress-dialog-fragment";
     private static final String KEY_ALERT_DIALOG_FRAG = "alert-dialog-fragment";
 
-    private boolean mBannerOverriden = false;
```
phillipm: I really don't understand how this field was being used in `updateCommCareBanner`. Pretty sure there was no need for it.

amstone326: Haha this looks like it was so convolutedly unnecessary.

PR #761, opened 2015-10-28
Make question waiting for data persist accross views

phillipm:
If you launch an image/audio/video chooser, rotate the device, and choose something, the data is lost. This occurs because rotating causes the questions to be recreated and that loses the flag that signifies the question is waiting for data. Use pending intent widget infrastructure that Danny wrote.   

http://manage.dimagi.com/default.asp?185734


phillipm:
@dannyroberts I'm using your pending intent widget infrastructure on all questions that use any sort of callout

app/src/org/odk/collect/android/views/ODKView.java
```diff
@@ -332,26 +330,17 @@ public void setFocus(Context context) {
...
         }
 
-        if (!set) {
-            Log.w(TAG, "Attempting to return data to a widget or set of widgets not looking for data");
-                     
-            for (QuestionWidget q : widgets) {
-                if (q instanceof IBinaryWidget) {
-                    ((IBinaryWidget) q).setBinaryData(answer);
-                    break;
-                }
```
phillipm: Don't understand why this `!set` conditional is here. I took it out, is that wrong?

app/src/org/odk/collect/android/widgets/IntentWidget.java
```diff
@@ -104,9 +98,6 @@ public void makeTextView() {
     public void makeButton(){
         setOrientation(LinearLayout.VERTICAL);
 
-        TableLayout.LayoutParams params = new TableLayout.LayoutParams();
-        params.setMargins(7, 5, 7, 5);
-
```
phillipm: Pretty sure this isn't needed

app/src/org/odk/collect/android/widgets/QuestionWidget.java
```diff
@@ -649,4 +649,7 @@ public Spannable forceMarkdown(String text){
     public Spannable stylize(String text){
         return MarkupUtil.styleSpannable(getContext(), text);
     }
+
+    public void setBinaryData(Object answer) {
+    }
```
phillipm: Removed `IBinaryWidget` interface and moved the `setBinaryData` method here. This lets us not need to check `instanceof IBinaryWidget` in `ODKView.setBinaryData`.

app/src/org/odk/collect/android/views/ODKView.java
```diff
@@ -332,28 +332,23 @@ public void setFocus(Context context) {
...
+        for (QuestionWidget q : widgets) {
+            if (questionFormIndex.equals(q.getFormId())) {
+                q.setBinaryData(answer);
+                pendingCalloutInterface.setPendingCalloutFormIndex(null);
+                return;
             }
         }
+        Logger.log(AndroidLogger.SOFT_ASSERT,
+                "Unable to find question widget to attach pending data to.");
```
phillipm: Added some soft asserts because this should never happen

app/src/org/odk/collect/android/views/ODKView.java
amstone326: It looks like that was there for if you had already gone through one full loop of the question widgets and not set binary data to any of them, then it did another loop through to just set data to all binary widgets, even if they didn't have the waiting for data flag set. Since you refactored this and removed the 2nd iteration through, it's not wrong to remove it.

app/src/org/odk/collect/android/widgets/AudioWidget.java
```diff
@@ -70,10 +71,9 @@ public void onClick(View v) {
                 Intent i = new Intent(android.provider.MediaStore.Audio.Media.RECORD_SOUND_ACTION);
                 i.putExtra(android.provider.MediaStore.EXTRA_OUTPUT,
                         android.provider.MediaStore.Audio.Media.EXTERNAL_CONTENT_URI.toString());
-                mWaitingForData = true;
                 try {
-                    ((Activity)getContext())
-                            .startActivityForResult(i, FormEntryActivity.AUDIO_VIDEO_FETCH);
+                    ((Activity)getContext()).startActivityForResult(i, FormEntryActivity.AUDIO_VIDEO_FETCH);
+                    pendingCalloutInterface.setPendingCalloutFormIndex(prompt.getIndex());
```
amstone326: I don't think it actually matters, but I usually try to make any calls to a startActivity() method the last call I make in a sequence.

app/src/org/odk/collect/android/widgets/IntentWidget.java
amstone326: What makes you think that? Could be totally true, but shouldn't take it out unless you're sure.

app/src/org/odk/collect/android/widgets/QuestionWidget.java
amstone326: Should just add a comment that this gets implemented in the subclasses that need it.

amstone326:
@phillipm Left a couple of very minor, optionally actionable comments. Other than that, can you just update the class-level comment for PendingCalloutInterface now that its usage is more broad than what Danny had there? Good to merge after that.

app/src/org/odk/collect/android/widgets/IntentWidget.java
phillipm: `params` is never passed into anything, so it doesn't seem used, but the linter doesn't say it is dead code. I just tested it and it is okay to remove.

app/src/org/odk/collect/android/widgets/AudioWidget.java
phillipm: The ordering is important here because I only want to call `setPendingCalloutFormIndex` if `startActivityForResult` doesn't raise a `ActivityNotFoundException` exception. I could refactor these statements be order-agnostic, but that would require more code, which I opted to avoid.

phillipm:
@amstone326 thanks, updated the PR based on your feedback

PR #762, opened 2015-10-29
Dynamic refreshing of question content in question lists

amstone326:
Previously, if the question text or select options for one question in a question list were dependent upon the answer to a prior question in the group, the question text / select options would NOT get updated when a user changed their answer to the prior question. This fixes that.

cross: https://github.com/dimagi/javarosa/pull/188


app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -605,46 +607,76 @@ private void processIntentResponse(Intent response, boolean cancelled) {
...
         saveAnswersForCurrentScreen(DO_NOT_EVALUATE_CONSTRAINTS);
-        
+
         if(!(mCurrentView instanceof ODKView)){
-            throw new RuntimeException("Tried to update form relevency not on compound view");
+            throw new RuntimeException("Tried to update form relevancy not on compound view");
         }
-        
+
         ODKView oldODKV = (ODKView)mCurrentView;
```
wpride: Know this isn't your variable name but feels like a misnomer since this is the current view

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -605,46 +607,76 @@ private void processIntentResponse(Intent response, boolean cancelled) {
...
-        
-        FormEntryPrompt[] newValidPrompts = mFormController.getQuestionPrompts();
-        Set<FormEntryPrompt> used = new HashSet<>();
-        
         ArrayList<QuestionWidget> oldWidgets = oldODKV.getWidgets();
+        ArrayList<Vector<SelectChoice>> oldSelectChoices = getOldSelectChoicesForEachWidget(oldWidgets);
```
wpride: Question: why create these separate array lists for the String and Select choice before the loop instead of calling these functions on 653 and 654? (Except for the individual widget instead of as a group)? At a higher level it seems not great to have these three concurrent sets that we rely on always being sorted the same

app/src/org/odk/collect/android/widgets/SelectOneWidget.java
```diff
@@ -173,20 +173,14 @@ public int getCheckedId() {
     @Override
     public void onCheckedChanged(CompoundButton buttonView, boolean isChecked) {
         onUserInteracton();
-        
-        widgetEntryChanged();
-        
-        if (!isChecked) {
-            // If it got unchecked, we don't care.
```
wpride: Lol this seems just like not true

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -605,46 +607,76 @@ private void processIntentResponse(Intent response, boolean cancelled) {
         }
     }
 
-    private void updateFormRelevencies(){
+    /**
+     * @return A list of the select choices for each widget in the list of old widgets, with the
+     * original order preserved
+     */
+    private ArrayList<Vector<SelectChoice>> getOldSelectChoicesForEachWidget(ArrayList<QuestionWidget> oldWidgets) {
```
phillipm: Would be nice to move the method definition to somewhere below its usage

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -605,46 +607,76 @@ private void processIntentResponse(Intent response, boolean cancelled) {
...
+        FormEntryPrompt[] newValidPrompts = mFormController.getQuestionPrompts();
+        Set<FormEntryPrompt> leftInOldList = new HashSet<>();
 
-            for(FormEntryPrompt prompt : newValidPrompts) {
-            	if(prompt.getIndex().equals(oldWidget.getPrompt().getIndex())) {
-            		stillRelevent = true;
-            		used.add(prompt);
-            	}
+        ArrayList<Integer> removeList = new ArrayList<>();
+        for (int i = 0; i < oldWidgets.size(); i++){
```
phillipm: If you follow @wpride 's advice of reducing the 3 for loops to just one you'll also be able to use a foreach here instead of the less readable index-based iteration you're currently using.

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -605,46 +607,76 @@ private void processIntentResponse(Intent response, boolean cancelled) {
...
+            FormEntryPrompt oldPrompt = oldWidgets.get(i).getPrompt();
+            String priorQuestionTextForThisWidget = oldQuestionTexts.get(i);
+            Vector<SelectChoice> priorSelectChoicesForThisWidget = oldSelectChoices.get(i);
+            boolean stillRelevant = false;
+
+            for (FormEntryPrompt newPrompt : newValidPrompts) {
```
phillipm: I would put this loop in a helper method

app/src/org/odk/collect/android/activities/FormEntryActivity.java
amstone326: Sorry I meant to leave a comment here explaining that -- was deliberate and necessary because of the referencing issue I was talking about. Since oldODKV is just a reference to the current view (as you pointed out; you're right that that should be renamed), once the for loop below starts, the select choices and question texts that the "old widgets" reference obviously start changing to be the updated ones, making it impossible to access them anymore.

app/src/org/odk/collect/android/activities/FormEntryActivity.java
amstone326: @phillipm See my response to Will above for why I couldn't do that

amstone326:
@phillipm, @wpride Updated with clearer variable names, a comment to explain the unintuitive ordering of stuff, and the helper method Phillip suggested.

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -605,47 +607,101 @@ private void processIntentResponse(Intent response, boolean cancelled) {
...
-
-            for(FormEntryPrompt prompt : newValidPrompts) {
-            	if(prompt.getIndex().equals(oldWidget.getPrompt().getIndex())) {
-            		stillRelevent = true;
-            		used.add(prompt);
-            	}
-            }
-            if(!stillRelevent){
-                removeList.add(i);
+        FormEntryPrompt[] newValidPrompts = mFormController.getQuestionPrompts();
```
ctsims: The issue that requires us to save the two arrays above is that we can't compare these directly (IE: "hasSameDisplayContent(FormPrompt oldFormPrompt)") because the prompt objects themselves are getting manipulated, right?

amstone326: Correct. I originally implemented it exactly like that, and then realized that it wasn't working for that reason.

amstone326:
Release Note: Fixed issues where various question elements (question label, Lookup Table contents, etc), if located in a question list, wouldn't update dynamically when a related question was changed.

PR #765, opened 2015-10-30
call dialog.dismiss() for no gps dialog

amstone326:
dialog.dismiss() wasn't getting called where it should be for the onClickListener in one of the dialogs, making it impossible to close. I also did a quick pass over all of the new dialogs to make sure there weren't any others left with the same problem.

Also makes text capitalization for the no gps dialog consistent, and removes a noop from 1 file.


orangejenny:
looks fine, will wait for tests

PR #767, opened 2015-11-02
Propagate parent detail variables to all child details

orangejenny:
Move https://github.com/dimagi/commcare-odk/pull/757 to master

### DIFF
--- a/app/src/org/commcare/android/framework/EntityDetailFragment.java
+++ b/app/src/org/commcare/android/framework/EntityDetailFragment.java
@@ -72,7 +72,7 @@ public class EntityDetailFragment extends Fragment {
         // Note that some of this setup could be moved into onAttach if it would help performance
         Detail childDetail = getChildDetail();
         TreeReference childReference = getChildReference();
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/android/framework/EntityDetailFragment.java
```diff
@@ -70,8 +71,8 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
 
         // Note that some of this setup could be moved into onAttach if it would help performance
         Detail childDetail = getChildDetail();
-        NodeEntityFactory factory = new NodeEntityFactory(childDetail, asw.getEvaluationContext());
         TreeReference childReference = getChildReference();
+        NodeEntityFactory factory = new NodeEntityFactory(childDetail, this.getFactoryContext());
```
ctsims: It worries me that the reference retrieved above (line 74) and below (line 116) are the same, but coupled. Seems worth passing in rather than duplicating to future proof against synchronicity issues between the two.

orangejenny: fixed

PR #768, opened 2015-11-02
Update processAction override for JavaRosa performance PR

phillipm:
Needed for https://github.com/dimagi/javarosa/pull/174


app/src/org/odk/collect/android/jr/extensions/PollSensorAction.java
```diff
@@ -100,6 +101,8 @@ public void run() {
                 requestLocationUpdates(providers);
             }
         });
+
+        return null;
```
phillipm: This returns null because the actual value setting happens in a callback. This isn't ideal, but shouldn't interfere with how the result of `processAction` is used in `FormDef`

PR #769, opened 2015-11-03
In entity subnode details, ignore blank rows

orangejenny:
Make data-driven entity detail screens behave like entity selection screens: don't display rows that have no data.

### DIFF
--- a/app/src/org/commcare/android/adapters/EntitySubnodeDetailAdapter.java
+++ b/app/src/org/commcare/android/adapters/EntitySubnodeDetailAdapter.java
@@ -13,7 +13,7 @@ import org.commcare.suite.model.Detail;
 import org.javarosa.core.model.instance.TreeReference;
 
 import java.util.List;
-import java.util.Vector;
+import java.util.ArrayList;
 
 /**
  * Created by jschweers on 8/24/2015.
--- a/app/src/org/commcare/android/adapters/EntitySubnodeDetailAdapter.java
+++ b/app/src/org/commcare/android/adapters/EntitySubnodeDetailAdapter.java
@@ -32,7 +32,7 @@ public class EntitySubnodeDetailAdapter implements ListAdapter, ModifiableEntity
     public EntitySubnodeDetailAdapter(Context context, Detail detail, List<TreeReference> references, List<Entity<TreeReference>> entities, ListItemViewModifier modifier) {
         this.context = context;
         this.detail = detail;
-        this.references = references;
+        this.modifier = modifier;
         this.initializeData(references, entities);
     }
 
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/android/adapters/EntitySubnodeDetailAdapter.java
```diff
@@ -32,8 +33,24 @@ public EntitySubnodeDetailAdapter(Context context, Detail detail, List<TreeRefer
...
+        for (Entity e : entities) {
+            boolean isValid = false;
+            for (int i = 0; i < e.getNumFields(); i++) {
+                if (e.isValidField(i)) {
+                    isValid = true;
+                }
+            }
+            if (isValid) {
+                this.entities.add(e);
+                this.references.add(references.get(entityIndex));
```
phillipm: The `isValid` boolean isn't needed: if you move these two `add`s into the body of the `e.isValidField(i)` conditional and then break afterwards it should be the same behavior without an intermediate boolean.

phillipm: I would move all this logic into a helper method

orangejenny:
@phillipm comments addressed

ctsims:
Hey,

Major question here: What is "Blank" in the context of a detail node,
exactly? All fields not being empty strings? Seems reasonable as kind of an
optimization, but I'd like for us to have a longer-term plan for filtering
here.

-Clayton

app/src/org/commcare/android/adapters/EntitySubnodeDetailAdapter.java
```diff
@@ -32,10 +33,28 @@ public EntitySubnodeDetailAdapter(Context context, Detail detail, List<TreeRefer
         this.context = context;
         this.detail = detail;
         this.references = references;
```
phillipm: I wouldn't perform this field set if you are then going to wipe it in `initializeData`

app/src/org/commcare/android/adapters/EntitySubnodeDetailAdapter.java
```diff
@@ -32,10 +33,28 @@ public EntitySubnodeDetailAdapter(Context context, Detail detail, List<TreeRefer
...
-        this.entities = entities;
-        this.modifier = modifier;
+        this.initializeData(references, entities);
     }
 
+    /**
+     * Populate entities and references. Will ignore any entities (and their
+     * associated references) without any valid fields.
+     */
+    private void initializeData(List<TreeReference> references, List<Entity<TreeReference>> entities) {
```
phillipm: Can you rename the method arguments here so that they don't conflict with class fields? Doing so will enable you to omit the `this` keyword.

app/src/org/commcare/android/adapters/EntitySubnodeDetailAdapter.java
```diff
@@ -32,10 +33,28 @@ public EntitySubnodeDetailAdapter(Context context, Detail detail, List<TreeRefer
...
+        this.references = new Vector<>();
+        this.entities = new Vector<>();
```
phillipm: ArrayLists are preferable over Vectors, right?

orangejenny:
@ctsims 

Yeah, right now it's looking for all blanks, which makes app building awkward. For the purpose of defining "blank row", it'd be useful to distinguish between data fields and label fields (possibly by making a new format option for labels?).

For more complex filtering, I'd imagine each field knowing whether or not it's relevant, and hiding rows that are all irrelevant (and label fields being always "irrelevant", rather than using their own form).

app/src/org/commcare/android/adapters/EntitySubnodeDetailAdapter.java
orangejenny: typo, that should have been `modifier`

app/src/org/commcare/android/adapters/EntitySubnodeDetailAdapter.java
orangejenny: I like using `this` to make it obvious that you're dealing with member variables. But agree that having the names match is confusing.

ctsims:
@orangejenny Yeah. I think what we want to probably do for filtering long-term is probably defining a set of filters over the entity outputs themselves, since this will let us run filters on the cached computation results, rather than as a defined filter, which we won't be able to cache/optimize.

Anyway, this seems like a reasonable first-pass in the meantime, although something we probably want to be careful to sunset once we have a more concrete way to do it.

orangejenny:
@ctsims Makes sense. Added a note to the docs for this; that's the best way I have atm to not forget.

PR #776, opened 2015-11-04
Auto-login for developer debugging

phillipm:
Allow developers to store username & password for an app such when debugging the app it auto-logs in for them.

Enabled in the developer preferences. When you login, it will save the last password such that the next time you need to login it does it for you. If you press the logout button it won't auto-login but will populate the field password field for you.

If you create an app with the following preferences in your `profile.ccpr` file:
```
    <property key="last_password" value="a_great_password"/>
    <property key="last_logged_in_user" value="clarice"/>
    <property key="cc-enable-auto-login" value="yes" force="true"/>
```
then the app will auto-login with those credentials when the app launches.


app/src/org/commcare/android/session/DevSessionRestorer.java
```diff
@@ -0,0 +1,57 @@
...
+    public static void tryAutoLoginPasswordSave(String password) {
+        if (autoLoginEnabled()) {
+            SharedPreferences prefs =
+                    CommCareApplication._().getCurrentApp().getAppPreferences();
+            prefs.edit().putString(CommCarePreferences.LAST_PASSWORD, password).commit();
+        }
+    }
+
+    private static boolean autoLoginEnabled() {
+        return BuildConfig.DEBUG && DeveloperPreferences.isAutoLoginEnabled();
```
phillipm: If we want anyone to be able to use this feature, not only devs who are doing debugging, we need to remove `BuildConfig.DEBUG` from this predicate.

PR #777, opened 2015-11-05
Perform restore from local xml file on app install

phillipm:
contains commits from https://github.com/dimagi/commcare-odk/pull/655


phillipm:
Only needed for parto demo app

PR #781, opened 2015-11-06
D3 scaffolding

orangejenny:
Opening PR for feedback on general approach, which is to use D3 inside of a WebView. Planning to do a bunch of small/medium PRs and merge them all into the `graphing` branch, then merge to master when everything is ready.

Pros:
- Greater range of functionality that native android graphing libraries (compared with AChartEngine, MPAndroidChart, and HelloCharts, which were the only three I found that met our most basic requirements)
- We get total control over look and feel
- D3 is very well supported and documented

Cons:
- D3's API is fairly low-level; it will take substantially more work to get it running than if we used an actual graphing library
- Introducing JavaScript to the code base adds system complexity
- I'm a bit worried about performance. Initial tests took ~5 seconds to load a graph. Using `hardwareAccelerated` and `LOAD_NO_CACHE` has cut this back to ~1 second.


app/src/org/commcare/android/view/GraphView.java
```diff
@@ -163,33 +169,28 @@ private void setPanAndZoom(boolean allow) {
...
+        
+        WebView.setWebContentsDebuggingEnabled(true);   // TODO: only if in dev
+        WebView webView = new WebView(mContext);
+        configureSettings(webView);
+
+        String html =
+                "<html>" +
+                    "<head>" +
+                        "<link rel='stylesheet' type='text/css' href='file:///android_asset/graphing/graph.css'></link>" +
+                        "<script type='text/javascript' src='file:///android_asset/graphing/underscore.min.js'></script>" +
```
orangejenny: underscore isn't really necessary, but it's convenient and ~6KB

phillipm:
I have limited knowledge in this area but it looks pretty good; nice to see that it is fairly concise to switch to JS graphs. Are you looking to have this merged? If so I'd say remove the commented out code and it should be good.

orangejenny:
I'd like to merge it, yeah, so I can then do more PRs on top of it. Will remove the commented-out code.

orangejenny:
Leaving the commented-out code for now, it'll be handy to have a reference when I re-implement full-screen graphs. Self-merging (just into the graphing branch, not master.)

One more comment on the approach: I'm now looking at using http://nvd3.org/ on top of D3. Looks like it supports ~everything we need and will save time.

PR #782, opened 2015-11-06
disable case tiles when in landscape mode of form entry

amstone326:


phillipm:
Are we sure we want to take this approach to increasing the visible form entry area while in landscape mode? What about making persistent case tiles collapse-able and then auto-collapsing while in landscape mode?

app/src/org/commcare/android/framework/BreadcrumbBarFragment.java
```diff
@@ -360,6 +363,10 @@ public void onResume() {
             }
         }
     }
+
+    private boolean inLandscapeFormEntry(Activity a) {
+        return (a instanceof FormEntryActivity) && getResources().getConfiguration().orientation == Configuration.ORIENTATION_LANDSCAPE;
```
phillipm: This function can be made static if you call `getResources()...` on the `a` argument

amstone326:
@phillipm Not sure what you mean by "making persistent case tiles collapse-able and then auto-collapsing while in landscape mode" or how it's different from this. Can you explain?

ctsims:
Case tiles are still very custom, so I'm inclined to not make any drastic
UI changes based on them without thinking hard about it. Anyone using case
tiles already knows how it's affecting their screens/projects, so I kind of
don't want to be making big changes to their experience.

No strong feelings though. Most case tile users are locked in portrait mode.

amstone326:
Ok fair, I think I'm gonna just close this for now then.

PR #784, opened 2015-11-06
Don't allow keyboard suggestions

wpride:
One fix for the lack of form entry space in the new UI

In classic Android fashion the supposed flag for this doesn't work and you have to use this "textFilter" flag:

http://stackoverflow.com/questions/4281838/how-to-disable-displaying-suggestions-on-the-soft-keyboard


amstone326:
So we want to just shut this off everywhere, not only in landscape or something like that?

Also should this have a parallel PR into 2.24 for the hotfix?

ctsims:
@amstone326 - I think it would be much more confusing for folks if it worked in some orientations but not others. It's an active/functional component, so it would be tough to figure out what's going on with it

@wpride 
We should just push to 2.24 and close this out (it'll get merged when the hotfix branch does)

PR #787, opened 2015-11-06
Fix for overzealous display of group label data

ctsims:
Implemented an adaptive layout technique for shrinking the question group label max size based on the screen dimensions, triggering on any major changes to the screen size (soft keyboard).

Fix for: http://manage.dimagi.com/default.asp?186516

### DIFF
--- a/app/src/org/commcare/android/framework/CommCareActivity.java
+++ b/app/src/org/commcare/android/framework/CommCareActivity.java
@@ -144,6 +144,17 @@ public abstract class CommCareActivity<R> extends FragmentActivity
         mGestureDetector = new GestureDetector(this, this);
     }
 
+    /**
+     * Call this method from an implementing activity to request a new event trigger for any time
+     * the available space for the core content view changes significantly, for instance when the
+     * soft keyboard is displayed or hidden.
+     *
+     * This method will also be reliably triggered upon the end of the first layout pass, so it
+     * can be used to do the initial setup for adaptive layouts as well as their updates.
+     *
+     * After this is called, major layout size changes will be triggered in the onMajorLayoutChange
+     * method.
+     */
     protected void requestMajorLayoutUpdates() {
         final View decorView = getWindow().getDecorView();
         decorView.getViewTreeObserver().addOnGlobalLayoutListener(new ViewTreeObserver.OnGlobalLayoutListener() {
@@ -153,16 +164,15 @@ public abstract class CommCareActivity<R> extends FragmentActivity
             @Override
             public void onGlobalLayout() {
                 Rect r = new Rect();
-                //r will be populated with the coordinates of your view that area still visible.
+                //r will be populated with the coordinates of your view that are visible after the
+                //recent change.
                 decorView.getWindowVisibleDisplayFrame(r);
 
                 int mainContentHeight = r.height();
 
-                //int viewFrameWindowDifference = Math.abs(mainContentHeight - (r.bottom - r.top));
-
                 int previousMeasurementDifference = Math.abs(mainContentHeight - mPreviousDecorViewFrameHeight);
 
-                if (previousMeasurementDifference > 100) { // if more than 100 pixels, its probably a keyboard...
+                if (previousMeasurementDifference > 100) {
                     onMajorLayoutChange(r);
                 }
                 mPreviousDecorViewFrameHeight = mainContentHeight;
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/android/framework/CommCareActivity.java
```diff
@@ -141,6 +144,52 @@ protected void onCreate(Bundle savedInstanceState) {
...
     }
 
+    protected void requestMajorLayoutUpdates() {
+        final View decorView = getWindow().getDecorView();
+        decorView.getViewTreeObserver().addOnGlobalLayoutListener(new ViewTreeObserver.OnGlobalLayoutListener() {
+
+            int mPreviousDecorViewFrameHeight = 0;
+
+            @Override
+            public void onGlobalLayout() {
```
wpride: If I'm understanding correctly: every time we setup the layout again, we get the currently available amount of space for the actual form. If this has changed by more than 100px we reason that the keyboard position has changed, and so update the space allotment for the group label. Is that the correct? If so, is there any reason not to check directly if the keyboard is up or not?

wpride: (also comment explaining this would be good)

app/src/org/commcare/android/framework/CommCareActivity.java
```diff
@@ -141,6 +144,52 @@ protected void onCreate(Bundle savedInstanceState) {
...
+
+                if (previousMeasurementDifference > 100) { // if more than 100 pixels, its probably a keyboard...
+                    onMajorLayoutChange(r);
+                }
+                mPreviousDecorViewFrameHeight = mainContentHeight;
+            }
+        });
+    }
+
+    protected void onMajorLayoutChange(Rect newRootViewDimensions) {
```
wpride: comment explaining when/how to override this?

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2377,4 +2385,87 @@ private void setTitleToLoading() {
...
+
+        float textSize = groupLabel.getTextSize();
+
+        int numberOfGroupLinesAllowed = (int)((spaceAvailable - defaultHeaderSpace) / textSize);
+
+        if(numberOfGroupLinesAllowed < 0) {
+            numberOfGroupLinesAllowed = 0;
+        }
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) {
```
wpride: Why?

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2377,4 +2385,87 @@ private void setTitleToLoading() {
...
+
+        int spaceAvailable = availableWindow - spaceRequested;
+
+        int defaultHeaderSpace = this.getResources().getDimensionPixelSize(R.dimen.content_min_margin) * 2;
+
+        float textSize = groupLabel.getTextSize();
+
+        int numberOfGroupLinesAllowed = (int)((spaceAvailable - defaultHeaderSpace) / textSize);
+
+        if(numberOfGroupLinesAllowed < 0) {
```
wpride: this is unnecessary given line 2433 below, right?

app/src/org/commcare/android/framework/CommCareActivity.java
ctsims: Hey,

Yeah, the idea is basically that you call this to request an update any time your size changes dramatically. 

It's not necessarily specific to keyboard popup, and I don't think we would suppress the event if the keyboard hadn't changed. The idea is essentially that if you have any layout code that is dependent on screen size, this will give you the right hook for when to re-evaluate your static screen size.

I can add docs

app/src/org/odk/collect/android/activities/FormEntryActivity.java
ctsims: getMaxLines is apparently only gettable on API16+

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2377,4 +2385,87 @@ private void setTitleToLoading() {
...
+            updateGroupViewVisibility();
+            groupLabel.setMaxLines(0);
+        } else {
+            this.mGroupForcedInvisible = false;
+            updateGroupViewVisibility();
+            groupLabel.setMaxLines(numberOfGroupLinesAllowed);
+        }
+    }
+
+    private void updateGroupViewVisibility() {
```
phillipm: It would be nice if this method took in arguments and used them to correctly set the state of `mGroupNativeVisibility` and `mGroupForcedInvisible`. That way the manipulation of these class-level flags is encapsulated in this method call.

PR #795, opened 2015-11-09
Restore last case list query

phillipm:
Store the last query for a case list in the current sessions top frame. Hence, if you continue forward in a session, when you back out to the same case list, the query is restored from the session's stack. If you back out past the case list selection view and then forward into it again, the query won't be restored; implementing the desired behavior. 

fixes: http://manage.dimagi.com/default.asp?183986
cross request: https://github.com/dimagi/commcare/pull/191

### DIFF
--- a/app/src/org/commcare/android/framework/CommCareActivity.java
+++ b/app/src/org/commcare/android/framework/CommCareActivity.java
@@ -4,7 +4,6 @@ import android.annotation.TargetApi;
 import android.app.Activity;
 import android.content.Context;
 import android.content.DialogInterface;
-import android.content.SharedPreferences;
 import android.graphics.Bitmap;
 import android.os.Build;
 import android.os.Bundle;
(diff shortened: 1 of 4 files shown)


app/src/org/commcare/android/framework/CommCareActivity.java
```diff
@@ -494,15 +483,12 @@ public void cancelCurrentTask() {
         stateHolder.cancelTask();
     }
 
-    protected void saveLastQueryString(String key) {
-        SharedPreferences settings = getSharedPreferences(CommCarePreferences.ACTIONBAR_PREFS, 0);
-        SharedPreferences.Editor editor = settings.edit();
-        editor.putString(key, lastQueryString);
-        editor.commit();
+    protected void restoreLastQueryString() {
+        lastQueryString = CommCareApplication._().getCurrentSession().getCurrentFrameStepExtra(KEY_LAST_QUERY_STRING);
```
amstone326: @phillipm The behavior here is such that the lastQueryString will be null if it wasn't explicitly set. Just confirming that that's what we want it to be?

phillipm: Hmmm, looks like that was also the behavior before with
```
lastQueryString = settings.getString(key, null);
```
so we should be good.

PR #797, opened 2015-11-09
Clear session on app swipe close

phillipm:
If you swipe close the android app and then re-open it the current commcare session is still around (used to show up in oldui breadcrumb bar, on new ui you have to inspect the object using the debugger). The user isn't restored to that session's current state, but rather taken to the home screen.

This PR catches the swipe close action via `onTaskRemoved` and resets the session. 

Fixes http://manage.dimagi.com/default.asp?168706
Contains commits from https://github.com/dimagi/commcare-odk/pull/796

### DIFF
--- a/app/src/org/commcare/dalvik/application/CommCareApplication.java
+++ b/app/src/org/commcare/dalvik/application/CommCareApplication.java
@@ -906,6 +906,7 @@ public class CommCareApplication extends Application {
         // class name because we want a specific service implementation that
         // we know will be running in our own process (and thus won't be
         // supporting component replacement by other applications).
+        startService(new Intent(this, CommCareSessionService.class));
         bindService(new Intent(this, CommCareSessionService.class), mConnection, Context.BIND_AUTO_CREATE);
         mIsBinding = true;
     }
@@ -1050,6 +1051,7 @@ public class CommCareApplication extends Application {
                 mIsBound = false;
                 // Detach our existing connection.
                 unbindService(mConnection);
+                stopService(new Intent(this, CommCareSessionService.class));
             }
         }
     }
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/dalvik/services/CommCareSessionService.java
```diff
@@ -121,13 +122,23 @@ public void onCreate() {
...
 
+    @Override
+    public void onTaskRemoved(Intent rootIntent) {
+        try {
+            CommCareApplication._().getCurrentSessionWrapper().reset();
+        } catch (SessionStateUninitException e) {
+            Log.e(AndroidLogger.SOFT_ASSERT,
+                    "Trying to wipe uninitialized session in session service tear-down");
+        }
+    }
```
phillipm: Is called when the app is swipe closed from Android's task manager

app/src/org/commcare/dalvik/application/CommCareApplication.java
```diff
@@ -908,6 +906,7 @@ public void onServiceDisconnected(ComponentName className) {
         // class name because we want a specific service implementation that
         // we know will be running in our own process (and thus won't be
         // supporting component replacement by other applications).
+        startService(new Intent(this, CommCareSessionService.class));
         bindService(new Intent(this, CommCareSessionService.class), mConnection, Context.BIND_AUTO_CREATE);
```
phillipm: Binding the service with an auto-start flag doesn't ensure that `onTaskRemoved` is called when the app is swipe-closed. Hence we have to manually call `startService` and then bind it.

app/src/org/commcare/dalvik/application/CommCareApplication.java
```diff
@@ -1052,6 +1051,7 @@ private void unbindUserSessionService() {
                 mIsBound = false;
                 // Detach our existing connection.
                 unbindService(mConnection);
+                stopService(new Intent(this, CommCareSessionService.class));
```
phillipm: Since we are now starting the service via `startService` we need to close it down using `stopService` because I don't think unbinding it will do that for us

PR #798, opened 2015-11-09
Refactor: make mCurrentView an ODKView not View instance

phillipm:
There was a lot of boilerplate around the fact that `FormEntryActivity.mCurrentView` wasn't of the type `ODKView`.

Also remove need to create `FormNavigationUI` instances, since the one public method it contains can be static.


app/src/org/odk/collect/android/views/ODKView.java
```diff
@@ -42,15 +41,12 @@
...
     // starter random number for view IDs
     private final static int VIEW_ID = 12345;  
-    
-    private static final String TAG = ODKView.class.getSimpleName();
 
     private final LinearLayout mView;
     private final LinearLayout.LayoutParams mLayout;
     private final ArrayList<QuestionWidget> widgets;
     private final ArrayList<View> dividers;
-    private ProgressBar mProgressBar;
```
amstone326: @phillipm Why did the mProgressBar and mProgressEnabled fields, and all logic surrounding them, get deleted?

phillipm: The progress bar now lives somewhere that is not the ODKView, so `updateProgressBar` was never called.

app/src/org/odk/collect/android/views/ODKView.java
```diff
@@ -434,10 +426,6 @@ public void removeWidget(View v){
      * @return Index of question's view in mView.
      */
     private int getViewIndex(int questionIndex) {
-        // Account for progress bar
-        if (mProgressEnabled) {
```
amstone326: @phillipm Are you sure this code isn't relevant anymore though? I don't know this part of the code, just seems like a weird thing to be removing

phillipm: yeah, `mProgressEnabled` was never set, so it was always false, so that conditional was never satisfied.

PR #799, opened 2015-11-09
Clear question widget listeners to prevent memory leaks

phillipm:
Question widget listeners were never getting unset, which leaked some memory.

It probably doesn't actually leak very much memory, regardless, unsetting listeners is an important pattern and we should get in the habit of using it. 

Includes commits from https://github.com/dimagi/commcare-odk/pull/798


amstone326:
@phillipm Good to merge as soon as my question on https://github.com/dimagi/commcare-odk/pull/798 is addressed, since it contains that commit

PR #802, opened 2015-11-10
Ensure AndroidInstanceInitializer uses the android version of data instance classes

phillipm:
If this is the correct pattern for making the android instance initializer use android version of data instance classes, then I will implement it for ledgers, fixtures, and sessions.

fixes http://manage.dimagi.com/default.asp?186528
cross-requested with https://github.com/dimagi/commcare/pull/192

NOTE: we need to remove `org.commcare.android.util.CommCareInstanceInitializer` because it isn't used any more.


phillipm:
@wpride Try to not merge PRs when there is a "don't pull" label.
In this case I wanted to include similar changes for ledgers, fixtures, and sessions

PR #803, opened 2015-11-10
Prevent NPE in GET_CASE switch of onActivityResult

phillipm:
Fix a null pointer exception. __I have no idea if blindly preventing the NPE is the right approach.__ If anyone is familiar with this, please weigh in.

Crash stacktrace:
https://commcare-acralyzer.cloudant.com/acralyzer/_design/acralyzer/index.html#/reports-browser/commcare-odk/bug/b08b624100441f02ede2ab35fb0e1bdc


amstone326:
Hey- just looking at this quickly, I'm pretty sure there's a bigger problem here. If we're returning from getting a case, the current session should definitely have a needed datum. Will look into the crash more as soon as I can.

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -549,9 +547,12 @@ else if(resultCode == RESULT_OK){
                 if (resultCode == RESULT_CANCELED) {
```
amstone326: @ctsims What does the TODO comment above the getSession() call mean? Is there like some edge case when the session needs to be loaded in another way? I'm hoping that what was null here was actually just currentSession, and not currentSession.getNeededDatum(), because that seems like a scarier failure mode.

amstone326: @phillipm Talked to Clayton about this, he agreed with what you were saying that an exception should get thrown if the session was null, so it's most likely that the getNeededDatum() is in fact null. We discussed what's likely causing it, and strategy for at least handling the situation more carefully, so I'll grab this PR and implement that soon. If you're curious about what it is, it's a quick explanation, just long to type.

amstone326:
@phillipm closing this, will replace with a new PR

PR #807, opened 2015-11-10
Fix crash when rotating form entry view w/ date/time pickers

phillipm:
If you rotate the a device while in a form that has TimePicker or DateTimePicker widgets then it crashes with `java.lang.IndexOutOfBoundsException: setSpan (2 ... 2) ends beyond length 0`

I was only able to trigger this crash on the [registration form of the mLabour app](https://www.commcarehq.org/a/mlabour/apps/view/6bbf69cec278cff14d152ae8d1a30fe8/modules-1/forms-0/). I tried to rotate on other forms that had TimePickers without success. 

This fix probably only works for devices running Android 3 and up, but that is okay with me, since it seems hard to encounter.

https://commcare-acralyzer.cloudant.com/acralyzer/_design/acralyzer/index.html#/reports-browser/commcare-odk/bug/1c3ed844920b9be36e634e2a31ec3933


phillipm:
@amstone326 This fixes the rotation crash on the mLabour Android app I mentioned to you earlier. I assume it isn't worth fixing and uploading an apk..

PR #808, opened 2015-11-10
Localize button text for update activity

phillipm:
"Check" and "Install" weren't available for localization:
image

Addresses http://manage.dimagi.com/default.asp?186391


amstone326:
@phillipm If we do a hotfix for https://github.com/dimagi/commcare-odk/pull/810, should we consider including this as well? It obviously doesn't break apps but does seem pretty shitty that people can't translate major components of the UI. (If we do hotfix, I'll also need to add those strings manually to the HQ commcare-translations doc).

phillipm:
@amstone326 :+1:

PR #810, opened 2015-11-11
Save form as complete if auto-advance question reaches End-of-form

phillipm:
Auto-advance questions at the end of form were resulting in the form being saved as incomplete. Instead auto-advancing past the end of form should be the same as pressing the 'next arrow' button 

Should consider hotfixing this.

http://manage.dimagi.com/default.asp?186847


amstone326:
Merged this, but agreed that it seems potentially hotfix-worthy. @ctsims what do you think?

PR #811, opened 2015-11-12
Capture and restore user sessions

phillipm:
Capture and restore user sessions. Piggyback off the auto-login dev option; only restoring captured sessions when that is enabled.

Doesn't support capturing and restoring the session stack, so restoring form linking type sessions won't work.

To capture the current session, run the following command from the command line:
`adb shell am broadcast -a org.commcare.dalvik.api.action.SessionCaptureAction`

Session will be restored upon login.

@amstone326 we should put this in the dev tool app so that users can submit bug reports that include the session of the commcare app open.

cross-requested with https://github.com/dimagi/commcare/pull/195

__Relevant to GS__: See https://confluence.dimagi.com/display/ccinternal/Advanced+Mobile+Debugging


app/src/org/commcare/android/session/DevSessionRestorer.java
```diff
@@ -47,11 +62,75 @@ public static void tryAutoLoginPasswordSave(String password) {
...
+            CommCareApplication._().getCurrentSession().serializeSessionState(serializedStream);
+        } catch (IOException e) {
+            Log.w(TAG, "Failed to serialize session");
+            return;
+        } catch (SessionStateUninitException e) {
+            Log.w(TAG, "Attempting to save a non-existent session");
+            return;
+        }
+        String serializedSession = Base64.encodeToString(baos.toByteArray(), Base64.DEFAULT);
+        prefs.edit().putString(CommCarePreferences.CURRENT_SESSION, serializedSession).commit();
```
phillipm: Does it matter that we are storing unencrypted session info in the app preferences?

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -371,6 +371,15 @@ else if(resultCode == RESULT_OK){
...
                 } else if(resultCode == RESULT_OK) {
                     if (!intent.getBooleanExtra(LoginActivity.ALREADY_LOGGED_IN, false)) {
+                        CommCareSession session = CommCareApplication._().getCurrentSession();
+                        if (session.getCommand() != null) {
+                            // restore the session state if there is a command.
+                            // For debugging and occurs when a serialized
+                            // session is stored upon login
+                            sessionNavigator.startNextSessionStep();
+                            return;
+                        }
```
phillipm: Is `session.getCommand() != null` the right check here?

amstone326: I'm not sure... You're essentially trying to check whether the session has any info in it yet, right? Another option would be to just set a flag in the case when you restore the session in CommCareApplication.bindUserSessionService(), and then we know for sure that it's correct. Otherwise, would want to double check with @ctsims that this is the right way of doing it

PR #812, opened 2015-11-12
Better logging and error message for form save failure

wpride:
See http://manage.dimagi.com/default.asp?186674

We should be logging this error anyways since we had no visibility. Unclear if we gain anything from percolating the error to the user level but I think not. 

@phillipm I saw you suggested one of your changes might've impacted this behavior (IE failing on form save instead of later) - this seems better to me (IE better than sending to HQ and failing there)


ctsims:
@wpride we are only pushing to logcat here? Can we push this to our forensic logs as well

PR #813, opened 2015-11-12
Properly respect markdown in select choices, refactor

wpride:
See 186542

Also removed a function we weren't using

cross https://github.com/dimagi/javarosa/pull/194


ctsims:
"Pull requests that have a failing status can’t be merged from a phone."

What the hell?

PR #814, opened 2015-11-12
Session fix 2

wpride:
fix for http://manage.dimagi.com/default.asp?185986

We weren't updating the current User (maintained by the session service) until logout/login so the custom user data wasn't getting udpated


app/src/org/commcare/android/tasks/DataPullTask.java
```diff
@@ -259,6 +259,7 @@ public void reportProgress(int progress) {
                         this.context.sendBroadcast(i);
                         
                         Logger.log(AndroidLogger.TYPE_USER, "User Sync Successful|" + username);
+                        CommCareApplication._().getSession().setCurrentUser(CommCareApplication._().getCurrentUserFromDB(username), password);
```
phillipm: Might be good to use the same db user fetching method that `updateUserSyncToken` uses, instead of using `getCurrentUserFromDB`.
Maybe something like:
```
    private void updateSessionUser(String password) throws SessionUnavailableException {
        SqlStorage<User> storage = CommCareApplication._().getUserStorage("USER", User.class);
        User u = storage.getRecordForValue(User.META_USERNAME, username);
        CommCareApplication._().getSession().setCurrentUser(u, password);
    }
```
I think using `getUserStorage` offers another layer of protection over `getRawStorage`, though I might be mistaken; I don't really understand all the different ways to access the db.

phillipm:
We should look into whether this fixes http://manage.dimagi.com/default.asp?185816 as well

app/src/org/commcare/dalvik/application/CommCareApplication.java
```diff
@@ -102,6 +102,7 @@
 import java.util.Comparator;
 import java.util.Date;
 import java.util.HashSet;
+import java.util.NoSuchElementException;
```
phillipm: Good to go after you remove this line

PR #816, opened 2015-11-13
Product-Level analytics

amstone326:
Includes reporting for all events that Amelia/Kevin/I identified as wanting in the first pass for product-level analytics. The GoogleAnalyticsUtils and GoogleAnalyticsFields classes provide good context and explanations for what was added, and I outlined/explained the hierarchy of event reporting in more detail here: https://docs.google.com/a/dimagi.com/spreadsheets/d/1-vdt6oiaHwNcbN7Dz6qc6o69RA3J5C-eLP85hf0z2po/edit?usp=sharing

@dimagi/product


app/AndroidManifest.xml
```diff
@@ -6,7 +6,9 @@
           android:versionName="2.26">
 
     <uses-sdk
-        tools:overrideLibrary="com.etsy.android.grid"
+        tools:overrideLibrary="com.etsy.android.grid, com.google.android.gms.measurement,
```
amstone326: Does anyone know the implications of using this? Will it just fail gracefully on versions that don't support the library, or do something more sinister?

app/src/org/commcare/android/analytics/GoogleAnalyticsUtils.java
```diff
@@ -0,0 +1,232 @@
...
+                .setCategory(category)
+                .setAction(action)
+                .setLabel(label)
+                .build());
+    }
+
+    /**
+     * Report a google analytics event that has a category, action, label, and value
+     */
+    public static void reportEvent(String category, String action, String label, int value) {
```
phillipm: instances of `reportEvent` can be made `private`.

app/src/org/commcare/dalvik/application/CommCareApplication.java
```diff
@@ -314,6 +323,25 @@ public CallInPhoneListener getCallListener() {
...
+            mTracker = getAnalyticsInstance().newTracker(DEV_TRACKING_ID);
+            mTracker.enableAutoActivityTracking(true);
+        }
+        String userId = getCurrentUserId();
+        if (!"".equals(userId)) {
+            mTracker.set("&uid", userId);
+        } else {
+            mTracker.set("&uid", null);
+        }
+        return mTracker;
```
phillipm: `getDefaultTracker` is called every time an event is reported. It is worth thinking about setting up the default tracker in a manner that doesn't require you to set the user id every time `getDefaultTracker` is called. Like if you setup that part of the tracker every time a new user logs in, you don't need to do it every time you call `getDefaultTracker`. That said, it might be a more complex implementation that isn't worth the effort. I'll leave it to you to decide.

app/src/org/commcare/android/analytics/TimedStatsTracker.java
```diff
@@ -0,0 +1,112 @@
+package org.commcare.android.analytics;
+
+import android.content.SharedPreferences;
+import android.util.Log;
+
+import org.commcare.android.javarosa.AndroidLogger;
+import org.commcare.dalvik.application.CommCareApp;
+import org.commcare.dalvik.application.CommCareApplication;
+import org.javarosa.core.services.Logger;
+import org.joda.time.DateTime;
```
phillipm: Many of these imports are unused. I think our plugin that auto removes these isn't around anymore :(

app/src/org/commcare/android/analytics/TimedStatsTracker.java
```diff
@@ -0,0 +1,112 @@
...
+    private static String KEY_LAST_FORM_START_TIME = "last-form-start-time";
+    private static String KEY_LAST_SESSION_START_TIME = "last-session-start-time";
+    private static String KEY_LAST_LOGGED_IN_USER = "last-logged-in-user";
+
+    public static void registerEnterForm(String formTitle) {
+        SharedPreferences.Editor editor =
+                CommCareApplication._().getCurrentApp().getAppPreferences().edit();
+        editor.putString(KEY_LAST_FORM_NAME, formTitle).
+                putLong(KEY_LAST_FORM_START_TIME, currentTime());
+        editor.commit();
```
phillipm: Might be nice to chain the `editor.commit()` call to the previous line for consistency.
Also, I'm a fan of putting the `.` on the new line with the chained method call, because that offers an additional piece of visual info to distinguish the chained call from a normal method invocation.

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2554,6 +2592,18 @@ private int getFontSizeInPx() {
                 getResources().getDisplayMetrics());
     }
 
+    private void reportFormEntry() {
+        TimedStatsTracker.registerEnterForm(getCurrentFormTitle());
+    }
+
+    private void reportFormExit() {
+        TimedStatsTracker.registerExitForm(getCurrentFormTitle());
+    }
```
phillipm: Do we care if the form was completed or not? I think `reportFormExit` is called even if the form wasn't completed or even saved.

app/src/org/commcare/android/analytics/TimedStatsTracker.java
```diff
@@ -0,0 +1,112 @@
...
+                        "last user reported as logged in");
+            }
+        } else {
+            Log.i(TAG, "Attempting to report ending a session when there is no last logged in " +
+                    "user in prefs");
+        }
+    }
+
+    private static int computeElapsedTimeInSeconds(long startTime, long endTime) {
+        return (int) (endTime - startTime / 1000);
```
phillipm: adding an extra pair of around `endTime - startTime` would be nice for those of us who never memorized arithmetic operator precedence, like me :)

phillipm: err, aren't the parens actually required in this case, since `/` has a higher precedence than `-`.

app/src/org/commcare/dalvik/activities/HomeButtons.java
```diff
@@ -136,6 +141,7 @@ public void onClick(View v) {
     private static View.OnClickListener getIncompleteButtonListener(final CommCareHomeActivity activity) {
         return new View.OnClickListener() {
             public void onClick(View v) {
+                reportButtonClick(GoogleAnalyticsFields.LABEL_INCOMPLETE_FORMS_BUTTON);
```
phillipm: Seems a little redundant to log an event here and also within the `FormRecordListActivity` that is launched by pressing this button.

app/src/org/commcare/dalvik/preferences/CommCarePreferences.java
```diff
@@ -102,38 +104,74 @@
     // Fields for setting print template
     private static final int REQUEST_TEMPLATE = 0;
     public final static String PRINT_DOC_LOCATION = "print_doc_location";
-    private final static String PREF_MANAGER_PRINT_KEY = "print-doc-location";
+
+    private final static String PREFS_APP_SERVER_KEY = "default_app_server";
```
phillipm: This is also defined in `ResourceInstallUtils`. We should unify them.

app/src/org/commcare/dalvik/preferences/CommCarePreferences.java
```diff
@@ -102,38 +104,74 @@
...
+    private final static String PREFS_DATA_SERVER_KEY = "ota-restore-url";
+    private final static String PREFS_SUBMISSION_URL_KEY = "PostURL";
+    private final static String PREFS_KEY_SERVER_KEY = "default_key_server";
+    private final static String PREFS_FORM_RECORD_KEY = "form-record-url";
+    private final static String PREFS_UPDATE_FREQUENCY_KEY = "cc-autoup-freq";
+    private final static String PREFS_FUZZY_SEARCH_KEY = "cc-fuzzy-search-enabled";
+    private final static String PREFS_LOCALE_KEY = "cur_locale";
+    private final static String PREFS_PRINT_KEY = "print-doc-location";
```
phillipm: The strings that these keys are set to are used in several places in the code base. Would be nice to have them point to these fields so we only need to change them in one place. not a blocker but would be a good follow up

phillipm:
Nice!
Finished a pass and left some questions and actionable items.

phillipm:
Heads up that a change removing the "Call Log" settings entry has been merged into master.

app/src/org/odk/collect/android/activities/FormEntryActivity.java
amstone326: I was thinking of this as just the average time spent in a single form entry session, regardless of whether or not the form was completed/saved. But the right answer really just depends on how Kevin/Amelia were intending to use this metric, so will defer to them. @dimagi/product -- Would you guys rather this include all form entry sessions, or only those that resulted in form save or completion?

app/src/org/commcare/android/analytics/TimedStatsTracker.java
amstone326: Yikes :( :( Thanks

app/src/org/commcare/dalvik/activities/HomeButtons.java
amstone326: Eh that's true in theory, but because of the hierarchical way in which the reporting works and how the product team will be making use of that, it makes sense to just provide it to them in both places.

app/src/org/odk/collect/android/activities/FormEntryActivity.java
amsagoff: I think every session, @krsdimagi ?

krsdimagi: @amstone326 definitely looking for every session.

app/src/org/commcare/dalvik/activities/HomeButtons.java
amstone326: :+1: cool, will leave as is

app/AndroidManifest.xml
ctsims: My understanding is that it prevents the builds from failing, but puts the onus on not-importing or referencing those files onto us, so we have to be cautious to clip them out of incompatible files.

What's the Min SDK version for these?

app/src/org/commcare/android/analytics/TimedStatsTracker.java
```diff
@@ -0,0 +1,109 @@
...
+ * Utils for keeping track of the length of some app events, and then reporting them to Google
+ * analytics
+ *
+ * @author amstone
+ */
+public class TimedStatsTracker {
+
+    private static final String TAG = TimedStatsTracker.class.getSimpleName();
+
+    private static String KEY_LAST_FORM_NAME = "last-form-name";
```
ctsims: I'm not 100% sure I think it's clean for us to record this. It's app/user specific data which makes it feel like it shouldn't bubble up to the analytics framework. It's also fairly tied towards a user's specific behavior, so if we're assuming this isn't "secure" data, it sketches me out that we'd be correlating position, time , and a form, that feels a lot like a patient identifying event.

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -1050,37 +1064,51 @@ public boolean onPrepareOptionsMenu(Menu menu) {
     public boolean onOptionsItemSelected(MenuItem item) {
         switch (item.getItemId()) {
             case MENU_PREFERENCES:
+                reportOptionsItemEntry(GoogleAnalyticsFields.LABEL_SETTINGS);
```
ctsims: Oof, this is pretty ugly. I think making an item ID -> Key map and adding the event against that map above the switch might be a cleaner way to implement this.

app/src/org/commcare/dalvik/activities/FormAndDataSyncer.java
```diff
@@ -128,30 +129,63 @@ public void syncData(boolean formsToSend,
             @Override
             protected void deliverResult(CommCareHomeActivity receiver, Integer result) {
                 receiver.getUIController().refreshView();
+                String reportSyncLabel;
+                int reportSyncValue;
 
                 //TODO: SHARES _A LOT_ with login activity. Unify into service
                 switch (result) {
                     case DataPullTask.AUTH_FAILED:
+                        reportSyncLabel = GoogleAnalyticsFields.LABEL_SYNC_FAILURE;
```
ctsims: hm, this looks pretty tough to manage to me. I wonder whether it's time to move these integer return codes to be enums that have tagged attributes that we can return such that we can just issue the results to the analytics framework with one call, rather than one-per-case.

app/src/org/commcare/dalvik/activities/HomeButtons.java
```diff
@@ -166,6 +172,7 @@ public void update(HomeCardDisplayData cardDisplayData,
     private static View.OnClickListener getLogoutButtonListener(final CommCareHomeActivity activity) {
         return new View.OnClickListener() {
             public void onClick(View v) {
+                reportButtonClick(GoogleAnalyticsFields.LABEL_LOGOUT_BUTTON);
```
ctsims: Is there any chance that we want to make a, like, 

"AnalyticsReportingOnClickListener()"

that we can use to handle this kind of thing a bit more generically? In theory we could use a few different strategies to get a String to Label the button

app/src/org/commcare/dalvik/application/CommCareApplication.java
```diff
@@ -793,6 +826,25 @@ public boolean matches(UserKeyRecord ukr) {
         CommCareApplication._().closeUserSession();
     }
 
+    public String getCurrentUserId() {
+        if (currentUserId != null) {
+            return currentUserId;
+        }
+        try {
+            currentUserId = this.getSession().getLoggedInUser().getUniqueId();
```
ctsims: why aren't we just getting this live each time? It doesn't feel like this is a super expensive call?

app/src/org/commcare/dalvik/application/CommCareApplication.java
```diff
@@ -793,6 +826,25 @@ public boolean matches(UserKeyRecord ukr) {
...
+            return currentUserId;
+        } catch (SessionUnavailableException e) {
+            return "";
+        }
+    }
+
+    private void refreshUserIdCache() {
+        try {
+            currentUserId = this.getSession().getLoggedInUser().getUniqueId();
+        } catch (SessionUnavailableException e) {
```
ctsims: If there's no session shouldn't it clear the user ID?

app/src/org/commcare/dalvik/preferences/DeveloperPreferences.java
```diff
@@ -43,27 +42,123 @@
...
     @Override
     public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, String key) {
-        if (key.equals(ENABLE_AUTO_LOGIN) && !isAutoLoginEnabled()) {
-            DevSessionRestorer.clearPassword(sharedPreferences);
-        } else if (key.equals(ENABLE_SAVE_SESSION) && !isSessionSavingEnabled()) {
-            DevSessionRestorer.clearSession();
+        String editPrefLabel = "";
+        switch(key) {
+            case SUPERUSER_ENABLED:
+                editPrefLabel = GoogleAnalyticsFields.LABEL_DEV_MODE;
```
ctsims: I'm confused, we build a big table of these above that we should be able to use to create this table, right? I don't think we should have to manually code it in again?

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -370,6 +380,7 @@ public void onClick(View v) {
             @Override
             public void onClick(View v) {
                 if (!NAV_STATE_QUIT.equals(v.getTag())) {
+                    GoogleAnalyticsUtils.reportFormNavBackward(GoogleAnalyticsFields.LABEL_ARROW);
                     FormEntryActivity.this.showPreviousView(true);
                 } else {
                     FormEntryActivity.this.triggerUserQuitInput();
```
ctsims: I'd actually love to know whether user's are using this v. 'back' v. " < Form Name Here" to exit forms, actually, can we record this too?

ctsims:
@amstone326 What's the minimum API verison of the google analytics API's, btw? A few places in the code make it look like it's > than our min, but the code doesn't seem to reflect that on the whole

app/src/org/commcare/android/analytics/TimedStatsTracker.java
amstone326: Ok, will remove

app/src/org/commcare/dalvik/application/CommCareApplication.java
amstone326: Haha, @phillipm specifically suggested caching this since it gets called every time I report an event. Should I revert back?.. Don't really feel strongly either way. It seemed a little bit dangerous to cache instead of just making sure it's fresh each time, but it is a lot of repetitive calls.

app/src/org/commcare/dalvik/activities/HomeButtons.java
amstone326: Not really sure what you mean here

app/src/org/commcare/dalvik/preferences/DeveloperPreferences.java
amstone326: Fixed

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
amstone326: Fixed here and in all similar places

app/src/org/commcare/dalvik/application/CommCareApplication.java
amstone326: Yea, I think I left that out b/c this only ever gets called right after we've bound the user session service, but should be there for good measure anyway.

app/src/org/commcare/dalvik/activities/FormAndDataSyncer.java
amstone326: Done

app/src/org/odk/collect/android/activities/FormEntryActivity.java
amstone326: Added in ec13e63e4288001a5aff2a705aa6e689b58442cd

amstone326:
@ctsims Addressed and implemented changes for all comments except the 1 that I didn't really understand. Left a question about my solution to the version compatibility issue.

app/src/org/commcare/dalvik/activities/HomeButtons.java
ctsims: Just that it seems verbose to add these in as code lines. In theory we could have this be a thing that happens automatically based on the button having an annotation or something if we reliably used a common onclicklistener. Now that the other stuff is less verbose I'm less concerned though

app/src/org/commcare/dalvik/application/CommCareApplication.java
```diff
@@ -118,8 +123,13 @@
         reportType = org.acra.sender.HttpSender.Type.JSON,
         httpMethod = org.acra.sender.HttpSender.Method.PUT)
 public class CommCareApplication extends Application {
+
     private static final String TAG = CommCareApplication.class.getSimpleName();
 
+    // Tracking ids for Google Analytics
+    private static final String LIVE_TRACKING_ID = "UA-69708208-1";
```
ctsims: hm, do we want these to get injected by gradle? Seems bad for us to have them be public TBH.

app/src/org/commcare/dalvik/application/CommCareApplication.java
ctsims: Oh, I see. Hm, my guess is that the google analytics code behind reporting the event is going to take more time than doing this 3 part call (I don't think these methods have any ~body, it's just the stack setup).

If we're concerned about runtime I'd validate how much time this particular call takes against the rest of the report, if it's not pronounced I vote we kill it.

build.gradle
```diff
@@ -13,17 +13,21 @@ import com.android.builder.core.DefaultManifestParser;
 buildscript {
     repositories {
         mavenCentral()
+        jcenter()
     }
 
     dependencies {
         classpath 'com.android.tools.build:gradle:1.5.0'
+        classpath 'com.google.gms:google-services:1.4.0-beta3'
```
ctsims: is there areason we're using the beta of his dependency? Seems sketchy...

ctsims:
@amstone326 Ok, this looks pretty legit now. I think we should likely kill the user ID caching, and there's the 2.2 test pending, but am happy with the code overall.

app/src/org/commcare/dalvik/application/CommCareApplication.java
amstone326: @ctsims How would I set that up?

phillipm: If it is alright to have just on release builds then you would add it to /home/jenkins/.gradle/gradle.properties on the jenkins mobile server. Then you would have to wire things up in similar ways to how ACRA and other keys are setup: https://github.com/dimagi/commcare-odk/blob/master/build.gradle#L77-L79

phillipm: That would also work if we want local dev build to have the key, you would just need to add the key to your local gradle.properties

build.gradle
amstone326: I just copied this directly from google's tutorial page on setting up google analytics for mobile. Not sure why they're suggesting beta but it appears there isn't actually a non-beta version of 1.4.0: https://jcenter.bintray.com/com/android/tools/build/gradle/. I tried updating to 1.5.0 just now and got a compatibility error that I couldn't solve off-hand, and don't think are worth sinking time into now.

app/src/org/commcare/dalvik/application/CommCareApplication.java
amstone326: Ok done

amstone326:
@ctsims I added the opt out language for legal, and added the functionality you wanted for tracking CC Version with app installs. This should be good to pull now that we know we're not doing anything for 2.2 compat!

ctsims:
Awesome!! Can we re-run on 2.3.7 and make sure we're good on that as well?
I think should be fine, but want to make sure the new odd inclusions are
working.

-Clayton

amstone326:
@ctsims What are "odd inclusions"?

ctsims:
The way we're injecting the new libraries, mostly (with the overwrite
stuff), but we may not need to do that anymore with the 2.2 exclusion?

amstone326:
Oh actually.. I should probably be actually changing our app's min SDK in
this PR, and then remove the override stuff I was doing, right?

PR #818, opened 2015-11-13
Target android 6

phillipm:
Handling changes described in https://developer.android.com/about/versions/marshmallow/android-6.0-changes.html

__Release Notes__: Update app to target Android 6 platform

Mostly update runtime permission acquisition:
 - Ask for all permissions at login screen
 - If external storage perms aren't granted, block login, since form entry fails without it.
 - on Android 6 and above, allow user to trigger permission request dialog via login menu. Enables supervisor to grant permissions in one go, so that users don't get confused while using the app.
 - If location permissions aren't granted, then location updates won't be registered during form entry or mapping activities. Currently no dialog pops up asking to enable the permissions, that is a TODO.

__NOTE__: I'm not checking external storage read/write permissions. External storage usage is used everywhere in our app, so it would be really hard to explicitly check. Not to worry: turns out that if you have a app service running and then you change the app permissions via the settings menu, then the whole app is restarted. This means there is currently no way to login if you have disabled external storage.

Also remove call log feature that wasn't working

screen

screen

screen


app/src/org/commcare/dalvik/application/CommCareApplication.java
```diff
@@ -356,6 +343,10 @@ public void initializeGlobalResources(CommCareApp app) {
     }
 
     public String getPhoneId() {
+        if (ContextCompat.checkSelfPermission(this, Manifest.permission.READ_PHONE_STATE) == PackageManager.PERMISSION_DENIED) {
+            return "000000000000000";
```
phillipm: Default to fake imei if we don't have correct permissions

app/src/org/odk/collect/android/jr/extensions/PollSensorAction.java
```diff
@@ -156,8 +169,11 @@ public void onLocationChanged(Location location) {
                     mModel.setValue(val == null ? null: AnswerDataFactory.templateByDataType(dataType).cast(val.uncast()), qualifiedReference);
                 }
             }
-            
-            if (location.getAccuracy() <= GeoUtils.GOOD_ACCURACY) {
+
+            Context context = CommCareApplication._().getApplicationContext();
+            if (location.getAccuracy() <= GeoUtils.GOOD_ACCURACY
+                    && (ContextCompat.checkSelfPermission(context, Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED ||
```
ctsims: This is repeated a bunch of times in this file and should be moved to a method

app/src/org/odk/collect/android/jr/extensions/PollSensorAction.java
```diff
@@ -110,13 +117,19 @@ public void run() {
...
+                        ContextCompat.checkSelfPermission(context, Manifest.permission.ACCESS_COARSE_LOCATION) == PackageManager.PERMISSION_GRANTED)) {
             mLocationManager.removeUpdates(PollSensorAction.this);
             return;
         }
         
         for (String provider : providers) {
-            mLocationManager.requestLocationUpdates(provider, 0, 0, PollSensorAction.this);
+            if ((provider.equals(LocationManager.GPS_PROVIDER) && ContextCompat.checkSelfPermission(context, Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED) ||
+                    (provider.equals(LocationManager.NETWORK_PROVIDER) && ContextCompat.checkSelfPermission(context, Manifest.permission.ACCESS_COARSE_LOCATION) == PackageManager.PERMISSION_GRANTED)) {
+                mLocationManager.requestLocationUpdates(provider, 0, 0, PollSensorAction.this);
```
ctsims: Does this mean we're just skipping the use of the background provider unless someone already accepted GPS permissions instead of asking users to do them?

app/src/org/commcare/dalvik/activities/LoginActivity.java
```diff
@@ -70,12 +81,24 @@
...
  */
 @ManagedUi(R.layout.screen_login)
-public class LoginActivity extends CommCareActivity<LoginActivity> implements OnItemSelectedListener {
+public class LoginActivity extends CommCareActivity<LoginActivity>
+        implements OnItemSelectedListener, RuntimePermissionRequester {
 
     private static final String TAG = LoginActivity.class.getSimpleName();
-    
+
+    private final static String[] appPermissions =
```
ctsims: Hm, are we not collecting these from something like the RuntimePermissionRequester implementing classes, for a reason? In theory we could add it as an annotation and reflect them all back out, which would prevent us from having to add them in a number of places. It would also let us collapse the logic into a unified "hasPermission()" method that grabbed the permissions from that list.

phillipm: Some activities need different permissions given different states (like https://github.com/dimagi/commcare-odk/pull/818/files#diff-3e9cdafcf0cfda6187cf54c80c8115deR172). That type of permission request dispatch wouldn't be possible with annotations, afaik.

app/src/org/odk/collect/android/jr/extensions/PollSensorAction.java
phillipm: I couldn't find a way to get a hold of an activity, which is needed to show the user the permission request dialog. 
I guess I could do some sort of check when form entry is launched to see if it collects geo data and ask there.
Any suggestions?

PR #820, opened 2015-11-16
Stop storing activities in StateFragment

phillipm:
Trying to sort out the memory leaks we are experiencing:
This PR moves us away from storing old activities in persistent fragments, which can often lead to memory leaks if you aren't careful.

Instead I have introduced the `ContainerFragment` which enables us to explicitly choose what we want to persist across activity configuration changes.

Since we no longer have access to the previous activity, I've started storing the updated state of `@UiElement` fields in the bundle that is passed into `CommCareActivity.onCreate` by way of `onSaveInstanceState`. This is a slight lose in terms of conciseness, but hopefully it will enable us to make progress on the memory leak issues. 

Also fixes app manager multimedia validation bug where rotating screen after initial validation will trigger another validation task to fire off.


app/src/org/commcare/android/framework/CommCareActivity.java
```diff
@@ -238,24 +228,6 @@ protected void loadFields(boolean restoreOldFields) {
...
-                                if (v == null) {
-                                    Log.d("loadFields", "NullPointerException when trying to find view with id: " +
-                                            +element.value() + " (" + getResources().getResourceEntryName(element.value()) + ") "
-                                            + " oldView is " + oldView + " for activity " + oldActivity +
-                                            ", element is: " + f + " (" + f.getName() + ")");
-                                }
-                                v.setVisibility(oldView.getVisibility());
-                                v.setEnabled(oldView.isEnabled());
-                                continue;
-                            }
```
phillipm: I can't figure out when this is needed? @ctsims can you provide context?

ctsims: This restores the state of fields from the previous state. IE: If you rotate, this ensures that if you've typed text into a field, changed it from visible to GONE, or disabled it, that it will be in the same state in the rotated view.

phillipm: @ctsims Ah, got it, thanks! I've updated the PR to persist this data via Bundles that are persisted throughout the activity lifecycle

app/src/org/commcare/dalvik/activities/RecoveryActivity.java
```diff
@@ -53,10 +53,8 @@
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         
-        if(this.getDestroyedActivityState() != null) {
-            //We just rotated or whatever, don't re-initialize everything
-        } else {
-            //Fresh Start, statewise.
+        if (savedInstanceState == null) {
```
ctsims: Just as a heads up, savedInstanceState is not a reliable mechanism and its use is discouraged, so I wouldn't generally use it as a flag for whether this is restarting.

ctsims:
@phillipm This all looks fine other than the reliance on onSaveInstanceState. The android docs say

"Note that it is important to save persistent data in onPause() instead of onSaveInstanceState(Bundle) because the latter is not part of the lifecycle callbacks, so will not be called in every situation as described in its documentation."

which has been very true in my experience. We should shift our storage/recovery to use an onPause mechanism rather than onSaveInstanceState if you are going to serialize in the mechanism you're using.

app/assets/locales/messages_ccodk_default.txt
```diff
@@ -28,6 +28,9 @@ install.current.version=Version currently installed: ${0}
 verify.title=CommCare Resource Verification
 verify.checking=Verifying Resources
 verify.progress=Verifying resources. ${0} resources verified of ${1} total
+verify.retry=Retry
+verify.check.message=Verifying media...
```
ctsims: I'm confused as to why we added new translations here instead of updating the text of the current ones

app/src/org/commcare/android/framework/CommCareActivity.java
```diff
@@ -141,6 +136,37 @@ protected void onCreate(Bundle savedInstanceState) {
...
+            if (savedInstanceState != null) {
+                ManagedUiFramework.restoreUiElements(this, savedInstanceState);
+            } else {
+                ManagedUiFramework.loadUiElements(this);
+            }
+        }
+    }
+
+    @Override
+    protected void onSaveInstanceState(Bundle outState) {
```
ctsims: no-op

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -206,7 +207,7 @@ protected void onCreate(Bundle savedInstanceState) {
                 setContentView(R.layout.entity_select_layout);
                 //So we're not in landscape mode anymore, but were before. If we had something selected, we 
                 //need to go to the detail screen instead.
-                if (oldActivity != null) {
+                if (savedInstanceState != null) {
```
ctsims: Should this check savedInstanceState or check for the fragment

PR #821, opened 2015-11-17
Don't start update activity using startActivityForResult

phillipm:
The home activity does not need to read any result data from the update activity so use `startActivity` instead of `startActivityForResult`


app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -386,17 +384,6 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
                     return;
                 }
                 break;
-            case UPGRADE_APP:
-                if (resultCode == RESULT_CANCELED) {
-                    return;
-                } else if(resultCode == RESULT_OK) {
```
amstone326: @phillipm I was actually wondering this and then forgot to ask- why do we no longer need to make the user logout after an update?

phillipm: I can't remember. @ctsims / @wpride do either of you foresee any issues coming up if we apply an update without logging out and back in?

amstone326: Worth noting that this has effectively already been removed because KEY_REQUIRE_REFRESH is no longer being sent, and there don't seem to be any problems coming up with updates, but worth understanding why it's ok.

ctsims: @pmates This is definitely capable of causing issues. After an app update
many, many things could have changed about the structure of the app. Home
screen buttons, text localizations, fixture data. I think home screen
translations are likely the most easy-to-replicate thing that would come
up, but don't remember whether there were buggier/more mission critical
components.

-Clayton

amstone326:
@phillipm Can we close this? Sounds like the logout logic needs to get added back in, which means this setup would stay

phillipm:
No, this PR is still valid. In a future PR I will logout directly from the update screen when the user chooses to perform the update installation. That action will also not need to use `startActivityForResult`

amstone326:
Ok, will merge then. Sounds like we should definitely get that other PR into 2.25 though.

PR #822, opened 2015-11-17
C3

orangejenny:
Replace AChartEngine with C3 (http://c3js.org/) for mobile graphing.

Apologies for length. Should be straightforward to review commit-by-commit.

This is set to merge into `graphing` because there are a few remaining issues to fix before merging to master:
- [ ] annotations displaying incorrectly
- [ ] time-based graphs don't work
- [ ] point-style not supported
- [ ] ability to highlight a single bar in bar graphs (new feature requested by TDH)
- [ ] non-full-screen graphs don't display titles (this one might not be possible, might just deprecate it)
- [ ] translucent colors not supported (might deprecate)
- [ ] need to implement a full-screen view
- [ ] performance isn't great, need to at least add a spinner indicator and ideally make performance better
(and anything else marked as TODO in this PR should be addressed in 2.25)


app/src/org/commcare/android/view/c3/Configuration.java
```diff
@@ -0,0 +1,92 @@
...
+ * the bulk of the work is done in the constructor. The instantiator
+ * can then call getConfiguration and getVariables to get at the JSON
+ * configuration and any JavaScript variables that configuration depends on.
+ *
+ * Created by jschweers on 11/16/2015.
+ */
+public class Configuration {
+    protected GraphData mData;
+    protected JSONObject mConfiguration;
+    protected OrderedHashtable<String, String> mVariables;
```
phillipm: These can be marked 'final'

app/src/org/commcare/android/view/c3/AxisConfiguration.java
```diff
@@ -0,0 +1,177 @@
...
+ */
+public class AxisConfiguration extends Configuration {
+    public AxisConfiguration(GraphData data) throws JSONException, InvalidStateException {
+        super(data);
+
+        JSONObject x = new JSONObject();
+        JSONObject y = new JSONObject();
+        JSONObject y2 = new JSONObject();
+
+        if (Boolean.valueOf(mData.getConfiguration("show-axes", "true")).equals(Boolean.FALSE)) {
```
phillipm: `(...).equals(Boolean.FALSE)` -> `!(...)`

phillipm: Might be nice to name this conditional test expression for readability:
`final boolean shouldShowAxes = Boolean.valueOf(mData.getConfiguration("show-axes", "true");`

app/src/org/commcare/android/view/c3/AxisConfiguration.java
```diff
@@ -0,0 +1,177 @@
...
+            JSONObject show = new JSONObject("{ show: false }");
+            x = show;
+            y = show;
+            y2 = show;
+        } else {
+            // Undo C3's automatic axis padding
+            JSONObject padding = new JSONObject("{top: 0, right: 0, bottom: 0, left: 0}");
```
phillipm: I would lift all this into a helper function called `addAxisToGraph(JSONObject x, JSONObject y, JSONObject y2)` or something

app/src/org/commcare/android/view/c3/AxisConfiguration.java
```diff
@@ -0,0 +1,177 @@
...
+ *
+ * Created by jschweers on 11/16/2015.
+ */
+public class AxisConfiguration extends Configuration {
+    public AxisConfiguration(GraphData data) throws JSONException, InvalidStateException {
+        super(data);
+
+        JSONObject x = new JSONObject();
+        JSONObject y = new JSONObject();
+        JSONObject y2 = new JSONObject();
```
phillipm: What is a secondary  y axis?

app/src/org/commcare/android/view/c3/Configuration.java
```diff
@@ -0,0 +1,92 @@
...
+     *
+     * @param description Something to identify the kind of value, used to augment any error message.
+     */
+    protected double parseDouble(String value, String description) throws InvalidStateException {
+        try {
+            Double numeric = Double.valueOf(value);
+            if (numeric.isNaN()) {
+                throw new InvalidStateException("Could not understand '" + value + "' in " + description);
+            }
+            return numeric.doubleValue();
```
phillipm: Lint warning about unneed to unbox, since unboxing is done automatically in new versions of Java

app/src/org/commcare/android/view/c3/DataConfiguration.java
```diff
@@ -0,0 +1,349 @@
...
+
+    // Hash of y-values id => 'y' or 'y2' depending on whether this data
+    // should be plotted against the primary or secondary y axis
+    private JSONObject mAxes = new JSONObject();
+
+    // Hash of y-values id => line, scatter, bar, area, etc.
+    JSONObject mTypes = new JSONObject();
+
+    // Hash of y-values id => series color
+    JSONObject mColors = new JSONObject();
```
phillipm: Can be private

phillipm: All of these class fields can be made `final`

app/src/org/commcare/android/view/c3/DataConfiguration.java
```diff
@@ -0,0 +1,349 @@
...
+
+    public DataConfiguration(GraphData data) throws JSONException, InvalidStateException {
```
phillipm: There are a lot of checks for the graph type in this constructor. Might be worth considering creating sub-classes for each type of graph

phillipm: Would recommend breaking up this 150 line constructor into various helper functions

app/src/org/commcare/android/view/c3/DataConfiguration.java
```diff
@@ -0,0 +1,349 @@
...
+        // C3 doesn't support bubble charts well, so radius data gets set up
+        // in separate variables rather than as part of the configuration
+        mVariables.put("radii", radii.toString());
+        mVariables.put("maxRadii", maxRadii.toString());
+    }
+
+    /**
+     * Add annotations, by creating a fake series with data labels turned on.
+     * @throws JSONException
+     * @throws InvalidStateException
```
phillipm: `@throws` annotations that don't include descriptions don't offer additional info over the method signature so it would be nice to remove them

app/src/org/commcare/android/view/c3/DataConfiguration.java
```diff
@@ -0,0 +1,349 @@
...
+        return false;
+    }
+
+    /**
+     * Comparator to sort XYPointData-derived objects by x value without parsing them.
+     * Useful for bar graphs, where x values are text.
+     *
+     * @author jschweers
+     */
+    private class StringPointComparator implements Comparator<XYPointData> {
```
phillipm: private classes should be made static whenever possible to prevent the automatic creation of a pointer to the encapsulating class, which sometimes leaks memory

app/src/org/commcare/android/view/c3/AxisConfiguration.java
orangejenny: An axis displayed on the right, with a different scale than the left/primary y axis

temp

app/src/org/commcare/android/view/c3/DataConfiguration.java
orangejenny: I'd rather not split into subclasses because the future direction of graphing is to mix different types of series in the same graph (admittedly, there's other code in this PR that expects type to be graph-level, but I'd prefer not to make even more). Agree that this is messy, though, will clean it up.

phillipm:
Left comments regarding code organization; not sure how much of it you want to follow up on, let me know when your done following up and I'll merge

orangejenny:
@phillipm Added some cleanup commits. DataConfiguration is still a little beastly, but I think it's as good as it's going to get right now.

PR #823, opened 2015-11-17
Check that session hasn't changed after callout

amstone326:
Clear the current session and return to home screen if the session's needed datum has changed (including being null) when we return from a session-related callout. Provides better handling of this crash: https://commcare-acralyzer.cloudant.com/acralyzer/_design/acralyzer/index.html#/reports-browser/commcare-odk/bug/b08b624100441f02ede2ab35fb0e1bdc

 Also add a popup dialog telling the user what has happened when the session is cleared.


app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -864,7 +883,8 @@ private void handleGetCommand(AndroidSessionWrapper asw) {
         } else {
             i = new Intent(getApplicationContext(), MenuList.class);
         }
-        i.putExtra(SessionFrame.STATE_COMMAND_ID, asw.getSession().getCommand());
+        i.putExtra(SessionFrame.STATE_COMMAND_ID, command);
+        addPendingDatumIdExtra(i, asw.getSession());
```
phillipm: causes and NPE if you select a module that doesn't use cases

amstone326:
@phillipm Fixed, thanks for catching. @ctsims Should probably look at this before merging though.

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -565,6 +579,19 @@ else if(resultCode == RESULT_OK){
...
+     * @return If the nature of the data that the session is waiting for has not changed since the
+     * callout that we are returning from was made
+     */
+    private boolean sessionStateUnchangedSinceCallout(CommCareSession session, Intent intent) {
+        boolean dataIdsEqual = session.getNeededData().equals(
+                intent.getStringExtra(KEY_PENDING_SESSION_DATA));
+        String intentDatum = intent.getStringExtra(KEY_PENDING_SESSION_DATUM_ID);
+        boolean datumIdsEqual = intentDatum  == null ? true :
+                intentDatum.equals(session.getNeededDatum().getDataId());
+        return dataIdsEqual && datumIdsEqual;
```
ctsims: These seem very close, it took me two reads to realize they weren't the same value

ctsims:
@amstone326 one minor change then good to go

amstone326:
@ctsims fixed

PR #824, opened 2015-11-18
Point style

orangejenny:
The bad thing about C3 is that to do something custom, I'm reduced to searching for C3-generated class names and then manipulating the SVG directly.

The good thing about C3 is that it's actively maintained and has something of a community, and there happens to already be an open issue for custom point shapes: https://github.com/masayuki0812/c3/issues/1071

screenshot_2015-11-18-14-31-27

### DIFF
--- a/app/assets/graphing/graph.js
+++ b/app/assets/graphing/graph.js
@@ -13,11 +13,11 @@ document.addEventListener("DOMContentLoaded", function(event) {
     // Set point size for bubble charts, and turn points off altogether
     // for other charts (we'll be using custom point shapes)
     config.point = {
-    	r: function(d) {
-    	    if (radii[d.id]) {
-    	        // Arbitrary max size of 30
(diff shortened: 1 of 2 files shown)


ppKrauss:
Hi, where the point style at https://c3js.org/reference.html  ?

orangejenny:
@ppKrauss C3 doesn't support point style, that's why this pull request was necessary.

ppKrauss:
Hi @orangejenny, sorry, I added a new [comment at the correct place](https://github.com/c3js/c3/issues/1071#issuecomment-510871991). Well, about your initiative, why your merge was stopped?

orangejenny:
@ppKrauss Ah, I see. Are you asking why didn't I submit this code as a PR to C3? Probably just lack of time. You're welcome to use this code.

ppKrauss:
2019-2015=**4 years** (!). This is not "lack of time", seems that C3js developers are abondoning it...

PR #829, opened 2015-11-18
Switch image inflation setting to being enabled from HQ

amstone326:
Switch new image inflation setting to being enabled from HQ instead of Developer Preferences


amstone326:
Release Note: New image system that keeps the sizes of user-provided images in CommCare consistent with the original file size, and consistent across devices.

(Assuming the HQ component of this gets merged before 2.25, which it definitely should).

PR #831, opened 2015-11-19
Correctly scale images (via vectors) to reduce app memory footprint

phillipm:
Reduces home screen heap footprint from ~80mb to ~30mb

Using vector graphics (.svg) is supported on API 21 and up. For those devices switching into landscape mode doesn't change memory usage.

On devices below API 21, gradle creates png versions of the vector graphics. In this case, going into landscape mode will bump memory consumption because the device is showing a larger density image. For instance, on a 4.2.2 Galaxy Nexus, memory consumption goes from 30 to 50 mb on the home screen when switching from portrait to landscape mode. 

Importing SVGs into Android Studio is extremely finicky, you pretty much have to use Adobe Illustrator (inkscape creates more complex svg code that AS can't read).

Since Android Studio creates xml versions of the SVGs imported, I have also left the originals in the `app/original_vector_graphics` directory

New home screen should be pretty much identical:
screen

__Release notes:__ Reduce app's memory consumption


orangejenny:
> Reduces home screen heap footprint from ~80mb to ~30mb

awesome

PR #834, opened 2015-11-19
Save custom data to session

wpride:
We need something like this to enable us to use case list callouts with double management specifically WRT biometrics. This lets us load the templates into the forms after not matching and wanting to register a new user with those strings


wpride:
cross https://github.com/dimagi/commcare/pull/196

phillipm:
can we close this?

PR #835, opened 2015-11-20
support fill-below and opacity

orangejenny:
C3 defaults area below a line to the same color as the line, and it doesn't support user-configured opacity for either line colors or area colors. Deal with this.

screenshot_2015-11-20-12-35-27


phillipm:
Nice! These graphs are pretty!

PR #837, opened 2015-11-21
full screen graphs

orangejenny:
Support double-tap to see the graph full screen.

In full-screen mode, you can zoom and can touch a point to get a tooltip. This PR adds to series configuration `is-data` (default true) so users can specify that a given series should *not* be included in the tooltip (or the legend) (for example, the fake "series" used to create parto's red and yellow areas, or thresholds in general).

It's a bit finicky; the touch to tooltip isn't always responsive (it's a hover event in C3, so Android is translating the touch event to a mouseover). I'm using the built-in WebView zoom because C3's zoom capabilities are "experimental" (do not work well), but unfortunately that means that you're zooming in on the entire view, not just zooming in on the data while the axes stay put & re-scale themselves.

screenshot_2015-11-20-19-09-47
screenshot_2015-11-20-19-09-55

### DIFF
--- a/app/src/org/commcare/dalvik/activities/GraphActivity.java
+++ b/app/src/org/commcare/dalvik/activities/GraphActivity.java
@@ -1,0 +1,45 @@
+package org.commcare.dalvik.activities;
+
+import android.annotation.TargetApi;
+import android.os.Build;
+import android.os.Bundle;
+import android.view.View;
+import android.view.ViewGroup;
+import android.view.Window;
+import android.webkit.WebView;
+import android.widget.LinearLayout;
+
+import org.commcare.android.framework.CommCareActivity;
+import org.commcare.android.util.InvalidStateException;
+import org.commcare.android.view.GraphView;
+import org.commcare.suite.model.graph.GraphData;
+
+/**
+ * Created by jschweers on 11/20/2015.
+ */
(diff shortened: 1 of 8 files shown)


app/src/org/commcare/dalvik/activities/GraphActivity.java
```diff
@@ -0,0 +1,47 @@
...
+            setTitle(title);
+        }
+
+        String html = extras.getString(GraphView.HTML);
+        mView = new GraphView(this, title, true);
+        try {
+            WebView view = (WebView) mView.getView(html);
+            setContentView(view);
+        } catch (InvalidStateException e) {
+            e.printStackTrace();
```
phillipm: Hmm, not sure, but it might be good to call `finish()` here, otherwise I suspect it just hang with a blank screen.

phillipm:
Looks good, will merge when conflicts are handled

orangejenny:
Has a minor dependency on https://github.com/dimagi/commcare/pull/197

app/src/org/commcare/dalvik/activities/GraphActivity.java
orangejenny: good call

orangejenny:
i was bound to step on my own toes eventually. merge conflict fixed.

PR #839, opened 2015-11-22
Don't trigger triggerables on constraint check

phillipm:
Don't evaluation triggerables when checking question constraints during form saving. Hide this change under a developer option, which can also be programatically enabled via the profile file by setting `cc-fire-triggers-on-save` to `yes` or `no`

Improves save performance of TDH "Child Classification" form by 11 seconds, from 12 down to  below 1 second.

cross-request with https://github.com/dimagi/javarosa/pull/196


app/src/org/odk/collect/android/tasks/SaveToDiskTask.java
```diff
@@ -365,20 +371,19 @@ public void setFormSavedListener(FormSavedListener fsl) {
...
-     * Goes through the entire form to make sure all entered answers comply with their constraints.
-     * Constraints are ignored on 'jump to', so answers can be outside of constraints. We don't
-     * allow saving to disk, though, until all answers conform to their constraints/requirements.
+     * Goes through the entire form to make sure all entered answers comply
+     * with their constraints.  Constraints are ignored on 'jump to', so
+     * answers can be outside of constraints. We don't allow saving to disk,
+     * though, until all answers conform to their constraints/requirements.
      */
-    private int validateAnswers(Boolean markCompleted) {
+    private int validateAnswersAndRefireTriggers(Boolean markCompleted) {
```
ctsims: These two methods share their entire structure other than 

answerQuestion -> checkCurrentQuestionConstraint

right? Seems unfortunate...

phillipm:
@ctsims updated PR to unify the two similar methods

PR #840, opened 2015-11-22
Standard aspect ratio for graphs

orangejenny:
AChartEngine displayed bar graphs squished unless they were square, so the old graphing code based aspect ratio on graph type. Drop this hackiness, use the same aspect ratio everywhere.

screenshot_2015-11-21-21-16-21


phillipm:
Good to go once merge conflicts are resolved

orangejenny:
mergeable

PR #841, opened 2015-11-22
Time graphs

orangejenny:
Support graphs that use datetimes for x values.

screenshot_2015-11-21-21-29-12

### DIFF
--- a/app/src/org/commcare/android/view/c3/Configuration.java
+++ b/app/src/org/commcare/android/view/c3/Configuration.java
@@ -73,24 +73,6 @@ public class Configuration {
     }
 
     /**
-     * Parse given string into double
-     *
-     * @param description Something to identify the kind of value, used to augment any error message.
-     */
-    protected double parseYValue(String value, String description) throws InvalidStateException {
-        return parseDouble(value, description);
-    }
-
-    /**
-     * Parse given string into double
-     *
-     * @param description Something to identify the kind of value, used to augment any error message.
-     */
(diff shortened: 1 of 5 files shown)


app/src/org/commcare/android/view/c3/Configuration.java
```diff
@@ -37,40 +38,14 @@ public JSONObject getConfiguration() {
...
-    /**
-     * Parse given string into double
-     *
-     * @param description Something to identify the kind of value, used to augment any error message.
+     * @param days The time, measured in days since the epoch.
      */
-    protected double parseRadiusValue(String value, String description) throws InvalidStateException {
-        return parseDouble(value, description);
+    protected String convertTime(double days) {
+        Date d = new Date((long)(days * 24 * 60 * 60 * 1000));
```
phillipm: Use `DateUtils.DAY_IN_MS` instead of `24 * 60 * 60 * 1000`
Also consider renaming `days` to `daysSinceEpoch`

orangejenny: nice, didn't know we had that constant lying around (or that it's actually `24 * 60 * 60 * 1000 + 1`...?)

phillipm: Oh, you scared me with that `+ 1` business! It is actually an `l` for `long`.

ctsims:
@orangejenny

Awesome work on these! The difference is subtle, but makes a major
improvement!

-Clayton

PR #850, opened 2015-11-24
Remove map lib dependency

phillipm:
Hopefully this allows users who don't have mapping android libs to install our app.

http://manage.dimagi.com/default.asp?182975


ctsims:
@phillipm Can you just wire this up to use required:false instead?

http://developer.android.com/guide/topics/manifest/uses-library-element.html#rq

I think we still need to declare the need

phillipm:
@ctsims updated

PR #853, opened 2015-11-24
Support for bar-color

orangejenny:
See https://github.com/dimagi/commcare/pull/202

The legend in the screenshot looks wonky because `line-color` wasn't set - because `bar-color` is point-specific, not series-specific, I don't know what color to draw the legend unless `line-color` is also specified, and C3 doesn't know anything about the bar-specific colors, so it gives the legend an arbitrary color.

screenshot_2015-11-24-18-39-44

### DIFF
--- a/app/assets/graphing/graph.js
+++ b/app/assets/graphing/graph.js
@@ -120,7 +120,19 @@ document.addEventListener("DOMContentLoaded", function(event) {
                 if (type === "bar") {
                     var bars = d3.selectAll(".c3-bars-" + yID + " path")[0];
                     for (var i = 0; i < bars.length; i++) {
-                        bars[i].style.opacity = lineOpacities[yID];
+                        // If there's a bar-specific color, set it
+                        if (barColors[yID] && barColors[yID][i]) {
+                            bars[i].style.fill = barColors[yID][i];
+                        }
+                        // Get opacity: bar-specific if it's there, otherwise series-specific
+                        var opacity;
+                        var debug = "from lineOpacities";
+                        if (barOpacities[yID]) {
+                            opacity = barOpacities[yID][i];
+                            debug = "from barOpacities";
+                        }
+                        opacity = opacity || lineOpacities[yID];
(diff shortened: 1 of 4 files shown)


app/assets/graphing/graph.js
```diff
@@ -63,7 +63,7 @@ document.addEventListener("DOMContentLoaded", function(event) {
 
                 }
                 var label = xLabels[key] || d;
-                return type === "time" ? label : Math.round(label);
+                return Math.round(label) || label;
```
orangejenny: unrelated bad merge was causing bar labels to display as "NaN"

app/src/org/commcare/dalvik/activities/EntityDetailActivity.java
```diff
@@ -97,8 +97,6 @@ protected void onCreate(Bundle savedInstanceState) {
             this.mEntityContext = new Pair<Detail, TreeReference>(shortDetail, mTreeReference);
         }
 
-        Entity<TreeReference> entity = factory.getEntity(mTreeReference);
```
orangejenny: random unused variable, but a somewhat expensive one to be generating unnecessarily

app/assets/graphing/graph.js
```diff
@@ -120,7 +120,19 @@ document.addEventListener("DOMContentLoaded", function(event) {
                 if (type === "bar") {
                     var bars = d3.selectAll(".c3-bars-" + yID + " path")[0];
                     for (var i = 0; i < bars.length; i++) {
-                        bars[i].style.opacity = lineOpacities[yID];
+                        // If there's a bar-specific color, set it
+                        if (barColors[yID] && barColors[yID][i]) {
+                            bars[i].style.fill = barColors[yID][i];
```
orangejenny: C3 actually supports bar-specific colors, but it's easier to just handle the colors here along with the opacities (which are not supported at all by C3).

phillipm:
Good to go once comments in https://github.com/dimagi/commcare/pull/202 are addressed

PR #857, opened 2015-11-25
Clarify End of Form Navigation Behavior

ctsims:
When  users swipe right on a page with a green nav arrow, the app now gives them a clear indication that the next step has to involve clicking the finish button


app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2193,14 +2193,30 @@ private void finishReturnInstance(boolean reportSaved) {
...
     }
 
+    protected boolean isBlockingUserInput() {
+        View cover = this.findViewById(R.id.form_entry_cover);
+
+        if(cover != null && cover.getVisibility() == View.VISIBLE) {
+            return true;
+        } else {
+            return false;
+        }
```
phillipm: change to `return cover != null && cover.getVisibility() == View.VISIBLE`

phillipm:
Can you center the animation? It looks strange that it expands rightwards

ctsims:
@phillipm Fixed your code comment.

Unfortunately we can't really center the animation. It's centered in the View right now, which includes padding, which we want for the click affordance, but makes it difficult to have it expand only the internal drawable.

PR #858, opened 2015-11-25
Remove deprecated grid view

phillipm:
Stop using https://github.com/dimagi/AndroidStaggeredGrid, which has been deprecated. Instead use a view holder pattern with a RecyclerView and StaggeredGridLayoutManager.

Saves a few megabytes of heap space for apps that don't show 'incomplete' or 'report' buttons by not inflating their images.

This PR is pretty big, let me know if you want to walk through the changes with me.

Before:

After (6.0 device):
screen

After (4.2.2 device):
screen


app/res/layout/grid_header_top_banner.xml
```diff
@@ -1,20 +1,21 @@
...
-        android:layout_marginTop="@dimen/activity_horizontal_margin"
         android:layout_marginBottom="@dimen/activity_horizontal_margin"
+        android:layout_marginTop="@dimen/activity_horizontal_margin"
+        android:adjustViewBounds="true"
         android:paddingLeft="@dimen/standard_spacer"
-        android:paddingRight="@dimen/standard_spacer"/>
+        android:paddingRight="@dimen/standard_spacer"
+        android:scaleType="fitCenter"
+        android:src="@drawable/commcare_logo"
+        android:visibility="visible"/>
```
phillipm: Changes here from code formatter

phillipm:
@biyeun While I was migrating away from some deprecated view libraries I slightly modified the home screen UI: slight corner rounding, different drop shadow, different spacing between buttons. Will you quickly look over the "After" image in the PR description and let me know if there is anything I should tweak? Thanks

biyeun:
chatted with phillip offline and this looks great :+1:

ctsims:
@biyeun @phillipm New shadows look nice, more subtle. 

What's with the white borders introduced in 4.2.2+? Is that intentional? They look jarringly different

phillipm:
@ctsims The 4.2.2 borders are probably a compat artifact due to the use of either rounded corners or drop shadows. I'll look into fixing it.

app/src/org/commcare/android/adapters/HomeScreenAdapter.java
```diff
@@ -1,150 +1,209 @@
...
+        StateListDrawable bgDrawable = getSLD(context, cardDisplayData.bgColor);
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) {
+            squareButtonViewHolder.cardView.setBackground(bgDrawable);
+        } else {
+            squareButtonViewHolder.cardView.setBackgroundDrawable(bgDrawable);
+        }
+    }
+
+    private static StateListDrawable getSLD(Context context,
```
amstone326: worth a comment on what this method is doing

app/src/org/commcare/android/adapters/HomeScreenAdapter.java
```diff
@@ -1,150 +1,209 @@
...
-        hiddenButtons[index] = isButtonHidden;
-        if (hasVisibilityChanged) {
-            if (isButtonHidden) {
-                visibleButtons.remove(buttons[index]);
-            } else {
-                visibleButtons.add(index, buttons[index]);
-            }
-        }
+    public int getSyncButtonPosition() {
+        // NOTE PLM: assumes sync button is always the second to last button.
```
amstone326: Why is this safe to assume?

app/src/org/commcare/dalvik/activities/LoginActivity.java
```diff
@@ -96,9 +97,6 @@
     
     @UiElement(R.id.screen_login_banner_pane)
     private View banner;
-    
-    @UiElement(R.id.str_version)
-    private TextView versionDisplay;
```
amstone326: Why is this getting removed? It looks like it doesn't currently show up on the login screen anyway, but do we want to fix that?

app/src/org/commcare/android/adapters/HomeScreenAdapter.java
```diff
@@ -1,150 +1,209 @@
...
-    public int getCount() {
-        return visibleButtons.size();
+    private HomeCardDisplayData getItem(int position) {
+        return buttonData[position - 1];
     }
 
-    @Override
-    public Object getItem(int position) {
-        return buttons[position];
+    private static String getLastPayloadString(List<Object> payload) {
```
amstone326: Can you add a comment for what this is doing?

app/src/org/commcare/dalvik/activities/HomeButtons.java
```diff
@@ -0,0 +1,212 @@
...
+
+    private static View.OnClickListener getSyncButtonListener(final CommCareHomeActivity activity) {
+        return new View.OnClickListener() {
+            public void onClick(View v) {
+                activity.syncButtonPressed();
+            }
+        };
+    }
+
+    private static TextSetter getSyncButtonNotificationText(final CommCareHomeActivity activity) {
```
amstone326: I'm finding the names of these 3 methods (this and the other 2 get____NotificationText() methods) confusing, given that they return a TextSetter for all of the text for 1 home card, but it sounds like it would just be returning a String (the notification text).

app/res/layout/home_screen.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
amstone326: Not really a big deal and definitely not worth changing if it proves at all annoying, but is there a reason this has to be 4 levels of nested layouts, when only the RecyclerView actually has content?

app/src/org/commcare/android/adapters/HomeScreenAdapter.java
```diff
@@ -1,150 +1,209 @@
...
+    private void bindCard(SquareButtonViewHolder squareButtonViewHolder,
+                          int i, List<Object> payload) {
+        HomeCardDisplayData cardDisplayData = getItem(i);
+        String notificationText = null;
+
+        if (payload != null) {
+            notificationText = getLastPayloadString(payload);
         }
+
+        cardDisplayData.textSetter.update(cardDisplayData,
```
amstone326: Very minor, but any reason the code for setting the view holder's text isn't just part of the setupViewHolder() method? Seems like it would logically be a part of that.

app/src/org/commcare/android/adapters/SquareButtonViewHolder.java
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * @author Phillip Mates (pmates@dimagi.com).
+ */
+public class SquareButtonViewHolder extends RecyclerView.ViewHolder {
+    public final SquareImageView imageView;
+    public final RelativeLayout cardView;
+    public final TextView textView;
+    public final TextView subTextView;
+
+    public SquareButtonViewHolder(View view) {
```
amstone326: It's kind of an odd pattern to have the constructor take in a View parameter, when you are always explicitly expecting it to be 1 very specific layout file (home_card.xml). Can you just make the constructor take no params and inflate that view within the class? Same goes for the HeaderItemView constructor.

app/src/org/commcare/android/view/SquareImageView.java
```diff
@@ -0,0 +1,37 @@
...
+    public SquareImageView(Context context, AttributeSet attrs) {
+        super(context, attrs);
+    }
+
+    public SquareImageView(Context context, AttributeSet attrs, int defStyleAttr) {
+        super(context, attrs, defStyleAttr);
+    }
+
+    @Override
+    protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
```
amstone326: can you add a comment as to the logic being employed in this method?

amstone326:
OK finished a pass on all of this. Awesome refactor, really well done and learned a bunch from looking over it. Left some actionable comments, but nothing major. Ready to merge once addressed.

app/src/org/commcare/dalvik/activities/LoginActivity.java
phillipm: This info can be accessed via the login screen by selecting the "About CommCare" menu entry. Pretty sure this is just debris from the old ui.

app/src/org/commcare/android/adapters/SquareButtonViewHolder.java
phillipm: I think this is the standard pattern. ViewHolders hold info and shouldn't do anything else, so I think view inflation shouldn't occur in this class. Plus the `parent` param from the adapter's `onCreateViewHolder` _might_ be important

app/src/org/commcare/android/view/SquareImageView.java
phillipm: Copied this code from `SquareButton`, don't exactly understand the scaling logic. Will defer to @ctsims.

app/src/org/commcare/dalvik/activities/HomeButtons.java
phillipm: good call!

app/res/layout/home_screen.xml
phillipm: Removed them, thanks for the catch!

phillipm:
PR comments addressed, ready to be looked at again/merged

app/src/org/commcare/android/view/SquareImageView.java
amstone326: Ah ok didn't realize, seems fine then.

app/src/org/commcare/android/adapters/SquareButtonViewHolder.java
amstone326: Got it, makes sense.

PR #864, opened 2015-11-27
Fix app update progress bar artifact

phillipm:
Fix issue where after performing an update check the progress bar shows partially complete:
screen

Also rename dev setting entry "Enable auto-login during debugging" -> "Auto-login While Debugging"

### DIFF
--- a/app/src/org/commcare/dalvik/activities/UpdateUIState.java
+++ b/app/src/org/commcare/dalvik/activities/UpdateUIState.java
@@ -120,7 +120,6 @@ class UpdateUIState {
         stopUpdateButton.setEnabled(false);
         installUpdateButton.setEnabled(false);
 
-        progressBar.setEnabled(false);
         updateProgressText("");
         updateProgressBar(0, 100);
     }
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/dalvik/activities/UpdateUIState.java
```diff
@@ -194,8 +188,8 @@ protected void updateProgressText(String msg) {
     }
 
     protected void updateProgressBar(int currentProgress, int max) {
-        progressBar.setProgress(currentProgress);
         progressBar.setMax(max);
+        progressBar.setProgress(currentProgress);
```
phillipm: The fix: switching the order of these set method calls. That was a fun one to track down!

app/src/org/commcare/dalvik/activities/UpdateUIState.java
```diff
@@ -165,7 +162,6 @@ protected void cancellingUiState() {
         stopUpdateButton.setEnabled(false);
         installUpdateButton.setEnabled(false);
 
-        progressBar.setEnabled(false);
```
phillipm: These `setEnabled` calls aren't really needed

PR #867, opened 2015-11-27
Fixes case searching.

josejulio:
Was only searching for first *searchTerm* on the last field.

Although I'm not sure how the FuzzySearch works. Should the score be accumulated across fields and searchTerms or is OK as is?

### DIFF
--- a/app/src/org/commcare/android/adapters/EntityListAdapter.java
+++ b/app/src/org/commcare/android/adapters/EntityListAdapter.java
@@ -254,7 +254,7 @@ public class EntityListAdapter implements ListAdapter {
                         String field = e.getNormalizedField(i);
                         if (field != "" && field.toLowerCase(currentLocale).contains(filter)) {
                             add = true;
-                            continue filter;
+                            break filter;
                         } else {
                             // We possibly now want to test for edit distance for
                             // fuzzy matching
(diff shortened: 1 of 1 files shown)


damagatchi:

app/src/org/commcare/android/adapters/EntityListAdapter.java
```diff
@@ -254,7 +254,7 @@ private void search() {
                         String field = e.getNormalizedField(i);
                         if (field != "" && field.toLowerCase(currentLocale).contains(filter)) {
                             add = true;
-                            continue filter;
```
josejulio: IF continuing, the *add* variable would be set to false, if the next search term doesn't match, it will end up here https://github.com/dimagi/commcare-odk/pull/867/files#diff-2a0c2809914e40cd56d9b605f3d6c0c5L273

app/src/org/commcare/android/adapters/EntityListAdapter.java
```diff
@@ -264,15 +264,12 @@ private void search() {
...
                                         add = true;
                                         score += match.second;
-                                        continue filter;
+                                        break filter;
                                     }
                                 }
                             }
                         }
                     }
-                    if (!add) {
```
josejulio: First searchTerm didn't match any field? Break the searchTerms loop.

wpride:
@josejulio Looked at this with @ctsims and unfortunately at the moment we think that for most of our applications we want the "AND" filtering functionality as its currently written. I'll work with you offline to discuss other possible avenues to get the behavior you need here

josejulio:
Thanks William, I though that it should be OR because of our call the other day.

PR #877, opened 2015-12-01
Augment session restoration

amstone326:
Added:
-A checkbox on login screen to indicate if you want to restore the saved session (only visible if there is a session currently saved for that app)
-An item in the options menus of EntitySelectActivity, MenuItem, MenuGrid, and FormEntryActivity for saving the current session
-An item in the options menu of CommCarePreferences (very hidden) for clearing the saved session

@phillipm A couple of things:
1. Should we make the "Save Session" items in the menus only visible based on a developer setting flag? I'm loathe to keep it coupled with auto-login, but we could just make it its own flag really quick.
2. From quick testing, it seems like the session restorer can't take you back to an entity detail screen, it just takes you to the Entity Select screen immediately prior to that. That makes total sense because when there is an entity detail enabled, nothing gets added onto the session stack to represent entity selection until you press confirm on the detail screen. I think this is totally fine for now, just a heads up (I didn't put the Save Session menu item in EntityDetailActivity for this reason)

### DIFF
--- a/app/src/org/commcare/android/framework/SaveSessionCommCareActivity.java
+++ b/app/src/org/commcare/android/framework/SaveSessionCommCareActivity.java
@@ -1,0 +1,36 @@
+package org.commcare.android.framework;
+
+import android.view.Menu;
+import android.view.MenuItem;
+
+import org.commcare.android.session.DevSessionRestorer;
+import org.javarosa.core.services.locale.Localization;
+
+/**
+ * Any CC activity from which we want the user to be able to save the current session via an item
+ * in the options menu should subclass this class
+ *
+ * @author amstone
+ */
+public abstract class SaveSessionCommCareActivity<R> extends SessionAwareCommCareActivity<R> {
+
+    // Choose a high number so that it will be higher than the menu ids of all other menu items
+    // created by subclasses
+    private static final int MENU_SAVE_SESSION = 9999;
+
+    @Override
+    public boolean onCreateOptionsMenu(Menu menu) {
+        super.onCreateOptionsMenu(menu);
+        menu.add(Menu.NONE, MENU_SAVE_SESSION, Menu.NONE, Localization.get("menu.save.session"));
+        return true;
+    }
+
+    @Override
+    public boolean onOptionsItemSelected(MenuItem item) {
+        if (item.getItemId() == MENU_SAVE_SESSION) {
+            DevSessionRestorer.saveSessionToPrefs();
+        }
+        return super.onOptionsItemSelected(item);
+    }
+
+}
--- a/app/src/org/commcare/android/framework/SaveSessionCommCareActivity.java
+++ b/app/src/org/commcare/android/framework/SaveSessionCommCareActivity.java
@@ -4,6 +4,7 @@ import android.view.Menu;
 import android.view.MenuItem;
 
 import org.commcare.android.session.DevSessionRestorer;
+import org.commcare.dalvik.preferences.DeveloperPreferences;
 import org.javarosa.core.services.locale.Localization;
 
 /**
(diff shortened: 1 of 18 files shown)


app/src/org/commcare/android/framework/SaveSessionCommCareActivity.java
```diff
@@ -0,0 +1,36 @@
...
+    public boolean onCreateOptionsMenu(Menu menu) {
+        super.onCreateOptionsMenu(menu);
+        menu.add(Menu.NONE, MENU_SAVE_SESSION, Menu.NONE, Localization.get("menu.save.session"));
+        return true;
+    }
+
+    @Override
+    public boolean onOptionsItemSelected(MenuItem item) {
+        if (item.getItemId() == MENU_SAVE_SESSION) {
+            DevSessionRestorer.saveSessionToPrefs();
```
phillipm: Probably want to return true here instead of calling the super method to make sure we stop processing item selections.

app/src/org/commcare/android/framework/SaveSessionCommCareActivity.java
```diff
@@ -0,0 +1,36 @@
...
+ * Any CC activity from which we want the user to be able to save the current session via an item
+ * in the options menu should subclass this class
+ *
+ * @author amstone
+ */
+public abstract class SaveSessionCommCareActivity<R> extends SessionAwareCommCareActivity<R> {
+
+    // Choose a high number so that it will be higher than the menu ids of all other menu items
+    // created by subclasses
+    private static final int MENU_SAVE_SESSION = 9999;
```
phillipm: I wish there was a nice way to ensure that menu item ids didn't conflict when building menus across inheritance schemes. I don't think there is though.

app/src/org/commcare/dalvik/application/CommCareApplication.java
```diff
@@ -874,7 +875,7 @@ public void onServiceConnected(ComponentName className, IBinder service) {
                     if (user != null) {
                         mBoundService.startSession(user);
                         attachCallListener();
-                        if (DevSessionRestorer.autoLoginEnabled()) {
+                        if (restoreSession) {
```
phillipm: Nice

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -50,6 +50,7 @@
 import android.widget.Toast;
 
 import org.commcare.android.framework.CommCareActivity;
+import org.commcare.android.framework.SaveSessionCommCareActivity;
 import org.commcare.android.framework.SessionAwareCommCareActivity;
```
phillipm: You should remove the `SessionAwareCommCareActivity`import wherever you use the `SaveSessionCommCareActivity` instead

phillipm:
@amstone326 I vote yes in response to 1)
Looks good after you address my one actionable comment

app/src/org/commcare/android/framework/SaveSessionCommCareActivity.java
amstone326: Yea wasn't happy with this at all but also had no alternatives

phillipm:
Just tested it out, thanks for taking the next step usability-wise! This is something that power app builders can now use instead of just mobile devs!

app/src/org/commcare/android/framework/SaveSessionCommCareActivity.java
amstone326: Whoa so this actually doesn't work, it causes an IndexOutOfBounds exception in some instances (unclear why only sometimes, I only just encountered it when I tried adding onPrepareOptionsMenu() and access the item at index 9999). I think I have to make this id Menu.First, and then have all the other classes that subclass this start their menu numbering at Menu.First + 1

amstone326:
@phillipm Should be good to go now

amstone326:
@dimagi/product Users can now turn on a developer setting that allows them to save the current session, and then upon login be brought back to the last point they were at in that session.

amsagoff:
thanks for the mention - assume this is for 2.26?

amstone326:
Yes! You can tell for sure by the milestone tag, which we're usually pretty
good about using.

phillipm:
When I use the "relevant for GS" tag I like to bold some text saying why in normal-ish english. Or link to some docs. Would be good to update https://confluence.dimagi.com/display/ccinternal/Advanced+Mobile+Debugging

amstone326:
**Relevant for GS**: This page (https://confluence.dimagi.com/display/ccinternal/Advanced+Mobile+Debugging) now contains updated documentation for both this feature and the auto-login feature. Would be great to make that page known to people in the next report-out!

amsagoff:
nice!!

PR #888, opened 2015-12-02
Fix progress dialog from being cut off

amstone326:
On older phones, the bottom of progress dialogs with indeterminate progress bars was getting cut off. In fixing that, the dialog text is no longer centered vertically relative to the progress bar if it is a single line of text, and I can't figure out how to fix that (struggled with it for a while). But this is much better than before and don't want to keep sinking too much time into that for now.


phillipm:
Nice, glad this got fixed.

PR #891, opened 2015-12-03
Remove duplicate call to refreshForNewApp() 

amstone326:
Realized that since onResume() is always called after onActivityResult(), it is unnecessary to have an additional call to refreshForNewApp() in onActivityResult(), since it will happen in onResume()


phillipm:
Ready to go once merge conflicts are resolved

amstone326:
@phillipm Updated

PR #894, opened 2015-12-04
Migrate fixtures to new serialization schema

phillipm:
A bug came up in how TreeElements are serialized (fixed by https://github.com/dimagi/javarosa/pull/204). Since fixtures store serialized tree elements in a database, we need to do a database migration, otherwise fixtures won't load upon updating to CommCare 2.25

This PR performs the fixture db migration by (manually) loading the fixture instance from the db using the old serialization scheme and then serializing using the new scheme and updating the db.

cross-requested with https://github.com/dimagi/javarosa/pull/204 and https://github.com/dimagi/commcare/pull/215


phillipm:
Changed to target master https://github.com/dimagi/commcare-odk/pull/905

PR #895, opened 2015-12-07
Regression test for 2.24.2 form save bug

phillipm:
Robolectric test to ensure forms with hidden repeat groups at the end of their forms still save, similar to the bug that required the 2.24.2 hotfix

The test:
 - launches form entry 
 - answer simple question
 - completes and saves form


app/src/org/commcare/android/util/AndroidInstanceInitializer.java
```diff
@@ -61,6 +61,10 @@ public String getVersionString() {
 
     @Override
     public String getDeviceId() {
-        return CommCareApplication._().getPhoneId();
+        String phoneId = CommCareApplication._().getPhoneId();
+        if (phoneId == null) {
+            return super.getDeviceId();
+        }
+        return phoneId;
```
phillipm: Robolectric returns `null` for imei, which in turn breaks some form metadata setup. This logic catches it `null` values and returns the call to super instead, which is `"----"`. Sorta messy, open to finding an alternative if someone thinks it is important.

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -173,6 +171,7 @@ protected void onCreate(Bundle savedInstanceState) {
         ACRAUtil.registerAppData();
         uiController = new HomeActivityUIController(this);
         sessionNavigator = new SessionNavigator(this);
+        formAndDataSyncer = new FormAndDataSyncer(this);
```
phillipm: Isolate form submission and data sync logic into something easily faked.

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -1543,4 +1381,25 @@ public boolean isBackEnabled() {
...
+
+    /**
+     * For Testing purposes only
+     */
+    public void setFormAndDataSyncer(FormAndDataSyncer formAndDataSyncer) {
+        if (BuildConfig.DEBUG) {
+            this.formAndDataSyncer = formAndDataSyncer;
+        } else {
+            throw new RuntimeException("On principal of design, only meant for testing purposes");
+        }
```
phillipm: Sad that cruft sometimes comes along with tests. Definitely interested in hearing opinions on better ways to do this.

app/src/org/odk/collect/android/widgets/StringWidget.java
```diff
@@ -222,4 +222,11 @@ public void onTextChanged(CharSequence s, int start, int before, int count) {
     public void setLastQuestion(boolean isLast){
            // nothing changes for Strings
     }
+
+    /**
+     * For testing purposes
+     */
+    public void setAnswer(String answer) {
+        mAnswer.setText(answer);
+    }
```
phillipm: Eventually we should make a whole form entry abstraction layer for filling out forms in a way similar to calabash. This is just a hacky start

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
amstone326: Awesome for cleaning up CCHomeActivity, also.

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
amstone326: Don't have any great ideas for an alternative. The way I've made this sort of situation a little bit cleaner in the past is to make the object that has a test version be an argument to the constructor, and pass in the appropriate version for each ase. But since this is an activity that's not really an option. Seems ok to leave like this for now.

unit-tests/src/org/commcare/android/tests/formnav/EndOfFormTest.java
```diff
@@ -0,0 +1,143 @@
...
+        // launch form entry
+        FormEntryActivity formEntryActivity =
+                Robolectric.buildActivity(FormEntryActivity.class).withIntent(formEntryIntent)
+                        .create().start().resume().get();
+
+        ImageButton nextButton = (ImageButton)formEntryActivity.findViewById(R.id.nav_btn_next);
+        // demonstrate form progress bug where it thinks the form is done until
+        // you enter an answer for the question, which effects the hidden
+        // repeat coming up next
+        assertTrue(nextButton.getTag().equals(FormEntryActivity.NAV_STATE_DONE));
```
amstone326: I'm not sure if I'm just lacking context on this, but am confused why you're testing for the *persistence* of bugs, both here and on line 119. Is there a reason they're not getting fixed?

unit-tests/src/org/commcare/android/tests/formnav/EndOfFormTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+        ShadowActivity shadowFormEntryActivity = Shadows.shadowOf(formEntryActivity);
+        while (!shadowFormEntryActivity.isFinishing()) {
+            Log.d(TAG, "Waiting for the form to save and the form entry activity to finish");
+        }
+
+        return shadowFormEntryActivity;
+    }
+
+    private void assertStoredFroms() {
```
amstone326: typo

unit-tests/src/org/commcare/android/tests/formnav/EndOfFormTest.java
phillipm: It is a hard bug to fix. I don't really know why I'm highlighting it in this tests; I guess as a personal note because I intend to eventually look into it and this is a good example of how to reproduce it.

amstone326: Hm, I guess this seems particularly odd to me because if someone other than you fixes this without knowledge of this test, the test will start to fail, which is confusing. I would at the very least add an explicit comment on what this is doing because it took me several times of reading it to understand that you were in fact testing for the existence of bugs. But my vote would likely be to have only a comment about it, and remove the actual tests.

phillipm: Cool, since it was confusing I'll just take it out.

PR #896, opened 2015-12-07
Don't show subheader if it has same text as header

phillipm:
Don't show menu list subheader if it has the same title as the action bar header.

Before:
image

After (ignore missing icons):
screen

http://manage.dimagi.com/default.asp?187640


ctsims:
Phillip: Did we run this by biyeun and make an intentional decision v.
swapping what the action bar header said?

phillipm:
@ctsims, yes it is in the fogbugz ticket

PR #897, opened 2015-12-07
Dismiss error dialog when the okay button is pressed

phillipm:
Pressing the OK button wasn't doing anything when the `shouldExit` param of `CommCareActivity.createErrorDialog` wasn't set.


amstone326:
Yikes sorry

PR #898, opened 2015-12-07
Show error message when startNextSessionStep fails

phillipm:
If you try to enter a part of the app that needs fixtures but you are in demo mode, the app crashes. Instead, tell the user that that part of the app isn't available and back up one step.

screen

cross-requested with https://github.com/dimagi/commcare/pull/216


app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -475,11 +475,27 @@ else if(resultCode == RESULT_OK){
...
-            sessionNavigator.startNextSessionStep();
+            startNextSessionStepSafe();
         }
         super.onActivityResult(requestCode, resultCode, intent);
     }
 
+    private void startNextSessionStepSafe() {
+        try {
+            sessionNavigator.startNextSessionStep();
+        } catch (RuntimeException e) {
```
phillipm: Might be worth only catching fixture specific exceptions, but that would require more work. Let me know what you think

PR #902, opened 2015-12-08
fix cut off labels for time graphs

orangejenny:
http://manage.dimagi.com/default.asp?188934

Before:
before

After:
after


orangejenny:
top y-label still a little cut off, but there isn't a trivial way to fix this, can deal with it by increasing y-max

PR #903, opened 2015-12-08
Small choice dialog fixes

amstone326:
-ListView selector wasn't working; add a selector to the individual buttons instead
-Make target area larger (had to do this with padding because the single_dialog_choice_view wasn't respecting its layout_height parameter; still not sure why)


phillipm:
This is a followup to https://github.com/dimagi/commcare-odk/pull/749

PR #905, opened 2015-12-09
Migrate fixtures to new serialization schema

phillipm:
A bug came up in how TreeElements are serialized (fixed by https://github.com/dimagi/javarosa/pull/208). Since fixtures store serialized tree elements in a database, we need to do a database migration, otherwise fixtures won't load upon updating to CommCare 2.26

This PR performs the fixture db migration by (manually) loading the fixture instance from the db using the old serialization scheme and then serializing using the new scheme and updating the db.

cross-requested with https://github.com/dimagi/javarosa/pull/208 and https://github.com/dimagi/commcare/pull/220


app/src/org/commcare/android/database/migration/FixtureSerializationMigration.java
```diff
@@ -0,0 +1,89 @@
...
+        long start = System.currentTimeMillis();
+        db.beginTransaction();
+        ConcreteAndroidDbHelper helper = new ConcreteAndroidDbHelper(c, db);
+        Cursor cur = null;
+        DataInputStream fixtureByteStream = null;
+        try {
+            SqlStorage<Persistable> userFixtureStorage =
+                    new SqlStorage<Persistable>("fixture", FormInstance.class, helper);
+            cur = db.query("fixture", new String[]{DbUtil.ID_COL}, null, null, null, null, null);
+            Vector<Integer> ids = SqlStorage.fillIdWindow(cur, DbUtil.ID_COL);
```
ctsims: You can do this process by using a storage iterator from SqlStorage.iterate(false) which will give you an iterator that will only load ID's. Would prefer to do that than expose direct DB queries.

app/src/org/commcare/android/database/migration/FixtureSerializationMigration.java
```diff
@@ -0,0 +1,89 @@
...
+
+                fixtureByteStream =
+                        new DataInputStream(new ByteArrayInputStream(userFixtureStorage.readBytes(id)));
+                fixture.migrateSerialization(fixtureByteStream, helper.getPrototypeFactory());
+                userFixtureStorage.update(id, fixture);
+            }
+            db.setTransactionSuccessful();
+            return true;
+        } catch (Exception e) {
+            // Even if it failed, let the migration think it was successful,
```
ctsims: Hm, in this case that might be the best we can do. The alternative here, though is that people block until a bugfix comes out, which might be safer. The softassert is a good pattern to follow though, seems reasonable to stick with that.

ctsims:
Looks good. One change listed (not mandatory if there's a good reason for the current process)

phillipm:
Updated

app/src/org/commcare/android/database/migration/FixtureSerializationMigration.java
```diff
@@ -0,0 +1,86 @@
...
+            // otherwise the app will crash. It's important to let user get to
+            // a point where they can sync data and then clear user data and
+            // restore, which ultimately has the same effect as running the
+            // fixture serialization migration.
+            db.setTransactionSuccessful();
+            Logger.log(AndroidLogger.SOFT_ASSERT, "fixture serialization db migration failed");
+            Logger.exception(e);
+            return false;
```
ctsims: If we're proceeding from here, actually, I think we _have_ to return true, right? Otherwise people will miss 7->8+ upgrade paths after this? Feels like if we're gonna return false we should have an explicit process to follow to recover from doing so in the upgrader (or at least a warning in the code to let people know that it could happen).

phillipm: Good catch. I've updated it to return `true` with a note saying that subsequent migrations may fail if they depend on fixtures.

PR #906, opened 2015-12-09
Store fixtures in files instead of DB

phillipm:
Fixtures used to be stored completely in the DB. This limited them to 1mb in size, which has become unrealistic. 
This PR moves fixture storage to a hybrid db / file approach: using the normal db setup but adding a column (`DatabaseHelper.FILE_COL`) that points to the filepath where the fixture is stored in a serialized, usually encrypted, manner. Also adds a table column for the encryption key (`DatabaseHelper.AES_COL`), so that encrypted io is contained in the storage layer.

Includes a database migration to the new setup.

Cross-requested with https://github.com/dimagi/javarosa/pull/209 and https://github.com/dimagi/commcare/pull/221

__Release Notes__: remove 1mb lookup table size restriction


unit-tests/src/org/commcare/android/database/SqlFileBackedStorageTests.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    @Test
+    public void testRecordLookup() {
+        SqlFileBackedStorage<FormInstance> userFixtureStorage =
+                CommCareApplication._().getCurrentApp().getFileBackedStorage("fixture", FormInstance.class);
+
+        // TODO PLM: this should return a result. I think the ':' isn't escaped properly in the query
+        Vector<FormInstance> shouldHaveFormResults = userFixtureStorage.getRecordsForValues(new String[]{FormInstance.META_ID}, new String[]{"commtrack:programs"});
+        Assert.assertTrue(shouldHaveFormResults.size() == 0);
```
phillipm: Any thoughts on this ^?
A fixture with an instance id of  `commtrack:programs` should exist. I think the `:` is messing up the query. We should really look into this!

unit-tests/src/org/commcare/android/database/SqlFileBackedStorageTests.java
```diff
@@ -0,0 +1,272 @@
...
+                removeOneEntry(i.nextID(), userFixtureStorage);
+            } else if (count == 1) {
+                idOne = i.nextID();
+            } else if (count == 2) {
+                removeTwoEntries(idOne, i.nextID(), userFixtureStorage);
+            } else {
+                // seems to be required; otherwise iterator loops forever. Not
+                // sure if it is a robolectric bug or a bug in our iterator
+                // that comes up when we iterate and delete at the same time
+                break;
```
phillipm: Worth noting ^

app/src/org/commcare/android/database/SqlStorage.java
```diff
@@ -413,12 +449,13 @@ public boolean isEmpty() {
...
-        String[] projection = includeData ? new String[]{DbUtil.ID_COL, DbUtil.DATA_COL} : new String[]{DbUtil.ID_COL};
-        Cursor c = db.query(table, projection, null, null, null, null, null);
-        return new SqlStorageIterator<>(c, this);
+        return null;
     }
 
+    protected Cursor getIterateCursor(SQLiteDatabase db, boolean includeData) {
+        String[] projection = includeData ? new String[]{DatabaseHelper.ID_COL, DatabaseHelper.DATA_COL} : new String[]{DatabaseHelper.ID_COL};
+        return db.query(table, projection, null, null, null, null, null);
+    }
```
phillipm: Split this logic into a helper method so subclass `FileBackedSqlStorage` can override it with appropriate logic.

app/src/org/commcare/android/database/SqlStorage.java
```diff
@@ -199,43 +210,19 @@ public T getRecordForValues(String[] rawFieldNames, Object[] values) throws NoSu
...
-                throw new InvalidIndexException("Invalid unique column " + scrubbedName + ". Multiple records found with value " + value.toString(), scrubbedName);
-            }
-            c.moveToFirst();
-            byte[] data = c.getBlob(c.getColumnIndexOrThrow(DbUtil.DATA_COL));
-            return newObject(data);
-        } finally {
-            if (c != null) {
-                c.close();
-            }
-        }
```
phillipm: This was all duplicated in `getRecordForValues`

app/src/org/commcare/android/crypt/EncryptionIO.java
```diff
@@ -0,0 +1,95 @@
...
+            throws FileNotFoundException {
+        final File path = new File(filename);
+        FileOutputStream fos = new FileOutputStream(path);
+        if (symetricKey == null) {
+            return fos;
+        } else {
+            try {
+                Cipher cipher = Cipher.getInstance("AES");
+                cipher.init(Cipher.ENCRYPT_MODE, symetricKey);
+                return new CipherOutputStream(fos, cipher);
```
ctsims: It seems odd to me that we buffer the read below, but not the write here. Is there a reason we don't?

app/src/org/commcare/android/crypt/EncryptionIO.java
```diff
@@ -0,0 +1,95 @@
...
+            //form into a _Byte array_, I don't think there's a lot of concern than 2GB of data
+            //are gonna sneak by.
+
+            return is;
+
+            //CTS - Removed the byte array length check here. Plenty of
+            //files are smaller than their contents (padded encryption data, etc),
+            //so you can't actually know that's correct. We should be relying on the
+            //methods we use to read data to make sure it's all coming out.
+        } catch (FileNotFoundException e) {
```
ctsims: You may have just copied this over, but is there a reason this returns null instead of forwarding the result like the previous method?

app/src/org/commcare/android/database/HybridFileBackedSqlHelpers.java
```diff
@@ -0,0 +1,108 @@
...
+        if (!newFile.createNewFile()) {
+            throw new RuntimeException("Trying to create a new file using existing filename; " +
+                    "Shouldn't be possible since we already checked for uniqueness");
+        }
+
+        return newFile;
+    }
+
+    private static File getUniqueFilename(File dbDir) {
+        String uniqueSuffix = "_" + (Math.random() * 5);
```
ctsims: Can we just use GUID's for these? seems more reliable/consistent than making up a structure.

app/src/org/commcare/android/database/HybridFileBackedSqlStorage.java
```diff
@@ -0,0 +1,544 @@
...
+                        " with names " + Arrays.toString(rawFieldNames) +
+                        " and values " + Arrays.toString(values));
+            } else if (queryCount > 1) {
+                throw new InvalidIndexException("Invalid unique column set" +
+                        Arrays.toString(rawFieldNames) +
+                        ". Multiple records found with value " +
+                        Arrays.toString(values), Arrays.toString(rawFieldNames));
+            }
+            cur.moveToFirst();
+            byte[] serializedObj = cur.getBlob(cur.getColumnIndexOrThrow(DatabaseHelper.DATA_COL));
```
ctsims: This logic seems replicated in quite a few places (read object, check null, deserialize from either file or blob),  like readBytes below and getRecordsForValues above

app/src/org/commcare/android/database/HybridFileBackedSqlStorage.java
```diff
@@ -0,0 +1,544 @@
...
+            // rather, needed in case persistable object is used after being written to storage.
+            persistable.setID((int)insertedId);
+
+            if (insertedId > Integer.MAX_VALUE) {
+                throw new RuntimeException("Waaaaaaaaaay too many values");
+            }
+
+            db.setTransactionSuccessful();
+        } catch (IOException e) {
+            e.printStackTrace();
```
ctsims: ? This should absolutely throw this exception, shouldn't it?

app/src/org/commcare/android/database/HybridFileBackedSqlStorage.java
```diff
@@ -0,0 +1,544 @@
...
+                    insertedId = db.insertOrThrow(table, DatabaseHelper.DATA_COL, contentValues);
+                } else {
+                    // store serialized object in file and file pointer in db
+                    File dataFile = HybridFileBackedSqlHelpers.newFileForEntry(dbDir);
+                    ContentValues contentValues = helper.getNonDataContentValues(persistable);
+                    contentValues.put(DatabaseHelper.FILE_COL, dataFile.getAbsolutePath());
+                    byte[] key = generateKeyAndAdd(contentValues);
+                    insertedId = db.insertOrThrow(table, DatabaseHelper.FILE_COL, contentValues);
+
+                    writeStreamToFile(bos, dataFile.getAbsolutePath(), key);
```
ctsims: There's no way to undo this op, currently. An atomic structure for this method would be

* Create DB row/record for "Remove file at PATH on next clean"
* Open Transaction
* Write Row
* Delete "Remove file" row
* Close transaction

app/src/org/commcare/android/database/HybridFileBackedSqlStorage.java
```diff
@@ -0,0 +1,544 @@
...
+
+    private void writeStreamToFile(ByteArrayOutputStream bos, String filename,
+                                   byte[] key) throws IOException {
+        DataOutputStream fileOutputStream = null;
+        try {
+            fileOutputStream = getOutputFileStream(filename, key);
+            bos.writeTo(fileOutputStream);
+        } finally {
+            if (fileOutputStream != null) {
+                fileOutputStream.close();
```
ctsims: My guess is we should wrap this in an IOException wrapper, since we generally swallow stream close IOExceptions. (The method signature is eating it right now)

app/src/org/commcare/android/database/HybridFileBackedSqlStorage.java
```diff
@@ -0,0 +1,544 @@
...
+                                        SQLiteDatabase db, int id) {
+        ContentValues updatedContentValues =
+                helper.getContentValuesWithCustomData(extObj, bos.toByteArray());
+        if (!objectInDb) {
+            // was stored in file: remove file and store in db
+            updatedContentValues.put(DatabaseHelper.FILE_COL, (String)null);
+            updatedContentValues.put(DatabaseHelper.AES_COL, (byte[])null);
+
+            File dataFile = new File(filename);
+            dataFile.delete();
```
ctsims: If this occurs but the transaction fails our record is now invalid. I would add this to a queue somewhere that manages files to delete so we can't end up in a state where the record becomes corrupted

app/src/org/commcare/android/database/HybridFileBackedSqlStorage.java
```diff
@@ -0,0 +1,544 @@
...
+            updatedContentValues.put(DatabaseHelper.FILE_COL, filename);
+            fileEncryptionKey = generateKeyAndAdd(updatedContentValues);
+        } else {
+            // was stored in a file all along, update file
+            updatedContentValues = helper.getNonDataContentValues(extObj);
+        }
+        db.update(table, updatedContentValues,
+                DatabaseHelper.ID_COL + "=?", new String[]{String.valueOf(id)});
+
+        writeStreamToFile(bos, filename, fileEncryptionKey);
```
ctsims: Same comments as above re:atomicity

app/src/org/commcare/android/database/HybridFileBackedSqlStorage.java
```diff
@@ -0,0 +1,544 @@
...
+
+    @Override
+    public void remove(int id) {
+        SQLiteDatabase db = getDbOrThrow();
+
+        db.beginTransaction();
+        try {
+            String filename = HybridFileBackedSqlHelpers.getEntryFilename(helper, table, id);
+            if (filename != null) {
+                File dataFile = new File(filename);
```
ctsims: Same comments as above re:atomicity

app/src/org/commcare/android/database/HybridFileBackedSqlStorage.java
```diff
@@ -0,0 +1,544 @@
...
+        }
+    }
+
+    @Override
+    public void remove(List<Integer> ids) {
+        if (ids.size() > 0) {
+            SQLiteDatabase db = getDbOrThrow();
+            db.beginTransaction();
+            try {
+                HybridFileBackedSqlHelpers.removeFiles(ids, helper, table);
```
ctsims: Same comments as above re:atomicity

app/src/org/commcare/android/database/HybridFileBackedStorageIterator.java
```diff
@@ -0,0 +1,46 @@
...
+     * @param storage The storage being queried
+     */
+    public HybridFileBackedStorageIterator(Cursor c,
+                                           HybridFileBackedSqlStorage<T> storage) {
+        super(c, storage, null);
+    }
+
+    @Override
+    public T nextRecord() {
+        byte[] blob =
```
ctsims: This doesn't seem to have a path to follow if the "includeData" flag was set to false.

app/src/org/commcare/android/database/SqlStorage.java
```diff
@@ -45,26 +47,35 @@
...
+        T e = null;
         try {
-            T e = ctype.newInstance();
-            if (e instanceof EncryptedModel) {
-                em = (EncryptedModel) e;
-            }
-        } catch (IllegalAccessException | InstantiationException ie) {
+            e = ctype.newInstance();
+        } catch (IllegalAccessException ie) {
+            ie.printStackTrace();
```
ctsims: Odd, why did you unfold these?

app/src/org/commcare/android/database/migration/FixtureSerializationMigration.java
```diff
@@ -48,16 +55,27 @@ public static boolean migrateFixtureDbBytes(SQLiteDatabase db, Context c) {
...
+            } else {
+                fixtureStorage =
+                        new UnencryptedHybridFileBackedSqlStorage<Persistable>("fixture", FormInstance.class, helper, baseDir);
+            }
+            SqlStorage<Persistable> oldUserFixtureStorage =
+                    new SqlStorage<Persistable>("oldfixture", FormInstance.class, helper);
+            int count = 0;
+            for (SqlStorageIterator i = oldUserFixtureStorage.iterate(false); i.hasMore(); ) {
                 int id = i.nextID();
+                Log.d(TAG, "migrating fixture " + count++);
```
ctsims: Seems dangerously cute. I'd pull it out, lest someone rely on count in the future, and someone else removes the log statement. Log statements don't feel like they should have side effects.

app/src/org/commcare/android/database/user/UserDatabaseUpgrader.java
```diff
@@ -214,7 +217,33 @@ private boolean upgradeSevenEight(SQLiteDatabase db, int oldVersion, int newVers
      * attributes.
      */
     private boolean upgradeEightNine(SQLiteDatabase db, int oldVersion, int newVersion) {
-        return FixtureSerializationMigration.migrateFixtureDbBytes(db, c);
+        Log.d(TAG, "starting user fixture migration");
```
ctsims: Is this identical to the other method other than one or two parameters? Feels like there's a lot of repeated code.

app/src/org/commcare/android/db/legacy/LegacySqlIndexedStorageUtility.java
```diff
@@ -126,16 +123,17 @@ public T getRecordForValues(String[] rawFieldNames, Object[] values) throws NoSu
             throw new InvalidIndexException("Invalid unique column set" + rawFieldNames + ". Multiple records found with value " + values.toString(), rawFieldNames.toString());
         }
         c.moveToFirst();
-        byte[] data = c.getBlob(c.getColumnIndexOrThrow(DbUtil.DATA_COL));
+        byte[] data = c.getBlob(c.getColumnIndexOrThrow(DatabaseHelper.DATA_COL));
+        int dbId = c.getInt(c.getColumnIndexOrThrow(DatabaseHelper.ID_COL));
         c.close();
-        return newObject(data);
+        return newObject(data, dbId);
```
ctsims: I'd prefer we move the newObject code in its old form to this class rather than change how this class functions in general. This was a snapshot of a procedural generation/translation process so it's really sketch for us to make changes to that process over time, and cleaner for the code to represent what the legacy object used to do

app/src/org/commcare/dalvik/application/CommCareApplication.java
```diff
@@ -701,12 +702,36 @@ public SQLiteDatabase getHandle() {
...
+        return new HybridFileBackedSqlStorage<>(storage, c, buildUserDbHandle(), getUserDbDir());
+    }
+
+    /**
+     * Get root directory for user-level file-backed sql storage layer
+     * (currently used for storing fixtures).
+     */
+    public String getUserDbDir() {
+        try {
+            return currentApp.storageRoot() + currentApp + "/" + getSession().getLoggedInUser().getUniqueId() + "/";
```
ctsims: This shouldn't actually be necessary, should it? Isn't there already a sandbox for user specific data or is that partition only in the database?

unit-tests/src/org/commcare/dalvik/application/CommCareTestApp.java
```diff
@@ -0,0 +1,24 @@
+package org.commcare.dalvik.application;
+
+import org.commcare.android.database.UnencryptedHybridFileBackedSqlStorage;
+import org.commcare.android.database.UnencryptedHybridFileBackedSqlStorageMock;
+import org.javarosa.core.services.storage.Persistable;
+
+/**
```
ctsims: Class level comment would be welcome here

app/src/org/commcare/android/crypt/EncryptionIO.java
phillipm: I thought in the case where you have all the data already lying around you don't need to use buffered streams?
Since this is a util method used in several contexts, not just in when writing to a byte array to a file in `HybridFileBackedSqlStorage`, I'll add the buffered stream. But am I confused in saying that in the case of `HybridFileBackedSqlStorage` it isn't needed?

ctsims: Writing to a buffered stream should still help with file writes. The data is in memory but we're still writing it in a way that is chunked out of convenience by whatever structure the incoming stream decides (IE: Regardless of whether it's 2 bytes or 134 bytes it'll still try to dump them whenever it has them). Using a buffered output steam lets the writer write in whatever way is convenient without actually starting the file I/O until it can write a nice clean chunk, which should speed things up on both ends.

Basically if one end of a stream is I/O backed that end of the stream should almost always be buffered unless we know for sure that it already is.

app/src/org/commcare/android/crypt/EncryptionIO.java
phillipm: Changed to let FileNotFoundException propagate up.
Note the TODOs I've added https://github.com/dimagi/commcare-odk/pull/906/files#diff-475fa51f6592fd0eb3c922e364db1d6eR137 and https://github.com/dimagi/commcare-odk/pull/906/files#diff-09bcd67419bcbc8e281fe0458496ae74R47

app/src/org/commcare/android/database/HybridFileBackedSqlHelpers.java
phillipm: Like `PropertyUtils.genGUID`? Sounds good

app/src/org/commcare/android/database/HybridFileBackedSqlStorage.java
phillipm: The logic is slightly different in each case:
 - in `getRecordsForValues` I would guess we want to compute the column indexes outside of the `while` loop
 - in `getRecordForValues` we avoid loading the entire byte array by using an input stream
 - while in `readBytes` we directly read out of the file into a byte array.

They could all be unified but I suspect it would lead to slight performance losses. Your call since you understand these implications better than myself.

app/src/org/commcare/android/database/HybridFileBackedStorageIterator.java
phillipm: Neither does the normal SqlIterator. I think the onus is currently on the caller to invoke this correctly.

app/src/org/commcare/android/database/SqlStorage.java
phillipm: "Multi-catch with these reflection exceptions requires API level 19"

phillipm: I mistakenly folded them in a previous PR...

app/src/org/commcare/android/database/user/UserDatabaseUpgrader.java
phillipm: What other method? `FixtureSerializationMigration.migrateFixtureDbBytes` is still called below.

app/src/org/commcare/dalvik/application/CommCareApplication.java
phillipm: I didn't see anything along those lines anywhere. Do you know where I should look?

unit-tests/src/org/commcare/dalvik/application/CommCareTestApp.java
phillipm: Left one that is sort of vague, would appreciate suggestions for ways to make it clearer.

app/src/org/commcare/android/database/HybridFileBackedSqlStorage.java
ctsims: Ah, ok,  I missed those subtleties, that sounds like a good justification to me.

app/src/org/commcare/android/database/HybridFileBackedStorageIterator.java
ctsims: Ah, ok. I thought the SqlIterator fetched on demand, if that's not true, all set on this.

app/src/org/commcare/android/database/SqlStorage.java
ctsims: Oh right, obnoxious. Cool, thanks!

app/src/org/commcare/android/database/user/UserDatabaseUpgrader.java
ctsims: Similar to AppDatabaseUpgrader.java's update method which seems to do the ~same thing with slightly different args, it looks like?

app/src/org/commcare/dalvik/application/CommCareApplication.java
ctsims: Hm, it must be that we store things in one big directory and then only load/decrypt the files that have relevant DB roots. Seems reasonable to partition if we can.

app/src/org/commcare/dalvik/application/CommCareApplication.java
```diff
@@ -701,12 +702,36 @@ public SQLiteDatabase getHandle() {
...
+        } catch (SessionUnavailableException e) {
+            return currentApp.storageRoot() + currentApp + "/";
```
ctsims: It worries me a lot that we get the DB in two different ways here depending on whether the session is available, since it feels like a process could try to get the storage and get a fully _different_ location when someone isn't logged in, then log in and the fixture they were parsing is gone. Seems a bit safer to error out here

unit-tests/src/org/commcare/dalvik/application/CommCareTestApp.java
ctsims: that clarified it for me. Thanks!

app/src/org/commcare/android/database/HybridFileBackedSqlStorage.java
```diff
@@ -0,0 +1,570 @@
...
+                    // store serialized object in file and file pointer in db
+                    String dataFilePath = HybridFileBackedSqlHelpers.newFileForEntry(dbDir).getAbsolutePath();
+                    setFileAsOrphan(db, dataFilePath);
+                    ContentValues contentValues = helper.getNonDataContentValues(persistable);
+                    contentValues.put(DatabaseHelper.FILE_COL, dataFilePath);
+                    byte[] key = generateKeyAndAdd(contentValues);
+                    insertedId = db.insertOrThrow(table, DatabaseHelper.FILE_COL, contentValues);
+
+                    writeStreamToFile(bos, dataFilePath, key);
+                    unsetFileAsOrphan(db, dataFilePath);
```
phillipm: Before I apply this change to other db ops: Does the ordering with `setFileAsOrphan`/`unsetFileAsOrphan` look correct in terms of making `write` atomic?

phillipm:
@ctsims Think I addressed your concerns with making db events atomic. I still need to find a good place and time to clear the orphaned file table, but otherwise should be good to review again.
Any suggestions on when/where to perform the orphan file clearing?

app/src/org/commcare/android/database/HybridFileBackedSqlStorage.java
```diff
@@ -0,0 +1,573 @@
...
+
+        db.beginTransaction();
+        try {
+            db.delete(table, null, null);
+            db.setTransactionSuccessful();
+        } finally {
+            db.endTransaction();
+        }
+
+        FileUtil.deleteFileOrDir(dbDir);
```
phillipm: With the remove db methods I elected to perform file removal after the transaction completed. I'm operating under the assumption that if the transaction isn't set to successful then `endTransaction` will raise a runtime exception and the file deletion logic will never be reached. Wanted to double check: this assumption is correct, right?

app/src/org/commcare/android/database/HybridFileBackedSqlHelpers.java
```diff
@@ -0,0 +1,169 @@
...
+            throw new RuntimeException("Trying to create a new file using existing filename; " +
+                    "Shouldn't be possible since we already checked for uniqueness");
+        }
+
+        return newFile;
+    }
+
+    private static File getUniqueFilename(File dbDir) {
+        String timeAsString =
+                DateTime.now().toString(DateTimeFormat.forPattern("MM_dd_yyyy_HH_mm_ss"));
```
ctsims: I don't think we need this if we just get a long enough GUID below. Is there a specific reason we want to use the time?

app/src/org/commcare/android/database/HybridFileBackedSqlHelpers.java
```diff
@@ -0,0 +1,169 @@
...
+                    cur.moveToNext();
+                }
+            }
+        } finally {
+            if (cur != null) {
+                cur.close();
+            }
+        }
+
+        removeFiles(files);
```
ctsims: We're basically assuming that the filenames are globally unique, but if not, this approach can result in weird behavior like

Table[A,B]

Delete all files: A succeeds, B fails, File System State: B, Table State[A,B]

App checks whether 'A' exists on filesystem before putting something there, sees that it doesn't, adds a new file to slot A

purge triggered

"New" A is removed.

Seems like a fine tradeoff, but it seems worth being explicit that this helper system only works with globally unique names

app/src/org/commcare/android/database/HybridFileBackedSqlStorage.java
ctsims: I think the bigger problem is if something (anything) kills the process after end transaction but before deleting the dir, which will result in the files never going away, since our record of them is gone. Wouldn't it make more sense to delete the full directory after deleting the records from the table, but before the transaction ends?

ctsims:
@phillipm super minor comments and then g2g

app/src/org/commcare/android/database/HybridFileBackedSqlStorage.java
phillipm: But what if the file/dir deletion partially completes and then fails? Then the db entry would persist because the transaction rolled back, but the files would be missing

app/src/org/commcare/android/database/HybridFileBackedSqlHelpers.java
phillipm: Ah, yeah, entirely possible; made note of it in a comment.

app/src/org/commcare/android/database/HybridFileBackedSqlHelpers.java
```diff
@@ -0,0 +1,174 @@
...
+        if (!newFile.createNewFile()) {
+            throw new RuntimeException("Trying to create a new file using existing filename; " +
+                    "Shouldn't be possible since we already checked for uniqueness");
+        }
+
+        return newFile;
+    }
+
+    private static File getUniqueFilename(File dbDir) {
+        String filename = PropertyUtils.genGUID(14);
```
ctsims: Sorry to keep being a pain in the ass about this. 14 chars is only about 72 bits of entropy, we wanna shoot for 128 as a general standard. There's actually a "generateUUID()" method specifically in the same class that will generate UUID v4 id's that I try to use for anything like this just to eliminate any concern.

app/src/org/commcare/android/database/HybridFileBackedSqlStorage.java
ctsims: It's better for the db entry to persist and to try and silently fail to delete non existent files than for the db entries to disappear and the files to remain, though. We can keep trying to delete blank files, but if we remove the index of the files we need to wipe, we won't even know that we need to wipe anything.

app/src/org/commcare/android/database/HybridFileBackedSqlStorage.java
```diff
@@ -0,0 +1,574 @@
...
+     * @param classType     type of object being stored in this database
+     * @param directoryName Name of storage root subdir where entry files are placed
+     */
+    public HybridFileBackedSqlStorage(String tableName,
+                                      Class<? extends T> classType,
+                                      AndroidDbHelper dbHelper,
+                                      String directoryName) {
+        super(tableName, classType, dbHelper);
+
+        final String subPath = GlobalConstants.FILE_CC_DB + directoryName + "/_" + tableName;
```
phillipm: Prefix `tableName` part of the directory with `_` to prevent the issue where if `tableName = ""` and we call `removeAll` it removes subdirectories associated with other tables

app/src/org/commcare/dalvik/application/CommCareApplication.java
```diff
@@ -704,12 +707,32 @@ public SQLiteDatabase getHandle() {
...
+
+    public <T extends Persistable> HybridFileBackedSqlStorage<T> getFileBackedUserStorage(String storage, Class<T> c) {
+        return new HybridFileBackedSqlStorage<>(storage, c, buildUserDbHandle(), getUserKeyRecordId());
+    }
+
+    public String getUserKeyRecordId() {
+        try {
+            return getSession().getUserKeyRecordUUID();
+        } catch (SessionUnavailableException e) {
+            throw new RuntimeException(e);
```
phillipm: Using the user key record uuid as the subdirectory name for a specific user level file-backed db directory. Made sure this same uuid is also used when performing db migrations from the old db setup to this new one.

app/src/org/commcare/dalvik/application/CommCareApplication.java
```diff
@@ -783,10 +806,10 @@ public boolean matches(UserKeyRecord ukr) {
         //TODO: We can just delete the db entirely.
 
         Editor sharedPreferencesEditor = CommCareApplication._().getCurrentApp().getAppPreferences().edit();
+        sharedPreferencesEditor.putString(CommCarePreferences.LAST_LOGGED_IN_USER, null).commit();
 
-        sharedPreferencesEditor.putString(CommCarePreferences.LAST_LOGGED_IN_USER, null);
-
-        sharedPreferencesEditor.commit();
+        // manually clear file-backed fixture storage to ensure files are removed
+        CommCareApplication._().getFileBackedUserStorage("fixture", FormInstance.class).removeAll();
```
phillipm: Making sure to remove files for file-backed databases when a "clear user data" is performed

phillipm: If we wanted to do this in one swoop, instead of calling it on all tables that are file-backed, we could delete the `app/some_app_id/file_db/ukr_uuid_for_user` directory

app/src/org/commcare/android/database/UnencryptedHybridFileBackedSqlStorage.java
```diff
@@ -0,0 +1,48 @@
...
+ * 1mb sql row limit.
+ *
+ * @author Phillip Mates (pmates@dimagi.com).
+ */
+public class UnencryptedHybridFileBackedSqlStorage<T extends Persistable>
+        extends HybridFileBackedSqlStorage<T> {
+    public UnencryptedHybridFileBackedSqlStorage(String table,
+                                                 Class<? extends T> ctype,
+                                                 AndroidDbHelper helper) {
+        super(table, ctype, helper, "app_level");
```
phillipm: Store all unencrypted file-backed db files in `app/some_app_id/file_db/app_level/_tableName`

phillipm:
@ctsims thanks for catching the :file_folder: :bomb: :boom: :boom: :boom:
Should be :ok_hand: :eyeglasses:

PR #910, opened 2015-12-09
Validate cases DAG before case purge

amstone326:
Soft assert when an invalid edge had to be removed

Cross: https://github.com/dimagi/commcare/pull/223, https://github.com/dimagi/javarosa/pull/212


amstone326:
Release Note: Improvements to recovery when phone and server are out of sync

PR #912, opened 2015-12-10
Make EntityView child view IDs unique

orangejenny:
http://manage.dimagi.com/default.asp?187729

@ctsims has context, but as explanation seems like there was an issue with view ids not being unique, so when you selected a case in awesome mode (which highlights that row, which triggers a redraw on the view) that row was re-added to the list in the wrong place (namely, as a child of the first row), causing that first row to expand in height and get distorted.

Putting in master because it's been broken since 2.23 and if the fix breaks anything, it'll likely be something obscure. Did some testing with scrolling case lists. Didn't get multimedia in case lists working (at all, not with this code applied), but I'm going to stop sinking time into this and let QA do its thing.

### DIFF
--- a/app/src/org/commcare/android/view/EntityView.java
+++ b/app/src/org/commcare/android/view/EntityView.java
@@ -81,7 +81,7 @@ public class EntityView extends LinearLayout {
         for (int i = 0; i < views.length; ++i) {
             if (mHints[i] == null || !mHints[i].startsWith("0")) {
                 views[i] = initView(e.getField(i), forms[i], new ViewId(rowId, i, false), e.getSortField(i));
-                views[i].setId(i);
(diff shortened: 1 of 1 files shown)


amstone326:
@orangejenny merging. Should let Kriti know on the ticket that it's not going to be fixed til 2.26.

orangejenny:
@amstone326 will do

ctsims:
@orangejenny @amstone326 - Is that timeline workable for mWellCare? am a
bit surprised we aren't putting in 2.25, given that it's such a small
change.

orangejenny:
Kriti said it was low priority (in ticket). I'm guessing they aren't actually using it and she just noticed it while testing.

PR #913, opened 2015-12-11
Fix for Android Indexes not processing correctly on local forms

ctsims:
The android code changed to stop firing the android specific model processors, which was preventing indexes from updating correctly. 

Depends on 
https://github.com/dimagi/commcare/pull/224/files


amstone326:
@ctsims Can merge this now if you want, but seems like we should be adding regression tests so this doesn't happen again?

ctsims:
@amstone326 Agreed on regressions. Don't think we should block on it yet

PR #914, opened 2015-12-15
Form record migration for Multiple Apps

amstone326:
Adds an app_id field to form records, so that we can grab only forms relevant to the current app (now that multiple apps is a thing).

@ctsims, There are 2 places that I didn't add filtering by the app id, but might need it:
1) AndroidSessionWrapper.getExistingIncompleteCaseDescriptor() -- I think this looks ok because that method is already filtering for only the forms that are part of the current SessionStateDescriptor, but want to be sure.
2) FormRecordCleanupTask -- Do we want form cleanup to only consider forms of the current app, or is it actually better to just go through them all whenever this gets triggered?


app/src/org/commcare/android/database/user/UserDatabaseUpgrader.java
```diff
@@ -38,49 +45,55 @@ public UserDatabaseUpgrader(Context c, boolean inSenseMode) {
 
     public void upgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
         if (oldVersion == 1) {
-            if (upgradeOneTwo(db, oldVersion, newVersion)) {
```
amstone326: all of these oldVersion and newVersion fields were unused

app/src/org/commcare/android/database/user/UserDatabaseUpgrader.java
```diff
@@ -242,4 +311,45 @@ private void markSenseIncompleteUnsent(final SQLiteDatabase db) {
...
+                if (r.getStatus() == ApplicationRecord.STATUS_INSTALLED) {
+                    count++;
+                }
+            }
+            return (count > 1);
+        }
+    }
+
+    public static ApplicationRecord getInstalledAppRecord() {
+        SqlStorage<ApplicationRecord> storage =
```
amstone326: @ctsims -- Want to confirm my intuition that any user db migrations will always occur after the global db migration, and that I can therefore assume that the version of app records in global storage at this point will always be ApplicationRecord, as opposed to ApplicationRecordV1

amstone326: ..which I'm realizing would also render unnecessary the 2 separate cases in multipleInstalledAppRecords() above. I think I can simplify that method if this is true.

app/src/org/commcare/android/db/legacy/LegacyInstallUtils.java
```diff
@@ -485,17 +485,20 @@ public SQLiteDatabase getHandle() {
...
-                    new SqlStorage<User>("USER", User.class, newAndroidDbHelper));
+                    new SqlStorage<>("USER", User.class, newAndroidDbHelper));
             
             Logger.log(AndroidLogger.TYPE_MAINTENANCE, "LegacyUser| Users copied. Copying form records");
             
-            final Map<Integer, Integer> formRecordMapping = SqlStorage.cleanCopy(new LegacySqlIndexedStorageUtility<FormRecord>("FORMRECORDS", FormRecord.class, ldbh),
-                new SqlStorage<FormRecord>("FORMRECORDS", FormRecord.class, newAndroidDbHelper), new CopyMapper<FormRecord>() {
+            final Map<Integer, Integer> formRecordMapping = SqlStorage.cleanCopy(
+                    new LegacySqlIndexedStorageUtility<>(
+                            FormRecord.STORAGE_KEY, FormRecord.class, ldbh),
```
amstone326: Linting to make use of the STORAGE_KEY field

app/src/org/commcare/android/database/global/DatabaseGlobalOpenHelper.java
```diff
@@ -27,7 +27,7 @@
      * V.3 - ApplicationRecord has new fields to support multiple app seating, FormsProvider
      * and InstanceProvider use per-app databases
      */
-    private static final int GLOBAL_DB_VERSION = 3;
+    public static final int GLOBAL_DB_VERSION = 3;
```
phillipm: Why did you make this public?

app/src/org/commcare/android/database/user/UserDatabaseUpgrader.java
phillipm: :+1:

app/src/org/commcare/android/database/user/UserDatabaseUpgrader.java
```diff
@@ -202,6 +214,62 @@ private boolean upgradeSevenEight(SQLiteDatabase db, int oldVersion, int newVers
         }
     }
 
+    /**
+     * Adding an appId field to FormRecords
```
phillipm: Would be good to also mention why this is needed: multiple apps

app/src/org/commcare/android/database/user/UserDatabaseUpgrader.java
```diff
@@ -202,6 +214,62 @@ private boolean upgradeSevenEight(SQLiteDatabase db, int oldVersion, int newVers
...
+     */
+    private boolean upgradeEightNine(SQLiteDatabase db) {
+        db.beginTransaction();
+        try {
+            if (multipleInstalledAppRecords()) {
+                // Cannot migrate FormRecords once this device has already started installing multiple
+                // applications, because there is no way to know which of those apps the existing
+                // FormRecords belong to
+                throw new MigrationException(true);
```
phillipm: What is the end result of this exception being raised? Does CC crash or is the user told to uninstall all their apps and try again? I think an alright failure mode (for power users) would be to fail silently during the migration and warn the user when they login that they need to re-install everything to fix a bug.

app/src/org/commcare/android/database/user/models/FormRecordV1.java
```diff
@@ -0,0 +1,78 @@
...
+import org.commcare.android.storage.framework.Table;
+import org.commcare.modern.models.EncryptedModel;
+import org.commcare.modern.models.MetaField;
+
+import java.util.Date;
+
+/**
+ * The version of a FormRecord that exists in databases on any pre-2.26 version of CommCare. This
+ * class is used to read and make available a FormRecord that exists in such a database
+ * (for database upgrade purposes).
```
phillipm: Would be good to reference the database migration number that uses this code

app/src/org/commcare/android/tasks/ZipTask.java
```diff
@@ -46,20 +47,6 @@
...
-    public static final long TRANSPORT_FAILURE = 4;
-    public static final long PROGRESS_ALL_PROCESSED = 8;
-
-    public static final long SUBMISSION_BEGIN = 16;
-    public static final long SUBMISSION_START = 32;
-    public static final long SUBMISSION_NOTIFY = 64;
-    public static final long SUBMISSION_DONE = 128;
-
-    public static final long PROGRESS_LOGGED_OUT = 256;
-    public static final long PROGRESS_SDCARD_REMOVED = 512;
```
phillipm: :)

app/src/org/commcare/android/database/global/DatabaseGlobalOpenHelper.java
amstone326: ...Was at one point trying to do something useless with it and reverted, forgot to change back. Thanks for catching.

app/src/org/commcare/android/database/user/UserDatabaseUpgrader.java
amstone326: @phillipm This gets propagated back to CommCareApplication and caught here: https://github.com/dimagi/commcare-odk/blob/master/app/src/org/commcare/dalvik/application/CommCareApplication.java#L656, which sets the value of dbState in CCApplication. Then when CCHomeActivity tries to dispatch, it will check the dbState and trigger a handled app exit with the appropriate error message if necessary (https://github.com/dimagi/commcare-odk/blob/master/app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java#L1141-L1151 in CCHomeActivity)

amstone326:
Addressed @phillipm's comments, but need answers to questions for @ctsims before it's merge-able.

amstone326:
@ctsims Waiting for an answer to 1 quick question

app/src/org/commcare/android/database/user/UserDatabaseUpgrader.java
ctsims: Yeah, migrations occur when the DB is opened for the first time, and we need to open the Global and App DB's to get the UserDB handles, so you should feel comfortable assuming that any Global DB migrations have already commited.

amstone326:
@ctsims ok ready to go now.

amstone326:
@ctsims Do you want to do a pass on this or can @phillipm merge it?

ctsims:
Am reviewing stuff today, will definitely prioritize getting in.

app/src/org/commcare/android/database/app/DatabaseAppOpenHelper.java
```diff
@@ -108,6 +108,13 @@ public static String indexOnTableWithPGUIDCommand(String indexName,
         return indexOnTableCommand(indexName, tableName, Resource.META_INDEX_PARENT_GUID);
     }
 
+    /**
+     * Build and return SQL command to add a column to a table
+     */
+    public static String addColumnToTable(String tableName, String columnName, String dataType) {
```
ctsims: I'd move this method to the dbutits, and I think the column name possibly needs to get sanitized, will check usage in code.

app/src/org/commcare/android/database/user/UserDatabaseUpgrader.java
```diff
@@ -208,15 +221,71 @@ private boolean upgradeSevenEight(SQLiteDatabase db, int oldVersion, int newVers
...
-    private boolean upgradeEightNine(SQLiteDatabase db, int oldVersion, int newVersion) {
+    private boolean upgradeEightNine(SQLiteDatabase db) {
         return FixtureSerializationMigration.migrateFixtureDbBytes(db, c);
     }
 
+    /**
+     * Adding an appId field to FormRecords, for compatibility with multiple apps functionality
+     */
+    private boolean upgradeNineTen(SQLiteDatabase db) {
+        db.beginTransaction();
```
ctsims: This could potentially take a long time. I'd probably request more time from the commcare service connector so the app doesn't time out trying to connect to the db (I think the fixture migration code does this)

app/src/org/commcare/android/database/user/UserDatabaseUpgrader.java
```diff
@@ -208,15 +221,71 @@ private boolean upgradeSevenEight(SQLiteDatabase db, int oldVersion, int newVers
...
+        try {
+            if (multipleInstalledAppRecords()) {
+                // Cannot migrate FormRecords once this device has already started installing
+                // multiple applications, because there is no way to know which of those apps the
+                // existing FormRecords belong to
+                throw new MigrationException(true);
```
ctsims: er, do we have a plan for dealing with this? If the app can't start up and we're going to lose these forms anyway, it seems like we don't gain much from either picking a random app and assigning all the forms to it or something?

app/src/org/commcare/android/database/user/UserDatabaseUpgrader.java
```diff
@@ -208,15 +221,71 @@ private boolean upgradeSevenEight(SQLiteDatabase db, int oldVersion, int newVers
...
+                        oldRecord.lastModified(),
+                        appId);
+                newRecord.setID(oldRecord.getID());
+                upgradedRecords.add(newRecord);
+            }
+
+            // Alter the FormRecord table to include an app id column
+            db.execSQL(DatabaseAppOpenHelper.addColumnToTable(
+                    FormRecord.STORAGE_KEY,
+                    FormRecord.META_APP_ID,
```
ctsims: Ok, yeah, should definitely be sanitizing either here or inside the addColumnToTable method

app/src/org/commcare/android/database/user/models/FormRecordV1.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.Date;
+
+/**
+ * This class represents the version of a FormRecord that exists on any devices running a pre-2.26
+ * version of CommCare, which was deprecated in user db version 9. This class is used to read a
+ * FormRecord that exists in such a database, in order to run a db upgrade.
+ *
+ * @author amstone
+ */
+@Table(FormRecord.STORAGE_KEY)
```
ctsims: For the legacy records it's good to manually code this in, since the value of static fields can change over time, and the values and structure of fixed models don't.

app/src/org/commcare/android/tasks/FormRecordCleanupTask.java
```diff
@@ -71,8 +71,10 @@ public FormRecordCleanupTask(Context context, CommCarePlatform platform, int tas
     protected Integer doTaskBackground(Void... params) {
         SqlStorage<FormRecord> storage = CommCareApplication._().getUserStorage(FormRecord.class);
 
-        Vector<Integer> recordsToRemove = storage.getIDsForValues(new String[] { FormRecord.META_STATUS}, new String[] { FormRecord.STATUS_SAVED });
-        int oldrecords = recordsToRemove.size();
+        Vector<Integer> recordsToRemove = storage.getIDsForValues(
+                new String[]{FormRecord.META_STATUS},
+                new String[]{FormRecord.STATUS_SAVED});
```
ctsims: Shouldn't this be app specific as well?

app/src/org/commcare/dalvik/application/CommCareApplication.java
```diff
@@ -860,9 +861,22 @@ public void onServiceConnected(ComponentName className, IBinder service) {
...
 
-                    //Don't let anyone touch this until it's logged in
+                    // Don't let anyone touch this until it's logged in
                     // Open user database
-                    mBoundService.prepareStorage(key, record);
+                    try {
+                        mBoundService.prepareStorage(key, record);
+                    } catch (MigrationException e) {
+                        if (e.isDefiniteFailure()) {
+                            CommCareApplication._().triggerHandledAppExit(CommCareApplication.this,
```
ctsims: This strategy still seems unfortunate to me, since there's no action a user can take that will set things back to a reasonable state. This will just keep crashing the app over and over, even though there are three things we can potentially do:

1) Clear the current user data and log that user in again
2) Clear app data and re-install the app (but not uninstall all apps)
3) Something smart per migration, like moving the forms somewhere that they can be submitted.

For this app, for instance, if people only had Saved/Incomplete forms they're basically fine wiping all of them out. It's only unsent forms that are a problem, right?

ctsims:
@amstone326 Awesome PR! Everything here looks good other than how we're handling multiple app records, I think we can do something a bit smarter on that one. Maybe we should catch up on voice and come up with a concrete plan.

app/src/org/commcare/android/database/user/models/FormRecordV1.java
amstone326: Makes sense

app/src/org/commcare/android/tasks/FormRecordCleanupTask.java
amstone326: @ctsims I had asked that in the PR comment above because I wasn't sure, but left it because no one said anything. Will change.

app/src/org/commcare/android/database/user/UserDatabaseUpgrader.java
amstone326: @ctsims What exactly should I be sanitizing for? Don't think I'm envisioning the vulnerability here properly.

amstone326:
@ctsims OK, ready now! woo

PR #915, opened 2015-12-16
Fix for images providing too much padding in case lists

ctsims:
Tested on a few screen sizes and it looked fine. Couldn't find a super super low density screen though

Old:
icons

New
lower_padding

ticket: http://manage.dimagi.com/default.asp?189511#1061324


phillipm:
This isn't targeting 2.25

ctsims:
@phillipm Tested, looks super great on the small screen, move the target to 2.26 (Where it was intended). unless we think 2.25 is a better target

PR #916, opened 2015-12-17
Awesome autoselect

orangejenny:
@ctsims 

Not much code, but I don't like having the awesome mode conditional in CommCareHomeActivity. Could simplify this PR by removing the pre-select logic I added to EntitySelectActivity.

Bug was that when CommCareHomeActivity detected the autoselect and launched an EntityDetailActivity, that detail activity would be immediately canceled because of this logic: https://github.com/dimagi/commcare-odk/blob/master/app/src/org/commcare/dalvik/activities/EntityDetailActivity.java#L109-L117


orangejenny:
(edited PR description for clarity)

orangejenny:
missed a commit. really ready for review now.

PR #917, opened 2015-12-17
Make secondary y axis respect show-axes configuration

orangejenny:
http://manage.dimagi.com/default.asp?189509

Putting in 2.25 after all because the bug turns out to not be specific to case list graphs and it's a tiny graphing-only change. That said, I don't know of anyone actually using `show-axes`.

### DIFF
--- a/app/src/org/commcare/android/view/c3/AxisConfiguration.java
+++ b/app/src/org/commcare/android/view/c3/AxisConfiguration.java
@@ -27,10 +27,6 @@ public class AxisConfiguration extends Configuration {
             x.put("type", "timeseries");
         }
 
-        // Display secondary y axis, regardless of if it has data; this makes the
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/android/view/c3/AxisConfiguration.java
```diff
@@ -186,6 +182,11 @@ private JSONObject getAxis(String prefix) throws JSONException, InvalidStateExce
         JSONObject config = new JSONObject();
         boolean isX = prefix.equals("x");
 
+        // X and primary Y axis show by default, but not secondary y. Force them all to show.
```
orangejenny: the fix is that this code now happens after the `if (!showAxes)` check above, rather than happening outside of this function and ignoring `showAxes`.

orangejenny:
wrong branch. closing.

PR #921, opened 2015-12-18
Rename install fragments

phillipm:
Every time I open these files I get confused as to which one does what. Hopefully these names are more intuitive.  

I renamed but also reformatted the layout files, so the diff isn't showing them as merely renamed.


app/res/layout/fragment_setup_install.xml
```diff
@@ -1,96 +0,0 @@
...
-                android:textStyle="bold"/>
-            <TextView
-                android:id="@+id/str_setup_message_2"
-                android:layout_width="fill_parent"
-                android:layout_height="wrap_content"
-                android:layout_marginLeft="@dimen/content_min_margin"
-                android:layout_marginRight="@dimen/content_min_margin"
-                android:layout_marginBottom="@dimen/content_min_margin"
-                android:gravity="center"
-                android:text="Please choose an installation method below."
```
phillipm: I removed occurrences of static text like this because we (should) always overwrite it with localized text dynamically.

app/src/org/commcare/android/fragments/InstallAppSelectFragment.java
```diff
@@ -23,10 +23,10 @@
  *
  * @author Daniel Luna (dluna@dimagi.com)
  */
-public class SetupInstallFragment extends Fragment {
+public class InstallAppSelectFragment extends Fragment {
```
amstone326: Nitpicking, but this makes it sound like you're choosing which app to install. Maybe SelectInstallModeFragment?

PR #925, opened 2015-12-22
Signature widget crashing on Android 6

phillipm:
Signature widget was crashing on my phone upon launch. Not sure what was causing it because it looks like it wasn't crashing on 2.25. Regardless this PR addresses the crash.

### DIFF
--- a/app/src/org/odk/collect/android/activities/DrawActivity.java
+++ b/app/src/org/odk/collect/android/activities/DrawActivity.java
@@ -3,7 +3,6 @@ package org.odk.collect.android.activities;
 import android.app.Activity;
 import android.content.Context;
 import android.content.pm.ActivityInfo;
-import android.content.res.Configuration;
 import android.graphics.Bitmap;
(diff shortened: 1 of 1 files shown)


app/src/org/odk/collect/android/activities/DrawActivity.java
```diff
@@ -170,40 +159,46 @@ protected void onCreate(Bundle savedInstanceState) {
...
 
-        pointPaint = new Paint();
+        Paint pointPaint = new Paint();
         pointPaint.setAntiAlias(true);
         pointPaint.setDither(true);
         pointPaint.setStyle(Paint.Style.FILL_AND_STROKE);
         pointPaint.setStrokeWidth(10);
         pointPaint.setColor(Color.BLACK);
 
+        drawView = new DrawView(this, OPTION_SIGNATURE.equals(loadOption),
```
phillipm: The fix is to move `DrawView` creation below `paint` and `pointPaint` setup

PR #926, opened 2015-12-22
Hotfix 2.25.1: Form record process parse failure

phillipm:
Issue: http://manage.dimagi.com/default.asp?189759

Regression test in PR that targets master: https://github.com/dimagi/commcare-odk/pull/927

Caused by https://github.com/dimagi/commcare-odk/pull/913

### DIFF
--- a/app/src/org/commcare/android/models/logic/FormRecordProcessor.java
+++ b/app/src/org/commcare/android/models/logic/FormRecordProcessor.java
@@ -11,7 +11,9 @@ import org.commcare.android.tasks.ExceptionReportTask;
 import org.commcare.android.util.FormUploadUtil;
 import org.commcare.core.process.XmlFormRecordProcessor;
 import org.commcare.dalvik.application.CommCareApplication;
+import org.commcare.data.xml.TransactionParser;
 import org.commcare.xml.AndroidTransactionParserFactory;
+import org.commcare.xml.LedgerXmlParsers;
 import org.javarosa.xml.util.InvalidStructureException;
 import org.javarosa.xml.util.UnfullfilledRequirementsException;
 import org.kxml2.io.KXmlParser;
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/android/models/logic/FormRecordProcessor.java
```diff
@@ -61,12 +63,20 @@ public FormRecord process(FormRecord record)
             FormUploadUtil.getDecryptCipher((new SecretKeySpec(record.getAesKey(), "AES")));
         InputStream is = new CipherInputStream(new FileInputStream(f), decrypter);
 
-        AndroidTransactionParserFactory factory = new AndroidTransactionParserFactory(c, null);
-
-        factory.initCaseParser();
-        factory.initStockParser();
```
phillipm: These calls occur already when the `AndroidTransactionParserFactory` calls its super method https://github.com/dimagi/commcare/blob/master/core/org/commcare/core/parse/CommCareTransactionParserFactory.java#L54-L60

app/src/org/commcare/android/models/logic/FormRecordProcessor.java
```diff
@@ -61,12 +63,20 @@ public FormRecord process(FormRecord record)
...
+                if (LedgerXmlParsers.STOCK_XML_NAMESPACE.equals(namespace) || "case".equalsIgnoreCase(name)) {
+                    return super.getParser(parser);
+                } else {
+                    return null;
+                }
+            }
+        };
 
-        XmlFormRecordProcessor.process(new AndroidSandbox(CommCareApplication._()), is, factory);
+        XmlFormRecordProcessor.process(null, is, factory);
```
phillipm: The sandbox is an unused arg, so replacing it with `null`. Should fix the commcare code to not require it as an arg (outside of hotfix scope)

app/src/org/commcare/android/models/logic/FormRecordProcessor.java
```diff
@@ -61,12 +63,20 @@ public FormRecord process(FormRecord record)
...
-        AndroidTransactionParserFactory factory = new AndroidTransactionParserFactory(c, null);
-
-        factory.initCaseParser();
-        factory.initStockParser();
+        AndroidTransactionParserFactory factory = new AndroidTransactionParserFactory(c, null) {
+            @Override
+            public TransactionParser getParser(KXmlParser parser) {
+                String namespace = parser.getNamespace();
+                String name = parser.getName();
+                if (LedgerXmlParsers.STOCK_XML_NAMESPACE.equals(namespace) || "case".equalsIgnoreCase(name)) {
```
phillipm: The fix is to use `AndroidTransactionParserFactory` parse logic, but limit it only to those parsers that were used before these changes: https://github.com/dimagi/commcare/pull/224/files#diff-f43ca72053194678359963951db09445

PR #927, opened 2015-12-22
Regression test for 2.25.1 hotfix

phillipm:
Regression tests for 2.25.1 hotfix PR: https://github.com/dimagi/commcare-odk/pull/926, as well as the bug fixed in https://github.com/dimagi/commcare-odk/pull/913
(includes changes from that PR so that we don't need to merge it from 2.25 into master)

I verified that this test fails when the changes from https://github.com/dimagi/commcare-odk/pull/926 are not present.

Contains commits from https://github.com/dimagi/commcare-odk/pull/895. Review & merge that PR first


unit-tests/src/org/commcare/android/tests/formsave/FormRecordProcessingTest.java
```diff
@@ -0,0 +1,175 @@
...
+        EvaluationContext ec = TestUtils.getEvaluationContextWithAndroidIIF();
+
+        CaseDbQueryTest.evaluate("instance('casedb')/casedb/case[@case_type = 'followup'][index/parent = 'worker_one']/@case_id",
+                "constant_id", ec);
+
+        fillOutFormWithCaseUpdate();
+
+        // TODO PLM: I would expect the following evaluation to not result in 'constant_id'
+        CaseDbQueryTest.evaluate("instance('casedb')/casedb/case[@case_type = 'followup'][index/parent = 'worker_one']/@case_id",
+                "constant_id", ec);
```
phillipm: This test saves a form that updates the case parent from `worker_one` to `worker_two`, hence I would expect this assert to fail, because the shouldn't a case only have one parent? And in this case it should be `worker_two`, as is asserted right below.

I think this exemplifies a bug in the system.

amstone326:
@phillipm Has conflicts

PR #932, opened 2015-12-28
Only start task progress dialogs once

phillipm:
https://github.com/dimagi/commcare-odk/pull/924 introduced a bug where form load progress dialogs were not dismissed upon form load completion. I think this bug was being caused by multiple calls to `startBlockingForTask`, once when the task starts and once on `CommCareActivity.onResumeFragments`.

The fix in this PR is to introduce `syncTaskBlockingWithDialogFragment` and call it in `CommCareActivity.onResumeFragments`. Its behavior performs any progress dialog transactions relating task changes that occurred while the activity was in the background.

### DIFF
--- a/app/src/org/commcare/android/framework/CommCareActivity.java
+++ b/app/src/org/commcare/android/framework/CommCareActivity.java
@@ -83,7 +83,13 @@ public abstract class CommCareActivity<R> extends FragmentActivity
      * Activity has been put in the background. Flag prevents dialogs
      * from being shown while activity isn't active.
      */
-    private boolean activityPaused;
+    private boolean areFragmentsPaused;
+
+    /**
+     * Mark when task tried to show progress dialog before fragments have resumed,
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/android/framework/CommCareActivity.java
```diff
@@ -572,8 +598,8 @@ public CustomProgressDialog getCurrentProgressDialog() {
     @Override
     public void dismissProgressDialog() {
         CustomProgressDialog progressDialog = getCurrentProgressDialog();
-        if (progressDialog != null && progressDialog.isAdded()) {
-            progressDialog.dismissAllowingStateLoss();
+        if (!areFragmentsPaused && progressDialog != null && progressDialog.isAdded()) {
+            progressDialog.dismiss();
```
phillipm: Forces us to be more careful with fragment transactions. We should always try to avoid `*AllowingStateLoss` because it is prone to introduce ui bugs.

PR #935, opened 2015-12-28
Dispatch activity for routing pre-home screen logic

phillipm:
Create a dedicated dispatch activity to route pre home screen logic. Avoids loading homescreen buttons into memory before they are used, which seems to make the app a bit more responsive. Also results in simpler login activity launch and finish logic.

Fixes auto-login bug where if you try to back out of the login screen when auto-login is enabled it logs you in and hides the app.

Fixes issue where home screen is sometimes shown with no buttons, which was introduced by https://github.com/dimagi/commcare-odk/pull/858

_Note_: I didn't test changes to external app launching, but the changes were pretty minimal so I don't expect issues.


app/src/org/commcare/dalvik/activities/UnrecoverableErrorActivity.java
```diff
@@ -43,7 +39,7 @@ protected Dialog onCreateDialog(int id) {
         AlertDialogFactory factory = new AlertDialogFactory(this, title, message);
         DialogInterface.OnClickListener buttonListener = new DialogInterface.OnClickListener() {
             public void onClick(DialogInterface dialog, int i) {
-               Intent intent = new Intent(UnrecoverableErrorActivity.this, CommCareHomeActivity.class);
+               Intent intent = new Intent(UnrecoverableErrorActivity.this, DispatchActivity.class);
```
phillipm: Pretty strange that we restart the app when it reaches an unrecoverable error. Especially because the dialog says "Shutdown CommCare".
This most likely creates a loop for the user, because when the app restarts I assume it will reach the same error. 
Anyone know why this is here?

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -991,163 +943,15 @@ protected void onResume() {
...
-                    getString(R.string.migration_failure_title), false);
-            return;
-        } else if (dbState == CommCareApplication.STATE_MIGRATION_QUESTIONABLE) {
-            CommCareApplication._().triggerHandledAppExit(this,
-                        getString(R.string.migration_possible_failure),
-                        getString(R.string.migration_failure_title), false);
-            return;
-        } else if (dbState == CommCareApplication.STATE_CORRUPTED) {
-            handleDamagedApp();
-        }
```
phillipm: This db state check logic has been moved to the dispatch activity. That means it is no longer called every time the home screen is resumed, I assume that is a good thing. It will be called every time the dispatch activity is resumed: before app install, before login, before home screen launch.

app/src/org/commcare/dalvik/activities/DispatchActivity.java
```diff
@@ -0,0 +1,311 @@
+package org.commcare.dalvik.activities;
```
ctsims: Does this have an associated layout/view? It will be shown as the parent activity if, say, the corruption dialog is popped up, so it seems like it should have some sort of associated background view (Like Aliza's layout that is shown when commcare is switching between seated apps.

app/src/org/commcare/dalvik/activities/UnrecoverableErrorActivity.java
ctsims: Most of the irrecoverable errors here are unrecoverable in the sense that there's no clear/safe action to follow to get the user back into a state where we are confident that things aren't going to commit bad state. It doesn't necessarily mean "The app can't run", just that we can't be sure that, say, logging you out or something similar is enough to put things right.

ctsims:
One minor comment, then g2g

app/src/org/commcare/dalvik/activities/DispatchActivity.java
phillipm: Corrupt dialogs can be shown on using this activity. I tested it and it displayed fine, see http://manage.dimagi.com/default.asp?189955 for an screenshot.
I wanted to avoid adding any inflation logic, because it really shouldn't _usually_ show anything.

amstone326:
Awesome, thanks for doing this. Had my half-finished version of this sitting around for way too long.

PR #941, opened 2015-12-29
Lint MarkupUtil & upgrade htmlspanner to custom fork that is compat w/ latest htmlcleaner

phillipm:
htmlcleaner was bumped from 2.2 to 2.16 in https://github.com/dimagi/commcare-odk/pull/934. Turns out anything above 2.2 doesn't work with htmlspanner. Looked into it and it is fixed in [this fork](https://github.com/sdcr/HtmlSpanner). This PR includes an `aar` artifact of that fork to be included in the build so that we can use htmlcleaner.

Also add appropriate access restrictions to `MarkupUtil` methods and remove unused methods.


ctsims:
@wspride Now that there's no CSS do we actually use the HTML capabilities anywhere, out of curiosity?

phillipm:
We still use css as far as I know, like in https://github.com/dimagi/commcare-odk/blob/master/app/src/org/commcare/android/util/Stylizer.java

ctsims:
Hm, yeah, I think that was the CSS thing we don't use anymore. Will to
confirm.

PR #945, opened 2015-12-30
Update setup and test docs in README

phillipm:
Just set up the mobile dev env with @benrudolph today and am updating the docs with some relevant info.


benrudolph:
:raised_hands:

PR #947, opened 2015-12-31
General linting

phillipm:
Showed @benrudolph Android Studio yesterday and ran the linter on our code base. He said HQ didn't have as many lint warnings as us and that made me embarrassed for mobile.

Took this as an op to also remove lint warnings from our lint config that we don't need to pay attention to: https://github.com/dimagi/commcare-odk/pull/946

Review commit by commit.
Most are automated changes made by android studio. Worth a glance over because in rare cases android studio makes code edit mistakes.


app/src/org/commcare/util/externalizable/AndroidClassHasher.java
```diff
@@ -13,14 +13,13 @@
...
-    MessageDigest mMessageDigester;
+    private final MessageDigest mMessageDigester;
     
     public AndroidClassHasher() {
         try {
             mMessageDigester = java.security.MessageDigest.getInstance("MD5");
         } catch (NoSuchAlgorithmException e) {
-            // TODO Auto-generated catch block
-            e.printStackTrace();
+            throw new RuntimeException(e);
```
phillipm: In order to synchronize safely over `mMessageDigester` I made it `final`. That meant that I needed to somehow set `mMessageDigester` or abort when the `NoSuchAlgorithmException` is raised. I chose to abort, since the whole system probably won't work if we don't have a class hasher. 

Do you agree with my decision?

app/src/org/commcare/android/tasks/templates/CommCareTask.java
```diff
@@ -7,28 +7,28 @@
 /**
  * @author ctsims
  */
-public abstract class CommCareTask<A, B, C, R> extends ManagedAsyncTask<A, B, C> {
+public abstract class CommCareTask<Params, Progress, Result, Receiver> extends ManagedAsyncTask<Params, Progress, Result> {
```
phillipm: Changes in this file aren't from lint warnings. Rather this is a readability change I've been meaning to do for a long time.

benrudolph:
très bien! killer. had a few questions, but other than that g2g from me

phillipm:
Woohoo, thanks for the review on this annoying PR! I've addressed your comments.

app/src/org/commcare/util/externalizable/AndroidClassHasher.java
benrudolph: sounds right to me (could be useful to add some more context into that error cause `NoSuchAlgorithm` doesn't make sense to those who are new to this codebase)

PR #952, opened 2016-01-05
File system installer intermediate file cleanup

phillipm:
Remove need to write to temp file and then rename temp file to final destination. Instead just write to final destination using library calls instead of hand written stream processing methods.


app/src/org/commcare/android/resource/installers/FileSystemInstaller.java
```diff
@@ -71,32 +66,25 @@ public void cleanup() {
...
         try {
-            OutputStream os;
             Reference localReference;
 
             //Moved this up before the local stuff, in case the local reference fails, we don't want to start dealing with it
-            InputStream input;
-            try {
-                input = ref.getStream();
-            } catch (FileNotFoundException e) {
+            File inputFile = new File(ref.getLocalURI());
```
ctsims: Hey,

So this won't work, since the reference here could be remote. getLocalURI only works for local files, not network files.

PR #954, opened 2016-01-05
Don't re-consume one-time intent extras

amstone326:
Prevent re-consumption of one-time intent extras when an activity is launched from history

### DIFF
--- a/app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
+++ b/app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
@@ -134,6 +134,7 @@ public class CommCareHomeActivity
     // The API allows for external calls. When this occurs, redispatch to their
     // activity instead of commcare.
     private boolean wasExternal = false;
+    private static final String WAS_EXTERNAL_KEY = "was_external";
 
     private int mDeveloperModeClicks = 0;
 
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -141,10 +142,18 @@
...
 
+    private boolean loginExtraWasConsumed;
+    private static final String EXTRA_CONSUMED_KEY = "login_extra_was_consumed";
+
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
+        if (savedInstanceState != null) {
+            wasExternal = savedInstanceState.getBoolean(WAS_EXTERNAL_KEY);
```
phillipm: If you set `wasExternal` here then remove it from `processFromExternalLaunch`

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -274,15 +285,8 @@ public HomeActivityUIController getUiController() {
...
         super.onSaveInstanceState(outState);
-        outState.putBoolean("was_external", wasExternal);
-    }
-
-    @Override
-    protected void onRestoreInstanceState(Bundle inState) {
-        super.onRestoreInstanceState(inState);
-        if (inState.containsKey("was_external")) {
-            wasExternal = inState.getBoolean("was_external");
-        }
```
phillipm: good call

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
amstone326: Ah, I wasn't paying close enough attention when I refactored this; I think it makes more sense to leave it in processFromExternalLaunch so I'll remove it from here.

amstone326:
@phillipm Addressed comment.

PR #955, opened 2016-01-05
Ask for all app permissions at ccz app install time

phillipm:
Fixes 2 bugs introduced when we bumped our app to target Android 6 (https://github.com/dimagi/commcare-odk/pull/818)

 - We weren't asking for storage permissions at app install time, which meant you couldn't install apps on Android 6 devices. The install activity now asks for all permissions that CommCare might need. We also check for all perms at login time to ensure functionality is smooth once logged in. 
 - On devices running API 15 and below, we shouldn't ask for READ_EXTERNAL_STORAGE permissions because it isn't available at those API levels and the compat library seems to always return false when checking if the device has those permissions.

### DIFF
--- a/app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
+++ b/app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
@@ -25,11 +25,12 @@ import android.widget.Toast;
 
 import org.commcare.android.database.global.models.ApplicationRecord;
 import org.commcare.android.fragments.ContainerFragment;
-import org.commcare.android.fragments.SetupEnterURLFragment;
-import org.commcare.android.fragments.SelectInstallModeFragment;
 import org.commcare.android.fragments.InstallConfirmFragment;
+import org.commcare.android.fragments.SelectInstallModeFragment;
+import org.commcare.android.fragments.SetupEnterURLFragment;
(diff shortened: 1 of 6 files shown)


app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -814,6 +824,26 @@ public void onRequestPermissionsResult(int requestCode,
...
+                }
+            }
         }
     }
+
+    private void showMissingPermissionState() {
+        // TODO PLM: instead of popping up the same message we should disable
+        // install buttons and show a message and button to re-request the
+        // permissions.
+        Permissions.acquireAllAppPermissions(this, this,
```
amstone326: Why is the missing permission state to just re-request all permissions?

amstone326:
Good to merge after answer to 1 question

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
phillipm: The app won't work if it doesn't have these permissions, so for now, the lowest bar for this state is just re-ask for perms. The TODO outlines a future state that is more user friendly.

PR #956, opened 2016-01-05
Dramatic Performance Improvements during large app install/init

ctsims:
Ticket: http://manage.dimagi.com/default.asp?189554
X-Requested with: https://github.com/dimagi/commcare/pull/233

Pushes checking the install status to its own thread during install (could still be improved) and makes the suite parser skip adding resources to the table during startup, which was unnecessary and wasted an enormous amount of time for apps with lots of resources.


phillipm:
Looks good. I'll merge once https://github.com/dimagi/commcare/pull/233 is ready

ctsims:
Internal Release Note:

Fixed bug causing apps with massive (Thousands) of media files to take huge amounts of time to start up.

PR #958, opened 2016-01-06
Fix dialog rotation bug

amstone326:
Fix for http://manage.dimagi.com/default.asp?190442, which was introduced by https://github.com/dimagi/commcare-odk/pull/932. 

1st commit contains the fix, 2nd commit is just clarifying existing code.

### DIFF
--- a/app/src/org/commcare/android/framework/CommCareActivity.java
+++ b/app/src/org/commcare/android/framework/CommCareActivity.java
@@ -62,16 +62,16 @@ public abstract class CommCareActivity<R> extends FragmentActivity
 
     TaskConnectorFragment<R> stateHolder;
 
-    //fields for implementing task transitions for CommCareTaskConnector
+    // Fields for implementing task transitions for CommCareTaskConnector
     private boolean inTaskTransition;
 
     /**
      * Used to indicate that the (progress) dialog associated with a task
      * should be dismissed because the task has completed or been canceled.
      */
-    private boolean shouldDismissDialog = true;
+    private boolean dismissLastDialogAfterTransition = true;
 
-    protected AlertDialogFragment dialogToShowOnResume;
+    protected AlertDialogFragment alertDialogToShowOnResume;
 
     private GestureDetector mGestureDetector;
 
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/android/framework/CommCareActivity.java
```diff
@@ -303,9 +304,8 @@ protected int getWakeLockLevel() {
      * occurred while the activity was paused.
      */
     private void syncTaskBlockingWithDialogFragment() {
-        if (dialogId < 0) {
```
amstone326: @phillipm This is the fix -- don't just blindly dismiss upon resuming, because it messes with the natural lifecycle of re-showing a dialog for a task that is in progress after rotation. Instead, only do so if we have explicitly marked that we tried to dismiss the dialog while paused. (Note that I'm actually not sure this is even necessary; I'm not sure what made you add this check to the previous PR, would be curious to know. But assuming it is necessary, this way of doing it seems safer).

app/src/org/commcare/android/framework/CommCareActivity.java
```diff
@@ -62,16 +62,16 @@
...
-    //fields for implementing task transitions for CommCareTaskConnector
+    // Fields for implementing task transitions for CommCareTaskConnector
     private boolean inTaskTransition;
 
     /**
      * Used to indicate that the (progress) dialog associated with a task
      * should be dismissed because the task has completed or been canceled.
      */
-    private boolean shouldDismissDialog = true;
+    private boolean dismissLastDialogAfterTransition = true;
```
phillipm: :+1:

app/src/org/commcare/android/framework/CommCareActivity.java
phillipm: Can't remember why I decided to go this route.
I like your fix! 
_I think_ it is necessary to do this sort of check in case the activity is paused but the task finishes or is canceled.

phillipm:
Thanks for the fix and linking to the originating PR

PR #959, opened 2016-01-06
Check for session expiration in home screen dispatch

amstone326:
The DispatchActivity refactor left the home screen dispatch logic such that it was no longer checking for the case where the user has been logged out, which I believe we still want to be doing. This PR fixes this by checking for session expiration in attemptDispatchHomeScreen(), and just calling finish() if so. The specific case that I noticed where this is necessary is clearing user data (the app was crashing whenever you did this because CommCareHomeActivity would try to set up the home screen after returning from the preference activity, when there was no longer any logged in user). But I imagine there are other ways in which this could happen, and we would want to handle those smoothly as well, so went for a catch-all fix.

### DIFF
--- a/app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
+++ b/app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
@@ -969,6 +969,7 @@ public class CommCareHomeActivity
             this.finish();
         }
     }
+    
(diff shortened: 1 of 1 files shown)


phillipm:
Nice! Just encountered this an thought it was some sort of crazy android 6 permission issue. Glad you tracked it down

PR #960, opened 2016-01-06
Move play store launcher bugfix from home to dispatch activity

phillipm:
Move over fix for [this bug](https://code.google.com/p/android/issues/detail?id=2373) from the home activity, which used to be the activity launched upon app start, to the dispatch activity, which is now the launched activity.

I didn't test this because I don't really have context on it and it would require deploying to the play store.... @ctsims let me know if you know of a good way to test it, or if it simple enough to expect to work.


ctsims:
@phillipm So the old behavior was:

* Restart phone or Force Close CommCare.
* Start CommCare from the play store (IE: Hit the "Open" button on the play store page)
* Go to a menu or something
* hit home button
* Launch CommCare from app launcher

And you'd end up with two Stacks for the app. Let me know if those steps don't replicate. I can look for the ticket to see if it was device specific.

phillipm:
Nice, was able to reproduce on master and verify it is addressed by this PR.

PR #961, opened 2016-01-06
Setup temp storage dirs after gaining external write perms 

phillipm:
Android 6 related install issue where the temporary storage directory created in `CommCareApplication` wasn't actually being created because we didn't yet acquire external storage permissions from the user.

This PR simply calls the temp storage setup method again after the perms are acquired from the user.

Other changes are addressing no-opy code

### DIFF
--- a/app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
+++ b/app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
@@ -825,6 +825,9 @@ public class CommCareSetupActivity extends CommCareActivity<CommCareSetupActivit
                     }
                 }
             }
+            // external storage perms were enabled, so setup temp storage,
+            // which fails in application setup without external storage perms.
+            CommCareApplication._().prepareTemporaryStorage();
         }
     }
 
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -836,6 +825,9 @@ public void onRequestPermissionsResult(int requestCode,
                     }
                 }
             }
+            // external storage perms were enabled, so setup temp storage,
+            // which fails in application setup without external storage perms.
+            CommCareApplication._().prepareTemporaryStorage();
```
phillipm: Fix

app/src/org/commcare/dalvik/activities/CommCareSetupActivity.java
```diff
@@ -174,8 +174,6 @@ protected void onCreate(Bundle savedInstanceState) {
                     this.uiState = UiState.READY_TO_INSTALL;
                     //Now just start up normally.
                 }
-            } else {
-                incomingRef = this.getIntent().getStringExtra(KEY_PROFILE_REF);
```
phillipm: only place `KEY_PROFILE_REF` is ever used. So the result is always going to be `null`, so just remove it

PR #962, opened 2016-01-07
Unify instances of restarting commcare

amstone326:
We had a few different places in the app where we performed a restart of CommCare, and were doing it in slightly different (and in 1 case very different) ways in each. This unifies those instances to use 1 single helper method that does the restart in what (I think) is the most proper way.

@phillipm Also (mostly) fixes http://manage.dimagi.com/default.asp?186379, by performing the reboot after uninstall in a way that actually clears the back stack. This fixes the crash that was happening when you press back after being redirected to the install screen, but does NOT fix the fact that you are redirected to the install screen rather than back to the app manager..at least not entirely. This was previously a known shortcoming of the way I had set up restarting after uninstall, in that restarting CommCare consistently brought the app back to normal CommCare rather than the app manager, and I couldn't find a way around it. The odd thing is that after refactoring the way the restart is happening, it is now sometimes returning to the app manager, and sometimes returning to normal CommCare, and I can't really identify a pattern that is causing the distinction... Electing not to go crazy over that for now because it's not a huge deal, but will revisit later. 

**Question**: I tried to make sure I found all instances where we were doing a restart, but if anyone knows of something I missed, let me know please.

### DIFF
--- a/app/src/org/commcare/dalvik/activities/CrashWarningActivity.java
+++ b/app/src/org/commcare/dalvik/activities/CrashWarningActivity.java
@@ -2,7 +2,6 @@ package org.commcare.dalvik.activities;
 
 import android.app.Activity;
 import android.content.Intent;
-import android.os.Build;
 import android.os.Bundle;
 import android.view.View;
 import android.widget.Button;
@@ -12,6 +11,7 @@ import android.widget.TextView;
 
 import org.commcare.android.util.CommCareExceptionHandler;
 import org.commcare.dalvik.R;
+import org.commcare.dalvik.application.CommCareApplication;
 import org.javarosa.core.services.locale.Localization;
 
 /**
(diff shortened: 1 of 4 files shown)


app/src/org/commcare/dalvik/activities/CrashWarningActivity.java
```diff
@@ -54,7 +54,7 @@ private void setupButtons() {
         closeButton.setText(Localization.get("crash.warning.button"));
         closeButton.setOnClickListener(new View.OnClickListener() {
             public void onClick(View v) {
-                restartCommCare();
+                CommCareApplication._().restartCommCare(CrashWarningActivity.this);
```
phillipm: Calls to `restartCommCare` shouldn't be delegated through an instance (like `_()`) because it is a static method.

amstone326: @phillipm fixed

phillipm:
Cool, I'll look out for the weird restart issue where the activity restarted isn't always the same.

PR #963, opened 2016-01-07
Remove explicit redirects to DispatchActivity + fix incorrect invalid state

amstone326:
Fixes 2 multiple apps-related issues:

1) We should never be explicitly redirecting to DispatchActivity from another activity, since all activities are launched by DispatchActivity anyway and it messes up the back stack. In places where we want to direct back to dispatch for multiple apps navigation purposes, replace an explicit redirect with a call to finish().

2) After the DispatchActivity refactor, it is actually valid to be in the dispatch() method with usableAppsPresent() but no seated app; we should just attempt to seat an available app in this case.


phillipm:
Thanks for cleaning this up! I forgot to test how the dispatch refactor interacted with the multiple apps workflow

PR #965, opened 2016-01-08
Prevent swipe action when detail from persistent case tile is expanded

phillipm:
__Contains commits from https://github.com/dimagi/commcare-odk/pull/964 , review that one first__

When you expand a tabbed case detail from a persistent case tile, swipe actions are still registered for the screen hidden by the case detail. So if you are in form entry and swipe to change the tab, it also advances you in the form. Or if you swipe back it sometimes exits the current activity.

screen

This PR blocks swipes from dispatching at the `CommCareActivity` level if something is obstructing the main view.

Also remove some unused methods...


ctsims:
reviewed commit-by-commit and looks good other than my comment. It looks like there's a bug that removed the use of whether back navigation is available, but I don't think we should remove the declarations.

phillipm:
Okay, reverted that commit, should be good to go now

PR #966, opened 2016-01-08
UI controllers work with @ManagedUI + UI controller for login activity

amstone326:
Pre-factoring for PIN password work -- 2 major changes:
1. A CommCareActivity can now utilize *both* the @ManagedUI framework and a UIController. This involved making a CommCareActivityUIController interface, and adjusting the util methods in ManagedUiFramework.java 
2. LoginActivity now uses a ui controller (along with the @ManagedUI annotation that it was already using)

### DIFF
--- a/app/src/org/commcare/android/framework/CommCareActivityUIController.java
+++ b/app/src/org/commcare/android/framework/CommCareActivityUIController.java
@@ -1,0 +1,12 @@
+package org.commcare.android.framework;
+
+/**
+ * Created by amstone326 on 1/9/16.
+ */
+public abstract class CommCareActivityUIController {
+
+    protected abstract void setupUI();
+
+    public abstract void refreshView();
+
+}
--- a/app/src/org/commcare/android/framework/CommCareActivityUIController.java
+++ b/app/src/org/commcare/android/framework/CommCareActivityUIController.java
@@ -1,12 +1,16 @@
 package org.commcare.android.framework;
 
 /**
- * Created by amstone326 on 1/9/16.
+ * Interface to be implemented by any class acting as a ui controller for a CommCareActivity
+ *
+ * IMPORTANT: Any CommCareActivity that uses a CommCareActivityUIController must override
+ * CommCareActivity.initUIController() and CommCareActivity.getUIController()
+ *
+ * @author Aliza Stone (astone@dimagi.com)
  */
 public interface CommCareActivityUIController {
 
     void setupUI();
-
     void refreshView();
 
 }
--- a/app/src/org/commcare/android/framework/CommCareActivityUIController.java
+++ b/app/src/org/commcare/android/framework/CommCareActivityUIController.java
@@ -3,8 +3,8 @@ package org.commcare.android.framework;
 /**
  * Interface to be implemented by any class acting as a ui controller for a CommCareActivity
  *
- * IMPORTANT: Any CommCareActivity that uses a CommCareActivityUIController must override
- * CommCareActivity.initUIController() and CommCareActivity.getUIController()
+ * IMPORTANT: Any CommCareActivity that uses a CommCareActivityUIController must implement the
+ * WithUIController interface
  *
  * @author Aliza Stone (astone@dimagi.com)
  */
--- a/app/src/org/commcare/android/framework/CommCareActivityUIController.java
+++ b/app/src/org/commcare/android/framework/CommCareActivityUIController.java
@@ -3,10 +3,10 @@ package org.commcare.android.framework;
 /**
  * Created by amstone326 on 1/9/16.
  */
-public abstract class CommCareActivityUIController {
+public interface CommCareActivityUIController {
 
-    protected abstract void setupUI();
+    void setupUI();
 
-    public abstract void refreshView();
+    void refreshView();
 
 }
--- a/app/src/org/commcare/android/framework/ManagedUiFramework.java
+++ b/app/src/org/commcare/android/framework/ManagedUiFramework.java
@@ -37,7 +37,7 @@ public class ManagedUiFramework {
      */
     public static void loadUiElements(CommCareActivity activity) {
 
-        Class classHoldingFields = getClassHoldingFields(activity);
+        Class classHoldingFields = getObjectHoldingFields(activity).getClass();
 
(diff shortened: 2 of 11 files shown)


app/src/org/commcare/android/framework/CommCareActivityUIController.java
```diff
@@ -0,0 +1,16 @@
+package org.commcare.android.framework;
+
+/**
+ * Interface to be implemented by any class acting as a ui controller for a CommCareActivity
+ *
+ * IMPORTANT: Any CommCareActivity that uses a CommCareActivityUIController must override
```
amstone326: Not thrilled about not having a better way of enforcing this -- if someone has an idea of a better way to set this up, let me know.

phillipm: You should be able to make an interface called something like `WithUIController` that defines `initUIController` and `getUIController`. Then make the home activity, login activity, etc implement that interface. In the CommCareActivity you can then redefine `usesUIController()` to `return this instanceof WithUIController;` and make appropriate calls to the `WithUIController` interface with casts (`((WithUIController)this).getUIController()`). 

This seems _slightly_ messier with the casts and all, but enforces the implementation constraints that you have in the comments.

app/src/org/commcare/android/framework/ManagedUiFramework.java
```diff
@@ -122,15 +135,22 @@ private static boolean isFieldInBundle(String elementKey, Bundle bundle) {
...
                 try {
                     f.setAccessible(true);
                     try {
-                        View v = (View)f.get(activity);
+                        View v;
+                        if (activity.usesUIController()) {
+                            v = (View)f.get(activity.getUIController());
+                        } else {
+                            v = (View)f.get(activity);
+                        }
```
phillipm: I would lift this out of the for loop, such that you get the proper class instance once. It also sort of  coincides with the result of `getClassHoldingFields`, so maybe some of the setup work can be shared.

app/src/org/commcare/dalvik/activities/LoginActivity.java
```diff
@@ -128,66 +82,17 @@
...
-        sld.addState(StateSet.WILD_CARD, colorDrawable);
-
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) {
-            loginButton.setBackground(sld);
-        } else {
-            loginButton.setBackgroundDrawable(sld);
-        }
-
-        loginButton.setTextColor(getResources().getColor(R.color.cc_neutral_bg));
-    }
```
phillipm: So happy to see the LoginActivity getting smaller in size!

app/src/org/commcare/dalvik/activities/LoginActivity.java
```diff
@@ -265,40 +121,35 @@ public void onRequestPermissionsResult(int requestCode,
...
             }
         }
-        loginButton.setEnabled(true);
-        errorBox.setVisibility(View.GONE);
-        errorBox.setText("");
-    }
-
-    private boolean isRestoreSessionChecked() {
-        return restoreSessionCheckbox.isChecked();
+        uiController.setPermissionsGrantedState();
```
phillipm: :+1:

app/src/org/commcare/dalvik/activities/LoginActivityUIController.java
```diff
@@ -0,0 +1,336 @@
...
+    @UiElement(value=R.id.restore_session_checkbox)
+    private CheckBox restoreSessionCheckbox;
+
+    @UiElement(R.id.app_selection_spinner)
+    private Spinner spinner;
+
+    @UiElement(R.id.welcome_msg)
+    private TextView welcomeMessage;
+
+    private LoginActivity activity;
```
phillipm: This can be `final` and a few methods in this class can be marked `private`

phillipm:
Solid refactor!
One idea to making the ui controller framework you've setup potentially clearer is to replace usages of `CommCareActivity.usesUIController` with a method called something like `.getUIController` (that name is taken but I'm having trouble coming up with a more descriptive name) that either returns the activity itself or the ui controller if there is one being used. I say this because it seems like you test for the existence of a ui controller and then promptly use that info to either use the ui controller or the activity to do some work.

Left another suggestion along with this one. If they don't seem worth making, I'm happy to merge whenever.

app/src/org/commcare/android/framework/CommCareActivityUIController.java
amstone326: Cool, seems like a good solution.

amstone326:
@phillipm Am going to forego your last suggestion for 2 reasons: 1) Not sure how it would work for the same method to either return the ui controller or just a CommCareActivity, since the type signatures would have to be different. I think it would have to just return either the ui controller or null, at which point it wouldn't be as clean of a change. 2) More importantly: After refactoring to use the WithUIController interface, the call to initUIController() in CommCareActivity can only be made after a cast to WithUIController, which means that the check for safely casting to WithUIController cannot be dependent upon a uiController already being initialized (hope that made sense).

Made all the other changes you suggested though, so should be ready.

phillipm:
Hmm, can you look at this [commit](https://github.com/dimagi/commcare-odk/commit/9b49949a270d3d7995673889a777ca4e69215101) I made on a different branch? I think it makes the changes I've outlined without running into issues. Allows for the removal of a few helpers, so I vote for that implementation, but happy to merge whatever you prefer.

amstone326:
@phillipm Cool, hadn't really thought about the fact that all of the places where getUIManager() gets used wouldn't require casting, which makes that work in a much cleaner way. Cherry-picked the commit here.

app/src/org/commcare/android/framework/CommCareActivityUIController.java
```diff
@@ -0,0 +1,16 @@
...
+ * CommCareActivity.initUIController() and CommCareActivity.getUIController()
```
phillipm: Can you update this comment to say WithUIController interface instead of the distinct methods?

amstone326: done

PR #969, opened 2016-01-11
add markdown to list widgets

wpride:
Fix for http://manage.dimagi.com/default.asp?190325#1069642

Never got markdown support in for this ListWidget type class because I didn't know it existed:
https://confluence.dimagi.com/display/commcarepublic/Advanced+CommCare+ODK+Formatting#AdvancedCommCareODKFormatting-GridView

Ideally these three classes would probably be refactored to share some common ancestor like "AbstractSelectWidget" that would have these helper functions. Not sure that's worthwhile considering how little used they are, for now QuestionWidget is the common ancestor

### DIFF
--- a/app/src/org/odk/collect/android/widgets/QuestionWidget.java
+++ b/app/src/org/odk/collect/android/widgets/QuestionWidget.java
@@ -368,12 +368,12 @@ public abstract class QuestionWidget extends LinearLayout implements QuestionExt
         }
     }
 
-    public void setChoiceText(TextView choiceText, Vector<SelectChoice> choices, int index){
-        String markdownText = mPrompt.getSelectItemMarkdownText(choices.get(index));
+    public void setChoiceText(TextView choiceText, SelectChoice choice){
+        String markdownText = mPrompt.getSelectItemMarkdownText(choice);
         if(markdownText != null){
             choiceText.setText(forceMarkdown(markdownText));
         } else{
-            choiceText.setText(mPrompt.getSelectChoiceText(choices.get(index)));
+            choiceText.setText(mPrompt.getSelectChoiceText(choice));
         }
     }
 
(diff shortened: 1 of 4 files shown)


wpride:
Release Note: Markdown works with List Widgets

app/src/org/odk/collect/android/widgets/QuestionWidget.java
```diff
@@ -352,7 +353,27 @@ private void stripUnderlines(TextView textView) {
...
+            textView.setMovementMethod(LinkMovementMethod.getInstance());
+            // Wrap to the size of the parent view
+            textView.setHorizontallyScrolling(false);
+        } else {
+            textView.setText(mPrompt.getLongText());
+        }
+    }
+
+    public void setChoiceText(TextView choiceText, FormEntryPrompt prompt, int index){
+        String markdownText = mPrompt.getSelectItemMarkdownText(mPrompt.getSelectChoices().get(index));
```
ctsims: getSelectChoices here can involve a few callouts and should be considered an expensive method. I'd save this result inline to prevent it from being re-fired on like 374. Good otherwise.

ctsims:
Note: Don't pull until above comment is addressed.

app/src/org/odk/collect/android/widgets/QuestionWidget.java
```diff
@@ -352,7 +355,27 @@ private void stripUnderlines(TextView textView) {
...
+            textView.setText(forceMarkdown(prompt.getMarkdownText()));
+            textView.setMovementMethod(LinkMovementMethod.getInstance());
+            // Wrap to the size of the parent view
+            textView.setHorizontallyScrolling(false);
+        } else {
+            textView.setText(mPrompt.getLongText());
+        }
+    }
+
+    public void setChoiceText(TextView choiceText, Vector<SelectChoice> choices, int index){
```
phillipm: What about changing the signature to `setChoiceText(TextView choiceText, SelectChoice choice)` and then invoking it like `setChoiceText(label, mItems.get(i)`? 
It seems a bit strange to me to pass in all the choices and the index and deref the choice list inside of the method twice.

PR #971, opened 2016-01-12
Refactor EntitySelectActivity

phillipm:
Pretty light-weight refactor of EntitySelectActivity: basically lifting code into helper methods.


app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -202,49 +195,61 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        inAwesomeMode = true;
+
+        rightFrame = (FrameLayout)findViewById(R.id.screen_compound_select_right_pane);
+
+        TextView message = (TextView)findViewById(R.id.screen_compound_select_prompt);
+        //use the old method here because some Android versions don't like Spannables for titles
+        message.setText(Localization.get("select.placeholder.message", new String[]{Localization.get("cchq.case")}));
+    }
 
+    private void setupPortraitDualPaneView(boolean isOrientationChange) {
```
amstone326: Why is this called a DualPaneView? I don't think we ever show 2 panes in portrait mode

phillipm: We do in "AwesomeMode"
I was trying to give it a more descriptive name

amstone326: I was pretty sure awesome mode only happens in landscape mode.. Is that wrong?

phillipm: Oh, right: When awesome mode is possible but we are in portrait mode, we still need to do different setup compared to when we aren't in awesome mode and in portrait mode.

amstone326: Agreed, but it's misleading to call that dualPaneView because it will never show 2 panes in portrait mode. The only difference with the setup in portrait mode when awesome mode is possible is that we're checking if, when previously in awesome mode, we had something selected, and if so calling out to the Detail Activity. It doesn't change what the portrait mode view actually looks like.

phillipm: :+1:

PR #974, opened 2016-01-13
Location sort

forest-tong:


app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -1158,40 +1160,35 @@ protected boolean onBackwardSwipe() {
...
-
-    private void triggerRebuild() {
+    public void loadEntities() {
         if (loader == null && !EntityLoaderTask.attachToActivity(this)) {
-            EntityLoaderTask theloader = new EntityLoaderTask(shortSelect, asw.getEvaluationContext());
-            theloader.attachListener(this);
-            theloader.execute(selectDatum.getNodeset());
+            EntityLoaderTask entityLoader = new EntityLoaderTask(shortSelect, asw.getEvaluationContext());
+            entityLoader.attachListener(this);
+            entityLoader.execute(selectDatum.getNodeset());
```
phillipm: :+1:

app/src/org/commcare/android/models/NodeEntityFactory.java
```diff
@@ -1,5 +1,9 @@
 package org.commcare.android.models;
 
+import android.content.Context;
```
phillipm: No longer needed

app/src/org/commcare/android/models/NodeEntityFactory.java
```diff
@@ -16,6 +20,7 @@
 public class NodeEntityFactory {
     private boolean mEntitySetInitialized = false;
     private static final Object mPreparationLock = new Object();
+    private HereFunctionHandler hereFunctionHandler;
```
phillipm: Can be marked as `final`

app/src/org/commcare/android/models/NodeEntityFactory.java
```diff
@@ -34,6 +40,8 @@ public Detail getDetail() {
         getDetail().populateEvaluationContextVariables(nodeContext);
         //return new AsyncEntity<TreeReference>(detail.getFields(), nodeContext, data);
 
+        nodeContext.addFunctionHandler(hereFunctionHandler);
```
phillipm: I would consider writing: 
`nodeContext.addFunctionHandler(EntitySelectActivity.hereFunctionHandler);`
which would allow you to eliminate the `hereFunctionHandler` class field.

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -161,6 +162,8 @@
     private boolean cancelled;
     private ContainerFragment<EntityListAdapter> containerFragment;
 
+    public static HereFunctionHandler hereFunctionHandler = new HereFunctionHandler();
```
phillipm: Can be marked `final`

phillipm:
@damagatchi retest this

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -481,13 +486,8 @@ private void refreshView() {
...
 
-            if (adapter == null &&
-                    loader == null &&
-                    !EntityLoaderTask.attachToActivity(this)) {
-                EntityLoaderTask theloader =
-                        new EntityLoaderTask(shortSelect, asw.getEvaluationContext());
-                theloader.attachListener(this);
-                theloader.execute(selectDatum.getNodeset());
+            if (adapter == null) {
+                loadEntities();
```
phillipm: Good call!

app/src/org/commcare/dalvik/geo/HereFunctionHandler.java
```diff
@@ -0,0 +1,149 @@
...
+        Vector p = new Vector();
+        p.addElement(new Class[0]);
+        return p;
+    }
+
+    public boolean rawArgs() {
+        return false;
+    }
+
+    public boolean realTime() {
```
phillipm: I recently removed this from the IFunctionHandler interface, so you can remove this implementation, since it will never be called

app/src/org/commcare/dalvik/geo/HereFunctionHandler.java
```diff
@@ -0,0 +1,149 @@
...
+    @Override
+    public void onStatusChanged(String provider, int status, Bundle extras) {}
+
+    // Methods for implementing IFunctionHandler
+
+    public String getName() {
+        return HERE_NAME;
+    }
+
+    public Vector getPrototypes() {
```
phillipm: Will you add `@Override` annotations to all the methods that implement interfaces or override super classes.
Helps with code readability

app/src/org/commcare/dalvik/geo/HereFunctionHandler.java
```diff
@@ -0,0 +1,149 @@
...
+import org.javarosa.core.model.data.GeoPointData;
+import org.odk.collect.android.utilities.GeoUtils;
+
+import java.util.Set;
+import java.util.Vector;
+
+/**
+ * HereFunctionHandler is a class that can be used to evaluate the here() XPath expression.
+ * In any context where here() needs to be evaluated, create a HereFunctionHandler and
+ * call addFunctionHandler to add it to the XPath evaluator.
```
phillipm: My approach when writing comments is to:
 - be concise. "HereFunctionHandler is a class that can be" can be inferred by the surrounding context
 - avoid describing how code is used. "call addFunctionHandler to add it ..." might get stale if we change the code but forget to update this comment

My preference would be something along the lines of "Allows evaluation contexts to support the here() XPath function, which returns the current location."

Feel free to ignore this and write comments however you feel proper; just wanted let you what I like :)

app/src/org/commcare/dalvik/geo/HereFunctionHandler.java
```diff
@@ -0,0 +1,149 @@
...
+
+    public void registerEntitySelectActivity(EntitySelectActivity entitySelectActivity) {
+        this.entitySelectActivity = entitySelectActivity;
+    }
+
+    // Setting up and subscribing to the location manager is delayed until here() is actually
+    // evaluated.
+    public Object eval(Object[] args, EvaluationContext ec) {
+        if(!locationUpdatesRequested) requestLocationUpdates();
+        if(location == null) return "";
```
phillipm: One of the coding styles we try to adhere to is to always use braces for conditional bodies. Though more verbose it makes it easier to add `else` statements and is consistent with the rest of the codebase

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -502,6 +506,8 @@ protected void onPause() {
         if (adapter != null) {
             adapter.unregisterDataSetObserver(mListStateObserver);
         }
+
+        hereFunctionHandler.stopLocationUpdates();
```
phillipm: One tricky thing about stopping in `onPause` is that you may want to resume later and continue acquiring a location. For instance, if you are getting a location and switch to another app for a sec and switch back, you will have paused and resumed this activity, but there is no restart code for the location acquisition logic. Doing this cleanly is sorta tricky, we can talk further about possible approaches.

app/src/org/commcare/dalvik/geo/HereFunctionHandler.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    @Override
+    public void onLocationChanged(Location location) {
+        // Do we need to check the accuracy of the location?
+        this.location = toGeoPointData(location);
+        Log.i("HereFunctionHandler", "location has been set to " + this.location.getDisplayText());
+        // Trigger refreshing of entity list.
+        if (entitySelectActivity != null) {
+            entitySelectActivity.loadEntities();
```
phillipm: Might eventually want to only do this reloading when the final location is acquired. That is, refresh only when we have acquired a satisfactorily accurate location.

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -277,6 +284,8 @@ public void onClick(View v) {
         if (!isUsingActionBar()) {
             searchbox.setText(lastQueryString);
         }
+
+        hereFunctionHandler.registerEntitySelectActivity(this);
```
phillipm: It is good practice in android to perform listener registration in `onResume` and then unregister the same listener in `onPause`. This helps prevent memory leaks due to dangling references.

app/src/org/commcare/dalvik/geo/HereFunctionHandler.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    private void requestLocationUpdates() {
+        mLocationManager = (LocationManager) context.getSystemService(Context.LOCATION_SERVICE);
+        Set<String> mProviders = GeoUtils.evaluateProviders(mLocationManager);
+
+        for (String provider : mProviders) {
+            if ((provider.equals(LocationManager.GPS_PROVIDER) && ContextCompat.checkSelfPermission(context, Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED) ||
+                    (provider.equals(LocationManager.NETWORK_PROVIDER) && ContextCompat.checkSelfPermission(context, Manifest.permission.ACCESS_COARSE_LOCATION) == PackageManager.PERMISSION_GRANTED)) {
+                // This is non-null if the calling activity pauses.
```
phillipm: I don't really understand this comment

app/src/org/commcare/dalvik/geo/HereFunctionHandler.java
```diff
@@ -0,0 +1,145 @@
...
+                if (location == null) {
+                    Location lastKnownLocation = mLocationManager.getLastKnownLocation(provider);
```
phillipm: what if `lastKnownLocation` is null?

phillipm:
Solid! You've managed to add new functionality to one of the messiest parts of our code :clap: 

After you address my comments let's talk about next steps. There are several things that will make this feature more usable:
 - only triggering location acquisition when the case list has location data.
 - stop refining the location after a certain accuracy is obtained
 - stop trying to get the location if the user moves away from the case list activity
 - caching the last location and using that during case list loading until the case list is done loading and we can reload using a more accurate location. Doing so will ensure that all cases report consistent distances, even if the distances are calculated against a less accurate location.

app/src/org/commcare/dalvik/geo/HereFunctionHandler.java
amstone326: Going to be super nitpicky and also throw in putting a space between if and (

phillipm: Yeah, might be good to setup our codestyle config in Android Studio, which will help you do this automatically (ctrl-alt-l in Android Studio). See https://confluence.dimagi.com/display/MD/Code+Conventions#CodeConventions-AndroidStudioCodingConventionSettings. I can help you out with it, because it is a little finicky.

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
forest-tong: The entities are reloaded in `onResume`, which calls, `eval`, which then calls `HereFunctionHandler.requestLocationUpdates`. Doesn't that restart location acquisition?

app/src/org/commcare/dalvik/geo/HereFunctionHandler.java
forest-tong: Basically what I meant to point out is that if `eval` is never called, then the location is never requested. But I already mentioned this at the top, and I guess it's pretty self-explanatory, so I'll take out the comment.

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
phillipm: Yeah, it depends on whether `loader` is `null` or not and if there are any pending tasks in `EntityLoaderTask`. I don't fully understand what is going on because that is a very circular piece of logic. I think that part of the code is trying to ensure entity loading continues if the entity select activity is destroyed temporarily (on screen rotation)

PR #979, opened 2016-01-15
PIN auth

amstone326:
cross: https://github.com/dimagi/javarosa/pull/230, https://github.com/dimagi/commcare/pull/243

How to run/test this:
-Install and log into an app, go into Developer Preferences and turn on "Give option to use PIN for login" (this is turned off by default, and individual domains will be able to set a custom field in their app properties on HQ to enable it).
-Log out again. The next time you log in, a dialog will pop up offering you the option to set a PIN now, set one later, or not to ask again. 
-Selecting set a PIN will take you to the pin create screen. On that same screen, the options menu contains a "Remember PW For Next Login" option. If you select this and confirm, you will be logged out immediately, and the next login will be primed.
-At any time (if the PIN setting is turned on), you can also access the PIN create screen via the home screen options menu. If you navigate to the PIN create screen this way, you will first be required to authenticate yourself with either your current PIN if you have one, or your password if not.


app/src/org/commcare/android/tasks/ManageKeyRecordTask.java
```diff
@@ -118,100 +132,113 @@ protected void deliverError(R receiver, Exception e) {
...
-            } else if(record.getType() == UserKeyRecord.TYPE_NEW) {
-                //See if we have another sandbox with this ID that is fully initialized.
-                if(storage.getIDsForValues(new String[] {UserKeyRecord.META_SANDBOX_ID, UserKeyRecord.META_KEY_STATUS}, new Object[] {record.getUuid(), UserKeyRecord.TYPE_NORMAL}).size() > 0) {
-                    
-                    Logger.log(AndroidLogger.TYPE_MAINTENANCE, "Marking new sandbox " + record.getUuid() + " as initialized, since it's already in use on this device");
-                    //If so, this sandbox _has_ to have already been initialized, and we should treat it as such.
-                    record.setType(UserKeyRecord.TYPE_NORMAL);
-                    storage.write(record);
+
+        for (UserKeyRecord normalRecord :
```
amstone326: just refactored this to use getRecordsForValue() for each TYPE, rather than using if statements to check for each type

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -270,149 +387,166 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
-                else if(resultCode == RESULT_OK){
-                    CommCareApplication._().notifyLogsPending();
-                    uiController.refreshView();
-                    return;    
-                }
-            case GET_INCOMPLETE_FORM:
-                //TODO: We might need to load this from serialized state?
-                if(resultCode == RESULT_CANCELED) {
-                    uiController.refreshView();
+                case PREFERENCES_ACTIVITY:
```
amstone326: just fixing tabbing

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -270,149 +387,166 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
+                        }
+                    }
+                    break;
+                case MODEL_RESULT:
+                    boolean fetchNext = processReturnFromFormEntry(resultCode, intent);
+                    if (!fetchNext) {
+                        return;
+                    }
+                    break;
+                case AUTHENTICATION_FOR_PIN:
```
amstone326: actual new cases start here

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -210,7 +217,116 @@ private void processFromLoginLaunch() {
...
+                                .edit()
+                                .putBoolean(CommCarePreferences.HAS_DISMISSED_PIN_CREATION, true)
+                                .commit();
+                        showPinFutureAccessDialog();
+                    }
+                });
+
+
+        dialog.setChoiceItems(new DialogChoiceItem[]{createPinChoice, nextTimeChoice, notAgainChoice});
+        dialog.addCollapsibleInfoPane(Localization.get("pin.dialog.extra.info"));
```
amstone326: User can click on a small icon button to see an explanation of what it means to set a PIN

app/res/layout/create_pin_view.xml
```diff
@@ -0,0 +1,52 @@
...
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:id="@+id/pin_prompt_text"
+        android:layout_centerHorizontal="true"
+        android:layout_margin="@dimen/standard_spacer_large"
+        android:textSize="@dimen/text_large" />
+
+    <EditText
+        android:id="@+id/pin_entry"
+        android:layout_height="60dp"
```
ctsims: seems oddly hardcoded to me.

app/res/layout/create_pin_view.xml
```diff
@@ -0,0 +1,52 @@
...
+        android:layout_height="wrap_content"
+        android:orientation="horizontal"
+        android:layout_alignParentBottom="true"
+        android:weightSum="2" >
+
+        <Button
+            android:layout_width="wrap_content"
+            android:layout_height="@dimen/min_button_height"
+            android:id="@+id/pin_cancel_button"
+            android:textSize="@dimen/font_size_dp_medium"
```
ctsims: We should default to using "font_size_sp_medium", etc. SP is basically the same as DP except when the user changes their phone's font size it respects the difference.

app/res/layout/pin_auth_view.xml
```diff
@@ -0,0 +1,66 @@
...
+
+    <TextView
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:id="@+id/auth_prompt_text"
+        android:layout_centerHorizontal="true"
+        android:layout_margin="@dimen/standard_spacer_large"
+        android:textSize="@dimen/text_large" />
+
+    <EditText
```
ctsims: You might be able to make this similarity easier by separating out the common elements here into their own "component" layout and using an <include> to bring them in. Will keep things the same without replicating code. screen_login.xml does that with the top banner if you want an example.

app/res/layout/screen_login.xml
```diff
@@ -86,21 +86,19 @@
                                 android:textSize="@dimen/text_medium"
                                 android:visibility="gone"/>
 
-                            <EditText
+                            <AutoCompleteTextView
```
ctsims: did we determine that there was no spinner that would do this based on concrete values? Autocomplete doesn't necessarily seem significantly better to me, since you still have to start typing, and it covers up the view below

app/res/layout/screen_login.xml
```diff
@@ -142,6 +140,14 @@
                                 android:textSize="16dp"
                                 android:visibility="gone" />
 
+                            <TextView
+                                android:id="@+id/primed_password_message"
+                                android:layout_width="fill_parent"
+                                android:layout_height="wrap_content"
+                                android:layout_margin="@dimen/standard_spacer"
+                                android:textSize="16dp"
```
ctsims: hard coded text size

app/src/org/commcare/android/crypt/CryptUtil.java
```diff
@@ -78,58 +80,44 @@ private static Cipher decodingCipher(String password) throws NoSuchAlgorithmExce
...
-        } catch (InvalidKeyException e) {
+        } catch (NoSuchAlgorithmException | InvalidKeySpecException | InvalidKeyException e) {
             throw new RuntimeException(e);
         } catch (NoSuchPaddingException e) {
             return null;
         }
     }
 
+
+    public static byte[] unWrapKey(byte[] wrapped, String password) {
```
ctsims: Do we need this wrapper v. just renaming? I think it's fine to assume that we dont' need to maintain the API for this method strongly

app/src/org/commcare/android/database/app/AppDatabaseUpgrader.java
```diff
@@ -140,4 +157,74 @@ private boolean upgradeFiveSix(SQLiteDatabase db) {
...
+            } else {
+                // Otherwise, sort the records in decreasing order of validTo date, and then mark
+                // the first one in the list as active
+                Collections.sort(records, new Comparator<UserKeyRecord>() {
+                    @Override
+                    public int compare(UserKeyRecord lhs, UserKeyRecord rhs) {
+                        return lhs.getValidTo().compareTo(rhs.getValidTo());
+                    }
+                });
+                records.get(0).setActive();
```
ctsims: This method needs to actually commit the record as well, right?

app/src/org/commcare/android/tasks/ManageKeyRecordTask.java
```diff
@@ -37,15 +39,17 @@
...
  * @author ctsims
  *
  */
 public abstract class ManageKeyRecordTask<R> extends HttpCalloutTask<R> {
-    final String username;
-    final String password;
+    private final String username;
+    private String password;
+    private String pin;
+    private final LoginActivity.LoginMode loginMode;
```
ctsims: I would probably introduce this coupling in the other direction or put it somewhere neutral (Just moving the enum definition that is). In general it's better it Tasks don't 'know' about activities

app/src/org/commcare/android/tasks/ManageKeyRecordTask.java
```diff
@@ -275,24 +301,35 @@ protected boolean processSuccesfulRequest() {
...
                     ukr.setID(existing.getID());
                     storage.write(ukr);
-                } catch(NoSuchElementException nsee) {
-                    //If there's no existing record, write this new one (we'll handle updating the status later)
+                } catch (NoSuchElementException nsee) {
+                    // If there's no existing record, write this new one (we'll handle updating the status later)
                     storage.write(record);
                 }
+
+                markOldRecordsInactive(storage, record.getUsername(), record.getUuid());
```
ctsims: I think this can be either "record" _or_ "ukr/existing" from above. In the non-catch block, "record" never receives a recordId (seems like that should be what determines whether the record isn't marked inactive?)

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -210,7 +217,116 @@ private void processFromLoginLaunch() {
             if (CommCareHomeActivity.isDemoUser()) {
                 showDemoModeWarning();
             }
+
+            pinModeDecisionLogic();
+        }
+    }
+
+    // See if we should launch the pin choice dialog
+    private void pinModeDecisionLogic() {
```
ctsims: This method name feels really awkward to me

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -210,7 +217,116 @@ private void processFromLoginLaunch() {
...
+
+        LoginActivity.LoginMode loginMode = LoginActivity.LoginMode.fromString(
+                getIntent().getStringExtra(LoginActivity.LOGIN_MODE));
+
+        if (loginMode == LoginActivity.LoginMode.PRIMED) {
+            launchPinCreateScreen(loginMode);
```
ctsims: If this is going to hit another activity it's probably a good idea to return afterwards to make sure the workflow doesn't get hijacked.

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
ctsims: I'd avoid passing the password through an intent, it results in the object being marshalled into a memory space we can't control and could expose info to a sneaky malicious process

app/src/org/commcare/dalvik/activities/LoginActivity.java
```diff
@@ -381,40 +391,43 @@ public void keysDoneOther(LoginActivity r, HttpCalloutOutcomes outcome) {
...
     
     private void done() {
         ACRAUtil.registerUserData();
-
         CommCareApplication._().clearNotifications(NOTIFICATION_MESSAGE_LOGIN);
 
         Intent i = new Intent();
+        i.putExtra(LOGIN_MODE, uiController.getLoginMode().toString());
+        if (uiController.getLoginMode() == LoginMode.PASSWORD) {
+            i.putExtra(PASSWORD_FROM_LOGIN, uiController.getEnteredPasswordOrPin());
```
ctsims: Same comment as above, we should avoid passing PW's around through intents

app/src/org/commcare/dalvik/activities/LoginActivity.java
```diff
@@ -606,4 +630,35 @@ public CommCareActivityUIController getUIController() {
...
+        PRIMED("primed-mode");
+
+        private String stringVersion;
+
+        LoginMode(String s) {
+            this.stringVersion = s;
+        }
+
+        @Override
+        public String toString() {
```
ctsims: Enums should be able to do this by default, no?

app/src/org/commcare/dalvik/activities/PinAuthenticationActivity.java
```diff
@@ -0,0 +1,173 @@
...
+import org.commcare.android.database.app.models.UserKeyRecord;
+import org.commcare.android.framework.ManagedUi;
+import org.commcare.android.framework.SessionAwareCommCareActivity;
+import org.commcare.android.framework.UiElement;
+import org.commcare.dalvik.R;
+import org.commcare.dalvik.application.CommCareApplication;
+import org.javarosa.core.services.locale.Localization;
+
+/**
+ * When a user selects Set/Reset PIN from the home activity options menu, they must first
```
ctsims: I wonder whether we truly need the capacity to check someone's PIN/PW after they've logged in. I know it's super common, but it's kind of awkward if you just logged someone in to change their PIN and you need to enter the PW twice, right? Maybe good to double check with product (fine to leave for now until we can UAT)

app/src/org/commcare/dalvik/application/CommCareApplication.java
```diff
@@ -1097,6 +1097,15 @@ public CommCareSessionService getSession() throws SessionUnavailableException {
         }
     }
 
+
+    public UserKeyRecord getRecordForCurrentUser() {
+        try {
+            return getSession().getUserKeyRecord();
+        } catch (SessionUnavailableException e) {
+            return null;
```
ctsims: Will this ever be ok? Not quite sure why this just wouldn't throw a SessionUnavailableException as part of the method signature

app/src/org/commcare/dalvik/dialogs/PaneledChoiceDialog.java
```diff
@@ -89,4 +90,30 @@ public void setOnDismissListener(DialogInterface.OnDismissListener listener) {
     public void dismiss() {
         dialog.dismiss();
     }
+
+    public void addCollapsibleInfoPane(String messageContent) {
```
ctsims: I love this being part of the framework! Can definitelyimagine this being super helpful.

app/src/org/commcare/dalvik/dialogs/PaneledChoiceDialog.java
```diff
@@ -89,4 +90,30 @@ public void setOnDismissListener(DialogInterface.OnDismissListener listener) {
...
+        extraInfoButton.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                toggleExtraInfoVisibility();
+            }
+        });
+    }
+
+    private void toggleExtraInfoVisibility() {
+        TextView extraInfoContent = (TextView)view.findViewById(R.id.extra_info_content);
```
ctsims: It'd be nice to animate this if we can. There's pretty portable logic to do so in the ODKView Class (possibly in CommCareActivity, actually)

app/src/org/commcare/dalvik/services/CommCareSessionService.java
```diff
@@ -83,6 +83,7 @@
     private final Object lock = new Object();
 
     private User user;
+    private int userKeyRecordID;
```
phillipm: Does this need to be refreshed in `setCurrentUser`?

phillipm: Oh, think I misunderstood this part of the code, ignore ^

app/res/layout/create_pin_view.xml
amstone326: @ctsims So the reason I did this was because a while back I ran into a bug where using sp font sizes was producing really undesirable behavior in the alert dialogs on some devices. This was the PR that fixed that: https://github.com/dimagi/commcare-odk/pull/754. Do you have any idea what might have been the reason for that? I've been hesitant about using sp for font size ever since.

amstone326: ..Oh no. I just thought about what you said. Does that mean that the issue was likely just that the specific tablet with the issue had its font size set to something weird?

app/res/layout/screen_login.xml
amstone326: As far as I could find in my searching, Android doesn't offer an out-of-the-box combobox. I'm sure I could make it custom in some way, but it would have taken a lot longer and it didn't seem worth the time right now. Especially given how tangential this is to the actual feature.

amstone326: I can remove the autocomplete entirely for now if we don't think it's a worthwhile addition; I agree that I don't know how much value it really adds.

app/src/org/commcare/android/crypt/CryptUtil.java
amstone326: Not totally sure what you're asking. unwrapByteArrayWithString() is a generic method used in 2 places: here (to unwrap an AES key with a password), and then to unwrap a password with a PIN. So I just kept this method named for its specific case for consistency/clarity. I could just call unwrapByteArrayWithString() directly, but not sure why that's better.

app/src/org/commcare/android/database/app/AppDatabaseUpgrader.java
amstone326: Hm yea I would think so, but I tested this migration before and it worked with PIN functionality, which it shouldn't have if this change didn't register, because then there wouldn't have been any active records found in LoginActivityController.getActiveRecordForUsername(). I fixed it anyway, but is there any reason it would have been working before?.. Otherwise I did something super weird.

amstone326: Ok nvm, definitely just did not test what I thought I did. Thanks for catching.

app/src/org/commcare/android/tasks/ManageKeyRecordTask.java
amstone326: moved the enum to its own file

app/src/org/commcare/android/tasks/ManageKeyRecordTask.java
amstone326: Well I'm using the uuid, not the recordId, to determine a match. Which seems OK because the whole point is that we won't have 2 UKRs with the same username AND same sandbox ID, we just might have some with other sandbox IDs. Is there a reason to switch to using recordID?

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
amstone326: @ctsims Yea I was concerned about that initially but talked with Phillip about it and neither of us could really come up with a better strategy. The passing of the un-hashed password here and in the other place you flagged is obviously essential to being able to set a PIN. What alternative would you suggest?

app/src/org/commcare/dalvik/activities/PinAuthenticationActivity.java
amstone326: @ctsims This is actually necessary. In general, whenever we want to assign a new PIN to a user, we need to have access to the un-hashed password in order to do so. When the user selects Set/Reset PIN from the options menu, we are at a point where we no longer have access to that un-hashed password (since they could go into the options menu at any point, well after the initial login), and therefore need to require them to enter either their PW or current PIN, in order to get it. 

The use case you're describing of having just logged in and then needing to re-enter your password shouldn't really come up because if you've logged in in password mode, and the PIN setting is enabled, the pin choice dialog will show up immediately upon login, and if you select "Yes, create a pin" from THAT screen, you don't have to re-auth. You only get shown the PinAuthenticationActivity if you select Set/Reset PIN from the home activity options menu.

app/src/org/commcare/dalvik/activities/LoginActivity.java
amstone326: Ok yea was overcomplicating that

app/src/org/commcare/dalvik/application/CommCareApplication.java
amstone326: Definitely wouldn't be OK, but I guess the reason I wasn't throwing it is because it would make the code in a lot of the places that use this method very awkward. The places that this gets called are all places that we absolutely should have an active session, and if we don't, there are going to be plenty of other problems anyway, and it wouldn't even be clear what to do if the exception was caught there. This does return null which is really just another version of the same error state that the caller can check for if it makes sense to (and some of them do). Do you think it is important to throw the exception instead?

app/src/org/commcare/dalvik/dialogs/PaneledChoiceDialog.java
amstone326: Not sure what you mean by animate.

app/src/org/commcare/dalvik/activities/PinAuthenticationActivity.java
amstone326: Worth noting that depending upon what you propose as an alternative to passing the un-hashed password through from login (in place of doing it via the intent), it's possible that we could still have it around from the initial login. Although it seems like that wouldn't be a great situation, and we should probably wipe the password from wherever we cache it as soon as possible no matter what, instead of allowing it to persist indefinitely just so the user doesn't have to re-auth here.

amstone326:
@ctsims OK addressed all comments with either a change, or a question/comment back if I didn't agree or didn't understand.

app/src/org/commcare/dalvik/dialogs/PaneledChoiceDialog.java
phillipm: Probably referring to something similar to the ability to expand persistent case tiles to show the case detail. Doing so slides the detail down in an animation.
See https://github.com/dimagi/commcare-odk/blob/master/app/src/org/commcare/android/framework/BreadcrumbBarFragment.java#L182-L204

phillipm:
Would be awesome to get some screenshots going

app/res/layout/screen_login.xml
ctsims: It's easy to turn off/on, right? My guess is that based on what I've seen it's more likely to cause state-confusion than help, so let's turn it off for now and think about how we want to manage this input in the future.

app/src/org/commcare/android/crypt/CryptUtil.java
ctsims: I think it's a bit more confusing to have two methods with identical signatures that do the same thing, personally. It makes it seem like one of them must do something the other isn't.

app/src/org/commcare/android/tasks/ManageKeyRecordTask.java
ctsims: Makes sense in that context, no need to change

app/src/org/commcare/android/database/app/models/UserKeyRecord.java
```diff
@@ -208,6 +267,56 @@ public boolean isPasswordValid(String password) {
...
+        if (pin != null) {
+            return isPinValid(pin);
+        } else {
+            return isPasswordValid(password);
+        }
+    }
+
+    public static byte[] wrapPasswordWithPin(String password, String pin) {
+        return CryptUtil.wrapByteArrayWithString(password.getBytes(), pin);
+    }
```
phillipm: This can be `private`. It is only called once, I would remove it and just do `passwordWrappedByPin = CryptUtil.wrapByteArrayWithString(password.getBytes(), pin);` in `assignPinToRecord`. The name of the var you are assigning the result of `CryptUtil.wrapByteArrayWithString` to gives enough context that you don't need to create a helper method

app/src/org/commcare/android/database/app/models/UserKeyRecord.java
```diff
@@ -208,6 +267,56 @@ public boolean isPasswordValid(String password) {
...
+
+    public static String unwrapPasswordWithPin(byte[] wrappedPassword, String pin) {
```
phillipm: can be `private`. Consider moving into code to `getUnhashedPasswordViaPin` since that is the only place it is called and the only thing that `getUnhashedPasswordViaPin` does is call this helper.

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
ctsims: Once someone logs in we keep their password cached currently so we can do web requests with it. You can retrieve it from CommCareApplication._().getSession().getLoggedInUser()

app/src/org/commcare/dalvik/activities/PinAuthenticationActivity.java
ctsims: Responded in the other place. We have access to the user's password in the local context for the length of their session so we can perform web requests with it, so we should just use that cached value.

app/src/org/commcare/android/database/app/AppDatabaseUpgrader.java
```diff
@@ -140,4 +158,80 @@ private boolean upgradeFiveSix(SQLiteDatabase db) {
...
+        // First, create a mapping from username --> list of UKRs
+        Map<String, List<UserKeyRecord>> usernamesToRecords = new HashMap<>();
+        for (UserKeyRecord record : migratedRecords) {
+            String username = record.getUsername();
+            List<UserKeyRecord> recordsForUsername = usernamesToRecords.get(username);
+            if (recordsForUsername == null) {
+                recordsForUsername = new ArrayList<>();
+            }
+            recordsForUsername.add(record);
+            usernamesToRecords.put(username, recordsForUsername);
```
phillipm: the `put` call is only needed inside of `recordsForUsername == null`, otherwise the list will be updated regardless.

app/src/org/commcare/dalvik/application/CommCareApplication.java
ctsims: Definitely. Returning null when the state is more or less inconceivable is generally an antipattern, and is always harder to track down as a source of errors. The try/catch loop can be tedious for sure. If the context we use this in are SessionAwareCommCareActivity classes, we could extend their root methods (onCreate/onResume/wherever the source of calls are) to have wrappers which catch this exception generically and perform a login redirect, which would prevent the need to do a bunch of explicit handling, since we know what we want to do ~any time the session is unavailable.

app/src/org/commcare/dalvik/dialogs/PaneledChoiceDialog.java
ctsims: exactly. Popping in text that changes the view size can be disorienting for people. It's good for us to smooth it over a bit.

app/src/org/commcare/android/database/app/models/UserKeyRecord.java
```diff
@@ -54,22 +58,42 @@
...
+
+    @Persisting(9)
+    private String rememberedPassword;
+
+    /**
+     * If there are multiple UKRs for a single username in app storage, we guarantee that only
+     * 1 will be marked as active
+     */
+    @Persisting(10)
+    private boolean isActive;
```
phillipm: :clap:

app/src/org/commcare/android/database/app/models/UserKeyRecord.java
```diff
@@ -54,22 +58,42 @@
...
     @MetaField(META_KEY_STATUS)
     @Persisting(7)
     private int type;
 
+    /** The un-hashed password wrapped by a numeric PIN **/
+    @Persisting(8)
+    private byte[] passwordWrappedByPin;
+
+    @Persisting(9)
+    private String rememberedPassword;
```
phillipm: W/o digging too much I don't understand what this is for. Is it set when using pin login and otherwise unset? A comment would help

app/src/org/commcare/android/database/app/models/UserKeyRecord.java
```diff
@@ -235,4 +344,86 @@ public boolean isCurrentlyValid() {
...
+        return !"".equals(rememberedPassword);
+    }
+
+    public String getPrimedPassword() {
+        return this.rememberedPassword;
+    }
+
+    public void clearPrimedPassword() {
+        this.rememberedPassword = "";
+    }
```
phillipm: Comments for (some of) these methods would be :star:

app/src/org/commcare/dalvik/activities/LoginMode.java
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Enum representing the different logical/UI states that you can be in while on the login screen
+ *
+ * @author Aliza Stone (astone@dimagi.com)
+ */
+public enum LoginMode {
+
+    PASSWORD, PIN, PRIMED;
+
+    public static LoginMode fromString(String s) {
```
phillipm: Isn't needed if you store/restore `LoginMode` using extra serialization: http://stackoverflow.com/a/5309967

app/src/org/commcare/dalvik/activities/PinAuthenticationActivity.java
```diff
@@ -0,0 +1,173 @@
...
+    private static final String TAG = PinAuthenticationActivity.class.getSimpleName();
+
+    public static final String PASSWORD_FROM_AUTH = "password-obtained-from-auth";
+
+    private LoginMode authMode;
+    private UserKeyRecord currentRecord;
+    private String passwordObtainedFromAuth;
+
+    @Override
+    public void onCreate(Bundle savedInstanceState) {
```
phillipm: should be `protected` not `public`

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -210,9 +217,121 @@ private void processFromLoginLaunch() {
...
+        f.showDialog();
+    }
+
+    private void launchPinAuthentication() {
+        Intent i = new Intent(this, PinAuthenticationActivity.class);
+        startActivityForResult(i, AUTHENTICATION_FOR_PIN);
+    }
+
+    private void launchPinCreateScreen(LoginMode loginMode) {
+        launchPinCreateScreen(loginMode, null);
```
phillipm: Instead of passing `null` in for `passwordFromAuth`, could you move the `getIntent().getStringExtra(LoginActivity.PASSWORD_FROM_LOGIN)` code up to here? 
Would remove the need for the `if (passwordFromAuth != null) {` line and avoid using `null` which is always a plus

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -270,152 +389,169 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
+                    }
+                    break;
+                case AUTHENTICATION_FOR_PIN:
+                    if (resultCode == RESULT_OK) {
+                        String password = intent.getStringExtra(PinAuthenticationActivity.PASSWORD_FROM_AUTH);
+                        launchPinCreateScreen(LoginMode.PASSWORD, password);
+                    }
+                    return;
+                case CREATE_PIN:
+                    boolean choseRememberPassword = (intent == null) ? false :
```
phillipm: Can be simplified to not need a ternary op

app/src/org/commcare/android/database/app/AppDatabaseUpgrader.java
amstone326: True, but given that `recordsForUsername.add(record)` is a common call, it would just make the code messier to separate that out.

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -270,152 +389,169 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
+                            intent.getBooleanExtra(CreatePinActivity.CHOSE_REMEMBER_PASSWORD, false);
+                    if (choseRememberPassword) {
+                        CommCareApplication._().closeUserSession();
```
phillipm: Why do we need to trigger a login here?

app/src/org/commcare/android/database/app/AppDatabaseUpgrader.java
phillipm: I don't follow. I'm suggesting:
```
if (recordsForUsername == null) {
    recordsForUsername = new ArrayList<>();
    usernamesToRecords.put(username, recordsForUsername);
}
recordsForUsername.add(record);
```

app/src/org/commcare/dalvik/activities/CreatePinActivity.java
```diff
@@ -0,0 +1,220 @@
...
+    private Button cancelButton;
+
+    @UiElement(value=R.id.pin_confirm_button)
+    private Button continueButton;
+
+    private static final String TAG = CreatePinActivity.class.getSimpleName();
+    public static final String CHOSE_REMEMBER_PASSWORD = "chose-remember-password";
+
+    private String unhashedUserPassword;
+    private LoginMode loginMode;
```
phillipm: Can be local to `onCreate`

app/src/org/commcare/dalvik/activities/CreatePinActivity.java
```diff
@@ -0,0 +1,220 @@
...
+    private UserKeyRecord userRecord;
+
+    // Indicates whether the user is entering their PIN for the first time, or is confirming it
+    private boolean inConfirmMode;
+    private String firstRoundPin;
+
+    @Override
+    public void onCreate(Bundle savedInstanceState) {
```
phillipm: `public` -> `protected`

app/src/org/commcare/dalvik/activities/CreatePinActivity.java
```diff
@@ -0,0 +1,220 @@
...
+                Localization.get("remember.password.for.next.login"));
+        return true;
+    }
+
+    @Override
+    public boolean onOptionsItemSelected(MenuItem item) {
+        if (item.getItemId() == MENU_REMEMBER_PW_AND_LOGOUT) {
+            launchRememberPasswordConfirmDialog();
+        }
+        return true;
```
phillipm: consider calling `return super.onOptionsItemSelected(item);` here

phillipm: Saying that because a lot of other code seems to do that; not sure exactly what the implications

app/src/org/commcare/dalvik/activities/CreatePinActivity.java
```diff
@@ -0,0 +1,220 @@
...
+
+    private static final String TAG = CreatePinActivity.class.getSimpleName();
+    public static final String CHOSE_REMEMBER_PASSWORD = "chose-remember-password";
+
+    private String unhashedUserPassword;
+    private LoginMode loginMode;
+    private UserKeyRecord userRecord;
+
+    // Indicates whether the user is entering their PIN for the first time, or is confirming it
+    private boolean inConfirmMode;
```
phillipm: Do you want to keep this state around if you rotate the device?

app/src/org/commcare/dalvik/activities/CreatePinActivity.java
```diff
@@ -0,0 +1,220 @@
...
+    private String firstRoundPin;
+
+    @Override
+    public void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        userRecord = CommCareApplication._().getRecordForCurrentUser();
+        if (userRecord == null) {
+            Log.i(TAG, "Something went wrong in CreatePinActivity. Could not find the current user " +
```
phillipm: To be :pouting_cat: : if you change `Log.i` to Log.e` you can remove "Something went wrong in CreatePinActivity" and convey the same info

app/src/org/commcare/dalvik/activities/CreatePinActivity.java
```diff
@@ -0,0 +1,220 @@
...
+import org.commcare.android.framework.SessionAwareCommCareActivity;
+import org.commcare.android.framework.UiElement;
+import org.commcare.dalvik.R;
+import org.commcare.dalvik.application.CommCareApplication;
+import org.commcare.dalvik.dialogs.AlertDialogFactory;
+import org.javarosa.core.services.locale.Localization;
+
+
+/**
+ * @author Aliza Stone (astone@dimagi.com)
```
phillipm: Would be nice to mention here that you can also reset the pin in this activity

app/src/org/commcare/dalvik/activities/CreatePinActivity.java
```diff
@@ -0,0 +1,220 @@
...
+            setResult(RESULT_CANCELED);
+            this.finish();
+            return;
+        }
+
+        loginMode = LoginMode.fromString(
+                getIntent().getStringExtra(LoginActivity.LOGIN_MODE));
+        if (loginMode == LoginMode.PASSWORD) {
+            unhashedUserPassword = getIntent().getStringExtra(LoginActivity.PASSWORD_FROM_LOGIN);
+        } else if (loginMode == LoginMode.PRIMED) {
```
phillipm: I'm having trouble figuring out when the login mode will be `PASSWORD` and when it will be `PRIMED`.

app/src/org/commcare/dalvik/activities/LoginActivity.java
```diff
@@ -381,40 +391,43 @@ public void keysDoneOther(LoginActivity r, HttpCalloutOutcomes outcome) {
...
+        };
+    }
 
-            return true;
-        }catch (Exception e) {
-            e.printStackTrace();
-            return false;
+    private int getMatchingUsersCount(String username) {
+        int count = 0;
+        for (UserKeyRecord record : storage()) {
```
phillipm: I would take this as an opportunity to remove the `storage()` and `storage` class field, which only used here. I say this only because I'm on a quest to remove that stupid db close error we get sometimes and have resorted to random tactics like removing as many unneeded references to db objects, to try and prevent ref leaks...

app/src/org/commcare/dalvik/activities/PinAuthenticationActivity.java
```diff
@@ -0,0 +1,173 @@
...
+/**
+ * When a user selects Set/Reset PIN from the home activity options menu, they must first
+ * authenticate themselves in order to be able to do so. This is the activity that launches to
+ * prompt the user to enter either their current PIN, if it exists, or their password if it does
+ * not, before proceeding to the CreatePinActivity
+ *
+ * @author Aliza Stone (astone@dimagi.com)
+ */
+@ManagedUi(R.layout.pin_auth_view)
+public class PinAuthenticationActivity extends
```
phillipm: This class gets an A+ in terms of aesthetic! super clean!

app/src/org/commcare/android/database/app/models/UserKeyRecord.java
amstone326: What are you looking to be added? They're pretty simple methods with descriptive names, not sure what I would comment. I added a comment to the rememberedPassword field as you suggested, which might provide the clarification you're looking for on all of these.

phillipm:
mega :muscle: :muscle: :muscle: 
Did a pass but didn't look super deep; assuming Clayton has the detailed logic covered.

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
amstone326: We're triggering a log out, not log in. The user selected the "remember password for next login" option, which saves the password and then logs you out.

app/src/org/commcare/android/database/app/AppDatabaseUpgrader.java
amstone326: ah, got it

app/src/org/commcare/android/database/app/models/UserKeyRecord.java
phillipm: Just reading the code w/o much context on the feature, I'm having trouble figuring out what "primed" means.

app/src/org/commcare/dalvik/activities/CreatePinActivity.java
amstone326: yes, good call

app/src/org/commcare/dalvik/activities/CreatePinActivity.java
amstone326: The login mode is passed down via intents from the LoginActivity -> DispatchActivity -> HomeActivity -> CreatePinActivity. The LoginActivity gets the LoginMode from the current state of the LoginActivityController

app/src/org/commcare/dalvik/activities/LoginActivity.java
amstone326: oh yea super unnecessary

app/src/org/commcare/android/framework/SessionAwareCommCareActivity.java
```diff
@@ -1,11 +1,30 @@
...
 public abstract class SessionAwareCommCareActivity<R> extends CommCareActivity<R> {
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        try {
+            onCreateAware(savedInstanceState);
+        } catch (SessionUnavailableException e) {
+            SessionActivityRegistration.redirectToLogin(this);
+        }
```
phillipm: I think we should try and find an alternative to exposing `onCreateAware` and wrapping its call in a try/catch.
When possible it is good to avoid wrapping large pieces of code in try/catch statements because doing so lets us be lazy about when things fail.

I'll poke around and see if I can come up with any ideas.

app/src/org/commcare/dalvik/activities/PinAuthenticationActivity.java
```diff
@@ -0,0 +1,156 @@
...
+            pinEntry.setText("");
+        } else {
+            Toast.makeText(this, Localization.get("pin.auth.failed.password"), Toast.LENGTH_LONG).show();
+            passwordEntry.setText("");
+        }
+    }
+
+    private void onSuccessfulAuth() {
+        Intent i = new Intent();
+        setResult(RESULT_OK, i);
```
phillipm: I don't think there is a reason to create an `Intent` and set it in the result if you don't put anything in it.

app/src/org/commcare/android/framework/SessionAwareCommCareActivity.java
phillipm: @ctsims @amstone326 here is the commit with my approach to this issue: https://github.com/dimagi/commcare-odk/commit/973f70d9730e28a77a47e0eb374f8e2cb9b94013

Main idea is to fail soft with certain isPinSet checks when the result only effects the UI, which won't even be shown because the user will be redirected to the login screen before it is shown.

I'm not in love with the solution but it localized exception handling instead of integrating it into our session framework, which I think is important.

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -149,9 +157,7 @@
     private static final String EXTRA_CONSUMED_KEY = "login_extra_was_consumed";
 
     @Override
-    protected void onCreate(Bundle savedInstanceState) {
-        super.onCreate(savedInstanceState);
-
+    protected void onCreateAware(Bundle savedInstanceState) throws SessionUnavailableException {
```
phillipm: Overriding implementations of `onCreateAware` should probably still call their super implementation, to allow for more seamless extensibility in the future.

app/src/org/commcare/android/framework/SessionAwareCommCareActivity.java
ctsims: We've gone a bit back and forth on it. While I agree that high level handling has a lot of downsides, I think that the semantic meaning of sessionunavailable basically always has a clear outcome in session aware activities, and I don't want us to fall into a trap of trying to sneak away from the current activity. In your example, for instance, the activity just finishes itself hoping that it'll end up back in the right place, but there's not really any guarantee you will. You could easily end up with a weird workflow where CommCare just exits, or the user ends up back on the home screen for reasons that are unclear to them.

90% of hte time "I wanted session data, but there wasn't a session" means "try to log back in", it seems totally reasonable to me for us to encode that in a way that doesn't need us to write custom logic each time.

app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
ctsims: agree

app/src/org/commcare/android/framework/SessionAwareCommCareActivity.java
phillipm: yeah, just calling `finish` is not a clean way out...
Basically we just need to defer until `onResume`, in which `CommCareSessionAwareActivity.onResume` will detect the loss of the session and redirect to the login screen.
Maybe this is what is bugging me: `CommCareSessionAwareActivity.onCreate`, but calling `onCreateAware`, is doing essentially what `CommCareSessionAwareActivity.onResume` is doing, aborting when the session is lost.

phillipm: Re-did the approach I linked to above to not call `finish()`, instead deferring to `onResume` where the user will be booted to the login screen via existing session aware logic (as far as I can tell)
https://github.com/dimagi/commcare-odk/commit/67467d121fe43acca18752acd35b3f5da436c3d2

phillipm:
Will you please update the '[CommCare Android Developer Options](https://confluence.dimagi.com/display/ccinternal/CommCare+Android+Developer+Options#CommCareAndroidDeveloperOptions-GiveoptiontousePINforlogin)' docs with this feature flag?

amstone326:
@phillipm Updated

PR #986, opened 2016-01-19
Location sort no loader

forest-tong:
Essentially, HereFunctionHandler exposes three major public methods, corresponding to allowing GPS/network use, forbidding GPS/network use, and refreshing the location. This allows the client to only have to worry about whether GPS/network use is permitted (basically corresponding to whether the activity is in focus). 

The logistics of dealing with the location manager and deciding when a good enough location has been achieved are dealt with internally in the HereFunctionHandler class.

In EntitySelectActivity, there is a new boolean flag, containsHereFunction, which is set to true once the first here() function is evaluated. This boolean flag is saved in onDestroy, so that in a screen orientation change, we know that the activity already has a here function. If the activity already has a here function, we can avoid refreshing the location, and immediately re-enable GPS/network use.

There is also a new boolean flag in EntitySelectActivity called locationChangedWhileLoading, which is pretty self-explanatory.

I put in a ton of log statements that should be removed right before pulling. (Edit: we decided to leave them in for future dev work.)

Cross-requested with https://github.com/dimagi/javarosa/pull/231.

Release Note: Allow case lists with location data to be sorted by distance from your current location.


app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -277,6 +287,10 @@ public void onClick(View v) {
         if (!isUsingActionBar()) {
             searchbox.setText(lastQueryString);
         }
+
+        if (!this.containsHereFunction) {
+            hereFunctionHandler.refreshLocation();
+        }
```
phillipm: Might be missing something but I feel like `refreshLocation` should be called when `containsHereFunction` is `true`, not `false`.

app/src/org/commcare/dalvik/geo/HereFunctionHandler.java
```diff
@@ -0,0 +1,174 @@
...
+    public void unregisterEntitySelectActivity() {
+        this.entitySelectActivity = null;
+    }
+
+    // The EntitySelectActivity must subscribe before this method is called if a fresh location is desired.
+    public Object eval(Object[] args, EvaluationContext ec) {
+        if (entitySelectActivity != null && !entitySelectActivity.containsHereFunction) {
+            refreshLocation();
+            allowGpsUse();
+            entitySelectActivity.containsHereFunction = true;
```
phillipm: It is generally bad form in Java land to modify the fields of another class directly. You should do this via getters and setters. While this introduces boilerplate, it makes for more extensible code: maybe you want to synchronize your setters across threads, maybe you want to cache computations and invalidate them when a setter is called, etc.
Most importantly, this is convention: readers of code will not expect for class field mutations to occur outside of that class, and not look for it when they are trying to understand the code.

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -1191,4 +1218,11 @@ private void stopTimer() {
             }
         }
     }
+
+    public void onLocationChanged() {
```
phillipm: Sorta confusing to give this method the same name as `LocationListener.onLocationChanged`. Makes me think that this method implements that interface.

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -1191,4 +1218,11 @@ private void stopTimer() {
             }
         }
     }
+
+    public void onLocationChanged() {
+        boolean loaded = loadEntities();
+        if (!loaded) {
+            locationChangedWhileLoading = true;
```
phillipm: This should be persisted between device orientation changes, right? If the location is updated and you rotate, then loading completes, it currently won't retrigger entity loading.

phillipm: Otherwise, I really like the way `locationChangedWhileLoading` is enabling reloading of the entity list!

app/src/org/commcare/dalvik/geo/HereFunctionHandler.java
```diff
@@ -0,0 +1,174 @@
...
+        if (!locationGoodEnough && !requestingLocationUpdates) {
+            requestLocationUpdates();
+        }
+    }
+
+    public void forbidGpsUse() {
+        if (requestingLocationUpdates) {
+            removeLocationUpdates();
+        }
+    }
```
phillipm: Consider renaming `forbidGpsUse` to `pauseGpsUse` (or `stopGpsUse`) and maybe `allowGpsUse` to `resumeGpsUse`. You could also consider moving the `EntitySelectActivity` calls to `allow/forbidGpsUse` into the `un/registerEntitySelectActivity` methods, since in `EntitySelectActivity` these calls occur in sequence. If you do that I would recommend renaming `registerEntitySelectActivity` to `registerEvalLocationListener` because it makes it a little more clear that you are registering a listener concerned with locations that is tied to an evaluation cycle.

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -901,6 +918,12 @@ public void onClick(View v) {
     }
 
     @Override
+    public void onSaveInstanceState(Bundle savedInstanceState) {
+        savedInstanceState.putBoolean(CONTAINS_HERE_FUNCTION, containsHereFunction);
+        super.onSaveInstanceState(savedInstanceState);
```
phillipm: Stylistically, I prefer to see calls to `super.currentMethod` come first when ordering isn't important, such as this case.

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -161,6 +162,14 @@
     private boolean cancelled;
     private ContainerFragment<EntityListAdapter> containerFragment;
 
+    // Function handler for handling XPath evaluation of the function here().
+    // Although only one instance is created, which is used by NodeEntityFactory,
+    // every instance of EntitySelectActivity registers itself (one at a time)
+    // to listen to the handler and refresh whenever a new location is obtained.
+    public static final HereFunctionHandler hereFunctionHandler = new HereFunctionHandler();
```
phillipm: I would make this private and expose it via a getter method. Allows you to better control the internal implementation later on, in case you want to return new instances each time or do other crazy stuff.

phillipm:
Looking good! Let me know when you finish addressing my comments and I can do another quick pass.

app/src/org/commcare/dalvik/geo/HereFunctionHandler.java
```diff
@@ -0,0 +1,174 @@
...
+    public void refreshLocation() {
+        this.locationGoodEnough = false;
+    }
+
+    @Override
+    public void onLocationChanged(Location location) {
+        this.location = toGeoPointData(location);
+        Log.i("HereFunctionHandler", "location has been set to " + this.location.getDisplayText());
+
+        if (this.location.getAccuracy() <= GeoUtils.ACCEPTABLE_ACCURACY) {
```
phillipm: `getAccuracy` isn't in the current master branch of javarosa. Be sure to create a cross-requested javarosa PR so when this gets merged it doesn't break everyone's master builds.

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
forest-tong: `refreshLocation` starts out false, and only gets set to true when `eval` is called. So in `onCreate`, the only time `refreshLocation` is true is when there's a screen rotation after `eval` has been called. And in that case, we don't want to reset the location, we just want to enable GPS/network use (which occurs in `onResume`).

app/src/org/commcare/dalvik/geo/HereFunctionHandler.java
forest-tong: Done.

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
forest-tong: Done.

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
forest-tong: Done.

app/src/org/commcare/dalvik/geo/HereFunctionHandler.java
forest-tong: I considered wording like `pauseGpsUse` or `resumeGpsUse,` but decided that these don't fit, because what I'm trying to do with this interface is have the client only worry about whether GPS/network use is permitted, and not have to worry about whether the HereFunctionHandler is actually using GPS/network. For example, if the HereFunctionHandler finished finding an accurate enough location, it would be misleading to say `resumeGpsUse` when you mean to say "use GPS if you still need it." If you think allow/forbid is unclear, I could consider permit/prohibit, or enable/disable, or something similar.

The reason `allowGpsUse` can't be moved into `registerEntitySelectActivity` is that we want to register an `EntitySelectActivity` even if it doesn't contain the `here` function (and if there's no `here` function, GPS use is forbidden).

I like the suggestion of renaming it to `registerEvalLocationListener.`

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
forest-tong: Done.

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
forest-tong: Done.

app/src/org/commcare/dalvik/geo/HereFunctionHandler.java
forest-tong: Done.

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
phillipm: Ahh, okay! It is a nonintuitive to be clearing the location in `onCreate` based on whether the entity loading has evaluated a `here()` xpath function yet or not. It seems like you want to clear the location every time you enter a newly launched `EntitySelectActivity`. This can be determined by checking if the `Bundle` argument passed into `onCreate` is `null`, meaning it was a new launch, not a screen rotation.

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -161,6 +163,26 @@
...
+    }
+
+    public boolean getContainsHereFunction() {
+        return containsHereFunction;
+    }
+
+    public void setContainsHereFunction(boolean containsHereFunction) {
+        this.containsHereFunction = containsHereFunction;
+    }
+
```
phillipm: Can you move these get/set methods a bit below the main methods of this Activity? For readability sake it is nice to see the most important methods (`onCreate`, `onResume`, etc) near the top of the class.

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
forest-tong: Good point. I've changed it.

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
forest-tong: Done.

app/src/org/commcare/dalvik/geo/HereFunctionHandler.java
phillipm: Cool, I'm happy with the way it is. What I was getting at is that since you are passing the entity select activity into the register/unregister methods, you can use `getContainsHereFunction()` to determine whether to use/stop gps all from within the here function handler. This would minimize the amount of HereFunctionHandler code you see in EntitySelectActivity

app/src/org/commcare/dalvik/geo/HereFunctionHandler.java
```diff
@@ -0,0 +1,174 @@
...
+    private void removeLocationUpdates() {
+        // stops the GPS. Note that this will turn off the GPS if the screen goes to sleep.
+        if (ContextCompat.checkSelfPermission(this.context, Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED ||
+                ContextCompat.checkSelfPermission(this.context, Manifest.permission.ACCESS_COARSE_LOCATION) == PackageManager.PERMISSION_GRANTED) {
+            mLocationManager.removeUpdates(this);
+            requestingLocationUpdates = false;
+        }
+    }
+
+    public static GeoPointData toGeoPointData(Location location) {
```
phillipm: This can be private

phillipm:
:tada: :satellite: :earth_americas: :earth_africa: :earth_asia:  :satellite: :tada:

amstone326:
@phillipm ^^ definitely trying to make sure you make the cut for the next emoji blog

PR #987, opened 2016-01-20
fix action bar override

wpride:
The case callout icon was getting overridden by the ActionMenu when it existed. This updates the MenuItem in the menu with the correct icon.

It feels like there's a better way to do this that I'm not thinking of but the inner class scopes are making things difficult. Also possible this should be refactored so that barcodeButton is itself a MenuItem? Or do we need the old layout around for pre-Honeycomb devices?


app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -768,6 +777,16 @@ public boolean onQueryTextChange(String newText) {
             }
         });
 
+        if (shortSelect != null) {
+            Action action = shortSelect.getCustomAction();
+            if (action != null) {
+                ViewUtil.addDisplayToMenu(this, menu, MENU_ACTION,
+                        action.getDisplay().evaluate());
+            }
+            makeCalloutClickListener(shortSelect.getCallout());
```
phillipm: This line causes a null pointer crash when I tried testing locally.

wpride:
@phillipm thanks, updated

app/src/org/commcare/dalvik/activities/EntitySelectActivity.java
```diff
@@ -263,6 +264,9 @@ public void onClick(View v) {
         }
 
         barcodeButton.setOnClickListener(barcodeScanOnClickListener);
+        if(callout.getImage() != null){
```
phillipm: This is throwing a NPE for me

PR #991, opened 2016-01-20
Refresh home buttons when language is changed

phillipm:
When you changed languages in the settings menu the home screen buttons didn't update immediately.

Fix for http://manage.dimagi.com/default.asp?191090

### DIFF
--- a/app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
+++ b/app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
@@ -273,7 +273,7 @@ public class CommCareHomeActivity
             case PREFERENCES_ACTIVITY:
                 // rebuild buttons in case language was changed
                 uiController.setupUI();
-                uiController.configUI();
+                rebuildMenus();
                 return;
             case MEDIA_VALIDATOR_ACTIVITY:
                 if(resultCode == RESULT_CANCELED){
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -271,7 +271,9 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
             // if handling new return code (want to return to home screen) but a return at the end of your statement
             switch(requestCode) {
             case PREFERENCES_ACTIVITY:
-                uiController.configUI();
+                // rebuild buttons in case language was changed
+                uiController.setupUI();
```
phillipm: The fix

amstone326: Hey, just for personal understanding, am confused how this fixes it. What call in setupUI() refreshes the language?

phillipm: It rebuilds the button data, which is currently only built once, when the adapter is created. Since the button data is only built once, the localized text will get stale if the language changes. So calling setup again here forces the button text to get re-localized.

amstone326: Oh wow I was looking at the UI controller for LoginActivity, hence my confusion.

PR #992, opened 2016-01-20
Prevent home screen button loading on user data clearing

phillipm:
Small bug where clearing user data causes app to crash due to changes in https://github.com/dimagi/commcare-odk/pull/991

### DIFF
--- a/app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
+++ b/app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
@@ -271,7 +271,7 @@ public class CommCareHomeActivity
             // if handling new return code (want to return to home screen) but a return at the end of your statement
             switch(requestCode) {
             case PREFERENCES_ACTIVITY:
-                if (resultCode == RESULT_CANCELED) {
+                if (resultCode != CommCarePreferences.RESULT_DATA_RESET) {
                     // rebuild home buttons in case language changed;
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/dalvik/activities/CommCareHomeActivity.java
```diff
@@ -271,8 +271,11 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
             // if handling new return code (want to return to home screen) but a return at the end of your statement
             switch(requestCode) {
             case PREFERENCES_ACTIVITY:
-                // rebuild buttons in case language was changed
-                uiController.setupUI();
+                if (resultCode == RESULT_CANCELED) {
+                    // rebuild home buttons in case language changed;
+                    // but only if we didn't just clear user data
+                    uiController.setupUI();
```
phillipm: Testing for when `resultCode` is `RESULT_CANCELED` is sort of a hack, but I'm okay with that...

ctsims: Hm, there are a lot of ways for us to convey this. Can't we just return a different result code for clearing data, like RESULT_DATA_RESET and capture around that?

phillipm: :guardsman:

PR #994, opened 2016-01-21
Fix LoginActivity fragments not restoring properly

phillipm:
__Contains commits from https://github.com/dimagi/commcare-odk/pull/993__

Turns out when you don't call `super.onSaveInstanceState` when overriding `onSaveInstanceState`, then none of your fragments get restored :boom:

The result is that rotating the screen while any login task is running detaches the task from the activity, making the progress dialog disappear

Fix for http://manage.dimagi.com/default.asp?200782


app/src/org/commcare/dalvik/activities/LoginActivity.java
```diff
@@ -132,7 +128,9 @@ public void onRequestPermissionsResult(int requestCode,
     }
 
     @Override
-    public void onSaveInstanceState(Bundle savedInstanceState) {
+    protected void onSaveInstanceState(Bundle savedInstanceState) {
+        super.onSaveInstanceState(savedInstanceState);
```
phillipm: The fix

amstone326: Yay! Confused by how this was causing it though, can you explain?

phillipm: I'm guessing that in the Android platform's implementation of `onSaveInstanceState` certain fragment info is committed to the bundle. By overriding `onSaveInstanceState` but not passing the bundle to the `super` implementation, we interrupted that code path, preventing the platform from storing whatever data it usually stores into that bundle.

phillipm:
Smaller version https://github.com/dimagi/commcare-odk/pull/999

PR #1003, opened 2016-01-25
Prevent NPE in "Settings" view

phillipm:
Getting an NPE on entering the settings view from the home screen.

Not sure if just blindly avoiding the NPE is the right solution, feel free to take over this PR and implement a more appropriate fix.


amstone326:
@phillipm Hey, definitely representative of a deeper problem so don't want to just blindly avoid this. I'll look into this and PR a different fix. Question though-- did you add null checks in both of the places you did because you explicitly saw crashes in both places? Or were you just being preventative with one of them?

phillipm:
On my device, calls to `prefManager.findPreference(PREFS_PRINT_DOC_LOCATION)` were returning `null`. That is why I added the null checks to those 2 parts of the code.

Feel free to either close this PR or it take over.

PR #1012, opened 2016-01-25
Update map api

forest-tong:


app/src/org/commcare/dalvik/activities/EntityMapActivity.java
```diff
@@ -1,288 +1,186 @@
...
+                    .position(entityLocation.second)
+                    .title(entityLocation.first.getFieldString(0))
+                    .snippet(entityLocation.first.getFieldString(1)));
+            markerReferences.put(marker, entityLocation.first.getElement());
+            builder.include(entityLocation.second);
+        }
+        final LatLngBounds bounds = builder.build();
+
+        // Move camera to be include all markers
+        // TODO: does this work for 0 or 1 marker?
```
forest-tong: Not really sure what's the best way to test this.

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
```diff
@@ -1,288 +1,186 @@
...
-                        //Ok, so now we have an address or not. If we _do_ have one, let's have some fun
-                        
-                        if(gp != null) {
-                            OverlayItem overlayItem = overlayFactory.generateOverlay(gp, e);
-                            mEntityOverlay.addOverlay(overlayItem, e.getElement());
-                        }
-                        else { 
-                            bogus++;
+                    String address = entity.getFieldString(i).trim();
+                    if(address != null && !"".equals(address)) {
```
phillipm: you don't need to check for null after calling `trim()`, since that will raise a null pointer exception and the null check will never be reached.
Looking at the implementation of `getFieldString`, it seems to always return a String instance, so you don't need the check in general

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
```diff
@@ -1,288 +1,186 @@
...
-import java.util.List;
-import java.util.NoSuchElementException;
+import java.util.HashMap;
 import java.util.Vector;
 
 /**
  * @author ctsims
  */
-public class EntityMapActivity extends MapActivity {
+@TargetApi(11)
```
phillipm: Can you move this to annotate `onCreate`, which is the only place it is needed?

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
```diff
@@ -1,288 +1,186 @@
...
-    private LocationManager mLocationManager;
-    
+    private EvaluationContext entityEvaluationContext;
     private CommCareSession session;
+    Vector<Entity<TreeReference>> entities;
+
+    Vector<Pair<Entity<TreeReference>, LatLng>> entityLocations;
+    HashMap<Marker, TreeReference> markerReferences = new HashMap<>();
+
+    GoogleMap mMap;
```
phillipm: These class fields can be marked `private`

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
```diff
@@ -1,288 +1,186 @@
...
-    private MyLocationOverlay mMyLocationOverlay;
-    private Geocoder mGeoCoder;
-    private LocationManager mLocationManager;
-    
+    private EvaluationContext entityEvaluationContext;
     private CommCareSession session;
+    Vector<Entity<TreeReference>> entities;
+
+    Vector<Pair<Entity<TreeReference>, LatLng>> entityLocations;
+    HashMap<Marker, TreeReference> markerReferences = new HashMap<>();
```
phillipm: can be marked `final`

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
```diff
@@ -1,288 +1,186 @@
...
+    private static final int MAP_PADDING = 50;  // Number of pixels to pad bounding region of markers
 
-    private MapView map;
-    private MyLocationOverlay mMyLocationOverlay;
-    private Geocoder mGeoCoder;
-    private LocationManager mLocationManager;
-    
+    private EvaluationContext entityEvaluationContext;
     private CommCareSession session;
+    Vector<Entity<TreeReference>> entities;
```
phillipm: this can be made a local variable inside of `onCreate` instead of a class level field

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
```diff
@@ -1,288 +1,186 @@
...
-            return BuildConfig.MAPS_API_KEY;
+
+    private LatLng getLatLngFromAddress(String address) {
+        LatLng location = null;
+        try {
+            GeoPointData data = new GeoPointData().cast(new UncastData(address));
+            if(data != null) {
+                location = new LatLng(data.getLatitude(), data.getLongitude());
+            }
+        } catch(Exception ex) {
```
phillipm: In general it is good to specialize catches as much as possible. I think if you add the `@NonNull` annotation to the `String address` method argument, then the only likely exception that can be raised here is `IllegalArgumentException` from within the `GeoPointData.cast` method. I would just catch that.

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
```diff
@@ -1,288 +1,186 @@
...
         }
-        return this.getString(debugId);
-    }
 
-    private EvaluationContext entityContext;
+        // Geocaching code removed
```
phillipm: Not needed; but we should come up with a deprecation strategy for `GeocodeCacheModel`. For now I would add a TODO to that class' javadoc comment saying it is no longer used and we need to do a db migration to remove the associated database table

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
phillipm: I would also add a `Log.i` statement in the catch

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
```diff
@@ -1,288 +1,186 @@
...
-        super.onStop();
-        mMyLocationOverlay.disableCompass(); 
-        mMyLocationOverlay.disableMyLocation();
+    public void onMapReady(final GoogleMap map) {
+        mMap = map;
+
+        // Find bounding region of markers
+        LatLngBounds.Builder builder = new LatLngBounds.Builder();
+        for (Pair<Entity<TreeReference>, LatLng> entityLocation: entityLocations) {
+            Marker marker = map.addMarker(new MarkerOptions()
```
phillipm: I would use `mMap` instead of `map` inside of this method. Doing so makes it clearer that you are modifying the internal state of the class, not just some local variable

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
phillipm: Maybe you can listen for the bounds changing via http://stackoverflow.com/a/13942063 and then store it such that if you navigate away from the activity and come back it can use those bounds.  I think you can store that info using `onSaveInstanceState`.

Related, what happens when you rotate the device while viewing the map? Does it recenter to the original bounds? If so, this will also fix that

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
```diff
@@ -1,288 +1,186 @@
...
-    protected void onStop() {
-        super.onStop();
-        mMyLocationOverlay.disableCompass(); 
-        mMyLocationOverlay.disableMyLocation();
+    public void onMapReady(final GoogleMap map) {
+        mMap = map;
+
+        // Find bounding region of markers
+        LatLngBounds.Builder builder = new LatLngBounds.Builder();
+        for (Pair<Entity<TreeReference>, LatLng> entityLocation: entityLocations) {
```
phillipm: I would lift this loop into a dedicated helper method

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
```diff
@@ -1,288 +1,186 @@
...
+        // Move camera to be include all markers
+        // TODO: does this work for 0 or 1 marker?
+        map.setOnMapLoadedCallback(new GoogleMap.OnMapLoadedCallback() {
+            @Override
+            public void onMapLoaded() {
+                map.animateCamera(CameraUpdateFactory.newLatLngBounds(bounds, MAP_PADDING));
+            }
+        });
+
+        map.setOnInfoWindowClickListener(this);
```
phillipm: You will want to unset this listener in `onPause` by calling `mMap.setOnInfoWindowClickListener(null)`. If you do that you will want to re-register it in `onResume`

phillipm: (for preventing memory leaks)

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
```diff
@@ -1,288 +1,186 @@
...
-            double lat = location.getLatitude();
-            double lng = location.getLongitude();
-
-            //lLat
-            boundHints[0] = lat -1;
-            //lLng
-            boundHints[1] = lng -1;
+    @Override
+    public void onInfoWindowClick(Marker marker) {
+        Intent i = new Intent(EntityMapActivity.this.getIntent());
```
phillipm: Using the full `EntityMapActivity.this.getIntent()` is only necessary when in an nested anonymous class. In this case `getIntent()` should suffice

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
```diff
@@ -1,288 +1,186 @@
...
+        TreeReference ref = markerReferences.get(marker);
+        SerializationUtil.serializeToIntent(i, EntityDetailActivity.CONTEXT_REFERENCE, ref);
 
-            //uLat
-            boundHints[2] = lat + 1;
+        setResult(RESULT_OK, i);
+        EntityMapActivity.this.finish();
```
phillipm: Same goes for `EntityMapActivity.this.finish();`

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
phillipm: Oh, ^ is in respect to another issue you mentioned.
In terms of testing the TODO above: try breaking out of the for-each loop above after one iteration, or not doing any iterations

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
```diff
@@ -1,288 +1,186 @@
...
+    }
 
-            //rLon
-            boundHints[3] = lng + 1;
+    private EvaluationContext getEvaluationContext() {
```
phillipm: This isn't needed: since `getEvaluationContext` is only called in `onCreate` you can make `entityEvaluationContext` a local variable and set it once like `entityEvaluationContext = session.getEvaluationContext(getInstanceInit());`

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
phillipm: Er, looks like the failed address loading is already logged in `onCreate`, should be good in respects to logging then.

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
```diff
@@ -1,288 +1,186 @@
...
+        if(entityEvaluationContext == null) {
+            entityEvaluationContext = session.getEvaluationContext(getInstanceInit());
         }
+        return entityEvaluationContext;
+    }
+
+    private AndroidInstanceInitializer getInstanceInit() {
```
phillipm: Also not really needed, removing this will enable you to make `session` a local variable instead of a class field

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
```diff
@@ -1,288 +1,186 @@
...
     @Override
-    protected void onCreate(Bundle icicle) {
-        super.onCreate(icicle);
-        String licenseKey = getLicenseKey();
-        map = new MapView(this, licenseKey);
-        
-        this.setContentView(map);
-        
-        mGeoCoder = new Geocoder(this);
+    protected void onCreate(Bundle savedInstanceState) {
```
phillipm: Would recommend trying to break up `onCreate` such that it uses 1 or 2 helper methods for some of the main loops below.

phillipm:
You can now remove the `EntityOverlayItemFactory`, `EntityOverlay`, and `BalloonItemizedOverlay` classes, as well as the whole `libraries/mapballoons` folder!

phillipm:
Looks :bomb:, it'll be sweet to tell our partners that maps work again!
After you checkout my comments we can look into getting the api keys setup on the build server

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
forest-tong: Wow, I didn't know about these new type annotations! Pretty cool.

phillipm: Yeah, been meaning to start using them more myself. They are only at the android level, so don't think they work with vanilla java, such as our javarosa/commcare repos.

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
forest-tong: The problem is that the loop is doing two things (adding the markers to map and calculating the bounding region of the markers), so I don't personally think that lifting it into a helper method would clarify things a lot.

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
forest-tong: I unset the listener in onPause, but is it necessary to re-register it in onResume? The callback should only happen when the map is initially loaded in onCreate.

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
forest-tong: In fact, is it even necessary to clone the intent? Or can we just do `Intent i = getIntent()`?

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
phillipm: I recommend doing it: might be possible to pause the activity before the map is done loading

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
phillipm: Not sure, you should try it out. You might even be able to do `Intent i = new Intent()`. If after making that change you are still able to reach the case detail view by pressing a drop-pin, then things should be good. I suspect the only way it would break is if not using `getIntent()` leaves out some extra that was put in at some earlier point.

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
forest-tong: Took out the method, but ended up leaving session a global variable so helper methods can access it.

forest-tong:
Addressed the comments. I'll remove the classes you mentioned in a separate PR.

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
forest-tong: I tried refactoring it to re-register it in onPause, but it ended up getting messy, because to avoid doing the whole bounds calculation twice you have to make new boolean flags. I'm not sure it's worth it. OnMapReady gets called before loading, so the only way the callback might not be registered is if onMapReady is actually interrupted, and it doesn't do any blocking stuff as far as I can tell.

app/src/org/commcare/android/database/user/models/GeocodeCacheModel.java
```diff
@@ -19,8 +19,10 @@
 import java.util.Date;
 
 /**
+ * TODO: Database migration to remove the associated database tables.
  * @author ctsims
  */
+@Deprecated
```
phillipm: This is sweet :clap: 
Android Studio now shows a strikethrough for usages of this class. Gonna use this one more

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
```diff
@@ -1,288 +1,185 @@
...
-                            OverlayItem overlayItem = overlayFactory.generateOverlay(gp, e);
-                            mEntityOverlay.addOverlay(overlayItem, e.getElement());
-                        }
-                        else { 
-                            bogus++;
+    /**
+     * Gets entity locations, and adds corresponding pairs to the vector entityLocations.
+     * @param detail
+     */
+    private void addEntityLocations(Detail detail) {
```
phillipm: :+1:

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
```diff
@@ -1,288 +1,185 @@
...
+    @Override
+    protected void onPause() {
+        super.onPause();
 
-            //lLat
-            boundHints[0] = lat -1;
-            //lLng
-            boundHints[1] = lng -1;
+        mMap.setOnMapLoadedCallback(null);  // Avoid memory leak in callback
+        if (mMap != null && (ContextCompat.checkSelfPermission(this, Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED
```
phillipm: either remove the null check to `mMap` or wrap ` mMap.setOnMapLoadedCallback(null);` in a null check. Currently `mMap != null` is equiv to a no-op

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
```diff
@@ -1,288 +1,185 @@
...
-                        
-                        if(gp != null) {
-                            OverlayItem overlayItem = overlayFactory.generateOverlay(gp, e);
-                            mEntityOverlay.addOverlay(overlayItem, e.getElement());
-                        }
-                        else { 
-                            bogus++;
+    /**
+     * Gets entity locations, and adds corresponding pairs to the vector entityLocations.
+     * @param detail
```
phillipm: Can you either remove the `@param detail` comment or add a description to it?

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
```diff
@@ -1,288 +1,185 @@
...
+                        .position(entityLocation.second)
+                        .title(entityLocation.first.getFieldString(0))
+                        .snippet(entityLocation.first.getFieldString(1)));
+                markerReferences.put(marker, entityLocation.first.getElement());
+                builder.include(entityLocation.second);
+            }
+            final LatLngBounds bounds = builder.build();
+
+            // Move camera to be include all markers
+            // TODO: does this work for 1 marker?
```
phillipm: Can you check this and remove it before we merge? Should be able to check by adding a `return` below https://github.com/dimagi/commcare-odk/blob/update-map-api/app/src/org/commcare/dalvik/activities/EntityMapActivity.java#L84

forest-tong: Yup, it works with just one marker.

app/res/layout/entity_map_view.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<fragment xmlns:android="http://schemas.android.com/apk/res/android"
+    android:name="com.google.android.gms.maps.MapFragment"
```
phillipm: Is there a reason you didn't use `com.google.android.gms.maps.SupportMapFragment` here instead as suggested here https://developers.google.com/maps/documentation/android-api/start ?

app/src/org/commcare/dalvik/activities/EntityMapActivity.java
```diff
@@ -1,288 +1,186 @@
...
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.entity_map_view);
 
-            @Override
-            protected void selected(TreeReference ref) {
-                Intent i = new Intent(EntityMapActivity.this.getIntent());
-                SerializationUtil.serializeToIntent(i, EntityDetailActivity.CONTEXT_REFERENCE, ref);
-                
-                setResult(RESULT_OK, i);
+        MapFragment mapFragment = (MapFragment) getFragmentManager()
```
phillipm: `getFragmentManager` should probably be `getSupportFragmentManager` to be compat with other devices

forest-tong:
Tried it out just now and it seems to work just like before. That felt too easy...

PR #1013, opened 2016-01-25
Move logic from LoginActivity to ManageKeyRecordTask

phillipm:
:haircut: for LoginActivity: move more key setup logic from login into key record task

Redoing work in https://github.com/dimagi/commcare-odk/pull/993 to avoid merge conflicts


app/src/org/commcare/android/tasks/ExternalManageKeyRecordTask.java
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * @author Phillip Mates (pmates@dimagi.com).
+ */
+public class ExternalManageKeyRecordTask extends ManageKeyRecordTask<ExternalManageKeyRecordTask.DummyDataPullController> {
+    public ExternalManageKeyRecordTask(Context c, int taskId, String username, String password,
+                                       LoginMode loginMode, CommCareApp app, boolean restoreSession) {
+        super(c, taskId, username, password, loginMode, app, restoreSession, false);
+    }
+
```
amstone326: would be worth adding an explanatory comment for why this class overrides all of its inherited methods with empty implementations. Or just rename it DummyManageKeyRecordTask (assuming this is a Dummy class?)

phillipm: :ok_hand:

PR #1027, opened 2016-01-28
Force logout to apply app update

phillipm:
Should have done this a while ago, but better late than never:
Disable the ability to apply an update without logging out; because I guess that is sketch.

Also modified the single app manager code to still work. Now applies any updates in the single app manager activity upon return from the update activity. @amstone326 probably want to check that out.

Removing the ability to apply the update in the update screen fixes this QA ticket http://manage.dimagi.com/default.asp?215637

### DIFF
--- a/app/assets/locales/messages_ccodk_default.txt
+++ b/app/assets/locales/messages_ccodk_default.txt
@@ -12,13 +12,11 @@ updates.check.begin=Checking for updates...
 updates.check.failed=Failed to download updates
 updates.check.start=Recheck
 updates.check.cancel=Stop checking
-updates.check.install=Install
 updates.check.cancelling=Cancelling check for updates
 updates.install.finished=App update was successful
 updates.check.network_unavailable=No network connectivity
-updates.staged.version=Update to version ${0}
+updates.staged.version=Update to version ${0} & log out
 updates.error=Error encountered during update download
-updates.applied.on.login=(downloaded update will automatically be applied on next login.)
 
 updates.pinned.download=Downloading updates
 updates.pinned.progress=${0}/${1}
--- a/app/assets/locales/messages_ccodk_default.txt
+++ b/app/assets/locales/messages_ccodk_default.txt
@@ -16,6 +16,7 @@ updates.check.cancelling=Cancelling check for updates
 updates.install.finished=App update was successful
 updates.check.network_unavailable=No network connectivity
 updates.staged.version=Update to version ${0} & log out
+updates.staged.version.app.manager=Update to version ${0}
 updates.error=Error encountered during update download
 
 updates.pinned.download=Downloading updates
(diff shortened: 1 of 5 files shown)


app/src/org/commcare/dalvik/activities/SingleAppManagerActivity.java
```diff
@@ -29,13 +32,17 @@
  * @author amstone
  */
 
-public class SingleAppManagerActivity extends Activity {
+public class SingleAppManagerActivity extends CommCareActivity<SingleAppManagerActivity> {
```
phillipm: Any issues with extending `CommCareActivity`? Makes dialog/task management way easier!

app/assets/locales/messages_ccodk_default.txt
```diff
@@ -12,13 +12,11 @@ updates.check.begin=Checking for updates...
 updates.check.failed=Failed to download updates
 updates.check.start=Recheck
 updates.check.cancel=Stop checking
-updates.check.install=Install
 updates.check.cancelling=Cancelling check for updates
 updates.install.finished=App update was successful
 updates.check.network_unavailable=No network connectivity
-updates.staged.version=Update to version ${0}
+updates.staged.version=Logout and update to ${0}
```
amstone326: Think you want this to say "Logout and update to *version* ${0}" ?

phillipm: I left out 'version' because I didn't want it to be too long. I can add it back if that seems confusing

app/src/org/commcare/dalvik/activities/SingleAppManagerActivity.java
amstone326: Nope, I just didn't originally because there was no reason to.

app/assets/locales/messages_ccodk_default.txt
amstone326: My opinion would be that it's worth leaving in

amstone326:
@phillipm Ok tested this out, a few comments:

-Worth looking into the thing I showed you with the apply updates button getting elongated by the longer text. I know it's a small thing but it does look a lot nicer without the extra space at the bottom..

-Super minor but would change the text to say "Log out and update..." instead of "Logout and update..."

-Even though the button text does in fact warn you that you're going to be logged out, there's something about it that feels sort of jarring. I could see a user thinking something went wrong if they pressed the button without reading it carefully, and then suddenly found themselves logged out when what they were expecting was an app update. Not sure what the best way to address that is. At the very least, I would add back some text underneath the buttons in parentheses that says like "(Your update will be applied on your next login)". But would probably say that we want something more. Is there a reason that we can't have the update apply when you press that button, and *then* log you out when it's done? I believe that is what used to happen before you refactored how updates work (@ctsims can confirm).

amstone326:
One other thing: When you're doing an update from the app manager, the text on the apply updates button can just say "Update to Version x" instead of "Log out and Update" because at that point you're not within an app, and the user actually explicitly gets a warning that they will be logged out of any app they are currently logged into before they can enter the UpdateActivity from the app manager.

phillipm:
@amstone326 :dart: with the :eyeglasses: :+1: 
Updated to apply update and then logout, "logout" -> "log out", potentially fixed button elongation (not really sure...), and show different text when launched by app manager. :tada:

amstone326:
Awesome, looks good!

PR #1028, opened 2016-01-29
Make form save messages distinguish between incomplete/complete saves

phillipm:
Meryn and the PPIUD team brought up an interesting point that users were getting confused by the fact that when you save a form as incomplete via the drop down menu, you get the same message as when you save a form as complete.

This PR makes it so when you save a form using the drop down menu, you get:
> Form saved as incomplete

And when you complete a form at the end-of-form nav, it shows:
> Form saved and ready for submission

Perhaps the "ready for submission" part is confusing because it makes it seem like the user has something they need to do?

screen1

screen

### DIFF
--- a/app/src/org/odk/collect/android/tasks/SaveToDiskTask.java
+++ b/app/src/org/odk/collect/android/tasks/SaveToDiskTask.java
@@ -44,7 +44,7 @@ import javax.crypto.spec.SecretKeySpec;
 public class SaveToDiskTask<R extends FragmentActivity> extends CommCareTask<Void, String, SaveToDiskTask.SaveStatus, R> {
     // callback to run upon saving
     private FormSavedListener mSavedListener;
-    private final Boolean mSave;
+    private final Boolean exitAfterSave;
     private final Boolean mMarkCompleted;
     // URI to the thing we are saving
     private Uri mUri;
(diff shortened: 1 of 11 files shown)


phillipm:
@amsagoff wondering, do you have suggestions on the string text I've chosen?

amsagoff:
@phillipm agree "ready for submission" implies some action from the user. What if we kept that as "form successfully saved" and instead changed the submission text to "form successfully completed"?

Though ultimately I think meryn's issue was with the end of form UI being confusing.

amsagoff:
@phillipm to make sure i understand - when you're on the last question and you save the form, does it get saved as incomplete or saved and submitted or...what

phillipm:
@amsagoff if the bar is green but you save the form using the drop-down menu, it still gets saved as incomplete. This feels wrong to me, but that is how it works currently.

I like your text suggestions, will update

amsagoff:
That actually feels right to me - in that case, I have a new suggestion: why doesn't the text say "form saved as incomplete" if that's what's really happening?

app/src/org/odk/collect/android/tasks/SaveToDiskTask.java
```diff
@@ -373,13 +379,17 @@ private int validateAnswers(boolean markCompleted, boolean fireTriggerables) {
                     saveStatus =
                             FormEntryActivity.mFormController.checkCurrentQuestionConstraint();
                 }
-                if (markCompleted && saveStatus != FormEntryController.ANSWER_OK) {
-                    return saveStatus;
+                if (markCompleted &&
+                        saveStatus != FormEntryController.ANSWER_OK && // not technically needed
```
amstone326: looks like you already noted this, but why leave this line in?

phillipm: Not sure, just took it out.

PR #1035, opened 2016-02-01
Remove new doc flags for restart intent

phillipm:
Strangely, on my particular device (Nexus 5x / Android 6) when I uninstall an app in the app manger the call to `CommCareApplication.restartCommCare` trashes my app, freezing everything until I force close it. 

Removing these flags fix it.
Not really sure what they do / why they were put here in the first place. @ctsims, do you know?

### DIFF
--- a/app/src/org/commcare/dalvik/application/CommCareApplication.java
+++ b/app/src/org/commcare/dalvik/application/CommCareApplication.java
@@ -278,11 +278,6 @@ public class CommCareApplication extends Application {
         intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP |
                 Intent.FLAG_ACTIVITY_SINGLE_TOP |
                 Intent.FLAG_ACTIVITY_RESET_TASK_IF_NEEDED);
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/dalvik/application/CommCareApplication.java
```diff
@@ -278,11 +278,6 @@ public static void restartCommCare(Activity activity) {
         intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP |
```
ctsims: http://developer.android.com/reference/android/app/Activity.html#finishAndRemoveTask()
http://developer.android.com/reference/android/app/Activity.html#finishAffinity()

phillipm: Those require API 16 and API 21. Do you suggest  using them if the API level is met?
Also seems strange to launch an activity from the activity we are closing. Wondering if using a pending intent w/ a timer would be a good approach: http://stackoverflow.com/a/17166729

amstone326:
@phillipm A pending intent approach is what was being used in the app manager's implementation of restart CommCare before I did the unification. Look at what's deleted in SingleAppManagerActivity here: https://github.com/dimagi/commcare-odk/pull/962/files

phillipm:
Just tested out the pending intent stuff and what is currently used seems to work slightly better. 
This is ready to :eyeglasses: again

PR #1039, opened 2016-02-01
fix bitly install

wpride:


wpride:
see https://github.com/dimagi/commcare-odk/pull/1038

PR #1040, opened 2016-02-01
Card fix 1

wpride:
http://manage.dimagi.com/default.asp?215785

### DIFF
--- a/app/res/layout/square_card.xml
+++ b/app/res/layout/square_card.xml
@@ -38,6 +38,7 @@
                 android:paddingLeft="@dimen/standard_spacer"
                 android:paddingRight="@dimen/standard_spacer"
                 android:textSize="@dimen/font_size_large"
+                android:maxLines="1"
                 app:text_size_small="@dimen/font_size_medium"/>
         </RelativeLayout>
 
--- a/app/src/org/commcare/android/adapters/HomeCardDisplayData.java
+++ b/app/src/org/commcare/android/adapters/HomeCardDisplayData.java
@@ -92,5 +92,10 @@ public class HomeCardDisplayData {
             squareButtonViewHolder.subTextView.setVisibility(View.GONE);
         }
     }
+
+    @Override
+    public String toString(){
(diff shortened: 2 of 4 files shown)


wpride:
I can't figure out why the reordering is happening, pretty deep in the SGLM code, though seems possibly related to this known bug: https://github.com/lucasr/twoway-view/issues/190

However the behavior we're seeing is different, screen rotation non-deterministically will trigger a reordering, then scrolling will always revert the menu to its original/correct ordering

This hacky workaround basically just invalidates the current spannable cache and makes a call to the grid manager's checkForGaps method:

https://android.googlesource.com/platform/frameworks/support/+/a910619/v7/recyclerview/src/android/support/v7/widget/StaggeredGridLayoutManager.java#216

phillipm:
The reordering issue on scroll might be related to this line: https://github.com/dimagi/commcare-odk/blob/master/app/src/org/commcare/dalvik/activities/HomeActivityUIController.java#L87

When I removed that line I didn't experience the reordering on scroll, only reordering when you pause and resume the activity

phillipm:
Actually, might be wrong on that one... It might also be fixed by adding `uiController.refreshView` to `CommCareHomeActivity.onResume`

wpride:
@phillipm tried that among many other similar "fixes"

phillipm:
Think I found something: I modified `ResizingTextView`, adding a call to `ResizingTextView.reset()` at this line: https://github.com/dimagi/commcare-odk/blob/master/app/src/org/commcare/android/view/ResizingTextView.java#L72. Seems to have resolved the issue for me.

wpride:
refreshview

@phillipm here's what I'm seeing just implementing that change - is this the desired/acceptable output? Buttons still misaligned, logout button on right side cc @ctsims @amstone326 @biyeun

wpride:
@phillipm just saw you opened https://github.com/dimagi/commcare-odk/pull/1046, will move this discussion over there and close this branch for now

PR #1044, opened 2016-02-01
Refactor: Offline install

phillipm:
- Don't show keyboard upon entering offline install screen
- Show blue 'install' button
- Trigger install after successful unzip. Currently the user has the option to proceed/cancel install after unzip. At that point I would expect all users to always press proceed, since they have already pressed a button titled `Install App`

Still work to be done in terms of fully migrating this to the new UI; mostly just wanted to prevent the keyboard from popping up cause that is mad annoying.

screen


app/src/org/commcare/dalvik/activities/InstallArchiveActivity.java
```diff
@@ -116,14 +114,12 @@ protected void deliverError(InstallArchiveActivity receiver, Exception e) {
             }
         };
 
-        File mFile = new File(filepath);
```
phillipm: unused

app/src/org/commcare/dalvik/activities/InstallArchiveActivity.java
```diff
@@ -161,13 +157,6 @@ protected void onResume() {
     }
 
     private void evalState() {
-        if (done) {
```
phillipm: always false

orangejenny:
Good to merge once conflicts are resolved.

PR #1046, opened 2016-02-02
Prevent reordering of home screen buttons due to text resizing

phillipm:
Note: doesn't prevent home screen buttons from staggering in a different order when in demo mode, but does prevent button reordering when the home screen is paused and resumed.

Follow up to https://github.com/dimagi/commcare-odk/pull/1040

Issue described in http://manage.dimagi.com/default.asp?215785

### DIFF
--- a/app/src/org/commcare/android/view/ResizingTextView.java
+++ b/app/src/org/commcare/android/view/ResizingTextView.java
@@ -70,6 +70,7 @@ public class ResizingTextView extends TextView {
         if (!mHasTriedSmallLayout && this.getLineCount() > 1) {
             setTextSizeToSmall();
         }
+        reset();
     }
 
     private void setTextSizeToSmall() {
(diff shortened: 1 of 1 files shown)


wpride:
from #1040 :

refreshview

@phillipm here's what I'm seeing just implementing that change - is this the desired/acceptable output? Buttons still misaligned, logout button on right side cc @ctsims @amstone326 @biyeun

biyeun:
yeah, those buttons should have consistent groups of height...that looks jarring

biyeun:
e.g. the vibrant colors should all be the same hieght

app/src/org/commcare/android/view/ResizingTextView.java
```diff
@@ -24,27 +23,29 @@
...
+
         setParams(context, attrs);
     }
 
     @Override
     protected void onTextChanged(CharSequence text, int start, int lengthBefore, int lengthAfter) {
-        this.reset();
+        if (!getText().equals(text)) {
+            resetOriginalTextSize();
+        }
```
phillipm: The fix

phillipm:
screen

PR #1048, opened 2016-02-03
setAdjustViewBounds true

wpride:
Fix for http://manage.dimagi.com/default.asp?216180
We need to allow this ImageView to resize itself according to the new scaling rules

### DIFF
--- a/app/src/org/odk/collect/android/views/media/MediaLayout.java
+++ b/app/src/org/odk/collect/android/views/media/MediaLayout.java
@@ -214,6 +214,7 @@ public class MediaLayout extends RelativeLayout {
                         mImageView.setPadding(10, 10, 10, 10);
                         mImageView.setImageBitmap(b);
                         mImageView.setId(23423534);
+                        mImageView.setAdjustViewBounds(true);
                         mediaPane = mImageView;
                     } else if (errorMsg == null) {
                         // An error hasn't been logged and loading the image failed, so it's likely


wpride:
fix

app/src/org/odk/collect/android/views/media/MediaLayout.java
```diff
@@ -214,6 +214,7 @@ else if (qrCodeContent != null ) {
                         mImageView.setPadding(10, 10, 10, 10);
                         mImageView.setImageBitmap(b);
                         mImageView.setId(23423534);
+                        mImageView.setAdjustViewBounds(true);
```
ctsims: @amstone326 Can you confirm with whether this jives with your image resizing work? I'm honestly not sure how it overlaps.

amstone326:
@ctsims It should be fine because all that's doing is telling the image to preserve its aspect ratio. But am a bit confused about what changed in this release to cause us to need that? And also why that fixed the issue VL was reporting. @wpride Can you weight in at all? What made you try that as the solution?

wpride:
@amstone326 I recalled having to add the same thing when I implemented the ResizingImageView:

https://github.com/dimagi/commcare-odk/blob/a7818db1f08d55395b6616610db28b6f866a5741/app/src/org/odk/collect/android/views/media/MediaLayout.java#L210

The only reason I can think that this just came up is that going down this code path requires that both cc-inflation-target-density and cc-resize-images be set to none. If they tested the target densities and other resizing methods this wouldn't come up

PR #1050, opened 2016-02-03
Remove duplicate implementation of getPrototypeFactory

phillipm:


app/src/org/commcare/android/database/DbUtil.java
```diff
@@ -95,9 +94,6 @@ public static PrototypeFactory getPrototypeFactory(Context c) {
                         if (!emptyc) {
                             continue;
                         }
-                        if (Externalizable.class.isAssignableFrom(prototype)) {
-                            classNames.add(cn);
-                        }
```
phillipm: Removed this because the class name is added regardless right below. Sound alright?

app/src/org/odk/collect/android/tasks/FormLoaderTask.java
```diff
@@ -265,7 +265,7 @@ private static FormDef deserializeFormDef(Context context, File formDef) {
             DataInputStream dis = new DataInputStream(new BufferedInputStream(fis));
 
             // read serialized formdef into new formdef
-            fd.readExternal(dis, ApkUtils.getPrototypeFactory(context));
+            fd.readExternal(dis, DbUtil.getPrototypeFactory(context));
```
phillipm: Note: this change means that we are now using an `AndroidPrototypeFactory` when loading forms, not a vanilla `PrototypeFactory`. I assume that is what we want, but am not positive.

orangejenny: @ctsims could confirm, but I'm confident enough that this is true to merge.

orangejenny:
@phillipm I just noticed that this is assigned to Clayton, but am assuming that's just to bring it to his attention, not because he's the only person you want to merge this. Let me know if that's not the case.

PR #1051, opened 2016-02-04
URL -> Code

wpride:
Update other locations where "Enter URL" is used instead of "Enter Code"

cc @kaapstorm


kaapstorm:
:+1: :+1: :+1:

PR #1057, opened 2016-02-04
Build: enable different build variants/package names with google analytics

phillipm:
Google analytics broke the QA beta build.

Fix that, in a super silly way, cause Gradle is :space_invader: :space_invader: :space_invader:


build.gradle
```diff
@@ -34,6 +35,11 @@ repositories {
     }
 }
 
+configurations {
+    // prevents libs from including commons-logging, which Android already does
+    all*.exclude group: 'commons-logging', module: 'commons-logging'
```
phillipm: addresses the hundreds of instances of the following warning that pop up at build time

> WARNING: Dependency commons-logging:commons-logging:1.1.1 is ignored for qabetaDebug as it may be conflicting with the internal version provided by Android.

phillipm:
@amstone326 Note that all the build variants use the same analytics credentials. This is probably not optimal if we start actually using standalone apps with prepackaged CCZs. I'll let you figure out how to switch up the appropriate values in `google-services.json`

amstone326:
@phillipm My plan for this is likely to just set the tracking id conditionally based upon the build type. Right now, I'm just using the DEV tracking ID for everything, because we're letting the 2.26 release be a test round: https://github.com/dimagi/commcare-odk/blob/master/app/src/org/commcare/dalvik/application/CommCareApplication.java#L346. For 2.27, I'm going to change that line to set the dev tracking ID for debug builds, and the live tracking ID for others. For standalone builds, we could either choose to report to the test dashboard, or just not report analytics for them at all (which might actually be the best choice).

phillipm:
@amstone326 perfect. Not sure if there is pre-built infrastructure available to set the tracking IDs. If not, the substitution functionality in this PR should do the trick.

amstone326:
@phillipm What do you mean by "pre-built infrastructure"? Something other than the way I'm setting the tracking ID in the code I posted?

phillipm:
The ID is hardcoded in [`google-services.json`](https://github.com/dimagi/commcare-odk/blob/master/google-services.json#L23) so you will probably need to set it there as well, right?

amstone326:
Ah, good point. I'm not sure which location takes precedence if it is set in both. My guess would be that it only uses the value from google-services.json as a default if you don't explicitly set one in code, but I'll have to look into it.

PR #1059, opened 2016-02-05
Synchronous server request activity

phillipm:
Feature to search for cases on a server and claim them, pulling down the claimed case locally.

cross-request: https://github.com/dimagi/commcare/pull/253
cross-request: https://github.com/dimagi/javarosa/pull/289

screen

screen

screen

screen


app/src/org/commcare/activities/EntitySelectActivity.java
```diff
@@ -173,6 +174,9 @@ protected void onCreate(Bundle savedInstanceState) {
         refreshTimer = new EntitySelectRefreshTimer();
         asw = CommCareApplication._().getCurrentSessionWrapper();
         session = asw.getSession();
+        // Don't show actions (e.g. 'register patient', 'claim patient') when
+        // in the middle on workflow triggered by an (sync) action.
+        hideActions = session.isSyncCommand(session.getCommand());
```
amstone326: I'm not sure if this is a good idea, but going to throw it out and we can discuss (or you can reject outright if you hate it)... I agree that it makes sense to reuse the normal case list and case detail display frameworks from the suite file for displaying the case claim options. However, it does feel a little weird to me to be using the exact same code path as the usual entity select screen, with just this boolean flag being the only distinction. You could create a new activity with a more fitting name (`CaseClaimSelectActivity` ?) that extends `EntitySelectActivity`, and then you have a clean way to play around with the UI of that activity a bit and distinguish it from the normal entity select screen. It's a bit of a picky distinction to ask for, but I think it just feels a little off to me to use the same activity to do 2 things that are conceptually very different in the app.

amstone326:
@phillipm Make sure to have HQ introduce logic that prevents a user from being able to claim cases that they already own

amstone326:
It probably also wouldn't hurt to change the mobile side such that we show a useful message (just like "this case is already on your phone") if a user somehow does get in a state that they try to do that, instead of just erroring.

app/src/org/commcare/activities/EntitySelectActivity.java
amstone326: (Just making a note of the fact that we decided to wait to address this until after the first round of UI/UX feedback)

app/src/org/commcare/activities/PostRequestActivity.java
```diff
@@ -0,0 +1,246 @@
...
+    private void makePostRequest() {
+        if (!hasTaskLaunched && !inErrorState) {
+            SimpleHttpTask postTask;
+            try {
+                postTask = new SimpleHttpTask(this, url, params, true);
+            } catch (ModernHttpRequester.PlainTextPasswordException e) {
+                enterErrorState(Localization.get("post.not.using.https", url.toString()));
+                return;
+            }
+            postTask.connect((CommCareTaskConnector)this);
```
amstone326: I'm confused as to why this cast is necessary... I see that it is b/c I just tried removing it in AS and got an error, but shouldn't this just work since `PostRequestActivity` implements `CommCareTaskConnector`.

amstone326: Is it because the implementation is indirect (`PostRequestActivity` implements `ConnectorWithHttpResponseProcessor` which implements `CommCareTaskConnector` ?)

app/assets/locales/messages_ccodk_default.txt
```diff
@@ -163,6 +163,23 @@ login.welcome.multiple=Welcome back! Please select an app and log in.
...
+query.dialog.title=Querying Server
+query.dialog.body=Searching server for relevant data
+query.response.format.error=Error processing server response: ${0}
+query.response.empty=Query response had no results
+post.request.button=Claim
+post.malformed.url=${0} is not a valid URL
+post.generic.error=An error occurred while preparing the HTTP request
+post.not.using.https=${0} is an invalid URL. You must use HTTPS.
+post.dialog.title=Claiming...
+post.dialog.body=Claiming choosen data from server
```
amstone326: typo, choosen --> chosen

app/src/org/commcare/activities/QueryRequestActivity.java
```diff
@@ -0,0 +1,335 @@
...
+            setResult(RESULT_CANCELED);
+            finish();
+        } else {
+            loadStateFromSavedInstance(savedInstanceState);
+
+            setupUI();
+        }
+    }
+
+    private static RemoteQuerySessionManager buildQuerySessionManager(AndroidSessionWrapper sessionWrapper) {
```
amstone326: would argue that this would be better as a static method of `RemoteQuerySessionManager.java`

app/src/org/commcare/activities/QueryRequestActivity.java
```diff
@@ -0,0 +1,335 @@
...
+                                  DisplayUnit displayUnit, boolean isLastPrompt) {
+        Hashtable<String, String> userAnswers =
+                remoteQuerySessionManager.getUserAnswers();
+        promptsLayout.addView(createPromptMedia(displayUnit));
+
+        EditText promptEditText = new EditText(this);
+        if (userAnswers.containsKey(promptId)) {
+            promptEditText.setText(userAnswers.get(promptId));
+        }
+        promptEditText.setSingleLine();
```
amstone326: can you add a comment as to why this is necessary? seems a little odd otherwise

app/src/org/commcare/activities/QueryRequestActivity.java
```diff
@@ -0,0 +1,335 @@
...
+
+    private MediaLayout createPromptMedia(DisplayUnit display) {
+        DisplayData displayData = display.evaluate();
+        String promptText =
+                Localizer.processArguments(displayData.getName(), new String[]{""}).trim();
+        TextView text = new TextView(getApplicationContext());
+        text.setText(promptText);
+
+        int padding = (int)getResources().getDimension(R.dimen.help_text_padding);
+        text.setPadding(0, 0, 0, 7);
```
amstone326: can you make this not hardcoded?

app/src/org/commcare/activities/QueryRequestActivity.java
```diff
@@ -0,0 +1,335 @@
...
+        } else if (isResponseEmpty(instanceOrError.first)) {
+            Toast.makeText(this, Localization.get("query.response.empty"), Toast.LENGTH_SHORT).show();
+        } else {
+            CommCareApplication._().getCurrentSession().setQueryDatum(instanceOrError.first);
+            setResult(RESULT_OK);
+            finish();
+        }
+    }
+
+    public static Pair<ExternalDataInstance, String> buildExternalDataInstance(InputStream instanceStream,
```
amstone326: a javadoc of some sort here would be nice

app/src/org/commcare/adapters/EntityListAdapter.java
```diff
@@ -391,7 +391,7 @@ public synchronized void signalKilled() {
      * Get action's index in detail's list of actions given position in adapter
      */
     public int getActionIndex(int positionInAdapter) {
-        return positionInAdapter - (getCurrentCountWithActions() - 1);
+        return positionInAdapter - getCurrentCount();
```
amstone326: confused by this change

app/src/org/commcare/tasks/SimpleHttpTask.java
```diff
@@ -0,0 +1,108 @@
...
+import java.io.InputStream;
+import java.net.URL;
+import java.util.HashMap;
+
+/**
+ * Makes get/post request and delegates http response to receiver on completion
+ *
+ * @author Phillip Mates (pmates@dimagi.com)
+ */
+public class SimpleHttpTask
```
amstone326: Makes sense not to do in this PR, but is this something that it would make sense to start using in other places where we are doing server communication?

amstone326:
@wpride If you have a chance, it would be good for you to do a quick pass of the parts of this PR that introduce new classes for handling http requests/responses, since I don't have much knowledge of best practices or how we currently do it

app/src/org/commcare/network/ModernHttpRequester.java
```diff
@@ -0,0 +1,224 @@
...
+        } else {
+            return buildUrlWithParams();
+        }
+    }
+
+    protected HttpURLConnection setupConnection(URL builtUrl) throws IOException {
+        Log.d(TAG, builtUrl.toString());
+        HttpURLConnection httpConnection = (HttpURLConnection)builtUrl.openConnection();
+        if (isPostRequest) {
+            setupConnectionInner(httpConnection);
```
amstone326: this line is repeated in the `else` statement, so can be moved to above the `if`

unit-tests/resources/commcare-apps/case_search_and_claim/bad-query-result.xml
```diff
@@ -0,0 +1,14 @@
+asodfhoadshgaofiw<patients>
```
amstone326: Is this string meant to be here?

unit-tests/resources/commcare-apps/case_search_and_claim/default/app_strings.txt
```diff
@@ -0,0 +1,45 @@
+app.display.name=village-health2
```
amstone326: I think you can remove this file; I think it's good for us to keep the test app files as bare-bones as possible

amstone326: ah nevermind, I forgot it gets referenced in the suite file. looks like all the other test apps have this file as well.

unit-tests/src/org/commcare/android/tests/activities/PostRequestActivityTest.java
```diff
@@ -0,0 +1,205 @@
...
+        String badURL = "http://bad.url.com";
+
+        PostRequestActivity postRequestActivity = buildPostActivity(badURL);
+
+        assertErrorMessage(postRequestActivity, true, Localization.get("post.not.using.https", badURL));
+    }
+
+    private static void assertErrorMessage(PostRequestActivity postRequestActivity,
+                                           boolean isVisible,
+                                           String hasMessage) {
```
amstone326: `hasMessage` seems like an odd variable name

unit-tests/src/org/commcare/android/tests/activities/PostRequestActivityTest.java
```diff
@@ -0,0 +1,205 @@
...
+        TextView errorMessage =
+                (TextView)postRequestActivity.findViewById(R.id.error_message);
+        if (isVisible) {
+            Assert.assertEquals(View.VISIBLE, errorMessage.getVisibility());
+        } else {
+            assertFalse(View.VISIBLE == errorMessage.getVisibility());
+        }
+        if (hasMessage != null) {
+            Assert.assertEquals(hasMessage, errorMessage.getText());
```
amstone326: it looks like it doesn't make a difference which you use, but it would be good to be consistent about using the `Assert.` prefix for calls to `assertEquals`, `assertFalse`, etc.

app/src/org/commcare/activities/PostRequestActivity.java
phillipm: Yeah, I have no idea why this is needed

app/src/org/commcare/activities/QueryRequestActivity.java
phillipm: How would I do that?
Sometime in the future we should migrate `MediaLayout` to an xml layout file, at which point we could remove most of this layout setup code.

app/src/org/commcare/adapters/EntityListAdapter.java
phillipm: I think `getCurrentCountWithActions() - 1` only worked when there was one action.

app/src/org/commcare/tasks/SimpleHttpTask.java
phillipm: Yes, though doing so is part of a initiative to migrate from our deprecated http libs (`HttpRequestGenerator`) to new http libs (`ModernHttpRequester`), which we should thoroughly test.

unit-tests/resources/commcare-apps/case_search_and_claim/bad-query-result.xml
phillipm: yeah, that is what makes this a bad query result

app/src/org/commcare/adapters/EntityListAdapter.java
amstone326: ah ok get it now

app/src/org/commcare/activities/QueryRequestActivity.java
amstone326: Agreed on the migrate. In the short term, you should be able to add the values you want to use to `dimens.xml` and then reference those here using `getResources().getDimension(R.dimen.xxx)`

phillipm:
@amstone326 thanks for review, lots of good points! I think I've addressed everything across the 3 PRs except for https://github.com/dimagi/commcare-odk/pull/1059#issuecomment-221902789 which I'd like to address in a future PR if that is okay.

PR #1069, opened 2016-02-11
Move logs to user storage + Improvements to force close logs

amstone326:
Changes to what we do with force closure reports:
-If submission of a force close fails on initial attempt, save it to storage (if possible, e.g. if user is logged in) so we can try again later during a normal `LogSubmissionTask`
-Make a dedicated `ForceCloseLogEntry` class that stores a lot more info than other log entries have (domain, app version, readable session string, restorable serialized session string, etc.), and start submitting force close logs in this format (but for now, also continues sending them in the old format, so that HQ can still process them until the new format is being handled)

And a change to how we handle all logs: When there is an active session, save them to user storage, rather than global storage


app/src/org/commcare/android/logging/ForceCloseLogSerializer.java
```diff
@@ -0,0 +1,100 @@
...
+                });
+            }
+        });
+    }
+
+    @Override
+    public void writeToDeviceReport(XmlSerializer serializer) throws IOException {
+        this.serializer = serializer;
+
+        serializer.startTag(DeviceReportWriter.XMLNS, "force_close_subreport");
```
ctsims: This means we're going to start missing these in the short term, right? Until we get HQ to start parsing these?

app/src/org/commcare/android/session/DevSessionRestorer.java
```diff
@@ -108,26 +108,36 @@ public static void saveSessionToPrefs() {
...
         }
-
         SharedPreferences prefs = ccApp.getAppPreferences();
+        String serializedSession = getSerializedSessionString();
+        if (!"".equals(serializedSession)) {
+            prefs.edit().putString(CommCarePreferences.CURRENT_SESSION, serializedSession).commit();
+        }
+    }
+
+    public static String getSerializedSessionString() {
```
ctsims: How similar is this method to your utility set above?

ctsims:
minor comment and question, otherwise good to pull once you're checked off on "don't pull"

app/src/org/commcare/android/logging/ForceCloseLogSerializer.java
amstone326: @ctsims Yea that's a good point. Would probably need to wait to pull this til that happens? Or I could add something in temporarily that puts them in the normal device logs as well. Not sure what makes the most sense there.

app/src/org/commcare/android/session/DevSessionRestorer.java
amstone326: @ctsims Sorry, not sure what you're asking about. What utility above?

app/src/org/commcare/android/logging/ForceCloseLogSerializer.java
ctsims: I think adding it to both when we log is probably the easiest short-term solution if it's not too onerous.

app/src/org/commcare/android/session/DevSessionRestorer.java
ctsims: Ah, I think I was confusing what you're calling the "readable session string" above, and the "Serialized Session String" here.

app/src/org/commcare/android/logging/ForceCloseLogger.java
```diff
@@ -112,12 +117,31 @@ public String getFilename() {
...
+        }
+    }
+
+    private static String getSubmissionUri() {
+        try {
+            SharedPreferences settings = CommCareApplication._().getCurrentApp().getAppPreferences();
+            return settings.getString(CommCarePreferences.PREFS_SUBMISSION_URL_KEY,
+                    CommCareApplication._().getString(R.string.PostURL));
+        } catch (Exception e) {
+            return CommCareApplication._().getString(R.string.PostURL);
```
phillipm: Are we trying to catch NPEs when there isn't an app seated? I would prefer explicit `null` checks or catching `NullPointerException` over the general `Exception`

app/src/org/commcare/android/logging/ForceCloseLogger.java
```diff
@@ -112,12 +117,31 @@ public String getFilename() {
...
+    private static void writeErrorToStorage(ForceCloseLogEntry entry) {
+        if (logStorage != null) {
+            logStorage.write(entry);
+        }
+    }
+
+    private static String getSubmissionUri() {
+        try {
+            SharedPreferences settings = CommCareApplication._().getCurrentApp().getAppPreferences();
+            return settings.getString(CommCarePreferences.PREFS_SUBMISSION_URL_KEY,
```
phillipm: There are a lot of places that have these two lines of code, might be good to eventually replace them with this method.

app/src/org/commcare/CommCareApplication.java
```diff
@@ -947,7 +957,16 @@ public void onServiceConnected(ComponentName className, IBinder service) {
                         }
                     }
 
-                    XPathErrorLogger.registerStorage(getUserStorage(XPathErrorEntry.STORAGE_KEY, XPathErrorEntry.class));
+                    // Switch the regular and forece close loggers over to using user storage while
+                    // there is a session
+                    Logger.registerLogger(new AndroidLogger(getUserStorage(AndroidLogEntry.STORAGE_KEY,
+                            AndroidLogEntry.class)));
+                    ForceCloseLogger.registerStorage(getUserStorage(ForceCloseLogEntry.STORAGE_KEY,
+                            ForceCloseLogEntry.class));
```
phillipm: These logger registrations could be lifted into a helper like `void setupLoggers(boolean isUserStorageAvailable)`

app/src/org/commcare/android/logging/ForceCloseLogSerializer.java
```diff
@@ -0,0 +1,102 @@
...
+import org.javarosa.core.model.utils.DateUtils;
+import org.javarosa.core.services.storage.EntityFilter;
+import org.javarosa.core.util.SortedIntSet;
+import org.xmlpull.v1.XmlSerializer;
+
+import java.io.IOException;
+import java.util.Hashtable;
+
+/**
+ * Convert xpath error logs to xml.
```
phillipm: needs to be updated

app/src/org/commcare/android/logging/ForceCloseLogSerializer.java
```diff
@@ -0,0 +1,102 @@
...
+            writeText("session_readable", forceCloseEntry.getReadableSession());
+            writeText("session_serialized", forceCloseEntry.getSerializedSessionString());
+        } catch (Exception e) {
+            e.printStackTrace();
+        } finally {
+            serializer.endTag(DeviceReportWriter.XMLNS, "force_close");
+        }
+    }
+
+    private void writeText(String element, String text)
```
phillipm: Duplicated in `XPathErrorSerializer`

app/src/org/commcare/android/logging/ForceCloseLogSerializer.java
```diff
@@ -0,0 +1,102 @@
...
+                        return IDs.contains(id) ? PREFILTER_INCLUDE : PREFILTER_EXCLUDE;
+                    }
+
+                    public boolean matches(LogEntry e) {
+                        throw new RuntimeException("can't happen");
+                    }
+                });
+            }
+        });
+    }
```
phillipm: Duplicated in `XPathErrorSerializer`. Might be worth making them both subclasses of the same base class or having some static purge setup methods they can both call

app/src/org/commcare/android/logging/ForceCloseLogger.java
```diff
@@ -26,61 +26,65 @@
...
-            report = null;
-        }
-
+    /**
+     * Attempts to send a force close report for the given exception to the server
+     * immediately. If this fails, instead writes the exception to storage as a ForceCloseLogEntry,
+     * so that we can attempt to send it again later during a normal LogSubmissionTask
+     */
+    private static void sendToServerOrStore(Throwable exception) {
+        ByteArrayOutputStream streamToWriteErrorTo = new ByteArrayOutputStream();
```
phillipm: This stream is never closed

phillipm: Nevermind, it is in `DeviceReportWriter.write()`

phillipm: Errr, unless the `DeviceReportWriter` constructor throws an `IOException`, then it isn't closed...

app/src/org/commcare/logging/AndroidLogEntry.java
```diff
@@ -37,30 +33,24 @@ public AndroidLogEntry() {
     }
 
     public AndroidLogEntry(String type, String message, Date date) {
-        this.type = type;
-        this.message = message;
-        this.date = date;
+        super(type, message, date);
```
phillipm: :+1:

app/src/org/commcare/logging/AndroidLogSerializer.java
```diff
@@ -78,11 +79,12 @@ public void writeToDeviceReport(XmlSerializer serializer) throws IOException {
         serializer.startTag(DeviceReportWriter.XMLNS, "log_subreport");
 
         try {
-            if (isSingleLog) {
-                serializeLog(entry);
-            } else {
-                if (Logger._() != null) {
-                    Logger._().serializeLogs(this);
```
phillipm: Seems like `serializeLogs` is not called anymore. Maybe you can remove the implementations from `AndroidLogger` then?

app/src/org/commcare/preferences/DevSessionRestorer.java
```diff
@@ -108,27 +108,37 @@ public static void saveSessionToPrefs() {
...
+        String serializedSession = getSerializedSessionString();
+        String formEntrySession = FormEntryActivity.getFormEntrySessionString();
+
+        ccApp.getAppPreferences().edit()
+                .putString(CommCarePreferences.CURRENT_SESSION, serializedSession)
+                .putString(CommCarePreferences.CURRENT_FORM_ENTRY_SESSION, formEntrySession)
+                .apply();
+    }
+
+    public static String getSerializedSessionString() {
```
phillipm: :+1:

phillipm:
Thanks for doing this, it was much needed!
Left some comments that should be addressed, then good to go!

app/src/org/commcare/android/logging/ForceCloseLogger.java
amstone326: I don't know for sure what it's trying to catch here, this was just refactored from this: https://github.com/dimagi/commcare-odk/blob/master/app/src/org/commcare/tasks/ExceptionReporting.java#L74-L80. I can't really imagine what else it would be protecting against besides the current app being null though, so feel comfortable changing it to that.

app/src/org/commcare/android/logging/ForceCloseLogger.java
amstone326: Will probably leave that for another PR

app/src/org/commcare/android/logging/ForceCloseLogSerializer.java
amstone326: whoops

app/src/org/commcare/android/logging/ForceCloseLogger.java
amstone326: Hm, confused what you mean on this for a couple of reasons. The `finally` block of `DeviceReportWriter.write()` does call `os.close()`, so not sure what you mean by the last comment. But is also worth noting that the default implementations of `close()` for both `OutputStream` and `ByteArrayOutputStream` do nothing, so unless we introduce an override, it's kind of irrelevant. I'm not sure what an overridden version of that method would do, but if you have an idea definitely share.

app/src/org/commcare/logging/AndroidLogSerializer.java
amstone326: Hm, I can't do that unless I make `AndroidLogger` abstract because it inherits those methods from `ILogger`, which is an interface. And it does look like j2me still makes use of the `serializeLogs` methods in `ILogger`, so those methods can't be totally removed. I could make the `AndroidLogger` implementations return `null` though, to indicate that they're unused. Not sure how I feel about that pattern, but open to it.

amstone326:
@phillipm thanks for the feedback. Implemented all suggestions except for a couple that I left questions on.

PR #1070, opened 2016-02-12
Capture and restore form entry session

phillipm:
Allows for capturing form entry actions and replaying them using the "[Enable Session Saving](https://confluence.dimagi.com/display/ccinternal/CommCare+Android+Developer+Options#CommCareAndroidDeveloperOptions-EnableSessionSaving)" dev option.

For now doesn't support:
 - replaying answers with "(" or ")" characters

cross-request: https://github.com/dimagi/javarosa/pull/248


app/src/org/commcare/views/QuestionsView.java
```diff
@@ -375,7 +375,9 @@ public void setOnFocusChangeListener(OnFocusChangeListener l) {
     public void teardownView() {
         for (QuestionWidget widget : widgets) {
             widget.unsetListeners();
+            widget.setOnCreateContextMenuListener(null);
```
amstone326: What are these changes doing?

phillipm: This was me randomly trying to fix some memory leaks that I was running into early on in this PR.
Worth leaving in because this might cause memory to leak. 
Regardless, Clayton helped me get around the memory leaks by taking an approach that didn't use Views at all when restoring question answers from the saved sessions.

PR #1078, opened 2016-02-17
Catch arity exceptions in form entry and hierarchy views

phillipm:
@esoergel was encountering xpath arity errors that were crashing commcare instead of showing the user the error, which is user addressable.

https://github.com/dimagi/commcare-odk/commit/3a9ad0004f8fb2e33cb505d04e272b6c67d9ce14 adds additional xpath arity exception checks; I'm unsure if they are necessary...

xpath arity checks caught in the form hierarchy activity show a toast message because that activity doesn't extend `CommCareActivity` (it needs to extend the non-Fragment compatible `ListActivity`).

http://manage.dimagi.com/default.asp?217408


esoergel:
Nice, thanks for the fix!

PR #1088, opened 2016-02-23
Exit form entry on form save error due to missing case data

phillipm:
Currently when form saving fails due to case update referencing non-existent case,  a cryptic pinned notification is shown and the user is left in form entry.

Instead, show the user a blocking error with a hopefully more helpful message and when they press okay, back out of form entry.

screen

screen

Additional changes in `InstanceProvider` to address potential null pointer exceptions.

Ticket: http://manage.dimagi.com/default.asp?213001

cross-request: https://github.com/dimagi/javarosa/pull/255
cross-request: https://github.com/dimagi/commcare/pull/262


app/src/org/commcare/android/tasks/templates/CommCareTask.java
```diff
@@ -78,9 +78,9 @@ protected void onPostExecute(Result result) {
                 connector.stopBlockingForTask(taskId);
                 if (unknownError != null) {
                     deliverError(connector.getReceiver(), unknownError);
-                    return;
+                } else {
+                    deliverResult(connector.getReceiver(), result);
```
phillipm: Allows dialog to be dismissed even when error occurs

app/src/org/odk/collect/android/activities/FormEntryActivity.java
```diff
@@ -2245,6 +2244,14 @@ private int getCurrentFormID() {
         return mFormController.getFormID();
     }
 
+    public void handleFormSaveError(Exception e) {
+        if (e instanceof IllegalStateException) {
+            // TODO PLM: send this error to HQ as a app build error, most likely a user level issue.
```
phillipm: Note the TODO. Don't really want to do this right now because it is extra work and this isn't even setup on HQ yet.

app/src/org/odk/collect/android/tasks/SaveToDiskTask.java
```diff
@@ -89,30 +91,46 @@ public SaveToDiskTask(Uri mUri, Boolean saveAndExit, Boolean markCompleted, Stri
...
-                return SaveStatus.SAVED_COMPLETE;
-            } else {
-                return SaveStatus.SAVED_INCOMPLETE;
-            }
+        try {
+            exportData(mMarkCompleted);
+        } catch (FileNotFoundException e) {
+            e.printStackTrace();
+            return new Pair<>(SaveStatus.SAVE_ERROR,
+                    "Something is blocking acesss to the submission file in " + FormEntryActivity.mInstancePath);
```
ctsims: Are these getting bubbbled up to the user? If so we possibly want them to be localizable...

app/src/org/commcare/dalvik/odk/provider/InstanceProvider.java
```diff
@@ -517,6 +537,10 @@ private void linkToSessionFormRecord(Uri instanceUri) {
             if (FormRecord.STATUS_COMPLETE.equals(current.getStatus())) {
                 try {
                     new FormRecordProcessor(getContext()).process(current);
+                } catch (InvalidStructureException | InvalidStorageStructureException e) {
+                    // record should be wiped when form entry is exited
+                    Logger.log(AndroidLogger.TYPE_ERROR_WORKFLOW, e.getMessage());
```
ctsims: I'm a bit surprised this is the logging category we chose. Are we bubbling these up in some new way?

app/src/org/odk/collect/android/tasks/SaveToDiskTask.java
phillipm: They are. Is this something we want to start doing? Should I also localize the messages here: https://github.com/dimagi/commcare/pull/262/files ?
They will also be shown to the user in the same format (see screenshot for this PR). It is a little tricky because of how the parser context is added (https://github.com/dimagi/javarosa/pull/255/files#diff-512400e8172a1fc10562b6d6bf852619R16)

app/src/org/commcare/dalvik/odk/provider/InstanceProvider.java
phillipm: I'm just copying this line from the catch directly below, which was the old code path

orangejenny:
Looks good, left a couple of minor questions.

phillipm:
@orangejenny  Thanks for the close review and questions :mag_right: :muscle: :muscle: :mag:
I responded inline and I think things should be good to go!

PR #1089, opened 2016-02-24
Check if CallOutActivity.onCreate bundle is null before referencing it

phillipm:
fix for https://commcare-acralyzer.cloudant.com/acralyzer/_design/acralyzer/index.html#/reports-browser/commcare-odk/bug/941a986da24b6c57417032737cfd4be9

might need to go into a 2.26 hotfix


ctsims:
@phillipm Do we want to start pushing these into a hotfix branch, rather than master?

PR #1094, opened 2016-02-25
Refactor: move callout and some UI logic out of EntitySelectActivity

phillipm:
Get `EntitySelectActivity` to below the 1000 line-count mark by moving callout logic into its own file and some ui setup logic into another file.


phillipm:
@orangejenny I've added the localization texts

orangejenny:
thanks!

PR #1095, opened 2016-02-25
Show to user that task is cancelling when the press 'stop'

phillipm:
Sometimes tasks take forever to cancel because we don't actually check the value of  `AsyncTask.isCancelled()` regularly. This is confusing for me, and probably users, because they think that the button is broken.

This PR changes the blocking task dialog text to signify when the task was cancelled.

screen

Eventually we should do a better job of stopping the process when cancel is pressed, but that is sorta hard.

http://manage.dimagi.com/default.asp?217900

### DIFF
--- a/app/src/org/commcare/dalvik/dialogs/CustomProgressDialog.java
+++ b/app/src/org/commcare/dalvik/dialogs/CustomProgressDialog.java
@@ -1,16 +1,5 @@
 package org.commcare.dalvik.dialogs;
 
-/**
- * @author amstone326
- *
- * Any progress dialog associated with a CommCareTask should use
- * this class to implement the dialog. Any class that launches such a task
- * should implement the generateProgressDialog() method of the DialogController
- * interface, and create the dialog in that method. The rest of the dialog's
- * lifecycle is handled by methods of the DialogController interface that are
- * fully implemented in CommCareActivity
- */
-
 import android.app.AlertDialog;
 import android.app.Dialog;
 import android.content.Context;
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/dalvik/dialogs/CustomProgressDialog.java
```diff
@@ -200,17 +196,37 @@ public void onClick(View v) {
...
+            TextView titleView = (TextView)pd.findViewById(R.id.progress_dialog_title).findViewById(R.id.dialog_title_text);
+            TextView messageView = (TextView)pd.findViewById(R.id.progress_dialog_message);
+            Button cancelButton = (Button)pd.findViewById(R.id.dialog_cancel_button);
+            setCancellingText(titleView, messageView, cancelButton);
+        }
+    }
+
+    private void setCancellingText(TextView titleTextView, TextView messageTextView, Button cancelButton) {
+        titleTextView.setText(cancellingPrefixText + " '" + title + "'");
+        messageTextView.setText(cancellingPrefixText + "...");
```
ctsims: This kind of verbage is full of some dangerous assumptions about the nature of the language involved. If we were in a right-to-left language this would be really confusing to put together.

Two options that might help:

1) Use the localization keys to do the construction, IE: 

Localization.get("cancelling.title", new String[] {titleText})
where
cancelling.title=Cancelling: ${0}

or something similar. 

Alternatively, you could only use the message view for the cancel message (although I'd put the "..." in the message itself for the same reason)

phillipm: :+1: updated, see the :rice_scene:

amstone326:
Nice, this was much-needed.

PR #1098, opened 2016-02-26
Schedule form relevancy update when DateWidget changes

phillipm:
Changing the value of a date picker question widget didn't update the relevancy of form questions. Enabling relevancy updates normally crashes the app when you spin the date changer because it is adding and removing question views willy-nilly which causes crashes. To get around that, schedule a change to be fired about a second later.

Tested on Android 6, 4.4.2, and 2.3.7 devices

http://manage.dimagi.com/default.asp?112607


app/src/org/commcare/views/widgets/DateWidget.java
```diff
@@ -79,24 +88,14 @@ public void onDateChanged(DatePicker view, int year, int month, int day) {
...
-                            return;
                         }
                     }
                 }
-
-                //TODO: Not here, the change isn't processed yet.
-                widgetEntryChanged();
-
+                if (!isRelevancyUpdateScheduled) {
+                    mainHandler.postDelayed(updateFormRelevancyRunnable, 750);
```
orangejenny: shouldn't `isRelevancyUpdateScheduled` be set to true here?

phillipm: :+1:

orangejenny:
Tests aren't running properly; I'm just going to merge this.

PR #1101, opened 2016-02-29
Make wifi direct screen scrollable

phillipm:
Allow for scrolling of the entire wifi direct view. Addresses issue where buttons weren't visible on smaller screens.

Use our UI Controller pattern (also seen in `CommCareHomeActivity, `UpdateActivity`, and `LoginActivity`) with WiFi Direct to control the visibility of buttons. This replaces the use of an adapter and staggered grid view, because it is harder to control scrolling using those constructs and I feel like this use case doesn't quite warrant an adapter.

screen

There is still an issue with button height in the `receive` wifi direct state:
screen
Not quite sure how to remedy it.

Also note that the drop shadows on the buttons are now gone :disappointed:, I can look into adding them back if you want to go forward with this PR

http://manage.dimagi.com/default.asp?218687

cc @wpride


phillipm:
Now I'm wondering if just wrapping the entire layout in a `ScrollView` would have rendered the same results... If so and you aren't into the UI Controller approach, feel free to close and go that route.

app/src/org/commcare/android/adapters/WiFiDirectUIController.java
```diff
@@ -0,0 +1,114 @@
...
+        this.activity = activity;
+    }
+
+    @Override
+    public void setupUI() {
+        setupButtonListeners();
+    }
+
+    private void setupButtonListeners() {
+        modeButton = (SquareButtonWithText)activity.findViewById(R.id.mode);
```
ctsims: Now that we have a pattern for our controllers we should see what we can do to use the annotation style harnessing that is in the old Activitiy framework to get rid of some of this verbosity.

PR #1106, opened 2016-03-03
Update form progress on location question answering

phillipm:
Form progress wasn't getting updated when location questions were being answered.

http://manage.dimagi.com/default.asp?218955


app/src/org/commcare/views/widgets/GeoPointWidget.java
```diff
@@ -238,6 +238,9 @@ public void setBinaryData(Object answer) {
                         ": " + truncateDouble(sa[2]) + "m\n" +
                         StringUtils.getStringSpannableRobust(getContext(), R.string.accuracy) +
                         ": " + truncateDouble(sa[3]) + "m");
+        // update form relevancies and such
+        // TODO PLM: probably want to call this for every class that overrides setBinaryData
+        widgetEntryChanged();
```
phillipm: @ctsims, thoughts on this TODO?

ctsims: I think all other binary data questions require the question widget to get recreated entirely when they return from their callout, so good to note but probably not necessary across the board.

PR #1108, opened 2016-03-03
Hotfix: revert mime library version bump

phillipm:
revert bump from mime 4.0 to 4.5.1 version bump from https://github.com/dimagi/commcare-odk/pull/934

That version bump caused form attachments to be sent to HQ in a format that HQ can't handle, causing data loss for form attachments.


czue:
:+1: thanks for super quick turnaround!

PR #1109, opened 2016-03-03
Pass PostURL when zipping forms and use for submission

wpride:
SD Card and WiFi Direct transfer will pass the original user's PostURL along with the form so that the submitting phone can use the "correct" URL

See http://manage.dimagi.com/default.asp?217957

@ctsims open to feedback on this approach

cc @czue


app/src/org/commcare/tasks/SendTask.java
```diff
@@ -91,6 +94,26 @@ protected Boolean doTaskBackground(Void... params) {
...
+
+                // see if we have a form.properties file to load the PostURL from
+                FilenameFilter filter = new FilenameFilter() {
+                    @Override
+                    public boolean accept(File dir, String filename) {
+                        return filename.equals(ZipTask.FORM_PROPERTIES_FILE);
+                    }
+                };
+
+                File[] fileContents = f.listFiles(filter);
```
phillipm: I would assume that `f` stands for 'file', but here you are calling a directory specific method `listFiles` on it. Would be helpful to rename `f` to `dir` or `formInstanceDir` (if that is an accurate description).

phillipm: `fileContents` -> `formPropertyFiles`

app/src/org/commcare/tasks/SendTask.java
```diff
@@ -91,6 +94,26 @@ protected Boolean doTaskBackground(Void... params) {
...
+                if(fileContents != null && fileContents.length > 0){
+                    Properties properties = FileUtil.loadProperties(fileContents[0]);
+                    if(properties != null && properties.getProperty(ZipTask.FORM_PROPERTY_POST_URL) != null){
+                        url = properties.getProperty(ZipTask.FORM_PROPERTY_POST_URL);
+                        Logger.log(TAG, "Successfully got form.property PostURL: " + url);
```
phillipm: This will send the log to HQ with the tag `SendTask`, which is the only time such a tag is sent to HQ. Consider using `AndroidLogger.TYPE_FORM_DUMP` to match other wifi/sd form dump logging calls.

app/src/org/commcare/tasks/SendTask.java
```diff
@@ -91,6 +94,26 @@ protected Boolean doTaskBackground(Void... params) {
...
+                    }
+                    // don't submit this file
+                    FileUtil.deleteFileOrDir(fileContents[0]);
+                }
```
phillipm: This code seems pretty self-contained, can you lift all of this into a helper method?

app/src/org/commcare/tasks/ZipTask.java
```diff
@@ -128,9 +135,37 @@ private long dumpInstance(File folder, SecretKeySpec key) throws FileNotFoundExc
                 }
             }
         }
+
+        writeProperties(myDir);
+
         return FormUploadUtil.FULL_SUCCESS;
     }
 
+    private void writeProperties(File file) {
```
phillipm: Can you add a comment describing why this post url property file is needed?

app/src/org/commcare/tasks/SendTask.java
wpride: I thought about it but I think this is cleaner because 1) we can only override url if all the requisite condition are met and 2) we'll probably add more properties at some point and I didn't want to a map back and forth

phillipm: 1) I'm okay with helper methods changing an object's state (wasn't necessarily advocating that the method should be `static`).
2) I like to use helper methods to help mitigate the code readability impact of code growth, so in my opinion lifting this code into helper method(s) is a way to be proactive about the fact that more properties may get added later.

app/src/org/commcare/tasks/SendTask.java
```diff
@@ -83,34 +87,55 @@ protected Boolean doTaskBackground(Void... params) {
...
                     allSuccessful = false;
-                    CommCareApplication._().reportNotificationMessage(NotificationMessageFactory.message(StockMessages.Send_MalformedFile, new String[]{null, f.getName()}, MALFORMED_FILE_CATEGORY));
-                    publishProgress(Localization.get("bulk.send.file.error", new String[]{f.getAbsolutePath()}));
+                    CommCareApplication._().reportNotificationMessage(NotificationMessageFactory.message(StockMessages.Send_MalformedFile, new String[]{null, formFolder.getName()}, MALFORMED_FILE_CATEGORY));
+                    publishProgress(Localization.get("bulk.send.file.error", new String[]{formFolder.getAbsolutePath()}));
                 }
                 counter++;
             } catch (SessionUnavailableException | FileNotFoundException fe) {
-                Log.e("E", Localization.get("bulk.send.file.error", new String[]{f.getAbsolutePath()}), fe);
+                Log.e("E", Localization.get("bulk.send.file.error", new String[]{formFolder.getAbsolutePath()}), fe);
```
phillipm: `"E"` -> `TAG`

app/src/org/commcare/tasks/SendTask.java
```diff
@@ -83,34 +87,55 @@ protected Boolean doTaskBackground(Void... params) {
...
             publishProgress(Localization.get("bulk.send.dialog.progress", new String[]{"" + (i + 1)}));
 
-            File f = files[i];
+            File formFolder = files[i];
 
-            if (!(f.isDirectory())) {
-                Log.e("send", "Encountered non form entry in file dump folder at path: " + f.getAbsolutePath());
-                CommCareApplication._().reportNotificationMessage(NotificationMessageFactory.message(StockMessages.Send_MalformedFile, new String[]{null, f.getName()}, MALFORMED_FILE_CATEGORY));
+            if (!(formFolder.isDirectory())) {
+                Log.e("send", "Encountered non form entry in file dump folder at path: " + formFolder.getAbsolutePath());
```
phillipm: `"send"` -> `TAG`
(might need to add `private static final String TAG = SendTask.class.getSimpleName();`)

phillipm:
:+1: 
@ctsims has more context, so he should do a pass too before merge

app/src/org/commcare/tasks/ZipTask.java
```diff
@@ -128,9 +135,38 @@ private long dumpInstance(File folder, SecretKeySpec key) throws FileNotFoundExc
...
+            Properties properties = new Properties();
+            SharedPreferences settings = CommCareApplication._().getCurrentApp().getAppPreferences();
+            // HQ likes us to submit forms to the "correct" app and user specific URL
+            String postUrl = settings.getString(CommCarePreferences.PREFS_SUBMISSION_URL_KEY,
+                    c.getString(R.string.PostURL));
+            properties.setProperty("PostURL", postUrl);
+            properties.store(outputStream, null);
+        } catch(IOException e){
+            // we'll just ignore this, not the end of the world
+            e.printStackTrace();
```
ctsims: I don't really think this is a great plan. There's no reason we shouldn't fail the process entirely if this doesn't work.

app/src/org/commcare/utils/FileUtil.java
```diff
@@ -382,4 +382,26 @@ public static String getDumpDirectory(Context c) {
...
+
+    public static Properties loadProperties(File file){
+        Properties prop = new Properties();
+        InputStream input = null;
+        try {
+            input = new FileInputStream(file);
+            prop.load(input);
+            return prop;
+        } catch (IOException ex) {
+            ex.printStackTrace();
```
ctsims: Also not comfortable with this.

PR #1112, opened 2016-03-04
Backmerge form attachment hotfixes

phillipm:
Merge https://github.com/dimagi/commcare-odk/pull/1108 and https://github.com/dimagi/commcare-odk/pull/1111 into master.

### DIFF
--- a/app/src/org/commcare/tasks/PurgeStaleArchivedFormsTask.java
+++ b/app/src/org/commcare/tasks/PurgeStaleArchivedFormsTask.java
@@ -8,6 +8,8 @@ import org.commcare.models.database.user.models.FormRecord;
 import org.commcare.utils.FormUploadUtil;
 import org.javarosa.core.services.Logger;
 import org.joda.time.DateTime;
+import org.joda.time.format.DateTimeFormatter;
+import org.joda.time.format.ISODateTimeFormat;
 
 import java.io.File;
 import java.io.FileNotFoundException;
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/tasks/PurgeStaleArchivedFormsTask.java
```diff
@@ -156,4 +168,106 @@ private static DateTime parseModifiedDate(String dateString)
         Date parsed = format.parse(dateString);
         return new DateTime(parsed);
     }
+
+    /**
+     * This is a temporary fix for a problem which occured with specific dates where forms with
+     * attachmetns were not received properly. It can be removed on June 1, 2016
```
phillipm: @ctsims Curious about the date you chose here.

app/src/org/commcare/tasks/PurgeStaleArchivedFormsTask.java
```diff
@@ -57,6 +66,9 @@ public static void launchPurgeTask() {
     @Override
     protected Void doInBackground(Void... params) {
         CommCareApp app = CommCareApplication._().getCurrentApp();
+
+        performUnsentAttachmentHotfix(app);
```
phillipm: Wondering what the price is for this procedure. If a device has thousands of saved forms it will take a while for this to run. That said it happens in the background and only happen once, so I guess we are fine.

PR #1113, opened 2016-03-04
Merge FileUtil and FileUtils

wpride:
We've had the FileUtils laying around from the collect merge for a while even though there's redundant methods and its annoying. Just merged all of the methods from FileUtils. Redundant method signatures were:

getMd5Hash ~ identical
createFolder - identical
copyFile(String, String) - this one was implemented differently, one using a FileChanel and one just using the copyFile(String, String, Cipher, Cipher) method with null ciphers and I/O streams. @ctsims might need to confirm these weren't intentionally different for some reason

EDIT: copyFile was actually slightly uglier since one version threw an IOException and the other didn't. I added try/catches in all the old collect classes so the behavior (log and proceed) should be the same.


app/src/org/commcare/utils/FileUtils.java
```diff
@@ -1,364 +0,0 @@
...
-     *
-     * @param sourceFile A file pointer to a file on the file system
-     * @param destFile   Either a file or directory. If a directory, the
-     *                   file name will be taken from the source file
-     */
-    public static void copyFile(File sourceFile, File destFile) {
-        if (sourceFile.exists()) {
-            if (destFile.isDirectory()) {
-                destFile = new File(destFile, sourceFile.getName());
-            }
```
phillipm: `FileUtil.copyFile` doesn't exhibit this behavior of allowing the destFile to be a directory. Looks like everywhere that calls this passes in a file, so it should be fine, except for in `DrawActivity` where it is unclear whether a dir or file is passed in.
I would leave this implementation in for simplicity and since it might be _slightly_ more efficient to use built-in channels instead of the hand-rolled stream writing in `AndroidStreamUtils`.

phillipm: Though it is good to allow the `IOException` to be thrown upwards...

phillipm:
:+1: after my comment is addressed

PR #1114, opened 2016-03-06
Wifi direct

wpride:
One major functionality change and a lot of refactor to make this functionality cleaner.

ADRA needs the ability to send forms via Wi-Fi Direct multiple times - IE the tablet collecting the forms via WiFiDirect from the front line workers itself will not ever get connectivity and thus needs to transfer these forms to yet another tablet that will be returned to the home office and sync with the server. This change supports this workflow. Essentially, after pulling any extant FormRecords from the database into the file system, this activity will now check the "to be submitted" folder for any forms it previously received via wifi direct. If these are found, we move these into the same folder as the form records we just pulled. Once there, forms from the two sources are functionally equivalent and we can proceed from there. This is accomplished here: https://github.com/dimagi/commcare-odk/compare/wifi-direct?expand=1#diff-1ec3d20f602697a29ebc1680b992c3a3R627

The biggest functional refactor is separating the process of moving files from the database to storage and zipping these files into separate tasks so that we can do the step above in between. there's some other cleanup besides.


wpride:
Release Note: Wi-Fi Direct now supports multiple transfers between devices

phillipm:
contains commits from: https://github.com/dimagi/commcare-odk/pull/1109

app/src/org/commcare/tasks/FormRecordToFileTask.java
```diff
@@ -0,0 +1,278 @@
...
+     */
+    private long getFileInstanceFromStorage(File formRecordFolder, SecretKeySpec decryptionKey)
+            throws FileNotFoundException, SessionUnavailableException {
+        File[] files = formRecordFolder.listFiles();
+        Logger.log(TAG, "Trying to get instance with: " + files.length + " files.");
+
+        File myDir = new File(storedFormDirectory, formRecordFolder.getName());
+        myDir.mkdirs();
+
+        if (files == null) {
```
phillipm: this is always false because an NPE will be thrown in the log statement above, with `files.lengtt`, if `files` was `null`

app/src/org/commcare/tasks/FormRecordToFileTask.java
```diff
@@ -0,0 +1,278 @@
...
+                if (file.getName().endsWith(ext)) {
+                    supported = true;
+                    break;
+                }
+            }
+            if (!supported) {
+                continue;
+            }
+            bytes += file.length();
+        }
```
phillipm: This `for`-loop can be lifted into a named helper method

app/src/org/commcare/tasks/FormRecordToFileTask.java
```diff
@@ -0,0 +1,278 @@
...
+            } else {
+                try {
+                    FileUtil.copyFile(file, new File(myDir, file.getName()));
+                } catch (IOException ie) {
+                    Log.d(TAG, "Copying file: " + file.getName() + " failed with: " + ie.getMessage());
+                    publishProgress(("File writing failed: " + ie.getMessage()));
+                    return FormUploadUtil.FAILURE;
+                }
+            }
+        }
```
phillipm: Lifting this `for`-loop into a helper method that throws an `IOException` will help remove the duplicated error handling present here.

app/src/org/commcare/tasks/FormRecordToFileTask.java
```diff
@@ -0,0 +1,278 @@
...
+        storedFormDirectory.mkdirs();
+
+        SqlStorage<FormRecord> storage = CommCareApplication._().getUserStorage(FormRecord.class);
+        Vector<Integer> ids = StorageUtils.getUnsentOrUnprocessedFormsForCurrentApp(storage);
+
+        if (ids.size() > 0) {
+            FormRecord[] records = new FormRecord[ids.size()];
+            for (int i = 0; i < ids.size(); ++i) {
+                records[i] = storage.read(ids.elementAt(i).intValue());
```
phillipm: `intValue()` isn't needed

app/src/org/commcare/tasks/FormRecordToFileTask.java
```diff
@@ -0,0 +1,278 @@
...
+            }
+
+            results = new Long[records.length];
+            for (int i = 0; i < records.length; ++i) {
+                //Assume failure
+                results[i] = FormUploadUtil.FAILURE;
+            }
```
phillipm: This loop isn't needed if you move the work into the loop above. They should iterate over the same range.

app/src/org/commcare/tasks/FormRecordToFileTask.java
```diff
@@ -0,0 +1,278 @@
...
+                                break;
+                            }
+                            continue;
+                        }
+
+                        //Check for success
+                        if (results[i].intValue() == FormUploadUtil.FULL_SUCCESS) {
+                            //FormRecordCleanupTask.wipeRecord(c, platform, record);
+                            //publishProgress(Localization.get("bulk.form.dialog.progress",new String[]{""+i, ""+results[i].intValue()}));
+                        }
```
phillipm: can you take out this dead code?

app/src/org/commcare/activities/CommCareWiFiDirectActivity.java
```diff
@@ -659,9 +730,9 @@ protected void deliverError(CommCareWiFiDirectActivity receiver, Exception e) {
             }
 
             @Override
-            protected void deliverResult(CommCareWiFiDirectActivity receiver, FormRecord[] result) {
-                if (result != null) {
-                    receiver.onZipSuccesful(result);
+            protected void deliverResult(CommCareWiFiDirectActivity receiver, Integer result) {
+                if (result == ZipTask.RESULT_SUCCESS) {
```
phillipm: Do you need to unbox the result with an `intValue()`? I'm not sure, but you manually unbox in other places.

app/src/org/commcare/tasks/FormRecordToFileTask.java
wpride: +1

app/src/org/commcare/activities/CommCareWiFiDirectActivity.java
ctsims: This definitely needs to be done with intValue, Integer comparisons are unpredictable (sometimes they are cached/interned sometimes not, so it can work one day but not the nexst)

app/src/org/commcare/activities/CommCareWiFiDirectActivity.java
```diff
@@ -595,21 +599,51 @@ public void prepareFileTransfer() {
...
+        File receiveFolder = new File(toBeSubmittedDirectory);
+        if (!receiveFolder.isDirectory() || !receiveFolder.exists()) {
+            // we haven't received any transfers, just return.
+            return;
+        }
+        // We've received transferred forms. Move them into the to be zipped directory.
+        File[] files = receiveFolder.listFiles();
+        try {
+            Log.d(TAG, "We have " + files.length + " toBeSubmitted files to transfer");
+            for (File instanceDirectory : files) {
```
phillipm: is `instanceDirectory` a directory or a file? Can you confirm it is a directory, and if not, rename to something that denotes a file.

phillipm: Oh, nevermind, read deeper and they are files or dirs... either way ` `copyFileDeep` is handling it.
I get so confused by file/dir ambiguities.

app/src/org/commcare/activities/CommCareWiFiDirectActivity.java
```diff
@@ -595,21 +599,51 @@ public void prepareFileTransfer() {
...
+                Log.d(TAG, "Transferred " + instanceDirectory + " to " + toBeTransferredDirectory
+                        + "/" + instanceDirectory.getName());
+            }
+        } catch(IOException e){
+            // if we catch an error, delete all the new files we've copied over
+            for (File instanceDirectory : files) {
+                File newDestination = new File(toBeTransferredDirectory, instanceDirectory.getName());
+                if(newDestination.exists()){
+                    FileUtil.deleteFileOrDir(newDestination);
+                }
```
phillipm: I would recommend making a new list with all the `files` entries transformed into the  `newDestination` versions outside of this try/catch. That way, you don't have to repeat `File newDestination = new File(toBeTransferredDirectory, instanceDirectory.getName());` twice, which feels cluttered when reading.

app/src/org/commcare/activities/CommCareWiFiDirectActivity.java
```diff
@@ -595,21 +599,51 @@ public void prepareFileTransfer() {
...
+            }
+        }
+        // Delete all the to be submitted files, we have them copied for transfer.
+        for (File instanceDirectory : files) {
+            if(instanceDirectory.exists()){
+                FileUtil.deleteFileOrDir(instanceDirectory);
+            }
+        }
```
phillipm: If you follow the advice above, then you can lift the 3 `for` loops in this method into a helper method that essentially looks like this loop, just taking in a `File[] files` param

app/src/org/commcare/activities/CommCareWiFiDirectActivity.java
```diff
@@ -637,9 +708,9 @@ protected void deliverError(CommCareWiFiDirectActivity receiver, Exception e) {
             }
 
             @Override
-            protected void deliverResult(CommCareWiFiDirectActivity receiver, FormRecord[] result) {
-                if (result != null) {
-                    receiver.onZipSuccesful(result);
+            protected void deliverResult(CommCareWiFiDirectActivity receiver, Integer result) {
+                if (result.intValue() == ZipTask.RESULT_SUCCESS) {
```
phillipm: Can you make `ZipTask` return an `enum` of possible outcomes instead of an `Integer` flag? This will make it easier to ensure all outcomes are handled and equality checks of the flags won't need to use unboxing because enum instances are always singletons; allowing for reference equality.

app/src/org/commcare/activities/CommCareWiFiDirectActivity.java
```diff
@@ -822,6 +894,10 @@ public CustomProgressDialog generateProgressDialog(int taskId) {
                 title = localize("wifi.direct.wipe.task.title").toString();
                 message = localize("wifi.direct.wipe.task.message").toString();
                 break;
+            case FormRecordToFileTask.PULL_TASK_ID:
+                title = localize("wifi.direct.pull.task.title").toString();
+                message = localize("wifi.direct.pull.task.message").toString();
```
phillipm: What is the difference between doing a `localize().toString()` and a `Localization.get()`? Does the first format markup and then discard it in the `toString`? If so we should consider encapsulating that logic in a documented method.

app/src/org/commcare/tasks/ZipTask.java
```diff
@@ -235,133 +110,26 @@ private boolean zipFolder(File[] files, String zipFile, ZipOutputStream zos) thr
...
-            ByteArrayOutputStream bos = new ByteArrayOutputStream();
-            e.printStackTrace(new PrintStream(bos));
-            return new String(bos.toByteArray());
-        } catch (Exception ex) {
-            return null;
+            String zipPath = toBeZippedPath;
+            File nf = new File(zipPath);
+            if (nf.exists()) {
+                nf.delete();
+           }
```
phillipm: Could you replace this with the same `FileUtil.deleteFileOrDir(toBeZippedDir);` as below?

app/src/org/commcare/tasks/ZipTask.java
```diff
@@ -235,133 +110,26 @@ private boolean zipFolder(File[] files, String zipFile, ZipOutputStream zos) thr
...
+    protected Integer doTaskBackground(Void... params) {
+        Log.d(TAG, "Doing UnzipTask");
         try {
-            ByteArrayOutputStream bos = new ByteArrayOutputStream();
-            e.printStackTrace(new PrintStream(bos));
-            return new String(bos.toByteArray());
-        } catch (Exception ex) {
-            return null;
+            String zipPath = toBeZippedPath;
+            File nf = new File(zipPath);
```
phillipm: There are 2 places in this method where you do `new File(toBeZippedPath)`, which is the only use of `toBeZippedPath`. I'd recommend doing it once in the constructor and storing it as a final class field instead of having `toBeZippedPath` be a class field.

app/src/org/commcare/tasks/ZipTask.java
```diff
@@ -200,21 +72,24 @@ private boolean zipTargetFolder(File targetFilePath, String zipFile) throws IOEx
         return false;
     }
 
-    private boolean zipFolder(File[] files, String zipFile, ZipOutputStream zos) throws IOException {
-        Log.d(TAG, "827 zipping folder with files: " + files[0] + ", zipFile: " + zipFile);
+    private void zipInstanceFolder(File[] toBeZippedFiles, String zipFilePath, ZipOutputStream zos)
```
phillipm: can be `static`

app/src/org/commcare/tasks/ZipTask.java
```diff
@@ -1,197 +1,69 @@
...
-            }
-        }
+    public ZipTask(String toBeZippedPath, String zipFilePath) {
+        taskId = ZIP_TASK_ID;
+        this.toBeZippedPath = toBeZippedPath;
+        this.zipFilePath = zipFilePath;
     }
 
-    private boolean zipTargetFolder(File targetFilePath, String zipFile) throws IOException {
+    private boolean zipParentFolder(File toBeZippedDirectory, String zipFilePath) throws IOException {
```
phillipm: can be `static`

app/src/org/commcare/tasks/ZipTask.java
phillipm: The `zipFilePath` argument is only used in a log statement, which feels a bit strange. Can you move the log statement outside of this method call and remove the `zipFilePath` arg?

app/src/org/commcare/tasks/ZipTask.java
```diff
@@ -200,21 +72,24 @@ private boolean zipTargetFolder(File targetFilePath, String zipFile) throws IOEx
...
     }
 
-    private boolean zipFolder(File[] files, String zipFile, ZipOutputStream zos) throws IOException {
-        Log.d(TAG, "827 zipping folder with files: " + files[0] + ", zipFile: " + zipFile);
+    private void zipInstanceFolder(File[] toBeZippedFiles, String zipFilePath, ZipOutputStream zos)
+            throws IOException {
+        Log.d(TAG, "Zipping instance folder with files: " + Arrays.toString(toBeZippedFiles)
+                + ", zipFilePath: " + zipFilePath);
+
         int BUFFER_SIZE = 1024;
```
phillipm: Is there a reason you aren't using the default buffer size specified by `BufferedInputStream`?

phillipm: Oh, I see, it is being used to write data from one stream to another, got it

app/src/org/commcare/tasks/ZipTask.java
```diff
@@ -223,7 +98,7 @@ private boolean zipFolder(File[] files, String zipFile, ZipOutputStream zos) thr
                 String fileName = pathParts[pathPartsLength - 1];
                 String fileFolder = pathParts[pathPartsLength - 2];
 
-                Log.d(TAG, "827 zipping folder with path: " + fileFolder + "/" + fileName);
+                Log.d(TAG, "Zipping instance folder with path: " + fileFolder + "/" + fileName);
 
                 ZipEntry entry = new ZipEntry(fileFolder + "/" + fileName);
                 zos.putNextEntry(entry);
```
phillipm: The docs for `putNextEntry` say that `closeEntry` should be called after the entry data has been written using `write`. `putNextEntry` does this for you if you forgot, but it won't for the last call entry that is added.

app/src/org/commcare/tasks/ZipTask.java
```diff
@@ -1,197 +1,69 @@
...
 
-            File[] fileArray = targetFilePath.listFiles();
-
-            for (File file : fileArray) {
-                File[] subFileArray = file.listFiles();
-                zipFolder(subFileArray, zipFile, out);
+            for (File formInstanceFolder : formInstanceFolders) {
+                File[] subFileArray = formInstanceFolder.listFiles();
+                zipInstanceFolder(subFileArray, zipFilePath, out);
             }
```
phillipm: Would recommend calling `out.finish()`

app/src/org/commcare/tasks/ZipTask.java
```diff
@@ -223,7 +98,7 @@ private boolean zipFolder(File[] files, String zipFile, ZipOutputStream zos) thr
                 String fileName = pathParts[pathPartsLength - 1];
                 String fileFolder = pathParts[pathPartsLength - 2];
 
-                Log.d(TAG, "827 zipping folder with path: " + fileFolder + "/" + fileName);
+                Log.d(TAG, "Zipping instance folder with path: " + fileFolder + "/" + fileName);
 
                 ZipEntry entry = new ZipEntry(fileFolder + "/" + fileName);
```
phillipm: Consider `fileFolder + "/" + fileName` -> `file.getPath()`

app/src/org/commcare/tasks/ZipTask.java
```diff
@@ -223,7 +98,7 @@ private boolean zipFolder(File[] files, String zipFile, ZipOutputStream zos) thr
                 String fileName = pathParts[pathPartsLength - 1];
                 String fileFolder = pathParts[pathPartsLength - 2];
 
-                Log.d(TAG, "827 zipping folder with path: " + fileFolder + "/" + fileName);
+                Log.d(TAG, "Zipping instance folder with path: " + fileFolder + "/" + fileName);
```
phillipm: `fileName` and `fileFolder` don't seem necessary

app/src/org/commcare/tasks/FormRecordToFileTask.java
```diff
@@ -0,0 +1,257 @@
...
+ *
+ * @author wspride
+ */
+public abstract class FormRecordToFileTask extends CommCareTask<String, String, Pair<Long, FormRecord[]>, CommCareWiFiDirectActivity> {
+    private static final String TAG = AndroidLogger.TYPE_FORM_DUMP;
+
+    private Context c;
+    private Long[] results;
+    // this is where the forms that have been pulled from FormRecord storage to the file system live
+    private File storedFormDirectory;
```
phillipm: `final`

app/src/org/commcare/tasks/FormRecordToFileTask.java
```diff
@@ -0,0 +1,257 @@
...
+        results = null;
+    }
+
+    private static final String[] SUPPORTED_FILE_EXTS = {".xml", ".jpg", ".3gpp", ".3gp"};
+
+    /**
+     * Turn a FormRecord folder from storage into a standard file representation in our file system.n
+     */
+    private long getFileInstanceFromStorage(File formRecordFolder, SecretKeySpec decryptionKey)
+            throws FileNotFoundException, SessionUnavailableException {
```
phillipm: `FileNotFoundException` and `SessionUnavailableException` are never thrown in this method

app/src/org/commcare/tasks/FormRecordToFileTask.java
```diff
@@ -0,0 +1,257 @@
...
+                        } catch (IOException e) {
+                            Logger.log(AndroidLogger.TYPE_ERROR_WORKFLOW, "Bizarre. Exception just getting the file reference. Not removing." + getExceptionText(e));
+                            continue;
+                        }
+
+                        //Good!
+                        //Time to transfer forms to storage!
+                        try {
+                            results[i] = getFileInstanceFromStorage(folder, new SecretKeySpec(record.getAesKey(), "AES"));
+                        } catch (FileNotFoundException e) {
```
phillipm: dead catch block, see https://github.com/dimagi/commcare-odk/pull/1114/files#r59388194

app/src/org/commcare/tasks/FormRecordToFileTask.java
```diff
@@ -0,0 +1,257 @@
+package org.commcare.tasks;
+
+import android.content.Context;
+import android.content.SharedPreferences;
+import android.os.Environment;
```
phillipm: Unused

app/src/org/commcare/tasks/FormRecordToFileTask.java
```diff
@@ -0,0 +1,257 @@
...
+
+    @Override
+    protected void onPostExecute(Pair<Long, FormRecord[]> result) {
+        super.onPostExecute(result);
+        //These will never get Zero'd otherwise
+        c = null;
+        results = null;
+    }
+
+    private static final String[] SUPPORTED_FILE_EXTS = {".xml", ".jpg", ".3gpp", ".3gp"};
```
phillipm: Move above method definitions, near other fields

app/src/org/commcare/tasks/FormRecordToFileTask.java
```diff
@@ -0,0 +1,257 @@
...
+    public static final int PULL_TASK_ID = 721356;
+
+    public FormRecordToFileTask(Context c, String formStoragePath) {
+        this.c = c;
+        this.storedFormDirectory = new File(formStoragePath);
+        taskId = PULL_TASK_ID;
+    }
+
+    @Override
+    protected void onPostExecute(Pair<Long, FormRecord[]> result) {
```
phillipm: This method is unneeded

app/src/org/commcare/tasks/FormRecordToFileTask.java
```diff
@@ -0,0 +1,257 @@
...
+ * This class transfers all of the FormRecords from form record storage and into our file system,
+ * decrypting files as necessary. This task should really be merged with DumpTask.
+ *
+ * @author wspride
+ */
+public abstract class FormRecordToFileTask extends CommCareTask<String, String, Pair<Long, FormRecord[]>, CommCareWiFiDirectActivity> {
+    private static final String TAG = AndroidLogger.TYPE_FORM_DUMP;
+
+    private Context c;
+    private Long[] results;
```
phillipm: If you remove `onPostExecute` then `results` can be local to `doTaskBackground`

app/src/org/commcare/tasks/FormRecordToFileTask.java
```diff
@@ -0,0 +1,257 @@
...
+/**
+ * This class transfers all of the FormRecords from form record storage and into our file system,
+ * decrypting files as necessary. This task should really be merged with DumpTask.
+ *
+ * @author wspride
+ */
+public abstract class FormRecordToFileTask extends CommCareTask<String, String, Pair<Long, FormRecord[]>, CommCareWiFiDirectActivity> {
+    private static final String TAG = AndroidLogger.TYPE_FORM_DUMP;
+
+    private Context c;
```
phillipm: Can be final if `onPostExecute` is removed

app/src/org/commcare/tasks/FormRecordToFileTask.java
```diff
@@ -0,0 +1,257 @@
...
+                } catch (IOException e) {
+                    e.printStackTrace();
+                }
+            }
+        }
+    }
+
+    @Override
+    protected Pair<Long, FormRecord[]> doTaskBackground(String... params) {
+        Log.d(TAG, "Doing zip task in background with params: " + params);
```
phillipm: To avoid lint warning: `param` -> `Arrays.toString(params)`

app/src/org/commcare/tasks/FormRecordToFileTask.java
```diff
@@ -0,0 +1,257 @@
...
+                    Logger.log(AndroidLogger.TYPE_ERROR_DESIGN, "Totally Unexpected Error during form submission" + getExceptionText(e));
+                }
+            }
+
+            long result = 0;
+            for (int i = 0; i < records.length; ++i) {
+                if (results[i] > result) {
+                    result = results[i];
+                }
+            }
```
phillipm: This loop should be lift into a named helper method that gives additional context to what is going on.

app/src/org/commcare/tasks/FormRecordToFileTask.java
```diff
@@ -0,0 +1,257 @@
...
+
+            return new Pair<>(result, records);
+
+        } else {
+            publishProgress("No forms to send.");
```
phillipm: Worth localizing?

app/src/org/commcare/tasks/FormRecordToFileTask.java
```diff
@@ -0,0 +1,257 @@
...
+                    this.cancel(false);
+                    return null;
+                } catch (Exception e) {
+                    //Just try to skip for now. Hopefully this doesn't wreck the model :/
+                    Logger.log(AndroidLogger.TYPE_ERROR_DESIGN, "Totally Unexpected Error during form submission" + getExceptionText(e));
+                }
+            }
+
+            long result = 0;
+            for (int i = 0; i < records.length; ++i) {
```
phillipm: Recommend using a `for-each`

app/src/org/commcare/tasks/FormRecordToFileTask.java
```diff
@@ -0,0 +1,257 @@
...
+        c = null;
+        results = null;
+    }
+
+    private static final String[] SUPPORTED_FILE_EXTS = {".xml", ".jpg", ".3gpp", ".3gp"};
+
+    /**
+     * Turn a FormRecord folder from storage into a standard file representation in our file system.n
+     */
+    private long getFileInstanceFromStorage(File formRecordFolder, SecretKeySpec decryptionKey)
```
phillipm: I would rename the `get` to `copy` to help distinguish that we aren't returning a file instance of any sort, but rather doing a copy and returning the status code of the operation

app/src/org/commcare/tasks/FormRecordToFileTask.java
```diff
@@ -0,0 +1,257 @@
...
+        super.onPostExecute(result);
+        //These will never get Zero'd otherwise
+        c = null;
+        results = null;
+    }
+
+    private static final String[] SUPPORTED_FILE_EXTS = {".xml", ".jpg", ".3gpp", ".3gp"};
+
+    /**
+     * Turn a FormRecord folder from storage into a standard file representation in our file system.n
```
phillipm: Worth mentioning what is being returned

app/src/org/commcare/tasks/FormRecordToFileTask.java
```diff
@@ -0,0 +1,257 @@
...
+
+            return new Pair<>(result, records);
+
+        } else {
+            publishProgress("No forms to send.");
+            return null;
+        }
+    }
+
+    private String getExceptionText(Exception e) {
```
phillipm: can be `static`

app/src/org/commcare/tasks/FormRecordToFileTask.java
```diff
@@ -0,0 +1,257 @@
...
+        for (File file : files) {
+            // This is not the ideal long term solution for determining whether we need decryption, but works
+            if (file.getName().endsWith(".xml")) {
+                FileUtil.copyFile(file, new File(targetDirectory, file.getName()), decryptCipher, null);
+            } else {
+                FileUtil.copyFile(file, new File(targetDirectory, file.getName()));
+            }
+        }
+    }
+    private boolean isSupportedFiletype(File file){
```
phillipm: `static`

app/src/org/commcare/tasks/FormRecordToFileTask.java
```diff
@@ -0,0 +1,257 @@
...
+        for (String ext : SUPPORTED_FILE_EXTS) {
+            if (file.getName().endsWith(ext)) {
+                return true;
+            }
+        }
+        return false;
+    }
+    private void logTransferBytes(File[] files){
```
phillipm: `static`

phillipm:
Finished another pass. Let me know when it is ready for a final pass

app/src/org/commcare/activities/CommCareWiFiDirectActivity.java
wpride: Yeah, that's always bugged me about Java

phillipm:
:metal:

PR #1116, opened 2016-03-08
Finalize product-level analytics for 2.27

amstone326:
Kevin said all of the data that's been coming into the test dashboard from 2.26 looks sound, so we're going ahead with moving to the live dashboard (but continuing to use the test dashboard for debug builds). Also makes the opt out button more hidden (per request from @jjackson).


amstone326:
@dimagi/product

krsdimagi:
Nice!

PR #1122, opened 2016-03-11
add test

orangejenny:
Tests for https://github.com/dimagi/commcare-odk/pull/1117


orangejenny:
This should be good to go. It's quick & dirty; I eventually ought to split the functions in XPathEvalTest out into utility methods.

PR #1124, opened 2016-03-11
Add project space ID to GA

amstone326:
Contains all commits from https://github.com/dimagi/commcare-odk/pull/1069 because it uses a util method that was moved in that PR, so don't merge until that one is. 

The actual changes in this PR are just the most recent commit.


amstone326:
@dimagi/product

phillipm:
Added the cross-request reference so that the tests will pass the next time you push a commit (or retrigger from jenkins)

amstone326:
@phillipm this can be merged now that #1069 was

PR #1126, opened 2016-03-11
Signature widget changes

amstone326:
-Move the 'Reset' and 'Cancel' buttons to the right side of the screen, so that the 'Save and Close' button stands alone

-Automatically navigate forward after signature capture ONLY if: a) a signature was actually captured (e.g. they pressed 'Save and Close' as opposed to cancel), b) the signature question is not on the  last screen, and c) the signature widget isn't in a question list
*(Having now done this and realized all of the conditions I needed to add in order to have this make sense, I'm open to being told that it's not a good idea to have the widget behave differently under different conditions. But wanted to share it and see what others think)*

-Remove some of the more embarrassing comments left in the `SignatureWidget` code from when I worked on it as an intern...

### DIFF
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -472,7 +472,7 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
                 boolean saved = ImageCaptureProcessing.processCaptureResponse(this, getInstanceFolder(), false);
                 if (saved) {
                     ImageButton nextButton = (ImageButton)this.findViewById(R.id.nav_btn_next);
-                    if (nextButton.getTag().equals(NAV_STATE_NEXT)) {
+                    if (nextButton.getTag().equals(NAV_STATE_NEXT) && !questionsView.isQuestionList()) {
                         next();
                     }
                 }
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -471,10 +471,8 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
             case SIGNATURE_CAPTURE:
                 boolean saved = ImageCaptureProcessing.processCaptureResponse(this, getInstanceFolder(), false);
                 if (saved) {
-                    ImageButton nextButton = (ImageButton)this.findViewById(R.id.nav_btn_next);
-                    if (nextButton.getTag().equals(NAV_STATE_NEXT) && !questionsView.isQuestionList()) {
-                        next();
-                    }
+                    // attempt to auto-advance if a signature was captured
+                    advance();
                 }
                 break;
             case IMAGE_CHOOSER:
@@ -2037,11 +2035,7 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
 
     @Override
     protected boolean onForwardSwipe() {
-        //We've already computed the "is there more coming" stuff intensely in the the nav details
-        //and set the forward button tag appropriately, so use that to determine whether we can
-        //swipe forward.
-        ImageButton nextButton = (ImageButton)this.findViewById(R.id.nav_btn_next);
-        if(nextButton.getTag().equals(NAV_STATE_NEXT)) {
+        if (canNavigateForward()) {
             GoogleAnalyticsUtils.reportFormNavForward(
                     GoogleAnalyticsFields.LABEL_SWIPE,
                     GoogleAnalyticsFields.VALUE_FORM_NOT_DONE);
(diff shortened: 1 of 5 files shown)


wpride:
We have some of this logic for the "quick" multi select appearance type which also auto advances when a selection is made. You can see that here https://github.com/dimagi/commcare-odk/blob/master/app/src/org/commcare/views/widgets/GridWidget.java#L145. Might make sense to unify this logic IE have both use the AdvanceToNextListener. It looks like the quick select widget will auto advance even if we're on a question list which seems not good. However looking at this and the "next" function it looks like we already check for end of form so your check in the control block might be redundant https://github.com/dimagi/commcare-odk/blob/master/app/src/org/commcare/activities/FormEntryActivity.java#L946

wpride:
Also I'm pro keeping funny comment artifacts with historical value intact, don't be a revisionist

phillipm:
I find it useful to see before and after screenshots when making ui changes. That was we don't need to spin up a build of this PR to understand what the changes look like. 

Screenshots are super easy to pull if you have a device open. This command stores the screenshot as `screen.png` in the current dir:
`adb shell screencap -p /sdcard/screen.png && adb pull /sdcard/screen.png && adb shell rm /sdcard/screen.png`

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -469,7 +469,13 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
                 ImageCaptureProcessing.processCaptureResponse(this, getInstanceFolder(), true);
                 break;
             case SIGNATURE_CAPTURE:
-                ImageCaptureProcessing.processCaptureResponse(this, getInstanceFolder(), false);
+                boolean saved = ImageCaptureProcessing.processCaptureResponse(this, getInstanceFolder(), false);
+                if (saved) {
+                    ImageButton nextButton = (ImageButton)this.findViewById(R.id.nav_btn_next);
+                    if (nextButton.getTag().equals(NAV_STATE_NEXT) && !questionsView.isQuestionList()) {
```
phillipm: `nextButton.getTag().equals(NAV_STATE_NEXT) ` is pretty 'low-level' code, it would be good to lift it into a named helper method and also use it in `onForwardSwipe`, which has the same line of code.

phillipm: Worth noting but nothing block: this will run into this issue http://manage.dimagi.com/default.asp?186991

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -2062,7 +2062,9 @@ public boolean onScroll(MotionEvent e1, MotionEvent e2, float distanceX, float d
 
     @Override
     public void advance() {
-        next();
+        if (!questionsView.isQuestionList() && canNavigateForward()) {
```
amstone326: @wpride This means that all usages of auto-advance would now only do so if it's not a question list, and if it's not on the last screen. I think it's easy to agree on the first condition being correct. The second is a little harder but also seems correct, since I don't think the act of answering a question should ever cause a form to submit.

amstone326: Oh also important to this- the condition you pointed out in `showNextView()` does not do what you thought it did (prevent navigating forward if it's on the last screen). I'm not sure why it doesn't, but when I remove the check for `canNavigateForward()`, the form gets submitted when an auto-advance question is on the last screen.

amstone326:
@phillipm 

Before:
screenshot_20160314-131729

After: 
screenshot_20160314-131533

app/src/org/commcare/activities/FormEntryActivity.java
wpride: Oh, hmm. That seems bad then because the auto advance grid widget definitely looks like it's counting on that not happening

amstone326: Oh ok well then I think this is definitely the right change to make

phillipm: Heads up that Ray expects auto-advance on the last question of the form to complete the form and submit it: http://manage.dimagi.com/default.asp?186847#BugEvent.1043572
Not sure if it is worth letting him know that this is changing or ask him how important it is to his workflow

amstone326: @phillipm Thanks for flagging. It's worth noting that the issue he was reporting was that the form was getting saved but not submitted, whereas this would just do nothing and wait for you submit as usual. Possible that he doesn't care all that much about the auto-submit, but was correctly reporting odd/buggy behavior. 

Either way, I'm still pretty inclined to say that answering a question should never trigger a form submission, and we should just let Ray know that this is going to change if we think he is relying on it. Do you want to communicate to him since you talk with him often, or should I just drop him a note?

phillipm: Yeah, I can follow up with him

amstone326: Cool, this should be good to go then

PR #1128, opened 2016-03-13
Add update label for Intent callout button

phillipm:
- Make it so barcode button shows "Update Barcode" once the scan has succeeded.
- Allow default barcode/intent button text to be localized. 
- Don't try to set any form data if callout was cancelled.
- Add additional error handling for intents that may have parcelled data that CommCare doesn't know how to handle; showing a toast error message to the user. Old behavior was that the app crashed.

- Allow both barcode and intent callout button text to be overriden by adding a `button-label` and `update-button-label` to the following intent question xml (no HQ support yet):
```
<odkx:intent xmlns:odkx="http://opendatakit.org/xforms" 
    id="Fingerprint" class="com.simprints.id.REGISTER"
    button-label="Scan Fingerprint"
    update-button-label="Rescan Fingerprint">
```

Note: this doesn't allow for localization support, and we should ideally be doing this through the intent question label, but that would require a vellum migration plan, since old apps don't expect this behavior. It would also require us to have an "update button" label for changing the text when intent data has been set.

Once this is merged I will add documentation about the button label setting feature to confluence. 

Show a cancel message when intent activities are cancelled:
screen

Show toast error message when the intent result can't be processed:
screen

screen


app/src/org/commcare/engine/extensions/IntentExtensionParser.java
```diff
@@ -42,7 +42,8 @@ public void handle(XFormParser p, Element e, Object parent) {
 
         Log.d(TAG, "0123 extention parser appearance is: " + appearance);
 
-        String label = e.getAttributeValue(null, "button-label");
+        String getButtonLabel = e.getAttributeValue(null, "button-label");
```
amstone326: variable name here could be clearer, not sure what the "get" part is meant to mean

amstone326: Ah I see, it's like "get" something from the intent. Makes more sense, I think it just comes off a little awkwardly because of the conflict with the standard meaning for `get`. I can't really think of a better option though

amstone326:
@phillipm Merged, adding documentation on the label stuff to confluence would be great.

PR #1129, opened 2016-03-14
Fix Nepali date widget from changing dates after navigating to/from widget

phillipm:
Currently, the Nepali date widget changes the date back one day every time you scroll to and from the widget. This is due to some sort of timezone/calendar day issue where if you are in Nepal and it is, say 10am on the 12th, and hence the 11th in Greenwich Mean Time, then restoring the stored date value restores to the 11th, not the 12th.

:calendar: :calendar:  :earth_asia: :earth_africa: :boom: :fire: :fire: :boom:


ctsims:
@phillipm Happy to pull. Looks like ther'es a failing check? Can we retrigger?

phillipm:
@ctsims This PR is currently just has the failing test case that exemplifies the bug that needs to be fixed.

PR #1130, opened 2016-03-14
Refactor: EntityView

phillipm:
Needed for a feature to attach callout data to case entries in the case list.

- Remove 'Text To Speech' logic and layouts since we the `tts` field is always `null`. I know @ctsims said we might add it back at some point, but I would like to remove it until a client actually contracts us to build it, otherwise it is dead code that isn't being maintained and adding complexity.
- Unify setup code for header and entry rows
- Refactor `ViewId`

Probably easiest to review commit-by-commit


app/src/org/commcare/views/EntityView.java
```diff
@@ -150,7 +144,7 @@ private View initView(Object data, String form, ViewId uniqueId, String sortFiel
         } else if (FORM_AUDIO.equals(form)) {
             String text = (String) data;
             AudioButton b;
-            if (text != null & text.length() > 0) {
+            if (text != null && text.length() > 0) {
```
amstone326: oof

app/src/org/commcare/views/EntityView.java
```diff
@@ -167,7 +167,7 @@ public void setSearchTerms(String[] terms) {
     }
 
     public void refreshViewsForNewEntity(Entity e, boolean currentlySelected, long rowId) {
-        for (int i = 0; i < e.getNumFields(); ++i) {
+        for (int i = 0; i < views.length; ++i) {
```
amstone326: Hey, a little confused what this change represents. Seems like if `e.getNumFields()` was incorrect, it should have been causing buggy behavior. What exactly is the distinction between the 2 expressions?

phillipm: They can be used interchangeably because `views` is setup like this:
```
this.views = new View[e.getNumFields()];
```
I'm just trying to make the internal representation have as few external dependencies as possible.

PR #1131, opened 2016-03-14
Restore last entered username

amstone326:
Problem being addressed: Currently whenever the `LoginActivity` resumes, we load the username entry box with the username of the last logged in user, if there is one. This can produce confusing behavior in a couple of scenarios: 
1) If you attempt to log in with a new username, and that login fails, and then the screen shuts off, when you turn it back on, the username box is populated with the last logged in username instead of the username you just tried.
2) Same as above, but without the login failure part - if you just go away in the middle of entering a username for any reason, and then come back, the box will have been re-populated with something other than what you left there.

The solution I have here: 
-In `LoginActivity.onPause()`, either store the currently entered username to prefs if it differs from the last logged in username, or clear it if it is empty.
-In `restoreLastUsername()`, check for the last entered username BEFORE checking for the last logged in username. By also clearing the last entered username upon the start of a session, this ensures that the last logged in username will take precedence ONLY UNTIL a different username has been tried.

### DIFF
--- a/app/src/org/commcare/activities/LoginActivityUIController.java
+++ b/app/src/org/commcare/activities/LoginActivityUIController.java
@@ -211,14 +211,9 @@ public class LoginActivityUIController implements CommCareActivityUIController {
         // Remove any error content from trying to log into a different app
         setStyleDefault();
 
-        final SharedPreferences prefs = CommCareApplication._().getCurrentApp().getAppPreferences();
-        String lastUser = prefs.getString(CommCarePreferences.LAST_LOGGED_IN_USER, null);
-        if (lastUser != null) {
-            // If there was a last user for this app, show it
-            username.setText(lastUser);
-            passwordOrPin.requestFocus();
-        } else {
-            // Otherwise, clear the username text so it does not show a username from a different app
+        if (!restoreLastUser()) {
+            // If we didn't have a username to restore for this app, clear the username text so it
+            // does not show a username from a different app
             username.setText("");
             username.requestFocus();
         }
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/activities/LoginActivityUIController.java
```diff
@@ -441,13 +436,31 @@ private void setupLoginButton() {
...
+     */
+    protected boolean restoreLastUsername() {
         SharedPreferences prefs = CommCareApplication._().getCurrentApp().getAppPreferences();
-        String lastUser = prefs.getString(CommCarePreferences.LAST_LOGGED_IN_USER, null);
-        if (lastUser != null) {
-            username.setText(lastUser);
+
+        // First try to restore the last username that was entered, but NOT successfully logged in
+        // with (this value gets cleared upon a successful login so that that can take precedence)
+        String lastEnteredUsername = prefs.getString(LoginActivity.KEY_LAST_ENTERED_USERNAME, null);
```
amstone326: Note that the only reason I'm retaining the more complicated workflow of checking for BOTH the last entered username and then the last logged in username (as opposed to just allowing the last entered username to serve both purposes) is for the differential behavior of setting the focus placement (at the end of the username entry box for the last-entered username, under the assumption that they may still be in the middle of typing it, vs. in the password entry box for last logged in username). If this doesn't seem like a good enough reason for the extra complexity, I could remove it, but I do think it's a nicer UX this way

phillipm:
If `uiController.restoreLastUser();` is the issue and is being called in `onCreate` then can't you store the currently entered username into the bundle that is passed into `onCreate` and make that take precedence?

amstone326:
Ugh ok this is actually just an embarrassingly over-complicated workaround for a regression I didn't realize I had caused with multiple apps. Submitting the correct PR to fix it in a sec.

PR #1132, opened 2016-03-14
Check for empty password on login

amstone326:
If a login is attempted with an empty password, immediately fail and show a specified error message, instead of going through a whole login attempt.


phillipm:
:100:
So excited that you did this! Gets me at least once a day

PR #1133, opened 2016-03-14
End of Form Button

ctsims:
Updates end of form navigation such that on the last page of a form the progress bar grows into a large "Continue" button rather than forcing the user to click on the smaller "next" button to improve visibility.

New Behavior (Sorry for super low framerate, GIF + Vysor isn't amazing for this)
end_of_form_nav


ctsims:
@amsagoff @biyeun Went with the inset button for now, to retain consistent padding

biyeun:
DOPE. :fire::fire::fire::fire::fire::fire::fire::fire:

phillipm:
Yeah, this is sweet! Curious to see what people think!

What if it said 'Complete' instead of 'Continue'?
Anyone have ideas of how to address the fact that the spacing between the green bar and its container is really small?

amsagoff:
What will happen if the last question is not required? Will the green
animate when you get to that screen? What will that look like?

amstone326:
@amsagoff I'm assuming it will work the same way that the turning-green of the arrow does currently, which is that the button turns green as soon as the form is valid to be submitted (so yes, it would animate as soon as you get to the last screen if the last question is not required)

app/src/org/commcare/activities/components/FormNavigationUI.java
```diff
@@ -122,7 +178,13 @@ public void onAnimationRepeat(Animation animation) {
 
             }
         });
-        nextButton.startAnimation(growShrinkAnimation);
+        finishButton.startAnimation(growShrinkAnimation);
+    }
+
+    public static void setFinishVisible(ClippingFrame finishButton) {
```
phillipm: This should be `private`

app/src/org/commcare/activities/components/FormNavigationUI.java
```diff
@@ -77,30 +85,78 @@ public static void updateNavigationCues(CommCareActivity activity,
...
 
         progressBar.setProgressDrawable(context.getResources().getDrawable(R.drawable.progressbar_full));
-        progressBar.setProgress(totalQuestions);
+        progressBar.setProgress(details.totalQuestions);
 
         Log.i("Questions", "Form complete");
     }
 
+    @TargetApi(Build.VERSION_CODES.HONEYCOMB)
+    private static void inflateFinishButton(Context context,
```
phillipm: "inflate" was a confusing word for me, because the view isn't being inflated in the android sense, but rather the view showing an inflation animation. Maybe a name like "animateFinishButtonInflation"?

app/src/org/commcare/views/ClippingFrame.java
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * A clipping frame is a simple frame layout that allows the specification (by attribute or
+ * android property) of an area which should be clipped. This allows the view to only display a
+ * portion of its content at a time, which can be used to animate "reveal" effects.
+ *
+ * Created by ctsims on 3/14/2016.
+ */
+public class ClippingFrame extends FrameLayout {
+
+    private Rect mClipBounds= new Rect(0,0,1,1);
```
phillipm: `final`

app/src/org/commcare/views/ClippingFrame.java
```diff
@@ -0,0 +1,119 @@
+package org.commcare.views;
+
+import android.content.Context;
+import android.content.res.TypedArray;
+import android.graphics.Canvas;
+import android.graphics.Rect;
+import android.util.AttributeSet;
+import android.widget.FrameLayout;
+import android.widget.RelativeLayout;
```
phillipm: unused

app/src/org/commcare/views/ClippingFrame.java
```diff
@@ -0,0 +1,119 @@
...
+            clipWidth = typedArray.getFloat(R.styleable.ClippingView_clip_width, 1);
+            clipHeight = typedArray.getFloat(R.styleable.ClippingView_clip_height, 1);
+        }
+    }
+
+    /**
+     * A property setter to be used by animations. Note that the name is sensitive and should
+     * not be changed.
+     */
+    public void setClipWidth(float clipWidth) {
```
phillipm: haha, really?! Android :crystal_ball:

phillipm:
:+1: :+1: 
Left some naming-related comments that would be nice to fix; otherwise good to go.

ctsims:
@phillipm updated

phillipm:
Thanks! Gonna go out with a CHW tomorrow and I'll put this on their device :)

amstone326:
@amsagoff Random thought, but to the extent that FMs are not already very in the habit of doing this, we might want to emphasize with the rollout of this that the text on the button is customizable. (I just came across a place in the an app that made me realize there are ways in which people sometimes use forms that make “Finish” seem out of context)

amstone326:
Oo also @phillipm how did it go when you tried this out with a CHW?

amsagoff:
Hey,

Is this still going out in 2.27, without being field tested?

phillipm:
It is in 2.27. I didn't see anyone getting confused with it while in Nepal, but the CHWs were trained to use the arrow instead of swiping anyways. I don't think it has been tested on anyone else...

amsagoff:
This is a red flag to me. If we change up the end of form experience I want
to make sure that we get it right to avoid continued churn and confusion.
We also will need to come up with a communication strategy. Does this need
to be in 2.27 or is it relatively straightforward to push to 2.28?

phillipm:
Agreed, I'll disable it by default and hide it behind a feature flag on 2.27 so FMs can test it in the field

phillipm:
https://github.com/dimagi/commcare-odk/pull/1208

PR #1134, opened 2016-03-14
Broad Catch for Resource Scan Crash

ctsims:
Fixes the crash when the app is loading resources in the middle of receiving an update.


phillipm:
Follow up to https://github.com/dimagi/commcare-odk/pull/1081

PR #1137, opened 2016-03-16
Grid Icons resolution fix

wpride:
Fix for http://manage.dimagi.com/default.asp?221195

Use a bigger icon when using grid menus

Adding don't pull for now for question: is it SOP to add the icon to every folder? Some subset of them? If so, what are the go-to sizes? (If we even have a policy)


ctsims:
@wpride One of the two icons should get added (form v. folder) depending on the type if there is no specified icon.

I actually am not sure what the default size "Should" be in this case, since it's so screen size specific. I think roughly 1/3 of the largest supported resolution would be the default box, with probably 30% margins?

wpride:
@ctsims yeah got the form v. folder part, was actually asking about ldpi v mdpi v ldpi v xhdpi - do we usually add an icon to each of these?

ctsims:
Ah, shouldn't need to with Vector Graphics

wpride:
@ctsims thanks, this is gtg then unless there's other feedback

PR #1138, opened 2016-03-16
Fix overriding of the last-entered username

amstone326:
Finally realized that this was a regression caused by multiple apps. For no clear reason, I was calling `refreshForNewApp()` (which calls `restoreLastUsername()`) in `onResume()`, instead of just upon successful completion of a call to `SeatAppActivity`, which was causing the last-entered username to be overridden by the last logged in username whenever `onResume()` was called. Sorry for not catching this sooner!

### DIFF
--- a/app/src/org/commcare/activities/LoginActivity.java
+++ b/app/src/org/commcare/activities/LoginActivity.java
@@ -283,6 +283,13 @@ public class LoginActivity extends CommCareActivity<LoginActivity>
         tryAutoLogin();
     }
 
+    @Override
+    protected void onActivityResult(int requestCode, int resultCode, Intent intent) {
+        if (requestCode == SEAT_APP_ACTIVITY) {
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -283,6 +283,13 @@ protected void onResumeFragments() {
         tryAutoLogin();
     }
 
+    @Override
+    protected void onActivityResult(int requestCode, int resultCode, Intent intent) {
+        if (requestCode == SEAT_APP_ACTIVITY && resultCode == RESULT_OK) {
```
phillipm: should call `super`: http://stackoverflow.com/a/11745655

amstone326: Ugh good catch thanks

PR #1141, opened 2016-03-18
Code to get the latest dev build

wpride:
FYI/saving in case we move forward with this somehow


app/src/org/commcare/activities/UpdateActivity.java
```diff
@@ -205,7 +205,8 @@ protected void startUpdateCheck() {
         }
 
         String ref = ResourceInstallUtils.getDefaultProfileRef();
-        updateTask.execute(ref);
+        String newestRef = ResourceInstallUtils.getLatestProfileRef();
```
amstone326: I'm guessing we would want this in an alternative update path, as opposed to replacing the default update behavior

wpride: Yeah this is just an FYI not mergable

amstone326:
:(

PR #1142, opened 2016-03-18
Fix NPE on multiple apps Multimedia Validation

phillipm:
If you:
- install an app w/o multimedia via multiple apps
- try to validate its multimedia
- select 'skip' when it fails
- switch to the normal CommCare app

Then you get a crash due to the intent passed to `DispatchActivity.onActivityResult` being `null`.

Also make sure to call `super.onActivityResult` in a few methods when no the result code isn't processed in the current `onActivityResult` implementation. This isn't necessary, but is the 'right thing' to do.

### DIFF
--- a/app/src/org/commcare/activities/CommCareVerificationActivity.java
+++ b/app/src/org/commcare/activities/CommCareVerificationActivity.java
@@ -225,6 +225,7 @@ public class CommCareVerificationActivity
             // we found some media, so try validating it
             newMediaToValidate = true;
         }
+        super.onActivityResult(requestCode, resultCode, intent);
     }
 
     private void handleVerificationSuccess() {
@@ -290,8 +291,8 @@ public class CommCareVerificationActivity
                 break;
             case R.id.screen_multimedia_retry:
                 verifyResourceInstall();
+                break;
         }
-
     }
 
     private String prettyString(String rawString) {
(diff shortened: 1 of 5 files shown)


app/src/org/commcare/activities/CommCareVerificationActivity.java
```diff
@@ -224,7 +224,10 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
         if (requestCode == GET_MULTIMEDIA && resultCode == Activity.RESULT_OK) {
             // we found some media, so try validating it
             newMediaToValidate = true;
+            return;
         }
+
+        super.onActivityResult(requestCode, resultCode, intent);
```
amstone326: Is super specifically _only_ supposed to get called when the result code isn't processed in this implementation?

phillipm: From what I can tell, it is one of those things that it actually doesn't matter and is only worth putting in for the sake of extensibiltiy. For instance `Activity.onActivityResult` is an empty implementation.
I think `FragmentActivity.onActivityResult` dispatches to associated fragments, so in that sense it might be more important... 
I think if something is processed via the overriding version, it probably doesn't also need to be passed up to super classes for processing.

PR #1147, opened 2016-03-22
Fix last username spill-over to newly installed app

amstone326:
Fix for http://manage.dimagi.com/default.asp?221630#1113023, which was introduced in https://github.com/dimagi/commcare-odk/pull/1138. 

Turns out there was one reason for calling `refreshForNewApp()` in `onResume()` that I had forgotten about: if you install an app from the app manager and then return to the login screen, the seated app has changed. This PR introduces a check for that case.


phillipm:
This targets `master` but the milestone says 2.27

amstone326:
Thanks, creating a new one

PR #1155, opened 2016-03-23
Avoid image scaling div by zero

phillipm:
Avoid a division by zero in image scaling code. Not sure if blindly avoiding the division is the correct response to this; for instance, getting a zero width value may signify a deeper issue. I didn't take the time to try to reproduce this.

Potential fix for http://manage.dimagi.com/default.asp?221782

### DIFF
--- a/app/src/org/commcare/utils/CachingAsyncImageLoader.java
+++ b/app/src/org/commcare/utils/CachingAsyncImageLoader.java
@@ -18,7 +18,7 @@ import android.widget.ImageView;
  */
 @SuppressLint("NewApi")
 public class CachingAsyncImageLoader implements ComponentCallbacks2 {
-    private final TCLruCache cache;
+    private final LruCache<String, Bitmap> cache;
     private static final int CACHE_DIVISOR = 2;
     private final Context context;
 
@@ -27,7 +27,7 @@ public class CachingAsyncImageLoader implements ComponentCallbacks2 {
                 Context.ACTIVITY_SERVICE);
         int memoryClass = (am.getMemoryClass() * 1024 * 1024) / CACHE_DIVISOR;        //basically, set the heap to be everything we can get
         this.context = context;
-        this.cache = new TCLruCache(memoryClass);
+        this.cache = new LruCache<>(memoryClass);
     }
 
     public void display(String url, ImageView imageView, int defaultResource,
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/utils/MediaUtil.java
```diff
@@ -257,7 +267,7 @@ private static Bitmap getBitmapScaledByTargetOrContainer(String imageFilepath,
         double widthScaleDownFactor =  (double)boundingWidth / originalWidth;
         // Choosing the larger of the scale down factors, so that the image still fills the entire
         // container
-        double dominantScaleDownFactor = Math.max(heightScaleDownFactor, heightScaleDownFactor);
+        double dominantScaleDownFactor = Math.max(widthScaleDownFactor, heightScaleDownFactor);
```
phillipm: @amstone326 `widthScaleDownFactor` wasn't being used, so I assume that you meant to use it here. Can you please verify that this is correct. If it isn't, please remove `widthScaleDownFactor`

app/src/org/commcare/utils/CachingAsyncImageLoader.java
```diff
@@ -110,6 +101,8 @@ public void onLowMemory() {
 
     @Override
     public void onTrimMemory(int level) {
+        // TODO PLM: this is probably never called because Context.registerComponentCallbacks
+        // is never used to register instances of CachingAsyncImageLoader
```
phillipm: @wpride Heads up regarding this code. This is sorta tricky because I'm unsure when the right time is to call  `Context.unregisterComponentCallbacks`. I could also be wrong here.

app/src/org/commcare/utils/CachingAsyncImageLoader.java
```diff
@@ -44,14 +44,6 @@ public void display(String url, ImageView imageView, int defaultResource,
...
     }
 
-    @SuppressLint("NewApi")
-    private class TCLruCache extends LruCache<String, Bitmap> {
-
-        public TCLruCache(int maxSize) {
-            super(maxSize);
-        }
-    }
-
```
phillipm: Pretty sure there is no need for this class. Making this nested class non-static means it can potentially leak memory (I don't think it does in this case, but worth removing so we don't have to worry about it)

app/src/org/commcare/utils/MediaUtil.java
amstone326: Ugh yes thank you for catching. This was the part of this code I refactored more recently, so it at least hasn't been around for long.

amstone326:
Hey, this looks ok to me, although I couldn't replicate, so I also don't know what the root cause of that `newWidth` dimension being 0 is. In reviewing this I did notice a small logical error in one of my methods (which I added a commit to fix), which could have potentially been causing buggy behavior in some edge-case scenarios, although I couldn't think of how it would be causing this particular scenario. I think either way this is a fine change to merge, and then VL can test out if this fixes the issue.

amstone326:
(Let me know if you agree that merging this is fine and I'll go ahead).

phillipm:
@amstone326 Looks good to me

PR #1156, opened 2016-03-23
Add OK button to 'About CommCare' dialog

phillipm:
Used a phone with a broken 'back' button the other day and got stuck on the 'About CommCare' button because it didn't have an exit button.
 
screen


orangejenny:
hurrah for improved user experience

PR #1158, opened 2016-03-23
Don't refresh case list if distance hasn't changed much

phillipm:
For the distance from here in the case list feature we don't need to refresh the case list on every location update if the location hasn't changed by very much. In this case I'm setting it to 30 meters.

Also lift 'location provider missing' dialog logic into own file.


app/src/org/commcare/utils/HereFunctionHandler.java
```diff
@@ -85,20 +90,33 @@ public void refreshLocation() {
...
+        Log.i("HereFunctionHandler", "location has been set to " + this.location);
 
         if (this.location.getAccuracy() <= GeoUtils.ACCEPTABLE_ACCURACY) {
             locationGoodEnough = true;
             forbidGpsUse();
         }
 
-        if (entitySelectActivity != null) {
+
+        if (shouldRefreshEntityList()) {
```
amstone326: I would argue that it makes more sense to write this as `if (entitySelectActivity != null && shouldRefreshEntityList())`, since the null-check isn't part of the logic implied by the method name; feels more like just its own separate concern/requirement. It also allows for short-circuiting in the event that `entitySelectActivity` is null.

phillipm: :ok_hand:

PR #1159, opened 2016-03-23
Notify user of self-indexing case in restore payload

phillipm:
Tell the user when their restore failed due to self-indexing case.

Error message is the following and is localizable:
screen

Following the pattern laid out with error handling in the save to disk task: https://github.com/dimagi/commcare-odk/pull/1088
The idea is to have the task return a result code and a string that has any error message details that will be passed to the user.

Addressing http://manage.dimagi.com/default.asp?188895

cross-request: https://github.com/dimagi/commcare/pull/275
cross-request: https://github.com/dimagi/javarosa/pull/278

Follow-up: update https://confluence.dimagi.com/display/ccinternal/Mobile+Error+Messages?src=search


app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -187,7 +188,7 @@ public void reportProgress(int progress) {
                         if (ukr == null) {
                             Logger.log(AndroidLogger.TYPE_ERROR_ASSERTION, "Shouldn't be able to not have a valid key record when OTA restoring with a key server");
                             this.publishProgress(PROGRESS_DONE);
-                            return PullTaskResult.UNKNOWN_FAILURE;
+                            return new Pair<>(PullTaskResult.UNKNOWN_FAILURE, "Unable to generate encryption key");
```
phillipm: Not going to localize this error message or the other 2 above because I don't think they really ever happen in practice.

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -217,7 +219,7 @@ public void reportProgress(int progress) {
                         CommCareApplication._().releaseUserResourcesAndServices();
                     }
                     Logger.log(AndroidLogger.TYPE_USER, "Bad Auth Request for user!|" + username);
-                    return PullTaskResult.AUTH_FAILED;
+                    return new Pair<>(PullTaskResult.AUTH_FAILED, "");
```
wpride: Possibly too-cumbersome suggestion: instead of using a Pair with sometimes empty string, would it make sense to turn PullTaskResult from an enum into a full-fledged object with a String field and boolean hasMessage field?

wpride:
Very cool, looks good to go to me pending response to my own design question

app/src/org/commcare/tasks/DataPullTask.java
phillipm: Good call! How about I do this in new PR and also update `SaveToDiskTask` to use the same object ([here](https://github.com/dimagi/commcare-odk/blob/master/app/src/org/commcare/tasks/SaveToDiskTask.java#L93-L95))

wpride: Sounds good!

PR #1160, opened 2016-03-23
Revert to allowing auto-advance selects to trigger form submission

amstone326:
In this PR https://github.com/dimagi/commcare-odk/pull/1126, I changed the behavior of auto-advance widgets to NOT perform an auto-advance if it is the last question in the form, so that auto-advancing would no longer trigger an automatic form submission. My thinking being that nothing should trigger a form submission other than a user explicitly pressing the submit button. However, @phillipm brought up the issue that some existing users (namely Ray it sounds like) are currently relying on this functionality in their apps, and would be upset to discover it has changed. 

On principle, I think my reasoning is technically correct, but given the reality that app builders are very unlikely to use the auto-advance setting on a question unless they know what they're doing and are going to explicitly train their users accordingly, it also seems fine to leave the old functionality. Just wanted to check if anyone else has strong opinions on what the right thing to do here is, before I revert the change (cc: @ctsims @wpride @amsagoff).

### DIFF
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -481,7 +481,7 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
                 boolean saved = ImageCaptureProcessing.processCaptureResponse(this, getInstanceFolder(), false);
                 if (saved) {
                     // attempt to auto-advance if a signature was captured
-                    advance();
+                    advance(false);
                 }
                 break;
             case IMAGE_CHOOSER:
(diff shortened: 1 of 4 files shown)


wpride:
I thought we decided that we explicitly wanted people to click the green check mark to submit forms in the new UI. Allowing workarounds seems against that

phillipm:
I think we should get the opinion of more people; seems like a change that 3 devs shouldn't be making without consulting the product team or even users.

amstone326:
That's fine with me, although we probably need to make a short-term decision sooner than that, for 2.27. Left as it is now, the auto-advance to submit a form will be disabled. If we want to leave the old behavior, this PR should get merged for now.

amsagoff:
I agree with @amstone326 's reasoning here. Curious to hear more about how Ray/others are relying on this, and then we can make a call on whether we want to productize that workflow or just support it for legacy projects.

phillipm:
Reached out to Ray. I'll report back with his response.

amstone326:
@phillipm Did you hear back from Ray on this?

phillipm:
Ray is all set. Closing this out, thanks for reconsidering this.

PR #1161, opened 2016-03-24
Feature flag for disabling swiping out of case detail tabs

phillipm:
Feature flag for preventing swipe actions in the case detail tab view from exiting the case detail view.

Can be enabled at the app level or in the devices developer options

Setting it to target 2.27 because it is a small, contained change that I think some FMs would appreciate. 

End goal is to do some sort of UX testing to determine whether this should be the default. 

screen

Follow-up: add documentation to https://confluence.dimagi.com/display/ccinternal/CommCare+Android+Developer+Options

### DIFF
--- a/app/res/xml/preferences_developer.xml
+++ b/app/res/xml/preferences_developer.xml
@@ -96,4 +96,11 @@
         android:defaultValue="yes"
         android:key="cc-form-payload-status"
         android:title="Load form payload status"/>
+    <ListPreference
+        android:enabled="true"
+        android:entries="@array/pref_enabled_labels"
+        android:entryValues="@array/pref_enabled_vals"
+        android:defaultValue="yes"
+        android:key="cc-detail-final-swipe-enabled"
+        android:title="Enable swipe in/out of case detail tabs"/>
 </PreferenceScreen>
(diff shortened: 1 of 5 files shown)


phillipm:
cc @orangejenny

orangejenny:
Hey, code looks fine, but I don't follow the use case for toggling this for individual projects. Seems it'd just get turned on/off based on whether or not a given FM likes swiping, which isn't a great way to make decisions about nav design.

But I do support gathering data to help decide whether or not the swiping is a good idea. We've gone back & forth on this feature a few times, but no one's done usability testing on it. Do you have concrete plans for doing testing while you're in the field?

phillipm:
Dev options can be enabled at the app level by setting a custom property similar to the last entry on [this page](https://confluence.dimagi.com/display/ccinternal/CommCare+Android+Developer+Options#CommCareAndroidDeveloperOptions-Firetriggersonformsave). I definitely need to update the docs to explain this.

No plans on how to test this. Rushika was asking for it on a chat group and I think we talked about changing this so I figured I'd PR it. I would be happy to reach out to a few projects that I know use case detail tabs and see if they are interested in this and then talk to the product & UAT teams about doing more testing.

phillipm:
Added analytics to track when swiping in and out of tabbed/non-tabbed case detail screen.

orangejenny:
Appreciate that you're tracking whether or not tabs are enabled; the combination of tabs & the swipe nav has always been kind of awkward, and I don't think we've ever gotten feedback on that specific combination.

amstone326:
Nice, fun to see someone else making use of the analytics code :)

PR #1162, opened 2016-03-24
Incomplete Forms Crash Fix

ctsims:
Fix for incomplete form display when traversing views

ticket:
http://manage.dimagi.com/default.asp?221833

### DIFF
--- a/app/src/org/commcare/adapters/IncompleteFormListAdapter.java
+++ b/app/src/org/commcare/adapters/IncompleteFormListAdapter.java
@@ -94,7 +94,7 @@ public class IncompleteFormListAdapter extends BaseAdapter implements FormRecord
         for (Suite s : platform.getInstalledSuites()) {
             for (Enumeration en = s.getEntries().elements(); en.hasMoreElements(); ) {
                 Entry entry = (Entry) en.nextElement();
-                if (entry instanceof FormEntry) {
+                if (!entry.isView()) {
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/adapters/IncompleteFormListAdapter.java
```diff
@@ -94,10 +94,14 @@ public IncompleteFormListAdapter(Context context,
         for (Suite s : platform.getInstalledSuites()) {
             for (Enumeration en = s.getEntries().elements(); en.hasMoreElements(); ) {
                 Entry entry = (Entry) en.nextElement();
-                if (!entry.isView()) {
-                    // Ensure that entry is actually <entry> and not a <view>,
-                    // which can't define a form
-                    names.put(((FormEntry)entry).getXFormNamespace(), entry.getText());
+                if (entry instanceof FormEntry) {
```
phillipm: This line should still be `if (!entry.isView()) {`

ctsims: Can you grab the ticket if that is true? That is crashing.

phillipm:
I've pushed a commit with my suggested change grabbed the ticket.

With this change I've tested using the reproduction steps and didn't see a crash. Feel free to post reproduction steps to the crash you are seeing on the ticket and I will address it, otherwise I'm fine letting QA retest after this is merged.

PR #1165, opened 2016-03-27
Support for Refresh To Latest Build utility

amstone326:
Code for supporting an additional utility to the Developer Toolkit that will enable one-click triggering of the following cycle (which proceeds automatically from start to finish):
-Save the current session, including any form entry progress
-Attempt to automatically update to the latest build 
-If the update is successful, log out and log back in as the last user, and then restore the saved session

My hope is that this willl significantly speedup the workflow for app-builders who are testing incremental changes to a form. (And if Cory/Will/Simon make the change they were talking about to enable updating to the latest version of an app without making a new build, the combination seems like it would be really powerful).

I still have to do the developer toolkit side of this now, but that will be really short and then I'll PR that too. This can definitely be reviewed/merged in the meantime. 

**To test:**
-Enable session saving in developer preferences (for now this won't work without that on, and gives a warning to turn it on first if you try to use it without doing so. I could obviously set it up such that this gets turned on automatically when someone triggers this action -- not sure if that is the correct thing to do or not).
-run this from the command line, with the device plugged in:
`adb shell am broadcast -a org.commcare.dalvik.api.action.RefreshToLatestBuildAction`

UX feedback is welcome; this feels like a delicate thing to get right.


app/src/org/commcare/activities/UpdateUIController.java
```diff
@@ -93,7 +93,7 @@ public void onClick(View v) {
         installUpdateButton.setOnClickListener(new View.OnClickListener() {
             @Override
             public void onClick(View v) {
-                activity.lauchUpdateInstallTask();
```
phillipm: lol! You always catch these!

app/src/org/commcare/activities/UpdateActivity.java
```diff
@@ -39,6 +40,8 @@
     private UpdateTask updateTask;
     private UpdateUIController uiController;
 
+    private boolean proceedAutomatically;
```
phillipm: Does this need to be stored in the instance bundle in case you rotate the screen? Or is it somehow magically stored in the intent?

app/src/org/commcare/activities/UpdateUIController.java
phillipm: lol! You always catch these!

app/src/org/commcare/activities/RefreshToLatestBuildActivity.java
```diff
@@ -0,0 +1,122 @@
...
+import org.commcare.dalvik.R;
+import org.commcare.preferences.DevSessionRestorer;
+import org.commcare.preferences.DeveloperPreferences;
+import org.commcare.utils.SessionUnavailableException;
+import org.commcare.views.dialogs.AlertDialogFactory;
+import org.javarosa.core.services.locale.Localization;
+
+/**
+ * Activity that is launched immediately upon reception of a RefreshToLatestBuildAction broadcast,
+ * triggering the necessary action sequence.
```
phillipm: This comment talks about how this class is used in our codebase, but doesn't describe what it does. I would find it helpful to include what this activity is doing / showing to the user.

I've read arguments that adding comments describing how code is used is harmful because as the codebase changes most developers neglect to keep these comments update-to-date. I tend to agree with this idea.

app/src/org/commcare/activities/UpdateActivity.java
amstone326: Haha, and you always catch these. Definitely needs to be stored.

amstone326:
Here's the PR for the toolkit app side of this: https://github.com/dimagi/commcare-developer-toolkit/pull/1

app/src/org/commcare/activities/RefreshToLatestBuildActivity.java
```diff
@@ -0,0 +1,126 @@
...
+import org.commcare.views.dialogs.AlertDialogFactory;
+import org.javarosa.core.services.locale.Localization;
+
+/**
+ * Triggers the following action sequence upon launching, which proceeds automatically from
+ * start to finish:
+ * -Save the current session, including any form entry progress
+ * -Attempt to automatically update to the latest build
+ * -If the update is successful, log out and log back in as the last user, and then restore the
+ * saved session
```
phillipm: :+1:

app/src/org/commcare/CommCareApplication.java
```diff
@@ -190,6 +190,7 @@
     private String messageForUserOnDispatch;
     private String titleForUserMessage;
 
+    private boolean latestBuildRefreshPending;
```
phillipm: Can you add a comment for this flag?

app/src/org/commcare/activities/UpdateActivity.java
```diff
@@ -49,6 +54,12 @@ protected void onCreate(Bundle savedInstanceState) {
 
         boolean isRotation = savedInstanceState != null;
         setupUpdateTask(isRotation);
+
+        proceedAutomatically = getIntent().getBooleanExtra(
+                RefreshToLatestBuildActivity.KEY_FROM_LATEST_BUILD_ACTIVITY, false);
+        if (proceedAutomatically) {
+            startUpdateCheck();
+        }
```
phillipm: You shouldn't need to call `startUpdateCheck()` here because it should have already been called in `setupUpdateTask()`

app/src/org/commcare/activities/UpdateActivity.java
```diff
@@ -170,19 +184,37 @@ public void handleTaskUpdate(Integer... vals) {
...
             // Gives user generic failure warning; even if update staging
             // failed for a specific reason like xml syntax
             uiController.checkFailedUiState();
+            if (proceedAutomatically) {
+                finishWithResult(RefreshToLatestBuildActivity.UPDATE_ERROR);
+                return;
+            }
         }
 
         unregisterTask();
```
phillipm: Probably good idea to call `unregisterTask()` even when `proceedAutomatically` is set to `true`

phillipm:
This is awesome! Good to go once my latest comments are addressed.

amstone326:
@phillipm Updated with changes. I also realized 2 small mistakes in the process, which I fixed:
-It actually isn't necessary to load the `proceedAutomatically` flag from `savedInstanceState` because it will always be accessible from `getIntent()`
-I wasn't wiping the `pendingBuildRefresh` flag when an error occurred

PR #1167, opened 2016-03-28
Fix for HQ not being able to distinguish demo data

ctsims:
HQ couldn't previously identify what data in the system was coming from demo users, simply that some submissions were forced to be sent without authentication, preventing it from handling them properly.

Also unifies redundant logic around not using authentication when demo users are logged in.

### DIFF
--- a/app/src/org/commcare/network/HttpRequestGenerator.java
+++ b/app/src/org/commcare/network/HttpRequestGenerator.java
@@ -68,15 +68,25 @@ public class HttpRequestGenerator {
      */
     public static final String AUTH_REQUEST_TYPE_NO_AUTH = "noauth";
 
+    private static final String SUBMIT_MODE = "submit_mode";
+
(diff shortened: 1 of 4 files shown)


app/src/org/commcare/network/HttpRequestGenerator.java
```diff
@@ -68,15 +68,25 @@
...
     public static final String AUTH_REQUEST_TYPE_NO_AUTH = "noauth";
 
+    private static final String SUBMIT_MODE = "submit_mode";
+
+    private static final String SUBMIT_MODE_DEMO = "demo";
+
     private Credentials credentials;
     PasswordAuthentication passwordAuthentication;
     private String username;
+    private String userType;
```
phillipm: Would argue that it would be cleaner to replace this with `private final boolean isDemoUserRequest`.
(requires you to set the field in the empty constructor, which I think is fine)

PR #1170, opened 2016-03-29
Fix update icons based on UAT feedback

phillipm:
One last piece of UAT feedback for the update screen that I forgot to implement:
People get confused by the green checkmark to apply a downloaded updated, thinking that the check means everything is set, no further action.

Solution is to switch the icon and color of the 'apply downloaded update' and 'recheck for updates' buttons.

The color choice is based off of what Biyuen said in the linked ticket:

> This should be documented somewhere, but my general usage has been blue for any primary action, gray for secondary action, and green for a completing action (triggering a success in a multi-step process) or confirming action. In HQ we also use green as the create action. These buttons should probably be blue

Now looks like the following 2 images

Update downloaded and ready to apply:
screen

Recheck server for update:
screen

@amsagoff @orangejenny @biyeun 

ticket: http://manage.dimagi.com/default.asp?220459


biyeun:
:+1:

PR #1172, opened 2016-03-29
Finalize multiple apps

amstone326:
cross-request: https://github.com/dimagi/commcare/pull/283
cross-request: https://github.com/dimagi/javarosa/pull/288


app/src/org/commcare/activities/SingleAppManagerActivity.java
```diff
@@ -140,12 +140,13 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
     /**
-     * Uninstalls the selected app
+     * Uninstalls the selected app and relaunches the App Manager
      */
     private void uninstall() {
         CommCareApplication._().expireUserSession();
         CommCareApplication._().uninstall(appRecord);
-        CommCareApplication.restartCommCare(SingleAppManagerActivity.this);
+        CommCareApplication.restartCommCare(SingleAppManagerActivity.this,
+                SingleAppManagerActivity.class);
```
amstone326: Oddly, this only does what I want (returns to the app manager screen) if I use `SingleAppManagerActivity.class`, and does not work if I use `AppManagerActivity.class`. I discovered this totally by accident and don't know why that is the case, but I'm going to leave it for now.. I tested several times and the behavior was reliable.

app/src/org/commcare/utils/MultipleAppsUtil.java
```diff
@@ -0,0 +1,128 @@
...
+import org.commcare.suite.model.SignedPermission;
+
+import java.util.ArrayList;
+
+/**
+ * Utility methods associated with multiple app seating functionality.
+ *
+ * @author Aliza Stone (astone@dimagi.com)
+ */
+public class MultipleAppsUtil {
```
amstone326: Just moved these methods from CommCareApplication to their own util class

PR #1173, opened 2016-03-30
Allow multimedia verification to be cancelled if launched voluntarily

phillipm:
Allow user to cancel multimedia verification process if they launched it voluntarily.
This includes via the settings menu, during app install in the app manager (@amstone326, is this okay behavior, if not I'll add more flags into `CommCareVerificationActivity), or during validation check from the app manager.

screen

cross-request: https://github.com/dimagi/commcare/pull/276

ticket: http://manage.dimagi.com/default.asp?222474

### DIFF
--- a/app/src/org/commcare/activities/CommCareVerificationActivity.java
+++ b/app/src/org/commcare/activities/CommCareVerificationActivity.java
@@ -184,6 +184,12 @@ public class CommCareVerificationActivity
         task.execute((String[])null);
     }
 
+    @Override
+    public void taskCancelled() {
+        setResult(RESULT_CANCELED);
+        finish();
+    }
+
     private void handleVerificationProblems(SizeBoundVector<MissingMediaException> problems) {
         String message = Localization.get("verification.fail.message");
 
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/activities/CommCareVerificationActivity.java
```diff
@@ -276,6 +282,9 @@ public CustomProgressDialog generateProgressDialog(int taskId) {
                             Localization.get("verification.checking"),
                             taskId);
             dialog.addProgressBar();
+            if (fromSettings || fromManager) {
```
amstone326: @phillipm Hey, the PR looks fine code-wise, but I'm not sure if it addresses the situation Matt was talking about. It sounds from the ticket like he's frustrated by the fact that you can't close out of the MM validation screen that pops up *automatically* immediately after install. But it looks from this line like the Cancel button still wouldn't be available in that case. Let me know if you think I'm misunderstanding Matt's description or the code, but otherwise I would probably vote that it's actually fine to just have the Cancel button there in all instances of this activity.

phillipm: Using the master build, I just tried to install an app via barcode w/ the 'include multimedia' unchecked. It didn't even show a 'verify multimedia' dialog. That makes me think he was referring to manually triggering multimedia. The reproduction steps from the ticket are also:

- Go to Settings
- Validate Media and the 'verifying multimedia' box automatically comes up.
- you used to be able to exit out of this by pushing the back button.

phillipm:
@amstone326 updated to not `finish` the activity when the verification task is cancelled, to allow for the workflow Matt described.

PR #1174, opened 2016-03-31
Ignore constraints on incomplete save

phillipm:
ticket: http://manage.dimagi.com/default.asp?222536

needed after this change: https://github.com/dimagi/commcare-odk/pull/1107

### DIFF
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -1342,8 +1342,7 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
         // save current answer; if headless, don't evaluate the constraints
         // before doing so.
         boolean wasScreenSaved =
-                saveAnswersForCurrentScreen(headless ? DO_NOT_EVALUATE_CONSTRAINTS : EVALUATE_CONSTRAINTS,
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -1342,8 +1342,7 @@ private void saveDataToDisk(boolean exit, boolean complete, String updatedSaveNa
         // save current answer; if headless, don't evaluate the constraints
         // before doing so.
         boolean wasScreenSaved =
-                saveAnswersForCurrentScreen(headless ? DO_NOT_EVALUATE_CONSTRAINTS : EVALUATE_CONSTRAINTS,
-                        complete, headless);
+                saveAnswersForCurrentScreen(DO_NOT_EVALUATE_CONSTRAINTS, complete, headless);
```
phillipm: This is the main save code-path, so it is sort of scary to be making a change like this late in the QA cycle.

Also seems counter-intuitive to make this change, but we are saved by additional constraint checked when the form starts saving: https://github.com/dimagi/commcare-odk/blob/master/app/src/org/commcare/tasks/SaveToDiskTask.java#L94

phillipm:
cc @ctsims

PR #1178, opened 2016-04-01
Show cancel button during data download part of sync

phillipm:
Allow user to cancel `DataPullTask` up until point where data has been downloaded and processing begins.

Don't allow user to cancel syncs that have been started automatically.

Avoid synchronization between button hiding and starting the data processing block. We avoid it because it isn't that bad if a user presses the cancel button after data processing has begun. If that happens, the task dialog says it is cancelling, butt the data processing still completes (because we don't check the `AsyncTask.isCancelled()` state in that block). Resulting in the user being slightly annoyed, if the data processing takes a long time.


app/src/org/commcare/fragments/TaskConnectorFragment.java
```diff
@@ -64,6 +65,9 @@ public void onDetach() {
     public void cancelTask() {
         if (currentTask != null) {
             currentTask.cancel(false);
+            if (currentTask instanceof DataPullTask) {
```
ctsims: can be refactored into commcare task

phillipm:
Created a 2.27 version https://github.com/dimagi/commcare-odk/pull/1226

PR #1181, opened 2016-04-04
UAT feedback for PIN auth

amstone326:
Implemented feedback from UAT:
1. When user signs in with an incorrect PIN, gives a more explicit error msg in the detail to make it clear that they're currently in PIN mode, but can switch to password mode if they've forgotten their PIN
2. Renamed "Switch to Password Mode" menu item on login screen to “Forgot PIN?” to make it clearer what that option is

Also fixed a bug that I introduced by checking for empty passwords.

There was 1 other piece of feedback that I wasn't sure how to implement, if anyone else has ideas. 1 of the UAT reviewers said that when they got to the "Change my PIN" screen, they initially tried to enter their new PIN, before realizing that they were expected to start by entering their existing PIN. It does say in large font on that screen "Enter your current PIN", so I wasn't really sure what else to do there, besides maybe bolding the word "current", but we don't really use any form of text styling elsewhere in the app, so felt weird introducing that pattern for the first time. I feel pretty fine with leaving this as-is since the feedback only came from 1 person, unless anyone has a good suggestion. 

UAT ticket: http://manage.dimagi.com/default.asp?212999
cc: @dimagi/product


app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -154,15 +154,23 @@ protected void onSaveInstanceState(Bundle savedInstanceState) {
      *                       upon successful login
      */
     protected void initiateLoginAttempt(boolean restoreSession) {
-        if (uiController.getEnteredPasswordOrPin().equals("")) {
-            raiseLoginMessage(StockMessages.Auth_EmptyPassword, false);
-            return;
+        LoginMode loginMode = uiController.getLoginMode();
+
+        if (uiController.getEnteredPasswordOrPin().equals("") &&
```
phillipm: I generally like to the `equals` call to be made on the hard-coded string (`""`) if possible to avoid any potential for a null pointer exception.

app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -154,15 +154,23 @@ protected void onSaveInstanceState(Bundle savedInstanceState) {
...
+        LoginMode loginMode = uiController.getLoginMode();
+
+        if ("".equals(uiController.getEnteredPasswordOrPin()) &&
+                loginMode != LoginMode.PRIMED) {
+            if (loginMode == LoginMode.PASSWORD) {
+                raiseLoginMessage(StockMessages.Auth_EmptyPassword, false);
+                return;
+            } else {
+                raiseLoginMessage(StockMessages.Auth_EmptyPin, false);
+                return;
```
phillipm: Can you lift the `return` from both branches of the if/else out below the conditional? It eliminates the need for 2 `return` statements.

PR #1184, opened 2016-04-04
Fix for NSD Attachment blocking

ctsims:
Workaround for NSD Services being a disaster and resulting in many-second blocking calls. Now will abandon the search after a brief window (it won't succeed after it's gotten into the bugged state anyway)

Ticket:
http://manage.dimagi.com/default.asp?222766#1120157

### DIFF
--- a/app/src/org/commcare/android/nsd/NSDDiscoveryTools.java
+++ b/app/src/org/commcare/android/nsd/NSDDiscoveryTools.java
@@ -130,6 +130,7 @@ public class NSDDiscoveryTools {
                     manager = (NsdManager)context.getSystemService(Context.NSD_SERVICE);
                 }
 
+                @Override
                 public void commit() {
                     mNsdManager = manager;
                 }
--- a/app/src/org/commcare/android/nsd/NSDDiscoveryTools.java
+++ b/app/src/org/commcare/android/nsd/NSDDiscoveryTools.java
@@ -7,6 +7,10 @@ import android.net.nsd.NsdServiceInfo;
 import android.os.Build;
 import android.util.Log;
 
+import org.commcare.logging.AndroidLogger;
+import org.commcare.utils.TimeBoundOperation;
+import org.javarosa.core.services.Logger;
+
 import java.net.InetAddress;
 import java.util.Collection;
 import java.util.HashMap;
@@ -37,7 +41,8 @@ public class NSDDiscoveryTools {
     public enum NsdState {
         Init,
         Discovery,
-        Idle
+        Idle,
+        Unsupported
     }
 
     private static NsdState state = NsdState.Init;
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/android/nsd/NSDDiscoveryTools.java
```diff
@@ -106,6 +113,41 @@ private static void doDiscovery(Context context) {
...
+
+            boolean connected = new TimeBoundOperation(500) {
+                NsdManager manager;
+
+                @Override
+                public void run() {
+                    manager = (NsdManager)context.getSystemService(Context.NSD_SERVICE);
+                }
+
+                public void commit() {
```
phillipm: Needs an `@Override`

app/src/org/commcare/utils/TimeBoundOperation.java
```diff
@@ -0,0 +1,76 @@
...
+            }
+        });
+
+        thread.start();
+        try {
+            thread.join(timeout);
+        }catch (InterruptedException e) {
+            //Execution failed
+            return false;
+        };
```
phillipm: unneeded `;`

app/src/org/commcare/utils/TimeBoundOperation.java
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * A time bound operation can only be executed once.
+ *
+ * Created by ctsims on 4/4/2016.
+ */
+public abstract class TimeBoundOperation {
+
+    final private long timeout;
+
+    private Thread thread;
```
phillipm: can be local to `execute`

app/src/org/commcare/utils/TimeBoundOperation.java
```diff
@@ -0,0 +1,76 @@
+package org.commcare.utils;
+
+/**
+ * A time bound operation will run a synchronized operation in the background, with the confidence
```
phillipm: "in the background": isn't it blocking the current thread? If so, maybe say something like "on a new, blocking thread"

app/src/org/commcare/android/nsd/NSDDiscoveryTools.java
```diff
@@ -37,7 +41,8 @@
     public enum NsdState {
         Init,
         Discovery,
-        Idle
+        Idle,
+        Unsupported
```
phillipm: Might want to check for `Unsupported` in `addListener`

ctsims: We actually won't know when we add the listener, which seems fine. The caller is asking to subscribe to events that will never happen, but I don't think that should cause any lifecycle issues.

ctsims:
@phillipm Feedback addressed, thanks for review!

PR #1188, opened 2016-04-06
Improve unsent forms subtext + fix subtext coloring

amstone326:
Chatted with Sheel about this ticket http://manage.dimagi.com/default.asp?211211 and determined that the only lingering change necessary was to make the Sync button subtext clearer. That subtext previously read "Sync with Server (x)", where x is the number of unsent forms. It seems silly to duplicate the main text that's already on the button, and not give a good description of what that number actually represents, so changing it to instead read "Unsent Forms: x".

(The other concern on this ticket -- the error text persisting and therefore obfuscating the unsent forms count -- appears to have been resolved at some point since the version the partner was on).

I also noticed while working on this that that somewhere in the process of this code getting refactored, some unreachable code was left in, and the subtext color was not always getting set to warning when it should be. Comments inline as to what I changed there.

### DIFF
--- a/app/assets/locales/messages_ccodk_default.txt
+++ b/app/assets/locales/messages_ccodk_default.txt
@@ -79,7 +79,7 @@ home.forms.incomplete=Incomplete
 home.forms.incomplete.indicator=${1} (${0})  
 home.forms.saved=Saved
 home.sync=Sync with Server
-home.sync.indicator=Unsent Forms: ${0}
+home.unsent.forms.indicator=Unsent Forms: ${0}
 home.logout=Log out of CommCare
 home.report=Report an Issue
 
--- a/app/assets/locales/messages_ccodk_default.txt
+++ b/app/assets/locales/messages_ccodk_default.txt
@@ -79,7 +79,7 @@ home.forms.incomplete=Incomplete
 home.forms.incomplete.indicator=${1} (${0})  
 home.forms.saved=Saved
 home.sync=Sync with Server
-home.sync.indicator=${1} (${0})
+home.sync.indicator=Unsent Forms: ${0}
 home.logout=Log out of CommCare
 home.report=Report an Issue
 
--- a/app/src/org/commcare/utils/SyncDetailCalculations.java
+++ b/app/src/org/commcare/utils/SyncDetailCalculations.java
@@ -37,7 +37,7 @@ public class SyncDetailCalculations {
         Pair<Long, String> lastSyncTimeAndMessage = getLastSyncTimeAndMessage();
 
         if (numUnsentForms > 0) {
-            Spannable syncIndicator = (activity.localize("home.sync.indicator",
+            Spannable syncIndicator = (activity.localize("home.unsent.forms.indicator",
                     new String[]{String.valueOf(numUnsentForms)}));
             squareButtonViewHolder.subTextView.setText(syncIndicator);
         } else {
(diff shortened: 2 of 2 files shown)


app/src/org/commcare/utils/SyncDetailCalculations.java
```diff
@@ -53,20 +58,15 @@ private static void showSyncMessage(TextView subTextView, int normalColor, int w
         } else {
             syncTimeMessage = DateUtils.formatSameDayTime(lastSyncTime, new Date().getTime(), DateFormat.DEFAULT, DateFormat.DEFAULT);
         }
+        return new Pair<>(lastSyncTime, Localization.get("home.sync.message.last", new String[]{syncTimeMessage.toString()}));
+    }
 
-        String message = "";
```
amstone326: This whole section is no longer ever relevant/used because if we're showing a sync message based upon the last sync time, that means there were no unsent forms.

app/src/org/commcare/utils/SyncDetailCalculations.java
```diff
@@ -29,22 +30,26 @@
...
         int numUnsentForms = formsStorage.getIDsForValue(FormRecord.META_STATUS, FormRecord.STATUS_UNSENT).size();
+
+        Pair<Long, String> lastSyncTimeAndMessage = getLastSyncTimeAndMessage();
+
         if (numUnsentForms > 0) {
             Spannable syncIndicator = (activity.localize("home.sync.indicator",
-                    new String[]{String.valueOf(numUnsentForms), cardDisplayData.text}));
+                    new String[]{String.valueOf(numUnsentForms)}));
             squareButtonViewHolder.subTextView.setText(syncIndicator);
-            squareButtonViewHolder.subTextView.setTextColor(activity.getResources().getColor(cardDisplayData.subTextColor));
```
amstone326: In the case where the subtext was being set based upon unsent forms, we were never checking to see if the background color should be the warning color, even though the implementation of `isSyncStronglyNeeded()` implies that we want to set the warning background when there are a lot of unsent forms.

amstone326:
Before:
screen

After:
screen

app/assets/locales/messages_ccodk_default.txt
```diff
@@ -79,7 +79,7 @@ home.forms.incomplete=Incomplete
 home.forms.incomplete.indicator=${1} (${0})  
 home.forms.saved=Saved
 home.sync=Sync with Server
-home.sync.indicator=${1} (${0})
+home.sync.indicator=Unsent Forms: ${0}
```
phillipm: Curious what happens when people don't update their app translations when updating to 2.27 and we try to localize `${1} (${0})`? Can we make sure we fail soft?

phillipm: Not actually sure this is an issue, just worth double checking...

app/src/org/commcare/utils/SyncDetailCalculations.java
phillipm: :+1:

phillipm:
Good to go after you check that the failure mode if the user provides a custom translation for `home.sync.indicator` that has two place-holders instead of one. My vote would be to implement a safe 'localization w/ placeholders' method that returns "INVALID LOCALIZATION" or something that helps the app building user known they need to update something. But maybe that would just confuse the user or is overly complicating our code...

@ctsims and @wpride: any thoughts on this issue, is it a real concern?

amstone326:
For anyone following along, @phillipm and I decided that the easiest way to prevent weirdness with old custom translations is just to change the key for that translation, rendering any old custom translations unused.

PR #1189, opened 2016-04-06
Represent datum data using OO

phillipm:
cross-request: https://github.com/dimagi/commcare/pull/282

### DIFF
--- a/app/src/org/commcare/models/AndroidSessionWrapper.java
+++ b/app/src/org/commcare/models/AndroidSessionWrapper.java
@@ -10,6 +10,7 @@ import org.commcare.models.database.user.models.FormRecord;
 import org.commcare.models.database.user.models.SessionStateDescriptor;
 import org.commcare.session.CommCareSession;
 import org.commcare.session.SessionFrame;
+import org.commcare.suite.model.EntityDatum;
 import org.commcare.suite.model.Entry;
 import org.commcare.suite.model.FormEntry;
 import org.commcare.suite.model.SessionDatum;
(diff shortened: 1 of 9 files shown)


app/src/org/commcare/models/AndroidSessionWrapper.java
```diff
@@ -308,28 +226,30 @@ public static AndroidSessionWrapper mockEasiestRoute(CommCarePlatform platform,
...
-                        //Record the degree of specificity of this selection for now (we'll
-                        //actually create the wrapper later
-                        curPredicates = countPredicates;
-                    } else {
-                        //There's already a path to this form. Only keep going 
-                        //if the current choice is less specific
-                        if (countPredicates >= curPredicates) {
-                            continue;
+                    // we only know how to mock a single case selection
+                    if (datum instanceof EntityDatum) {
```
phillipm: Note, these changes are inside of the `mockEasiestRoute` method, which is a brittle helper for a debug feature: manually load form payloads onto device. This feature is rarely used so we don't need to be cautious here.

app/src/org/commcare/tasks/FormRecordLoaderTask.java
```diff
@@ -260,4 +274,83 @@ public boolean registerPriority(FormRecord record) {
         }
     }
 
+    private static String getTitleFromSession(AndroidSessionWrapper androidSessionWrapper) {
+        //TODO: Most of this mimicks what we need to do in entrydetail activity, remove it from there
+        //and generalize the walking
+
+        // get a copy of the session
+        CommCareSession session = new CommCareSession(androidSessionWrapper.getSession());
```
wpride: Why do we need to get a copy of the session? Is it ok that its a shallow copy (which I think it is?)

phillipm: See https://github.com/dimagi/commcare-odk/pull/1187 , which addresses https://github.com/dimagi/commcare-odk/pull/1187/files#diff-eebcc4a423830db1c3a3aa4d9d827c6cL201 which seems pretty sketch to me, even if it is behaviorally okay.

PR #1190, opened 2016-04-07
Use ResultAndError to store task results that have error messages

phillipm:
Follow-up to request by @wpride : https://github.com/dimagi/commcare-odk/pull/1159#discussion_r58723208


wpride:
awesome

PR #1192, opened 2016-04-07
Be less agressive in notifying http submission progress (round II)

phillipm:
Re-apply changes from https://github.com/dimagi/commcare-odk/pull/1179 because we accidentally reverted it.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -19,7 +19,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.0.0-beta7'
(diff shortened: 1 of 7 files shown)


build.gradle
```diff
@@ -19,7 +19,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.0.0-beta7'
+        classpath 'com.android.tools.build:gradle:2.0.0'
```
phillipm: Probably a good idea to bump to the major release

PR #1193, opened 2016-04-08
Refactor: HorizontalMediaView

phillipm:
- Break-up method into smaller helpers
- Remove navigation icon code, which is never executed because `int navStyle` is always set to `NAVIGATION_NONE`

perhaps easiest reviewed commit-by-commit

### DIFF
--- a/app/src/org/commcare/views/HorizontalMediaView.java
+++ b/app/src/org/commcare/views/HorizontalMediaView.java
@@ -14,7 +14,6 @@ import org.commcare.suite.model.DisplayData;
 import org.commcare.suite.model.DisplayUnit;
 import org.commcare.utils.MediaUtil;
 import org.commcare.views.media.AudioButton;
-import org.javarosa.core.model.condition.EvaluationContext;
 import org.javarosa.core.reference.InvalidReferenceException;
 import org.javarosa.core.reference.ReferenceManager;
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/views/HorizontalMediaView.java
```diff
@@ -27,137 +26,112 @@
...
-    }
-
-    private HorizontalMediaView(Context c, EvaluationContext ec) {
         super(c);
-        mTextView = null;
-        mAudioButton = null;
-        mImageView = null;
-        this.ec = ec;
-        this.iconDimension = (int)getResources().getDimension(R.dimen.menu_icon_size);
+        this.iconDimension = (int) getResources().getDimension(R.dimen.menu_icon_size);
```
amstone326: Thought our official style policy on casting was no space, but can't remember where we actually have that stuff documented so not positive..

phillipm: haha, it is!
There was a brief point where I changed android studio versions and forgot to install our code format config file.

amstone326: Ok wanna change that back really quick and then I'll merge?

phillipm:
updated

PR #1194, opened 2016-04-08
Move modern http requests to own file

phillipm:
Move simple (modern) get request code into own file and refactored how we make requests without authentication to be more explicit.
 
Spent a little time thinking about how to add test coverage to this and couldn't think of a good way.

Verified by hand that the http (auth) code changed still works:
 - Installing apps from the server
 - Submitting forms as normal and demo users
 - Submitting logs


app/src/org/commcare/network/HttpRequestGenerator.java
```diff
@@ -86,32 +83,37 @@ public HttpRequestGenerator(User user) {
...
         }
+    }
 
-        this.userType = userType;
+    protected static String buildDomainUser(String username) {
+        if (username != null) {
+            SharedPreferences prefs = CommCareApplication._().getCurrentApp().getAppPreferences();
+
+            if (prefs.contains(USER_DOMAIN_SUFFIX)) {
+                username += "@" + prefs.getString(USER_DOMAIN_SUFFIX, null);
```
wpride: Shouldn't we error hard here if we don't have a domain rather than do test@null.commcarehq.org and fail later?

app/src/org/commcare/network/HttpRequestGenerator.java
```diff
@@ -86,32 +83,37 @@ public HttpRequestGenerator(User user) {
...
-        if (prefs.contains(USER_DOMAIN_SUFFIX)) {
-            domainedUsername += "@" + prefs.getString(USER_DOMAIN_SUFFIX, null);
-        }
+    private HttpRequestGenerator(String username, String password, String userType) {
+        String domainedUsername = buildDomainUser(username);
+        this.password = password;
+        this.userType = userType;
 
-        if (!User.TYPE_DEMO.equals(userType)) {
+        if (username != null && !User.TYPE_DEMO.equals(userType)) {
```
wpride: Why do we need to check username nullity?

app/src/org/commcare/network/HttpRequestGenerator.java
```diff
@@ -86,32 +83,37 @@ public HttpRequestGenerator(User user) {
...
             this.credentials = new UsernamePasswordCredentials(domainedUsername, password);
-            passwordAuthentication = new PasswordAuthentication(domainedUsername, password.toCharArray());
             this.username = username;
+        } else {
+            this.credentials = null;
+            this.username = null;
```
wpride: think member variables are instantiated to null automatically

app/src/org/commcare/network/SimpleGetRequest.java
```diff
@@ -0,0 +1,88 @@
...
+            Authenticator.setDefault(new Authenticator() {
+                @Override
+                protected PasswordAuthentication getPasswordAuthentication() {
+                    return new PasswordAuthentication(
+                            HttpRequestGenerator.buildDomainUser(username),
+                            password.toCharArray());
+                }
+            });
+        }
+        return makeModernRequest(url);
```
wpride: A bit confused about the null vs. IOException result here, might be worth a comment explaining. So we return null if the exception is an auth issue in which case we fall back to the apache libraries, otherwise we bubble the IOException up to the top level for handling? Does the Auth issue not consistently through a more specific IOException we can catch? The combination of null and exception just feels a bit clunky to me

app/src/org/commcare/network/HttpRequestGenerator.java
phillipm: Needs to here since `username` is `final`

app/src/org/commcare/network/HttpRequestGenerator.java
phillipm: See `buildNoAuthGenerator`

app/src/org/commcare/network/HttpRequestGenerator.java
phillipm: Probably, but I was just copying existing code and didn't want to modify it too much

app/src/org/commcare/network/SimpleGetRequest.java
phillipm: I agree that this is convoluted code.
I was just lifting code unchanged out of `HttpRequestGenerator` into this file, so I didn't modify anything here. The eventual goal is to stop using apache http code completely, which is deprecated. So I'd like to leave cleaning this up until we get to that point.

PR #1196, opened 2016-04-11
Removing crypto logic from robolectric tests

amstone326:
Don't perform encryption of user keys/passwords in robolectric tests, so that we don't encounter exceptions due to the standard jdk missing certain crypto implementations.


phillipm:
:raised_hands: :muscle: :muscle: :muscle: :raised_hands: 
Now running android tests won't be limited to only the mobile devs running Linux (me)

PR #1201, opened 2016-04-15
Fix user notifications for localization exceptions

amstone326:
The user-facing error activity was not always being displayed when a crash was caused by a NoLocalizedTextException, because we weren't checking the Throwable's cause(s).

@phillipm Seems like something that could be worth sneaking into 2.27?

### DIFF
--- a/app/src/org/commcare/utils/CommCareExceptionHandler.java
+++ b/app/src/org/commcare/utils/CommCareExceptionHandler.java
@@ -47,7 +47,7 @@ public class CommCareExceptionHandler implements UncaughtExceptionHandler {
      * they can fix.
      */
     private boolean warnUserAndExit(Throwable ex) {
-        if (ex instanceof NoLocalizedTextException) {
+        if (causedByLocalizationException(ex)) {
             Intent i = new Intent(ctx, CrashWarningActivity.class);
             i.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/utils/CommCareExceptionHandler.java
```diff
@@ -57,4 +57,14 @@ private boolean warnUserAndExit(Throwable ex) {
...
     }
+
+    private static boolean causedByLocalizationException(Throwable ex) {
+        if (ex == null) {
+            return false;
+        }
+        if (ex instanceof NoLocalizedTextException) {
+            return true;
+        }
+        return causedByLocalizationException(ex.getCause());
```
phillipm: can you collapse these conditional branches into a single boolean expression?

amstone326: Not actually seeing how I can do that and get the identical result.. What would you change it to?

phillipm: `return ex != null && (ex instanceof NoLocalizedTextException || causedByLocalizationException(ex.getCause()))`

amstone326: Ah yea ok that reads better too. Should we put in 2.27?

phillipm: yeah probably good to have this fixed.

PR #1205, opened 2016-04-16
Execute update task in parallel

phillipm:
Execute update task in parallel. Theis brought up the fact that when the update task is downloading updates in the background, other user actions are slowed down. This is due to not using the `AsyncTask.executeOnExecutor` task launching method.

Don't change other tasks to use the parallel call because it is a non-trivial change that we should QA in general.


amstone326:
@phillipm Are we sure it's safe to make even this change without QAing it?

phillipm:
I'm fairly sure this is a safe change, all the work done in the update task is downloading data into a temporary upgrade table. The actually update transaction occurs on next login. I verified that without this, if you try to load a case list while also updating the app it won't load the cases. With this change, the cases load and

PR #1206, opened 2016-04-16
Use normal PrototypeFactory in Robolectric tests

phillipm:
Make sure Robolectric tests that use `CommCareTestApplication` share more prototype factory code with the live app.

Would have caught the bug fixed by https://github.com/dimagi/commcare-odk/pull/1215 

Includes commits from https://github.com/dimagi/commcare-odk/pull/1224

### DIFF
--- a/app/src/org/commcare/CommCareApplication.java
+++ b/app/src/org/commcare/CommCareApplication.java
@@ -61,6 +61,7 @@ import org.commcare.logging.analytics.TimedStatsTracker;
 import org.commcare.models.AndroidClassHasher;
 import org.commcare.models.AndroidSessionWrapper;
 import org.commcare.models.database.AndroidDbHelper;
+import org.commcare.models.database.DbUtil;
 import org.commcare.models.database.HybridFileBackedSqlHelpers;
 import org.commcare.models.database.HybridFileBackedSqlStorage;
 import org.commcare.models.database.MigrationException;
@@ -105,6 +106,7 @@ import org.javarosa.core.services.locale.Localization;
 import org.javarosa.core.services.storage.EntityFilter;
 import org.javarosa.core.services.storage.Persistable;
 import org.javarosa.core.util.PropertyUtils;
+import org.javarosa.core.util.externalizable.PrototypeFactory;
 
 import java.io.File;
 import java.lang.ref.WeakReference;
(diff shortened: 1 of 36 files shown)


app/src/org/commcare/CommCareApplication.java
```diff
@@ -1476,4 +1478,7 @@ public boolean checkPendingBuildRefresh() {
         return false;
     }
 
+    public PrototypeFactory getPrototypeFactory(Context c) {
```
wpride: Why is this function necessary? (instead of just calling DbUtil.getPrototypeFactory()

phillipm: To allow for overriding in the robolectric tests via `CommCareTestApplication`

wpride: Interesting - and you cant override static methods I'm guessing?

Having to get the singleton CommCareApplication to call this method that doesn't actually need the instance seems wrong to me. Am I missing something or is this just a tradeoff we have to accept to have tests?

app/src/org/commcare/interfaces/HttpRequestEndpoints.java
```diff
@@ -0,0 +1,27 @@
...
+import java.io.InputStream;
+import java.net.URL;
+import java.util.Date;
+
+/**
+ * Types of http requests made by CommCare mobile to server
+ *
+ * @author Phillip Mates (pmates@dimagi.com)
+ */
+public interface HttpRequestEndpoints {
```
wpride: Very cool

app/src/org/commcare/models/database/DbUtil.java
```diff
@@ -28,6 +28,7 @@
     public final static String orphanFileTableName = "OrphanedFiles";
 
     private static PrototypeFactory factory;
+    private static final String[] packageNames = new String[]{"org.javarosa", "org.commcare", "org.odk.collect"};
```
wpride: org.odk.collect - can be removed yeah?

app/src/org/commcare/CommCareApplication.java
phillipm: It does feel wrong and there are various tools to get around it, like dependency injection frameworks.

I would like to revisit this at some point, but for now it is the best way I could come up with.

And yeah, overriding static methods isn't a thing in Java.

app/src/org/commcare/models/database/DbUtil.java
phillipm: We had to add it back due to not currently having the ability to migrate the keys in a prototype factory. I have an idea of how to do that, and will eventually follow through and remove this.

PR #1210, opened 2016-04-18
Tests: use same entry point for all app install logic

phillipm:
Removing a lot of duplicate app install code from odk unit tests.


amstone326:
nice, this is much cleaner!

PR #1213, opened 2016-04-20
Log intent callouts to server

wpride:
I meant to log these to the server

### DIFF
--- a/app/src/org/odk/collect/android/jr/extensions/IntentCallout.java
+++ b/app/src/org/odk/collect/android/jr/extensions/IntentCallout.java
@@ -7,6 +7,7 @@ import android.os.BadParcelableException;
 import android.os.Bundle;
 import android.util.Log;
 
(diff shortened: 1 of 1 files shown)


app/src/org/odk/collect/android/jr/extensions/IntentCallout.java
```diff
@@ -114,7 +116,7 @@ public Intent generate(EvaluationContext ec) {
                 }
             }
         }
-        Log.d(TAG, "Generated Intent: " + i.toString());
+        Logger.log(AndroidLogger.TYPE_FORM_ENTRY, "Generated Intent: " + i.toString());
```
amstone326: Being picky but seems like TYPE_FORM_ENTRY isn't the best fit here

amstone326: (nvm, explained offline)

PR #1214, opened 2016-04-20
prevent OOM error when processing image capture

amstone326:
Addresses this crash from the ICDS force-close logs:
```java.lang.OutOfMemoryError at android.graphics.BitmapFactory.nativeDecodeStream(Native Method) at android.graphics.BitmapFactory.decodeStreamInternal(BitmapFactory.java:709) at android.graphics.BitmapFactory.decodeStream(BitmapFactory.java:685) at android.graphics.BitmapFactory.decodeFile(BitmapFactory.java:447) at android.graphics.BitmapFactory.decodeFile(BitmapFactory.java:486) at org.odk.collect.android.utilities.FileUtils.scaleAndSaveImage(FileUtils.java:128) at org.odk.collect.android.activities.components.ImageCaptureProcessing.moveAndScaleImage(ImageCaptureProcessing.java:36) at org.odk.collect.android.activities.FormEntryActivity.processCaptureResponse(FormEntryActivity.java:484) at org.odk.collect.android.activities.FormEntryActivity.onActivityResult(FormEntryActivity.java:443) at android.app.Activity.dispatchActivityResult(Activity.java:5643) at android.app.ActivityThread.deliverResults(ActivityThread.java:3573) at android.app.ActivityThread.handleSendResult(ActivityThread.java:3620) at android.app.ActivityThread.access$1400(ActivityThread.java:169) at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1325) at android.os.Handler.dispatchMessage(Handler.java:102) at android.os.Looper.loop(Looper.java:136) at android.app.ActivityThread.main(ActivityThread.java:5479) at java.lang.reflect.Method.invokeNative(Native Method) at java.lang.reflect.Method.invoke(Method.java:515) at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1283) at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1099) at dalvik.system.NativeStart.main(Native Method)```

Did not reproduce the specific crash, but definitely just looks like processing an image capture is sometimes causing an OOM error because the original captured image is too large. The only question I have here is whether it's proper to just downsize the user's captured image without telling them we did so. I guess the alternative would be to show a non-crashing error message to the user that tells them their image capture can't be processed due to memory constraints. I'm personally inclined toward downsizing the image, because the alternative leaves the user without any real path forward to fix the issue. But wanted to acknowledge that this solution does something potentially odd. 

(I did test out going through some image capture widgets with this change to make sure it still worked soundly).

### DIFF
--- a/app/src/org/commcare/utils/FileUtil.java
+++ b/app/src/org/commcare/utils/FileUtil.java
@@ -435,7 +435,7 @@ public class FileUtil {
             return false;
         }
 
-        Bitmap bitmap = BitmapFactory.decodeFile(originalImage.getAbsolutePath());
+        Bitmap bitmap = MediaUtil.inflateImageSafe(originalImage.getAbsolutePath());
         Bitmap scaledBitmap = getBitmapScaledByMaxDimen(bitmap, maxDimen);
         if (scaledBitmap != null) {
             // Write this scaled bitmap to the final file location
--- a/app/src/org/commcare/utils/FileUtil.java
+++ b/app/src/org/commcare/utils/FileUtil.java
@@ -12,12 +12,15 @@ import android.os.Environment;
 import android.provider.DocumentsContract;
 import android.provider.MediaStore;
 import android.util.Log;
+import android.util.Pair;
 
+import org.commcare.logging.AndroidLogger;
 import org.commcare.resources.model.MissingMediaException;
 import org.commcare.resources.model.Resource;
 import org.javarosa.core.reference.InvalidReferenceException;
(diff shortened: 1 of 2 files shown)


wpride:
Nice catch. I agree with your assessment that we should be automatically scaling down. Seems like it might be worthwhile to log though so we can find evidence this is happening IE if someone reports that their image submissions are grainy

amstone326:
@wpride good idea, updated

PR #1215, opened 2016-04-20
Include odk classes in prototype factory class setup

phillipm:
Fix form deserialization bug introduced earlier in 2.27 where classes that start with the `org.odk.collect` package weren't included in the prototype factory class list.

Caused form deserialization to always fail, and revert back to xml form loading, when the form had intent questions in it.

Bug introduced by https://github.com/dimagi/commcare-odk/pull/1050

### DIFF
--- a/app/src/org/commcare/models/database/DbUtil.java
+++ b/app/src/org/commcare/models/database/DbUtil.java
@@ -11,6 +11,7 @@ import net.sqlcipher.database.SQLiteDatabaseHook;
 import org.commcare.models.AndroidPrototypeFactory;
 import org.commcare.modern.database.DatabaseHelper;
 import org.javarosa.core.util.PrefixTree;
+import org.javarosa.core.util.externalizable.Externalizable;
 import org.javarosa.core.util.externalizable.PrototypeFactory;
 
 import java.io.File;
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/models/database/DbUtil.java
```diff
@@ -76,9 +77,7 @@ public static PrototypeFactory getPrototypeFactory(Context c) {
             String cn = en.nextElement();
             try {
                 for (String packageName : packageNames) {
-
-                    if (cn.startsWith(packageName) && !cn.startsWith("org.commcare.dalvik") && !cn.contains(".test.") && !cn.contains("readystatesoftware")) {
```
phillipm: I removed `!cn.startsWith("org.commcare.dalvik")` because `org.commcare.dalvik` isn't a package anymore. Bye dalvik!

app/src/org/commcare/models/database/DbUtil.java
```diff
@@ -94,7 +93,9 @@ public static PrototypeFactory getPrototypeFactory(Context c) {
                         if (!emptyc) {
                             continue;
                         }
-                        classNames.add(cn);
+                        if (Externalizable.class.isAssignableFrom(prototype)) {
+                            classNames.add(cn);
+                        }
```
phillipm: Realizing, now that I understand the prototype factory code better, that https://github.com/dimagi/commcare-odk/pull/1050/files#r51792765 deleted the wrong `classNames.add(cn)` code

PR #1216, opened 2016-04-21
Allow for custom acknowledgement in 'about commcare'

phillipm:
ICDS feature to allow for adding custom acknowledgement in 'About CommCare' dialog.

Requires setting `custom.acknowledgement` translation key on HQ.
Requires double `\n\n` to do a new line break.

Adds the markdown enabled "Foo bar hello bold?".
screen

When `custom.acknowledgement` is unset the about dialog looks like:
screen


amstone326:
Cool! Have you double checked that the appearance of this doesn't change if the custom.acknowledgedment translation string is empty?

phillipm:
@amstone326 , Ha! Good point; made an incorrect assumption that crashed the app...
Also updated the description with a screenshot.

PR #1219, opened 2016-04-22
Framework for allowing controlling when task is cancelable

phillipm:


app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -104,249 +111,291 @@ private DataPullTask(String username, String password, String server, Context co
...
+                // checked locally
+                //TODO: Keys were lost somehow.
+                sue.printStackTrace();
+                responseError = PullTaskResult.UNKNOWN_FAILURE;
+                return;
             }
 
-            PullTaskResult responseError = PullTaskResult.UNKNOWN_FAILURE;
+            // Only purge cases if we already had a logged in user. Otherwise we probably can't read the DB.
+            CaseUtils.purgeCases();
```
phillipm: Wondering if we can move this out of the `setup` method, because at this point the cancel button is being shown, but won't dismiss until this finishes, so there may be some lag for the user if they press the cancel button and there are a lot of cases to purge

app/src/org/commcare/views/dialogs/CustomProgressDialog.java
```diff
@@ -156,11 +157,9 @@ public Dialog onCreateDialog(Bundle savedInstanceState) {
...
         messageView.setText(message);
 
-        if (usingCancelButton) {
-            Button cancelButton = setupCancelButton(view);
-            if (wasCancelPressed) {
-                setCancellingText(titleView, messageView, cancelButton);
-            }
+        setupCancelButton(view);
+        if (wasCancelPressed) {
+            setCancellingText(titleView, messageView, cancelButton);
```
ctsims: We can't guarantee that the action was cancellable when this was pressed, we likely need to wait for a signal that says "cancel accepted" or something similar.

ctsims: nevermind, missed that this was incoming from a previous state.

app/src/org/commcare/activities/CommCareActivity.java
```diff
@@ -406,6 +406,11 @@ public void taskCancelled() {
 
     public void cancelCurrentTask() {
         stateHolder.cancelTask();
+
```
ctsims: This actually needs to fire the cancel signal, _and_ validate whether the signal can be processed (IE: Are we past the threshold where we're already processing)

app/src/org/commcare/tasks/templates/CommCareTask.java
```diff
@@ -34,7 +42,24 @@ public CommCareTask() {
...
+                return null;
+            }
+            mainBackgroundWork(params);
+
+            if (isCancelled()) {
+                return null;
+            }
+
+            // hide cancel button during commit phase
+            disableCancelButton();
```
ctsims: This also needs to denote that "Nothing should process the 'cancel' signal anymore, even if the disable signal hasn't propogated out yet"

app/src/org/commcare/activities/CommCareActivity.java
```diff
@@ -406,6 +406,11 @@ public void taskCancelled() {
 
     public void cancelCurrentTask() {
         stateHolder.cancelTask();
+
+        if (stateHolder.canDetachFromCancelledTask()) {
+            // let cancelled task burn out in the background
+            dismissProgressDialog();
```
ctsims: should also detach the task and orphan it

app/src/org/commcare/fragments/TaskConnectorFragment.java
```diff
@@ -67,6 +68,22 @@ public void cancelTask() {
...
+    /**
+     * @return Task is in a state where, if cancelled, it can dismiss its
+     * blocking dialog and allow the user them to continue using the app
+     */
+    public boolean canDetachFromCancelledTask() {
+        return currentTask == null || currentTask.canStopUIBlockOnCancel();
+    }
+
+    public boolean isDialogCancelButtonEnabled() {
+        return isTaskDialogCancelButtonEnabled;
```
ctsims: this can come straight from the task itself (even as a direct, synchronous method call), and there's no need to double-store the state in a new boolean

app/src/org/commcare/tasks/templates/CommCareTask.java
```diff
@@ -50,6 +75,49 @@ protected final Result doInBackground(Params... params) {
...
+
+    /**
+     * Phase of task that cannot be cancelled.
+     */
+    protected Result commit(Params... params) {
+        return doTaskBackground(params);
+    }
+
+    private void enableCancelButton() {
+        showCancelButton = true;
```
ctsims: I think can just be "isCancellable"

app/src/org/commcare/tasks/templates/CommCareTask.java
```diff
@@ -50,6 +75,49 @@ protected final Result doInBackground(Params... params) {
...
+        canDismissOnCancel = true;
+    }
+
+    private void disableCancelButton() {
+        canDismissOnCancel = false;
+        showCancelButton = false;
+        publishCancelDialogUpdate();
+    }
+
+    public boolean canStopUIBlockOnCancel() {
```
ctsims: This method name requires a lot of parsing

app/src/org/commcare/tasks/templates/CommCareTask.java
```diff
@@ -197,4 +265,57 @@ public void disconnect() {
...
+    private static Handler getHandler() {
+        synchronized (CommCareTask.class) {
+            if (sHandler == null) {
+                sHandler = new InternalUIHandler();
+            }
+            return sHandler;
+        }
+    }
+
+    private static class InternalUIHandler extends Handler {
```
ctsims: Check out Activity.runOnUiThread. Let me know if that won't work for this

PR #1224, opened 2016-04-25
Test: Code coverage for DataPullTask

phillipm:
Upping the test coverage of `DataPullTask`. Doesn't necessarily check logical correctness of code, but covers most error handling code paths.


app/src/org/commcare/interfaces/HttpRequestEndpoints.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+import java.util.Date;
+
+/**
+ * @author Phillip Mates (pmates@dimagi.com)
+ */
+public interface HttpRequestEndpoints {
```
phillipm: I don't like this name; any ideas? `CommCareHttpRequests`?

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -156,7 +156,7 @@ protected void onCancelled() {
             //This should be per _user_, not per app
             prefs.edit().putLong("last-ota-restore", new Date().getTime()).commit();
 
-            HttpRequestGenerator requestor = new HttpRequestGenerator(username, password);
+            HttpRequestEndpoints requestor = dataPullRequester.getHttpGenerator(username, password);
```
phillipm: route http generator creation through `DataPullResponseFactory` so we can easily override it in tests.

PR #1226, opened 2016-04-26
Show cancel button during data download part of sync

phillipm:
Allow user to cancel `DataPullTask` up until point where data has been downloaded and processing begins.

Don't allow user to cancel syncs that have been started automatically.

Avoid synchronization between button hiding and starting the data processing block. We avoid it because it isn't that bad if a user presses the cancel button after data processing has begun. If that happens, the task dialog says it is cancelling, butt the data processing still completes (because we don't check the `AsyncTask.isCancelled()` state in that block). Resulting in the user being slightly annoyed, if the data processing takes a long time.

### DIFF
--- a/app/src/org/commcare/activities/LoginActivity.java
+++ b/app/src/org/commcare/activities/LoginActivity.java
@@ -249,6 +249,8 @@ public class LoginActivity extends CommCareActivity<LoginActivity>
                             receiver.updateProgress(Localization.get("sync.progress.downloading"), DataPullTask.DATA_PULL_TASK_ID);
                         } else if (update[0] == DataPullTask.PROGRESS_DOWNLOADING) {
                             receiver.updateProgress(Localization.get("sync.process.downloading.progress", new String[]{String.valueOf(update[1])}), DataPullTask.DATA_PULL_TASK_ID);
+                        } else if (update[0] == DataPullTask.PROGRESS_DOWNLOADING_COMPLETE) {
+                            receiver.hideTaskCancelButton();
                         } else if (update[0] == DataPullTask.PROGRESS_PROCESSING) {
                             receiver.updateProgress(Localization.get("sync.process.processing", new String[]{String.valueOf(update[1]), String.valueOf(update[2])}), DataPullTask.DATA_PULL_TASK_ID);
                             receiver.updateProgressBar(update[1], update[2], DataPullTask.DATA_PULL_TASK_ID);
--- a/app/src/org/commcare/network/HttpRequestGenerator.java
+++ b/app/src/org/commcare/network/HttpRequestGenerator.java
@@ -26,10 +26,10 @@ import org.apache.http.protocol.BasicHttpContext;
 import org.apache.http.protocol.ExecutionContext;
 import org.apache.http.protocol.HttpContext;
 import org.commcare.CommCareApplication;
+import org.commcare.android.database.user.models.ACase;
 import org.commcare.cases.util.CaseDBUtils;
 import org.commcare.logging.AndroidLogger;
 import org.commcare.models.database.SqlStorage;
-import org.commcare.android.database.user.models.ACase;
 import org.commcare.utils.GlobalConstants;
 import org.javarosa.core.model.User;
 import org.javarosa.core.model.utils.DateUtils;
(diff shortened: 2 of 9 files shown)


app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -224,8 +232,13 @@ public void reportProgress(int progress) {
                         wasKeyLoggedIn = true;
                     }
 
-
                     this.publishProgress(PROGRESS_AUTHED, 0);
+                    if (isCancelled()) {
+                        // About to enter data commit phase; last chance to finish early if cancelled
+                        return PullTaskResult.UNKNOWN_FAILURE;
+                    }
+                    this.publishProgress(PROGRESS_DOWNLOADING_COMPLETE, 0);
```
phillipm: Sort of considered changing `PROGRESS_DOWNLOADING_COMPLETE` to `DISABLE_DIALOG_CANCEL` or something. Any opinions?

app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -249,6 +249,8 @@ protected void deliverUpdate(LoginActivity receiver, Integer... update) {
                             receiver.updateProgress(Localization.get("sync.progress.downloading"), DataPullTask.DATA_PULL_TASK_ID);
                         } else if (update[0] == DataPullTask.PROGRESS_DOWNLOADING) {
                             receiver.updateProgress(Localization.get("sync.process.downloading.progress", new String[]{String.valueOf(update[1])}), DataPullTask.DATA_PULL_TASK_ID);
+                        } else if (update[0] == DataPullTask.PROGRESS_DOWNLOADING_COMPLETE) {
+                            receiver.hideTaskCancelButton();
```
phillipm: Before this PR the login data pull task could be cancelled, but it wouldn't do anything: just wait for the normal code to finish. Now that we can abort http requests, cancelling is actually useful because it aborts the login http request faster. Also provides same guarantee as syncing that task won't be cancelled while data is being commited.

app/src/org/commcare/network/HttpRequestGenerator.java
```diff
@@ -160,10 +165,12 @@ public HttpResponse makeCaseFetchRequest(String baseUri, boolean includeStateFla
...
-        HttpGet request = new HttpGet(uri);
-        AndroidHttpClient.modifyRequestToAcceptGzipResponse(request);
-        addHeaders(request, syncToken);
-        return execute(client, request);
+        currentRequest = new HttpGet(uri);
+        AndroidHttpClient.modifyRequestToAcceptGzipResponse(currentRequest);
+        addHeaders(currentRequest, syncToken);
+        HttpResponse response = execute(client, currentRequest);
+        currentRequest = null;
+        return response;
```
phillipm: Only setting the `currentResponse` in `makeCaseFetchRequest`, because we don't care about aborting other types of http requests (yet)

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -204,6 +206,10 @@ public void reportProgress(int progress) {
                 //Either way, don't re-do this step
                 this.publishProgress(PROGRESS_CLEANED);
 
+                if (isCancelled()) {
+                    // avoid making the http request if user cancelled the task
+                    return PullTaskResult.UNKNOWN_FAILURE;
```
ctsims: Does this response just not get used? Seems like adding a cancelled result wouldn't be terrible, but if I recall this doesn't go anywhere if the task was cancelled?

app/src/org/commcare/tasks/DataPullTask.java
ctsims: I think it's fine to not specifically articulate what action should happen, v. knowing that downloading is complete disables the button

ctsims:
Looks good! One quick quesiton in there and then ready.

app/src/org/commcare/tasks/DataPullTask.java
phillipm: Yeah, results returned when cancelled get dropped. They don't need to be though, since later android APIs support passing values to `onCancelled`.
In this case I could return `null` or extend `PullTaskResult` to have a (unused) cancel result, both aren't perfect, but probably better than `UNKNOWN_FAILURE`. Follow-up suggestions?

PR #1230, opened 2016-04-26
Consumer apps config changes

amstone326:
cross-request: https://github.com/dimagi/commcare/pull/296


app/src/org/commcare/engine/resource/installers/SingleAppInstallation.java
```diff
@@ -74,7 +75,9 @@ protected void deliverResult(CommCareSetupActivity receiver,
                     @Override
                     protected void deliverUpdate(CommCareSetupActivity receiver,
                                                  int[]... update) {
-                        receiver.updateResourceProgress(update[0][0], update[0][1], update[0][2]);
+                        if (!CommCareApplication._().isConsumerApp()) {
```
amstone326: since we're trying to obfuscate away as much of the underlying process as possible for consumer apps, we don't want to change what the dialog says based on these updates. This will instead just leave up the generic "Initializing your application" message.

app/src/org/commcare/preferences/CommCarePreferences.java
```diff
@@ -474,4 +477,12 @@ private void startFileBrowser() {
                     Localization.get("no.file.browser"), false);
         }
     }
+
+    public static String getKeyServer() {
+        if (CommCareApplication._().isConsumerApp()) {
+            return null;
```
amstone326: so that we don't attempt to do any remote key management in a consumer app

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -85,30 +85,21 @@
...
-                         boolean restoreOldSession) {
+    public DataPullTask(String username, String password,
+                         String server, Context context, DataPullRequester dataPullRequester) {
         this.server = server;
         this.username = username;
         this.password = password;
         this.context = context;
         this.taskId = DATA_PULL_TASK_ID;
-        this.dataPullRequester = new DataPullResponseFactory();
-        this.restoreSession = restoreOldSession;
```
amstone326: this field was alway false, so removed

app/src/org/commcare/activities/MenuList.java
```diff
@@ -67,6 +71,11 @@ public String getActivityTitle() {
         return null;
     }
 
+    @Override
+    public boolean isBackEnabled() {
+        return !(CommCareApplication._().isConsumerApp() && isRootModuleMenu);
+    }
```
phillipm: Let's plan on refactoring `MenuGrid` and `MenuList` at some point to subclass the same parent, so this, and other code, doesn't need to be duplicated

app/src/org/commcare/engine/resource/AndroidResourceManager.java
```diff
@@ -65,11 +65,11 @@ public AndroidResourceManager(AndroidCommCarePlatform platform) {
      * @return UpdateStaged upon update download, UpToDate if no new update,
      * otherwise an error status.
      */
-    public AppInstallStatus checkAndPrepareUpgradeResources(String profileRef) {
+    public AppInstallStatus checkAndPrepareUpgradeResources(String profileRef, int profileAuthority) {
```
phillipm: Can you add `profileAuthority` to the method javadoc?

app/src/org/commcare/tasks/UpdateTask.java
```diff
@@ -236,6 +238,10 @@ public void setAsAutoUpdate() {
         wasTriggeredByAutoUpdate = true;
     }
 
+    public void setAsLocalUpdate() {
+        isLocalUpdate = true;
+    }
```
phillipm: Instead of having this setter, can you set `isLocalUpdate` in the constructor based on the value of `CommCareApplication._().isConsumerApp()` (and also make `isLocalUpdate` be a `final` field)

scripts/consumer_apps_builder.py
```diff
@@ -0,0 +1,48 @@
+
```
phillipm: Before finalizing this, can you add a comment describing at a high level what this script does?

app/res/layout/blank_screen_login.xml
```diff
@@ -0,0 +1,21 @@
...
+        android:visibility="gone" />
+
+    <EditText
+        android:id="@+id/edit_password"
+        android:layout_width="fill_parent"
+        android:layout_height="wrap_content"
+        android:inputType="textPassword"
+        android:visibility="gone" />
+
+</LinearLayout>
```
phillipm: Can you put a screenshot of what this blank login looks like in the PR description?

app/src/org/commcare/CommCareApplication.java
```diff
@@ -1476,4 +1476,8 @@ public boolean checkPendingBuildRefresh() {
         return false;
     }
 
+    public boolean isConsumerApp() {
```
phillipm: Can you add a comment describing what consumer apps are?

app/src/org/commcare/activities/CommCareHomeActivity.java
```diff
@@ -235,6 +248,38 @@ private void processFromLoginLaunch() throws SessionUnavailableException {
...
     }
 
+    private void checkForChangedLocalRestoreFile() {
+        try {
+            User loggedInUser = CommCareApplication._().getSession().getLoggedInUser();
+            if (!loggedInUser.getLastSyncToken().equals(getSyncTokenOfLocalRestoreFile())) {
+                formAndDataSyncer.performLocalRestore(this, loggedInUser.getUsername(), loggedInUser.getCachedPwd());
+            }
+        } catch (SessionUnavailableException sue) {
+            return;
```
phillipm: unneeded `return`

app/src/org/commcare/activities/CommCareHomeActivity.java
```diff
@@ -235,6 +248,38 @@ private void processFromLoginLaunch() throws SessionUnavailableException {
...
+        }
+    }
+
+    private String getSyncTokenOfLocalRestoreFile() {
```
phillipm: Can you move this and `checkForChangedLocalRestoreFile` into a helper class?

app/src/org/commcare/network/DebugDataPullResponseFactory.java
```diff
@@ -1,8 +1,5 @@
-package org.commcare.tasks.network;
+package org.commcare.network;
 
-import org.commcare.network.DataPullRequester;
-import org.commcare.network.HttpRequestGenerator;
-import org.commcare.network.RemoteDataPullResponse;
 import org.commcare.tasks.DataPullTask;
 
 import java.io.IOException;
```
phillipm: Now that this, and `DebugDataPullResponse` are being used in live code it would be good to remove mention of debugging in the name?

app/src/org/commcare/activities/DispatchActivity.java
```diff
@@ -66,6 +68,11 @@ protected void onCreate(Bundle savedInstanceState) {
         if (savedInstanceState != null) {
             shortcutExtraWasConsumed = savedInstanceState.getBoolean(EXTRA_CONSUMED_KEY);
         }
+
+        // Note: It is important that this check happens in onCreate, and not at the time that
+        // the value is actually used, because we only want to check for changed app files on a
+        // new creation of the DispatchActivity, not every time it resumes
+        shouldCheckForLocalAppFilesChange = CommCareApplication._().isConsumerApp();
```
phillipm: Does it matter if you rotate the screen? Would this trigger update checks more than wanted?

app/src/org/commcare/activities/DispatchActivity.java
```diff
@@ -105,12 +112,22 @@ protected void onSaveInstanceState(Bundle outState) {
         outState.putBoolean(EXTRA_CONSUMED_KEY, shortcutExtraWasConsumed);
     }
 
+    private void checkForChangedCCZ() {
+        Intent i = new Intent(getApplicationContext(), UpdateActivity.class);
+        startActivity(i);
```
phillipm: When this update activity is launched, can the user back out of it? Is that okay?

app/src/org/commcare/activities/FormAndDataSyncer.java
```diff
@@ -1,6 +1,7 @@
 package org.commcare.activities;
 
 import android.annotation.SuppressLint;
+import android.app.Activity;
```
phillipm: Unneeded

app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -45,8 +47,12 @@
 import org.commcare.views.notifications.NotificationMessage;
 import org.commcare.views.notifications.NotificationMessageFactory;
 import org.commcare.views.notifications.NotificationMessageFactory.StockMessages;
+import org.javarosa.core.reference.InvalidReferenceException;
+import org.javarosa.core.reference.ReferenceManager;
 import org.javarosa.core.services.locale.Localization;
+import org.kxml2.io.KXmlParser;
 
+import java.io.IOException;
```
phillipm: Unneeded

app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -245,7 +246,13 @@ protected void onResumeFragments() {
             return;
         }
 
-        tryAutoLogin();
+        if (CommCareApplication._().isConsumerApp()) {
+            uiController.setUsername("t1");
+            uiController.setPasswordOrPin("123");
+            localLoginOrPullAndLogin(false);
```
phillipm: Can you lift this into a helper method? Also, is there a reason for the `t1` / `123` values or is that something that is yet to be addressed?

app/src/org/commcare/tasks/UpdateTask.java
phillipm: Hmm, and instead of the `isLocalUpdate` field, you can have `authority` be a class field

app/src/org/commcare/activities/LoginActivity.java
phillipm: I see, `t1` / `123` are referenced in the build scripts. Not sure if it is worth it, but you can always inject them here from the build configs via `buildConfigField`

phillipm:
📦 📦 📦 📱 💵 👍 👍 Looks great!

app/src/org/commcare/activities/MenuList.java
amstone326: Agreed

app/src/org/commcare/tasks/UpdateTask.java
amstone326: I don't think I want to do it exactly like that because I was envisioning that in the future, we might use the local update functionality for something other than consumer apps, so I'm loathe to make those two things coupled. However, I agree that it makes more sense for `UpdateTask` to store an `authority` field directly, and can pass that in via the constructor instead of a separate setter method.

scripts/consumer_apps_builder.py
amstone326: done

app/res/layout/blank_screen_login.xml
amstone326: The entire startup process, which includes installing the .ccz (so we are technically on the `SetupActivity` screen) and restoring a user from the local restore file (so we are technically on the `LoginActivity` screen) looks like this
screenshot_20160512-131427

app/src/org/commcare/activities/DispatchActivity.java
amstone326: Good call, added a protection for that. One day I will catch rotation errors for myself :(

app/src/org/commcare/activities/DispatchActivity.java
amstone326: Good call to check, just tested it though and you can't (y)

app/src/org/commcare/activities/LoginActivity.java
amstone326: ..this is just me completely forgetting to do this properly so thank you for flagging

amstone326:
@phillipm Thanks for the quick review, made your proposed changes/fixes. Still have a couple of things left to do on this before it can be merged though.

app/src/org/commcare/network/HttpRequestEndpointsMock.java
```diff
@@ -31,7 +31,13 @@ public static void setCaseFetchResponseCodes(Integer[] responseCodes) {
 
     @Override
     public HttpResponse makeCaseFetchRequest(String baseUri, boolean includeStateFlags) throws ClientProtocolException, IOException {
```
amstone326: @phillipm I changed the implementation of this method to default to returning a response with result code 200 if `catchFetchResponseCodeStack` is empty, so that I could continue using this in the way I had been before without needing to call `setCaseFetchResponseCodes`, which feels wrong to do in live code. Let me know if you see any issue with this resolution.

amstone326:
@phillipm This should be good to go now! I'm definitely going to have some tweaks to make to the build script once I try to get it up and running on jenkins, but I think that's fine to do in a separate PR, and if this gets merged then I can set up the jenkins job to target master. You should just double check the note about resolving a merge conflict that I tagged you in.

app/src/org/commcare/engine/resource/AndroidResourceManager.java
```diff
@@ -141,7 +143,7 @@ private void loadProfileViaTemp(Resource upgradeProfile)
             UnresolvedResourceException,
             InstallCancelledException {
         tempUpgradeTable.destroy();
-        loadProfileIntoTable(tempUpgradeTable, profileRef);
+        loadRemoteAuthorityProfileIntoTable(tempUpgradeTable, profileRef);
```
phillipm: I think that this should be `loadProfileIntoTable(tempUpgradeTable, profileRef, authority);`. 
If CommCare is closed half way through a local update, then `upgradeTable.getResourceWithId(CommCarePlatform.APP_PROFILE_RESOURCE_ID);` (in calling method) won't be `null`, and hence `loadProfileViaTemp` will be called, meaning that we will try to update from a remote resource.

app/src/org/commcare/network/HttpRequestEndpointsMock.java
```diff
@@ -1,4 +1,4 @@
-package org.commcare.android.mocks;
+package org.commcare.network;
```
phillipm: Will you rename the class to exclude `Mock`? When I read 'mock' I assume I'm looking at non-prod code

phillipm: Err, maybe it is okay to leave it in... Since you are also using `HttpResponseMock`

app/src/org/commcare/utils/ConsumerAppsUtil.java
```diff
@@ -0,0 +1,54 @@
...
+            if (!loggedInUser.getLastSyncToken().equals(ConsumerAppsUtil.getSyncTokenOfLocalRestoreFile())) {
+                context.getFormAndDataSyncer().performLocalRestore(
+                        context, loggedInUser.getUsername(), loggedInUser.getCachedPwd());
+            }
+        } catch (SessionUnavailableException sue) {
+
+        }
+    }
+
+    public static String getSyncTokenOfLocalRestoreFile() {
```
phillipm: `private`

app/src/org/commcare/utils/ConsumerAppsUtil.java
```diff
@@ -0,0 +1,54 @@
+package org.commcare.utils;
+
+import org.commcare.CommCareApplication;
+import org.commcare.activities.CommCareHomeActivity;
+import org.commcare.activities.FormAndDataSyncer;
```
phillipm: unused

scripts/consumer_apps_builder.py
```diff
@@ -0,0 +1,103 @@
...
+	os.chdir('../')
+
+
+def build_apks_from_resources():
+	for (app_dir_name, sub_dir_list, files_list) in os.walk(PATH_TO_STATIC_RESOURCES_DIR):
+		if '.git' not in app_dir_name and app_dir_name != PATH_TO_STATIC_RESOURCES_DIR:
+			build_apk_from_directory_contents(app_dir_name, files_list)
+
+
+def build_apk_from_directory_contents(app_sub_dir, files_list):
```
phillipm: Can you replace the tabs in this file with whitespace?

app/src/org/commcare/engine/resource/AndroidResourceManager.java
amstone326: good call, fixed. I actually just got rid of the `loadRemoteAuthorityProfileIntoTable` method entirely since it was only being used in 1 other place.

scripts/consumer_apps_builder.py
```diff
@@ -0,0 +1,103 @@
...
+	return tree.getroot().get("name")
+
+
+def move_apk(app_id):
+	subprocess.call(["mkdir", "-p", "./build/outputs/consumer_apks"]) 
+	subprocess.call(["mv", "./build/outputs/apk/commcare-odk-standalone-debug.apk", "./build/outputs/consumer_apks/{}.apk".format(app_id)])
+
+
+checkout_or_update_static_resources_repo()
+build_apks_from_resources()
```
phillipm: Can you put these methods under a `if __name__ == "__main__":` conditional, so that you can import the file in python without running them?

Like https://github.com/dimagi/mobile-deploy/blob/master/checkout_cross_request_repo.py#L54-L55

scripts/consumer_apps_builder.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def build_apks_from_resources():
+	for (app_dir_name, sub_dir_list, files_list) in os.walk(PATH_TO_STATIC_RESOURCES_DIR):
+		if '.git' not in app_dir_name and app_dir_name != PATH_TO_STATIC_RESOURCES_DIR:
+			build_apk_from_directory_contents(app_dir_name, files_list)
+
+
+def build_apk_from_directory_contents(app_sub_dir, files_list):
+	full_path_to_config_file = os.path.join(app_sub_dir, files_list[0])
```
phillipm: For future PR: might want to somehow check that the first entry is indeed the config file

PR #1231, opened 2016-04-28
Fix double app icon

phillipm:
https://github.com/dimagi/commcare-odk/pull/1227 introduced the issue where 2 `CommCare` launcher icons show up in the Android app list:

screen

This fixes that.

Also, bump to gradle plugin to 2.1.0 (time to update to Android Studio 2.1 :) )
And update Google support libs from `23.2.1` to `23.3.0`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -51,10 +51,10 @@ dependencies {
     compile project(':commcare')
     compile fileTree(dir: 'app/libs', include: '*.jar')
     compile(name: 'htmlspanner-custom', ext: 'aar')
-    compile 'com.android.support:cardview-v7:23.2.1'
-    compile 'com.android.support:recyclerview-v7:23.2.1'
-    compile 'com.android.support:support-v4:23.2.1'
-    compile 'com.android.support:gridlayout-v7:23.2.1'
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -51,10 +51,10 @@ dependencies {
     compile project(':commcare')
     compile fileTree(dir: 'app/libs', include: '*.jar')
     compile(name: 'htmlspanner-custom', ext: 'aar')
-    compile 'com.android.support:cardview-v7:23.2.1'
-    compile 'com.android.support:recyclerview-v7:23.2.1'
-    compile 'com.android.support:support-v4:23.2.1'
-    compile 'com.android.support:gridlayout-v7:23.2.1'
+    compile 'com.android.support:cardview-v7:23.3.0'
```
amstone326: @phillipm Since these have been finicky in the past, do you think we need to test that everything looks fine with the new versions?

phillipm: We should be good: in the [changelog](https://developer.android.com/tools/support-library/index.html) it looks like they are just bug fixes. I don't remember having issues updating the support lib versions in the past.

PR #1233, opened 2016-04-29
Fix grid widget warning msg being off-screen

amstone326:
Fix for http://manage.dimagi.com/default.asp?223603. 

This definitely fixes the problem, but I'm a little concerned by what it might break, since I can't figure out what the reason for having that code in there was in the first place. I don't think this was actually written by any of us, so I'm inclined to make the change unless anyone knows its purpose. 

Before: 
screenshot_20160429-172222

After:
screenshot_20160429-171221

### DIFF
--- a/app/src/org/commcare/views/widgets/GridWidget.java
+++ b/app/src/org/commcare/views/widgets/GridWidget.java
@@ -43,7 +43,7 @@ public class GridWidget extends QuestionWidget {
     // The possible select choices
     private final String[] choices;
 
-    // The Gridview that will hol the icons
+    // The Gridview that will hold the icons
     private final GridView gridview;
 
     // Defines which icon is selected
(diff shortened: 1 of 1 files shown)


ctsims:
I agree with your concern about the view. 

It is likely laid out as a fixed size due to odd implementations of how it grows. I think we should add the grid view to our regression test plan and come up with a representative usage (IE: a 3x3 grid of images or something) that fits our expectation of how it is used so we can have a gold standard to compare changes against.

Maybe Jeremy has an example that could work for that? I think he uses it quite a bit.

amstone326:
@ctsims Added a question to the QA app and discovered that this does cause an issue if you want your icons to go onto a 2nd row. Really struggling with what's causing it though, would be good to show you when you have a few minutes.

PR #1234, opened 2016-04-29
Resume inline video where it left off after rotation

amstone326:
Fix for http://manage.dimagi.com/default.asp?224767

### DIFF
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -43,6 +43,7 @@ import android.widget.ImageButton;
 import android.widget.RelativeLayout;
 import android.widget.TextView;
 import android.widget.Toast;
+import android.widget.VideoView;
 
 import org.commcare.CommCareApplication;
 import org.commcare.activities.components.FormFileSystemHelpers;
@@ -53,6 +54,7 @@ import org.commcare.activities.components.FormRelevancyUpdating;
 import org.commcare.activities.components.ImageCaptureProcessing;
 import org.commcare.dalvik.BuildConfig;
 import org.commcare.dalvik.R;
+import org.commcare.views.media.MediaLayout;
 import org.odk.collect.android.jr.extensions.IntentCallout;
 import org.odk.collect.android.jr.extensions.PollSensorAction;
 import org.commcare.interfaces.AdvanceToNextListener;
@@ -172,6 +174,8 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
     private static final String KEY_HAS_SAVED = "org.odk.collect.form.has.saved";
     public static final String KEY_FORM_ENTRY_SESSION = "form_entry_session";
     public static final String KEY_RECORD_FORM_ENTRY_SESSION = "record_form_entry_session";
+    private static final String KEY_WIDGET_WITH_VIDEO_PLAYING = "index-of-widget-with-video-playing-on-pause";
+    private static final String KEY_POSITION_OF_VIDEO_PLAYING = "position-of-video-playing-on-pause";
 
     /**
      * Intent extra flag to track if this form is an archive. Used to trigger
@@ -222,6 +226,9 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
     // used to limit forward/backward swipes to one per question
     private boolean isAnimatingSwipe;
     private boolean isDialogShowing;
+
+    private int indexOfWidgetWithVideoPlaying = -1;
+    private int positionOfVideoProgress = -1;
 
     private FormLoaderTask<FormEntryActivity> mFormLoaderTask;
     private SaveToDiskTask mSaveToDiskTask;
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -1533,6 +1545,41 @@ protected void onPause() {
...
+
+    private void restoreInlineVideoState() {
+        if (indexOfWidgetWithVideoPlaying != -1) {
+            QuestionWidget widgetWithVideoToResume = questionsView.getWidgets().get(indexOfWidgetWithVideoPlaying);
+            VideoView inlineVideo = (VideoView)widgetWithVideoToResume.findViewById(MediaLayout.INLINE_VIDEO_PANE_ID);
+            if (inlineVideo != null) {
+                inlineVideo.seekTo(positionOfVideoProgress);
+                inlineVideo.start();
+            } else {
+                Logger.log(AndroidLogger.SOFT_ASSERT,
```
amstone326: I don't actually foresee how this could happen, but wanted to prevent the NPE/log it just in case.

app/src/org/commcare/views/media/MediaLayout.java
```diff
@@ -40,6 +41,9 @@
 public class MediaLayout extends RelativeLayout {
     private static final String t = "AVTLayout";
 
+    @IdRes
```
phillipm: :+1:

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -1533,6 +1545,41 @@ protected void onPause() {
...
+                    positionOfVideoProgress = inlineVideo.getCurrentPosition();
+                    return;
+                }
+            }
+        }
+    }
+
+    private void restoreInlineVideoState() {
+        if (indexOfWidgetWithVideoPlaying != -1) {
+            QuestionWidget widgetWithVideoToResume = questionsView.getWidgets().get(indexOfWidgetWithVideoPlaying);
```
phillipm: What happens if you play a move, rotate, move to the next question screen, and rotate again? I feel like this `get` would potentially fail because you would try to restore the video using a question index that potentially doesn't exist.

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -1533,6 +1545,41 @@ protected void onPause() {
...
+            VideoView inlineVideo = (VideoView)widgetWithVideoToResume.findViewById(MediaLayout.INLINE_VIDEO_PANE_ID);
```
phillipm: What happens if there are two video widgets on one screen? They both use the same id, so how do we know which one to start?

app/src/org/commcare/activities/FormEntryActivity.java
amstone326: Hm, not following. At that point `indexOfWidgetwithVideoPlaying` has been reset to -1 by the first call to `restoreInlineVideoState()`, so this doesn't get called.

app/src/org/commcare/activities/FormEntryActivity.java
amstone326: Also confused, that's the point of having and using the `indexOfWidgetWithVideoPlaying` field

app/src/org/commcare/activities/FormEntryActivity.java
phillipm: Ah, gotcha, the second rotate will call `saveInlineVideoState` before trying to restore!

app/src/org/commcare/activities/FormEntryActivity.java
phillipm: I'm was wondering about the behavior of `findViewById` when two views have the same id

app/src/org/commcare/activities/FormEntryActivity.java
amstone326: It will, but I don't think that's relevant...What matters is just that `restoreInlineVideoState` sets the value of `indexOfWidgetWithVideoPlaying` back to -1 (line 1580 below) after restoring. (Not trying to nitpick, just want to make sure I'm understanding your concern and that it's not actually a problem).

app/src/org/commcare/activities/FormEntryActivity.java
amstone326: That can't happen, a single `QuestionWidget` view can only ever have 1 inline video view added (walking through the code in `MediaLayout` a bit should convince you of this if you want to double check).

app/src/org/commcare/activities/FormEntryActivity.java
phillipm: Yeah, sorry, my exclamation was me understanding how the code (correctly) worked. Looks good

app/src/org/commcare/activities/FormEntryActivity.java
phillipm: :+1:

amstone326:
Release Note: Inline videos resume where they left off after the screen is rotated

PR #1235, opened 2016-04-29
Fix cutting off of images in GridWidget 

amstone326:
Fixes issue caused by https://github.com/dimagi/commcare-odk/pull/1233, while retaining the fix that that PR was for.

Since a GridView is intended to be a scrolling container, rather than to wrap all of its contents, we have to explicitly set the height of the GridView based upon how much space we need to fit all of the images.

### DIFF
--- a/app/src/org/commcare/views/widgets/GridWidget.java
+++ b/app/src/org/commcare/views/widgets/GridWidget.java
@@ -153,7 +153,6 @@ public class GridWidget extends QuestionWidget {
         Display display = ((WindowManager)getContext().getSystemService(Context.WINDOW_SERVICE))
                 .getDefaultDisplay();
         int screenWidth = display.getWidth();
-        int screenHeight = display.getHeight();
 
         // Use the user's choice for num columns, otherwise decide based upon what will fit.
         int maxColumnsThatWillFit = screenWidth / maxImageWidth;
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/views/widgets/GridWidget.java
```diff
@@ -78,7 +79,8 @@ public GridWidget(Context context, FormEntryPrompt prompt,
         // The max width of an icon in a given column. Used to line
```
wpride: max width and height

app/src/org/commcare/views/widgets/GridWidget.java
```diff
@@ -87,7 +89,7 @@ public GridWidget(Context context, FormEntryPrompt prompt,
         // Build view
         for (int i = 0; i < mItems.size(); i++) {
             SelectChoice sc = mItems.get(i);
-            // Read the image sizes and set maxColumnWidth. This allows us to make sure all of our
+            // Read the image sizes and set maxImageWidth. This allows us to make sure all of our
```
wpride: "and set maxImageWidth and maxImageHeight"

app/src/org/commcare/views/widgets/GridWidget.java
amstone326: the comment is just referring to the width though (the fact that it's used to line up and auto-fit the columns)

app/src/org/commcare/views/widgets/GridWidget.java
amstone326: makes sense to change here though

app/src/org/commcare/views/widgets/GridWidget.java
wpride: Ok then separate comment for the height maybe? At a high level they're both used to configure the GridView appearance, right?

amstone326:
Comments added

PR #1238, opened 2016-05-02
Make alert dialogs persist across orientation change

amstone326:
Fix for http://manage.dimagi.com/default.asp?222011.

Turns out that a lot of the infrastructure was already there for this via some combination of past work by @phillipm and I, but seems like we got our wires a little crossed at some point, and just weren't using the existing infrastructure everywhere we should have been. The implementations of the `showAlertDialog()` and `showPendingAlertDialog()` methods (from the `DialogController` interface) in `CommCareActivity` automatically handle persisting alert dialogs on orientation change. The problem was just that in many places, we were still showing dialogs via the `AlertDialogFactory`'s `showDialog()` method, instead of `CommCareActivity`'s `showAlertDialog()` method.

So, changes made by this PR:

1. Use `showAlertDialog()` to show dialogs wherever possible (i.e. for any dialogs whose context is `CommCareActivity`. For dialogs launched in a normal activity, I left the old show methods in use. We could talk about having an implementation of `DialogController` in non-CommCareActivities at some point, but I didn't want to do that in this PR, since it's already pretty large).

2. Previously, the `DialogController` alert dialog infrastructure was only set up to handle dialogs generated from an `AlertDialogFactory`. This meant that `PaneledChoiceDialog`s and a couple of other custom dialogs (including the About CommCare dialog, which was the actual originator of this ticket) could not be shown that way. In order to expand the utility of this infrastructure, I created an abstract class called `CommCareAlertDialog`, from which an `AlertDialogFragment` can be constructed (which is the class of dialogs that `DialogController` knows how to manage), and then made `StandardAlertDialog` (renaming of `AlertDialogFactory`), `PaneledChoiceDialog`, and `CustomViewDialog` (a new class for dialogs like the About CommCare dialog) all subclass that.

### DIFF
--- a/app/src/org/commcare/views/dialogs/CommCareAlertDialog.java
+++ b/app/src/org/commcare/views/dialogs/CommCareAlertDialog.java
@@ -1,0 +1,62 @@
+package org.commcare.views.dialogs;
+
+import android.app.AlertDialog;
+import android.content.DialogInterface;
+import android.view.View;
+
+/**
+ * Created by amstone326 on 4/30/16.
+ */
+public abstract class CommCareAlertDialog {
+
+    protected AlertDialog dialog;
+    protected DialogInterface.OnCancelListener cancelListener;
+    protected DialogInterface.OnDismissListener dismissListener;
+
+    protected boolean isCancelable;
+    protected View view;
+
+    public void finalizeView() {
+        dialog.setCancelable(isCancelable);
+        dialog.setView(view);
+    }
+
+    public void performCancel(DialogInterface dialog) {
+        if (cancelListener != null) {
+            cancelListener.onCancel(dialog);
+        }
+    }
+
+    public void performDismiss(DialogInterface dialog) {
+        if (dismissListener != null) {
+            dismissListener.onDismiss(dialog);
+        }
+    }
+
+    public boolean isCancelable() {
+        return isCancelable;
+    }
+
+    public void setOnCancelListener(DialogInterface.OnCancelListener listener) {
+        isCancelable = true;
+        dialog.setOnCancelListener(listener);
+        cancelListener = listener;
+    }
+
+    public void setOnDismissListener(DialogInterface.OnDismissListener listener) {
+        dialog.setOnDismissListener(listener);
+        dismissListener = listener;
+    }
+
+    public AlertDialog getDialog() {
+        return dialog;
+    }
+
+    // IMPORTANT: Should ONLY be used to show dialogs in activities that are NOT CommCareActivities
+    // (and therefore cannot use the DialogController infrastructure set up there)
+    public void showDialog() {
+        finalizeView();
+        dialog.show();
+    }
+
+}
--- a/app/src/org/commcare/views/dialogs/CommCareAlertDialog.java
+++ b/app/src/org/commcare/views/dialogs/CommCareAlertDialog.java
@@ -13,6 +13,8 @@ public abstract class CommCareAlertDialog {
     protected DialogInterface.OnCancelListener cancelListener;
     protected DialogInterface.OnDismissListener dismissListener;
 
+    // false by default, can be overriden by calling setOnCancelListener(), or by subclass
+    // definitions if desired
     protected boolean isCancelable;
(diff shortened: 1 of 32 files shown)


app/src/org/commcare/views/dialogs/CommCareAlertDialog.java
```diff
@@ -0,0 +1,69 @@
...
+    public void performCancel(DialogInterface dialog) {
+        if (cancelListener != null) {
+            cancelListener.onCancel(dialog);
+        }
+    }
+
+    public void performDismiss(DialogInterface dialog) {
+        if (dismissListener != null) {
+            dismissListener.onDismiss(dialog);
+        }
```
phillipm: I feel like there was a reason, but why don't we just call `dialog.dismiss()` here?

amstone326: I actually don't know; I was just imitating the `performCancel` method you had added to `AlertDialogFactory` a while ago

app/src/org/commcare/views/dialogs/CommCareAlertDialog.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    public boolean isCancelable() {
+        return isCancelable;
+    }
+
+    public void setOnCancelListener(DialogInterface.OnCancelListener listener) {
+        isCancelable = true;
+        dialog.setOnCancelListener(listener);
```
phillipm: Might be nice to move `dialog.setOnCancelListener(cancelListener)` to `finalizeView`, then you don't need to override `setOnCancelListener` in `CustomViewAlertDialog`. Shouldn't need a null check, looking at the implementation of `setOnCancelListener`.

app/src/org/commcare/views/dialogs/CommCareAlertDialog.java
```diff
@@ -0,0 +1,69 @@
...
+ * The showing and persistence of all CommCareAlertDialogs should be managed by the implementation
+ * of the DialogController interface in CommCareActivity wherever possible (e.g. wherever the
+ * context of a dialog is a CommCareActivity)
+ *
+ */
+public abstract class CommCareAlertDialog {
+
+    protected AlertDialog dialog;
+    protected DialogInterface.OnCancelListener cancelListener;
+    protected DialogInterface.OnDismissListener dismissListener;
```
phillipm: can be `private`

app/src/org/commcare/views/dialogs/CommCareAlertDialog.java
phillipm: hmm, https://github.com/dimagi/commcare-odk/pull/745

app/src/org/commcare/views/dialogs/CommCareAlertDialog.java
```diff
@@ -0,0 +1,69 @@
...
+        dismissListener = listener;
+    }
+
+    public AlertDialog getDialog() {
+        return dialog;
+    }
+
+    // IMPORTANT: Should ONLY be used to show dialogs in activities that are NOT CommCareActivities
+    // (and therefore cannot use the DialogController infrastructure set up there)
+    public void showDialog() {
```
phillipm: Though a long name, it might be helpful to rename this to `showNonPersistentDialog`

phillipm:
Awesome, thanks for doing this refactor! Some minor comments in `CommCareAlertDialog`, otherwise good to go

app/src/org/commcare/views/dialogs/CommCareAlertDialog.java
amstone326: Oo yea no I like that suggestion a lot, I was uncomfortable with the fact that someone could still end up using this accidentally fairly easily.

app/src/org/commcare/views/dialogs/CommCareAlertDialog.java
amstone326: Yea good call

amstone326:
@phillipm Updated!

PR #1240, opened 2016-05-03
Handle an XPathException in SessionNavigator properly

amstone326:
While http://manage.dimagi.com/default.asp?225812 turned out to be an HQ issue at its root, it revealed some issues (one very large and one small) with how we were handling an XPathException thrown by `handleCompute()` in `SessionNavigator`. This PR resolves those 2 issues (comments in-line on the PR as to what was fixed).

cross-request: https://github.com/dimagi/commcare/pull/297

### DIFF
--- a/app/src/org/commcare/activities/CommCareActivity.java
+++ b/app/src/org/commcare/activities/CommCareActivity.java
@@ -376,14 +376,7 @@ public abstract class CommCareActivity<R> extends FragmentActivity
         }
     }
 
-    /**
-     * Display exception details as a pop-up to the user.
-     *
-     * @param e Exception to handle
-     */
-    public void displayException(Exception e) {
-        String title =  Localization.get("notification.case.predicate.title");
-        String message = Localization.get("notification.case.predicate.action", new String[]{e.getMessage()});
+    public void displayException(String title, String message) {
         DialogInterface.OnClickListener listener = new DialogInterface.OnClickListener() {
             @Override
             public void onClick(DialogInterface dialog, int i) {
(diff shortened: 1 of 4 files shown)


app/src/org/commcare/activities/CommCareActivity.java
```diff
@@ -376,14 +376,7 @@ public void stopTaskTransition() {
...
 
-    /**
-     * Display exception details as a pop-up to the user.
-     *
-     * @param e Exception to handle
-     */
-    public void displayException(Exception e) {
-        String title =  Localization.get("notification.case.predicate.title");
-        String message = Localization.get("notification.case.predicate.action", new String[]{e.getMessage()});
+    public void displayException(String title, String message) {
```
amstone326: This method was previously showing a very specific type of error, even though it had the generic name `displayException`. The new method below, `displayCaseListFilterException`, now has the behavior that `displayException` had previously, and this method is now appropriately generic.

app/src/org/commcare/activities/CommCareHomeActivity.java
```diff
@@ -821,8 +821,9 @@ public void processSessionResponse(int statusCode) {
             case SessionNavigator.LAUNCH_CONFIRM_DETAIL:
                 launchConfirmDetail(asw);
                 break;
-            case SessionNavigator.EXCEPTION_THROWN:
-                displayException(sessionNavigator.getCurrentException());
```
amstone326: It did not make sense to be calling this `displayException` method for this error, since it was showing a very specific error message that was unrelated. Instead, since this is an xpath error, handle it as a user-facing error the way we do for other exceptions of this type.

PR #1245, opened 2016-05-05
Launch auto-update using parallel thread pool

phillipm:
This will need to make it into whatever apk ICDS uses, so maybe in the next 2.27 hotfix or the 'CommCare Conservative' build

Completes the work started in https://github.com/dimagi/commcare-odk/pull/1205


amstone326:
@phillipm Shouldn't this be targeting 2.27 instead of master?

phillipm:
It is unclear if we are doing a 2.27 hotfix or making a new (conservative) CommCare build for ICDS, so in the meantime I decided to target master.
We can cherry pick the commit into either of those branches when needed.

PR #1250, opened 2016-05-11
Allow for copy & paste of navigation and form entry sessions

phillipm:
screen

Allow user to edit a form entry session and copy out the navigation and form entry sessions. 
This is really rough around the edges right now because navigation sessions are serialized to non-human readable text, so the you can't modify that part of a session. Regardless, it will still be useful for reproducing bug reports.

Eventually we can make navigation sessions serialize to human readable strings so they can be read/modified by hand.

First commit is xml auto-formatting


app/src/org/commcare/preferences/DevSessionRestorer.java
```diff
@@ -153,6 +153,9 @@ public static void clearSession() {
     }
 
     private static void clearSession(SharedPreferences prefs) {
-        prefs.edit().remove(CommCarePreferences.CURRENT_SESSION).commit();
+        prefs.edit()
+                .remove(CommCarePreferences.CURRENT_SESSION)
+                .remove(CommCarePreferences.CURRENT_FORM_ENTRY_SESSION)
```
phillipm: Clear form entry session when session saving is disabled

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -1722,6 +1722,7 @@ private void handleFormLoadCompletion(FormController fc) {
         try {
             FormEntrySessionReplayer.tryReplayingFormEntry(mFormController.getFormEntryController(),
                     formEntryRestoreSession);
+            formEntryRestoreSession = null;
```
phillipm: Fixes bug where if you rotate the screen after restoring a form replay session it would crash commcare due to trying to replay an empty session

amstone326:
What do you mean by "navigation sessions" in the PR description?

phillipm:
Session saving is split up between form entry sessions and pre-form entry sessions. I was thinking of pre-form entry session state as the navigation needed to reach the form, hence me calling it 'navigation session'

app/src/org/commcare/preferences/DeveloperPreferences.java
```diff
@@ -116,6 +143,27 @@ public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, Strin
         }
     }
 
+    private static String getSessionStateFromPrefs() {
```
amstone326: would prefer this to be called something like `getSavedSessionStateAsString()`

app/res/xml/preferences_developer.xml
```diff
@@ -45,6 +43,11 @@
         android:entryValues="@array/pref_enabled_vals"
         android:key="cc-enable-session-saving"
         android:title="Enable Session Saving"/>
+    <EditTextPreference
```
amstone326: What do you think about having this only be visible if session saving is enabled?

amstone326: @phillipm Pinging on this question, and also the build is still failing, but when I go to click on 'Details', it says the build is not found. Any idea what's up?

phillipm: @amstone326 Updated the PR to remove this preference when session saving isn't enabled

PR #1267, opened 2016-05-18
Checking if detail callout is null before using

amstone326:
Fix null pointers introduced by https://github.com/dimagi/commcare-odk/pull/1119.

@phillipm I'm gonna have someone merge this now because I need master to not be crashing in order to work on consumer apps stuff, but want to make sure with you that this is an OK resolution of this.

### DIFF
--- a/app/src/org/commcare/adapters/EntityListAdapter.java
+++ b/app/src/org/commcare/adapters/EntityListAdapter.java
@@ -261,7 +261,9 @@ public class EntityListAdapter implements ListAdapter {
                     getCalloutDataForEntity(entity));
         } else {
             emv.setSearchTerms(currentSearchTerms);
-            emv.setExtraData(detail.getCallout().getResponseDetailField(), getCalloutDataForEntity(entity));
+            if (detail.getCallout() != null) {
+                emv.setExtraData(detail.getCallout().getResponseDetailField(), getCalloutDataForEntity(entity));
+            }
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/adapters/EntityListAdapter.java
```diff
@@ -261,7 +261,9 @@ private View getEntityView(Entity<TreeReference> entity, EntityView emv, int pos
                     getCalloutDataForEntity(entity));
         } else {
             emv.setSearchTerms(currentSearchTerms);
-            emv.setExtraData(detail.getCallout().getResponseDetailField(), getCalloutDataForEntity(entity));
+            if (detail.getCallout() != null) {
```
phillipm: probably also need to check if `detail.getCallout().getResponseDetailField()` is `null`

app/src/org/commcare/views/EntityView.java
```diff
@@ -85,7 +85,9 @@ private EntityView(Context context, Detail d, Entity e,
             views.add(addCell(col, field, forms.get(col), mHints.get(col), sortField, -1, true));
         }
 
-        addExtraData(d.getCallout().getResponseDetailField(), extraData);
+        if (d.getCallout() != null) {
```
phillipm: Same as above

phillipm:
Sorry about that, I actually run into this this morning as weel

PR #1270, opened 2016-05-18
Make SessionUnavailableException a RuntimeException

phillipm:
Make `SessionUnavailableException` extend `RuntimeException` because it is usually just caught and rethrown as a RuntimeException. In addition, there is generally no way to recover from the session ending and storage being closed.

We could make `CommCareExceptionHandler` restart CommCare when it handles `SessionUnavailableException` throws, such that it will look like the session just ended to the user and they were auto-logged out. Only caveat being that they will still see the "CommCare has unexpectedly closed" dialog.
I'm not doing this because I want to still report these crashes to ACRA, which requires calling the parent handler. In the future if we feel like we have handled session unavailable crashes to the point that we don't need to see the crash reports, we can use this pattern


app/src/org/commcare/activities/DispatchActivity.java
```diff
@@ -140,34 +140,33 @@ private void dispatch() {
...
-                } else {
-                    launchHomeScreen();
-                }
+                CommCareApplication._().getSession();
             } catch (SessionUnavailableException sue) {
                 launchLoginScreen();
             }
+
+            ApplicationRecord currentRecord = currentApp.getAppRecord();
+            if (currentApp.getAppResourceState() == CommCareApplication.STATE_CORRUPTED) {
```
amstone326: Hm, so when I originally wrote this, I intentionally ordered it to check for these 2 states (if the app resource state is corrupted, or the currently seated app is not usable), _before_ checking for session expiration, so that you couldn't end up on the login screen with a corrupted or unusable app as the seated app. I don't have hard evidence that this is necessary, but I think we probably want to still preserve that behavior.

amstone326: Oo nevermind, I was reviewing commit-by-commit, looks like you reverted this change

app/src/org/commcare/engine/cases/CaseUtils.java
```diff
@@ -64,11 +64,7 @@ public static void purgeCases() {
         }
 
         SQLiteDatabase db;
-        try {
-            db = CommCareApplication._().getUserDbHandle();
-        } catch (SessionUnavailableException e) {
-            throw new UserStorageClosedException(e.getMessage());
```
amstone326: Hm so this actually changes what exception is thrown; Are you sure that this `UserStorageClosedException` isn't caught and used to trigger specific behavior anywhere?

app/src/org/commcare/models/AndroidSessionWrapper.java
```diff
@@ -163,13 +163,7 @@ public void commitStub() {
         SqlStorage<FormRecord> storage = CommCareApplication._().getUserStorage(FormRecord.class);
         SqlStorage<SessionStateDescriptor> sessionStorage = CommCareApplication._().getUserStorage(SessionStateDescriptor.class);
 
-        SecretKey key;
-        try {
-            key = CommCareApplication._().createNewSymmetricKey();
-        } catch (SessionUnavailableException e) {
-            // the user db is closed
-            throw new UserStorageClosedException(e.getMessage());
```
amstone326: same comment as above

app/src/org/commcare/models/AsyncNodeEntityFactory.java
```diff
@@ -105,12 +105,7 @@ private void primeCache() {
 
         long now = System.currentTimeMillis();
 
-        SQLiteDatabase db;
-        try {
-            db = CommCareApplication._().getUserDbHandle();
-        } catch (SessionUnavailableException e) {
-            throw new UserStorageClosedException(e.getMessage());
```
amstone326: same

amstone326: (and several more instances below, won't keep commenting on all of them)

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -441,13 +436,6 @@ private void storeSuccessfulSyncTime(SharedPreferences prefs) {
         } catch (StorageFullException e) {
             e.printStackTrace();
             failureReason = e.getMessage();
-        }
-
-        //These last two aren't a sign that the incoming data is bad, but
-        //we still can't recover from them usefully
-        catch (SessionUnavailableException e) {
-            e.printStackTrace();
```
amstone326: Are you sure it's ok that a `SessionUnavailableException` in this context no longer triggers the logging and return statements below (lines 451-452)?

app/src/org/commcare/tasks/DumpTask.java
```diff
@@ -237,7 +236,8 @@ protected Boolean doTaskBackground(String... params) {
                             }
                         }
                     } catch(SessionUnavailableException sue) {
-                        throw sue;
```
amstone326: Cannot figure out what the logic going on here was haha

app/src/org/commcare/tasks/LogSubmissionTask.java
```diff
@@ -140,13 +140,7 @@ private boolean serializeLogs(SqlStorage<DeviceReportRecord> storage) {
         //update the last recorded record
         settings.edit().putLong(CommCarePreferences.LOG_LAST_DAILY_SUBMIT, new Date().getTime()).commit();
 
-        DeviceReportRecord record;
-        try {
-            record = DeviceReportRecord.generateNewRecordStub();
-        } catch (SessionUnavailableException e) {
-            Logger.log(AndroidLogger.TYPE_MAINTENANCE, "User database closed while trying to submit");
```
amstone326: Also losing this log statement

app/src/org/commcare/utils/CommCareUtil.java
```diff
@@ -90,15 +90,11 @@ public static void triggerLogSubmission(Context c) {
...
             Toast.makeText(c, "Couldn't submit logs! Invalid submission URL...", Toast.LENGTH_LONG).show();
         } else {
-            try {
-                LogSubmissionTask reportSubmitter =
-                        new LogSubmissionTask(true,
-                                CommCareApplication._().getSession().startDataSubmissionListener(R.string.submission_logs_title),
-                                url);
-                reportSubmitter.execute();
-            } catch (SessionUnavailableException e) {
-                Toast.makeText(c, "Couldn't submit logs! No longer logged in", Toast.LENGTH_LONG).show();
```
amstone326: losing this notification

amstone326:
@phillipm Am I understanding properly that CommCare will now just crash outright whenever a `SessionUnavailableException` is thrown? I realize that that was already happening in some places before this PR, but if we're going to refactor anyway, it seems a lot better to make it so that the user is redirected to the login screen with a message to notify them that their session expired. It sounds like you were explaining why that was hard to do in the PR description, but I'm not really following why it's not possible.

app/src/org/commcare/tasks/DataPullTask.java
phillipm: Good call, there is chance it occurs when trying to pull data on login. I'll add it back

app/src/org/commcare/tasks/DumpTask.java
phillipm: It makes sure that `SessionUnavailableException` are not caught by the `catch (Exception e)` statement below, and instead handled with the logic I've just moved into the first catch

app/src/org/commcare/utils/CommCareUtil.java
phillipm: Don't think this message would be useful for a user to see.

app/src/org/commcare/engine/cases/CaseUtils.java
phillipm: I've removed `UserStorageClosedException` because it was just a `RuntimeException` front for `SessionUnavailableException`

phillipm:
@amstone326 Yes, since it is hard to catch all `SessionUnavailableException` throws at the top level (we would need to wrap every `Activity` entry point in a `try`/`catch`) I think that crashing is fine, and generally what we currently do.

When this is released we can monitor ACRA and identify specific points where we might be able to handle an session exception and redirect to the login screen, such as https://github.com/dimagi/commcare-odk/pull/1252 does. Not the cleanest way forward, but isn't deviated much from current behavior

app/src/org/commcare/engine/cases/CaseUtils.java
amstone326: 👍

app/src/org/commcare/tasks/DumpTask.java
amstone326: Oh yea sorry, totally understood what you did, just confused as to the point of what was going on before.

PR #1271, opened 2016-05-18
Fix for Busted Menu Audio Layout

ctsims:
The previous audio button behavior has been broken for what appears to be a long time. The audio buttons would drift off the screen if menu text got too long.

Fix for: http://manage.dimagi.com/default.asp?225035

Implementing the new Audio UI to Biyuen's spec was surprisingly not well supported, had to implement a custom layout rule. 

NOTE: I'm going to figure out whether this needs to make it into 2.27, will probably splice in the change if so.

Examples of the new behavior for short and long text, along with the broken behavior, and the pre-new-ui behavior (far right) that is really awkward on larger screens:
image

### DIFF
--- a/app/src/org/commcare/views/ShrinkingLinearLayout.java
+++ b/app/src/org/commcare/views/ShrinkingLinearLayout.java
@@ -1,0 +1,87 @@
+package org.commcare.views;
+
+import android.annotation.TargetApi;
+import android.content.Context;
+import android.content.res.TypedArray;
+import android.os.Build;
+import android.text.Layout;
+import android.util.AttributeSet;
+import android.view.Gravity;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.LinearLayout;
+
+import org.commcare.dalvik.R;
+import org.commcare.utils.AndroidUtil;
+
+/**
+ * A shrinking linear layout is capable of denoting one child view which can change its size
+ * if the full space is not needed.
+ *
+ * Normally a linear layout requires either fully laying out (IE: Pre-determined static sizes or
+ * ratios) all of its children _or_ accepting that if the linear layout grows larger than its parent
+ * some views will be cut off.
+ *
+ * When designing a shrinking layout, the view can be laid out fully "expanded" (IE: All of the
+ * child views are the largest they will ever be), then one view can be chosen to shrink to its
+ * "natural" width if it doesn't need the full space. This allows for things like text fields with
+ * an element next to the text that can grow horizontally _until_ running into the fully "expanded"
+ * layout.
+ *
+ * Limitations:
+ * This layout is incrementally slower than a traditional linear layout.
+ * The view will currently only shrink horizontally
+ *
+ * Created by ctsims on 5/18/2016.
+ */
+public class ShrinkingLinearLayout extends LinearLayout {
+    int shrinkingViewId = -1;
+
+    public ShrinkingLinearLayout(Context context, AttributeSet attrs) {
+        super(context, attrs);
+        loadViewConfig(context, attrs);
+    }
+
+    private void loadViewConfig(Context context, AttributeSet attrs) {
+        TypedArray typedArray = context.obtainStyledAttributes(attrs, R.styleable.ShrinkingLinearLayout);
+        shrinkingViewId = typedArray.getResourceId(R.styleable.ShrinkingLinearLayout_shrinkable_view, -1);
+    }
+
+
+    @Override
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/views/ShrinkingLinearLayout.java
```diff
@@ -0,0 +1,81 @@
...
+        TypedArray typedArray = context.obtainStyledAttributes(attrs, R.styleable.ShrinkingLinearLayout);
+        shrinkingViewId = typedArray.getResourceId(R.styleable.ShrinkingLinearLayout_shrinkable_view, -1);
+    }
+
+
+    @Override
+    protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
+        View dynamicView = this.findViewById(shrinkingViewId);
+
+        if(shrinkingViewId == -1 || dynamicView == null) {
```
amstone326: can you put a space between `if` and `(` (here and below)

amstone326:
@ctsims Hey so I'm confused as to what part of this PR is actually fixing the issue of the audio icon getting cut off. It seems like the `ShrinkingLinearLayout` resolves the issue of the icon being way over to the right side of the screen when the text only takes up part of the screen width, but I don't see how it resolves the reverse issue. Would you mind clarifying, just for sake of understanding?

app/src/org/commcare/views/ShrinkingLinearLayout.java
```diff
@@ -0,0 +1,81 @@
...
+
+    public ShrinkingLinearLayout(Context context, AttributeSet attrs) {
+        super(context, attrs);
+        loadViewConfig(context, attrs);
+    }
+
+    private void loadViewConfig(Context context, AttributeSet attrs) {
+        TypedArray typedArray = context.obtainStyledAttributes(attrs, R.styleable.ShrinkingLinearLayout);
+        shrinkingViewId = typedArray.getResourceId(R.styleable.ShrinkingLinearLayout_shrinkable_view, -1);
+    }
```
phillipm: Linter wants you to include `typedArray.recycle();` at the end of this method

app/src/org/commcare/views/ShrinkingLinearLayout.java
```diff
@@ -0,0 +1,81 @@
...
+ * layout.
+ *
+ * Limitations:
+ * This layout is incrementally slower than a traditional linear layout.
+ * The view will currently only shrink horizontally
+ *
+ * Created by ctsims on 5/18/2016.
+ */
+public class ShrinkingLinearLayout extends LinearLayout {
+    int shrinkingViewId = -1;
```
phillipm: `private`

ctsims:
@amstone326 @phillipm feedback addressed

amstone326:
@ctsims Will merge, but can you answer my question?

ctsims:
@amstone326 Thanks for flagging, sorry to miss that question:

Linear layouts function in one of two modes:

Mode 1: the layout adds its children in order, providing no real advice as to how long they should "Try" to be until one runs off the screen

^ This is what the old layout was doing

Mode 2: The layout tells its children "You get 20% of the width, you get 30% of the width, etc, etc" and the children will always be that width, alternatively it can say, "Child A gets 20%, Child B gets 30%, Child C gets 'the rest'", but the overall view has to be as wide as the screen.

^ This is the "Other broken example" which will never get something "Pushed" off the screen but also means your views can't be next to each other 'until' they need to fill up the width.

The way the PR addresses the "getting cut off" part of the problem is that it has you set everything up in Mode 2 (IE: The view is fully maximized in width and it is clear what % of the screen each child gets), but then also lets you provide a heuristic to say "One of these views might actually want to be way less big". It essentially takes the static view (Mode 2), measures all of the children in anticipation of drawing, but then measures one specific child in a vacuum, and if it's smaller than the "% of the screen" layout it forces it to be the size it 'wants' to be.

Kind of awkward to explain, might be easier to whiteboard if you have a few minutes.

amstone326:
@ctsims Hey that actually made a lot of sense for the most part. The only thing I'm still not following is what in this PR is actually making this happen: "The view is fully maximized in width and it is clear what % of the screen each child gets". I see that you added `android:layout_weight="1"` to the `row_txt` view, but based upon the documentation here https://developer.android.com/guide/topics/ui/layout/linear.html#Weight, I would expect you to have to give a `layout_weight` attribute to the `row_img` view as well, in order for the image to not get cut off when the text is long. What am I missing?

PR #1276, opened 2016-05-23
Redo demo warning dialog

phillipm:
| Old | New |
|---|---|
| demo-old | demo-new |

Make demo warning persist across screen rotation

**Additionally**: 
 - updates to google-services 9.0, which has no notable changes to maps or analytics: https://developers.google.com/android/guides/releases 
 - update from SQLCipher 3.3.1-2 to 3.4.0. I read over the [changelog](https://discuss.zetetic.net/t/sqlcipher-3-4-0-release/1273) and looks like we are good to go, (checked that we don't use the now deprecated `PRAGMA cipher`). Touts something like up to a 20% performance improvement over 3.3.1

### DIFF
--- a/templates/google-services.json
+++ b/templates/google-services.json
@@ -15,7 +15,7 @@
         }
       },
       "oauth_client": [],
-      "api_key": [],
+      "api_key": [{ "current_key": "" }],
       "services": {
         "analytics_service": {
           "status": 2,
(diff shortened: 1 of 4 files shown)


templates/google-services.json
```diff
@@ -15,7 +15,7 @@
         }
       },
       "oauth_client": [],
-      "api_key": [],
+      "api_key": [{ "current_key": "" }],
```
phillipm: Required to update from `com.google.gms:google-services:2.0.0` to `com.google.gms:google-services:3.0.0`

amstone326:
Good call, thanks for doing this

ctsims:
Nice! Thanks.

PR #1278, opened 2016-05-23
Redo case list action UI

phillipm:
While in the field we noticed people would erroneously click on actions when trying to select a case after performing a search. The best way to solve this would probably be to use [floating action buttons](https://www.google.com/design/spec/components/buttons-floating-action-button.html) or somehow get the actions to stick to the bottom of the screen once the bottom of the list is reached. Both seem pretty difficult given UI and backwards compatibility constraints.

**release notes**: improve case list action UI

**question**: In the screenshots below you will see that the icon image is scaled up (and pixelated) in the old setup, and in the new setup is not scaled. Not sure why this is happening or which one is better.
It also looks like the image effects the height of the action card. Any ideas of how to constrain this?

| before | after |
|---|---|
| screen  |  screen |

| before: normal list | before: searching |
|---|---|
| screen |  screen |

| after: normal list | after: searching |
|---|---|
| screen | screen |


phillipm:
@orangejenny @biyeun @dimagi/product Thoughts on this mobile UI change? 

I'm still not terribly pleased with it but I think it is better than what we currently have. 

The screenshots are exemplifying how we might soon have more than one action at the bottom of a case list, like "register" and "search for case"

biyeun:
Instead of the line, I would make the background around the cards dark and the cards themselves white

biyeun:
also the left-right padding around the cards is too wide---this should be narrower and the same as the distance between the cards vertically

amsagoff:
I was just looking at this last week and wishing we could update the design. Awesome to see!

phillipm:

amstone326:
@phillipm It looks like you made some UI changes from Biyeun's suggestion; would you mind adding updated screenshots? Or are the changes very minor?

phillipm:
The screenshots are up to date. decided to not follow all the suggestions for simplicity

app/res/layout/action_card.xml
```diff
@@ -0,0 +1,77 @@
...
+                android:layout_height="wrap_content"
+                android:layout_centerVertical="true"
+                android:layout_toRightOf="@+id/text"
+                android:background="@null"
+                android:focusable="false"
+                android:focusableInTouchMode="false"
+                android:padding="@dimen/min_button_padding"
+                android:visibility="gone"/>
+
+            <View
```
amstone326: What does this view correspond to in the screenshot? Seems like it would be some sort of vertical gray bar on the card, but don't see anything like that

app/src/org/commcare/adapters/EntityListAdapter.java
```diff
@@ -198,18 +207,33 @@ public TreeReference getItem(int position) {
...
+            case DIVIDER_TYPE:
+                return -2;
+            default:
+                throw new RuntimeException("Invalid view type");
         }
-        return references.indexOf(current.get(position).getElement());
+    }
+
+    private Action getAction(int position) {
+        return detail.getCustomActions().get(position - (dividerPosition + 1));
```
amstone326: the extra parens in this expression are unnecessary. If you did it that way because `dividerPosition + 1` is its own logical entity, you could just make that expression its own variable

amstone326:
@phillipm Just have the 2 above comments/questions and then good to merge

app/res/layout/action_card.xml
phillipm: Oh, I don't know why that isn't showing. I removed it and verified the UI looks the same as the screenshots

phillipm:
@amstone326 thanks, updated the PR

app/src/org/commcare/adapters/EntityListAdapter.java
```diff
@@ -198,18 +207,34 @@ public TreeReference getItem(int position) {
...
+    private Action getAction(int position) {
+        int baseActionPosition = dividerPosition + 1;
+        return detail.getCustomActions().get(position - baseActionPosition);
     }
 
     @Override
     public int getItemViewType(int position) {
-        if (actionsCount > 0 && position >= actionsStartPosition) {
-            return 1;
+        if (actionsCount > 0) {
```
wpride: NIT but this was a little hard to parse might be good for a comment

app/src/org/commcare/views/HorizontalMediaView.java
```diff
@@ -1,137 +0,0 @@
...
-import org.javarosa.core.reference.ReferenceManager;
-import org.javarosa.core.services.locale.Localizer;
-
-import java.io.File;
-
-/**
- * This layout for the GenericMenuFormAdapter allows you to load an image, audio, and text
- * to menus.
- *
- * @author wspride
```
wpride: :( he had a good run

PR #1279, opened 2016-05-24
Small startup optimization: don't print tables in prod code

phillipm:
- Only print resource tables on debug builds. Takes ~.25 seconds during startup time.
 - Move Android prototype factory setup code to its own file for better readability
 - Remove unneeded class name `contains` checks for `.test.` and `readystatesoftware`

### DIFF
--- a/app/src/org/commcare/models/database/DbUtil.java
+++ b/app/src/org/commcare/models/database/DbUtil.java
@@ -8,106 +8,13 @@ import android.util.Log;
 import net.sqlcipher.database.SQLiteDatabase;
 import net.sqlcipher.database.SQLiteDatabaseHook;
 
-import org.commcare.models.AndroidPrototypeFactory;
 import org.commcare.modern.database.DatabaseHelper;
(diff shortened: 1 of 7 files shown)


app/src/org/commcare/models/database/DbUtil.java
```diff
@@ -8,107 +8,14 @@
...
-            loadClass(cn, classNames);
-        }
-
-        return classNames;
-    }
-
-    public static void loadClass(String cn, List<String> classNames) {
-        try {
-            for (String packageName : packageNames) {
-                if (cn.startsWith(packageName) && !cn.contains(".test.") && !cn.contains("readystatesoftware")) {
```
phillipm: removed `&& !cn.contains(".test.") && !cn.contains("readystatesoftware")` from the moved code.

amstone326: I'm sure you're right because I have no idea what this did, but why is it no longer needed?

PR #1281, opened 2016-05-24
Allow for migration of class names in prototype factory

phillipm:
Renaming a class that implements `Externalizable` will break updates from older app versions when the old app has serialized that class into data it wants to read.

For instance, if you save a form as incomplete (with an intent callout questions) on 2.78 and change the location of `IntentCallout` in 2.79 and update to 2.79, then when the app tries to load the saved form it won't be able to find the `IntentCallout` class during the deserialization process.

This is because externalizable classes are indexed by a hash of their full package name.

This PR does 2 things:
 - Add a test that fails if you try to rename any classes that implements `Externalizable`. It will fail if you add a new `Externalizable` implementing class without adding it to the test.
 - Allow you to manually define a migration such that the tests pass and subsequent CommCare updates will be able to find the class path.

cross-request: https://github.com/dimagi/javarosa/pull/297


unit-tests/src/org/commcare/android/tests/processing/FormStorageTest.java
```diff
@@ -26,12 +31,187 @@
 @RunWith(CommCareTestRunner.class)
 public class FormStorageTest {
 
+    private static final List<String> classNames =
```
amstone326: I assume this is just a list you compiled manually of classes that implement `Externalizable`?

amstone326: Could rename to `externalizableClassNames`

unit-tests/src/org/commcare/CommCareTestApplication.java
```diff
@@ -59,32 +60,38 @@ public CommCareApp getCurrentApp() {
...
             throw new RuntimeException(e);
         }
 
-        testPrototypeFactor = new AndroidPrototypeFactory(tree);
-        return testPrototypeFactor;
+        testPrototypeFactory = new AndroidPrototypeFactory(tree);
+        return testPrototypeFactory;
     }
 
+    public static List<String> getTestPrototypeFactoryClasses() {
```
amstone326: Confused by the naming of this method in general so an explanation would be helfpul, but it should likely say something more than just "get" since it has side effects from calling `addExternalizableClassesFromDir`

unit-tests/src/org/commcare/CommCareTestApplication.java
```diff
@@ -80,6 +80,10 @@ public PrototypeFactory getPrototypeFactory(Context c) {
         return testPrototypeFactory;
     }
 
+    /**
+     * Get externalizable classes from *.class files in build dirs. Used to
+     * build PrototypeFactory that mirrors a prod environment
```
amstone326: I think you probably meant to add this to the `get` method instead?

amstone326: Ah no nevermind I got tripped up by the word "get", but makes sense now. Could make change "Get" to "Pull" for clarification

PR #1282, opened 2016-05-25
Split out graphing helper methods

orangejenny:
https://github.com/dimagi/commcare-odk/pull/1277/files/7a5e307a32af5bc318f4b83153c818b03cb84c46#r64393918

@phillipm

### DIFF
--- a/app/src/org/commcare/views/EntityDetailView.java
+++ b/app/src/org/commcare/views/EntityDetailView.java
@@ -4,6 +4,7 @@ import android.app.Activity;
 import android.content.Context;
 import android.content.Intent;
 import android.graphics.Bitmap;
+import android.os.Build;
 import android.view.Display;
 import android.view.GestureDetector;
 import android.view.Gravity;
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/views/EntityDetailView.java
```diff
@@ -434,4 +369,95 @@ private int getScreenWidth() {
...
     }
+
+    private void addSpinnerToGraph(WebView graphView, ViewGroup graphLayout) {
+        final ProgressBar spinner = new ProgressBar(this.getContext(), null, android.R.attr.progressBarStyleLarge);
+        spinner.setLayoutParams(new FrameLayout.LayoutParams(ViewGroup.LayoutParams.WRAP_CONTENT, ViewGroup.LayoutParams.WRAP_CONTENT, Gravity.CENTER));
+        GraphLoader graphLoader = new GraphLoader((Activity) this.getContext(), spinner);
+
+        // Set up interface that JavaScript will call to hide the spinner
+        // once the graph has finished rendering.
+        ((WebView)graphView).addJavascriptInterface(graphLoader, "Android");
```
phillipm: Would be good to get rid of the lint warning by either adding `@SuppressLint("AddJavascriptInterface")` to the method (with a comment as to why it should be safe) or adding code to disable the spinner JS code when running on devices below sdk level 17.

phillipm: `((WebView)graphView)` -> `graphView`

phillipm:
Yay, so much more readable!

app/src/org/commcare/views/EntityDetailView.java
orangejenny: > Would be good to get rid of the lint warning by either

Ended up doing both: skipping the method pre-jellybean, and suppressing the warning to make AS happy.

PR #1283, opened 2016-05-25
Show error message when lookup table for select question missing

phillipm:
If a user tries to open a form for which they don't have access to any of the entries in a lookup table select question, CommCare used to crash. Instead, show this error box and back out of the form: 
screen

cross-requested: https://github.com/dimagi/javarosa/pull/299

### DIFF
--- a/app/src/org/commcare/activities/components/FormNavigationUI.java
+++ b/app/src/org/commcare/activities/components/FormNavigationUI.java
@@ -6,7 +6,6 @@ import android.annotation.TargetApi;
 import android.content.Context;
 import android.graphics.Rect;
 import android.os.Build;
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/activities/components/FormNavigationUI.java
```diff
@@ -51,7 +50,7 @@ public static void updateNavigationCues(CommCareActivity activity,
         FormNavigationController.NavigationDetails details;
         try {
             details = FormNavigationController.calculateNavigationStatus(formController, view);
-        } catch (XPathTypeMismatchException e) {
+        } catch (XPathException e) {
```
phillipm: The fix

PR #1285, opened 2016-05-27
Fix group label not restoring after being hidden

phillipm:
Refactor of https://github.com/dimagi/commcare-odk/pull/787 and bug fix for http://manage.dimagi.com/default.asp?227683

 🐟 🐟 🐟 review commit-by-commit 🐟 🐟 🐟

### DIFF
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -1124,22 +1124,21 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
         mViewPane.addView(questionsView, lp);
 
         questionsView.startAnimation(mInAnimation);
-
-        FrameLayout header = (FrameLayout)findViewById(R.id.form_entry_header);
-
-        TextView groupLabel = ((TextView)header.findViewById(R.id.form_entry_group_label));
-
-        this.mGroupNativeVisibility = false;
-        FormLayoutHelpers.updateGroupViewVisibility(this, mGroupNativeVisibility, mGroupForcedInvisible);
-
         questionsView.setFocus(this);
 
+        setupGroupLabel();
+    }
+
+    private void setupGroupLabel() {
+        mGroupNativeVisibility = false;
+        FormLayoutHelpers.updateGroupViewVisibility(this, false, mGroupForcedInvisible);
         SpannableStringBuilder groupLabelText = questionsView.getGroupLabel();
 
         if (groupLabelText != null && !groupLabelText.toString().trim().equals("")) {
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/activities/components/FormLayoutHelpers.java
```diff
@@ -22,13 +23,26 @@
...
         TextView groupLabel = ((TextView)header.findViewById(R.id.form_entry_group_label));
 
-        int contentSize = newRootViewDimensions.height();
+        int numberOfGroupLinesAllowed =
+                getNumberOfGroupLinesAllowed(groupLabel, newRootViewDimensions, activity);
+
+        if (TextViewCompat.getMaxLines(groupLabel) != numberOfGroupLinesAllowed) {
+            shouldHideGroupLabel = numberOfGroupLinesAllowed == 0;
+            groupLabel.setMaxLines(numberOfGroupLinesAllowed);
+            updateGroupViewVisibility(header, groupLabel, hasGroupLabel, shouldHideGroupLabel);
```
phillipm: As far as I can tell the bug was that the last 2 arguments in `updateGroupViewVisibility(header, groupLabel, hasGroupLabel, shouldHideGroupLabel);` were flipped.

This code was hard to understand, but during my refactors I realized that the names given didn't match the argument names for `updateGroupViewVisibility`, so I tried flipping them to match and that fixed the bug reported.

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -1124,22 +1123,21 @@ private void showView(QuestionsView next, AnimationType from, boolean animateLas
...
+        FormLayoutHelpers.updateGroupViewVisibility(this, false, shouldHideGroupLabel);
         SpannableStringBuilder groupLabelText = questionsView.getGroupLabel();
 
         if (groupLabelText != null && !groupLabelText.toString().trim().equals("")) {
+            TextView groupLabel = (TextView)findViewById(R.id.form_entry_group_label);
             groupLabel.setText(groupLabelText);
-            this.mGroupNativeVisibility = true;
-            FormLayoutHelpers.updateGroupViewVisibility(this, mGroupNativeVisibility, mGroupForcedInvisible);
+            hasGroupLabel = true;
+            FormLayoutHelpers.updateGroupViewVisibility(this, true, shouldHideGroupLabel);
```
amstone326: why is it ok to change the 2nd argument here to always be `true`?

phillipm: the line above sets it to `true` so it is always true. Mostly did this because the linter was complaining.

PR #1289, opened 2016-06-01
Calendar prototype

sjain28:
Basic functionality for 2 standalone widgets. Both still need tweaks. 

cross-request: https://github.com/dimagi/commcare-core/pull/349


app/src/org/commcare/views/widgets/GregorianWidget.java
```diff
@@ -0,0 +1,136 @@
...
+
+    @Override
+    protected UniversalDate fromMillis(long millisFromJavaEpoch) {
+        return constructUniversalDate(new DateTime(millisFromJavaEpoch));
+    }
+
+    @Override
+    protected String[] getMonthsArray() {
+        if(myMonths == null){
+            myMonths = new String[]{"January", "February", "March", "April", "May", "June", "July",
```
wpride: At some point we're definitely going to need this to be localizable - looks like there are a few ways to do this, here looks good to start http://stackoverflow.com/questions/14832151/how-to-get-month-name-from-calendar

app/src/org/commcare/views/widgets/GregorianWidget.java
```diff
@@ -0,0 +1,136 @@
...
+        myCal = Calendar.getInstance();
+
+
+        //TODO: Validate input from text fields: Use a TextWatcher for this, use myCal to validate days in month
+        //TODO: Update month array pointer based on text entered..check to see how super does this though
+    }
+
+    @Override
+    protected void initText(){
+        dayTxt = (EditText)findViewById(R.id.daytxt);
```
wpride: My IDE is complaining about this being a TextView - this is because, confusingly, Android resource id's aren't domained by the file so if you have +id/daytxt in gregorian_date_widget.xml and universal_date_widget.xml then that reference will become ambiguous. So you'll need new IDs in your layout file

app/src/org/commcare/views/widgets/GregorianWidget.java
wpride: And since you're referencing this array in multiple places it should probably be in some static helper

app/src/org/commcare/views/widgets/GregorianWidget.java
```diff
@@ -0,0 +1,136 @@
...
+        LayoutInflater vi = (LayoutInflater)context.getSystemService(Context.LAYOUT_INFLATER_SERVICE);
+        View vv = vi.inflate(R.layout.gregorian_date_widget, null);
+        addView(vv);
+    }
+
+    @Override
+    protected void updateDateDisplay(long millisFromJavaEpoch) {
+        UniversalDate dateUniv = fromMillis(millisFromJavaEpoch);
+        dayTxt.setText(String.format("%02d", dateUniv.day));
+        monthTxt.setText(monthsArray[dateUniv.month - 1]);
```
wpride: Be careful with this (and on lines 59 and 72) - this sort of implicit knowledge can get dangerous. I'd consider wrapping all these month-index-shifts in some helper function

app/src/org/commcare/views/widgets/WidgetFactory.java
```diff
@@ -61,7 +61,11 @@ public QuestionWidget createWidgetFromPrompt(FormEntryPrompt fep, Context contex
                             questionWidget = new EthiopianDateWidget(context, fep);
                         } else if (appearance != null && appearance.toLowerCase().equals("nepali")) {
                             questionWidget = new NepaliDateWidget(context, fep);
-                        } else {
+                        }else if(appearance != null && appearance.toLowerCase().equals("greg")){ //TESTING GREGORIAN WIDGET
+                            questionWidget = new GregorianWidget(context, fep);
+                        }else if(appearance != null && appearance.toLowerCase().equals("calendar")){
+                            questionWidget = new CalendarWidget(context, fep);
```
wpride: Higher level question: how does CalendarWidget and GregorianWidget differ? Is it possible GregorianWidget should be a superclass of CalendarWidget?

app/res/layout/gregorian_date_widget.xml
```diff
@@ -0,0 +1,117 @@
...
+        <Button
+            android:id="@+id/monthupbtn"
+            android:layout_width="150dp"
+            android:layout_height="wrap_content"
+            android:layout_marginLeft="5dp"
+            android:layout_toRightOf="@id/dayupbtn"
+            android:background="@drawable/date_button_top"
+            android:text="+"
+            android:textSize="@dimen/text_largest"
+            android:textColor="#666"/>
```
wpride: all colors should go in colors.xml

app/res/layout/gregorian_date_widget.xml
```diff
@@ -0,0 +1,117 @@
...
+            android:layout_width="50dp"
+            android:layout_height="wrap_content"
+            android:background="@drawable/date_button_top"
+            android:text="+"
+            android:textSize="@dimen/text_largest"
+            android:textColor="@color/darkest_grey"/>
+
+        <EditText
+            android:id="@+id/daytxt"
+            android:layout_width="50dp"
```
wpride: Make sure you take a look at these on small and x-large screens too - can help you find some if you need

app/src/org/commcare/views/widgets/GregorianWidget.java
sjain28: Resolved.

app/src/org/commcare/views/widgets/GregorianWidget.java
sjain28: Resolved.

app/src/org/commcare/views/widgets/WidgetFactory.java
sjain28: They're completely unrelated - GregorianWidget is a Gregorian counterpart to Nepali and Ethiopian...all 3 of these inherit from the same superclass. 

CalendarWidget's GUI and internal logic are totally different. Will rename these classes to be more clear.

app/res/layout/gregorian_date_widget.xml
sjain28: Resolved.

app/res/layout/gregorian_date_widget.xml
```diff
@@ -99,7 +100,8 @@
             android:gravity="center"
             android:padding="3dp"
             android:textSize="@dimen/text_larger"
-            android:inputType="number"/>
+            android:inputType="number"
+            android:maxLength="4"/>
```
wpride: Month maxLenth is 4?

wpride:
@sjain28 Usually when we submit a PR that includes a UI change/addition we include an image of all the new screens so that people can view and comment without building themselves - could you add this? This might be useful: http://stackoverflow.com/questions/20983351/taking-screenshot-on-emulator-from-android-studio

wpride:
General UI thoughts:

- The text on the calendar widget could be bigger
- The buttons on the calendar widget could be bigger
- Instead of the grey with sharp edges I think we've been moving towards more of a blue/white aesthetic with bevelled edges
- Applies to text backgrounds too
- The calendar-toggle-button could be bigger (probably commcare blue too?)
- The button with a "Red X" I think should be an ImageView (like the calendar) with a red X image instead

But again, all of these changes are probably good to bring to Biyeun/Jenny

wpride:
General functional Thoughts:
- When you click the calendar button and the calendar comes up, you have to hit the grey "X" in order to accept the date - this seems unintuitive, possibly should be a green check instead?
- the calendar widget doesn't have an "X" button to give a null answer (maybe by design?)
- the second prototype autocomplete (which is really cool btw) the suggestions are very small, will be hard to click for some users
- you should probably disable keyboard suggestions when auto completing: http://stackoverflow.com/questions/4281838/how-to-disable-displaying-suggestions-on-the-soft-keyboard
- Is it intentional or by design that autocomplete doesn't start until the second letter is typed?
- If I type a bad an unallowed month (IE "Will) and then swipe forward, it reverts back to the initial suggested month - is that by design? Possible we should block that as an unallowed answer? 

Good questions for @dimagi/product and kevin

app/res/layout/calendar_widget.xml
```diff
@@ -0,0 +1,142 @@
...
+            android:layout_height="wrap_content"
+            android:layout_gravity="center_horizontal"
+            android:paddingLeft="5dp"
+            android:text="SAT" />
+
+    </LinearLayout>
+
+    <GridView
+        android:id="@+id/calendar_grid"
+        android:layout_width="280dp"
```
wpride: So this should probably be matchParent or something else related to the full width of the screen

app/res/layout/calendar_widget.xml
```diff
@@ -0,0 +1,142 @@
...
+        android:layout_height="340dp"
+        android:columnWidth="40dp"
```
wpride: then this should be proportional using the android:weightSum and android:layout_width properties

app/res/layout/calendar_widget.xml
```diff
@@ -0,0 +1,142 @@
...
+            android:layout_gravity="center_horizontal"
+            android:paddingLeft="5dp"
+            android:text="FRI" />
+
+        <TextView
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_gravity="center_horizontal"
+            android:paddingLeft="5dp"
+            android:text="SAT" />
```
wpride: These days should be localizable

app/res/layout/calendar_widget.xml
```diff
@@ -0,0 +1,142 @@
...
+        android:layout_marginTop="10dp"
+        android:gravity="center">
+
+        <TextView
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:id = "@+id/calendarweekday"
+            android:layout_centerVertical="true"
+            android:layout_centerHorizontal="true"
+            android:text="Day"
```
wpride: localizable

app/src/org/commcare/views/widgets/Prototype1.java
```diff
@@ -0,0 +1,148 @@
...
+    private void initButtons() {
+        openCalButton = new ImageButton(getContext());
+
+        openCalButton.setImageResource(R.drawable.avatar_vellum_date);
+
+        RelativeLayout layout = (RelativeLayout) findViewById(R.id.widgetinfo);
+        layout.addView(openCalButton);
+        RelativeLayout.LayoutParams params = (RelativeLayout.LayoutParams) openCalButton.getLayoutParams();
+        params.addRule(RelativeLayout.RIGHT_OF, R.id.gregdayofweek);
+        params.width = 60;
```
wpride: Slightly repeating above comments, but usually we wouldn't hard code but would have things be proportional to the screen size

app/src/org/commcare/views/widgets/Prototype1.java
```diff
@@ -0,0 +1,148 @@
...
+import java.util.Calendar;
+import java.util.Date;
+
+
+/**
+ * Created by Saumya on 6/1/2016.
+ */
+public class Prototype1 extends QuestionWidget {
+
+    private CalendarWidget myCal;
```
wpride: Think we're moving away from the mClassMemberVariable naming convention

app/src/org/commcare/views/widgets/Prototype3.java
```diff
@@ -0,0 +1,138 @@
...
+public class Prototype3 extends QuestionWidget{
+
+    private Prototype2 myPro2;
+    private LinearLayout myLayout;
+
+    public Prototype3(Context context, FormEntryPrompt prompt){
+        //TODO: Add buttons
+        super(context, prompt);
+        myPro2 = new Prototype2(context, prompt);
+        myPro2.removeQuestionText();
```
wpride: Why do we need/want to do this?

app/src/org/commcare/views/widgets/Prototype3.java
```diff
@@ -0,0 +1,138 @@
...
+import org.commcare.dalvik.R;
+import org.javarosa.core.model.data.DateData;
+import org.javarosa.core.model.data.IAnswerData;
+import org.javarosa.form.api.FormEntryPrompt;
+
+import java.util.Calendar;
+import java.util.Date;
+
+/**
+ * Created by Saumya on 6/2/2016.
```
wpride: class level comments on each of these would be good

app/src/org/commcare/views/widgets/Prototype3.java
```diff
@@ -0,0 +1,138 @@
...
+
+    public Prototype3(Context context, FormEntryPrompt prompt){
+        //TODO: Add buttons
+        super(context, prompt);
+        myPro2 = new Prototype2(context, prompt);
+        myPro2.removeQuestionText();
+
+        LayoutInflater inflater = (LayoutInflater)context.getSystemService
+                (Context.LAYOUT_INFLATER_SERVICE);
+        myLayout = (LinearLayout) inflater.inflate(R.layout.prototype3, null);
```
wpride: could be a better variable name

app/src/org/commcare/views/widgets/Prototype1.java
```diff
@@ -0,0 +1,148 @@
...
+            @Override
+            public void onClick(View v) {
+                closeCalendar();
+            }
+        });
+
+        //TODO: Add green "Submit" button. Not done so far because it's unclear what its point is. Navigation to next question is handled already.
+    }
+
+    public ImageButton getCalendarButton(){
```
wpride: unused

app/src/org/commcare/views/widgets/Prototype1.java
```diff
@@ -0,0 +1,148 @@
...
+            myCal.setDate(new DateData(new Date()));
+        }
+
+        myGreg.setFocus(getContext());
+        myGreg.setVisibility(GONE);
+        myCal.setVisibility(VISIBLE);
+    }
+
+    private void closeCalendar(){
+        myGreg.setDate((DateData) myCal.getAnswer());
```
wpride: Interesting - so the gregorian calendar is the final source of truth here? Feels slightly awkward that the CalendarWidget and GregorianWidget both have their own date values that we track and then coalesce. (Bigger comment at PR level)

wpride:
Higher Level design feedback: 

Might be worth thinking about a refactor whereby the CalendarWidget is no longer itself a QuestionWidget, but rather a separate UI/functional piece that can be optionally added to a widget (perhaps a Fragment: https://developer.android.com/reference/android/app/Fragment.html). then it could operate on the same core Date object as the GregorianDateWidget itself.

also curious: It seems like Prototype1 contains a GregorianDateWidget, P2 contains P1, and P3 contains P2. Did you consider using an inheritance pattern here?

orangejenny:
yeah, screenshots would be great

orangejenny:
Thoughts on ux:
- "X" button should probably have a different style (check with Biyeun), same for the arrow buttons in the calendar
- (Prototype 1) Calendar icon jumps around as the length of the day of week string changes
- Month dropdown could use more spacing around the individual dropdown items (text is too close to the left edge, and the items are a little short overall, seems easy to tap the wrong one)
- In calendar, grey out the days that aren't in the current month
- In calendar, don't show weeks that are entirely in a different month (June 2016 calendar also shows July 3-9)
- In calendar, add background color or some other visual differentiator to day of week headers
- In calendar, highlight (delicately, maybe a lighter blue? biyeun probably has an opinion on color) today's date
- When calendar is showing, don't show the day of week that's currently in line with the checkmark
- (Prototype 2) Year text box is too small
- (Prototype 2) Calendar icon is squashed, too wide, should be a circle
- Validation: shouldn't silently blank out bad dates, should show error message
- Calendar icon is distorted in landscape mode
- (Prototype 1) Alignment of day of week text & calendar icon is weird in landscape mode (they get displayed towards the middle of the screen, but they should probably still be aligned with the left and right edges of the widget)
- (Prototype 3) Would be nice if the << < > >> buttons didn't jump around as the length of the day of week text changes
- Test on whatever the smallest likely android screen size is
- (Prototype 3) In the calendar, go to Sept 2016, and the "2016" is split onto two lines
- (Prototype 3) In the calendar, pick a Wednesday, and the word breaks over two lines

biyeun:
Ok had a thorough look-through this evening. @sjain28 great functionality and speedy execution! yes!

A couple notes:
- The calendar widget should be a pop up. Otherwise, the navigation along the same plane as the form is confusing to the user. Think of it like pieces of paper and the form is a long scroll that moves left and right. The current calendar widget doesn't fit inside that mental model, therefore it makes more sense that it floats on top of the form in a dialog. @sjain28 would be a good exercise to skim through the google [material design docs](https://material.google.com/), which we base our UIs on. :)

Here are my suggested edits. I'll send the pngs + pdfs to the people involved in this discussion via email too.

calendar-prototype-1
calendar-prototype-2
calendar-prototype-2-type
calendar-prototype-3
calendar-overlay

biyeun:
realizing on the first prototype  we should probably approach the button layout like: 
calendar-prototype-1 2

krsdimagi:
@sjain28 let's hold off on some of these higher fidelity design elements until you have implemented some of the feedback from @orangejenny and @wspride.  As we test out our prototypes, we may find that we have a winner and don't need to take the time up front with all of the visual elements if we don't have to.  @biyeun feel free to push back if this doesn't make sense - the images look fantastic!

biyeun:
@krsdimagi makes sense. I was also thinking that there might be use cases where each one of these prototypes could be more valid / useful than the other.

sjain28:
Cross requested with https://github.com/dimagi/javarosa/pull/302

wpride:
These look and behave great Saumya, well done. Two small UI NITs:

device-2016-06-16-114638

The buttons on this are different sizes, and the calendar button on prototypes 2 and 3 aren't the CommCare teal color. Some more space between the two buttons would probably be good too.

Second, on the calendar pop up the green check mark is a bit small and hard to click - might be good to size up.

sjain28:
#Screenshots; 

screenshot_20160617-172702
screenshot_20160617-172709
screenshot_20160617-172714
screenshot_20160617-172717
screenshot_20160617-173013

wpride:
Looks really great Saumya! Nice work.

app/assets/locales/messages_ccodk_default.txt
```diff
@@ -400,6 +400,15 @@ template.malformed.chevron=The template file set for your print job has improper
 cannot.set.template=Could not set your print template
 no.file.browser=Your device does not have a file browser installed. Please download one from the playstore and then try again.
 
+start.recording=Start Recording
```
amstone326: It looks like these 2 strings must have accidentally gotten into this PR from the audio PR

app/assets/locales/messages_ccodk_default.txt
```diff
@@ -400,6 +400,15 @@ template.malformed.chevron=The template file set for your print job has improper
 cannot.set.template=Could not set your print template
 no.file.browser=Your device does not have a file browser installed. Please download one from the playstore and then try again.
 
+start.recording=Start Recording
+stop.recording=Stop Recording
```
amstone326: It looks like these 2 strings must have accidentally gotten into this PR from the audio PR

app/res/layout/calendar_widget.xml
```diff
@@ -0,0 +1,174 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="fill_parent"
+    android:layout_height="wrap_content"
+    android:orientation="vertical"
+    android:gravity="center"
+    android:layout_gravity="center_horizontal"
+    android:paddingLeft = "20dp"
```
amstone326: the convention for layout files is to remove the spaces here

app/res/layout/calendar_widget.xml
```diff
@@ -0,0 +1,174 @@
...
+    <LinearLayout
+        android:layout_width="fill_parent"
+        android:layout_height="wrap_content"
+        android:layout_gravity="center_horizontal"
+        android:gravity="center"
+        android:layout_marginTop="5dp"
+        android:orientation="horizontal">
+
+        <ImageButton
+            android:id="@+id/prevmonthbutton"
```
amstone326: this is really nitpicky, but would be great to use the underscore naming pattern for all of these ids, i.e. prev_month_button

app/res/layout/calendar_widget.xml
```diff
@@ -0,0 +1,174 @@
...
+        <TextView
+            android:layout_width="0dp"
+            android:textStyle="bold"
+            android:layout_height="wrap_content"
+            android:layout_weight="1"
+            android:layout_gravity="center_horizontal"
+            android:id = "@+id/day2"/>
+
+        <TextView
+            android:id = "@+id/day3"
```
amstone326: Same comment as above re:spaces (I won't keep commenting in all the places I see it)

app/res/layout/recording.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8"?>
```
amstone326: Also looks like this is probably from the audio widget PR?

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -1030,6 +1031,9 @@ private void showNextView(boolean resuming) {
      */
     private void showPreviousView(boolean showSwipeAnimation) {
         // The answer is saved on a back swipe, but question constraints are ignored.
+
```
amstone326: new lines added here are unnecessary

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -1146,6 +1150,7 @@ private void setupGroupLabel() {
      */
     private void showConstraintWarning(FormIndex index, String constraintText,
                                        int saveStatus, boolean requestFocus) {
+
```
amstone326: same here and below

app/src/org/commcare/views/widgets/GregorianDateWidget.java
```diff
@@ -0,0 +1,336 @@
...
+            @Override
+            public void beforeTextChanged(CharSequence s, int start, int count, int after) {}
+            @Override
+            public void onTextChanged(CharSequence s, int start, int before, int count) {}
+
+            @Override
+            public void afterTextChanged(Editable s) {
+                String contents = s.toString();
+
+                if(contents.length() > 1) {
```
amstone326: check what happens if input is 0, +, etc.

app/src/org/commcare/views/widgets/GregorianDateWidget.java
```diff
@@ -0,0 +1,336 @@
...
+    protected void initText(){
+        dayOfWeek = (TextView) findViewById(R.id.gregdayofweek);
+        dayTxt = (EditText)findViewById(R.id.daytxtfield);
+        yearTxt = (EditText)findViewById(R.id.yeartxtfield);
+
+        MonthAdapter monthAdapter = new MonthAdapter(getContext(), monthList);
+        monthTxt = (AutoCompleteTextView) findViewById(R.id.monthtxtfield);
+        monthTxt.setAdapter(monthAdapter);
+
+        dayTxt.addTextChangedListener(new TextWatcher() {
```
wpride: Would be nice to refactor these three TextWatchers into their own DateTextWatcher class that abstracts out the validation check and value casting

app/src/org/commcare/views/widgets/GregorianDateWidget.java
```diff
@@ -0,0 +1,336 @@
...
+            @Override
+            public void beforeTextChanged(CharSequence s, int start, int count, int after) {}
+            @Override
+            public void onTextChanged(CharSequence s, int start, int before, int count) {}
+
+            @Override
+            public void afterTextChanged(Editable s) {
```
amstone326: high-level discussion we had here- may want to consider making the widget's corrective behavior stronger than it currently is (i.e. immediately revert to the closest valid day/month upon clicking away if user input is invalid)

app/src/org/commcare/views/widgets/GregorianDateWidget.java
```diff
@@ -0,0 +1,336 @@
...
+        if(day > selectedDate.getActualMaximum(Calendar.DAY_OF_MONTH)){
+            dayTxt.setText(String.valueOf(selectedDate.getActualMaximum(Calendar.DAY_OF_MONTH)));
+            day = Integer.parseInt(dayTxt.getText().toString());
+        }
+
+        return toMillisFromJavaEpoch(year, month, day, millisOfDayOffset);
+    }
+
+    private void checkEmptyText() {
+        if(dayTxt.getText().toString().equals("")){
```
phillipm: `.equals("")` can also be `.isEmpty()`, which feels a bit more readable.

app/src/org/commcare/views/widgets/AbstractUniversalDateWidget.java
```diff
@@ -436,7 +450,7 @@ private void updateDateDisplay(long millisFromJavaEpoch) {
     /**
      * Update the widget helper date text (useful for those who don't know the other calendar)
      */
-    private void updateGregorianDateHelperDisplay() {
+    protected void updateGregorianDateHelperDisplay() {
```
phillipm: Since you've exposed this method to subclasses, I would rename it to not include the word `Helper`

app/src/org/commcare/views/widgets/GregorianDateWidget.java
```diff
@@ -0,0 +1,336 @@
...
+
+        if(day > selectedDate.getActualMaximum(Calendar.DAY_OF_MONTH)){
+            dayTxt.setText(String.valueOf(selectedDate.getActualMaximum(Calendar.DAY_OF_MONTH)));
+            day = Integer.parseInt(dayTxt.getText().toString());
+        }
+
+        return toMillisFromJavaEpoch(year, month, day, millisOfDayOffset);
+    }
+
+    private void checkEmptyText() {
```
wpride: Rename to what this actually does

app/src/org/commcare/views/widgets/AbstractUniversalDateWidget.java
amstone326: @phillipm What's the correlation between those 2 things?

app/src/org/commcare/views/widgets/GregorianDateWidget.java
```diff
@@ -0,0 +1,336 @@
...
+        dayOfWeek.setText(selectedDate.getDisplayName(Calendar.DAY_OF_WEEK, Calendar.LONG, Locale.getDefault()));
+    }
+
+    @Override
+    protected long getCurrentMillis() {
+
+        checkEmptyText();
+
+        int day = Integer.parseInt(dayTxt.getText().toString());
+        int month = monthArrayPointer + 1;
```
wpride: Wrap this function in a method

app/src/org/commcare/views/widgets/GregorianDateWidget.java
```diff
@@ -0,0 +1,336 @@
...
+        int year = Integer.parseInt(yearTxt.getText().toString());
+
+        if(day > selectedDate.getActualMaximum(Calendar.DAY_OF_MONTH)){
+            dayTxt.setText(String.valueOf(selectedDate.getActualMaximum(Calendar.DAY_OF_MONTH)));
```
wpride: possibly redundant?

app/src/org/commcare/views/widgets/AbstractUniversalDateWidget.java
amstone326: meaning why shouldn't a subclassed method use the word helper?

app/src/org/commcare/views/widgets/GregorianDateWidget.java
```diff
@@ -0,0 +1,336 @@
...
+        //Invalid month
+        if(!monthList.contains(monthTxt.getText().toString())){
+            setFocus(getContext());
+            return new InvalidData(Localization.get("invalid.month"), new DateData(selectedDate.getTime()));
+        }
+
+        //Invalid day (too high)
+        if(Integer.parseInt(dayTxt.getText().toString()) > selectedDate.getActualMaximum(Calendar.DAY_OF_MONTH)){
+            setFocus(getContext());
+            return new InvalidData(Localization.get("high.date") + String.valueOf(selectedDate.getActualMaximum(Calendar.DAY_OF_MONTH)), new DateData(selectedDate.getTime()));
```
wpride: Use localization parameters

app/src/org/commcare/views/widgets/GregorianDateWidget.java
```diff
@@ -0,0 +1,336 @@
...
+
+    @Override
+    protected UniversalDate decrementYear(long millisFromJavaEpoch) {
+        DateTime dt = new DateTime(millisFromJavaEpoch).minusYears(1);
+        return constructUniversalDate(dt);
+    }
+
+    @Override
+    protected UniversalDate fromMillis(long millisFromJavaEpoch) {
+        return constructUniversalDate(new DateTime(millisFromJavaEpoch));
```
phillipm: You should verify that the bug fixed in https://github.com/dimagi/commcare-odk/pull/1144 for the Nepali date widget doesn't occur here. Reproduction steps are in the description and in the test files. Let me know if you want to sit down and look into this.

app/src/org/commcare/views/widgets/CalendarFragment.java
```diff
@@ -0,0 +1,256 @@
...
+        myLayout.setMinimumWidth((int)(displayRectangle.width() * 0.9f));
+
+        return myLayout;
+
+    }
+
+    public void setArguments(Calendar cal){
+        myCal = cal;
+    }
+    public interface DismissListener{
```
amstone326: rename this to be clearer/more specific

app/src/org/commcare/views/widgets/CalendarFragment.java
```diff
@@ -0,0 +1,256 @@
...
+
+        initDisplay();
+        initWeekDays();
+        initOnClick();
+        refresh();
+
+        Rect displayRectangle = new Rect();
+        Window window = getActivity().getWindow();
+        window.getDecorView().getWindowVisibleDisplayFrame(displayRectangle);
+        myLayout.setMinimumWidth((int)(displayRectangle.width() * 0.9f));
```
phillipm: Check how this looks in landscape mode

app/src/org/commcare/views/widgets/CalendarFragment.java
```diff
@@ -0,0 +1,256 @@
...
+
+        return myLayout;
+
+    }
+
+    public void setArguments(Calendar cal){
+        myCal = cal;
+    }
+    public interface DismissListener{
+        void onDismiss();
```
phillipm: Distinguish between pressing 'OK' and 'Cancel'

app/src/org/commcare/views/widgets/GregorianDateWidget.java
```diff
@@ -0,0 +1,336 @@
...
+
+    private EditText dayTxt;
+    private AutoCompleteTextView monthTxt;
+    private EditText yearTxt;
+    private TextView dayOfWeek;
+    private Calendar selectedDate;
+    private String[] monthNames;
+    private List<String> monthList;
+    private final int MIN_YEAR = 1900;
+    private final String DAYFORMAT = "%02d";
```
orangejenny: naming convention: include underscores

app/src/org/commcare/views/widgets/CalendarFragment.java
```diff
@@ -0,0 +1,256 @@
...
+        ImageButton closer = (ImageButton) myLayout.findViewById(R.id.closecalendar);
+        closer.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                dismiss();
+            }
+        });
+    }
+
+    public void refresh(){
```
phillipm: A high-level comment would be useful

amstone326: split this up into well-named helper methods

app/src/org/commcare/views/widgets/CalendarFragment.java
```diff
@@ -0,0 +1,256 @@
...
+
+        //Backtracking calendar to the most recent Sunday
+        populator.add(Calendar.DAY_OF_MONTH, -monthStartWeekDay);
+
+        while(dateList.size() < totalDays){
+            dateList.add(populator.getTime());
+            populator.add(Calendar.DAY_OF_MONTH, 1);
+        }
+
+        int remainingDays = (8-populator.get(Calendar.DAY_OF_WEEK))%7;
```
amstone326: make the numbers you're using here static final class variables with a descriptive name

sjain28:
Pending big questions: 
-Have heard several different opinions about validation of user input, should make a point to bring this up during testing

sjain28:
Pending big questions: 
-Validation choices should be brought up during testing
-Check for timezone bug #1144

krsdimagi:
Hey Saumya!  Here are the major takeaways/changes we can make at this point post-hallway testing.  We can definitely go to voice to clear up any confusion, but these were the most-cited issues that we can use to improve usability at this point:

- The arrow functionality of the third prototype wasn’t well-received - we can drop this one and focus on the other two prototypes now.
- For both prototypes: the X/clear functionality was not understood.  Because of this, we can drop the ability to clear.
- Default value should always be today’s date (rather than incrementing based on the last entry option)
- For text entry:
  - It was not very usable because when you click on an entry field it should select the entire field and            the cursor should start at the beginning in order to avoid any typing problems, instead of the current behavior that allows you to select the middle of an entry.
  - Additionally, there was a bug on my motorola where after changing the year it would cut off the last number for some reason - this could be some kind of device-specific sizing issue, I can show you if this doesn’t make sense.
  - When in landscape mode, there was a “done” option as well as a check mark option - would it be possible to have this in portrait mode?  Users were nervous about clicking the check mark, and didn’t know if this would lock in an incorrect date, potentially.
- The calendar:
  - This was the most-appreciated feature.  Could we move it to the upper-righthand corner of each widget instead of under the date entry, as that is the expected location?
  - Testers wanted to be able to do text entry on the popout in the year and date, so they wouldn’t have to click the arrow buttons only - it might be nice to see if we can add this functionality.
  - It would be good to have an x next to the check mark on the popout, as only having a check mark made users nervous in the event that they just wanted to leave the calendar.
  - The calendar is not a fixed size - it would be good to fix the size so that when going from month-to-month it doesn’t seem to move around.

wpride:
@krsdimagi Awesome feedback Kevin, thanks. Definitely want to agree with Saumya on the clear date functionality - the ability to submit an empty date is one of the most in demand changes and in my mind a major win from this refactor. If the "X" was unclear I think we should rethink the design rather than scrap the functionality altogether. Maybe @biyeun or @orangejenny would have ideas.

Sorry if I'm out of the loop on this. @sjain28 Are next steps clear or should we schedule a follow up meeting?

amstone326:
@krsdimagi It looks like Will isn't in the meeting invite for this afternoon. Could you add him?

phillipm:
@damagatchi please retest this

phillipm:

phillipm:
Hey, with all the `PNG` image resources added in this PR and the audio recorder, do you have the original `SVG` versions? There is a way to perform auto-scaling by importing `SVG` files into Android Studio and using them as vector resources.

The current issue with just using one large `PNG` image resource is that on different screen dimensions Android will over consume memory resources. In the past this has cause devices to consistently crash. If you don't have the `SVG` versions (and Biyeun doesn't either) we can use a tool to create different versions for different screen densities.

app/assets/locales/messages_ccodk_default.txt
```diff
@@ -406,6 +406,13 @@ template.malformed.chevron=The template file set for your print job has improper
 cannot.set.template=Could not set your print template
 no.file.browser=Your device does not have a file browser installed. Please download one from the playstore and then try again.
 
+empty.fields=Please complete all fields
+low.year=Year must be > 1900
+invalid.month=Month is invalid
+high.date=Date can be at most
+low.date=Date must be at least 1
+high.year=Year can be at most
```
phillipm: Can you prefix all these keys with something like `calendar` or `date`? Something to give users who want to override them for different languages some context

app/res/layout/gregorian_date_widget.xml
```diff
@@ -0,0 +1,190 @@
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="fill_parent"
+    android:layout_height="wrap_content"
+    android:gravity="center"
+    android:orientation="vertical">
```
phillipm: This file seems to be unused

app/res/layout/calendar_widget.xml
```diff
@@ -0,0 +1,210 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="fill_parent"
+    android:layout_height="fill_parent"
+    android:orientation="vertical"
+    android:layout_gravity="center_horizontal"
+    android:paddingLeft="20dp"
+    android:paddingRight="20dp">
```
phillipm: This file seems to be unused

app/res/layout/autocomplete_month.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<TextView xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
+</TextView>
```
phillipm: Seems to be unused

app/src/org/commcare/views/widgets/DatePrototypeFactory.java
```diff
@@ -0,0 +1,19 @@
+package org.commcare.views.widgets;
+
+import android.content.Context;
+
+import org.javarosa.form.api.FormEntryPrompt;
+
+public class DatePrototypeFactory {
```
phillipm: I assume that this factory was useful when testing multiple prototypes. Now that we have settled on one, does it make sense to remove it and explicitly construct the `GregorianDateWidget` in `WidgetFactory`?

app/res/layout/spinner_gregorian_widget.xml
```diff
@@ -0,0 +1,188 @@
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="fill_parent"
+    android:layout_height="wrap_content"
+    android:gravity="center"
+    android:orientation="vertical">
```
phillipm: File doesn't seem to be used

app/res/layout/spinner_calendar_widget.xml
```diff
@@ -0,0 +1,180 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="fill_parent"
+    android:layout_height="fill_parent"
+    android:orientation="vertical"
+    android:layout_gravity="center_horizontal"
+    android:paddingLeft="20dp"
+    android:paddingRight="20dp">
```
phillipm: File doesn't seem to be used

app/src/org/commcare/views/widgets/AbstractUniversalDateWidget.java
```diff
@@ -205,30 +204,45 @@ public void onClick(View v) {
...
+//        btnYearUp.setOnTouchListener(new EDWTouchListener(btnYearUp, mYearHandler));
+//        btnYearDown.setOnTouchListener(new EDWTouchListener(btnYearUp, mYearHandler));
+//
+//        // button key listeners
+//        btnDayUp.setOnKeyListener(new EDWKeyListener(btnDayUp, mDayHandler));
+//        btnDayDown.setOnKeyListener(new EDWKeyListener(btnDayUp, mDayHandler));
+//        btnMonthUp.setOnKeyListener(new EDWKeyListener(btnMonthUp, mMonthHandler));
+//        btnMonthDown.setOnKeyListener(new EDWKeyListener(btnMonthUp, mMonthHandler));
+//        btnYearUp.setOnKeyListener(new EDWKeyListener(btnYearUp, mYearHandler));
+//        btnYearDown.setOnKeyListener(new EDWKeyListener(btnYearUp, mYearHandler));
```
phillipm: Can we remove this commented out code or uncomment it?

phillipm: If you remove it you should also remove the `mDayHandler`, `mMonthHandler`, `mYearHandler` code

app/src/org/commcare/views/widgets/AbstractUniversalDateWidget.java
phillipm: Ah, I think I misunderstood the use of the word `Helper`. I thought it was a helper method, but it is a method that shows helper text to the user. Ignore my initial suggestion.

app/src/org/commcare/views/widgets/CalendarFragment.java
```diff
@@ -0,0 +1,342 @@
...
+
+        //Backtracking calendar to the most recent Sunday
+        cal.add(Calendar.DAY_OF_MONTH, daysSincePreviousSunday);
+
+        return totalDays;
+    }
+
+    private class CalendarAdapter extends ArrayAdapter<Date>{
+
+        private LayoutInflater mInflater;
```
phillipm: can be `final`

app/src/org/commcare/views/widgets/CalendarFragment.java
```diff
@@ -0,0 +1,342 @@
...
+    private Spinner monthSpinner;
+    private Spinner yearSpinner;
+    protected LinearLayout myLayout;
+
+    protected Calendar calendar;
+    protected CalendarCloseListener myListener;
+
+    protected long todaysDateInMillis;
+    protected static final int DAYSINWEEK = 7;
+    protected static final String TIME = "TIME";
```
phillipm: All this fields can be made `private`

app/src/org/commcare/views/widgets/CalendarFragment.java
```diff
@@ -0,0 +1,342 @@
...
+    public void clear() {
+        calendar = Calendar.getInstance();
+        refresh();
+    }
+
+    public void setDate(DateData newDate){
+        Date nextDate = (Date) newDate.getValue();
+        calendar.setTimeInMillis(nextDate.getTime());
+        refresh();
+    }
```
phillipm: `setDate`, `clear`, and `getValue` seem to be unused

app/src/org/commcare/views/widgets/CalendarFragment.java
```diff
@@ -0,0 +1,342 @@
...
+        @Override
+        public View getView(int position, View view, ViewGroup parent){
+            if(view == null){
+                view = mInflater.inflate(R.layout.calendar_date, null);
+            }
+
+            TextView text = (TextView) view;
+
+            Date date = getItem(position);
+            text.setText(String.valueOf(date.getDate()));
```
phillipm: `getDate` is deprecated. Is there a non-deprecated method you can call with the same result?

app/src/org/commcare/views/widgets/CalendarFragment.java
```diff
@@ -0,0 +1,342 @@
...
+
+            Date current = calendar.getTime();
+
+            if(date.equals(current)){
+                text.setTextColor(getResources().getColor(R.color.white));
```
phillipm: `getColor(int)` is deprecated, should use `getColor(int, Theme)` and just pass `null` in for the `Theme` (as described in the docs of the `getColor` method)

app/src/org/commcare/views/widgets/CalendarFragment.java
```diff
@@ -0,0 +1,342 @@
...
+                text.setBackgroundColor(getResources().getColor(R.color.cc_attention_positive_color));
+            }
+            else if(date.getMonth() != current.getMonth()){
```
phillipm: `getMonth` is deprecated, is there an alternative we can use?

app/src/org/commcare/views/widgets/CalendarFragment.java
```diff
@@ -0,0 +1,342 @@
...
+
+/**
+ * Created by Saumya on 5/29/2016.
+ * DialogFragment for a popup calendar icon
+ * Uses support library for compatibility with pre-honeycomb devices
+ * Layout and logic inspired by Ahmed Al-Amir at https://www.toptal.com/android/android-customization-how-to-build-a-ui-component-that-does-what-you-want
+ */
+public class CalendarFragment extends android.support.v4.app.DialogFragment {
+
+    protected GridView myGrid;
```
phillipm: can you think of a more descriptive name for this? What is it a grid of?

app/src/org/commcare/views/widgets/CalendarFragment.java
```diff
@@ -0,0 +1,342 @@
...
+    protected ImageButton cancel;
+    protected Button today;
+    private Spinner monthSpinner;
+    private Spinner yearSpinner;
+    protected LinearLayout myLayout;
```
phillipm: `myLayout` -> something like `baseLayout`

app/src/org/commcare/views/widgets/CalendarFragment.java
```diff
@@ -0,0 +1,342 @@
...
+
+    protected Calendar calendar;
+    protected CalendarCloseListener myListener;
```
phillipm: `myListener` -> `calendarCloseListener`

app/src/org/commcare/views/widgets/CalendarFragment.java
```diff
@@ -0,0 +1,342 @@
...
+
+        disableScreenRotation();
+        initDisplay();
+        initWeekDays();
+        initOnClick();
+        refresh();
+        setWindowSize();
+
+        return myLayout;
+    }
```
phillipm: I try to order my class methods in terms of usage. I've found that to be a pattern that leads to pretty good readability. 
So for instance, `onCreateView` is one of the most important method, so it comes first in the class. This method calls `N` other methods, which should be defined right below, so that when I read `onCreateView`, I can look right below it to follow what the helper methods are doing. If one of those helper methods makes use of another helper method, I usually put it right below (so it is like a depth first traversal).

app/src/org/commcare/views/widgets/CalendarFragment.java
```diff
@@ -0,0 +1,342 @@
...
+    public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState){
+        getDialog().getWindow().requestFeature(Window.FEATURE_NO_TITLE);
+        inflateView(inflater, container);
+
+        if(savedInstanceState != null && savedInstanceState.containsKey(TIME)){
+            calendar = Calendar.getInstance();
+            calendar.setTimeInMillis(savedInstanceState.getLong(TIME));
+        }
+
+        disableScreenRotation();
```
phillipm: I'm curious, are you disabling screen orientation because it makes everything way easier (a perfectly valid reason) or for some other reason?

app/src/org/commcare/views/widgets/CalendarFragment.java
```diff
@@ -0,0 +1,342 @@
...
+        int totalDays = getNumDaysInMonth(populator);
+        populateListOfDates(dateList, populator, totalDays);
+
+        yearSpinner.setSelection(calendar.get(Calendar.YEAR)-GregorianDateWidget.MINYEAR);
+        monthSpinner.setSelection(calendar.get(Calendar.MONTH));
+        myGrid.setAdapter(new CalendarAdapter(getContext(), dateList));
+    }
+
+    //Populates an arraylist with dates
+    private void populateListOfDates(ArrayList<Date> dateList, Calendar populator, int totalDays) {
```
phillipm: This can be `static`. It would also be a good method to have unit tests for.

app/src/org/commcare/views/widgets/CalendarFragment.java
```diff
@@ -0,0 +1,342 @@
...
+        int remainingDays = ((DAYSINWEEK + 1)-(populator.get(Calendar.DAY_OF_WEEK)))%DAYSINWEEK;
+
+        for(int i = 0; i < remainingDays; i ++){
+            dateList.add(populator.getTime());
+            populator.add(Calendar.DAY_OF_MONTH, 1);
+        }
+    }
+
+    //Calculates the total number of days from most recent Sunday to the first Saturday following the end of the current month to fill the calendar view
+    private int getNumDaysInMonth(Calendar cal){
```
phillipm: This can be `static`

phillipm: Marking methods as `static` is important because they are easier refactor, unit test, and very slightly faster to invoke

app/src/org/commcare/views/widgets/CalendarFragment.java
```diff
@@ -0,0 +1,342 @@
...
+                text.setBackgroundColor(getResources().getColor(R.color.cc_attention_positive_color));
+            }
+            else if(date.getMonth() != current.getMonth()){
+                text.setTextColor(getResources().getColor(R.color.grey_dark));
+                text.setBackgroundColor(getResources().getColor(R.color.grey_lighter));
+            }
+            else{
+                text.setTextColor(getResources().getColor(R.color.black));
+                text.setBackgroundColor(getResources().getColor(R.color.transparent));
+            }
```
phillipm: The color highlight code could be lifted into a helper method called something like `highlightDate`

app/src/org/commcare/views/widgets/CalendarFragment.java
```diff
@@ -0,0 +1,342 @@
...
+            Date current = calendar.getTime();
+
+            if(date.equals(current)){
+                text.setTextColor(getResources().getColor(R.color.white));
+                text.setBackgroundColor(getResources().getColor(R.color.cc_attention_positive_color));
+            }
+            else if(date.getMonth() != current.getMonth()){
+                text.setTextColor(getResources().getColor(R.color.grey_dark));
+                text.setBackgroundColor(getResources().getColor(R.color.grey_lighter));
+            }
```
phillipm: Took me a second to figure out what this code was doing. Would be helpful to have a comment saying "graying out days from other months"

app/src/org/commcare/views/widgets/GregorianDateWidget.java
```diff
@@ -0,0 +1,386 @@
...
+    private Spinner monthSpinner;
+    private ImageButton openCalButton;
+
+    protected List<String> monthList;
+    protected int maxYear;
+    protected long todaysDateInMillis;
+    private long timeBeforeCalendarOpened;
+
+    private CalendarFragment myCalendarFragment;
+    private FragmentManager fm;
```
phillipm: Linter says a lot of these fields can be `final`/`private`

app/src/org/commcare/views/widgets/GregorianDateWidget.java
```diff
@@ -0,0 +1,386 @@
...
+
+        String [] monthNames = new String[12];
+        final Map<String, Integer> monthMap = calendar.getDisplayNames(Calendar.MONTH, Calendar.LONG, Locale.getDefault());
+        monthList = new ArrayList<>(monthMap.keySet());
+            Collections.sort(monthList, new Comparator<String>(){
+                @Override
+                public int compare(String a, String b){
+                    return monthMap.get(a) - monthMap.get(b);
+                }
+            });
```
phillipm: I think you have similar sort code in `CalendarFragment`. Could probably be lifted to a static helper.

phillipm:
💯 📅 👍 👍 awesome work! You did a great job fitting this new feature into our existing infrastructure, which is no easy feat!

app/src/org/commcare/views/widgets/DatePrototypeFactory.java
sjain28: It's still useful while we're making the cancel button optional via appearance attribute

app/src/org/commcare/views/widgets/CalendarFragment.java
sjain28: Just because it makes everything way easier with these fragments.

phillipm:
🎉 🎆 🎉

PR #1294, opened 2016-06-01
Stop using reflection to serialize objects

phillipm:
Seeing hundreds of `Caused by: java.lang.IllegalAccessException: access to field not allowed` ACRA crashes on Android 4 devices. Looks like a lot of devices from the HSPH PPIUD project are affected, which might be causing data loss issues.

I wasn't able to reproduce the issue so opting to remove use of reflection, which adds a good chunk of code but should be more performant.


ctsims:
Ooof, I'm really not super thrilled about the maintenance overhead of this.

Note that many other aspects of the runtime function off of reflection (Namely the UI callback framework), so if reflection is straight-broken the app just won't run.

Reflection is an Android Java Language feature, I really don't want to kill our ability to use it unless we're getting actual bug reports.

phillipm:
Crashes are specific to using `java.lang.reflect.Field.set`, so other instances of reflection should be fine.

While this does introduce more code, it is consistent with patterns used by the javarosa and commcare codebases, so we should be able to maintain it in a similar fashion.

ctsims:
Might be helpful for us to talk through strategy long term after the j2me split. The serialization code is a dinosaur from a different time and once we are no longer beholden to some of its restrictions we may find ourselves able to do significantly better.

Can we reach out to HSPH PPIUD and actually dig into who is seeing these when?

phillipm:
Closing because a less intrusive fix was found: https://github.com/dimagi/commcare-odk/pull/1303

PR #1303, opened 2016-06-06
Synchronize Persisted's Field.setField calls

phillipm:
Fixes serialization crashes of the following nature that seem to be caused by object serialization synchronization issues when multiple threads are involved. Specifically, since we cache class fields that we do reflection on, issues arise if one thread sets a field to accessible, and before it can access the field another thread sets it to inaccessible.

```
org.commcare.android.adapters.IncompleteFormListAdapter.resetRecords(IncompleteFormListAdapter.java:164)
at org.commcare.dalvik.activities.FormRecordListActivity.refreshView(FormRecordListActivity.java:312)
at org.commcare.dalvik.activities.FormRecordListActivity.onCreate(FormRecordListActivity.java:223)
at android.app.Activity.performCreate(Activity.java:5451)
at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1093)
at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2377)
... 11 more
Caused by: org.javarosa.core.util.externalizable.DeserializationException: for fieldappId
at org.commcare.android.storage.framework.Persisted.readExternal(Persisted.java:41)
at org.commcare.android.database.SqlStorage.newObject(SqlStorage.java:227)
... 19 more
Caused by: java.lang.IllegalAccessException: access to field not allowed
at java.lang.reflect.Field.setField(Native Method)
at java.lang.reflect.Field.set(Field.java:585)
at org.commcare.android.storage.framework.Persisted.readVal(Persisted.java:107)
at org.commcare.android.storage.framework.Persisted.readExternal(Persisted.java:38)
... 20 more
```

https://github.com/dimagi/commcare-odk/pull/1298/files#diff-b72fa8b3bcb18ed59a32788845cc84b1R272 reproduces the crash above and I've verified in that PR that these changes prevent the crash.

Since this is a timing issue, it is hard to test. But before hotfixing we should verify that the following work fine:
 - [x] Opening forms while old form purging is running in the background
 - [x] Opening a form while a form submission is happening in the background
 - [x] No noticeable slow-down loading a hundred saved forms.

### DIFF
--- a/app/src/org/commcare/android/storage/framework/Persisted.java
+++ b/app/src/org/commcare/android/storage/framework/Persisted.java
@@ -51,8 +51,7 @@ public class Persisted implements Persistable, IMetaData {
                 orderings = new ArrayList<>();
                 fieldOrderings.put(this.getClass(), orderings);
             }
-        }
-        synchronized (orderings) {
+
             if (orderings.size() == 0) {
                 for (Field f : this.getClass().getDeclaredFields()) {
                     if (f.isAnnotationPresent(Persisting.class)) {
--- a/app/src/org/commcare/android/storage/framework/Persisted.java
+++ b/app/src/org/commcare/android/storage/framework/Persisted.java
@@ -176,6 +176,7 @@ public class Persisted implements Persistable, IMetaData {
         ArrayList<String> fields = new ArrayList<>();
 
         for (Field f : this.getClass().getDeclaredFields()) {
+            synchronized (f) {
             try {
                 f.setAccessible(true);
 
@@ -186,10 +187,11 @@ public class Persisted implements Persistable, IMetaData {
             } finally {
                 f.setAccessible(false);
             }
-
+            }
         }
 
         for (Method m : this.getClass().getDeclaredMethods()) {
+            synchronized (m) {
             try {
                 m.setAccessible(true);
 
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/android/storage/framework/Persisted.java
```diff
@@ -98,6 +97,7 @@ public int getID() {
     }
 
     private void readVal(Field f, Object o, DataInputStream in) throws DeserializationException, IOException, IllegalAccessException {
+        synchronized (fieldOrderings) {
```
ctsims: You should be able to just synchronize 'f' here instead, which should protect any of these methods which are trying to manipulate that particular field while still allowing others to be read/written.

phillipm: Updated and updated on https://github.com/dimagi/commcare-odk/pull/1304 where I also ran the tests for the fix.

phillipm: Also note that making your suggested changes means we can't also correctly synchronize over the metadata field access since those aren't cached.

phillipm: Ah, realized that if we assume that `Class.getDeclaredFields()` always returns the same `Field` instances, then we can synchronize over them. Did this in 
https://github.com/dimagi/commcare-odk/pull/1303/commits/4204c98579f9f1b5c3406a370fb98dcba697881b , should be good to go.

PR #1304, opened 2016-06-06
2.28 version of Synchronize Persisted's Field.setField calls

phillipm:
2.28 version of https://github.com/dimagi/commcare-odk/pull/1303
Includes test that fails before the fix is applied and code indentation, so ?w=1

The last commit fixes the debug feature that allows for loading saved forms onto the device from a payload.

### DIFF
--- a/app/res/xml/preferences_developer.xml
+++ b/app/res/xml/preferences_developer.xml
@@ -99,7 +99,7 @@
         android:key="cc-auto-purge"
         android:title="Detect and Trigger Purge on Form Save"/>
     <ListPreference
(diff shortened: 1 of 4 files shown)


app/res/xml/preferences_developer.xml
```diff
@@ -99,7 +99,7 @@
         android:key="cc-auto-purge"
         android:title="Detect and Trigger Purge on Form Save"/>
     <ListPreference
-        android:defaultValue="yes"
+        android:defaultValue="saved"
```
amstone326: what is this doing?

PR #1305, opened 2016-06-07
Fix crash after uninstall

amstone326:


amstone326:
only the last commit is needed, opening PR for simpler solution

PR #1306, opened 2016-06-07
Fix session unavailable crash after uninstall

amstone326:
@phillipm Embarrassed that it took me so long to figure this out, but here goes: the problem was that the `unredirectedSessionExpiration` getting registered by the app uninstall was getting wiped out by the call to `System.exit()` in `restartCommCare`. I played around a bit and realized that it was fine to remove this in the case of uninstalling, as long as I also made sure that the app that's being uninstalled gets unseated first, if it is the currently seated app.


phillipm:
Sounds tricky! Glad the fix ended up being straight forward

PR #1307, opened 2016-06-08
Don't modify incoming object in list union code

phillipm:
Case loading code was caching a Vector instance, which was then getting modified when it was union'ed with another list. The fix is to make the union code not modify the incoming objects.

Fix for http://manage.dimagi.com/default.asp?229108

### DIFF
--- a/app/src/org/commcare/adapters/MenuAdapter.java
+++ b/app/src/org/commcare/adapters/MenuAdapter.java
@@ -118,12 +118,12 @@ public class MenuAdapter implements ListAdapter {
     private void addRelevantCommandEntries(Menu m, Vector<MenuDisplayable> items,
                                            Hashtable<String, Entry> map)
             throws XPathSyntaxException {
+        EvaluationContext ec = asw.getEvaluationContext();
         for (String command : m.getCommandIds()) {
             errorXpathException = "";
             XPathExpression mRelevantCondition = m.getCommandRelevance(m.indexOfCommand(command));
             if (mRelevantCondition != null) {
                 errorXpathException = m.getCommandRelevanceRaw(m.indexOfCommand(command));
-                EvaluationContext ec = asw.getEvaluationContext();
                 Object ret = mRelevantCondition.eval(ec);
                 try {
                     if (!XPathFuncExpr.toBoolean(ret)) {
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/adapters/MenuAdapter.java
```diff
@@ -135,9 +135,6 @@ private void addRelevantCommandEntries(Menu m, Vector<MenuDisplayable> items,
                     Logger.log(AndroidLogger.TYPE_ERROR_CONFIG_STRUCTURE, msg);
                     throw new RuntimeException(msg);
                 }
-                if (!XPathFuncExpr.toBoolean(ret)) {
-                    continue;
-                }
```
phillipm: Removing this because I have no idea what it would do given https://github.com/dimagi/commcare-odk/pull/1307/files#diff-bc3a12589746386b2d565b70c55f4adfL129

app/src/org/commcare/adapters/MenuAdapter.java
```diff
@@ -118,12 +118,12 @@ private boolean menuIsRelevant(Menu m) throws XPathSyntaxException {
     private void addRelevantCommandEntries(Menu m, Vector<MenuDisplayable> items,
                                            Hashtable<String, Entry> map)
             throws XPathSyntaxException {
+        EvaluationContext ec = asw.getEvaluationContext();
         for (String command : m.getCommandIds()) {
             errorXpathException = "";
             XPathExpression mRelevantCondition = m.getCommandRelevance(m.indexOfCommand(command));
             if (mRelevantCondition != null) {
                 errorXpathException = m.getCommandRelevanceRaw(m.indexOfCommand(command));
-                EvaluationContext ec = asw.getEvaluationContext();
```
phillipm: Moving this out of the loop, don't see why it should be in the loop body

app/src/org/commcare/utils/AndroidUtil.java
```diff
@@ -65,9 +66,8 @@ public static void initializeStaticHandlers() {
 
             joined.retainAll(other);
 
-            a.clear();
-            a.addAll(joined);
-            return a;
+            result.addAll(joined);
+            return result;
```
phillipm: The fix

PR #1310, opened 2016-06-09
Localize repeat dialog

phillipm:
Allow repeat dialog to be localized in standard fashion.

| Before add first group | After add first group |
| --- | --- |
| a_new_repeat | b_new_repeat |

| Before add another group | After add another group |
| --- | --- |
| a_another_repeat | b_another_repeat | 

Follow-up for http://manage.dimagi.com/default.asp?229190


amstone326:
@phillipm Can you delete all of the now-unused strings from `strings.xml`?

phillipm:
@amstone326 I've removed the old repeat strings

PR #1313, opened 2016-06-09
Fix login redirection bug

amstone326:
After a lot of debugging and scratching our heads, Clayton and I finally decided a while back that this was the result of an Android-level bug, in which they are not adhering to the contract that an activity will not resume until the activity that it has called out to with `startActivityForResult()` has in fact returned a result.

The fix is to manually make sure that we don't relaunch the LoginActivity if we are currently waiting for a result from the LoginActivity.

Added the 'don't pull' label until we collectively agree that we want to go this route. @ctsims @phillipm

### DIFF
--- a/app/src/org/commcare/activities/DispatchActivity.java
+++ b/app/src/org/commcare/activities/DispatchActivity.java
@@ -54,7 +54,6 @@ public class DispatchActivity extends FragmentActivity {
     private static final String KEY_APP_FILES_CHECK_OCCURRED = "check-for-changed-app-files-occurred";
     private static final String KEY_WAITING_FOR_ACTIVITY_RESULT = "waiting-for-login-activity-result";
 
-    // Used for soft assert for login redirection bug
     private boolean waitingForActivityResultFromLogin;
 
     boolean alreadyCheckedForAppFilesChange;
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/activities/DispatchActivity.java
```diff
@@ -212,15 +214,17 @@ private void createNoStorageDialog() {
     }
 
     private void launchLoginScreen() {
-        if (waitingForActivityResultFromLogin) {
-            Logger.log(AndroidLogger.SOFT_ASSERT, "Login redirection bug occurred; " +
-                    "DispatchActivity is attempting to launch a new LoginActivity while it is " +
-                    "still waiting for a result from another one.");
+        if (!waitingForActivityResultFromLogin) {
```
phillipm: Would be helpful to mention this is needed due to a Android platform bug

phillipm:
👍 looks good to me

amstone326:
@ctsims If you're ok with this as well then it can get pulled.

amstone326:
Release Note: Fix occasional bug where a successful login redirected to the login screen instead of home screen

PR #1315, opened 2016-06-10
Audio prototype

sjain28:


sjain28:
Widget opens to this screen: 
widget home screen

After pressing "Start Recording", this dialog opens: 
start recording

During recording: 
during recording

During audio playback: 
during playback

When playback is paused: 
playback paused

biyeun:
Some style suggestions @sjain28 
cc @dimagi/product @wpride 
audio-01
audio-02
audio-03
audio-04
audio-05
audio-06

wpride:
@biyeun strikes again!

ctsims:
@biyeun 

This looks awesome! 

Two questions: 

1) What's the purpose of the string name for the widget? In general our users won't really have any meaning behind a filename, so it seems like it's most likely to be confusing.

2) We didn't convey this super clearly, but the "Choose File" option is optional, and honestly generally probably not desirable for most audio recording workflows. Would you recommend just eliminating the button when it's gone or is there something we should do different for symetry/whitespace balance?

biyeun:
@ctsims thanks for the feedback. 

1) the intention of the string name was to give the user some visual feedback that audio is present from recording or uploading. 

2) Since the uploader is optional, I might have to revisit that part of the UI

sjain28:
@amstone326 Google analytics added in the most recent commit

ctsims:
@biyeun Ah, that makes sense, having the ability to spot that there is audio there and recorded in a clear (center stage) way seems important. @sjain28

sjain28:
@biyeun, do you have an idea of what the main widget would look like without the choose file button? 
According to @ctsims, we will probably make that button optional.

biyeun:
@sjain28 I will block off time to take a look at this tomorrow

app/res/drawable/progress_bar.xml
```diff
@@ -0,0 +1,20 @@
...
+        android:innerRadiusRatio="3"
+        android:thicknessRatio="14">
+
+        <size
+            android:height="200dp"
+            android:width="200dp"/>
+
+        <gradient
+            android:type="sweep"
+            android:startColor="#00000000"
```
wpride: think this can just be "@color/black"

app/src/org/commcare/views/widgets/AudioPrototype.java
```diff
@@ -0,0 +1,211 @@
...
+import org.commcare.logic.PendingCalloutInterface;
+import org.commcare.utils.StringUtils;
+import org.javarosa.core.services.locale.Localization;
+import org.javarosa.form.api.FormEntryPrompt;
+
+/**
+ * Created by Saumya on 6/3/2016.
+ * An alternative audio widget that records and plays audio natively without callout to any external application
+ */
+public class AudioPrototype extends AudioWidget implements RecordingFragment.RecordingCompletionListener {
```
wpride: Probably can rename this CommCareAudioWidget, probably past the prototype phase

app/src/org/commcare/views/widgets/AudioPrototype.java
```diff
@@ -0,0 +1,211 @@
...
+        setBinaryData(recorder.getFileName());
+        mPlayButton.setEnabled(true);
+        mPlayButton.setBackgroundResource(R.drawable.play);
+        recordingNameText.setTextColor(getResources().getColor(R.color.black));
+        recordingNameText.setText(Localization.get("recording.custom"));
+    }
+
+    @Override
+    protected void playAudio() {
+        Uri myPath = Uri.parse(mInstanceFolder + mBinaryName);
```
wpride: NIT: myPath -> filePath

app/src/org/commcare/views/widgets/AudioPrototype.java
```diff
@@ -0,0 +1,211 @@
...
+        final MediaPlayer player = MediaPlayer.create(getContext(), myPath);
```
wpride: Q: why do we create the media player here, then pass it around/re-declare it in various onClick methods? Does it not work to declare a final class-level MediaPlayer and have every method use that?

app/src/org/commcare/views/widgets/AudioPrototype.java
```diff
@@ -0,0 +1,211 @@
...
+            mPlayButton.setBackgroundResource(R.drawable.play_disabled);
+        }
+        mPlayButton.setEnabled(enabled);
+    }
+
+    @Override
+    protected void reloadFile(){
+        super.reloadFile();
+        recordingNameText.setTextColor(getResources().getColor(R.color.black));
+        if(mBinaryName.contains(CUSTOM_TAG)){
```
wpride: not clear what this is doing - this is just the default name if they don't specify one, in which case we use the localization string?

app/src/org/commcare/views/widgets/RecordingFragment.java
```diff
@@ -0,0 +1,330 @@
...
+
+/**
+ * Created by Saumya on 6/15/2016.
+ * A popup dialog fragment that handles recording_fragment and saving of audio files without external callout
+ */
+
+public class RecordingFragment extends android.support.v4.app.DialogFragment{
+
+    private String fileName;
+    private final String FILE_EXT = "/Android/data/org.commcare.dalvik/temp/Custom Recording.mp4";
```
wpride: Might make the space an underscore, spaces are handled weirdly on some OSes

app/src/org/commcare/views/widgets/RecordingFragment.java
```diff
@@ -0,0 +1,330 @@
...
+    private RecordingCompletionListener listener;
+    private MediaPlayer player;
+    
+    public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState){
+        layout = (LinearLayout) inflater.inflate(R.layout.recording_fragment, container);
+        disableScreenRotation();
+        prepareButtons();
+        prepareText();
+        setWindowSize();
+        fileName = Environment.getExternalStorageDirectory().getAbsolutePath() + FILE_EXT;
```
wpride: If you have two audioprototype widgets, this looks like it would record over the first one?

app/src/org/commcare/views/widgets/RecordingFragment.java
```diff
@@ -0,0 +1,330 @@
...
+        toggleRecording.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                GoogleAnalyticsUtils.reportAudioPaused();
+                pauseAudioPlayer(player);
+            }
+        });
+    }
+
+    private void pauseAudioPlayer(MediaPlayer player){
```
wpride: (same question about the MediaPlayer passing)

wpride:
this PR looks really great @sjain28 👍

app/src/org/commcare/views/widgets/AudioPrototype.java
sjain28: @wpride we want to display "Custom Recording" if they load a file recorded in the widget, and the file name if they choose a file from the device. 
This is checking if it's a custom recording.

app/res/drawable/progress_bar.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
```
phillipm: Can you rename this file to mention audio/recording?

app/src/org/commcare/logging/analytics/GoogleAnalyticsFields.java
```diff
@@ -22,6 +22,17 @@
...
+
+    //Actions for CATEGORY_AUDIO_WIDGET only
+    public static final String ACTION_START_RECORDING_DIALOG = "Click the button to open recording popup";
+    public static final String ACTION_CHOOSE_FILE = "Click the button to choose an audio file";
+    public static final String ACTION_START_RECORD = "Click the button to start recording audio";
+    public static final String ACTION_STOP_RECORD = "Click the button to stop recording audio";
+    public static final String ACTION_SAVE_RECORDING = "Click the button to save an audio recording";
+    public static final String ACTION_PLAY_AUDIO = "Click the button to play audio";
+    public static final String ACTION_PAUSE_AUDIO = "Click the button to pause audio playback";
+    public static final String ACTION_RECORD_AGAIN = "Click the button to Record Again";
```
phillipm: Nice!

app/src/org/commcare/views/widgets/AudioPrototype.java
```diff
@@ -0,0 +1,211 @@
...
+import org.javarosa.form.api.FormEntryPrompt;
+
+/**
+ * Created by Saumya on 6/3/2016.
+ * An alternative audio widget that records and plays audio natively without callout to any external application
+ */
+public class AudioPrototype extends AudioWidget implements RecordingFragment.RecordingCompletionListener {
+
+    private RecordingFragment recorder;
+    private FragmentManager fm;
```
phillipm: These two fields can be `final`

app/src/org/commcare/views/widgets/AudioPrototype.java
```diff
@@ -0,0 +1,211 @@
...
+        // launch audio filechooser intent on click
+        mChooseButton.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                GoogleAnalyticsUtils.reportAudioFileChosen();
+                Intent i = new Intent(Intent.ACTION_GET_CONTENT);
+                i.setType("audio/*");
+                try {
+                    ((Activity)getContext()).startActivityForResult(i, FormEntryActivity.AUDIO_VIDEO_FETCH);
+                    recordingNameText.setTextColor(getResources().getColor(R.color.black));
```
phillipm: `getColor` is deprecated

build.gradle
```diff
@@ -68,6 +68,7 @@ dependencies {
     }
     compile 'org.apache.httpcomponents:httpclient-android:4.3.5.1'
     compile 'net.sourceforge.htmlcleaner:htmlcleaner:2.16'
+    compile 'com.liulishuo.magicprogresswidget:library:1.0.9'
```
phillipm: Doesn't seem to be used

app/src/org/commcare/views/widgets/AudioPrototype.java
```diff
@@ -0,0 +1,211 @@
...
+        recorder.setListener(this);
+    }
+
+    @Override
+    protected void initializeButtons(final FormEntryPrompt prompt){
+        LayoutInflater vi = (LayoutInflater)getContext().getSystemService(Context.LAYOUT_INFLATER_SERVICE);
+        layout = (LinearLayout) vi.inflate(R.layout.audio_prototype, null);
+
+        mPlayButton = (ImageButton) layout.findViewById(R.id.play_audio);
+        ImageButton mCaptureButton = (ImageButton) layout.findViewById(R.id.capture_button);
```
phillipm: variables that are local to a method should not be prefixed with `m`

phillipm:
Some high level usage questions: have you tried rotating the screen or locking the phone while recording or in playback? Have you tried moving to the next question screen while playing back audio?

app/src/org/commcare/views/widgets/AudioWidget.java
```diff
@@ -38,18 +38,24 @@
...
+    protected String recordedFileName;
+    protected String mBinaryName;
+    protected final String mInstanceFolder;
+    protected String customFileTag;
 
     public AudioWidget(Context context, final FormEntryPrompt prompt, PendingCalloutInterface pic) {
         super(context, prompt);
 
+        initializeButtons(prompt);
+        setupLayout();
```
phillipm: Nice, thanks for breaking up this constructor into helpers

app/src/org/commcare/views/widgets/RecordingFragment.java
phillipm: Should be `static`

app/src/org/commcare/views/widgets/RecordingFragment.java
```diff
@@ -0,0 +1,330 @@
...
+ */
+
+public class RecordingFragment extends android.support.v4.app.DialogFragment{
+
+    private String fileName;
+    private final String FILE_EXT = "/Android/data/org.commcare.dalvik/temp/Custom Recording.mp4";
+
+    private LinearLayout layout;
+    private ImageButton toggleRecording;
+    private ImageButton discardRecording;
```
phillipm: Doesn't need to be class level field, can be moved into `prepareButtons`

app/src/org/commcare/views/widgets/RecordingFragment.java
```diff
@@ -0,0 +1,330 @@
...
+    private ProgressBar recordingProgress;
+
+    private Chronometer recordingDuration;
+    private long currentTimeMillis;
+
+    private MediaRecorder recorder;
+    private RecordingCompletionListener listener;
+    private MediaPlayer player;
+    
+    public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState){
```
phillipm: `@Override`

app/src/org/commcare/views/widgets/RecordingFragment.java
```diff
@@ -0,0 +1,330 @@
...
+            toggleRecording.setBackgroundResource(R.drawable.play);
+            toggleRecording.setOnClickListener(new View.OnClickListener() {
+                @Override
+                public void onClick(View v) {
+                    playAudio();
+                }
+            });
+            saveRecording.setEnabled(true);
+            instruction.setText(Localization.get("after.recording"));
+        }
```
phillipm: the 'record again' setup logic should get lifted into a helper

app/src/org/commcare/views/widgets/RecordingFragment.java
```diff
@@ -0,0 +1,330 @@
...
+            recorder.release();
+            this.recorder = null;
+        }
+    }
+
+    public String getFileName(){
+        return fileName;
+    }
+
+    private void disableScreenRotation() {
```
phillipm: This can be made into a `static` method that takes in the activity as an argument. That way it can be reused in the calendar code as well

app/res/drawable/progress_bar.xml
sjain28: It's not limited to audio recording - Could be used as a better looking alternative to the default Android progress bar anywhere.

sjain28:
@phillipm Yep, have tried all of that.

app/src/org/commcare/views/widgets/AudioPrototype.java
sjain28: @phillipm it's deprecating starting at API 23 - Can I continue to use it since we target level 9?

phillipm: If you change it to `getColor(R.color.black, null)` you should be good. While in some cases it isn't a huge deal to call deprecated code, it is generally good to avoid it. Calls to the Android platform can break in really crazy ways and sometimes that is due to using deprecated calls, so always good to rule it out when possible.

PR #1316, opened 2016-06-10
Remove red text on blue sync button

amstone326:
Addresses http://manage.dimagi.com/default.asp?190347. Make warning text on sync button orange (Biyeun's suggestion) instead of red.


wpride:
:fire: :tight: 🔥

PR #1317, opened 2016-06-13
Fix confusing behavior when enabling developer options

amstone326:
Phillip recently added an OK button to the About CommCare dialog, but pressing it did not contribute to the count for enabling Developer Options. This PR makes it so that either pressing OK or pressing 'back' contribute to the count (canceling a dialog calls both `onCancel` and `onDismiss`, so placing the code in the `OnDismissListener` only is sufficient). 

Also fix a small bug in `CommCareAlertDialog`

### DIFF
--- a/app/src/org/commcare/views/dialogs/CommCareAlertDialog.java
+++ b/app/src/org/commcare/views/dialogs/CommCareAlertDialog.java
@@ -19,8 +19,7 @@ public abstract class CommCareAlertDialog {
     protected DialogInterface.OnCancelListener cancelListener;
     private DialogInterface.OnDismissListener dismissListener;
     protected View view;
-    // false by default, can be overridden by calling setOnCancelListener(), or by subclass
-    // definitions if desired
+    // false by default, can be overridden if desired
     protected boolean isCancelable;
 
     public void finalizeView() {
@@ -28,6 +27,7 @@ public abstract class CommCareAlertDialog {
         if (isCancelable) {
             dialog.setOnCancelListener(cancelListener);
         }
+        dialog.setOnDismissListener(dismissListener);
         dialog.setView(view);
     }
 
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/views/dialogs/CommCareAlertDialog.java
```diff
@@ -47,13 +51,16 @@ public boolean isCancelable() {
...
+        isCancelable = true;
+    }
+
     public void setOnCancelListener(DialogInterface.OnCancelListener listener) {
         isCancelable = true;
         cancelListener = listener;
     }
 
     public void setOnDismissListener(DialogInterface.OnDismissListener listener) {
-        dialog.setOnDismissListener(listener);
```
amstone326: having this here causes a crash in some instances, since in `CustomViewAlertDialog` the dialog is not initialized until we show it; changed to follow the same pattern as `setOnCancelListener` where we just set the class field here, and then assign it to the dialog in `finalizeView()`

app/src/org/commcare/views/dialogs/CommCareAlertDialog.java
```diff
@@ -19,16 +19,20 @@
...
     protected View view;
-    // false by default, can be overridden by calling setOnCancelListener(), or by subclass
-    // definitions if desired
+    // false by default, can be overridden if desired
     protected boolean isCancelable;
 
     public void finalizeView() {
         dialog.setCancelable(isCancelable);
-        if (isCancelable) {
+        if (cancelListener != null) {
```
wpride: seems slightly redundant having both cancelListener and isCalcelable when you're checking cancelListener != null - is there a case where the dialog is cancelable but there isn't a listener?

amstone326: Yea, the use-case for that was actually added in this PR. Line 1284 of `CommCareHomeActivity` above calls `makeCancelable()`, but doesn't set an `onCancelListener()`. I did it this way because when a dialog is canceled by the user, BOTH `onCancel()` and `onDismiss()`actually get called (the reverse is not true). So I wanted to only add an `onDismissListener` to the About CommCare dialog, to prevent the logic from getting called twice in the case of the dialog being canceled. But I still want the dialog to be cancelable, hence allowing a dialog to be cancelable without a listener.

phillipm:
Thanks for fixing this!

PR #1318, opened 2016-06-13
Refactor DataPullTask

amstone326:
Prefactor for async restores work

This is a pretty large refactor in terms of how different the code looks now, so if whoever is reviewing it wants to sit down and talk about any of the changes to make sure they do in fact preserve the original behavior, I'm happy to do that.


app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -102,258 +101,296 @@ public DataPullTask(String username, String password,
...
+            this.publishProgress(PROGRESS_DONE);
+            return new ResultAndError<>(PullTaskResult.UNKNOWN_FAILURE,
+                    "Unable to get or generate encryption key");
+        } else {
+            factory.initUserParser(wrappedEncryptionKey);
         }
 
-        // Wrap in a 'try' to enable a 'finally' close that releases the
-        // sessionAliveLock.
+        if (isCancelled()) {
```
amstone326: Does anyone know if this check can come earlier? In the old code, this came after the initialization of the transaction parser factory and the encryption key so I left it here, but I'm not sure why that's necessary.

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -102,258 +101,296 @@ public DataPullTask(String username, String password,
...
-
-            boolean loginNeeded = true;
-            boolean useRequestFlags = false;
-            try {
-                loginNeeded = !CommCareApplication._().getSession().isActive();
-            } catch (SessionUnavailableException sue) {
-                //expected if we aren't initialized.
+            ResultAndError<PullTaskResult> resultFromRequest = makeRequestAndHandleResponse(factory);
+            if (resultFromRequest != null) {
+                return resultFromRequest;
```
amstone326: This is a kind of odd pattern that I introduced in order to preserve the precise behavior in the original code, in terms of when an error result was returned. In the old code, some errors would return a failure result immediately, while others would be allowed to fall through to the code on lines 174-176 below. Since I couldn't tell if that original design was intentional or incidental, I didn't want to change it. So any errors that were falling through to those lines in the old code now return null in their helper methods so that it will fall through here.

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -102,258 +101,296 @@ public DataPullTask(String username, String password,
...
+
+    private void initUKRForLogin() {
+        if (shouldGenerateFirstKey()) {
+            SecretKey newKey = CryptUtil.generateSemiRandomKey();
+            if (newKey == null) {
+                return;
             }
+            String sandboxId = PropertyUtils.genUUID().replace("-", "");
+            ukrForLogin = new UserKeyRecord(username, UserKeyRecord.generatePwdHash(password),
+                    newKey.getEncoded(), new Date(), new Date(Long.MAX_VALUE), sandboxId);
```
amstone326: This line represents an actual change from the old code https://github.com/dimagi/commcare-odk/blob/master/app/src/org/commcare/tasks/DataPullTask.java#L181, where we were wrapping the encrypted key before passing it into the `UserKeyRecord` constructor. I removed that operation here because I'm pretty sure that that was incorrect; the `UserKeyRecord` constructor does not indicate that it expects a wrapped key, and we perform the wrapping function again on line 209 above after we extract it from the ukr. But @ctsims you should double check and confirm.

amstone326:
@ctsims I'm assigning this to you because this is a pretty substantial refactor of very essential code. There is now pretty good test coverage of this area of the code, thanks to @phillipm creating the `DataPullTaskTests`, but I would still feel better if this got a sanity check from you.

app/src/org/commcare/tasks/DataPullTask.java
ctsims: I think it's best practice to check for cancellation as close to before the "blocking" operation as possible to ensure the user has the maximum amount of time to 'undo' the operation, so moving it earlier prevents that from being an option

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -102,258 +101,296 @@ public DataPullTask(String username, String password,
...
+        byte[] key;
+        if (loginNeeded) {
+            initUKRForLogin();
+            if (ukrForLogin == null) {
+                return null;
+            }
+            key = ByteEncrypter.wrapByteArrayWithString(ukrForLogin.getEncryptedKey(), password);
+        } else {
+            //Only purge cases if we already had a logged in user. Otherwise we probably can't read the DB.
+            CaseUtils.purgeCases();
```
ctsims: I don't think this is the best place for this call now. We need this to happen before we perform a sync request for sure, but if we've refactored this modularly in this process, I think we should call this after getEncryptionKey() has returned with a key outside of this method.

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -102,258 +101,296 @@ public DataPullTask(String username, String password,
...
+
+        handleLoginNeededOnSuccess();
+        this.publishProgress(PROGRESS_AUTHED, 0);
+
+        if (isCancelled()) {
+            // About to enter data commit phase; last chance to finish early if cancelled.
+            return new ResultAndError<>(PullTaskResult.UNKNOWN_FAILURE, "");
+        }
+
+        this.publishProgress(PROGRESS_DOWNLOADING_COMPLETE, 0);
```
ctsims: This should come after pullResponse.writeResponseToCache() executes

ctsims: hm, oh, probably wrong here, actually, I imagine it's 0 -> 100 as the second arg, nevermind

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -102,258 +101,296 @@ public DataPullTask(String username, String password,
...
+            String syncToken = readInput(cache.retrieveCache(), factory);
+            updateUserSyncToken(syncToken);
+
+            onSuccessfulSync();
+            return new ResultAndError<>(PullTaskResult.DOWNLOAD_SUCCESS);
+        } catch (XmlPullParserException e) {
+            e.printStackTrace();
+            Logger.log(AndroidLogger.TYPE_USER,
+                    "User Sync failed due to bad payload|" + e.getMessage());
+            return new ResultAndError<>(PullTaskResult.BAD_DATA, e.getMessage());
```
ctsims: Don't these need to reverse the logout if it occured?

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -102,258 +101,296 @@ public DataPullTask(String username, String password,
...
+        } catch (UnfullfilledRequirementsException e) {
+            e.printStackTrace();
+            Logger.log(AndroidLogger.TYPE_ERROR_ASSERTION,
+                    "User sync failed oddly, unfulfilled reqs |" + e.getMessage());
+            return null;
+        } catch (IllegalStateException e) {
+            e.printStackTrace();
+            Logger.log(AndroidLogger.TYPE_ERROR_ASSERTION,
+                    "User sync failed oddly, ISE |" + e.getMessage());
+            return null;
```
ctsims: A bit confused as to why this and the one above are nulls

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -102,258 +101,296 @@ public DataPullTask(String username, String password,
...
+        } else if (returnCode == PROGRESS_RECOVERY_FAIL_SAFE || returnCode == PROGRESS_RECOVERY_FAIL_BAD) {
+            wipeLoginIfItOccurred();
             this.publishProgress(PROGRESS_DONE);
-            return new ResultAndError<>(responseError);
-        } finally {
-            CommCareSessionService.sessionAliveLock.unlock();
+            return new ResultAndError<>(PullTaskResult.UNKNOWN_FAILURE, failureReason);
+        } else {
+            wipeLoginIfItOccurred();
+            return null;
```
ctsims: Should this return an unknown error?

app/src/org/commcare/tasks/DataPullTask.java
ctsims: There are a couple of reasons for why it was split up, but I believe the most pressing was simply to unify the cleanup code

app/src/org/commcare/tasks/DataPullTask.java
amstone326: Ah cool, makes a lot of sense

app/src/org/commcare/tasks/DataPullTask.java
amstone326: Yep, the ukr gets written in `onSuccessfulSync()`. 

I'm also not sure why that line has been working because I thought the same thing. Is it possible that code path doesn't get followed very much or at all?.. We should probably discuss this one a bit more in person

app/src/org/commcare/tasks/DataPullTask.java
amstone326: Cool, will move

app/src/org/commcare/tasks/DataPullTask.java
amstone326: Hm, I just copied this over exactly from the existing code. You're saying you think it should be 100 instead of 0 anyway?

app/src/org/commcare/tasks/DataPullTask.java
amstone326: Hm yea, none of them were doing so in the original code, but definitely seems like it would make sense to add that.

app/src/org/commcare/tasks/DataPullTask.java
amstone326: For the reason I explained above - in the original code (https://github.com/dimagi/commcare-odk/blob/master/app/src/org/commcare/tasks/DataPullTask.java#L278-L284) neither of these blocks return anything, causing them to fall through to the return code at the end. If you think that wasn't intentional or important, I can have them return something here.

app/src/org/commcare/tasks/DataPullTask.java
amstone326: Same reasoning as above. I can explain in person if I'm not doing a good job of it here.

amstone326:
@ctsims Implemented the changes that were straightforward. We should meet quickly to chat about the other issues.

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -102,258 +101,302 @@ public DataPullTask(String username, String password,
...
+            return new ResultAndError<>(PullTaskResult.DOWNLOAD_SUCCESS);
+        } else if (returnCode == PROGRESS_RECOVERY_FAIL_SAFE || returnCode == PROGRESS_RECOVERY_FAIL_BAD) {
+            wipeLoginIfItOccurred();
             this.publishProgress(PROGRESS_DONE);
-            return new ResultAndError<>(responseError);
-        } finally {
-            CommCareSessionService.sessionAliveLock.unlock();
+            return new ResultAndError<>(PullTaskResult.UNKNOWN_FAILURE, failureReason);
+        } else {
+            wipeLoginIfItOccurred();
```
amstone326: remove this line

amstone326:
er, hold up, build is failing

amstone326:
ok, we're good

PR #1319, opened 2016-06-13
Consumer apps tweaks

amstone326:
1. Make generating the package identifier from the app name robust to app names that contain illegal characters
2. Add a mechanism to the build script for building only some of the apps in the consumer_apps directory

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -516,13 +516,19 @@ def numbersToLetters(String version) {
 
 def getStandalonePackageIdentifier(ccDomainSafe, isConsumerApp) {
     if (project.hasProperty('application_name')) {
-        return ccDomainSafe + "." + application_name.replaceAll(" ", "").toLowerCase();
+        return ccDomainSafe + "." + cleanseAppNameForPackageIdentifier(application_name)
     } else if ("true".equals(isConsumerApp)) {
         throw new InvalidUserDataException("An application_name property must be provided for consumer app standalone builds")
     } else {
         return ccDomainSafe;
     }
 }
+
+
+def cleanseAppNameForPackageIdentifier(appName) {
+    return appName.replaceAll("[ '.:]", "").replaceAll("\\\\", "").toLowerCase()
+}
+
 
 def getStandaloneApplicationName(ccDomainSafe, isConsumerApp) {
     if (project.hasProperty('application_name')) {
--- a/scripts/consumer_apps_builder.py
+++ b/scripts/consumer_apps_builder.py
@@ -38,10 +38,16 @@ def checkout_or_update_static_resources_repo():
     os.chdir('../')
 
 
-def build_apks_from_resources(build_type):
+def build_apks_from_resources(build_type, subset_of_apps_to_build):
     for (app_dir_name, sub_dir_list, files_list) in os.walk(PATH_TO_STATIC_RESOURCES_DIR):
         if '.git' not in app_dir_name and app_dir_name != PATH_TO_STATIC_RESOURCES_DIR:
-            build_apk_from_directory_contents(app_dir_name, files_list, build_type)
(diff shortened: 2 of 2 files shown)


scripts/consumer_apps_builder.py
```diff
@@ -100,7 +106,11 @@ def assemble_apk(domain, build_number, username, password, build_type):
 
 def get_app_name_from_profile():
     tree = ET.parse(PATH_TO_ASSETS_DIR_FROM_ODK + '/direct_install/profile.ccpr')
-    return tree.getroot().get("name")
+    return escape_apostrophes(tree.getroot().get("name").encode('utf-8'))
+
+
+def escape_apostrophes(s):
+    return s.replace("'", "\\'")
```
amstone326: required because down the line, application_name ends up in the `values.xml` file, where any quotes need to be escaped

build.gradle
```diff
@@ -516,14 +516,20 @@ def numbersToLetters(String version) {
...
     } else if ("true".equals(isConsumerApp)) {
         throw new InvalidUserDataException("An application_name property must be provided for consumer app standalone builds")
     } else {
         return ccDomainSafe;
     }
 }
 
+
+def cleanseAppNameForPackageIdentifier(appName) {
+    return appName.replaceAll("[ '.:]", "").replaceAll("\\\\", "").toLowerCase()
```
amstone326: none of these characters are allowed in package names, so remove them in case they exist in the original app name

wpride: Why the two replaceAll calls? (As opposed to one regex to match all the invalid chars)

scripts/consumer_apps_builder.py
wpride: But confirming " are ok?

wpride: (also maybe that comment belongs in the file?)

scripts/consumer_apps_builder.py
```diff
@@ -38,10 +38,16 @@ def checkout_or_update_static_resources_repo():
     os.chdir('../')
 
 
-def build_apks_from_resources(build_type):
+def build_apks_from_resources(build_type, subset_of_apps_to_build):
     for (app_dir_name, sub_dir_list, files_list) in os.walk(PATH_TO_STATIC_RESOURCES_DIR):
         if '.git' not in app_dir_name and app_dir_name != PATH_TO_STATIC_RESOURCES_DIR:
-            build_apk_from_directory_contents(app_dir_name, files_list, build_type)
+            if subset_of_apps_to_build is None or in_list(subset_of_apps_to_build, app_dir_name):
```
wpride: I don't see where subset_of_apps_to_build is ever used besides to check if its present/valid or not - not really clear what its supposed to be doing/what values it can take?

scripts/consumer_apps_builder.py
amstone326: No, quotes would have the same issue, as would a bunch of other punctuation. I decided not to go overboard with adding every possible case here, since apostrophes are the only one that showed up in an application name so far (and seem like the only one likely to).

Adding the comment to the file, though

build.gradle
amstone326: No great reason, just because I found it confusing that I needed 4 '\' characters to actually represent a regex for '\', so I left it separate because I wasn't sure how they'd interact

scripts/consumer_apps_builder.py
amstone326: Sorry I should have explained this, I'll add a comment in the file. This is just a debugging feature for me. This script is set up to pull down the consumer-apps-resources repository and build all of the apps that are in there. But I wanted to have the option of only building some of them while testing, just to make things faster when I was working on a bug in only 1 or 2 of the apps. If I were using this I would just pass in a comma-separated list of directory names as a command line argument to the script invocation:

`python consumer_apps_builder.py d app_Savings,app_WomensHealth`

PR #1320, opened 2016-06-14
Use built-in String.split() in Android code

amstone326:
As part of performance improvements for ICDS Daily Feeding Form: Use static injection to use the  built-in String.split() method in Android, while retaining our custom implementation for j2me. Also add tests to assert equivalence between the 2 implementations.

A subset of these changes will be included in the 2.27 LTS release as well, but I haven't set that up yet (waiting to talk with @ctsims on the exact right way to do that)

cross-request: https://github.com/dimagi/javarosa/pull/304


amstone326:
Release Note: Improve form load time for some forms

PR #1326, opened 2016-06-21
Uninstall and login ccc commands

phillipm:
Added uninstall and login commands to the `ccc` script.

Usage:
 - `ccc uninstall d75d71b21495e81de4bfd659be6edb60`. The `d75d..` is the app ID (from HQ url) of the installed app to be uninstalled.
 - `ccc login test 123` logs in using test/123 for the currently seated app. Note that this enables auto-login/password saving debug feature flag.


app/src/org/commcare/provider/DebugControlsReceiver.java
```diff
@@ -0,0 +1,53 @@
...
+import org.commcare.CommCareApplication;
+import org.commcare.activities.LoginActivity;
+import org.commcare.android.database.global.models.ApplicationRecord;
+import org.commcare.preferences.DevSessionRestorer;
+
+/**
+ * Process broadcasts requesting to
+ * - uninstall app
+ * - save the current commcare user session.
+ *
```
amstone326: can you update the javadoc to include logging in

app/src/org/commcare/provider/DebugControlsReceiver.java
```diff
@@ -0,0 +1,54 @@
...
+    private static void captureSession() {
+        DevSessionRestorer.saveSessionToPrefs();
+        DevSessionRestorer.enableAutoLogin();
+    }
+
+    private static void uninstallApp(String appId) {
+        ApplicationRecord appRecord = CommCareApplication._().getAppById(appId);
+        if (appRecord != null) {
+            CommCareApplication._().expireUserSession();
+            CommCareApplication._().uninstall(appRecord);
```
amstone326: I understand why you can't call `restartCommCare()` here like the actual uninstall method does (https://github.com/dimagi/commcare-odk/blob/master/app/src/org/commcare/activities/SingleAppManagerActivity.java#L175) since you don't have an origin activity, but just a warning that not doing so could produce slightly odd behavior

PR #1330, opened 2016-06-22
2.28 consumer apps tweaks

amstone326:
Final changes for consumer apps:
-Show a progress dialog during the update check portion of an update, so that the screen isn't just blank
-Fix an activity navigation bug that was occurring after updating to a new .ccz 
-At least for v1 release, don't do anything with the new restore file on an upgrade
-Don't show the logged in notification for a consumer apps session, and never time out the session

### DIFF
--- a/app/src/org/commcare/activities/CommCareHomeActivity.java
+++ b/app/src/org/commcare/activities/CommCareHomeActivity.java
@@ -172,10 +172,6 @@ public class CommCareHomeActivity
         processFromExternalLaunch(savedInstanceState);
         processFromShortcutLaunch();
         processFromLoginLaunch();
-
-        if (CommCareApplication._().isConsumerApp()) {
-            ConsumerAppsUtil.checkForChangedLocalRestoreFile(this);
-        }
     }
 
     private void loadInstanceState(Bundle savedInstanceState) {
--- a/app/src/org/commcare/activities/CommCareHomeActivity.java
+++ b/app/src/org/commcare/activities/CommCareHomeActivity.java
@@ -1120,7 +1120,6 @@ public class CommCareHomeActivity
             handlePendingSync();
         } else if (CommCareApplication._().isConsumerApp() && !sessionNavigationProceedingAfterOnResume) {
             // so that the user never sees the real home screen in a consumer app
-            sessionNavigationProceedingAfterOnResume = false;
             enterRootModule();
         } else {
             // Display the normal home screen!
--- a/app/src/org/commcare/activities/CommCareHomeActivity.java
+++ b/app/src/org/commcare/activities/CommCareHomeActivity.java
@@ -374,9 +374,9 @@ public class CommCareHomeActivity
     void enterRootModule() {
         Intent i;
         if (useGridMenu(org.commcare.suite.model.Menu.ROOT_MENU_ID)) {
-            i = new Intent(getApplicationContext(), MenuGrid.class);
+            i = new Intent(this, MenuGrid.class);
         } else {
-            i = new Intent(getApplicationContext(), MenuList.class);
+            i = new Intent(this, MenuList.class);
         }
         addPendingDataExtra(i, CommCareApplication._().getCurrentSessionWrapper().getSession());
         startActivityForResult(i, GET_COMMAND);
@@ -863,9 +863,9 @@ public class CommCareHomeActivity
         String command = asw.getSession().getCommand();
 
         if (useGridMenu(command)) {
-            i = new Intent(getApplicationContext(), MenuGrid.class);
+            i = new Intent(this, MenuGrid.class);
         } else {
-            i = new Intent(getApplicationContext(), MenuList.class);
+            i = new Intent(this, MenuList.class);
         }
         i.putExtra(SessionFrame.STATE_COMMAND_ID, command);
         addPendingDataExtra(i, asw.getSession());
@@ -1121,7 +1121,7 @@ public class CommCareHomeActivity
             handlePendingSync();
         } else if (CommCareApplication._().isConsumerApp() && !sessionNavigationProceedingAfterOnResume) {
             // so that the user never sees the real home screen in a consumer app
-            enterRootModule();
+            //enterRootModule();
         } else {
             // Display the normal home screen!
             uiController.refreshView();
--- a/app/src/org/commcare/activities/CommCareHomeActivity.java
+++ b/app/src/org/commcare/activities/CommCareHomeActivity.java
@@ -1099,7 +1099,6 @@ public class CommCareHomeActivity
             refreshActionBar();
         }
         attemptDispatchHomeScreen();
-
         sessionNavigationProceedingAfterOnResume = false;
     }
 
@@ -1121,7 +1120,8 @@ public class CommCareHomeActivity
             handlePendingSync();
         } else if (CommCareApplication._().isConsumerApp() && !sessionNavigationProceedingAfterOnResume) {
             // so that the user never sees the real home screen in a consumer app
-            //enterRootModule();
+            sessionNavigationProceedingAfterOnResume = false;
+            enterRootModule();
         } else {
             // Display the normal home screen!
             uiController.refreshView();
--- a/app/src/org/commcare/activities/DispatchActivity.java
+++ b/app/src/org/commcare/activities/DispatchActivity.java
@@ -35,6 +35,7 @@ public class DispatchActivity extends FragmentActivity {
     private static final int LOGIN_USER = 0;
     private static final int HOME_SCREEN = 1;
     public static final int INIT_APP = 2;
+    public static final int UPDATE_CONSUMER_APP = 3;
 
     /**
      * Request code for automatically validating media.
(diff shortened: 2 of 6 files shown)


app/src/org/commcare/activities/CommCareHomeActivity.java
```diff
@@ -172,10 +172,6 @@ protected void onCreateSessionSafe(Bundle savedInstanceState) {
         processFromExternalLaunch(savedInstanceState);
         processFromShortcutLaunch();
         processFromLoginLaunch();
-
-        if (CommCareApplication._().isConsumerApp()) {
```
amstone326: Decided with Clayton that we're not going to do anything with the new restore file on upgrades

app/src/org/commcare/activities/CommCareHomeActivity.java
```diff
@@ -1125,6 +1120,7 @@ private void attemptDispatchHomeScreen() {
             handlePendingSync();
         } else if (CommCareApplication._().isConsumerApp() && !sessionNavigationProceedingAfterOnResume) {
             // so that the user never sees the real home screen in a consumer app
+            sessionNavigationProceedingAfterOnResume = false;
```
amstone326: the fact that I was missing this statement is what was causing the weird activity navigation bugs, although to be honest I am not sure how it manifested the way it did.

app/src/org/commcare/activities/DispatchActivity.java
```diff
@@ -124,23 +125,24 @@ private void dispatch() {
...
                 // Recurse in order to make the correct decision based on the new state
                 dispatch();
+                return;
             } else {
                 Intent i = new Intent(getApplicationContext(), CommCareSetupActivity.class);
                 this.startActivityForResult(i, INIT_APP);
             }
         } else {
             // Note that the order in which these conditions are checked matters!!
+            if (CommCareApplication._().isConsumerApp() && !alreadyCheckedForAppFilesChange) {
```
amstone326: moving this check to a more proper place

app/src/org/commcare/activities/UpdateActivity.java
```diff
@@ -183,27 +184,26 @@ public void handleTaskUpdate(Integer... vals) {
...
     public void handleTaskCompletion(AppInstallStatus result) {
+        if (CommCareApplication._().isConsumerApp()) {
+            dismissProgressDialog();
+        }
+
         if (result == AppInstallStatus.UpdateStaged) {
             uiController.unappliedUpdateAvailableUiState();
             if (proceedAutomatically) {
                 launchUpdateInstallTask();
-                return;
```
amstone326: realized that these 3 return statements were unnecessary, and removing them allowed me not to duplicate the call to `unregisterTask()` (which also occurs immediately below)

app/src/org/commcare/activities/UpdateActivity.java
```diff
@@ -247,6 +247,14 @@ protected void startUpdateCheck() {
         uiController.downloadingUiState();
     }
 
+    private void initUpdateTaskProgressDisplay() {
```
amstone326: In normal updates, the update check portion does not use a progress dialog, it just shows a progress bar embedded in the main activity view. But since I am showing a blank screen for consumer app updates, I want the generic consumer app progress dialog to show, so I need to control it manually.

app/src/org/commcare/activities/UpdateActivity.java
```diff
@@ -183,27 +184,26 @@ public void handleTaskUpdate(Integer... vals) {
 
     @Override
     public void handleTaskCompletion(AppInstallStatus result) {
+        if (CommCareApplication._().isConsumerApp()) {
```
amstone326: see comment for `initUpdateTaskProgressDisplay()` below

app/src/org/commcare/activities/CommCareHomeActivity.java
```diff
@@ -378,9 +374,9 @@ private void goToFormArchive(boolean incomplete, FormRecord record) {
     void enterRootModule() {
         Intent i;
         if (useGridMenu(org.commcare.suite.model.Menu.ROOT_MENU_ID)) {
-            i = new Intent(getApplicationContext(), MenuGrid.class);
+            i = new Intent(this, MenuGrid.class);
```
amstone326: made these changes when I was debugging the activity navigation issue, but I think this is more correct anyway? Can change back if that's not true

app/src/org/commcare/activities/DispatchActivity.java
```diff
@@ -124,23 +125,24 @@ private void dispatch() {
...
+                checkForChangedCCZ();
+                return;
```
phillipm: `return` is not needed here and creates a lint warning

app/src/org/commcare/activities/DispatchActivity.java
```diff
@@ -114,8 +115,8 @@ protected void onSaveInstanceState(Bundle outState) {
 
     private void checkForChangedCCZ() {
         alreadyCheckedForAppFilesChange = true;
-        Intent i = new Intent(getApplicationContext(), UpdateActivity.class);
-        startActivity(i);
+        Intent i = new Intent(this, UpdateActivity.class);
+        startActivityForResult(i, UPDATE_CONSUMER_APP);
```
phillipm: I don't think you need the change `startActivity` -> `startActivityForResult`

app/src/org/commcare/activities/UpdateActivity.java
```diff
@@ -315,13 +323,12 @@ protected void deliverError(UpdateActivity receiver,
...
+            return ConsumerAppsUtil.getGenericConsumerAppsProgressDialog(taskId, false);
+        } else if (taskId != DIALOG_UPGRADE_INSTALL) {
             Log.w(TAG, "taskId passed to generateProgressDialog does not match "
                     + "any valid possibilities in CommCareSetupActivity");
             return null;
-        }
-        if (CommCareApplication._().isConsumerApp()) {
-            return ConsumerAppsUtil.getGenericConsumerAppsProgressDialog(taskId, false);
         } else {
             return generateNormalUpdateDialog(taskId);
```
phillipm: Would be nice if this said `UpdateInstall` or `Install` instead of `Update`, because this dialog is only shown when installing an update, not downloading it.

app/src/org/commcare/activities/UpdateActivity.java
phillipm: Can you add a shortened version of this description as a comment?

app/src/org/commcare/activities/UpdateActivity.java
```diff
@@ -247,6 +247,14 @@ protected void startUpdateCheck() {
         uiController.downloadingUiState();
     }
 
+    private void initUpdateTaskProgressDisplay() {
+        if (CommCareApplication._().isConsumerApp()) {
+            showProgressDialog(DIALOG_CONSUMER_APP_UPGRADE);
```
phillipm: This seems a little sketch because you are trying to attach a dialog to a task that doesn't extend `CommCareTask`. Hence all the fragment management based off of `CommCareTask.taskId` _might_ not work.

phillipm: For instance, have you tried rotating the device while this dialog is showing?

app/src/org/commcare/services/CommCareSessionService.java
```diff
@@ -301,6 +303,11 @@ public void run() {
      * progess then don't do anything.
      */
     private void timeToExpireSession() {
+        if (CommCareApplication._().isConsumerApp()) {
+            // Never expire the session if this is a consumer app
+            return;
+        }
```
phillipm: Instead of doing this can you just not setup/run `maintenanceTimer` if it is a consumer app?

app/src/org/commcare/activities/DispatchActivity.java
amstone326: Ah yea you're right, this was one of the things I tried to resolve the bug, but I'll get rid of it

app/src/org/commcare/services/CommCareSessionService.java
amstone326: good call

app/src/org/commcare/activities/UpdateActivity.java
amstone326: Hm, so most of the actual dialog management, at least in terms of activity lifecycle stuff, is done by `CommCareActivity`, which `UpdateActivity` does extend. The only thing that is controlled by a `CommCareTask` is the automatic showing and dismissing of the dialog based upon the life cycle of the task, so I'm doing that part manually at the points where I know the task is starting and stopping. Agreed that it's not an amazing way to do it, but it does seem to be pretty robust (I did test it with rotating).

amstone326:
@phillipm Implemented feedback, but the bug I was struggling with has re-manifested itself :( So don't pull for now, will look into it again when I'm less sad...

amstone326:
@phillipm This is fixed now :) The issue was that we were not unregistering a pending session expiration when we started a new session, so that if the user got logged in again before we had a chance to respond to the session expiration directive, we would redirect to login when it was incorrect to do so.

PR #1338, opened 2016-06-28
Remove J2ME-compat 'split'

phillipm:
cross-request: https://github.com/dimagi/commcare-core/pull/320


unit-tests/src/org/commcare/utils/StaticInjectionTests.java
```diff
@@ -1,125 +0,0 @@
...
-/**
- * Test that the output of the j2me- and Android- implementations of methods that use static
- * injection for the Android version have identical behavior
- *
- * Created by amstone326 on 6/14/16.
- */
-@Config(application = CommCareApplication.class,
-        constants = BuildConfig.class)
-@RunWith(CommCareTestRunner.class)
-public class StaticInjectionTests {
```
phillipm: Sad to remove these tests, but I feel like we probably don't need them anymore

PR #1339, opened 2016-06-28
Refactor: width calculations for case list columns

phillipm:
Break-up case list column width calculations into helper methods and moves them to `ViewUtil`.

Last commit is the fix in the 2.28 PR: https://github.com/dimagi/commcare-android/pull/1340


app/src/org/commcare/views/ViewUtil.java
```diff
@@ -123,4 +85,99 @@ public static int getColorDrawableColor(ColorDrawable drawable) {
...
+            if (width != -1) {
+                claimedSpace += width;
+            } else {
+                indeterminateColumns++;
+            }
+        }
+        return new Pair<>(claimedSpace, indeterminateColumns);
+    }
+
+    private static boolean claimedSpaceNotTight(int fullSize, int claimedSpace,
```
amstone326: The method name here confused me until I thought about it a bit. What about just simplifying to 'widthReadjustmentNeeded', since the comment you have there explains the specifics anyway?

app/src/org/commcare/views/ViewUtil.java
```diff
@@ -123,4 +85,99 @@ public static int getColorDrawableColor(ColorDrawable drawable) {
...
+        // as percentage of the entire screen width.
+        int[] widths = new int[hints.size()];
+        parseWidths(hints, widths, fullSize);
+
+        Pair<Integer, Integer> constraints = buildConstraints(widths);
+        int claimedSpace = constraints.first;
+        int indeterminateColumns = constraints.second;
+
+        if (claimedSpaceNotTight(fullSize, claimedSpace, indeterminateColumns)) {
+            // Either more space has been claimed than the screen has room for,
```
amstone326: would suggest making this comment a javadoc for the `claimedSpaceNotTight` method

app/src/org/commcare/views/ViewUtil.java
phillipm: Yeah, had trouble coming up with a good name. Good suggestion, updated

PR #1340, opened 2016-06-28
Remove padding from case list columns that have small widths

phillipm:
Case list columns have the ability to set width percentages. On small devices, it might turn out that the width calculated for a column is smaller or super close to the padding size of that column. 

This PR removes the column's horizontal padding when its size is too close to the width allotted for the column.

Before:
screenshot_20160628-170653

After:
screenshot_20160628-170528

Ticket: http://manage.dimagi.com/default.asp?230152


amstone326:
Can you include the ticket this came from?

phillipm:
@amstone326 oops, here you go: http://manage.dimagi.com/default.asp?230152

PR #1346, opened 2016-06-30
add space after session key

wpride:
Fix for bug that was breaking all incomplete forms that used session data

### DIFF
--- a/app/src/org/commcare/android/database/user/models/SessionStateDescriptor.java
+++ b/app/src/org/commcare/android/database/user/models/SessionStateDescriptor.java
@@ -91,7 +91,7 @@ public class SessionStateDescriptor extends Persisted implements EncryptedModel 
         //TODO: Serialize into something more useful. I dunno. JSON/XML/Something
         StringBuilder descriptor = new StringBuilder();
         for (StackFrameStep step : session.getFrame().getSteps()) {
-            descriptor.append(step.getType());
+            descriptor.append(step.getType()).append(" ");
             if (SessionFrame.STATE_COMMAND_ID.equals(step.getType())) {
                 descriptor.append(step.getId());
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/android/database/user/models/SessionStateDescriptor.java
```diff
@@ -91,7 +91,7 @@ private String createSessionDescriptor(CommCareSession session) {
         //TODO: Serialize into something more useful. I dunno. JSON/XML/Something
         StringBuilder descriptor = new StringBuilder();
         for (StackFrameStep step : session.getFrame().getSteps()) {
-            descriptor.append(step.getType());
+            descriptor.append(step.getType()).append(" ");
             if (SessionFrame.STATE_COMMAND_ID.equals(step.getType())) {
                 descriptor.append(step.getId());
```
phillipm: I think we need a space here too

phillipm: I'm actually a little confused because I moved the space addition to the bottom of the loop body in https://github.com/dimagi/commcare-android/pull/1059/files#diff-cd3d036d05167f3e910a1381ea329b91R106

phillipm: Oh, nevermind, disregard this whole comment chain!

wpride: I think the one at the end handles for all the afterwards ones, but this is in between the key and the value

phillipm:
Can you recreate this PR to target `commcare_2.28`?
I've included the fix in https://github.com/dimagi/commcare-android/pull/1347 which has tests and other fixes; and targets `master`

PR #1348, opened 2016-06-30
Async restores

amstone326:
Context/Feature Description: 

-Current state, prior to this feature: OTA restore can be very slow for users with a large number of cases. If an incremental sync fails due to timeout of network, the work done during the request is lost and the next request must start again from the beginning. The user also does not see any indication of why the sync request is taking so long or how much longer it may take.

-New state with this feature: HQ will generate the restore response asynchronously, and send a retry response if the process is taking too long, which includes an estimation of the progress completed so far, and a time period after which to try again. When mobile retries, HQ will send the completed restore response if available, or another retry response if not. On the mobile side, this allows us to show a dialog to the user if a retry response is received, displaying the progress that has been made so far while the server continues to work on generating the restore response, thus creating a much better user experience during restores that take a long time.

-----

Screenshots / Other minor UI changes: As part of the UI work for this feature, I decided to try to create clearer distinctions between the 3 potential "phases" of a sync by giving each a distinct title (previously we used the "Communicating with Server" title for all phases). Clayton is also working on coming up with icons that we might place somewhere on the dialog for each phase (the thinking being that when users report bugs with syncing, we want them to be able to tell us what part of the sync it occurred in, and they are more likely to recall what icon they saw than anything else)

Phase 1 - contacting server (I removed the progress bar from this dialog because it always remains empty):
screenshot_20160713-174535

Phase 2 (NEW PHASE, will not always occur) - waiting for server progress:
screenshot_20160713-174622

Phase 3 - processing data from server:
screenshot_20160713-174653
 
-----

I still need to write an integration test for this feature, but that can't be done until HQ has created a way for mobile to request getting a specific response back to a sync request (Farid is going to work on this).


app/src/org/commcare/activities/SyncUIHandling.java
```diff
@@ -63,23 +63,35 @@ public static void handleSyncResult(ConnectorWithResultCallback activity,
...
                                         Integer... update) {
-        if (update[0] == DataPullTask.PROGRESS_STARTED) {
+        int progressCode = update[0];
+        if (progressCode == DataPullTask.PROGRESS_STARTED) {
             activity.updateProgress(Localization.get("sync.progress.purge"), DataPullTask.DATA_PULL_TASK_ID);
-        } else if (update[0] == DataPullTask.PROGRESS_CLEANED) {
+        } else if (progressCode == DataPullTask.PROGRESS_CLEANED) {
             activity.updateProgress(Localization.get("sync.progress.authing"), DataPullTask.DATA_PULL_TASK_ID);
-        } else if (update[0] == DataPullTask.PROGRESS_AUTHED) {
+            activity.updateProgressBarVisibility(false);
```
amstone326: the progress bar was not actually utilized in this phase or in the PROGRESS_DOWNLOADING phase, so I think it makes sense to not show it, instead of just showing it as empty for a while

amstone326:
@proteusvacuum fyi :tada:

proteusvacuum:
yay!

app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -602,24 +602,7 @@ public void handlePullTaskUpdate(Integer... update) {
...
-            hideTaskCancelButton();
-        } else if (update[0] == DataPullTask.PROGRESS_PROCESSING) {
-            updateProgress(Localization.get("sync.process.processing", new String[]{String.valueOf(update[1]), String.valueOf(update[2])}), DataPullTask.DATA_PULL_TASK_ID);
-            updateProgressBar(update[1], update[2], DataPullTask.DATA_PULL_TASK_ID);
-        } else if (update[0] == DataPullTask.PROGRESS_RECOVERY_NEEDED) {
-            updateProgress(Localization.get("sync.recover.needed"), DataPullTask.DATA_PULL_TASK_ID);
-        } else if (update[0] == DataPullTask.PROGRESS_RECOVERY_STARTED) {
-            updateProgress(Localization.get("sync.recover.started"), DataPullTask.DATA_PULL_TASK_ID);
-        }
+        SyncUIHandling.handleSyncUpdate(this, update);
```
phillipm: 😹

app/src/org/commcare/network/RemoteDataPullResponse.java
```diff
@@ -128,4 +129,8 @@ protected long guessDataSize() {
         }
         return -1;
     }
+
+    public String getRetryHeaderValue() {
+        return response.getFirstHeader("Retry-After").getValue();
```
phillipm: Assuming we are fine with this failing hard if the header value isn't around

amstone326: Interesting question.. it feels like yes because it means HQ just totally messed up the response, but I could see an argument for doing something better..

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -245,6 +225,57 @@ private static boolean shouldGenerateFirstKey() {
         return keyServer == null || keyServer.equals("");
     }
 
+    private ResultAndError<PullTaskResult> getRequestResultOrRetry(AndroidTransactionParserFactory factory) {
+        if (retryLimit != -1 && numTries >= retryLimit) {
+            // for testing purposes only
```
phillipm: why?

amstone326: why what?

phillipm: sorry; I'm confused why this is for testing purposes only. It seems like a legit thing to be checking.

amstone326: Ah. It just wasn't part of the spec or something we discussed introducing at all. I guess in theory we could decide on a limit for number of times that we want to allow us to get retry responses from HQ, but I actually don't think we do, because it's hard to know what how many of those it might take, depending on how huge the restore is and how taxed the server is.

I think it may make more sense once you read through the tests as to why I added this for those only.

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -509,6 +569,32 @@ private void updateUserSyncToken(String syncToken) throws StorageFullException {
...
+                    if (parser.getName().toLowerCase().equals("progress")) {
+                        serverProgressCompletedSoFar = Integer.parseInt(parser.getAttributeValue(null, "done"));
+                        serverProgressTotal = Integer.parseInt(parser.getAttributeValue(null, "total"));
+                        return;
+                    }
+                }
+                eventType = parser.next();
+            } while (eventType != KXmlParser.END_DOCUMENT);
+        } catch (IOException | XmlPullParserException e) {
+            Logger.log(AndroidLogger.TYPE_USER, "Error while parsing progress values of retry result");
```
phillipm: Slightly curious what happens if we get here in terms of the restore process? Does the progress bar enter a specific state? Will the restore continue without updating the progress vars, or will that cause things to abort?

app/src/org/commcare/network/HttpRequestEndpointsMock.java
```diff
@@ -41,6 +41,11 @@ public HttpResponse makeCaseFetchRequest(String baseUri, boolean includeStateFla
     }
 
     @Override
+    public HttpResponse makeAsyncCaseFetchRequest() {
+        return HttpResponseMock.buildHttpResponseMockForAsyncRestore();
```
phillipm: Instead of this, I would check `LocalDataPullResponseFactory.INSTANCE.isAsyncRestore()` in `HttpRequestEndpointsMock.makeCaseFetchRequest`. 
It would be nice to avoid adding the `makeAsyncCaseFetchRequest` declaration to `HttpRequestEndpoints` just for testing purposes.

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -582,8 +668,39 @@ public void reportDownloadProgress(int totalRead) {
         publishProgress(DataPullTask.PROGRESS_DOWNLOADING, totalRead);
     }
 
+    /**
+     * In order to achieve the impression of the progress bar updating smoothly during the wait
+     * period after a retry response is received, use a timer task to report incremental progress
+     * from the state we were last in to the progress value that the server just reported back to
+     * us, splitting up the wait time into uniform time periods for reporting each unit of progress.
+     */
+    private void startReportingServerProgress() {
```
phillipm: You should be able to lift this into a helper class that has `serverProgressCompletedSoFar` and `lastReportedServerProgressValue` as fields and takes in an `AsyncTask` it can publish results to. You could then also move the retry response parsing code into it. I'm suggesting this idea mainly to reduce the size of `DataPullTask`, which is already large, and isolate the retry code so it is easier to reason about.

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -245,6 +225,57 @@ private static boolean shouldGenerateFirstKey() {
         return keyServer == null || keyServer.equals("");
     }
 
+    private ResultAndError<PullTaskResult> getRequestResultOrRetry(AndroidTransactionParserFactory factory) {
+        if (retryLimit != -1 && numTries >= retryLimit) {
+            // for testing purposes only
+            return new ResultAndError<>(PullTaskResult.RETRY_LIMIT_EXCEEDED);
+        }
```
phillipm: Instead of adding a custom `RETRY_LIMIT_EXCEEDED` return code for tests, can you try counting retry limit and number of tries in `HttpRequestEndpointsMock` and returning something like `DataPullTask.CONNECTION_TIMEOUT` in the mock and asserting it in the async test?

In the past I've added test code to the prod code base, but I think it is good to avoid whenever possible.

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -582,8 +668,39 @@ public void reportDownloadProgress(int totalRead) {
...
+                    if (lastReportedServerProgressValue == serverProgressCompletedSoFar) {
+                        reportServerProgressTimer.cancel();
+                        reportServerProgressTimer.purge();
+                        return;
+                    }
+                    DataPullTask.this.publishProgress(PROGRESS_SERVER_PROCESSING,
+                            ++lastReportedServerProgressValue, serverProgressTotal);
+                }
+            }, 0, intervalAllottedPerProgressUnit);
+    }
```
phillipm: Saw this in action, was pretty :metal:

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -582,8 +668,39 @@ public void reportDownloadProgress(int totalRead) {
...
+        if (amountOfProgressToCoverThisCycle == 0) {
+            return;
+        }
+        long intervalAllottedPerProgressUnit = millisUntilNextAttempt / amountOfProgressToCoverThisCycle;
+
+        final Timer reportServerProgressTimer = new Timer();
+        reportServerProgressTimer.schedule(new TimerTask() {
+                @Override
+                public void run() {
+                    if (lastReportedServerProgressValue == serverProgressCompletedSoFar) {
```
phillipm: Not sure how important it is to synchronize around modifying and reading `serverProgressCompletedSoFar` and `serverProgressTotal`

app/src/org/commcare/tasks/DataPullTask.java
amstone326: Yea good call, I think I just missed this. I'm going to make both this case and the case you mentioned above (where the "Retry-After" header doesn't exist) return the `PullTaskResult.BAD_DATA` field, like I'm already doing when the retry header value is invalid.

app/src/org/commcare/network/HttpRequestEndpointsMock.java
amstone326: Good call!

app/src/org/commcare/tasks/DataPullTask.java
amstone326: done!

app/src/org/commcare/network/HttpRequestEndpointsMock.java
```diff
@@ -31,13 +31,21 @@ public static void setCaseFetchResponseCodes(Integer[] responseCodes) {
 
     @Override
     public HttpResponse makeCaseFetchRequest(String baseUri, boolean includeStateFlags) throws ClientProtocolException, IOException {
-        int responseCode;
-        if (caseFetchResponseCodeStack.size() > 0) {
-            responseCode = caseFetchResponseCodeStack.remove(0);
+        if (LocalDataPullResponseFactory.INSTANCE.isAsyncRestore) {
+            return makeAsyncCaseFetchRequest();
```
phillipm: Since `makeAsyncCaseFetchRequest` is only called here and is one line, maybe you could just inline its body here?

app/src/org/commcare/tasks/AsyncRestoreHelper.java
```diff
@@ -0,0 +1,132 @@
...
+import java.io.InputStream;
+import java.util.Timer;
+import java.util.TimerTask;
+
+/**
+ * Keeps tracking of everything DataPullTask needs to know and do in the case that the server is
+ * performing an async restore, including parsing the retry response and reporting incremental
+ * server progress during the retry wait period
+ */
+public class AsyncRestoreHelper {
```
phillipm: 👍 👍

amstone326:
@phillipm Thanks for the helpful review, all of your comments should be addressed now!

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -578,12 +599,22 @@ public void onSuccess() {
...
+    }
+
+    public AsyncRestoreHelper getAsyncRestoreHelper() {
+        return this.asyncRestoreHelper;
     }
 
     public enum PullTaskResult {
         DOWNLOAD_SUCCESS(-1),
+        RETRY_NEEDED(-1),
+        RETRY_LIMIT_EXCEEDED(-1),
```
phillipm: No longer used

proteusvacuum:
:balloon: :tada:

PR #1352, opened 2016-06-30
Finish and return activity after redirect to login

amstone326:


app/src/org/commcare/activities/CommCareHomeActivity.java
```diff
@@ -1098,6 +1098,10 @@ private boolean checkAndStartUnsentFormsTask(final boolean syncAfterwards,
     @Override
     protected void onResume() {
         super.onResume();
+        if (redirectedToLogin) {
+            finish();
+            return;
```
amstone326: @phillipm Need your opinion here- I realized that what we discussed earlier about having the `SessionAwareCommCareActivity` call `finish()` isn't sufficient, because the child activity will still proceed with whatever code path is in its `onResume()` method, even after `finish()` has been called. In order to truly halt all further execution of code in the activity, you have to call `finish()` AND `return`. But obviously, `return` has to be called from within the child activity's own `onResume` method... so I think in order to make this work, every single class that implements `SessionAwareCommCareActivity` would have to include the above lines of code in its `onResume` method.

So, 2 questions: Can you think of a better way of doing this that I'm not thinking of? And if not, do you think the mess of this is worth doing?

phillipm: Hmm, I was hoping that `finish` would still interrupt the code-path at some point. I guess it can only do so after `onResume` releases control.

I don't have any better ideas. I forget, what was the importance of stopping activity setup code from continuing to run when the session expires?

amstone326: @phillipm So the point of setting up the whole session aware infrastructure was to avoid causing a `SessionUnvailableException` if the user returns to code that assumes there is an active session, after having been logged out from the app manager. Calling the `redirectToLogin` function doesn't really resolve this problem if the code path is still allowed to continue running, because it might execute a statement that relies on an active session. In fact now that I think about it, this was probably contributing to the confusing crashes we were encountering after uninstalling an app from the manager and then coming back to CommCare. We were both thinking that calling `redirectToLogin` alone was sufficient, but I don't think it is.

amstone326: @phillipm, when you have a chance, let me know what your thoughts are based on that information. My inclination is that this is the correct thing to do, even if the code for it will be a bit messy. But would be good to know if you feel otherwise before I go ahead and do it.

phillipm: We could also reuse this pattern for `onResume` as well: https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/activities/SessionAwareCommCareActivity.java#L18-L22

phillipm: Or rather, make `SessionAwareCommCareActivity.onResume` only call `onResumeSessionSafe` if the session hasn't expired, otherwise redirecting to login.

amstone326:
@phillipm I think your last suggestion is the best option, so went ahead and implemented that. Let me know if this doesn't look like what you were envisioning, but I think I'm happy with this solution.

PR #1360, opened 2016-07-06
Fix bugs related to showing message after successful form dump or send

amstone326:
Fix for http://manage.dimagi.com/default.asp?231306 and http://manage.dimagi.com/default.asp?231600, which were the same bug. Comments in-line as to what fixed the issue

### DIFF
--- a/app/src/org/commcare/activities/AdvancedActionsActivity.java
+++ b/app/src/org/commcare/activities/AdvancedActionsActivity.java
@@ -45,11 +45,12 @@ public class AdvancedActionsActivity extends SessionAwarePreferenceActivity {
     private final static int WIFI_DIRECT_ACTIVITY = 1;
     private final static int DUMP_FORMS_ACTIVITY = 2;
 
-    public final static int RESULT_DATA_RESET = RESULT_FIRST_USER + 1;
-    public final static int RESULT_FORMS_PROCESSED = RESULT_FIRST_USER + 2;
+    public final static int RESULT_DATA_RESET = CommCareHomeActivity.RESULT_RESTART + 1;
+    public final static int RESULT_FORMS_PROCESSED = CommCareHomeActivity.RESULT_RESTART + 2;
 
     public final static String FORM_PROCESS_COUNT_KEY = "forms-processed-count";
     public final static String FORM_PROCESS_MESSAGE_KEY = "forms-processed-message";
+    static final String KEY_NUMBER_DUMPED = "num_dumped";
 
     private final static Map<String, String> keyToTitleMap = new HashMap<>();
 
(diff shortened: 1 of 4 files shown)


app/src/org/commcare/activities/CommCareHomeActivity.java
```diff
@@ -1271,7 +1271,7 @@ public static void createPreferencesMenu(Activity activity) {
     }
 
     private void startAdvancedActionsActivity() {
-        startActivity(new Intent(this, AdvancedActionsActivity.class));
+        startActivityForResult(new Intent(this, AdvancedActionsActivity.class), ADVANCED_ACTIONS_ACTIVITY);
```
amstone326: the primary issue was that `AdvancedActionsActivity` was not being started for result, so the `handleAdvancedActionResult` callback was never being called.

app/src/org/commcare/activities/AdvancedActionsActivity.java
```diff
@@ -45,11 +45,12 @@
     private final static int WIFI_DIRECT_ACTIVITY = 1;
     private final static int DUMP_FORMS_ACTIVITY = 2;
 
-    public final static int RESULT_DATA_RESET = RESULT_FIRST_USER + 1;
-    public final static int RESULT_FORMS_PROCESSED = RESULT_FIRST_USER + 2;
+    public final static int RESULT_DATA_RESET = CommCareHomeActivity.RESULT_RESTART + 1;
+    public final static int RESULT_FORMS_PROCESSED = CommCareHomeActivity.RESULT_RESTART + 2;
```
amstone326: Other issue was that these result codes conflicted with `CommCareHomeActivity.RESULT_RESTART` resulting in incorrect processing upon return from `AdvancedActionsActivity`

app/src/org/commcare/activities/AdvancedActionsActivity.java
```diff
@@ -246,30 +247,30 @@ public void onClick(DialogInterface dialog,
...
-                    Intent returnIntent = new Intent();
-                    returnIntent.putExtra(FORM_PROCESS_COUNT_KEY, dumpedCount);
-                    returnIntent.putExtra(FORM_PROCESS_MESSAGE_KEY, "bulk.form.dump.success");
-                    setResult(RESULT_FORMS_PROCESSED, returnIntent);
-                    finish();
-                }
-                break;
-            default:
-                break;
+        if (requestCode == WIFI_DIRECT_ACTIVITY || requestCode == DUMP_FORMS_ACTIVITY) {
```
amstone326: unified these into the same codepath since almost all of the code was the same.

app/src/org/commcare/activities/AdvancedActionsActivity.java
```diff
@@ -45,11 +45,12 @@
...
     private final static int DUMP_FORMS_ACTIVITY = 2;
 
-    public final static int RESULT_DATA_RESET = RESULT_FIRST_USER + 1;
-    public final static int RESULT_FORMS_PROCESSED = RESULT_FIRST_USER + 2;
+    public final static int RESULT_DATA_RESET = CommCareHomeActivity.RESULT_RESTART + 1;
+    public final static int RESULT_FORMS_PROCESSED = CommCareHomeActivity.RESULT_RESTART + 2;
 
     public final static String FORM_PROCESS_COUNT_KEY = "forms-processed-count";
     public final static String FORM_PROCESS_MESSAGE_KEY = "forms-processed-message";
+    static final String KEY_NUMBER_DUMPED = "num_dumped";
```
amstone326: was unnecessary and added complexity to the processing code to have `WifiDirectActivity` and `CommCareFormDumpActivity` use different keys for the number_dumped value, so unified them to both use this key.

PR #1347, opened 2016-06-30
SessionStateDescriptor tests and cleanup

phillipm:
Regression test for https://github.com/dimagi/commcare-android/pull/1346 and cleanup as to how we serialize sync/query steps in the `SessionStateDescriptor`. That is, completely skip including sync/query steps in the descriptor because it shouldn't be needed

PR #1361, opened 2016-07-06
Backmerge 2.28

amstone326:


phillipm:
I think the tests are failing due to the issue I mentioned in https://github.com/dimagi/commcare-android/pull/1347#issuecomment-229768553

amstone326:
@phillipm Should we just ignore it for now then?

phillipm:
If we merge it, then all subsequent PR tests will fail. I try fixing it this afternoon after I take care of some 2.28 bugs.

PR #1364, opened 2016-07-07
Add migration stragety to InstanceProvider.insert

phillipm:
2.28 version of https://github.com/dimagi/commcare-android/pull/1358 with minimal changes and no tests (since they require more changes, even though all those changes are to test files...)

Should be viewed with ?w=1

Addresses the following crash that happens when a user key record expires on HQ and HQ sends down a new one.

The underlying issue was that a while ago I moved a bunch of instance save logic into the `InstanceProvider` where it should be, but didn't realize that instances were migrated in a very specific way when key records expired. Hence, the instance migration code was handling all its form record upkeep, but the instance insert code was also trying to manage form records by linking it to the session's form record (which in this case doesn't exist).

```
Unexpected error while migrating database: android.database.SQLException: Failed to insert row into content://org.commcare.android.provider.odk.instances/instances at
org.commcare.provider.InstanceProvider.insert(InstanceProvider.java:222) at
android.content.ContentProvider$Transport.insert(ContentProvider.java:220) at
android.content.ContentResolver.insert(ContentResolver.java:1190) at
org.commcare.models.database.user.UserSandboxUtils.migrateData(UserSandboxUtils.java:147) at
org.commcare.tasks.ManageKeyRecordTask.lookForAndMigrateOldSandbox(ManageKeyRecordTask.java:556) at
org.commcare.tasks.ManageKeyRecordTask.doPostCalloutTask(ManageKeyRecordTask.java:445) at
org.commcare.network.HttpCalloutTask.doTaskBackground(HttpCalloutTask.java:107) at
org.commcare.network.HttpCalloutTask.doTaskBackground(HttpCalloutTask.java:29) at
org.commcare.tasks.templates.CommCareTask.doInBackground(CommCareTask.java:37) at
android.os.AsyncTask$2.call(AsyncTask.java:288) at
java.util.concurrent.FutureTask.run(FutureTask.java:237) at
android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:231) at
java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112) at
java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587) at
java.lang.Thread.run(Thread.java:841)
```

### DIFF
--- a/app/src/org/commcare/provider/InstanceProvider.java
+++ b/app/src/org/commcare/provider/InstanceProvider.java
@@ -195,13 +195,7 @@ public class InstanceProvider extends ContentProvider {
             values.put(InstanceProviderAPI.InstanceColumns.STATUS, InstanceProviderAPI.STATUS_INCOMPLETE);
         }
 
-        // Should we link this instance to the session's form record, or create
-        // a new, unindexed one?
-        boolean linkToSession = true;
-        if (values.containsKey(InstanceProviderAPI.UNINDEXED_SUBMISSION)) {
-            values.remove(InstanceProviderAPI.UNINDEXED_SUBMISSION);
-            linkToSession = false;
-        }
+        InstanceProviderInsertType insertType = InstanceProviderInsertType.getInsertionType(values);
 
         SQLiteDatabase db = mDbHelper.getWritableDatabase();
         long rowId = db.insert(INSTANCES_TABLE_NAME, null, values);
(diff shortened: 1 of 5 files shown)


app/src/org/commcare/tasks/ManageKeyRecordTask.java
```diff
@@ -410,6 +409,7 @@ private void markOldRecordsInactive(SqlStorage<UserKeyRecord> storage, String us
         for (UserKeyRecord r : allRecordsWithSameUsername) {
             if (!r.getUuid().equals(uuidOfActiveRecord)) {
                 r.setInactive();
+                storage.write(r);
```
phillipm: Bug fix that should be double checked. I'm not sure if this was causing things to break or not, but it sorta sketched me out. `r.setInactive` was being set, but the change wasn't being committed to storage. 

Original change was made in https://github.com/dimagi/commcare-android/pull/979

app/src/org/commcare/tasks/ManageKeyRecordTask.java
```diff
@@ -281,30 +281,29 @@ private void cleanupUserKeyRecords() {
                 newRecord.setType(UserKeyRecord.TYPE_NORMAL);
                 storage.write(newRecord);
             }
+        }
 
+        for (UserKeyRecord recordPendingDelete :
```
phillipm: This was confusing to me: the loop that removes user key records that are set as `TYPE_PENDING_DELETE` was inside another clean-up loop that processes new user key records. This means that it was never getting run. I test this thoroughly in https://github.com/dimagi/commcare-android/pull/1358/files#diff-e0416b2e041f9a9012311119486809ccR115

app/src/org/commcare/provider/InstanceProvider.java
```diff
@@ -211,28 +205,15 @@ public Uri insert(@NonNull Uri uri, ContentValues initialValues) {
...
-                storage.write(r);
+            switch (insertType) {
+                case SESSION_LINKED:
+                    finalizeSessionLinkedInsertion(instanceUri, uri);
+                    break;
+                case UNINDEXED_IMPORT:
+                    finalizeUnindexedInsertion(values, instanceUri);
+                    break;
+                case SANDBOX_MIGRATED:
+                    break;
```
phillipm: The fix is to add this migration branch that doesn't do anything to the form record.

PR #1368, opened 2016-07-08
Broadcast case id on case create/update events

phillipm:
This is needed so that Simprints knows

- when they should mark fingerprint registration as committed.
- when a case is selected, allowing them to refine their matching algorithm

cross-request: https://github.com/dimagi/commcare-core/pull/329

### DIFF
--- a/app/src/org/commcare/activities/EntityDetailActivity.java
+++ b/app/src/org/commcare/activities/EntityDetailActivity.java
@@ -235,7 +235,7 @@ public class EntityDetailActivity
     private void announceCaseSelect() {
         Intent selectIntentBroadcast = new Intent("org.commcare.dalvik.api.action.case.selected");
         selectIntentBroadcast.putExtra("case_id", getIntent().getStringExtra(SessionFrame.STATE_DATUM_VAL));
-        sendBroadcast(selectIntentBroadcast);
+        sendBroadcast(selectIntentBroadcast, "org.commcare.dalvik.provider.cases.read");
     }
 
     @Override
--- a/app/src/org/commcare/activities/EntityDetailActivity.java
+++ b/app/src/org/commcare/activities/EntityDetailActivity.java
@@ -224,6 +224,11 @@ public class EntityDetailActivity
      * Move along to form entry.
      */
     private void select() {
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/activities/EntityDetailActivity.java
```diff
@@ -224,6 +224,11 @@ protected boolean onBackwardSwipe() {
      * Move along to form entry.
      */
     private void select() {
+        // announce that case was selected
+        Intent selectIntentBroadcast = new Intent("org.commcare.dalvik.api.action.case.selected");
```
amstone326: could move this into a well-named helper and remove the need for an explanatory comment

app/src/org/commcare/models/FormRecordProcessor.java
```diff
@@ -85,7 +85,8 @@ public TransactionParser getParser(KXmlParser parser) {
 
         //Let anyone who is listening know!
         Intent i = new Intent("org.commcare.dalvik.api.action.data.update");
-        this.c.sendBroadcast(i);
+        i.putStringArrayListExtra("cases", factory.getCreatedAndUpdatedCases());
```
amstone326: is this definitely something that's ok to be broadcasting?

phillipm: As far as I know this seems fine. @ctsims any opinions?

amstone326:
@phillipm this is still failing after triggering a rebuild on jenkins, would suggest maybe making a noop change and pushing to re-trigger from here?

app/src/org/commcare/models/FormRecordProcessor.java
ctsims: Interesting... We do treat case Id's as private data generally. IS it clear whether this broadcast is permissioned? There's an Android "read CommCare data" permission that we require for the content provider. If it's set for the broadcasts as well this seems good to go.

phillipm: Ah, good call. I've added the case read permission and an example of it being read in https://github.com/dimagi/commcare-tester-app/pull/4

PR #1371, opened 2016-07-11
Increase form entry widget's button size

phillipm:
I've been finding it really really hard to press widget buttons during form entry. I think this is more of an issue on devices with high screen density. Increasing the padding helps. It does look just a tad too big on lower density devices, so open to suggestions.

| low density current UI | low density larger button UI |
|-------|-----------|
| small_button | big_button |

| high density current UI | high density larger button UI |
|-------|-----------|
| new_small_button | new_big_button |


ctsims:
You can parameterize the padding differently by adding it to the dimens
file. Can show you how this works if you want.

amstone326:
@phillipm Should I wait to merge until you talk to Clayton about an alternate strategy?

phillipm:
yeah

PR #1372, opened 2016-07-11
Fix IntentWidget 'quick' auto-advance + refactors

phillipm:
Form entry intent callouts have the ability to auto-launch using the ['quick' appearance attribute](https://confluence.dimagi.com/pages/viewpage.action?pageId=32637620). This feature was broken, in that it overwrote any data set by the intent callout result back to `null` when it auto-advanced to the next screen. This was happening because `IntentWidget.getAnswer` returns the value in the `TextView` for the widget, which isn't updated before advancing to the next screen forces a save.


app/src/org/commcare/views/widgets/IntentWidget.java
```diff
@@ -152,12 +156,7 @@ public void clearAnswer() {
 
     @Override
     public IAnswerData getAnswer() {
-        String s = mStringAnswer.getText().toString();
-        if (s.equals("")) {
-            return null;
-        } else {
-            return new StringData(s);
-        }
+        return mPrompt.getAnswerValue();
```
phillipm: The fix

app/src/org/commcare/views/widgets/WidgetFactory.java
```diff
@@ -40,64 +40,11 @@ public QuestionWidget createWidgetFromPrompt(FormEntryPrompt fep, Context contex
...
-                    String intentId = appearance.substring("intent:".length());
-                    IntentCallout ic = form.getExtension(AndroidXFormExtensions.class).getIntent(intentId, form);
-                    //Hm, so what do we do if no callout is found? Error? For now, fail fast
-                    if (ic == null) {
-                        throw new RuntimeException("No intent callout could be found for requested id " + intentId + "!");
-                    }
-                    //NOTE: No path specific stuff for now
-                    Intent i = ic.generate(form.getEvaluationContext());
-                    questionWidget = new IntentWidget(context, fep, i, ic, pendingCalloutInterface);
+                    questionWidget = buildIntentWidget(appearance, fep, context);
```
phillipm: Changes in this file just make it use helper methods

PR #1374, opened 2016-07-12
Extend ccc script to trigger 412 recovery syncs

phillipm:
`ccc invalidate` will set a fake value for the case DB hash. On the next sync this value will be used (and then removed), such that HQ will send down a 412, triggering a full recovery on the mobile device.

Will be used for integration tests.

cross-request: https://github.com/dimagi/commcare-core/pull/335

### DIFF
--- a/scripts/ccc
+++ b/scripts/ccc
@@ -14,6 +14,7 @@ HELP_MSG = """'capture' saves the current session for restoring later.
 'install some_app.ccz' pushes and installs the ccz to device.
 'uninstall app_id' uninstalls an installed app.
 'login username password' logs into the currently seated app
+'invalidate' forces recovery on next sync by providing fake case db hash.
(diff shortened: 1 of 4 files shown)


scripts/ccc
```diff
@@ -156,9 +168,9 @@ def main():
     command = sys.argv[1]
     dispatch = {'capture': lambda: capture(),
                 'uninstall': lambda: uninstall(sys.argv[2]),
-                'login': lambda: login(sys.argv[2], sys.argv[3]),
+                'invalidate': lambda: stage_recover(),
                 'remote': lambda: remote_install(*sys.argv[2:5]),
-                'help': lambda: HELP_MSG,
+                'help': lambda: print(HELP_MSG),
```
phillipm: I think this is the first thing that makes this script explicitly `python2` incompatible...

amstone326:
@phillipm Build is failing

PR #1376, opened 2016-07-13
Fix app install progress count

phillipm:
For for issue where the max value in the progress bar during app installs would often get stuck at `3`.

### DIFF
--- a/app/src/org/commcare/tasks/ResourceEngineTask.java
+++ b/app/src/org/commcare/tasks/ResourceEngineTask.java
@@ -48,9 +48,6 @@ public abstract class ResourceEngineTask<R>
     // launch this task
     private final boolean shouldSleep;
 
-    // last time in system millis that we updated the status dialog
-    private long lastTime = 0;
-
     protected String vAvailable;
     protected String vRequired;
     protected boolean majorIsProblem;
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/tasks/ResourceEngineTask.java
```diff
@@ -129,7 +126,7 @@ public void compoundResourceAdded(final ResourceTable table) {
         synchronized (statusLock) {
             // if last time isn't set or is less than our spacing count, do not
             // perform status update. Also if we are already running one, just skip this.
-            if (statusCheckRunning || System.currentTimeMillis() - lastTime < ResourceEngineTask.STATUS_UPDATE_WAIT_TIME) {
```
amstone326: I'm a bit confused how this could have been causing it to get stuck, would you mind explaining?

Also, should delete the `STATUS_UPDATE_WAIT_TIME` variable, which is now unused.

phillipm: Clayton added this threaded process so that updating the install progress wouldn't block actually performing the install. He limited such that it only ran once every second. Since I've updated entire recalculations to only happen when you add a 'compound' resources (media suite, suite, and profile resources), we shouldn't need a timeout prevent how often this runs because for most apps it should only run 2-3 times an install.

PR #1377, opened 2016-07-13
Move case claim code into commcare-core

phillipm:
Move modern http request code into `commcare-core` so it can be shared by CloudCare, the CLI, and `commcare-android`.

cross-request: https://github.com/dimagi/commcare-core/pull/337


phillipm:
@wpride here is the refactor that moves the core case claim http code into core so you can re-use it in FormPlayer. You will still need to instrument some things around the session to fire off the http requests, but it should hopefully be way easier now.

wpride:
@phillipm I am sorry I only have but one rock to give

PR #1378, opened 2016-07-13
Ellipses

sjain28:
Adds ellipses to the end of long form/module names, instead of cutting the names off. 
In response to http://manage.dimagi.com/default.asp?222398#1182919
and http://manage.dimagi.com/default.asp?191063

New Behavior: 

grid
list

### DIFF
--- a/app/res/layout/menu_list_item_modern.xml
+++ b/app/res/layout/menu_list_item_modern.xml
@@ -15,7 +15,10 @@
         android:id="@+id/row_txt"
         android:layout_width="wrap_content"
         android:layout_weight="1"
-        style="@style/ListContentV2"/>
+        style="@style/ListContentV2"
+        android:singleLine="false"
+        android:ellipsize="end"
+        android:maxLines="2"/>
 
     <org.commcare.views.media.AudioButton
         android:id="@+id/row_soundicon"
--- a/app/res/layout/menu_list_item_modern.xml
+++ b/app/res/layout/menu_list_item_modern.xml
@@ -16,7 +16,6 @@
         android:layout_width="wrap_content"
         android:layout_weight="1"
         style="@style/ListContentV2"
-        android:singleLine="false"
         android:ellipsize="end"
         android:maxLines="2"/>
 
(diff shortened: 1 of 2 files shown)


amstone326:
Ah yay!! This is long-needed. There's an old ticket lingering around somewhere (different from the one you linked) that was also about this that I should find so we can notify them when this is merged.

app/res/layout/menu_list_item_modern.xml
```diff
@@ -15,7 +15,9 @@
         android:id="@+id/row_txt"
         android:layout_width="wrap_content"
         android:layout_weight="1"
-        style="@style/ListContentV2"/>
+        style="@style/ListContentV2"
+        android:ellipsize="end"
+        android:maxLines="2"/>
```
wpride: Was setting the maxLines required to use ellipsize?

sjain28: Yeah, it only ellipsizes if the text spans the max number of lines.

amstone326:
Here it is: http://manage.dimagi.com/default.asp?191063. @sjain28 Once this merged, you should comment on this ticket and notify them that you've implemented a solution! I think they were hoping that we would do something to make the whole form name show up, but I think we determined that that wasn't something we were willing/able to do, at least not without a lot more effort, so this seems like the best possible alternative!

ctsims:
@sjain28 Nice! Great job on the quick turnaround.

Just checked on a big-ish tablet and it looks like two-lines is all that fits on that as well, so that limitation should be fine for us, I think. Better for the UI to be clear on it.

amstone326:
@sjain28 Is this ready to be reviewed? Just wasn't sure because it doesn't have the tag

sjain28:
@amstone326 yeah it is - Will add tags to all reviewable PRs.

amstone326:
@sjain28 Would you be up for talking to Clark about the best way to add this to the test plan?

sjain28:
@amstone326 it's added

amstone326:
@sjain28 Awesome! Thanks

PR #1380, opened 2016-07-13
Reduce GPS auto-capture timeout / Stop polling after form completes

phillipm:
Reduce GPS auto-capture timeout from `5` minutes to `2` minutes.

Change acceptable accuracy from `5` meters to `10` meters. Assuming that since my device had trouble getting to `5` meters in Boston on a network, older devices would have way more trouble getting anything near that in the field.

Add ability to unregister GPS auto-capture polling when a form completes. This will save a bunch of battery hopefully.

Add a singleton class that controls getting location updates from the location manager and forwarding them to all the poll sensor actions that have been registered with it. This is needed because unregistering the location manager requires having a reference around to all the objects that have been registered with that manager. This can be difficult if we have a poll action that starts polling on form start and another that stops it on form completion.

When reviewing please look out for memory leaks caused by dangling references. I've tried to be careful about this, but always good to get more eyes on it. If the form is cancelled instead of completed, there will be a dangling reference from the `PollSensorController` to all the `PollSensorAction`s registered to it by the form until the timeout timer clears them. 

**release notes**: stop GPS auto-capture upon form exit


app/src/org/commcare/android/javarosa/PollSensorAction.java
```diff
@@ -49,27 +50,10 @@
     public static final String XPATH_ERROR_ACTION = "poll_sensor_xpath_error_action";
     private TreeReference target;
 
-    private LocationManager mLocationManager;
+    private static LocationManager mLocationManager;
```
phillipm: Has to be static because there will be 2 `PollSensorAction` instances per form when there are `xforms-ready` and `xforms-revalidate` actions

phillipm:
@orangejenny After talking to Clayton about my previous approach I realized that there was a better way to stop polling on form exit: create a singleton that manages location for polling and allows poll actions to register with it. Doing this allows us to cancel polling on both form completion and cancellation (last approach couldn't handle this) by stopping the singleton location controller in `CommCareHomeAcitivity.onActivityResult`  when returning from `FormEntryActivity`.

This means that your HQ PR isn't actually needed.. sorry that I didn't realize this earlier! Can we revert the changes that put a `xforms-revalidate` pollsensor action into forms?

ctsims:
Hey @phillipm this approach checks out to me overall.

One final tweak: In my mind it would be a bit cleaner to tie the location controller in the FormEntryActivity lifecycle itself. If we, for instance, called out to the form entry activity from a different context from the home activity in the future it's a bit awkward for it to be linked.

Can we move the stopLocationPolling call to FormEntryActivity and tie it to the onPause event when activity.isFinishing is true? That would accomplish the same thing without leaking the state abstraction to the lifecycle of another object.

phillipm:
Good call, updated and tested it

PR #1381, opened 2016-07-13
Recover form view state after rotation or callout

amstone326:
-After rotating the screen during form entry, restores focus to the most recently updated widget 
-After performing a callout during form entry, restores focus to the widget that the callout was for (i.e.  image/video/sound capture/choose)

### DIFF
--- a/app/src/org/commcare/views/QuestionsView.java
+++ b/app/src/org/commcare/views/QuestionsView.java
@@ -4,6 +4,7 @@ import android.content.Context;
 import android.content.SharedPreferences;
 import android.graphics.Typeface;
 import android.os.Build;
+import android.os.Bundle;
 import android.preference.PreferenceManager;
 import android.text.Spannable;
 import android.text.SpannableStringBuilder;
@@ -139,7 +140,7 @@ public class QuestionsView extends ScrollView
             qw.setChangedListener(this);
         }
         
-        updateLastQuestion();
+        markLastStringWidget();
 
         addView(mView);
     }
@@ -231,9 +232,9 @@ public class QuestionsView extends ScrollView
         super.onMeasure(widthMeasureSpec, heightMeasureSpec);
     }
     
-    private void updateConstraintRelevancies(){
-        if(hasListener){
-            wcListener.widgetEntryChanged();
+    private void updateConstraintRelevancies(QuestionWidget changedWidget) {
+        if (hasListener) {
+            wcListener.widgetEntryChanged(changedWidget);
         }
     }
 
@@ -320,8 +321,10 @@ public class QuestionsView extends ScrollView
         }
     }
 
-    public void setFocus(Context context) {
-        if (widgets.size() > 0) {
+    public void setFocus(Context context, int indexOfLastChangedWidget) {
+        if (indexOfLastChangedWidget != -1) {
+            widgets.get(indexOfLastChangedWidget).setFocus(context);
+        } else if (widgets.size() > 0) {
             widgets.get(0).setFocus(context);
         }
     }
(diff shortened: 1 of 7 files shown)


app/src/org/commcare/views/QuestionsView.java
```diff
@@ -320,10 +322,42 @@ private void addHintText(String hintText) {
...
-            widgets.get(0).setFocus(context);
+    public void setFocus(Context context, int indexOfLastChangedWidget) {
+        QuestionWidget widgetToFocus = null;
+        if (indexOfLastChangedWidget != -1) {
+            widgetToFocus = widgets.get(indexOfLastChangedWidget);
+        } else if (widgets.size() > 0) {
+            widgetToFocus = widgets.get(0);
         }
+        if (widgetToFocus != null) {
+            scrollToWidget(widgetToFocus);
```
amstone326: Scrolling to a widget is needed for any widget that does not involve some sort of text entry (and therefore will not get scrolled to automatically by calling 'setFocus' on it)

app/src/org/commcare/views/QuestionsView.java
```diff
@@ -320,10 +322,42 @@ private void addHintText(String hintText) {
...
-        if (widgets.size() > 0) {
-            widgets.get(0).setFocus(context);
+    public void setFocus(Context context, int indexOfLastChangedWidget) {
+        QuestionWidget widgetToFocus = null;
+        if (indexOfLastChangedWidget != -1) {
+            widgetToFocus = widgets.get(indexOfLastChangedWidget);
+        } else if (widgets.size() > 0) {
+            widgetToFocus = widgets.get(0);
         }
+        if (widgetToFocus != null) {
```
amstone326: I call both of the methods below in order to focus on any widget because: 
-Scrolling to a widget is needed for any widget that does not involve some sort of text entry (and therefore will not get scrolled to automatically by calling 'setFocus' on it)
-But, `setFocus` still has some smaller side effects for those widgets, such as hiding the keyboard.

app/src/org/commcare/views/QuestionsView.java
```diff
@@ -320,10 +322,42 @@ private void addHintText(String hintText) {
...
+            scrollToWidget(widgetToFocus);
+            widgetToFocus.setFocus(context);
+        }
+    }
+
+    private void scrollToWidget(final QuestionWidget widget) {
+        new Handler().post(new Runnable() {
+            @Override
+            public void run() {
+                QuestionsView.this.scrollTo(0, widget.getTop());
```
amstone326: Think it's an open question whether it's best to scroll to the top of a widget or the bottom (or somewhere in between?). Input from others would be great.

app/src/org/commcare/views/QuestionsView.java
```diff
@@ -340,7 +374,6 @@ public void setBinaryData(Object answer, PendingCalloutInterface pendingCalloutI
         for (QuestionWidget q : widgets) {
             if (questionFormIndex.equals(q.getFormId())) {
                 q.setBinaryData(answer);
-                pendingCalloutInterface.setPendingCalloutFormIndex(null);
```
amstone326: I removed this call because it causes `getPendingWidget` in `FormEntryActivity` to return null at the point that I call it in `restoreFocusToCalloutQuestion`. I'm pretty sure this is fine because `onResumeSessionSafe` of `FormEntryActivity` will always calls `mFormController.setPendingCalloutFormIndex(null)` after returning from the callout. But just wanted to flag in case anyone sees an issue. I think this was just here as a more proactive, but not strictly necessary, place to set the index back to null.

app/src/org/commcare/views/widgets/SpinnerMultiWidget.java
```diff
@@ -111,7 +111,7 @@ public void onClick(DialogInterface dialog, int which, boolean isChecked) {
                                 selections[which] = isChecked;
 
                                 if (hasListener()) {
-                                    widgetChangedListener.widgetEntryChanged();
+                                    widgetChangedListener.widgetEntryChanged(SpinnerMultiWidget.this);
                                 }
```
phillipm: I'm wondering: can every insteance of `if (hasListener()) {
widgetChangedListener.widgetEntryChanged(this); }` can just be replaced by a call to `QuestionWidget.widgetEntryChanged()`?

phillipm:
Awesome! Ready to merge after addressing my question

app/src/org/commcare/views/widgets/SpinnerMultiWidget.java
amstone326: oo interesting, that seems like it would make sense. I can try it out

amstone326: updated!

PR #1386, opened 2016-07-14
Fix for bug breaking multimedia submissions with SD card transfer

wpride:
Two bugs:
First, when there was a subdirectory in the form instance directory, the activity would fail here but create the impression that it was successful. No potential for data loss, but misleading and non functional. The solution was the filter folders out before performing the copy.

Second, the name of files to be zipped was reversed so instead of folder/name it would be name/folder. https://github.com/dimagi/commcare-android/pull/1386/files#diff-af6d6e22c0533d597d5358277debfedeR100. Not clear why this sometimes works and sometimes does not. Investigating further. Should be hotfixed ASAP

### DIFF
--- a/app/src/org/commcare/tasks/ZipTask.java
+++ b/app/src/org/commcare/tasks/ZipTask.java
@@ -97,7 +97,7 @@ public abstract class ZipTask extends CommCareTask<Void, String, ZipTask.ZipTask
 
                 int pathPartsLength = pathParts.length;
 
-                String filePath = pathParts[pathPartsLength -1] + "/" + pathParts[pathPartsLength - 2];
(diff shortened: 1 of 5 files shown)


app/src/org/commcare/tasks/ZipTask.java
```diff
@@ -97,7 +97,7 @@ private static void zipInstanceFolder(File[] toBeZippedFiles, ZipOutputStream zo
 
                 int pathPartsLength = pathParts.length;
 
-                String filePath = pathParts[pathPartsLength -1] + "/" + pathParts[pathPartsLength - 2];
+                String filePath = pathParts[pathPartsLength -2] + "/" + pathParts[pathPartsLength - 1];
```
wpride: @phillipm

PR #1390, opened 2016-07-14
Wsp/sd fix 228

wpride:
Two bugs:
First, when there was a subdirectory in the form instance directory, the activity would fail here but create the impression that it was successful. No potential for data loss, but misleading and non functional. The solution was the filter folders out before performing the copy.

Second, the name of files to be zipped was reversed so instead of folder/name it would be name/folder. https://github.com/dimagi/commcare-android/pull/1386/files#diff-af6d6e22c0533d597d5358277debfedeR100. Not clear why this sometimes works and sometimes does not. Investigating further. Should be hotfixed ASA

### DIFF
--- a/app/src/org/commcare/activities/CommCareWiFiDirectActivity.java
+++ b/app/src/org/commcare/activities/CommCareWiFiDirectActivity.java
@@ -660,8 +660,6 @@ public class CommCareWiFiDirectActivity
 
 
     private void onRecordPullCompleted(Pair<Long, FormRecord[]> result, CommCareWiFiDirectActivity receiver) {
-
-        // for the time being we're going to ignore the result of the record pull and proceed regardless
         myStatusText.setText(localize("wifi.direct.pull.successful"));
         if(result != null){
             if(result.first > 0){
--- a/app/src/org/commcare/activities/CommCareWiFiDirectActivity.java
+++ b/app/src/org/commcare/activities/CommCareWiFiDirectActivity.java
@@ -58,6 +58,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.util.Arrays;
 import java.util.Vector;
 
 /**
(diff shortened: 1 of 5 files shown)


app/src/org/commcare/tasks/DumpTask.java
```diff
@@ -39,35 +41,47 @@
...
     private long dumpInstance(File folder, SecretKeySpec key) throws FileNotFoundException {
-        File[] files = folder.listFiles();
-        
+
+        Logger.log(TAG, "Dumping form instance at folder: " + folder);
+
+        File[] files = folder.listFiles(new FileFilter() {
+            @Override
+            public boolean accept(File file) {
+                return file.isFile();
```
wpride: these lines filter out folders so that we don't try copying them

app/src/org/commcare/tasks/FormRecordToFileTask.java
```diff
@@ -62,7 +63,12 @@ public FormRecordToFileTask(Context c, String formStoragePath) {
      * Return an int status code from FormUploadUtil corresponding to the outcome of the transfer
      */
     private long copyFileInstanceFromStorage(File formRecordFolder, SecretKeySpec decryptionKey) {
-        File[] files = formRecordFolder.listFiles();
+        File[] files = formRecordFolder.listFiles(new FileFilter() {
+            @Override
+            public boolean accept(File file) {
+                return file.isFile();
```
wpride: also filters out folders

app/src/org/commcare/tasks/ZipTask.java
```diff
@@ -97,7 +97,7 @@ private static void zipInstanceFolder(File[] toBeZippedFiles, ZipOutputStream zo
 
                 int pathPartsLength = pathParts.length;
 
-                String filePath = pathParts[pathPartsLength -1] + "/" + pathParts[pathPartsLength - 2];
+                String filePath = pathParts[pathPartsLength -2] + "/" + pathParts[pathPartsLength - 1];
```
wpride: this line reverses the path/name order to the correct order (see above)

app/src/org/commcare/activities/CommCareWiFiDirectActivity.java
```diff
@@ -658,11 +659,19 @@ private void onUnzipSuccessful(Integer result) {
     }
 
 
-    private void onRecordPullCompleted(Pair<Long, FormRecord[]> result) {
+    private void onRecordPullCompleted(Pair<Long, FormRecord[]> result, CommCareWiFiDirectActivity receiver) {
+
         // for the time being we're going to ignore the result of the record pull and proceed regardless
```
amstone326: this comment seems stale now

app/src/org/commcare/tasks/DumpTask.java
```diff
@@ -243,14 +260,18 @@ protected Boolean doTaskBackground(String... params) {
...
+
                 long result = 0;
                 for(int i = 0 ; i < records.length ; ++ i) {
                     if(results[i] > result) {
                         result = results[i];
                     }
                 }
-                
+
+            if(result > 0){
```
amstone326: what is this addition doing?

PR #1391, opened 2016-07-15
Allow fixtures in suite.xml to be updated on app update

phillipm:
cross-request: https://github.com/dimagi/commcare-core/pull/341

Includes test that verifies that an app update with change to a suite.xml fixture indeed updates the fixture in storage


ctsims:
Left a comment on the X-requested PR: This approach is strictly better than the current world, but there's still a potential scenario in which someone can receive and be "upgraded" but fail to have the new fixture.  We can address that in the future if we want to do this for the 99% likely case (we should tag this as a future problem to solve though)

PR #1394, opened 2016-07-18
Fix for Gradle Builds on Windows

ctsims:
Gradle config changes were introducing unescaped slashes into the config files.


wpride:
👍 on ✅

PR #1396, opened 2016-07-18
Fix SMS Install

wpride:
Looks like SMS install broke at some point when the requirements for a CommCareTask changed. This fixes that. Also adds the spongy castle algorithm to the proguard keep file since those were being eaten. 

Don't think we need a hotfix unless @gcapalbo disagrees

Needs https://github.com/dimagi/commcare-android/pull/1406


app/proguard.cfg
```diff
@@ -46,6 +47,7 @@
 -keep,includedescriptorclasses class org.odk.collect.android.** { *; }
 -keep,includedescriptorclasses class org.javarosa.** { *; }
 -keep class in.uncod.android.bypass.** { *; }
+-keep class org.spongycastle.** { *; }
```
wpride: @amstone326 this is the line that solves the NoSuchAlgorithmException issue

gcapalbo:
(offlined, answer is yes we should hotfix)

app/src/org/commcare/activities/CommCareSetupActivity.java
```diff
@@ -607,6 +607,7 @@ protected void deliverError(CommCareSetupActivity receiver, Exception e) {
                                     }
                                 }
                             };
+                    mTask.connect(this);
```
phillipm: If you change `RetrieveParseVerifyMessageTask mTask` to `RetrieveParseVerifyMessageTask<CommCareSetupActivity> mTask` it gets rid of the lint warning here

app/src/org/commcare/tasks/RetrieveParseVerifyMessageTask.java
```diff
@@ -18,9 +18,9 @@ public RetrieveParseVerifyMessageTask(RetrieveParseVerifyMessageListener listene
     }
 
     @Override
-    protected String doTaskBackground(String[] params) {
+    protected String doTaskBackground(Object... params) {
         try {
-            String url = SigningUtil.trimMessagePayload(params[0]);
+            String url = SigningUtil.trimMessagePayload(params[0].toString());
```
phillipm: I don't understand why you have to pass this in as an `Object`

amstone326:
@wpride So is this getting hotfixed?

wpride:
@amstone326 yes but its not urgent and there's still one bug I'm working out

amstone326:
Got it sounds good

PR #1397, opened 2016-07-18
Case alerts

sjain28:
Allows user to set alerts based on the values of case properties.


app/src/org/commcare/activities/ReminderThread.java
```diff
@@ -0,0 +1,139 @@
...
+        synchronized(myThread) {
+            if(!myThread.isAlive()) {
+                myThread.start();
+            }
+        }
+    }
+    
+    public void stopService() {
+        synchronized(myThread) {
+            mContinuePolling = false;
```
ctsims: It would be helpful to interrupt the thread inside of this as well to force it to process the threat stoppage.

app/src/org/commcare/activities/CommCareReminderService.java
```diff
@@ -0,0 +1,130 @@
...
+        
+        //CharSequence text = "Session Expires: " + DateFormat.format("MMM dd h:mmaa", sessionExpireDate);
+
+        //We always want this click to simply bring the live stack back to the top
+        Intent callable = new Intent(this, CommCareHomeActivity.class);
+        callable.setAction("android.intent.action.MAIN");
+        callable.addCategory("android.intent.category.LAUNCHER");  
+
+        // The PendingIntent to launch our activity if the user selects this notification
+        PendingIntent contentIntent = PendingIntent.getActivity(this, 0, callable, 0);
```
ctsims: I think we can skip the visible notification of this service for the reminder thread and have it just run in the background;.

app/src/org/commcare/activities/CommCareReminderService.java
```diff
@@ -0,0 +1,130 @@
...
+
+        mReminderJob = new ReminderThread(CommCareApplication._());
+        
+        mReminderJob.startPolling(alerts);
+
+        //Send the notification.
+        this.startForeground(NOTIFICATION, notification);
+    }
+    
+    public void stop() {
```
ctsims: What ends up ending this service exactly? It's not super clear what the path is once, say, someone logs out to the service ending

app/src/org/commcare/activities/CommCareHomeActivity.java
```diff
@@ -158,10 +164,24 @@
...
+
     @Override
     protected void onCreateSessionSafe(Bundle savedInstanceState) {
         super.onCreateSessionSafe(savedInstanceState);
 
+        for(Suite s: CommCareApplication._().getCurrentApp().getCommCarePlatform().getInstalledSuites()){
+            allAlerts.addAll(s.getAlerts());
+        }
+
+        launchService(allAlerts);
```
ctsims: I think we want to avoid launching this particular service unless we did get alerts out of the suite. IE: For most apps it would be best if the new service didn't launch at all.

app/src/org/commcare/activities/CommCareHomeActivity.java
```diff
@@ -158,10 +164,24 @@
...
+    private Object serviceLock = new Object();
+    boolean mIsBound = false;
+    boolean mIsBinding = false;
+    private Vector<Alert> allAlerts = new Vector<>();
+
     @Override
     protected void onCreateSessionSafe(Bundle savedInstanceState) {
         super.onCreateSessionSafe(savedInstanceState);
 
+        for(Suite s: CommCareApplication._().getCurrentApp().getCommCarePlatform().getInstalledSuites()){
```
ctsims: I'd add this loop to the inside of the launch service method itself

app/src/org/commcare/activities/CommCareHomeActivity.java
```diff
@@ -158,10 +164,24 @@
 
     private boolean sessionNavigationProceedingAfterOnResume;
 
+    //REMINDER STUFF
```
ctsims: I think we might want to compartmentalize this code a bit more cleanly somewhere other than this activity, since this code and process isn't really bound to the lifecycle of the CommCareHomeActivity.

The Session Service code actually ends up just being able to live inside of that class, which may be a better place for a lot of these helper objects.

app/src/org/commcare/activities/CommCareHomeActivity.java
```diff
@@ -158,10 +164,24 @@
...
+    private CommCareReminderService mBoundService;
+    private ServiceConnection mConnection;
+    private Object serviceLock = new Object();
+    boolean mIsBound = false;
+    boolean mIsBinding = false;
+    private Vector<Alert> allAlerts = new Vector<>();
+
     @Override
     protected void onCreateSessionSafe(Bundle savedInstanceState) {
```
ctsims: I'm not super familiar with this entry point, is its lifecycle cleanly bound to the notion of a user login/logout event in a way that is reliable? I'm a bit surprised to see things firing from here v. somewhere in the CommCareSessionService or somewhere less tied to user navigation

app/src/org/commcare/activities/ReminderThread.java
```diff
@@ -0,0 +1,139 @@
...
+                try{
+                    condition = XPathParseTool.parseXPath(xPathConditionString);
+                }catch (Exception e){
+                    e.printStackTrace();
+                    return;
+                }
+
+                //Prepare evaluation context
+                Hashtable<String, DataInstance> instances = new Hashtable<>();
+                instances.put("casedb", new ExternalDataInstance("jr://instance/casedb","casedb"));
```
ctsims: This seems good to bootstrap from here for testing, but I think we want to move loading these "connector" objects to be driven by data from inside the Alert object itself.

Let me know if you want some pointers for how to set that up.

app/src/org/commcare/activities/ReminderThread.java
```diff
@@ -0,0 +1,139 @@
...
+                EvaluationContext ec = CommCareApplication._().getCurrentSessionWrapper().getEvaluationContext(instances);
+
+                //Prepare TreeReferences
+                String caseXPathRef = "instance('casedb')/casedb/case[@case_type='" + caseTypeString + "']";
```
ctsims: Rather than loading the caseType as the external element here, I think we want to shift the Alert to load this full string (the "nodeset" reference), which would give us more flexibility in setting what types of reminders are available

app/src/org/commcare/activities/ReminderThread.java
```diff
@@ -0,0 +1,139 @@
...
+                TreeReference caseType = XPathReference.getPathExpr(caseXPathRef).getReference(); //This is more of an abstract reference to a bunch of stuff in the tree
+                List<TreeReference> refs = ec.expandReference(caseType); //When we expand the reference within the context ec we get the x number of references that actually "exist"
```
ctsims: We will need to re-run this expression either once-per loop below or periodically.

In the event that, say, a user adds a new case of a type, or removes one, the list of "resulting" references will change, so we could get bugs or miss new nodes for evaluation if we don't renew it.

app/src/org/commcare/activities/ReminderThread.java
```diff
@@ -0,0 +1,139 @@
...
+
+                while(mContinuePolling) {
+                    long current = System.currentTimeMillis();
+
+                    //Evaluate condition
+                    for(TreeReference ref : refs) {
+                        EvaluationContext internal = new EvaluationContext(ec, ref);
+                        Log.d("Result", condition.eval(internal).toString());
```
ctsims: For runtime considerations, I'd cache the results of "condition.eval()" into an object, since you do it again right below and those evaluations could take a considerable amonut of time to evalute

app/src/org/commcare/activities/ReminderThread.java
```diff
@@ -0,0 +1,139 @@
...
+
+                        if((Boolean) condition.eval(internal)){
```
ctsims: It would be helpful to catch and throw a very direct error if this result object isn't a Boolean type. it's quite difficult to debug those issues for end users who are configuring these expressions, so having as much context as possible is very helpful. Can chat about good mechanisms to make that work.

app/src/org/commcare/activities/ReminderThread.java
```diff
@@ -0,0 +1,139 @@
...
+                String caseXPathRef = "instance('casedb')/casedb/case[@case_type='" + caseTypeString + "']";
+                TreeReference caseType = XPathReference.getPathExpr(caseXPathRef).getReference(); //This is more of an abstract reference to a bunch of stuff in the tree
+                List<TreeReference> refs = ec.expandReference(caseType); //When we expand the reference within the context ec we get the x number of references that actually "exist"
+
+                while(mContinuePolling) {
+                    long current = System.currentTimeMillis();
+
+                    //Evaluate condition
+                    for(TreeReference ref : refs) {
+                        EvaluationContext internal = new EvaluationContext(ec, ref);
```
ctsims: Since these expressions are so much more potentially error prone, it would be great to wrap the inside of this loop (or the inside of the while loop before the thread.sleep() in a try/catch, since otherwise the reminder thread can terminate and there's no clear way to restart it.

app/src/org/commcare/activities/CommCareHomeActivity.java
sjain28: The CommCareHomeActivity gets launched on login, and destroyed upon logout.

app/src/org/commcare/activities/ReminderThread.java
sjain28: Do both of these lines need to get re-run, or just the expandReference in line 76?

ctsims: Just expandReference

sjain28:
@ctsims ReminderThread lifecycle is now bound to SessionService, and thread can be stopped if errors occur. Currently using notifications to tell users if thread stops - A more aggressive option like a popup dialog requires access to an activity in the ReminderThread.

PR #1406, opened 2016-07-20
Add multidexing

wpride:
Needed for https://github.com/dimagi/commcare-android/pull/1396

I think keeping the bouncy castle classes put us over the limit. @ctsims @phillipm any idea if this will introduce other issues?


wpride:
this appears to be unneeded and was an issue with my local setup

PR #1407, opened 2016-07-20
Backmerge CommCare 2.28.2 and CommCare 2.28.3 point release code

ctsims:
@wspride Can you confirm that this has everything from 2.28.2?

### DIFF
--- a/app/src/org/commcare/activities/CommCareFormDumpActivity.java
+++ b/app/src/org/commcare/activities/CommCareFormDumpActivity.java
@@ -77,7 +77,7 @@ public class CommCareFormDumpActivity extends SessionAwareCommCareActivity<CommC
             public void onClick(View v) {
 
                 formsOnSD = getDumpFiles().length;
-                Logger.log(AndroidLogger.TYPE_FORM_DUMP, "Send task found " + formsOnSD + " forms on the SD card.");
+                Logger.log(AndroidLogger.TYPE_FORM_DUMP, "Send task found " + formsOnSD + " forms on the phone.");
 
                 //if there're no forms to dump, just return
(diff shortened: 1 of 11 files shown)


phillipm:
This includes changes from 2.28.1, which I didn't backmerge because there is a more [comprehensive PR](https://github.com/dimagi/commcare-android/pull/1358) that encapsulates those changes currently targeting master. I'm going to close this and recreate it to avoid potential merge conflicts

phillipm:
Err, not 2.28.1, sorry, but this PR still has more commits than just 2.28.2 and 2.28.3

ctsims:
@phillipm Ok, that makes sense.

In general it's been my best-practices approach that release  branches should always merge directly back into master to avoid human-attention errors where bugfixes get "missed" more or less. Do you feel like that isn't worth the overhead of issues like this? I'm a bit mixed on the outcomes WRT things like lower-impact bugfixes which are engineering to be minor.

phillipm:
Ah, yeah, I think I started using cherry-pick more than I should have and it creates new SHA1's, so that is why there are more commits in this PR than expected.

Regarding the split between https://github.com/dimagi/commcare-android/pull/1364 and  https://github.com/dimagi/commcare-android/pull/1358 , the `2.28` version was added at the last minute, so I simplified it a lot over the version targeting `master`. In the future I will go for the direct backmerge and reconcile the changes on the more comprehensive, open branch

ctsims:
@phillipm Cool. Yeah, I don't have strong feelings either way, and totally recognize that sometimes it's cleaner / less error prone to cherry-pick/replicate, just wanted to double check on the best practice for our releases these days, thanks!

PR #1403, opened 2016-07-19
Handle session expiring when setting up home activity menu

phillipm:
Lots of 2.28 ACRA crashes with this one

```
java.lang.RuntimeException: Unable to start activity ComponentInfo{org.commcare.dalvik/org.commcare.activities.CommCareHomeActivity}: org.commcare.utils.SessionUnavailableException
at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2790)
at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2855)
at android.app.ActivityThread.access$900(ActivityThread.java:181)
at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1474)
at android.os.Handler.dispatchMessage(Handler.java:102)
at android.os.Looper.loop(Looper.java:145)
at android.app.ActivityThread.main(ActivityThread.java:6117)
at java.lang.reflect.Method.invoke(Native Method)
at java.lang.reflect.Method.invoke(Method.java:372)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1399)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1194)
Caused by: org.commcare.utils.SessionUnavailableException
at org.commcare.CommCareApplication.getSession(CommCareApplication.java:1187)
at org.commcare.activities.CommCareHomeActivity.onPrepareOptionsMenu(CommCareHomeActivity.java:1205)
at android.app.Activity.onPreparePanel(Activity.java:2982)
at android.support.v4.app.FragmentActivity.onPrepareOptionsPanel(FragmentActivity.java:529)
at android.support.v4.app.FragmentActivity.onPreparePanel(FragmentActivity.java:518)
at com.android.internal.policy.impl.PhoneWindow.preparePanel(PhoneWindow.java:648)
at com.android.internal.policy.impl.PhoneWindow.doInvalidatePanelMenu(PhoneWindow.java:994)
at com.android.internal.policy.impl.PhoneWindow$1.run(PhoneWindow.java:322)
at com.android.internal.policy.impl.PhoneWindow.doPendingInvalidatePanelMenu(PhoneWindow.java:968)
at com.android.internal.policy.impl.PhoneWindow.restoreHierarchyState(PhoneWindow.java:2284)
at android.app.Activity.onRestoreInstanceState(Activity.java:1097)
at android.app.Activity.performRestoreInstanceState(Activity.java:1042)
[...]

PR #1411, opened 2016-07-20
Hotfix 2.28.4 (Part I): Small fixes for some ACRA crashes & signature widget

phillipm:
2.28 versions of the following PRs:

- https://github.com/dimagi/commcare-android/pull/1405
- https://github.com/dimagi/commcare-android/pull/1403
- https://github.com/dimagi/commcare-android/pull/1408

### DIFF
--- a/app/src/org/commcare/activities/CommCareHomeActivity.java
+++ b/app/src/org/commcare/activities/CommCareHomeActivity.java
@@ -1176,7 +1176,9 @@ public class CommCareHomeActivity
             User u = CommCareApplication._().getSession().getLoggedInUser();
             return (User.TYPE_DEMO.equals(u.getUserType()));
         } catch (SessionUnavailableException e) {
-            return true;
+            // Default to a normal user: this should only happen if session
+            // expires and hasn't redirected to login.
+            return false;
         }
     }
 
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/activities/CommCareHomeActivity.java
```diff
@@ -1172,8 +1172,14 @@ void displayMessage(String message, boolean bad, boolean suppressToast) {
     }
 
     public static boolean isDemoUser() {
-        User u = CommCareApplication._().getSession().getLoggedInUser();
-        return (User.TYPE_DEMO.equals(u.getUserType()));
+        try {
+            User u = CommCareApplication._().getSession().getLoggedInUser();
+            return (User.TYPE_DEMO.equals(u.getUserType()));
+        } catch (SessionUnavailableException e) {
+            // Default to a normal user: this should only happen if session
```
wpride: Shouldn't we redirect to login in this case? Is there the possibility a Demo user could end up with some functionality they shouldn't and end up crashing?

wpride:
one question then gtm

app/src/org/commcare/activities/CommCareHomeActivity.java
phillipm: Redirection to login should happen in `onResume` of any activity that extends `SessionAwareCommCareActivity`. Adding this `try/catch` ensures we don't crash if `isDemoUser` is invoked somewhere before `onResume` gets called.

phillipm: But you're right, it is good to be cautious about this default value! I just checked and the uses of `isDemoeUser` toggle warning dialogs and enable the settings menu. If those don't or do show up when they shouldn't we should be fine.

PR #1417, opened 2016-07-22
Setup for async restore integration test

amstone326:


app/src/org/commcare/network/HttpRequestGenerator.java
```diff
@@ -168,6 +169,12 @@ public HttpResponse makeCaseFetchRequest(String baseUri, boolean includeStateFla
         //Add items count to fetch request
         serverUri = serverUri.buildUpon().appendQueryParameter("items", "true").build();
 
+        if (CommCareApplication._().shouldInvalidateCacheOnRestore()) {
+            serverUri = serverUri.buildUpon().appendQueryParameter("overwrite_cache", "true").build();
```
phillipm: Can you add a comment describing what this does / why we want it?

amstone326: I had figured it was sort of apparent what it does from the name of the parameter. Should I add something that says it's currently only used for testing purposes?

phillipm: Yeah, I think the why is very useful in this context. For instance, I have no idea how `serverUri.buildUpon().appendQueryParameter("items", "true").build();` is used by HQ and since it isn't in our code base, I can't just go look it up.

PR #1420, opened 2016-07-27
Persist install error messages

amstone326:
Addresses http://manage.dimagi.com/default.asp?223803.

Instead of using toasts to show error messages during app install, place the error message on the screen itself. (The error text will disappear if you go to any other screen and then return).

Before:
screenshot_20160727-113857

After:
screenshot_20160727-114044


phillipm:
yay! 🎉

PR #1423, opened 2016-07-27
Zebra Bluetooth Printer Integration 

ctsims:
Wires in our Zebra Printing library and attaches the concept of a compound intent callout, which allows some intent callouts to process bulk inputs (but not outputs) for processing.

External Printing code available at:
https://github.com/dimagi/zebra-print-android

It is not wired into our core code to allow some cleaner separation of some awkward library includes that the zebra SDK requires.


app/AndroidManifest.xml
```diff
@@ -49,7 +51,7 @@
 
     <application
         android:allowBackup="false"
-        tools:replace="android:label,android:icon"
+        tools:replace="android:label,android:icon, android:allowBackup"
```
phillipm: what is this needed for?

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -634,6 +654,13 @@ private void updateFormRelevancies() {
         ArrayList<Integer> shouldRemoveFromView = new ArrayList<>();
         // Loop through all of the old widgets to determine which ones should stay in the new view
         for (int i = 0; i < oldWidgets.size(); i++) {
+
+            //Intent widgets need to be fully rebuilt to update their intent callouts
+            //depending on model changes.
+            if(oldWidgets.get(i) instanceof IntentWidget) {
+                shouldRemoveFromView.add(i);
+                continue;
+            }
```
phillipm: Nice! I ran into this bug recently

app/AndroidManifest.xml
ctsims: The external library doesn't share our backup restriction but sets it, this overrides that flag to prevent a build conflict.

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -1158,6 +1186,38 @@ private void setupGroupLabel() {
...
+
+    private void fireCompoundIntentDispatch() {
+        CompoundIntentList i = this.questionsView.getAggregateIntentCallout();
+        if(i == null) {
+            hideCompoundIntentCalloutButton();
+            Log.e(TAG, "Multiple intent dispatch button shouldn't have been shown");
+            return;
+        }
+
+        this.startActivityForResult(i.getCompoundedIntent(), INTENT_COMPOUND_CALLOUT);
```
phillipm: Is there a reason to start this using `startActivityForResult` and not `startActivity`? Doesn't seem like  `INTENT_COMPOUND_CALLOUT` is used

ctsims: startActivityForResult has different backstack behavior, mostly. We don't do compound processing yet, but if you fire up with startActivity() on its own the system manages keeping state around in the previous activity differently in a way that can drop more state

phillipm: Can you add that as a comment? Not knowing that I would have totally replaced `startActivityForResult` with `startActivity` in a refactor.

app/src/org/commcare/activities/TemplatePrinterActivity.java
```diff
@@ -60,39 +72,114 @@ protected void onCreate(Bundle savedInstanceState) {
...
+                printStyle = TEMPLATE_STYLE_ZPL;
+            } else {
+                printStyle = TEMPLATE_STYLE_HTML;
+            }
+        }
+
+        //Since this and the callout activities are raised as "dialog" activities, they will
+        //recreate themselves on rotation. If we detect that we need to not "re-kick-off" the
+        //activity, it will result in duplicate activities.
+        if(printStyle == TEMPLATE_STYLE_ZPL) {
```
phillipm: `==` -> `.equals` or better, just lift this logic into the place where you set `printStyle` to `TEMPLATE_STYLE_ZPL`

app/src/org/commcare/activities/TemplatePrinterActivity.java
```diff
@@ -60,39 +72,114 @@ protected void onCreate(Bundle savedInstanceState) {
...
+            if(savedInstanceState != null) {
+                return;
+            }
+        }
+
+        if(TEMPLATE_STYLE_ZPL.equals(printStyle)) {
```
phillipm: This seems like the same check as right above

app/src/org/commcare/activities/TemplatePrinterActivity.java
```diff
@@ -182,6 +269,20 @@ public void onPageFinished(WebView view, String url) {
...
+    protected void onActivityResult(int requestCode, int resultCode, Intent data) {
+        super.onActivityResult(requestCode, resultCode, data);
+        if(requestCode == CALLOUT_ZPL) {
+            Intent response = new Intent();
+            if(resultCode != Activity.RESULT_CANCELED) {
+                response.putExtra(IntentCallout.INTENT_RESULT_VALUE, "");
+            }
+            this.setResult(resultCode, response);
+            this.finish();
+            return;
```
phillipm: `return` is unneeded

app/src/org/commcare/logic/PendingCalloutInterface.java
```diff
@@ -16,5 +16,9 @@
 public interface PendingCalloutInterface {
     FormIndex getPendingCalloutFormIndex();
 
+    boolean isPendingCalloutBulk();
+
     void setPendingCalloutFormIndex(FormIndex pendingCalloutFormIndex);
+
+    void setPendingCalloutIsBulk();
```
phillipm: This is never called

app/src/org/commcare/logic/PendingCalloutInterface.java
```diff
@@ -16,5 +16,9 @@
 public interface PendingCalloutInterface {
     FormIndex getPendingCalloutFormIndex();
 
+    boolean isPendingCalloutBulk();
```
phillipm: Can you add a comment explaining what a bulk callout is?

app/src/org/commcare/utils/CompoundIntentList.java
```diff
@@ -0,0 +1,70 @@
...
+        }
+        ArrayList<String>  indices = intent.getStringArrayListExtra(EXTRA_COMPOUND_DATA_INDICES);
+
+        indices.add(index);
+        intent.putExtra(index, toAdd.getExtras());
+        intent.putExtra(EXTRA_COMPOUND_DATA_INDICES, indices);
+        return true;
+    }
+
+    private void copyFromIntent(Intent template, String index) {
```
phillipm: `index` is never referenced in the method's body

app/src/org/commcare/logic/PendingCalloutInterface.java
ctsims: Removed that concept, sorry, I forgot that I had changed how I was preventing that response processing.

PR #1426, opened 2016-07-29
implement the locale data source system

wpride:
Allow Android to continue using its own arrays localization system in Ethiopian and Nepali widgets. This means that forms using those widgets won't be required to define their own localization strings (for older apps). 

Follow up to https://github.com/dimagi/commcare-android/pull/1415

cross-request: https://github.com/dimagi/commcare-core/pull/350


app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -279,6 +281,9 @@ protected void onCreate(Bundle savedInstanceState) {
 
         loadStateFromBundle(savedInstanceState);
 
+        // Need to override CalendarUtil's month localizer
+        CalendarUtils.setArrayDataSource(new AndroidArrayDataSource(this));
```
ctsims: There's a single place in the app we use for a bunch of these, isn't there? Would be nice to not leave it somewhere so central.

app/src/org/commcare/utils/AndroidArrayDataSource.java
```diff
@@ -0,0 +1,24 @@
...
+
+    private Context context;
+
+    public AndroidArrayDataSource(Context context){
+        this.context = context;
+    }
+
+    @Override
+    public String[] getArray(String key) {
+        int resourceId = context.getResources().getIdentifier(key, "array", context.getPackageName());
```
ctsims: Hm, I think this should probably fall back somehow in its current form, or throw a much more explicit error if it fails.

Before we had a guaranteed no-error, since the array was fetched from ID resource. Now it could get removed and our error message would be incredibly vague. Especially since this can now fail due to a change in commcare-core that is quite invisible to the end-user here.

ctsims:
@wpride Added a couple comments.  Would be good to also have a test for this.

PR #1427, opened 2016-07-29
Tests for Storage Backed Predicate Fix

ctsims:
Tests for CommCare Core Fix

cross-request: https://github.com/dimagi/commcare-core/pull/351

### DIFF
--- a/unit-tests/src/org/commcare/android/tests/queries/CaseDbQueryTest.java
+++ b/unit-tests/src/org/commcare/android/tests/queries/CaseDbQueryTest.java
@@ -66,7 +66,19 @@ public class CaseDbQueryTest {
                 "test_case_id_child_2", ec);
     }
 
-    public static void evaluate(String xpath, String expectedValue, EvaluationContext ec) {
+    @Test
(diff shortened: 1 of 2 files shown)


unit-tests/src/org/commcare/android/tests/queries/CaseDbQueryTest.java
```diff
@@ -66,7 +66,19 @@ public void testCaseIndexQueries() {
                 "test_case_id_child_2", ec);
     }
 
-    public static void evaluate(String xpath, String expectedValue, EvaluationContext ec) {
+    @Test
+    public void testCaseOptimizationTriggers() {
```
phillipm: Can we move this to test to `commcare-core`? It looks like it doesn't have anything Android specific

ctsims: I don't actually think we can, since I'm not sure what the SQL-based optimization status is on Android. CloudCare wasn't demonstrating the bug, for instance.

PR #1437, opened 2016-08-05
Multiple apps tweaks

amstone326:
Primary changes made:
1. Compute the md5 hash of a user's email before reporting it to analytics, so that we are not violating GA terms of service.
2. Change the language and wiring of the `GlobalPrivilegeClaimingActivity` so that it is generic to all possible privileges, rather than expecting a specific privilege.
3. Remove all language about a "superuser" privilege, since we are abandoning that.


amstone326:
@ctsims Assigned this to you since you have the context for the changes

amstone326:
**New screenshots:**

What the menu item looks like:
screenshot_20160810-112153

UI of privilege claiming activity before scanning barcode:
screenshot_20160810-112158

UI of privilege claiming activity after scanning barcode:
screenshot_20160810-112335

app/src/org/commcare/utils/EncryptionUtils.java
```diff
@@ -597,4 +597,19 @@ private static boolean writeSubmissionManifest(
 
         return true;
     }
+
+    public static String getMD5HashAsString(String plainText) {
```
ctsims: It might be a bit annoying for us to replicate this exact process in another environment, can we maybe use the Base64 utility to convert the hash byte[] to a base64 string? That's super easy to replicate in, say, Python with common libs.

ctsims:
Looks good, one comment in there.

amstone326:
@ctsims Updated

ctsims:
@amstone326 Great! Ready to pull. Do you know what's up with the failing test, though?

amstone326:
@ctsims Yea, read what Phillip said on the mobile-dev chat

app/src/org/commcare/logging/analytics/GoogleAnalyticsUtils.java
```diff
@@ -311,7 +312,8 @@ public static void reportAppManagerAction(String action) {
     }
 
     public static void reportPrivilegeEnabled(String privilegeName, String username) {
-        reportEvent(GoogleAnalyticsFields.CATEGORY_PRIVILEGE_ENABLED, privilegeName, username);
+        reportEvent(GoogleAnalyticsFields.CATEGORY_PRIVILEGE_ENABLED, privilegeName,
+                Base64.encode(username.getBytes()));
```
ctsims: Fuck, sorry, I said this in a confusing way.

We still need to Md5 the value, since otherwise it can still be read. What we want is to MD5 -> Base64, to get a consistent string version of the MD5 hashed value.

amstone326: @ctsims Ah no should have realized. Fixed now

PR #1438, opened 2016-08-08
Offline demo user restore

phillipm:
Enable CC apps to contain a restore payload for demo users. This will allow users to log in as the demo user without an internet connection.

cross-request: https://github.com/dimagi/commcare-core/pull/359


unit-tests/src/org/commcare/android/tests/DemoUserRestoreTest.java
```diff
@@ -0,0 +1,103 @@
...
+        CommCareHomeActivity homeActivity =
+                Robolectric.buildActivity(CommCareHomeActivity.class)
+                        .withIntent(homeActivityIntent).setup().get();
+        ShadowActivity shadowActivity = Shadows.shadowOf(homeActivity);
+        // demo users don't have a '...' menu
+        assertFalse(shadowActivity.getOptionsMenu().hasVisibleItems());
+    }
+
+    @Test
+    public void testDemoUserRestoreAndUpdate() {
```
phillipm: @amstone326 This test is now properly setup if you want to do TDD. It is failing in 2 of the main places where the 'demo user offline restore' feature needs more work: disabling home screens settings menu (not actually sure we want to disable this for the new, more functional demo user) and properly handling updates to the restore.

app/src/org/commcare/activities/AdvancedActionsActivity.java
```diff
@@ -223,9 +226,9 @@ private void startConnectionTest() {
         startActivity(i);
     }
 
-    private void clearUserData() {
+    public static void clearUserData(final Activity activity) {
```
amstone326: so that this can be called by `AppManagerAdvancedSettings` as well

amstone326:
One thing pending on this, although it doesn't need to block merge. Once [this PR](https://github.com/dimagi/commcare-hq/pull/13244) is merged, I need to make sure that submissions from the demo user stop receiving a 403 response (which they do currently).

amstone326:
Either @wspride or @ctsims should probably review this, since Phillip and I both did a portion of the dev

ctsims:
@amstone326 @phillipm I'll start looking, but it would be super helpful on these PR's to also get some sort of documentation about what spec changes were implemented (either by updating the wiki spec or by linking the spec somewhere), since otherwise it's hard for me to know where to look for edge cases

app/src/org/commcare/activities/AdvancedActionsActivity.java
phillipm: I'm curious, does this UX extension has anything to do with the offline restore feature?

app/src/org/commcare/android/resource/installers/OfflineUserRestoreAndroidInstaller.java
```diff
@@ -0,0 +1,97 @@
...
+    @Override
+    public void readExternal(DataInputStream in, PrototypeFactory pf)
+            throws IOException, DeserializationException {
+        super.readExternal(in, pf);
+    }
+
+    @Override
+    public void writeExternal(DataOutputStream out) throws IOException {
+        super.writeExternal(out);
+    }
```
phillipm: `readExternal` and `writeExternal` can be removed because they just call the `super` implementations.

app/src/org/commcare/preferences/CommCarePreferences.java
```diff
@@ -487,12 +487,7 @@ private void startDeveloperOptions() {
     }
 
     public static String getKeyServer() {
-        if (CommCareApplication._().isConsumerApp()) {
-            // So that we don't attempt to do any remote key management in a consumer app
-            return null;
-        } else {
-            return CommCareApplication._().getCurrentApp().getAppPreferences().getString("key_server", null);
-        }
```
phillipm: 👍

app/src/org/commcare/activities/AdvancedActionsActivity.java
amstone326: Yea. Since demo users don't have an options menu, there was no way to clear user data for the demo user once you logged in as them. I ran into wanting to be able to do so numerous times while doing dev, and I imagine it could be useful for others as well.

amstone326:
@ctsims This is the feature spec, although only a small portion of it is relevant to mobile: https://docs.google.com/document/d/1rpt4Tq98Lw3fl1s3x-yRraUkwAqjxBEer_n2c0TB-lE/edit#heading=h.q3wyha58n4s0. 

The major mobile change spec-wise is that app files can now contain a `demo-user-restore.xml` file, and if it is present, the suite file contains a new resource reference that looks like this:
```<user-restore>
        <resource descriptor="demo user restore"
            id="demo user restore"
            version="6">
            <location authority="local">./demo-user-restore.xml</location>
            <location authority="remote">./demo-user-restore.xml</location>
        </resource>
    </user-restore>```

Where is the best place for me to add this in our wiki pages?

File: app/src/org/commcare/activities/LoginActivity.java [hunk 5]
```diff
@@ -390,6 +397,24 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+            DemoUserBuilder.build(this, CommCareApplication._().getCurrentApp());
+            tryLocalLogin(DemoUserBuilder.DEMO_USERNAME, DemoUserBuilder.DEMO_PASSWORD, false,
+                    false, LoginMode.PASSWORD, true);
+        }
+    }
+
+    private void loginCczDemoUser(OfflineUserRestore offlineUserRestore) {
+        String username = offlineUserRestore.getUsername();
+        String password = offlineUserRestore.getPassword();
+        performingCczDemoUserLogin = true;
```
ctsims: I'm a bit worried about this boolean's status being a dependence inside of this class, as I don't think this class really uses member variables for this kind of state. Is there a specific reason that this flag isn't getting passed through the method instead?

app/src/org/commcare/android/resource/installers/OfflineUserRestoreAndroidInstaller.java
```diff
@@ -0,0 +1,97 @@
...
+
+    public OfflineUserRestoreAndroidInstaller(String localDestination, String upgradeDestination) {
+        super(localDestination, upgradeDestination);
+    }
+
+    @Override
+    public boolean initialize(AndroidCommCarePlatform instance, boolean isUpgrade) throws ResourceInitializationException {
+        if (localLocation == null) {
+            throw new ResourceInitializationException("The user restore file location is null!");
```
ctsims: This should crash have crashed before now, right? the installer shouldn't be able to save itself without this flag being set, right? I feel like it's better to crash on save than it is to crash during runtime

app/src/org/commcare/android/resource/installers/OfflineUserRestoreAndroidInstaller.java
```diff
@@ -0,0 +1,97 @@
...
+        }
+        if (isUpgrade) {
+            wipeSandboxForCurrentDemoUser(instance);
+        }
+        OfflineUserRestore offlineUserRestore = new OfflineUserRestore(localLocation);
+        instance.registerDemoUserRestore(offlineUserRestore);
+        return true;
+    }
+
+    private void wipeSandboxForCurrentDemoUser(AndroidCommCarePlatform instance) {
```
ctsims: This should be centralized somewhere other than this file, I think. Probably next to (possibly sharing code with) the other clear user data functionality?

amstone326: updated

app/src/org/commcare/android/resource/installers/OfflineUserRestoreAndroidInstaller.java
amstone326: @phillipm wrote this portion, but will just remove this check then if that seems right?

app/src/org/commcare/activities/LoginActivity.java
amstone326: Addressed by https://github.com/dimagi/commcare-android/pull/1438/commits/b592fb2a1d958e1bc0f3e603b7fea5eba41dc584

app/src/org/commcare/android/resource/installers/OfflineUserRestoreAndroidInstaller.java
amstone326: @phillipm Confirming this sounds OK to you

phillipm: 👍 
not sure why I had that check there

app/src/org/commcare/android/resource/installers/OfflineUserRestoreAndroidInstaller.java
```diff
@@ -0,0 +1,74 @@
...
+    public OfflineUserRestoreAndroidInstaller() {
+        // for externalization
+    }
+
+    public OfflineUserRestoreAndroidInstaller(String localDestination, String upgradeDestination) {
+        super(localDestination, upgradeDestination);
+    }
+
+    @Override
+    public boolean initialize(AndroidCommCarePlatform instance, boolean isUpgrade) throws ResourceInitializationException {
```
ctsims: So this is a tricky point, but I don't know that the semantics of this method work quite right for this case.

the isUpgrade flag here lets the app know to process the initialization in-place because there was an upgrade, but it is 

A) Not guaranteed to fire (IE: an upgrade can "complete" in the engine, and then an error can restart the process, and the app will initialize without firing the isUpgrade version of this method)
B) Only happens after an update is finalized and has become the current running mode

B is important because it means that even if this method was required to fire before the app could proceed, if there were an error in the demo data, the app could never get back on track, because an error in the restore data would prevent the update from continuing.

Any work that has to be guaranteed to happen has to occur in the customInstall or root install methods. or their lifecycle needs to somehow reflect its own guarantees that something was processed.

amstone326: Moved this code to `customInstall` and checked that it still works as I expect.

amstone326:
@ctsims Can you expand on what you mean by "not validating the resource in the installer anywhere"? I think that's the only unresolved issue left! 😄

app/src/org/commcare/android/resource/installers/OfflineUserRestoreAndroidInstaller.java
```diff
@@ -40,10 +40,13 @@ public OfflineUserRestoreAndroidInstaller(String localDestination, String upgrad
...
+    }
+
+    private OfflineUserRestore initDemoUserRestore() throws ResourceInitializationException {
         try {
-            OfflineUserRestore offlineUserRestore = new OfflineUserRestore(localLocation);
-            instance.registerDemoUserRestore(offlineUserRestore);
-            return true;
+            return new OfflineUserRestore(localLocation);
         } catch (UnfullfilledRequirementsException e) {
             throw new ResourceInitializationException(e.getMessage());
```
ctsims: One more very, very minor point: We should be wrapping these exceptions properly (IE: initializing the throwable source in the new exception) rather than just adding the same string message, otherwise we lose the most relevant part of the stack trace. This is a general habit we should get into with error handling

amstone326:
Hey @ctsims, so I performed both tests that we talked about yesterday, and they worked as expected:

-With an invalid user restore file, the update fails at the initialization attempt in `customInstall()`, so we never try to wipe the current sandbox, and everything remains as it was on the current version. 

-With an invalid xform that gets parsed after the restore restore file has already been installed, the only thing that is lost is that any local data that had been created under the demo user gets wiped (since in this case the sandbox has already been wiped, and then will get re-instantiated on next login). This seems fine to me since the expectation is that local demo data will also be wiped on a successful update, so if you were attempting to update, you were prepared for that to happen.

app/src/org/commcare/android/resource/installers/OfflineUserRestoreAndroidInstaller.java
```diff
@@ -0,0 +1,90 @@
...
+    public boolean initialize(AndroidCommCarePlatform instance, boolean isUpgrade) throws ResourceInitializationException {
+        instance.registerDemoUserRestore(initDemoUserRestore());
+        return true;
+    }
+
+    private OfflineUserRestore initDemoUserRestore() throws ResourceInitializationException {
+        try {
+            return new OfflineUserRestore(localLocation);
+        } catch (UnfullfilledRequirementsException e) {
+            throw new ResourceInitializationException(e.getMessage());
```
ctsims: Can we wrap these too? Good to pull afterwards

amstone326: Agh, sorry for so much back and forth. Thanks for the super detailed review, really appreciate it. Should be g2g now

phillipm:
🎆 🎉 🎉 🎆

PR #1442, opened 2016-08-10
Tell user when their attachment is too large to upload

phillipm:
| Media is large but not too large | Media is to large to upload |
| ----- | ---- |
| warning | oversized |

Implementing this fix is complicated by the fact that we don't want to store the oversized attachment if we aren't going to upload it because that would force the user to clear it, which requires a long click. 
Hence when the widget gets recreated when resuming from the callout, that state gets lost because the widget gets all its state from the associated `FormEntryPrompt`. Thus, my approach is to check the filesize in `setBinaryData`, which is called in `onActivityResult`, and set the `oversizedMediaSize` value. Then when the value is committed to the form instance in `getAnswer` we check `oversizedMediaSize` and return `InvalidData` if it is set. The next time the widget is recreated we check if the prompt is storing an `InvalidData` value and reshow the warning. At this point `oversizedMediaSize` is also cleared, so `getAnswer` returns `null` until reset.

Fix for http://manage.dimagi.com/default.asp?234483

### DIFF
--- a/app/src/org/commcare/views/widgets/MediaWidget.java
+++ b/app/src/org/commcare/views/widgets/MediaWidget.java
@@ -17,6 +17,8 @@ import org.javarosa.form.api.FormEntryPrompt;
 import java.io.File;
 
 /**
+ * Generic logic for capturing or choosing audio/video/image media
+ *
  * @author Phillip Mates (pmates@dimagi.com)
  */
 public abstract class MediaWidget extends QuestionWidget {
@@ -25,17 +27,18 @@ public abstract class MediaWidget extends QuestionWidget {
     protected Button mCaptureButton;
     protected Button mPlayButton;
     protected Button mChooseButton;
+    protected String mBinaryName;
 
     protected final PendingCalloutInterface pendingCalloutInterface;
-
-    protected String mBinaryName;
     protected final String mInstanceFolder;
+
     private int oversizedMediaSize;
 
-    public MediaWidget(Context context, FormEntryPrompt p, PendingCalloutInterface pic) {
-        super(context, p);
+    public MediaWidget(Context context, FormEntryPrompt prompt,
+                       PendingCalloutInterface pendingCalloutInterface) {
+        super(context, prompt);
 
-        this.pendingCalloutInterface = pic;
+        this.pendingCalloutInterface = pendingCalloutInterface;
 
         mInstanceFolder =
                 FormEntryActivity.mInstancePath.substring(0,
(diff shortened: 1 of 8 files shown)


app/src/org/commcare/views/widgets/QuestionWidget.java
```diff
@@ -623,10 +639,18 @@ public boolean hasListener() {
...
     }
 
-    public void checkFileSize(File file){
-        if (FileUtil.isFileOversized(file)) {
+    /**
+     * @return True if file is too big to upload.
+     */
+    protected boolean checkFileSize(File file){
+        if (FileUtil.isFileToLargeToUpload(file)) {
+            String fileSize = FileUtil.getFileSize(file) / 1024 + "";
```
amstone326: why are you dividing by 1024 again here, when `getFileSize()` already does that?

app/src/org/commcare/views/widgets/MediaWidget.java
```diff
@@ -0,0 +1,165 @@
...
+        addView(mCaptureButton);
+        addView(mChooseButton);
+        addView(mPlayButton);
+    }
+
+    @Override
+    public IAnswerData getAnswer() {
+        if (mBinaryName != null) {
+            return new StringData(mBinaryName);
+        } else if (oversizedMediaSize > 0) {
```
amstone326: Shouldn't this check happen first? Or will `mBinaryName` always be null when `oversizedMediaSize > 0`?

app/src/org/commcare/views/widgets/MediaWidget.java
```diff
@@ -0,0 +1,165 @@
...
+        String binaryPath = createFilePath(binaryURI);
+        File source = new File(binaryPath);
+        boolean isToLargeToUpload = checkFileSize(source);
+
+        if (mBinaryName != null) {
+            deleteMedia();
+        }
+
+        if (isToLargeToUpload) {
+            oversizedMediaSize = (int)source.length() / (1024 * 1024);
```
amstone326: can you make it clearer why we're dividing by `(1024 * 1024)` here?

amstone326:
@phillipm Hey, so I haven't done a super thorough review of the whole PR yet, but just from doing an overview, this seems like it may be a larger PR than is justified to be sneaking into 2.29 when QA is pretty much finished, given that this wasn't a regression in 2.29 (at least that's my understanding, correct me if I'm wrong). What would your thoughts be on pushing this to 2.30? Or if you feel strongly about getting it in, at least shrinking the PR to only the essential commits, rather than including a whole refactor of essential multimedia classes?

phillipm:
retargeting for 2.30 in https://github.com/dimagi/commcare-android/pull/1446

PR #1447, opened 2016-08-11
Add warning if file to big to upload

phillipm:
| Media is large but not too large | Media is to large to upload |
| ----- | ---- |
| warning | screen |

partial fix to http://manage.dimagi.com/default.asp?234483
More comprehensive fix that actually removes the video from the question at https://github.com/dimagi/commcare-android/pull/1446 and targeting 2.30


app/src/org/commcare/utils/FileUtil.java
```diff
@@ -536,6 +536,10 @@ public static boolean isFileOversized(File mf) {
         return length > WARNING_SIZE;
     }
 
+    public static boolean isFileToBigToUpload(File mf) {
```
amstone326: To --> Too :-)

app/src/org/commcare/views/widgets/QuestionWidget.java
```diff
@@ -624,7 +625,10 @@ public boolean hasListener() {
     }
 
     public void checkFileSize(File file){
-        if (FileUtil.isFileOversized(file)) {
+        if (FileUtil.isFileToBigToUpload(file)) {
+            long overByAmount = (file.length() - FormUploadUtil.MAX_BYTES) / 1024;
```
amstone326: Do we think that `overByAmount` will actually be a useful/actionable piece of information for users? As opposed to just saying it's too big to upload without specifying by how much?

phillipm: I figured it would be useful to see some sort of feedback about how much over the attachment is. Open to taking it out or showing a more comprehensive message like in https://github.com/dimagi/commcare-android/pull/1446

ctsims:
@phillipm Is it possible to add rather than change this message? We actually added the 5mb intentionally because HQ has a hard time receiving posts above that size, so splitting into two messages: "5mb is likely too big" and "15mb is defintiely too big" would be helpful

phillipm:
@ctsims I left the old message intact; I've updated the screenshots to reflect that.

ctsims:
@phillipm Oh, sorry, I totally missed that, sorry!

PR #1452, opened 2016-08-12
Update sqlcipher

phillipm:
Seems like a good idea to update sqlcipher to 3.5.3 given the [release notes](https://discuss.zetetic.net/t/sqlcipher-for-android-3-5-3-release/1549):

> SQLCipher for Android 3.5.3 was just released, it includes a fix that addresses over reads from the cursor window. The over read would occur in the scenarios addressed in the 3.5.2 release where calls such as getInt(…) and getDouble(…) for columns that have a text affinity of TEXT did not encode the content properly to the Java client. We have added further tests to the SQLCipher for Android test suite3 to prevent any regressions in the future. We recommend that you upgrade to this release immediately.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -62,7 +62,7 @@ dependencies {
     compile 'ch.acra:acra:4.9.0@aar'
     compile 'joda-time:joda-time:2.9.4'
     compile 'net.sf.kxml:kxml2:2.3.0'
-    compile 'net.zetetic:android-database-sqlcipher:3.5.2@aar'
+    compile 'net.zetetic:android-database-sqlcipher:3.5.3@aar'
     compile 'org.apache.james:apache-mime4j:0.7.2'
     compile('org.apache.httpcomponents:httpmime:4.3.6') {
         exclude module: 'httpclient'


amstone326:
@phillipm Oof, definitely agreed that the release note sounds like we should update now, but given that our last P1 was due to a library upgrade that had unforeseen consequences, should we be concerned about doing this post-QA?

phillipm:
It looks like the difference between `v3.5.2` and `v.3.5.3` is 3 commits:
 - https://github.com/sqlcipher/android-database-sqlcipher/commit/7edd89142517908dffa5d19642901b83f586cf15?w=1
 - https://github.com/sqlcipher/android-database-sqlcipher/commit/da1cc4e366f59dac251db2bf09892618f82568a8
 - https://github.com/sqlcipher/android-database-sqlcipher/commit/a7760e9bd108572a1590f888335408800cf8e8e8

The last commit makes changes to the body of `openDatabase` and `getWritableDatabase`, which we call, but we should be fine because it doesn't change the method signature, which is what messed up the form attachments.

phillipm:
Merging this before QA resumes tomorrow

PR #1456, opened 2016-08-18
Allow more than 2 apps by default in debug builds

phillipm:
Not sure if y'all feel like this is a good idea or not, but I think it would be useful to remove the cap of 2 apps installed for debug builds of CommCare. Throughout my dev workflow I often reinstall CommCare and then want to use multiple apps, which requires me to reclaim super user multiple apps privileges every time.


amstone326:
👍 I'm good with this

PR #1457, opened 2016-08-18
Add list bounds check in QuestionsView.setFocus

phillipm:
Fix for an ACRA crash:
```
java.lang.IndexOutOfBoundsException: Invalid index 1, size is 1
at java.util.ArrayList.throwIndexOutOfBoundsException(ArrayList.java:255)
at java.util.ArrayList.get(ArrayList.java:308)
at org.commcare.views.QuestionsView.setFocus(QuestionsView.java:329)
at org.commcare.activities.FormEntryActivity.showView(FormEntryActivity.java:1182)
at org.commcare.activities.FormEntryActivity.refreshCurrentView(FormEntryActivity.java:736)
at org.commcare.activities.FormEntryActivity.onActivityResult(FormEntryActiv
```

@amstone326 I'm just adding a patch to prevent the crash. Not sure if it is worth digging into how this may come up. Maybe something with questions being removed after relevancy condition recalculations?

### DIFF
--- a/app/src/org/commcare/views/QuestionsView.java
+++ b/app/src/org/commcare/views/QuestionsView.java
@@ -325,7 +325,7 @@ public class QuestionsView extends ScrollView
 
     public void setFocus(Context context, int indexOfLastChangedWidget) {
         QuestionWidget widgetToFocus = null;
-        if (indexOfLastChangedWidget != -1) {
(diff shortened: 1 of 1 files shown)


amstone326:
Hey yea that sounds very likely to be the cause to me. There is a world in which we could try to keep a count of how many widgets get removed due to relevancy condition changes and factor that in here, but I don't think that's worth doing right now, so going to merge this.

PR #1459, opened 2016-08-18
Add flag to respect exact image bounds

amstone326:
Add a flag that can indicate to us to scale an image down to more precise dimensions based on the given container dimensions, and set that flag to true in places where we are inflating a display image that should be shown within very specific dimensions. 

Fixes http://manage.dimagi.com/default.asp?235370

### DIFF
--- a/app/src/org/commcare/adapters/MenuAdapter.java
+++ b/app/src/org/commcare/adapters/MenuAdapter.java
@@ -270,9 +270,10 @@ public class MenuAdapter implements ListAdapter {
     }
 
     public void setupImageView(ImageView mIconView, MenuDisplayable menuDisplayable) {
-        String imageURI = menuDisplayable.getImageURI();
-        Bitmap image = MediaUtil.inflateDisplayImage(context, imageURI);
         if (mIconView != null) {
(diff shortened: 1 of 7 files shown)


amstone326:
Screenshot of the action item images that were messed up after the fix:
screenshot_20160818-163454

app/src/org/commcare/adapters/MenuAdapter.java
```diff
@@ -270,9 +270,10 @@ public void setupTextView(TextView textView, MenuDisplayable menuDisplayable) {
     }
 
     public void setupImageView(ImageView mIconView, MenuDisplayable menuDisplayable) {
-        String imageURI = menuDisplayable.getImageURI();
-        Bitmap image = MediaUtil.inflateDisplayImage(context, imageURI);
         if (mIconView != null) {
+            int iconDimension = (int)context.getResources().getDimension(R.dimen.menu_icon_size);
+            Bitmap image = MediaUtil.inflateDisplayImage(context, menuDisplayable.getImageURI(),
+                    iconDimension, iconDimension);
```
amstone326: Not actually using the new flag here because `mIconView` imposes the precise dimensions anyway, but realized that the MenuAdapter code should at least be using `iconDimension` as a rough bounding dimension

PR #1461, opened 2016-08-19
Don't start app update if there isn't connectivity

phillipm:


ctsims:
@phillipm Also opened https://github.com/dimagi/commcare-android/pull/1466 to unify how we are checking to match android guidelines

PR #1462, opened 2016-08-19
Hide keyboard on login attempt

phillipm:
This change ensures keyboard is dismissed on demo user login, which didn't used to happen.


ctsims:
@phillipm Does this happen even if the login fails? It's actually very inconvenient to have to pop the keyboard back up to re-type the username/password if there is a problem, and our users are very bad at understanding how to summon the keyboard, so I worry a lot that this will cause problems.

phillipm:
@ctsims In checking the existing code, the keyboard was always dismissed. I've updated it to only dismiss on successful logins, which does prevent some of the warning messages from being fully seen. I guess that seems like an alright trade-off for not having to re-click the text box, right?

amstone326:
@phillipm I don't have full context on the reason for this PR, but I don't think I agree with that. Having to re-click the text box is an annoyance, but it is something that users will know to do when it's necessary. Conversely I doubt most users will think to hide the keyboard in order to check if there is a message there, which means when a login fails, they won't have any visual indication of the fact that it failed or of the cause.

phillipm:
@amstone326 agreed. Updated to the PR to keep the old behavior but also dismiss the keyboard on demo logins

PR #1472, opened 2016-08-26
Blocking actions infrastructure

amstone326:
Fix for http://manage.dimagi.com/default.asp?235475

Very simple infrastructure for keeping track of events that should block other things until they have completed. Current usage is to block any navigation in form entry while form relevancy updates triggered by `onDateChanged` are still being processed.

@ctsims If you want to point me to the place in the TDH app where you said that text entry could benefit from this, I can look into adding that to this as well (or just do it in a separate PR).


app/src/org/commcare/activities/FormEntryActivityUIController.java
```diff
@@ -562,6 +573,10 @@ protected void next() {
         }
     }
 
+    protected boolean shouldIgnoreNavigationAction() {
```
ctsims: I was a bit surprised to see this get created v. added to shouldIgnoreSwipeAction, seems like they are accomplishing the same thing, right? Shouldn't this just get added to the list of reasons a "nav action" gets ignored?

app/src/org/commcare/activities/FormEntryActivityUIController.java
```diff
@@ -372,6 +379,10 @@ protected void showNextView() {
     }
 
     private void showNextView(boolean resuming) {
+        if (shouldIgnoreNavigationAction()) {
```
ctsims: Below we block the navigation on "next" instead of "showNextView", is there a reason it makes sense to handle this processing in two different places in the chain, exactly?

app/src/org/commcare/activities/FormEntryActivityUIController.java
amstone326: No, they're not. `shouldIgnoreSwipeAction` is used in the way its name indicates; it only acts to prevent *swipe* actions. I don't know why this is and chose not to mess with it

app/src/org/commcare/views/QuestionsView.java
```diff
@@ -65,12 +66,14 @@
...
 
+    private BlockingActionsManager blockingActionsManager;
+
     /**
      * If enabled, we use dividers between question prompts
      */
     private static final boolean SEPERATORS_ENABLED = false;
 
-    public QuestionsView(Context context) {
+    public QuestionsView(Context context, BlockingActionsManager blockingActionsManager) {
```
ctsims: Is there a reason we pass the blocking actions manager down the chain directly v. adding it to the widget factory? It seems like the widget factory is a cleaner place to keep the refrence

app/src/org/commcare/activities/FormEntryActivityUIController.java
amstone326: Similar to my answer above. `next()` is only used in a small subset of navigation scenarios. It is only called by `onForwardSwipe` and `advance` (which is used for widgets that auto-advance)

app/src/org/commcare/views/QuestionsView.java
amstone326: Sorry, not following. The `BlockingActionsManager` has to originate in the `FormEntryActivityUIController` because that is where we check it. So I just followed the natural path that got me from `FormEntryActivityUIController` --> `WidgetFactory.createWidgetFromPrompt`, which seems to require going through `QuestionsView`, since that is the link between them

app/src/org/commcare/views/widgets/DateWidget.java
```diff
@@ -66,6 +69,8 @@ private DatePicker buildDatePicker(boolean isQuestionWriteable) {
             public void run() {
                 widgetEntryChanged();
                 isRelevancyUpdateScheduled = false;
+                blockingActionsManager.registerActionCompletion(
```
ctsims: In general with these kinds of callback-centric workflows it's pretty dangerous for the way that the callback to be completed to rely on the callback being handled in this kind of manner.

ctsims: So this structure can cause problems in a couple of ways

app/src/org/commcare/activities/FormEntryActivityUIController.java
```diff
@@ -372,6 +380,11 @@ protected void showNextView() {
...
+            isAnimatingSwipe = false;
```
ctsims: I had to add this to fix a bug next() and showNextView() could accidentally end up leaving the animating flag to true.

app/src/org/commcare/views/QuestionsView.java
```diff
@@ -139,7 +143,7 @@ public QuestionsView(Context context, FormEntryPrompt[] questionPrompts,
             widgets.add(qw);
             mView.addView(qw, mLayout);
             
-            qw.setChangedListener(this);
+            qw.setChangedListeners(this, blockingActionsManager);
```
ctsims: Moved this to the question widget centrally as part of the listener setup (which is what connects it "to the screen" more or less, so should be the same path), lets us access it broadly from each widget.

app/src/org/commcare/views/widgets/QuestionWidget.java
```diff
@@ -601,13 +604,31 @@ public FormIndex getFormId(){
...
     }
 
+    protected void fireDelayed(DelayedBlockingAction delayedBlockingAction) {
+        if(this.blockingActionsManager != null) {
+            blockingActionsManager.queue(delayedBlockingAction);
+        }
+    }
+
+    protected void widgetEntryChangedDelayed() {
+        fireDelayed(new DelayedBlockingAction(System.identityHashCode(this), 400) {
```
ctsims: I moved this to 400 after experimenting a bit. It seems like a good fit now that the events chain.

ctsims:
@amstone326 Rewrote a bit to reflect a few things

1) The events can now be targeted per widget rather than per "event type". Otherwise if there are two date entries on a screen a change to both in succession will prevent one's delayed event from firing

2) I moved the state logic for things to the delayed event itself, this helped (3), and also made it the case that it the state receiver was no longer connected we didn't miss the event close

3) I set it up so that if I fire a bunch of events over and over instead of the new events not happening (original behavior from phillip) the new events expire the old event. This results in a situation where a new change before the old event fired "Takes it's place". this makes it the case that if I scroll around back and forth a lot the original scroll won't eventually fire in the middle of my scrolling, but it's perceived as one long event.

4) Restructured the listener to end up in the QuestionWidget root, which should make it easier for new widgets to use

I think it's best if you do a pass on the new code and pull if you think it looks good.

app/src/org/commcare/views/widgets/QuestionWidget.java
amstone326: What is the difference between using `System.identityHashCode(this)` vs. just `this`?

ctsims: Hm, a reasonable point. Mostly just dislike keeping around raw object references when we don't have to (to prevent accidental memory leaks, etc). but does seem a bit convoluted.

amstone326:
@ctsims Thanks for the explanation.. would not have come up with that on my own, but good to see an example of it. I made some linting-type changes but nothing else, so going to merge

PR #1474, opened 2016-08-29
Allow configuration of GPS auto-capture accuracy via HQ

phillipm:
Allows one to configure the GPS auto-capture accuracy cutoff by setting the `cc-gps-auto-capture-accuracy` property (to a double) using the ['custom properties' HQ feature flag](https://confluence.dimagi.com/display/ccinternal/CommCare+Android+Developer+Options). If not set, defaults to 10.0 meters.

@dimagi/product Might be nice to consider making this more widely available for configuration on HQ. Like exposing a value box when auto-capture is enabled on HQ that lets you set the accuracy. This configuration is helpful because the desired GPS accuracy varies from project to project and if misconfigured has negative effects on device battery life.


amsagoff:
Nice! Added to HQ roadmap

PR #1476, opened 2016-08-31
Enable offline updates from a .ccz

amstone326:


app/src/org/commcare/preferences/DeveloperPreferences.java
```diff
@@ -315,4 +317,8 @@ public static String formLoadPayloadStatus() {
     public static boolean isDetailTabSwipeActionEnabled() {
         return doesPropertyMatch(DETAIL_TAB_SWIPE_ACTION_ENABLED, CommCarePreferences.YES, CommCarePreferences.YES);
     }
+
+    public static boolean isOfflineUpdateEnabled() {
+        return doesPropertyMatch(OFFLINE_UPDATE_ENABLED, CommCarePreferences.YES, CommCarePreferences.YES);
```
amstone326: I currently have the default value set as Enabled, because I find it to be a pain to enable developer options, and because this is well-hidden as it is (in the options menu of the `UpdateActivity`), but would be open to switching it if that sounds wrong

phillipm:
Looks good!
I don't have opinion on whether this is enabled by default or not.
Have you tried running this on invalid CCZs? Like a version of a CCZ that is earlier than your current version or a completely different app?

amstone326:
@phillipm Yep, in both cases it just shows a message that says "App is up to date"

app/src/org/commcare/preferences/DeveloperPreferences.java
ctsims: Is there a reason to flag this at all if we are going to enable it by default? It's hard for me to imagine someone getting a ton of value from disabling the flag.

amstone326: My thinking was that projects may want to disable it via the custom flag thing on HQ. Does that sound improbable?

ctsims: What's the expected reason it might be disabled? It's actually a pretty sketchy attack vector in its current form, because someone could use the app manager to get to the update screen, then "replace" someone's app install with a malicious version (moving the submit/sync server to their Man in the Middle passthrough) without needing to be logged in, but I think if we're worried about that we shouldn't turn it on by default.

amstone326: Gotcha. @phillipm proposed making the offline install option only available when you are doing an update from within a user session, and never accessible from the app manager (with the exception of debug builds perhaps? just to make it easier for ourselves). Does that sound reasonable?

PR #1477, opened 2016-08-31
Change accessibility to offline update

amstone326:
@ctsims Proposed solution for offline update security concerns - only show the offline update option if we are logged in, or in debug mode


ctsims:
Hey, thanks for the update, I think this makes sense for now. In the future we'll want to be authenticating install hashes anyway for good security, but this should be totally adequate for our current risks.

PR #1479, opened 2016-09-05
Show 406 error messages to user

phillipm:
Mobile side of https://github.com/dimagi/commcare-hq/pull/12728
Show error message passed down from HQ on `406` http responses.

This will allow us to send custom messages down to our users.

screen

Thanks for the HQ side of this @mkangia 🎉


mkangia:
@phillipm Nice! Thats better information. Next would be a message for inactive users as well?

PR #1481, opened 2016-09-06
Move 'Enable Updating to Un-Starred Builds' to normal preferences

phillipm:
@dimagi/product How do you feel about the text of this feature?  <strike>I feel like 'Use latest HQ build' might be better than 'Use Newest App Version from HQ'. </strike> Went with @amstone326 suggestion of 'Enable Updating to Un-Starred Buildes'.

screen

### DIFF
--- a/app/res/xml/commcare_preferences.xml
+++ b/app/res/xml/commcare_preferences.xml
@@ -33,7 +33,7 @@
         android:entries="@array/pref_enabled_labels"
         android:entryValues="@array/pref_enabled_vals"
         android:key="cc-newest-version-from-hq"
-        android:title="Use Newest App Version From HQ"/>
+        android:title="Enable Updating to Un-Starred Builds"/>
     <Preference
         android:key="disable-analytics"/>
     <Preference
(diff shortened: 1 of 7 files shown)


amstone326:
What about "Enable Updating to Un-Starred Builds"? I feel like the majority of users will benefit from explicitly stating that that is the functionality change

amstone326:
@phillipm Hey, a little late to be thinking of this, but just randomly occurred to me- It feels a little dangerous that this setting is now so accessible in the app, because it could be pretty bad if users accidentally enabled it while in the field. I think it makes sense to at least move it from the 'Settings' menu to the 'Advanced' menu, and maybe add a warning of some sort when you try to change it? What do you think?

amstone326:
@amsagoff @ctsims Can you let me know if you think the above concern is legit? Phillip raised the concern that to move it again immediately after moving it once would be frustrating for people using it, which I agree with. But maybe we could just make it harder to edit, by having some sort of warning dialog that requires confirming before you get to the edit view?

amsagoff:
I don't have strong feelings on settings vs advanced settings, but in
general if there's a tradeoff between confusing existing users and putting
something in the place we think is best long-term, we should do the latter.
Most users are internal right now, so that goes double for this feature.

amstone326:
@amsagoff Makes sense. Are you not that concerned about the possibility of it being accidentally changed by users who shouldn't though? You would know best whether that sort of thing is worth worrying about

amsagoff:
I'm not super concerned, but think that it's valid to be. One thing we
could do is check out analytics for how many clicks there are on items in
"Settings" vs. "Advanced" and see if the click rate on random items is
significantly different.

PR #1483, opened 2016-09-07
Enable session saving when running 'ccc update'

phillipm:
The `ccc update` command is one way to trigger Aliza's 'Update to latest build and restore current location' feature. When called this way it only works if you have already set the 'enable session saving' dev preference. 

Instead of not working when it isn't set, this PR changes the behavior to automatically enable it, saving some setup/keystrokes.


amstone326:
Was just getting annoyed by this yesterday- thanks

PR #1485, opened 2016-09-07
Purge form records that update a non-existant case

wpride:
fixes http://manage.dimagi.com/default.asp?236251

### DIFF
--- a/app/src/org/commcare/tasks/ProcessAndSendTask.java
+++ b/app/src/org/commcare/tasks/ProcessAndSendTask.java
@@ -15,6 +15,7 @@ import org.commcare.views.notifications.NotificationMessageFactory;
 import org.commcare.views.notifications.ProcessIssues;
 import org.javarosa.core.model.User;
 import org.javarosa.core.services.Logger;
+import org.javarosa.xml.util.InvalidStorageStructureException;
 import org.javarosa.xml.util.InvalidStructureException;
 import org.javarosa.xml.util.UnfullfilledRequirementsException;
 import org.xmlpull.v1.XmlPullParserException;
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/tasks/ProcessAndSendTask.java
```diff
@@ -194,6 +195,13 @@ private boolean checkFormRecordStatus(FormRecord[] records)
                         //Otherwise, the SD card just got removed, and we need to bail anyway.
                         throw e;
                     }
+                } catch(InvalidStorageStructureException e) {
+                    // Thrown when updating a case that isn't present
+                    // Do same behavior as InvalidStructureException (wipe record)
+                    CommCareApplication._().reportNotificationMessage(NotificationMessageFactory.message(ProcessIssues.BadTransactions), true);
+                    Logger.log(AndroidLogger.TYPE_ERROR_DESIGN, "Removing form record due to case transaction data|" + getExceptionText(e));
+                    FormRecordCleanupTask.wipeRecord(c, record);
```
phillipm: What if we quarantined these forms (`processor.updateRecordStatus(record, FormRecord.STATUS_LIMBO);`)? If a user reports a form not making it to HQ, it might be helpful to be able to see which ones on the device so they can copy over data to a new form. Also possible that this is a sync related bug and a 412 sync would allow them to recover the form to a state that they can successfully submit it.

phillipm:
Didn't realize that this is reverting back to old expected behavior.
I will follow up with a test that saves a form that edits a case as incomplete, complete another form that closes that case, and then completes the incomplete form, asserting that the form is wiped and the user is warned.

PR #1489, opened 2016-09-12
Allow detail to tell entity list to focus to bottom

amstone326:
Custom for UATBC adherence calendar, so that when the calendar loads they are automatically brought to today's date, rather than having to scroll all the way down to get there.

cross-request: https://github.com/dimagi/commcare-core/pull/396


phillipm:
@damagatchi test this

phillipm:
@damagatchi test this please

PR #1491, opened 2016-09-12
Use detail's <focus> element to set focus in entity list

amstone326:
cross-request: https://github.com/dimagi/commcare-core/pull/397

@ctsims This is currently targeting 2.30, but if it seems like a little bit much to add last minute, we can leave the existing implementation for now, and put this into master. Just let me know.


amstone326:
@damagatchi test this please

amstone326:
Hm, realized there is definitely an advantage to putting this in 2.30, because otherwise UATBC will need to get an app update out with the case tile changed to use the new strategy precisely when they update to 2.31. That sounds like it could be painful, unless we are already expecting to have to control that process for other reasons?

PR #1496, opened 2016-09-19
Enable showing actions in the app bar

amstone326:
Currently, any actions that are defined within a `<detail>` block are shown as an item at the end of the entity list, and as an item in the options menu of the entity select screen. This PR makes it so that:

-If the entity list is being shown in grid view, actions are NOT shown as items at the end of the list, since that results in a very poor UI
-To compensate for that, actions are now configurable such that they can show up as an icon in the app bar. If the `<action>` block is defined with an `action-bar-icon` attribute, the specified icon will show up in the app bar of the entity select screen, in place of there being an item hidden in the options menu.

Screenshot below with an example of how this looks.

The last 2 commits also fix a couple of bugs that I found while working on this.

cross-request: https://github.com/dimagi/commcare-core/pull/411

### DIFF
--- a/app/src/org/commcare/activities/EntitySelectActivity.java
+++ b/app/src/org/commcare/activities/EntitySelectActivity.java
@@ -112,8 +112,8 @@ public class EntitySelectActivity extends SaveSessionCommCareActivity
     private EntityListAdapter adapter;
     private LinearLayout header;
     private SearchView searchView;
-    private MenuItem searchItem;
-    private MenuItem barcodeItem;
+    private MenuItem searchMenuItem;
+    private MenuItem barcodeMenuItem;
 
     private EntityDatum selectDatum;
 
@@ -140,7 +140,7 @@ public class EntitySelectActivity extends SaveSessionCommCareActivity
     private Detail shortSelect;
 
     private DataSetObserver mListStateObserver;
-    private OnClickListener barcodeScanOnClickListener;
+    public OnClickListener barcodeScanOnClickListener;
     private boolean isCalloutAutoLaunching;
 
     private boolean resuming = false;
@@ -756,9 +756,9 @@ public class EntitySelectActivity extends SaveSessionCommCareActivity
             @TargetApi(Build.VERSION_CODES.HONEYCOMB)
             @Override
             public void onActionBarFound(MenuItem searchItem, SearchView searchView, MenuItem barcodeItem) {
-                EntitySelectActivity.this.searchItem = searchItem;
+                EntitySelectActivity.this.searchMenuItem = searchItem;
                 EntitySelectActivity.this.searchView = searchView;
-                EntitySelectActivity.this.barcodeItem = barcodeItem;
+                EntitySelectActivity.this.barcodeMenuItem = barcodeItem;
                 // restore last query string in the searchView if there is one
                 if (lastQueryString != null && lastQueryString.length() > 0) {
                     if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.ICE_CREAM_SANDWICH) {
@@ -803,7 +803,8 @@ public class EntitySelectActivity extends SaveSessionCommCareActivity
                 }
             }
             if (shortSelect.getCallout() != null && shortSelect.getCallout().getImage() != null) {
-                EntitySelectCalloutSetup.setupImageLayout(this, barcodeItem, shortSelect.getCallout().getImage());
+                // Replace the barcode scan callout with our custom callout
+                EntitySelectCalloutSetup.setupImageLayout(this, barcodeMenuItem, shortSelect.getCallout().getImage());
             }
         }
     }
@@ -828,7 +829,7 @@ public class EntitySelectActivity extends SaveSessionCommCareActivity
     private void setSearchText(CharSequence text) {
         if (isUsingActionBar()) {
             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.ICE_CREAM_SANDWICH) {
-                searchItem.expandActionView();
+                searchMenuItem.expandActionView();
             }
             searchView.setQuery(text, false);
         } else {
--- a/app/src/org/commcare/activities/EntitySelectActivity.java
+++ b/app/src/org/commcare/activities/EntitySelectActivity.java
@@ -159,7 +159,9 @@ public class EntitySelectActivity extends SaveSessionCommCareActivity
     private static final HereFunctionHandler hereFunctionHandler = new HereFunctionHandler();
     private boolean containsHereFunction = false;
     private boolean locationChangedWhileLoading = false;
-    private boolean hideActions;
+
+    private boolean hideActionsFromOptionsMenu;
+    private boolean hideActionsFromEntityList;
 
     // Handler for displaying alert dialog when no location providers are found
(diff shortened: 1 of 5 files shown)


amstone326:
For example, in the adherence calendar for UATBC, if I add an action with an action-bar-icon to the detail definition, it does not show up in the calendar view, and the +/- icon shows up in the app bar.
screenshot_20160921-100643

app/src/org/commcare/views/ViewUtil.java
```diff
@@ -15,27 +15,30 @@
...
+    public static void addActionToMenu(Context context, Action action, Menu menu, int menuId,
+                                       int menuGroupId) {
+        DisplayData display = action.getDisplay().evaluate();
         MenuItem item = menu.add(menuGroupId, menuId, menuId,
                 Localizer.clearArguments(display.getName()).trim());
-        if (display.getImageURI() != null) {
+        if (action.hasActionBarIcon() && Build.VERSION.SDK_INT >= Build.VERSION_CODES.HONEYCOMB) {
+            Bitmap b = MediaUtil.inflateDisplayImage(context, action.getActionBarIconReference());
+            item.setIcon(new BitmapDrawable(context.getResources(), b));
+            item.setShowAsAction(MenuItem.SHOW_AS_ACTION_ALWAYS);
```
amstone326: I have slight reservations on the use of `SHOW_AS_ACTION_ALWAYS` here. The alternative is `SHOW_AS_ACTION_IF_ROOM`, the difference between the 2 being what it sounds like. I wanted to use the ALWAYS field here so that these actions would take precedence in the app bar over the search and barcode scan icons in the case of limited room (which I think they should, since the app config is saying it's important enough to its workflow to be obvious to the user). Note that any icons that get pushed out of the app bar still show up in the options menu, they are just not immediately visible. The downside of this is that if a user were to add several actions with an `action-bar-icon` in the same detail, they would end up looking very cluttered. Since the probability of that is low, and the resolution path would be pretty obvious to app builders (including fewer actions in the app bar), I think this is OK, but just wanted to flag it. This is an example of what it would look like for the custom actions to push the search icon off of the app bar and into the options menu:
screenshot_20160921-100606

app/src/org/commcare/activities/EntitySelectActivity.java
```diff
@@ -308,64 +339,39 @@ private void restoreExistingSelection(boolean isOrientationChange) {
         }
     }
 
-    private void setupToolbar(AdapterView view) {
```
amstone326: After staring at this code for a little while, I finally realized that this was only relevant for pre-honeycomb devices, because on any device where the call to `tryToAddSearchActionToAppBar` in `onCreateOptionsMenu` is successful, the toolbar created by this method gets hidden. I wanted to make this clearer in the code, so renamed this to `setupPreHoneycombFooter`, and moved all of the code that wasn't strictly relevant to that to other places (mainly `setupUI`)

amstone326: I also confirmed on a pre-honeycomb device that the footer still shows up properly

app/src/org/commcare/activities/EntitySelectActivity.java
```diff
@@ -716,7 +728,7 @@ private void returnWithResult(Intent intent) {
     public void afterTextChanged(Editable incomingEditable) {
         final String incomingString = incomingEditable.toString();
         final String currentSearchText = getSearchText().toString();
-        if (!"".equals(currentSearchText) && incomingString.equals(currentSearchText)) {
```
amstone326: this made it so that when you changed the input from 1 character to none, that change would not register in any way

app/src/org/commcare/activities/EntitySelectActivity.java
```diff
@@ -210,21 +217,15 @@ public void onChanged() {
     }
 
     private void setSearchBannerState() {
-        if (!"".equals(adapter.getSearchQuery())) {
+        if (!"".equals(adapter.getSearchQuery()) || adapter.isFilteringByCalloutResult()) {
```
amstone326: Not sure why this was set up so that the clear search button would not appear in the results banner for normal searches, but I can't see any reason why it should be that way. On newer devices, this isn't a big deal, because Android's SearchView has its own clear button. But on older devices that use our custom search view footer, the clear button on the search banner is the only place where it is possible to clear your search.

unit-tests/src/org/commcare/android/tests/caselist/EntityListCalloutDataTest.java
```diff
@@ -27,6 +29,7 @@
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
+import org.robolectric.Robolectric;
```
phillipm: Newly added imports are unneeded

unit-tests/src/org/commcare/android/tests/caselist/EntityListCalloutDataTest.java
```diff
@@ -112,9 +115,7 @@ public void testCalloutResultWithNoColumnTest() {
 
     private void performFingerprintCallout() {
         // make entity list callout to 'fingerprint identification'
-        ImageButton calloutButton =
-                (ImageButton)entitySelectActivity.findViewById(R.id.barcodeButton);
-        calloutButton.performClick();
+        entitySelectActivity.barcodeScanOnClickListener.onClick(null);
```
phillipm: Is there a particular reason for this change? Would be nice to leave `barcodeScanOnClickListener` as `private`

app/src/org/commcare/activities/EntitySelectActivity.java
amstone326: Oh interesting. So does that mean that there has never been a way to clear your search on older devices?

unit-tests/src/org/commcare/android/tests/caselist/EntityListCalloutDataTest.java
amstone326: Yes. This test was previously using the version of the callout button that is only visible on old devices. By changing the `setupPreHoneycombFooter()` method to only init those views when they are actually being used, I rendered the R.id.barcodeButton unusable in this test, since it is never initialized. The new implementation is a version-agnostic way to test this, since both UI implementations use the same onClickListener

app/src/org/commcare/activities/EntitySelectActivity.java
```diff
@@ -267,13 +268,49 @@ private void setupUI(boolean isOrientationChange) {
...
             EntitySelectViewSetup.setupDivider(this, listView, shortSelect.usesEntityTileView());
         }
-
         visibleView.setOnItemClickListener(this);
 
-        setupToolbar(visibleView);
+        initUIComponents();
+        restoreLastQueryString();
+        persistAdapterState(visibleView);
+        setupPreHoneycombFooter(attemptInitCallout());
```
amstone326: fixed

app/src/org/commcare/activities/EntitySelectActivity.java
amstone326: Hmm ok, in that case I'll revert this change. Doesn't make sense to add that functionality for older devices after all this time if it's never been there.

PR #1497, opened 2016-09-19
Make auto GPS capture timeout configurable

amstone326:
@dimagi/product Would make sense to add this in the same place that the configurability for auto-capture accuracy is going to be placed (as per Phillip's comment [here](https://github.com/dimagi/commcare-android/pull/1474))

### DIFF
--- a/app/src/org/commcare/preferences/CommCarePreferences.java
+++ b/app/src/org/commcare/preferences/CommCarePreferences.java
@@ -35,6 +35,7 @@ import org.javarosa.core.util.NoLocalizedTextException;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 
 public class CommCarePreferences
         extends SessionAwarePreferenceActivity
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/preferences/CommCarePreferences.java
```diff
@@ -420,6 +420,16 @@ public static double getGpsCaptureAccuracy() {
         }
     }
 
+    public static int getGpsAutoCaptureTimeout() {
+        SharedPreferences properties = CommCareApplication._().getCurrentApp().getAppPreferences();
+        try {
+            return Integer.parseInt(properties.getString(GPS_AUTO_CAPTURE_TIMEOUT,
```
phillipm: Do we want wrap this value in `TimeUnit.MINUTES.toMillis`, or have users input time in milliseconds?

sheelio: I think seconds or minutes (either) makes sense for this.

amstone326: Ah good call, thanks Phillip. I'll change it to assume that the user is inputing in minutes

PR #1498, opened 2016-09-20
Update to Android Studio 2.2

phillipm:
Android Studio 2.2 was released and is full of lots of good stuff.
Not sure if this will require us all to be on Android Studio 2.2 or not.

I think this will require you to install set Java 8 as your default version of java

Update to Android 7: https://developer.android.com/about/versions/nougat/android-7.0.html
Update Robolectric to use Android SDK 23
Update Zxing: https://github.com/zxing/zxing/releases/tag/zxing-3.3.0
Update SQLCipher: https://discuss.zetetic.net/t/sqlcipher-for-android-3-5-4-release/1676
Update play services: https://developers.google.com/android/guides/releases
Update google support library: https://developer.android.com/topic/libraries/support-library/revisions.html

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,8 +7,8 @@ import org.apache.tools.ant.filters.ReplaceTokens
 
 // This build script assumes the following directory structure:
 // - somewhere/your/code/directory/is
-// -- commcare-odk (github: https://github.com/dimagi/commcare-odk/)
-// -- commcare (github: https://github.com/dimagi/commcare/)
+// -- commcare-android (github: https://github.com/dimagi/commcare-android/)
+// -- commcare-core (github: https://github.com/dimagi/commcare-core/)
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -360,12 +359,6 @@ task switchGoogleServicesCommcare(type: Copy) {
     description = 'Makes build use normal google-services.json file'
     println project.ext.COMMCARE_APP_ID
 
-    /**
-     * This is an awkward workaround for a Windows build error that occurs when you try
-     * to copy something into the root directory (containing the .gradle folder). Don't
-     * change it without testing on a windows machine
-     */
-    outputs.files.setFrom(file("$projectDir/google-services.json"))
```
phillipm: @ctsims I removed this because it was causing issues. Can you check to see if Gradle 3.1 fixed the issue for you on Windows?

PR #1500, opened 2016-09-20
Demo user failing test

amstone326:
Adds a robolectric test for updating an app that has a demo user restore file.


unit-tests/src/org/commcare/android/util/CaseLoadUtils.java
```diff
@@ -44,7 +42,7 @@ public static EntitySelectActivity launchEntitySelectActivity(String command) {
 
         // start the entity select activity
         return Robolectric.buildActivity(EntitySelectActivity.class)
-                .withIntent(entitySelectIntent).create().start().resume().get();
+                .withIntent(entitySelectIntent).setup().get();
```
phillipm: @amstone326 the culprit, as far as I can tell. Definitely took a bit to track down, but I guess it is important to call `visible()` on robolectric activities before trying to interact with them in certain ways: http://robolectric.org/activity-lifecycle/
(`setup()` calls `visible()`)

app/src/org/commcare/CommCareApplication.java
```diff
@@ -841,13 +841,14 @@ public void clearUserData() {
...
 
-        // manually clear file-backed fixture storage to ensure files are removed
-        CommCareApplication._().getFileBackedUserStorage("fixture", FormInstance.class).removeAll();
-
         CommCareApplication._().closeUserSession();
     }
 
     public void wipeSandboxForUser(final String username) {
+        // manually clear file-backed fixture storage to ensure files are removed
+        CommCareApplication._().getFileBackedUserStorage("fixture", FormInstance.class).removeAll();
```
phillipm: Since the user db was being cleared before this was called I don't think it was doing anything, because the file paths were cleared before this code could remove the files. Hence leaving dangling fixture data (for fixtures over 1mb). Moving it here ensures that files are deleted before the db is cleared.

unit-tests/src/org/commcare/android/util/CaseLoadUtils.java
amstone326: yikes 😞

amstone326:
@phillipm Thank you so much for tracking this down!

PR #1501, opened 2016-09-23
Don't use "sense" mode for anythign in Android

amstone326:


phillipm:
🎉 Was always confused by what 'sense mode' was.

PR #1502, opened 2016-09-29
Add session nav 'skip' command

phillipm:
cross-request: https://github.com/dimagi/commcare-core/pull/417


phillipm:
verified that android tests pass locally (they are currently broken on the CI server)

PR #1503, opened 2016-10-04
Call checkFileSize on the correct image

amstone326:
Fix for http://manage.dimagi.com/default.asp?238084

### DIFF
--- a/app/src/org/commcare/views/widgets/ImageWidget.java
+++ b/app/src/org/commcare/views/widgets/ImageWidget.java
@@ -157,14 +157,16 @@ public class ImageWidget extends QuestionWidget {
             int screenWidth = display.getWidth();
             int screenHeight = display.getHeight();
 
-            // Check if we have a raw folder, and if so pull the image to display from there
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/views/widgets/ImageWidget.java
```diff
@@ -157,14 +157,16 @@ public void onClick(View v) {
...
+            // better quality
             File toDisplay = new File(mInstanceFolder + "/raw/" + mBinaryName);
             if (!toDisplay.exists()) {
-                // Otherwise, just use the image in the instance folder
-                toDisplay = new File(mInstanceFolder + "/" + mBinaryName);
+                toDisplay = imageBeingSubmitted;
             }
 
-            checkFileSize(toDisplay);
+            checkFileSize(imageBeingSubmitted);
```
amstone326: This is the fix -- A while back, I made it so that even when image capture resizing is being used, the original full-sized image is still the one that is displayed in the app, just because there isn't any reason not to and it will look nicer. This means that `checkFileSize()` needs to be called on the image in the instance folder that is going to ultimately be submitted, not the display image, since they are potentially different.

PR #1505, opened 2016-10-05
Allow hash-based labels on y axis for time graphs

orangejenny:
Labels should only be parsed as times when it's a time-based graph *and* they're for the x axis, since the y axis for time-based graphs is still numeric.

http://manage.dimagi.com/default.asp?238840


orangejenny:
The build failure is `groovy.lang.MissingPropertyException: Could not get unknown property 'processQabetaDebugGoogleServices' for root project 'commcare-android' of type org.gradle.api.Project` ...does this matter?

amstone326:
@orangejenny All mobile builds are failing right now, robolectric tests are broken after updates to the build server. Phillip is going to look into it when he's back online

PR #1506, opened 2016-10-07
Show clearer message when error encountered installing suite file

phillipm:
Instead of showing the message `'Couldn't find your application. Please check and retype your code or verify your internet connection'` upon trying to install an invalid suite, show the following:

install | pinned notification
---- | ----
screen | pinned

screen

For updates, it sort of looks like a wall of text. Should I move the error into a pinned notification?

cross-request: https://github.com/dimagi/commcare-core/pull/421

fix for http://manage.dimagi.com/default.asp?222137 and http://manage.dimagi.com/default.asp?237014


amstone326:
Hey, haven't reviewed code yet, but definitely agree with the proposal to move the message body for update errors into a pinned notification. Can work the same way as the install errors do. I would limit the text on the screen to ```"Failed to download updates. This usually means your app has a configuration error (See Notification for Details)"```

app/src/org/commcare/activities/CommCareSetupActivity.java
```diff
@@ -155,16 +156,32 @@
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
-        this.fromManager = this.getIntent().
-                getBooleanExtra(AppManagerActivity.KEY_LAUNCH_FROM_MANAGER, false);
+        fromManager = getIntent().getBooleanExtra(AppManagerActivity.KEY_LAUNCH_FROM_MANAGER, false);
 
-        if (checkForMultipleAppsViolation()) {
```
amstone326: Would you mind actually leaving this logic the way it was? I know you don't like this pattern in general, but in scenarios like this where it's really important that no code be called after this line if it returns `true`, it feels like it will be much easier for someone in the future to accidentally add code after this `if` statement that messes that up.

amstone326:
@phillipm Hey this is awesome, thanks for doing this. 2 high-level things:

1. It seems a little confusing that we now have both an `InvalidResourceStructureException` and an `InvalidStructureException` in our code base that mean slightly different things. Would be great to distinguish these better (or else just combine them if there isn't actually a difference?)

2. Would you mind adding a test for this, either at the robolectric level or to the QA plan? It sounds from previous discussions like we lost track of this changing at some point in the past, would be good to make sure that doesn't happen again.

phillipm:
@amstone326 ready to check out again. I added install / update robolectric tests and renamed `InvalidResourceStructureException`

app/src/org/commcare/tasks/UpdateTask.java
```diff
@@ -108,8 +108,7 @@ public void startPinnedNotification(Context ctx) {
         } catch (InvalidResourceStructureException e) {
             ResourceInstallUtils.logInstallError(e,
                     "Structure error ocurred during install|");
-            String structureError = "CommCare found an issue with the '" + e.resourceName + "' resource:\n" + e.getMessage();
-            return new ResultAndError<>(AppInstallStatus.UnknownFailure, structureError);
+            return new ResultAndError<>(AppInstallStatus.UnknownFailure, e.resourceName + "==" + e.getMessage());
```
amstone326: Ahh, gotcha. Would you mind adding a comment to that effect? Good to pull after!

phillipm: @amstone326 👍

PR #1510, opened 2016-10-12
Downgrade Robolectric SDK to 21

phillipm:
Revert robolectric target sdk bump from https://github.com/dimagi/commcare-android/pull/1498

Add robolectric add-on modules http://robolectric.org/using-add-on-modules/ (not entirely sure if we need these)

Bump android gradle plugin to latest Android Studio release

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,7 +18,7 @@ buildscript {
     }
 
     dependencies {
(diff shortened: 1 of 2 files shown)


amstone326:
latest commit added a bunch of logging, was that intentional?

phillipm:
ooops, fixed

PR #1517, opened 2016-10-19
Perform sms install cursor traversal in AsyncTask

phillipm:
Move cursor code that traverses the SMS inbox into the AsyncTask that processes the SMS body. This prevents the UI thread from blocking while getting SMSs (not that I noticed it blocking, but more on principle)

Yet to be tested because I can't get figure out how to trigger an sms install on HQ (I sent a request but my phone didn't receive it)


wpride:
@phillipm To test this I just asked @gcapalbo to generate the text for me, then sent to the device myself

gcapalbo:
@phillipm i think we already offlined this but let me know if you still need help testing this

PR #1518, opened 2016-10-19
Async case list search deadlock fix

phillipm:
Fix for a bug that was coming up where searching an async case list for the first time would freeze when you typed more than 1 character in quick succession.

The solution is to be stricter about cancelling the existing search process if a new search process is started; like when you type the second character of a query.

### DIFF
--- a/app/src/org/commcare/adapters/EntityStringFilterer.java
+++ b/app/src/org/commcare/adapters/EntityStringFilterer.java
@@ -83,7 +83,9 @@ public class EntityStringFilterer extends EntityFiltererBase {
             }
             Entity<TreeReference> e = fullEntityList.get(index);
             if (isCancelled()) {
-                break;
+                db.setTransactionSuccessful();
+                db.endTransaction();
+                return;
             }
 
             boolean add = false;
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/adapters/EntityStringFilterer.java
```diff
@@ -83,7 +83,9 @@ private void buildMatchList() {
             }
             Entity<TreeReference> e = fullEntityList.get(index);
             if (isCancelled()) {
-                break;
```
ctsims: Just realized this is a break, not a return, which makes me question a bit again whether I quite understand what's going on, since a break should _definitely_ still resulted in a succesful/end transaction pair being called, but this change should be an improvement either way.

amstone326:
@phillipm Do you know if this fixes the issues that REC was seeing that prompted me to create this ticket http://manage.dimagi.com/default.asp?238922?

phillipm:
@amstone326 Sounds like they were very similar issues. You could ask Jessica to see if she can reproduce the issue on 2.31.

amstone326:
Ah no I forgot it can't be because their issues were happening as soon as you press the search button

PR #1520, opened 2016-10-21
Audio playback UI update

phillipm:
| old | new |
| --- | --- |
| case_list_old_audio | case_list_new_audio |
| module_list_old_audio | module_list_new_audio |
| form_entry_old_audio | form_entry_new_audio |

**Release Notes**: update audio playback UI


app/src/org/commcare/views/EntityDetailView.java
```diff
@@ -70,14 +71,20 @@
...
+    private final AudioPlaybackButton audioButton;
     private final View valuePane;
     private View currentView;
     private final LinearLayout detailRow;
     private final LinearLayout.LayoutParams origValue;
     private final LinearLayout.LayoutParams origLabel;
     private final LinearLayout.LayoutParams fill;
 
+    @IdRes
+    private static final int imageViewID = 23422634;
```
amstone326: Could make these var names ALL_CAPS per our usual convention

app/src/org/commcare/views/media/AudioPlaybackButton.java
```diff
@@ -198,53 +225,62 @@ private String getAudioFilename() {
...
-            default:
-                Log.w(TAG, "Current playback state set to unexpected value");
-        }
+            }
+        };
     }
 
     private void startPlaying() {
-        AudioController.INSTANCE.playCurrentMediaEntity();
+        Pair<Integer, Integer> posAndduration = AudioController.INSTANCE.playCurrentMediaEntity();
```
amstone326: `posAndduration` --> `posAndDuration`

app/src/org/commcare/views/media/AudioController.java
```diff
@@ -28,24 +31,24 @@
      * Button that corresponds to the currentEntity media. Pressing the button
      * should trigger playback control methods here.
      */
-    private AudioButton currentButton;
+    private AudioPlaybackReset currentAudioReset;
```
amstone326: I'm not following why a whole new interface was needed for this. Why can't the `AudioPlaybackButton` still be the variable that is stored here, and have `resetPlaybackState` be a method of that class?

app/src/org/commcare/views/media/AudioPlaybackButton.java
```diff
@@ -260,4 +296,34 @@ private void pausePlaying() {
         currentState = MediaState.Paused;
         refreshAppearance();
     }
+
+    private void animateProgress(int milliPosition, int milliDuration) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.HONEYCOMB) {
+            ProgressBar progressBar = (ProgressBar)findViewById(R.id.circular_progress_bar);
+            animation = ObjectAnimator.ofInt(progressBar, "progress", 0, 500);
```
amstone326: What are the hard-coded 0 and 500 values for?

app/src/org/commcare/views/media/AudioController.java
phillipm: It was needed during an intermediate state of class refactoring.
Generally I'm in favor of using interfaces to pair-down a class to the collection of minimally needed methods. In this case, `AudioPlaybackButton` only had 1 other public method, so that benefits of using an interface didn't over-weigh the mental overhead of the indirection, so I'll take it out.

app/res/layout/menu_list_item_modern.xml
```diff
@@ -1,28 +1,30 @@
...
         android:layout_width="wrap_content"
         android:layout_weight="1"
-        style="@style/ListContentV2"
         android:ellipsize="end"
         android:maxLines="2"/>
 
-    <org.commcare.views.media.AudioButton
+    <org.commcare.views.media.AudioPlaybackButton
         android:id="@+id/row_soundicon"
+        android:clipChildren="false"
```
phillipm: oops, but in this case it isn't needed :) just an artifact of me blindly trying to make UI things work on Android.

app/src/org/commcare/views/media/AudioPlaybackButton.java
```diff
@@ -260,4 +294,36 @@ private void pausePlaying() {
         currentState = MediaState.Paused;
         refreshAppearance();
     }
+
+    private void animateProgress(int milliPosition, int milliDuration) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.HONEYCOMB) {
+            ProgressBar progressBar = (ProgressBar)findViewById(R.id.circular_progress_bar);
+            int startPosition = 0;
+            int progressDuration = 500;
```
amstone326: Sorry I should have been more explicit with my question here, but what I'm confused by is why the progressDuration is always 500?

phillipm: I think having a larger duration makes for a smoother animation, but not entirely sure..

amstone326: Hm, I'm still confused. Does the call to `animation.setDuration(milliDuration)` not just override the `500` value that was set in the previous line? Or do those "durations" actually mean different things?

phillipm: They seem to be different things. I tried taking out the `animation.setDuration(milliDuration)` call and it didn't work anymore.

amstone326: So from reading the documentation for `ofInt` and `setDuration`, my interpretation is that you want to flip these around. The [documentation for ofInt](https://developer.android.com/reference/android/animation/ObjectAnimator.html#ofInt(T, android.util.Property<T, java.lang.Integer>, int...)) says that `values` is "A set of values that the animation will animate between over time." While the [documentation for setDuration](https://developer.android.com/reference/android/animation/ObjectAnimator.html#setDuration(long)) says that `duration` is "The length of the animation, in milliseconds." Would you take the same conclusion from that?

phillipm: From what I understand `ObjectAnimator.ofInt(progressBar, "progress", x, y);` uses reflection to animate the `progress` property of the `ProgressBar` argument. The last two integer arguments (`x`/`y`) specify the start/end range for that property. I would expect the larger the range, the finer the granularity of the animation. 
`setDuration` sets how long the animation should take to get from the start position to the end position. In this case, the animation duration should correspond to the audio playback time, which is how it is now.
The `ObjectAnimator` documentation is sort of bad, so not positive...

amstone326: Ah ok you're right, that makes more sense now. Thanks for the clarification

PR #1521, opened 2016-10-21
Create large audio playback widget w/ seekbar

phillipm:
Add a audio playback widget type that shows progress information and allows seeking through the audio track by tapping (not dragging) the progress bar.

screen

Doesn't support drag seeking because that conflicts with navigation swipes

Expected itext:

```
<text id="seekable-audio-label">
  <value>large audio</value>
  <value form="expanded-audio">jr://file/commcare/audio/module0_en.mp3</value>
</text>
```


app/src/org/commcare/views/media/ExpandedAudioPlaybackView.java
```diff
@@ -0,0 +1,149 @@
...
+        if (AudioController.INSTANCE.mediaForInstance(ExpandedAudioPlaybackView.this)) {
+            AudioController.INSTANCE.seekTo(progress);
+        }
+        return false;
+    }
+
+    private void setupProgressAnimation(int currentPositionMillis) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.HONEYCOMB) {
+            final int startPosition = 0;
+            animation = ObjectAnimator.ofInt(seekBar, "progress", startPosition, playbackDurationMillis);
```
amstone326: So I think the way you did things here is what I was originally expecting in `startProgressBar` of AudioPlaybackButton, and what I was trying but failing to express well in our discussion on that PR. In that one, you picked an arbitrary value for the `progressDuration` value that is passed to `ofInt` (https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/views/media/AudioPlaybackButton.java#L303), whereas here you use the duration of the audio that is being played. Looking back at your explanation on that PR, it seems like what you did there might actually be more correct, in that you want to make sure that `progressDuration` is a large value in order to create a smooth animation, and it doesn't really matter if that value corresponds to the audio playback's length. Do you still think that's right?

app/src/org/commcare/views/media/AudioPlaybackButtonBase.java
```diff
@@ -0,0 +1,294 @@
...
+
+    protected abstract void pauseProgressBar();
+
+    protected abstract int getLayout();
+
+    @Override
+    protected void onWindowVisibilityChanged(int visibility) {
+        super.onWindowVisibilityChanged(visibility);
+
+        if (visibility != View.VISIBLE) {
```
phillipm: good catch, I'll try to find a fix for that

app/src/org/commcare/views/media/ExpandedAudioPlaybackView.java
phillipm: I have no idea why, but if I change it to match the other playback button, it doesn't work, so I'm going to leave it as is.

app/src/org/commcare/views/media/AudioPlaybackButtonBase.java
amstone326: @phillipm Good to merge this now once this issue is resolved

phillipm:
@amstone326 ready to go again. Had to make a lot of changes to make animation resumes work because a new button instance is created on rotation, so we need to reattach it to the currently playing entity.

PR #1525, opened 2016-10-24
Refactor: Make form upload result an enum instead of int

phillipm:
Catch `401` response codes on form upload and show warning to user:
screen

**Question**: Can someone think of a concise way to portray to the user that the data they collect under this username/password combo can not be synced going forward? 
(Oooh, unless they revert the password back to the old value...)

Contains commits from https://github.com/dimagi/commcare-android/pull/1375


amstone326:
@phillipm Good to merge after answer to my 1 question

PR #1526, opened 2016-10-25
Decode Base64 encoded URL in SMS install message

phillipm:
Change SMS install text body format from:
`[commcare app - do not delete] https://www.commcarehq.org/a/gc/sms/app_info/7c7d49fbef59b703fb468e20d52a21e4/`
to
`[commcare app - do not delete] aHR0cHM6Ly93d3cuY29tbWNhcmVocS5vcmcvYS9nYy9zbXMvYXBwX2luZm8vN2M3ZDQ5ZmJlZjU5YjcwM2ZiNDY4ZTIwZDUyYTIxZTQv`

Where the payload is a Base64 encoded version of the URL.

This prevents users from accidentally clicking the URL in the text, which confusingly points to a CC profile file.

I will update the code documentation once https://github.com/dimagi/commcare-android/pull/1517 get's merged, because otherwise there will be merge conflicts... 

@gcapalbo @orangejenny @wpride


gcapalbo:
@phillipm can we also support the old style for backwards-compatibility? some people might still have the old style sms on their phone, so if we could check if the string starts with "http" and use the old way that would be good. once enough time passes we could drop the old way entirely.

phillipm:
@gcapalbo yup, good call. Updated.

app/src/org/commcare/utils/SigningUtil.java
```diff
@@ -33,9 +36,16 @@
...
- * Created by wpride1 on 9/11/15.
+ * @author Will Pride (wpride@dimagi.com)
  */
 public class SigningUtil {
+    private final static ArrayList<String> WHITELISTED_URL_HOSTS = new ArrayList<>();
+
+    static {
+        WHITELISTED_URL_HOSTS.add("commcarehq.org");
+        WHITELISTED_URL_HOSTS.add("www.commcarehq.org");
+        WHITELISTED_URL_HOSTS.add("india.commcarehq.org");
```
wpride: does it make sense to add staging/swiss?

unit-tests/src/org/commcare/utils/SigningUtilTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+    @Test
+    public void validateBaseEncodedURLTest() throws UnsupportedEncodingException, Base64DecoderException {
+        // decode a valid URL
+        String goodUrlBase = "https://www.commcarehq.org/a/gc/sms/app_info/7c7d49fbef59b703fb468e20d52a21e4/";
+        String goodUrl = Base64.encode(goodUrlBase.getBytes("UTF-8"));
+        assertEquals(goodUrlBase, SigningUtil.decodeUrl(goodUrl));
+
+        // try to decode a 'malicious', of non-commcarehq origin, URL
+        String badUrlBase = "https://www.corncarehq.org/a/gc/sms/app_info/7c7d49fbef59b703fb468e20d52a21e4/";
```
wpride: damn you corncare

app/src/org/commcare/utils/SigningUtil.java
```diff
@@ -45,6 +45,8 @@
         WHITELISTED_URL_HOSTS.add("commcarehq.org");
         WHITELISTED_URL_HOSTS.add("www.commcarehq.org");
         WHITELISTED_URL_HOSTS.add("india.commcarehq.org");
+        WHITELISTED_URL_HOSTS.add("swiss.commcarehq.org");
```
phillipm: Good call, updated the PR. Can you double check my whitelist regex?

app/src/org/commcare/utils/SigningUtil.java
```diff
@@ -33,9 +37,14 @@
...
  * the CommCareHQ private key
  *
- * Created by wpride1 on 9/11/15.
+ * @author Will Pride (wpride@dimagi.com)
  */
 public class SigningUtil {
+    private final static ArrayList<Pattern> WHITELISTED_URL_HOSTS_REGEX = new ArrayList<>();
+
+    static {
+        WHITELISTED_URL_HOSTS_REGEX.add(Pattern.compile("^*.commcarehq.org"));
```
orangejenny: `^*` isn't valid, should be `"^.*.commcarehq.org"` or could use `".commcarehq.org$"` (or `"\.commcarehq\.org$"` to be explicit that you actually want periods)

app/src/org/commcare/utils/SigningUtil.java
```diff
@@ -50,12 +59,53 @@
...
+        boolean isURLWhitelisted = false;
+        for (Pattern urlPattern : WHITELISTED_URL_HOSTS_REGEX) {
+            if (urlPattern.matcher(host).find()) {
+                isURLWhitelisted = true;
+                break;
+            }
+        }
+
+        if (!isURLWhitelisted) {
+            throw new RuntimeException(url + " is not an approved URL.");
```
ctsims: Woah, just caught this, we should probably just throw a notification for this, not a runtime exception (or throw a typed exception here and catch/raise an notification later). Otherwise I can prevent your phone from ever being able to start up CommCare by sending you an invalid URL over SMS, which will make it crash on startup.

phillipm: The sms processing async task is wrapped in a broad try/catch to prevent this. I've updated it to throw a specific RuntimeException subclass, but I've left the umbrella try/catch in place. It still reports a generic 'sms install failed' toast message to the user if this comes up.

app/src/org/commcare/utils/SigningUtil.java
```diff
@@ -50,12 +59,53 @@
...
+        URL url;
+        try {
+            url = new URL(urlString);
+        } catch (MalformedURLException e) {
+            throw new RuntimeException(urlString + " is not a valid URL.");
+        }
+
+        String host = url.getHost();
+        boolean isURLWhitelisted = false;
+        for (Pattern urlPattern : WHITELISTED_URL_HOSTS_REGEX) {
```
wpride: Any reason to keep this as an ArrayList now? YAGNI and all

phillipm: Updated

wpride:
@ctsims holding off merge until you remove the Changes requested flag

PR #1535, opened 2016-10-27
Use same password for all ccz demo users

amstone326:
cross-request: https://github.com/dimagi/commcare-core/pull/436


amstone326:
@phillipm This fixes the issue where logging in after swipe closing the app would re-process the restore. The issue was that whenever the app does a full restart, a new CommCareApp object is initialized, which means a new CommCarePlatform object is initialized, and that is what stores the OfflineUserRestore, so we were re-generating that each time the app did a full restart. 

Now, any state that exists in the OfflineUserRestore is able to be restored when the object is recreated, assuming the underlying restore file is the same.

PR #1537, opened 2016-10-28
update Libs & Android API to 25

phillipm:
- Target Android API 25, which is a minor release that adds support for Google's new Pixel devices, so shouldn't have any large changes. 
- Support lib from `24.2.1` to `25.0.1`: https://developer.android.com/topic/libraries/support-library/revisions.html
- use simprints from jcenter instead of hand building every new release
- Play Services from `9.6.1` to `9.8.0`: https://developers.google.com/android/guides/releases#october_2016_-_v98
- use latest gradle plugin for android studio 2.2.2 
- update ACRA from `4.9.0` to `4.9.1`: https://github.com/ACRA/acra/wiki/ChangeLog
- update robolectric from `3.1.2` to `3.1.4`


phillipm:

PR #1514, opened 2016-10-14
Refix windows gradle issue

ctsims:
Revert change from
https://github.com/dimagi/commcare-android/pull/1498

which removes code that lets gradle builds work on windows.

PR #1538, opened 2016-10-28
Refix windows gradle issue

ctsims:
Moving 
https://github.com/dimagi/commcare-android/pull/1514

to 2.31 to ensure that windows builds will work historically


phillipm:
Test failure should be a farce. 
@ctsims can you verify this resolves your Windows build issues?

PR #1541, opened 2016-10-28
Use array length instead of hard code

wpride:
Fixes http://manage.dimagi.com/default.asp?240363

cc @proteusvacuum


proteusvacuum:
Woo! (are there no tests for this?)

wpride:
I doubt it, think this is still pretty much in hack-world

phillipm:
lol

ctsims:
It should be quite trivial to generate robolectric tests for this, fwiw

ctsims:
@wspride Are we going to add any tests around this? The change looks good, but agree that test coveage would be very good.

wpride:
@ctsims @phillipm pushed a test that triggers this exception on master - however, the test still passes because the crash happens on a not-main thread. Have either of you run into this before?

PR #1544, opened 2016-10-28
Fix for Ethiopian Month names in English

ctsims:
The Transliteration that was being used for ethiopian dates in english was quite strange and was matching up poorly to the best available sources.

Fix for: http://manage.dimagi.com/default.asp?240253


phillipm:
@ctsims can you look into the failing test?

ctsims:
@phillipm Fixed and passing now

PR #1547, opened 2016-11-01
check for SQL value > 1mb at android level

wpride:
Check that we're not writing a blob greater than 1mb in to SQL at the android level

cross-request: https://github.com/dimagi/commcare-core/pull/439


wpride:

wpride:

wpride:

app/src/org/commcare/models/database/SqlStorage.java
```diff
@@ -503,6 +505,10 @@ public void write(Persistable p) {
             update(p.getID(), p);
             return;
         }
+        int persistableSize = TableBuilder.toBlob(p).length;
```
phillipm: You probably want to check this in `update` as well.

phillipm: I think calling `toBlob` can be expensive because it re-serializes the entire object, which might get us into trouble with large cases / forms etc(?).
This change means that `toBlob` is being called twice, once in `DatabaseHelper.addDataToValues` and once here.
Can you think of a clean way to avoid this? 
Maybe passing in a `isSizeBounded` flag into the core code and only checking & throwing size exceptions on that? (definitely adds cruft to the code, but maybe okay in this situation..)

wpride:
@phillipm good call, think the new placement is more clean

PR #1548, opened 2016-11-02
RoboElectric catch async exceptions

wpride:
Contains https://github.com/dimagi/commcare-android/pull/1547, will split out if that isn't merged soon

Updates the CommCareTestApplication to check that all asynchronous activities have not crashed. RoboElectric would fail silently if an exception occurred off the main thread.

### DIFF
--- a/unit-tests/src/org/commcare/CommCareTestApplication.java
+++ b/unit-tests/src/org/commcare/CommCareTestApplication.java
@@ -53,7 +53,7 @@ public class CommCareTestApplication extends CommCareApplication implements Test
 
     private String cachedUserPassword;
 
-    public static ArrayList<Throwable> asyncExceptions = new ArrayList<>();
+    public ArrayList<Throwable> asyncExceptions = new ArrayList<>();
 
     @Override
     public void onCreate() {
--- a/unit-tests/src/org/commcare/CommCareTestApplication.java
+++ b/unit-tests/src/org/commcare/CommCareTestApplication.java
@@ -40,6 +40,7 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 
+import static junit.framework.Assert.fail;
 import static org.junit.Assert.assertTrue;
 
 /**
@@ -52,7 +53,7 @@ public class CommCareTestApplication extends CommCareApplication implements Test
 
     private String cachedUserPassword;
 
-    public static boolean asyncTestsPassed = true;
+    public static ArrayList<Throwable> asyncExceptions = new ArrayList<>();
 
     @Override
     public void onCreate() {
@@ -64,7 +65,7 @@ public class CommCareTestApplication extends CommCareApplication implements Test
         Thread.setDefaultUncaughtExceptionHandler(new Thread.UncaughtExceptionHandler() {
             @Override
             public void uncaughtException(Thread thread, Throwable ex) {
-                asyncTestsPassed = false;
+                asyncExceptions.add(ex);
                 Assert.fail(ex.getMessage());
             }
         });
@@ -225,7 +226,12 @@ public class CommCareTestApplication extends CommCareApplication implements Test
     @Override
     public void afterTest(Method method) {
         Robolectric.flushBackgroundThreadScheduler();
-        assertTrue(asyncTestsPassed);
+        if (asyncExceptions.size() > 0) {
+            for(Throwable throwable: asyncExceptions) {
+                throwable.printStackTrace();
+                fail("Test failed with " + asyncExceptions.size() + " asyncronous exceptions.");
+            }
+        }
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


ctsims:
Looks like 99% of the test failures are coming because we don't have a clear way to clean up the DB state after each test

https://github.com/robolectric/robolectric/issues/1890

unit-tests/src/org/commcare/CommCareTestApplication.java
```diff
@@ -214,4 +221,20 @@ public ModernHttpRequester buildModernHttpRequester(Context context, URL url,
     public DataPullRequester getDataPullRequester() {
         return LocalDataPullResponseFactory.INSTANCE;
     }
+
+    @Override
+    public void afterTest(Method method) {
+        Robolectric.flushBackgroundThreadScheduler();
+        assertTrue(asyncTestsPassed);
```
ctsims: This ends up bubbling through to the end as "the failure", so it's probably worth retaining one (I know there might be more than one, but still good to have at least one) of the exceptions to this point to pass as the assertion failure. It would make the output less confusing, I think.

wpride: Good call

wpride:
@ctsims I actually think that happened because I had left the old bug in accidentally (when confirming the regression test worked) and then the "failed" variable was still set for each subsequent run.

wpride:
Changed the variable from static to class which should cause these to not cascade

unit-tests/src/org/commcare/CommCareTestApplication.java
```diff
@@ -214,4 +222,25 @@ public ModernHttpRequester buildModernHttpRequester(Context context, URL url,
...
         return LocalDataPullResponseFactory.INSTANCE;
     }
+
+    @Override
+    public void afterTest(Method method) {
+        Robolectric.flushBackgroundThreadScheduler();
+        if (asyncExceptions.size() > 0) {
+            for(Throwable throwable: asyncExceptions) {
+                throwable.printStackTrace();
+                fail("Test failed with " + asyncExceptions.size() + " asyncronous exceptions.");
```
ctsims: I think we actually want to display the exception itself as the failure. Isn't there a "Fail" message which directly takes in an exception?

wpride: Not that I can see, I think generally they except you to just not catch exceptions if that's the failure you want

wpride: One option would be to re-throw in the afterTest() method but the output doesn't look significantly more useful

ctsims: Ok, in that case can you just change the message to be more clear? IE:

"Test failed due to [2] threads crashing off the main thread. See error log for more details"

ctsims:
Looks good to me. Will pull as soon as the label indicates things are g2g.

unit-tests/src/org/commcare/CommCareTestApplication.java
```diff
@@ -29,25 +30,32 @@
...
  */
-public class CommCareTestApplication extends CommCareApplication {
+public class CommCareTestApplication extends CommCareApplication implements TestLifecycleApplication {
     private static final String TAG = CommCareTestApplication.class.getSimpleName();
     private static PrototypeFactory testPrototypeFactory;
     private static final ArrayList<String> factoryClassNames = new ArrayList<>();
 
     private String cachedUserPassword;
 
+    public ArrayList<Throwable> asyncExceptions = new ArrayList<>();
```
phillipm: `public` -> `private final`

unit-tests/src/org/commcare/CommCareTestApplication.java
```diff
@@ -214,4 +223,26 @@ public ModernHttpRequester buildModernHttpRequester(Context context, URL url,
     public DataPullRequester getDataPullRequester() {
         return LocalDataPullResponseFactory.INSTANCE;
     }
+
+    @Override
+    public void afterTest(Method method) {
+        Robolectric.flushBackgroundThreadScheduler();
+        if (asyncExceptions.size() > 0) {
```
phillipm: `.size() > 0` -> `isEmpty()`

unit-tests/src/org/commcare/CommCareTestApplication.java
```diff
@@ -29,25 +30,32 @@
...
+import java.lang.reflect.Method;
 import java.net.URL;
 import java.util.HashMap;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 
+import static junit.framework.Assert.fail;
+import static org.junit.Assert.assertTrue;
```
phillipm: unused

PR #1549, opened 2016-11-03
Don't swallow exceptions that occur during media validation

amstone326:
Fixes 2 problems:
1. Exceptions that occurred during `VerificationTask` were being swallowed, 
2. The way in which they were swallowed actually resulted in the `VerificationTask` _always_ succeeding if there was a swallowed exception, even if there had been media validation failures as well.

Partial fix for http://manage.dimagi.com/default.asp?240552 (Partial because I haven't figured out what's actually causing the underlying exception there)

### DIFF
--- a/app/src/org/commcare/tasks/VerificationTask.java
+++ b/app/src/org/commcare/tasks/VerificationTask.java
@@ -16,7 +16,7 @@ import org.javarosa.core.util.SizeBoundVector;
  * @author ctsims
  */
 public abstract class VerificationTask<Reciever>
-        extends CommCareTask<String, int[], SizeBoundVector<MissingMediaException>, Reciever>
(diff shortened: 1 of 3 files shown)


amstone326:
It's possible that this is worthy of a point release, given that it's going to prevent this project from loading their multimedia properly. @ctsims, what do you think?

app/src/org/commcare/tasks/VerificationTask.java
```diff
@@ -16,35 +16,30 @@
...
+        setTableListeners(global);
+        global.verifyInstallation(problems);
+        unsetTableListeners(global);
 
-            if (problems.size() > 0) {
-                return problems;
-            }
-            return null;
-        } catch (Exception e) {
-            // TODO: make non-resource missing failures have a better exception
```
wpride: ugh...

PR #1553, opened 2016-11-04
Infrastructure for using root module menu as home screen

amstone326:
(Would NOT recommend reviewing this commit-by-commit)

-A bunch of simplifying and reorganizing of sync-related code, including adding the notion of a `SyncCapableCommCareActivity`, in order to make syncing from other places in the app more straightforward. 
-Add a developer preference that, if enabled, uses the root module menu as the app's home screen. This means that the normal CC home screen never shows, and icons are added to the top bar on the main module screen for syncing and logging out instead:
screenshot_20161107-111916


amstone326:
fyi @ctsims, assuming we might want to iterate on the exact UI a bit, but just gave a first go at it for now

app/src/org/commcare/activities/MenuBase.java
```diff
@@ -74,7 +83,50 @@ protected boolean onBackwardSwipe() {
...
+        addHomeScreenActionsToTopBar(menu);
+        return true;
+    }
+
+    private void addHomeScreenActionsToTopBar(android.view.Menu menu) {
+        if (menuIsBeingUsedAsHomeScreen()) {
+            ViewUtil.addItemToActionBar(menu, MENU_LOGOUT, MENU_GROUP_HOME_SCREEN_ACTIONS, "Logout",
+                    R.drawable.ic_logout_action_bar);
+            ViewUtil.addItemToActionBar(menu, MENU_SYNC, MENU_GROUP_HOME_SCREEN_ACTIONS, "Sync",
+                    R.drawable.ic_sync_action_bar);
```
phillipm: Do we want these to be shown for consumer apps?

amstone326: No

amstone326: Sorry that sent prematurely.. No, the consumer apps flag is separate, and for that we still don't want any notion of any of the home screen actions.

ctsims:
Sweet! Definitely a good first pass. Might be good for us to brainstorm this week on how we want to be thinking about the sync experience.

PR #1556, opened 2016-11-07
Add developer option for updating to latest saved app version

amstone326:
Thanks to some iterating with @esoergel and @orangejenny, it is now possible to update to the latest saved version of an app on HQ (i.e. without a build having been created). However, there is still a significant limitation with this, which has been outlined in full in this ticket: http://manage.dimagi.com/default.asp?237199. The summary is that the way HQ's resource versioning system currently works, updates to "latest saved" will re-download ALL app resources (as opposed to only those that have actually changed since the last version), which means it will be pretty inefficient for apps with multimedia. 

Given this limitation, it doesn't make sense to advertise this feature to GS yet. However, I wanted to still make it physically possible to enable in the UI, since it could still be useful to the tech team while debugging or app building. It's a bit of an unknown how soon it will be possible for HQ to resource the remaining work for this, but I'll try to keep nudging it. 

Documentation added to Confluence here: https://confluence.dimagi.com/display/ccinternal/Advanced+Mobile+Debugging#AdvancedMobileDebugging-ChangingyourUpdateTarget


app/res/xml/commcare_preferences.xml
```diff
@@ -32,7 +32,7 @@
         android:defaultValue="no"
         android:entries="@array/pref_enabled_labels"
         android:entryValues="@array/pref_enabled_vals"
-        android:key="cc-newest-version-from-hq"
+        android:key="cc-update-to-unstarred-builds"
```
phillipm: I'm wondering if there is any way that this preference key change could cause issues for people who have already enabled this feature. Did you check this?

amstone326: Hm, I didn't. I imagine the extent of the issue will be that if you already had it enabled, you will have to re-enable it, since the preference for which a "YES" value was saved is no longer the one we are using. I don't think I'm concerned by that since this setting is only ever used for debugging purposes (as opposed to being a setting that projects turn on permanently for all of their users and expect it to stay that way). In other words, it seems like OK behavior to me for an app builder to have to re-enable this particular setting after updating to a new version of CommCare. What do you think?

PR #1560, opened 2016-11-09
Enhancements to root menu home screen

amstone326:


amstone326:
Screenshots of what `RootMenuHomeActivity` looks like:
screenshot_20161115-165009
screenshot_20161115-165013

amstone326:

app/src/org/commcare/activities/components/MenuList.java
```diff
@@ -0,0 +1,112 @@
...
+import org.commcare.dalvik.R;
+import org.commcare.fragments.BreadcrumbBarFragment;
+import org.commcare.session.SessionFrame;
+import org.commcare.suite.model.Entry;
+import org.commcare.suite.model.Menu;
+
+public class MenuList implements AdapterView.OnItemClickListener {
+
+    protected CommCareActivity activity;
+    protected AdapterView adapterView;
```
phillipm: `AdapterView` -> `AdapterView<ListAdapter>` to prevent lint warning

app/src/org/commcare/activities/components/MenuList.java
```diff
@@ -0,0 +1,112 @@
...
+        }
+        menuView.setupMenuInActivity(activity, menuId);
+        menuView.beingUsedInHomeScreen = beingUsedInHomeScreen;
+    }
+
+    public int getLayoutFileResource() {
+        return R.layout.screen_suite_menu;
+    }
+
+    protected void setupMenuInActivity(CommCareActivity activity, String menuId) {
```
phillipm: can be private

app/src/org/commcare/adapters/NavDrawerAdapter.java
```diff
@@ -0,0 +1,43 @@
...
+
+    Context context;
+
+    public NavDrawerAdapter(Context context, NavDrawerItem[] drawerItems) {
+        super(context, android.R.layout.simple_list_item_1, drawerItems);
+        this.context = context;
+    }
+
+    @Override
+    public View getView(int position, View convertView, ViewGroup parent) {
```
phillipm: should have nonnull annotations:
`public @NonNull View getView(int position, View convertView, @NonNull ViewGroup parent) {`

app/src/org/commcare/adapters/NavDrawerAdapter.java
```diff
@@ -0,0 +1,43 @@
...
+        View v = convertView;
+        if (v == null) {
+            v = View.inflate(context, R.layout.nav_drawer_item_view, null);
+        }
+        NavDrawerItem item = this.getItem(position);
```
phillipm: `getItem` has a `@Nullable` annotation, so you should probably check if it is null before using it

app/src/org/commcare/activities/components/NavDrawerItem.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Represents a single item in the RootMenuHomeActivity's nav drawer
+ *
+ * @author Aliza Stone
+ */
+public class NavDrawerItem {
+
+    public String id;
+    public String text;
```
phillipm: these (and `iconResource`) can be `final`

app/src/org/commcare/activities/MenuActivity.java
```diff
@@ -0,0 +1,60 @@
...
+    private static final String MENU_STYLE_GRID = "grid";
+    
+    @Override
+    protected void onCreateSessionSafe(Bundle savedInstanceState) {
+        super.onCreateSessionSafe(savedInstanceState);
+        String menuId = getIntent().getStringExtra(SessionFrame.STATE_COMMAND_ID);
+        if (menuId == null) {
+            menuId = Menu.ROOT_MENU_ID;
+        }
+        if (menuId == Menu.ROOT_MENU_ID && DispatchActivity.useRootMenuHomeActivity()) {
```
phillipm: `==` -> `.equals`

app/src/org/commcare/activities/MenuActivity.java
```diff
@@ -0,0 +1,60 @@
...
+        }
+        // If not, check style attribute for this particular menu block
+        AndroidCommCarePlatform platform = CommCareApplication.instance().getCommCarePlatform();
+        String commonDisplayStyle = platform.getMenuDisplayStyle(currentCommand);
+        return MENU_STYLE_GRID.equals(commonDisplayStyle);
+    }
+
+    @Override
+    public String getActivityTitle() {
+        //return adapter.getMenuTitle();
```
phillipm: would be good to remove commented out code

app/src/org/commcare/activities/RootMenuHomeActivity.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    @Override
+    public boolean onOptionsItemSelected(MenuItem item) {
+        if (usingNavDrawer() && item.getItemId() == android.R.id.home) {
+            if (navDrawerController.isDrawerOpen()) {
+                navDrawerController.closeDrawer();
+            } else {
+                navDrawerController.openDrawer();
+            }
```
phillipm: It would be nice to move more of the nav drawer control code into the `NavDrawerController` class. For instance, the 3 lines above could be placed in something like `NavDrawerController.toggleDrawer()`. 
Also, you could move the `outState.putBoolean(KEY_DRAWER_WAS_OPEN, navDrawerController.isDrawerOpen());` into the controller class so that when you look at this class you don't have to think about the state of the nav drawer at all. This would also allow you to hide `reopenDrawerInOnResume`

app/src/org/commcare/adapters/NavDrawerAdapter.java
amstone326: why is that?

amstone326: Or rather, I see that the method declaration for the parent has it, but why is it important for us to copy that?

app/src/org/commcare/activities/MenuActivity.java
amstone326: this was here from before my refactor so I was hesitant to do so, but probably makes sense to

amstone326:
@phillipm Addressed all of your comments in the last 3 commits, thanks for the suggestions!

app/src/org/commcare/adapters/NavDrawerAdapter.java
phillipm: Nullable/NonNull annotations provide static safety guarantees about code. These guarantees let the IDE remind you when you can skip null checks or should perform null checks. The linter complains here because you are wiping those guarantees when you don't need to be.

PR #1561, opened 2016-11-10
Rename _() methods because in Java 9 _ is not a valid identifier

phillipm:
Getting rid of a lot of build warnings

cross-request: https://github.com/dimagi/commcare-core/pull/448


amstone326:
@phillipm Hey, being picky, but would you be opposed to using `instance()` for all of these instead of `getInstance()`. Just because we use it in so many places, it seems nice to keep it a little more concise.

phillipm:
@amstone326 sounds good, updated

amstone326:
awesome thanks!

PR #1566, opened 2016-11-16
Visual sync indicators on alternative home screen

amstone326:
This PR adds a couple of visual indicators for an app's sync state to the `RootMenuHomeActivity` (all other sync-capable activities are untouched/unchanged). The 2 changes are:

1) The icon in the right corner of the action bar that triggers a sync when pressed now also uses different icons/animations to indicate the app's current "sync status", which can be 1 of 4 things: sending forms, pulling data, has pending forms to send, or up-to-date. The first 2 states use basic animations; the screenshot below is a snapshot of the animation for pulling data, and the animation for sending forms is the same but in the opposite direction.
screenshot_20161129-121642

This is what the icon looks like when there are pending forms to send:
screenshot_20161129-124754

And this is what the icon looks like when everything is up-to-date:
screenshot_20161129-121434

2) When CommCare is sending forms to the server, a progress bar will appear beneath the action bar to indicate the status of that task:
screenshot_20161129-121518

NOTE: One limitation of this is that in the current state, I do not handle re-attaching the progress bar and animations on rotation. I am intentionally not worrying about that in this PR (it's not necessary for PlusCare's current needs), but definitely intend to address it in a later PR when I have more time.

If you want to test this out yourself, you just need to turn on the `cc-use-root-menu-as-home-screen` Developer Preference, or use this app, which already has it enabled: https://www.commcarehq.org/a/aliza-test/apps/view/86249082b422087135af20b97485e519/.


app/res/layout/grid_menu_layout.xml
```diff
@@ -14,9 +14,18 @@
     <FrameLayout
         android:layout_width="match_parent"
         android:layout_height="match_parent"
-        android:layout_alignParentBottom="true"
         android:layout_below="@id/universal_frame_tile">
 
+        <ProgressBar
```
amstone326: since the `RootMenuHomeActivity` uses this layout file as its content view, the holder for the submission progress bar goes here

app/src/org/commcare/activities/SyncCapableCommCareActivity.java
```diff
@@ -159,6 +166,165 @@ public static void handleSyncUpdate(CommCareActivity activity,
...
+        addSyncItemToActionBar(menu);
+        return true;
+    }
+
+    private void addSyncItemToActionBar(Menu menu) {
+        if (shouldShowSyncItemInActionBar() &&
+                Build.VERSION.SDK_INT >= Build.VERSION_CODES.HONEYCOMB) {
+            currentSyncMenuItem = menu.add(MENU_GROUP_SYNC_ACTION, MENU_SYNC, MENU_SYNC, "Sync");
+            currentSyncMenuItem.setShowAsAction(MenuItem.SHOW_AS_ACTION_ALWAYS);
+            switch (syncStateForIcon) {
```
wpride: Did you consider removing this as a class variable and running computeSyncState() just in time when the value is needed?

app/src/org/commcare/activities/SyncCapableCommCareActivity.java
```diff
@@ -159,6 +166,165 @@ public static void handleSyncUpdate(CommCareActivity activity,
...
+    private void triggerSyncIconRefresh(int trigger) {
+        if (shouldShowSyncItemInActionBar()) {
+            computeSyncState(trigger);
+            rebuildOptionsMenu();
+        }
+    }
+
+    private void computeSyncState(int trigger) {
+        switch(trigger) {
+            case TRIGGER_END_DATA_PULL:
```
wpride: These three variables are used identically, could be replaced with one

app/src/org/commcare/activities/SyncCapableCommCareActivity.java
```diff
@@ -159,6 +166,159 @@ public static void handleSyncUpdate(CommCareActivity activity,
...
+        }
+    }
+
+    @Override
+    public void stopBlockingForTask(int id) {
+        super.stopBlockingForTask(id);
+        if (id == ProcessAndSendTask.SEND_PHASE_ID_NO_DIALOG ||
+                id == ProcessAndSendTask.PROCESSING_PHASE_ID_NO_DIALOG) {
+            triggerSyncIconRefresh(TRIGGER_END_SEND_FORMS);
+        } else if (id == DataPullTask.DATA_PULL_TASK_ID) {
```
wpride: the if and else clauses here will both result in the same behavior below

app/src/org/commcare/tasks/ProcessAndSendTask.java
```diff
@@ -68,12 +76,13 @@ public ProcessAndSendTask(Context c, String url, boolean inSyncMode) {
...
         this.url = url;
         this.processor = new FormRecordProcessor(c);
+        this.formSubmissionListeners = new ArrayList<>();
         if (inSyncMode) {
             this.sendTaskId = SEND_PHASE_ID;
             this.taskId = PROCESSING_PHASE_ID;
         } else {
-            this.sendTaskId = -1;
-            this.taskId = -1;
+            this.sendTaskId = SEND_PHASE_ID_NO_DIALOG;
```
wpride: Already off-lined with @amstone326 about this and agreed out of the purview of this PR but seems like splitting the process and send portions of this task would be a good refactor. The usage of ID to denote state here gets a bit confusing/messy IMO

PR #1567, opened 2016-11-20
Include IMEI in key fetch URL

phillipm:
Adding IMEI as a query parameter into the key fetch url for auditing large deployments .

Feture requested in http://manage.dimagi.com/default.asp?242009

### DIFF
--- a/app/src/org/commcare/network/HttpRequestGenerator.java
+++ b/app/src/org/commcare/network/HttpRequestGenerator.java
@@ -197,6 +197,10 @@ public class HttpRequestGenerator implements HttpRequestEndpoints {
             url = url.buildUpon().appendQueryParameter("last_issued", DateUtils.formatTime(lastRequest, DateUtils.FORMAT_ISO8601)).build();
         }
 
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/network/HttpRequestGenerator.java
```diff
@@ -197,6 +197,10 @@ public HttpResponse makeKeyFetchRequest(String baseUri, Date lastRequest) throws
             url = url.buildUpon().appendQueryParameter("last_issued", DateUtils.formatTime(lastRequest, DateUtils.FORMAT_ISO8601)).build();
         }
 
+        // include IMEI in key fetch request for auditing large deployments
+        url = url.buildUpon().appendQueryParameter("device_id",
+                CommCareApplication.instance().getPhoneId()).build();
```
phillipm: I wonder if we should just use the current approach of adding the device id to requests: https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/network/HttpRequestGenerator.java#L205-L212

PR #1571, opened 2016-11-21
Fix map select feature

phillipm:
If a GPS capture question has the appearance attribute `maps`, then you are shown a map where you can select a location independent of your current GPS location. This stopped working when we updated to Google's new map APIs.

Since this feature is not really used, I left some things rough:

 - If you tap a location, it selects a new location, but the accuracy is hard-coded to 10 meters. A more robust implementation would scale the accuracy by the current zoom level.
 - When your location updates or you tap a new location, the zoom level is scaled by the accuracy. This works but could use some finessing to scale better.

Behavioral changes:

 - The activity no longer finishes after a certain accuracy is obtained. From my understanding the whole point of the activity is to not pick your current location.

Had to rewrite a large chunk of the code, so might be useful to disregard the diff and just look at the new `GeoPointMapActivity` implementation.

screen

Fix for http://manage.dimagi.com/default.asp?237541

### DIFF
--- a/app/res/layout/geopoint_layout.xml
+++ b/app/res/layout/geopoint_layout.xml
@@ -11,7 +11,7 @@
         android:layout_width="fill_parent"
         android:layout_height="fill_parent"
         android:clickable="true"
-        android:apiKey="017Xo9E6R7WmcCITvo-lU2V0ERblKPqCcguwxSQ"/>
+        />
     <LinearLayout
         android:id="@+id/location_buttons"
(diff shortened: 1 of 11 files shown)


phillipm:
Note, if you want to test out google maps features with a debug build, you need to get a debug api key from google's developer console webpage and place it in your `~/.gradle/gradle.properties` file under the variable `GOOGLE_PLAY_MAPS_API_KEY=...your api key...`

PR #1576, opened 2016-11-23
Audio recorder widget filename fix: don't save files as mp40

phillipm:
Audio files captured using the built-in audio capture widget were getting saved as mp40, which meant they weren't getting uploaded to CommCare HQ.

### DIFF
--- a/app/src/org/commcare/views/widgets/RecordingFragment.java
+++ b/app/src/org/commcare/views/widgets/RecordingFragment.java
@@ -39,8 +39,9 @@ import java.io.IOException;
 public class RecordingFragment extends android.support.v4.app.DialogFragment {
 
     private String fileName;
-    private static final String FILE_EXT =
-            "/Android/data/org.commcare.dalvik/temp/Custom_Recording.mp4";
+    private static final String FILE_BASE =
+            "/Android/data/org.commcare.dalvik/temp/Custom_Recording";
+    private static final String FILE_EXT = ".mp4";
 
     private LinearLayout layout;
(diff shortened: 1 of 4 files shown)


app/src/org/commcare/views/widgets/RecordingFragment.java
```diff
@@ -63,7 +64,7 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
         prepareButtons();
         prepareText();
         setWindowSize();
-        fileName = Environment.getExternalStorageDirectory().getAbsolutePath() + FILE_EXT + listener.getFileExtension();
+        fileName = Environment.getExternalStorageDirectory().getAbsolutePath() + FILE_BASE + listener.getFileExtension() + FILE_EXT;
```
amstone326: Looking at the implementation of getFileExtension() in CommCareAudioWidget, it looks like getFileExtension is a really misleading method name. Isn't it just providing some unique identifier for the given audio widget? If that's correct, can we rename it?

amstone326: Looking at the implementation of getFileExtension() in CommCareAudioWidget, it looks like getFileExtension is a really misleading method name. Isn't it just providing some unique identifier for the given audio widget? If that's correct, can we rename it?

amstone326:
@phillipm Actually I realized this has to get in before Friday and you're not gonna be around before then, so merging this PR and opening another 1 to make the method name change.

PR #1581, opened 2016-11-24
Working towards a smaller FormEntryActivity

phillipm:
The `FormEntryActivity` class is 1800 lines long, this PR starts breaking it up into several different classes to make it more readable.


app/src/org/commcare/activities/components/FormEntrySessionWrapper.java
```diff
@@ -0,0 +1,99 @@
...
+
+import static org.commcare.activities.FormEntryActivity.mFormController;
+
+/**
+ * Wrapper for the form entry action recorder, allowing for saving and
+ * restoring to the form entry activity.
+ *
+ * @author Phillip Mates (pmates@dimagi.com)
+ */
+public class FormEntrySessionWrapper {
```
amstone326: What's the reason this code can't just be part of `FormEntrySession`?

app/src/org/commcare/activities/components/FormEntryInstanceUtils.java
```diff
@@ -0,0 +1,145 @@
...
+import org.commcare.provider.FormsProviderAPI;
+import org.commcare.provider.InstanceProviderAPI;
+import org.commcare.utils.StringUtils;
+import org.commcare.views.dialogs.DialogChoiceItem;
+import org.commcare.views.dialogs.PaneledChoiceDialog;
+
+/**
+ * @author Phillip Mates (pmates@dimagi.com)
+ */
+public class FormEntryInstanceUtils {
```
amstone326: What do you think about putting all of these methods into `FormEntryInstanceState.java`? I feel like there is a balance to strike here between making mammoth classes smaller and having an unnecessary number of small classes that don't have a clear reason for being separate entities and make it kind of cumbersome to navigate through code.

app/src/org/commcare/activities/components/FormEntrySessionWrapper.java
phillipm: It contains Android specific code and `FormEntrySession` needs to live in `commcare-core`

app/src/org/commcare/activities/components/FormEntryInstanceUtils.java
phillipm: 👍

phillipm:
@amstone326 I've updated the PR with your suggested changes

amstone326:

PR #1590, opened 2016-11-30
Module badges

amstone326:
cross-request: https://github.com/dimagi/commcare-core/pull/476


amstone326:
@ctsims This is how the badges look currently
screenshot_20161201-120431

app/res/layout/menu_grid_item.xml
```diff
@@ -1,28 +1,55 @@
...
+
+        <RelativeLayout
+            android:id="@+id/badge_view"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:layout_alignParentBottom="true"
+            android:visibility="gone">
+
+            <ImageView
+                android:layout_width="@dimen/numeric_badge_width"
```
amstone326: this is a not-great solution to the fact that I can't size the badge ImageView dynamically based upon the badge's text, since the TextView isn't actually inside of the ImageView in the xml, it's just overlaid on top of it (all sorts of weird stuff started to happen when I tried to set it up that way - I think maybe you can't add elements within an ImageView?). So right now I'm just setting the `numeric_badge_width` dimension to be twice the `numeric_badge_font_size`, under the assumption that the text being shown will be 2 digits long or less... Very open to a better suggestion.

amstone326:
Updated screenshot with red color palette instead: 
screenshot_20161202-102540

app/res/values/dimens.xml
```diff
@@ -26,6 +26,12 @@
     <dimen name="min_button_height">52dp</dimen>
     <dimen name="spacer">10dp</dimen>
 
+    <dimen name="menu_icon_size">60dp</dimen>
+    <dimen name="menu_grid_icon_size">120dp</dimen>
+    <dimen name="menu_grid_icon_size_with_badge">136dp</dimen>
+    <dimen name="numeric_badge_font_size">15dp</dimen>
+    <dimen name="numeric_badge_width">30dp</dimen>
```
amstone326: @ctsims I was previously using `sp` for `numeric_badge_font_size` and `numeric_badge_width` because generally we want `sp` for text (and the badge_width has to be based on the font size), but based on the conversation we just had, I think we may actually want to use `dp` here instead, so that the size of the badge stays proportional with the size of the entire menu item on different screens. Does that sound right to you?

app/res/layout/menu_grid_item.xml
```diff
@@ -1,28 +1,55 @@
...
+                android:layout_height="@dimen/numeric_badge_width"
+                android:layout_centerInParent="true"
+                android:src="@drawable/badge_circle"/>
+
+            <TextView
```
ctsims: Is there a reason we did a TextView over an imageview rather than a TextView with the image drawable background? Assuming it's because it's hard to keep the aspect ratio/size consistent if it's framed by the textview rather than a static image?

app/res/values/dimens.xml
ctsims: Hm, agree that they should either both be SP or both be DP. It is definitely the case that our _icons_ wont' scale with sp, so makes a lot of sense to me to frame/anchor the size of these components to the icon rather than the menu text. Sounds good!

PR #1591, opened 2016-12-02
Go back to checking for all possible ProcessAndSendTask ids

amstone326:
@wpride The simplification I tried to make here after talking with you actually doesn't work for all cases, so need to revert. I can explain more in person if you want, it's a little hard to write out.


amstone326:
@damagatchi retest please

amstone326:

PR #1598, opened 2016-12-09
Add license reporting plugin

wpride:
run with ./gradlew downloadLicenses


phillipm:
What does this do?

wpride:
You should have the email thread about this, if you run that command you'll get a list of the licenses we're dependent on

PR #1609, opened 2017-01-02
Try SMS install only when on 'install method chooser' menu

phillipm:
SMS install was running whenever you rotated the device, even if you were in the middle of another install. This was causing the install progress dialog to disappear.

Fix for http://manage.dimagi.com/default.asp?244434


gcapalbo:
@phillipm @wspride will the sms install still happen automatically when you open commcare for the first time if the install sms is present on the device?

phillipm:
Yeah, when you launch the app it will run. The change prevents the sms install code from running after you've left the initial 'installation selection method' screen.

PR #1613, opened 2017-01-06
Combobox widget

amstone326:
I pinged Kriti and Shayoni to get feedback on how this looks to them, but the PR can still be reviewed and merged in the mean time (since it needs to be in by Friday; I can always make tweaks based on feedback after that if need be).

To test out yourself, just change the appearance attribute of  a select-one question to either "combobox", which will use the `StandardComboboxAdapter`, or "combobox permissive", which will use the `PermissiveComboboxAdapter`. UATBC will be using the latter.

cross-request: https://github.com/dimagi/commcare-core/pull/506


app/src/org/commcare/adapters/PermissiveComboboxAdapter.java
```diff
@@ -0,0 +1,96 @@
...
+     * @return true if choiceLowerCase contains any word within textEntered (the "words" of
+     * textEntered are obtained by splitting textEntered on " ")
+     */
+    private boolean choiceShouldBeShown(String choiceLowerCase, CharSequence textEntered) {
+        if ("".equals(textEntered) || textEntered == null) {
+            return true;
+        }
+        String[] enteredTextIndividualWords = textEntered.toString().split(" ");
+        for (String word : enteredTextIndividualWords) {
+            if (!choiceLowerCase.contains(word.toLowerCase())) {
```
ctsims: As a heads up, btw, we have a method

StringUtils.fuzzyMatch() that we can consider using if they need an even more permissive matching

app/src/org/commcare/views/Combobox.java
```diff
@@ -0,0 +1,138 @@
...
+    private CharSequence lastAcceptableStringEntered = "";
+    private int lastValidCursorLocation;
+    private boolean fixingInvalidEntry;
+    private ComboboxAdapter customAdapter;
+
+    public Combobox(Context context, Vector<String> choices, boolean addEmptyFirstChoice,
+                    boolean permissive, int fontSize) {
+        super(context);
+        this.choices = choices;
+        setupAdapter(context, fontSize, addEmptyFirstChoice, permissive);
```
ctsims: It seems like passing in an adapter itself, and then setting the items in "Setup adapter" might be a more extensible way to manage this in the future if we need to allow for a few different matching styles?

ctsims: (definitely not a big deal though)

app/src/org/commcare/views/Combobox.java
```diff
@@ -0,0 +1,138 @@
...
+public class Combobox extends AutoCompleteTextView {
+
+    private Vector<String> choices;
+    private CharSequence lastAcceptableStringEntered = "";
+    private int lastValidCursorLocation;
+    private boolean fixingInvalidEntry;
+    private ComboboxAdapter customAdapter;
+
+    public Combobox(Context context, Vector<String> choices, boolean addEmptyFirstChoice,
+                    boolean permissive, int fontSize) {
```
ctsims: It seems very, very strange for us to pass in fontSize here. The AutoCompleteTextView should have font size as part of its parameter set already, it feels like we shouldn't be moving those style params directly into our code if there are other ways to convey them

app/src/org/commcare/views/widgets/ComboboxWidget.java
```diff
@@ -0,0 +1,123 @@
...
+    private void addListeners() {
+        comboBox.setOnItemClickListener(new AdapterView.OnItemClickListener() {
+            @Override
+            public void onItemClick(AdapterView<?> parent, View view, int position, long id) {
+                widgetEntryChanged();
+            }
+        });
+
+        comboBox.addTextChangedListener(new TextWatcher() {
+            @Override
```
ctsims: Does mean there's no signal for "the user typing out a correct entry, not clicking on the entry in the list, and then clicking on the next question" which results in the value being set? Clicking the ">" or swiping would do it, since that triggers a "get answer", but I wonder if, say, losing focus should also attempt to set it?

app/src/org/commcare/views/widgets/ComboboxWidget.java
```diff
@@ -0,0 +1,123 @@
...
+        }
+    }
+
+    @Override
+    public void clearAnswer() {
+        comboBox.setText("");
+    }
+
+    @Override
+    public void setFocus(Context context) {
```
ctsims: I think this should set focus to the textbox

app/src/org/commcare/views/Combobox.java
amstone326: @ctsims So the reason I did this is because I was copying what `SpinnerWidget` does here: https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/views/widgets/SpinnerWidget.java#L52. It ends up looking pretty bad to just use the AutoCompleteTextView's native font size, because it's way smaller than the other text on the screen. But I agree that it looks out-of-place from the perspective of the `Combobox` view's code, which doesn't know anything about widget world. Is it sufficient in your mind to change the constructor to not take in font size, and then add a static constructor method called like `createComboboxForUseInWidget()` that does set the font size?

app/src/org/commcare/views/widgets/ComboboxWidget.java
amstone326: There's actually an important reason I didn't do this, but it's surprisingly convoluted to explain in writing. I'll come explain in person when you have a chance

app/src/org/commcare/views/widgets/ComboboxWidget.java
amstone326: (Resolved by adding a call to `widgetEntryChanged()` in the onFocusListener

amstone326:

amstone326:
@ctsims Ok so bad news is I've made a fair amount of changes since you last reviewed (on top of the changes you asked for). Good news is this should make it easier for Kriti and Shayoni to figure out what they want. I've set it up now so that it's much easier to add new implementations of `ComboboxAdapter` with a custom filter, so that I can just give them a way to try out different filtering systems and see what they like better (switching between them is just a matter of changing the appearance attribute). Right now I'm going to ask them to try out 2 options:

1. `MultiWordComboboxAdapter` -- The same as what I was calling `PermissiveComboboxAdapter` before.

2. `FuzzyMatchComboboxAdapter` -- Same as the multi-word adapter, but checks for fuzzy matches in addition to direct matches. If you are using this adapter, the as-you-type corrections are turned off; I realized that since fuzzy match only works once the number of typed letters reaches a certain threshold and is close to the number of letters in the comparison string, it doesn't make any sense to restrict typing here.

amstone326:

app/src/org/commcare/adapters/ComboboxAdapter.java
```diff
@@ -0,0 +1,130 @@
...
+ * @author Aliza Stone
+ */
+public class ComboboxAdapter extends ArrayAdapter<String> {
+
+    private float customTextSize;
+    protected final String[] allChoices;
+    protected String[] currentChoices;
+    protected ComboboxFilterRule filterRule;
+
+    public ComboboxAdapter(final Context context, final String[] objects, FilterType type) {
```
ctsims: It seems odd to me that we wouldn't just pass in the filter class itself, since the classes don't require a constructor. Isn't that the general android pattern? IE: https://developer.android.com/reference/android/widget/TextView.html#setFilters(android.text.InputFilter[])

app/src/org/commcare/views/EntityView.java
```diff
@@ -482,7 +482,8 @@ public static Spannable highlightSearches(String[] searchTerms, Spannable raw,
                             int indexInDisplay = normalizedDisplayString.indexOf(currentSpan);
                             int length = (curEnd - curStart);
 
-                            if (indexInDisplay != -1 && StringUtils.fuzzyMatch(currentSpan, searchText).first) {
+                            if (indexInDisplay != -1 &&
+                                    org.commcare.cases.util.StringUtils.fuzzyMatch(currentSpan, searchText).first) {
```
ctsims: It seems weird that this got moved to this line, rather than the import changing?

app/src/org/commcare/views/widgets/ComboboxWidget.java
```diff
@@ -0,0 +1,163 @@
...
+        });
+    }
+
+    private void fillInPreviousAnswer(FormEntryPrompt prompt) {
+        if (prompt.getAnswerValue() != null) {
+            String previousAnswerValue = ((Selection)prompt.getAnswerValue().getValue()).getValue();
+            for (int i = 0; i < choices.size(); i++) {
+                if (choices.get(i).getValue().equals(previousAnswerValue)) {
+                    comboBox.setText(choiceTexts.get(i));
+                    // setting this text will cause the dropdown to open up, which we don't want
```
ctsims: This came up in a couple of places and seems pretty awkward/unintuitive to me. I wonder if it might be worth putting the behavior more explicitly into the comboBox somehow. Is this a general awkwardness around the AutoCompleteTextView in general?

app/src/org/commcare/views/widgets/ComboboxWidget.java
```diff
@@ -0,0 +1,163 @@
...
+            return null;
+        } else {
+            return new InvalidData("The text entered is not a valid answer choice",
+                    new SelectOneData(new Selection(enteredText)));
+        }
+    }
+
+    @Override
+    public void clearAnswer() {
+        comboBox.setText("");
```
ctsims: Doesn't this also need the "dismissDropDown" call?

app/src/org/commcare/adapters/ComboboxAdapter.java
amstone326: Yep you're totally right, the structure of passing in the enum was left over from a version where I had many subclasses of `ComboboxAdapter` instead of the filter classes, and those did have a bunch of arguments to their constructor. Will change.

app/src/org/commcare/views/EntityView.java
amstone326: It's because the android version of StringUtils is also being imported in this class so you can't do both as a named import.

app/src/org/commcare/views/widgets/ComboboxWidget.java
amstone326: Ok I swear that I did this for a reason but when I just re-tested it now, I was able to remove both of the places that I was calling this without issue... Ghosts

amstone326:

amstone326:
@ctsims Addressed your remaining comments!

amstone326:
Release Note: Add appearance attributes for single select questions that will display the question as a combo box (combined text entry and drop-down list)

PR #1621, opened 2017-01-25
Accuracy string fix

wpride:
Fixes http://manage.dimagi.com/default.asp?246312

We were including the location provider in the args even though this wasn't expected in the translation string

I'm not clear from git blame where this got introduced - as far as I can tell neither the translation string nor this line have changed since 2.32. Very odd.


amstone326:
Probably has something to do with https://github.com/dimagi/commcare-android/pull/1606, although I can't immediately see how

amstone326:
Or maybe it's just been like this for a little bit? It's small enough that it could feasibly have been missed in QA and not reported by a user in the 1 release that it was present in.

PR #1622, opened 2017-01-26
Query Planner

ctsims:
Very simply query planner which give us a place to migrate optimizations we've been using for predicate evaluation and make informed decisions about which to use/not use when evaluating complex queries.

Also provides a Query Profiler which makes it possible to understand how a query was executed, and profile the runtime of that evaluation.

cross-request: https://github.com/dimagi/commcare-core/pull/511


ctsims:

ctsims:

ctsims:

PR #1627, opened 2017-02-03
Install app from list of available apps for a user

amstone326:
Allow users to install CommCare apps by submitting a request to the server for all apps that their user (mobile or web) has access to, and then simply selecting an app from that list. Once a user has submitted a request for available apps with some user on a given device, that list will persist until they choose to "See apps for a different user" in the options menu on that screen.

Question: This is currently accessible only via the options menu on the normal install screen (so in the same place as offline install). I'm not sure whether this is something we want to add a full install method button for yet (like the "Scan Application Barcode" and "Enter Code"), or just leave it there for now.

Screenshots below!!

cross-request: https://github.com/dimagi/commcare-core/pull/516


amstone326:

amstone326:
Screenshots:

screenshot_20170324-113426
screenshot_20170324-113430
screenshot_20170324-113346

app/res/values/strings.xml
```diff
@@ -49,7 +49,7 @@
     <string name="privilege_not_enabled_text">Special privileges are not enabled on your device.</string>
     <string name="privilege_claim_failed">Could not claim privileges</string>
     <string name="privilege_instructions">If you have a superuser account on HQ, you can enable certain privileges on your device by going to the \'Enable Privileges on Mobile\' tab of your \'My Account Settings\' page, and then scanning the barcode that appears there.</string>
-
+    <string name="see_available_apps">See Available Apps</string>
```
wpride: is the reason for having this here instead of in `android_startup_strings` that you want to reference this directly from a layout XML file?

app/src/org/commcare/activities/CommCareSetupActivity.java
```diff
@@ -384,6 +387,12 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
                     result = data.getStringExtra(InstallArchiveActivity.ARCHIVE_JR_REFERENCE);
                 }
                 break;
+            case GET_APPS_FROM_HQ:
```
wpride: Should there be error messaging (Toast notification) in the event of the request failing?

app/src/org/commcare/activities/PostRequestActivity.java
```diff
@@ -110,7 +110,7 @@ private void makePostRequest() {
         if (!hasTaskLaunched && !inErrorState) {
             SimpleHttpTask postTask;
             try {
-                postTask = new SimpleHttpTask(this, url, params, true);
+                postTask = new SimpleHttpTask(this, url, params, true, null);
```
wpride: Maybe just my own stylistic taste but I usually make an overloaded method when I find myself writing the same defaults twice as here and https://github.com/dimagi/commcare-android/pull/1627/files#diff-372dfe29c871df8f1f5baca49b82907eR182

app/src/org/commcare/network/HttpUtils.java
```diff
@@ -1,9 +1,9 @@
 package org.commcare.network;
 
 import android.content.SharedPreferences;
-import android.support.v4.util.Pair;
 
 import org.commcare.CommCareApplication;
+import org.commcare.modern.util.Pair;
```
wpride: Was this intentional?

app/src/org/commcare/preferences/GlobalPrivilegesManager.java
```diff
@@ -76,4 +94,42 @@ private static String getPrivilegeDisplayName(String privilegeName) {
         }
     }
 
+    public static void storeRetrievedAvailableApps(Vector<AppAvailableForInstall> availableAppsRetrieved) {
+        if (availableAppsRetrieved != null && availableAppsRetrieved.size() > 0) {
+            ByteArrayOutputStream baos = new ByteArrayOutputStream();
```
wpride: Should wrap this whole thing in a `try` and close these in a `finally`

app/src/org/commcare/preferences/GlobalPrivilegesManager.java
```diff
@@ -76,4 +94,42 @@ private static String getPrivilegeDisplayName(String privilegeName) {
...
+            DataOutputStream serializedStream = new DataOutputStream(baos);
+            try {
+                ExtUtil.write(serializedStream, new ExtWrapList(availableAppsRetrieved));
+                String serializedAppsList = Base64.encodeToString(baos.toByteArray(), Base64.DEFAULT);
+                getGlobalPrivilegesRecord().edit()
+                        .putString(RETRIEVED_AVAILABLE_APPS, serializedAppsList).commit();
+                serializedStream.close();
+            } catch (IOException e) {
```
wpride: error handling?

app/src/org/commcare/preferences/GlobalPrivilegesManager.java
```diff
@@ -76,4 +94,42 @@ private static String getPrivilegeDisplayName(String privilegeName) {
         }
     }
 
+    public static void storeRetrievedAvailableApps(Vector<AppAvailableForInstall> availableAppsRetrieved) {
```
wpride: Seems like these don't necessarily fit in with the `GlobalPriviledgesManager` class (and ideally would be stored in a SQLite DB rather than the SharedPreferences?) - but seems fine for a temporary, low-touch home

app/res/values/strings.xml
amstone326: Yea

app/src/org/commcare/activities/CommCareSetupActivity.java
amstone326: There is lots of error messaging for if the request fails in the `InstallFromListActivity` itself. If they've gotten to this point, it either means that the request succeeded and the InstallFromListActivity was automatically finished after they selected 1 of the apps, or they manually pressed back without selecting an app, and thus we don't need/want to do anything here.

app/src/org/commcare/activities/PostRequestActivity.java
amstone326: Sorry I'm not sure what you mean here. This line and the line you linked to are calling the same constructor with the same set of arguments. What would be the reason for overloading anything?

app/src/org/commcare/network/HttpUtils.java
amstone326: Yes; I forget the exact chain of code that made this necessary, but I changed which type of `Pair` I was using somewhere else (because `android.support.v4.util.Pair` isn't accessible in `core`), and there was some reason that it was cleaner/easier to have the type here match.

app/src/org/commcare/preferences/GlobalPrivilegesManager.java
amstone326: Merp yea you're probably right about the DB thing, not sure why that didn't occur to me. I should probably do that refactor now, I think it would be pretty messy to release it this way and then try to migrate to that later.

amstone326:
@wpride Responded and/or made the requested changes. The switch to using a DB for storing the apps is gonna be a much larger time investment though. Would you be open to merging this set of PRs (once all other issues are addressed) and I'll make that change in a new PR?

amstone326:

amstone326:
Annouce to field-dev: Added the ability to install an app from a list of apps that are available to your user, which you can retrieve as either a mobile or web user. Once the list has been retrieved once, it persists on the device until you uninstall CommCare, providing quick and easy access to all of the apps you might want to install, without having to scan a barcode.

PR #1630, opened 2017-02-06
Add option to update to latest saved state of an app

amstone326:
In follow-up to Cal's hackathon work https://github.com/dimagi/commcare-hq/pull/14737, add the option to update to the latest saved state of an app.

I chatted with @orangejenny for a bit today on the best way to do the UI for this, taking into account the fact that we didn't want to confuse existing workflows but also wanted it to be semi-discoverable. We settled on putting a menu item called "Update Options" in 2 places: one in the main "Settings" menu (as a replacement for the "Enable Updating to Un-Starred Builds" setting that is currently there) and one in the options menu of the Update Activity (since that is the most logical place that someone might look for it). Screenshots of both of these are below. The UIs of the 2 menus are slightly different because 1 is part of a native Android preference screen and 1 isn't, but I'm inclined not to worry about that.

UI from main settings screen:
screenshot_20170206-152141
screenshot_20170206-152144

UI from update activity:
screenshot_20170206-152041
screenshot_20170206-152050

@orangejenny What are your thoughts on the "NEW!" text I added next to the saved state option? I sort of like it but don't have any basis for whether that's a good practice or not. I wish we had a convenient way to link to documentation from mobile, but I can't think of anything.

### DIFF
--- a/app/assets/locales/android_translatable_strings.txt
+++ b/app/assets/locales/android_translatable_strings.txt
@@ -69,6 +69,10 @@ home.menu.pin.change=Change My PIN
 
 menu.update.from.ccz=Offline Update
 menu.update.from.hub=Update from Local Hub
+menu.update.options=Update Options
+update.option.starred=Latest starred build
+update.option.build=Latest build
+update.option.saved=Latest saved state (NEW!)
 
 home.developer.options.enabled=Developer Mode Enabled
 
--- a/app/res/values/strings.xml
+++ b/app/res/values/strings.xml
@@ -142,6 +142,16 @@ Some images in CommCare are graciously used under Creative Commons Licensing fro
         <item>400</item>
         <item>480</item>
     </string-array>
+    <string-array name="update_target_labels">
+        <item>Latest starred build</item>
+        <item>Latest build</item>
+        <item>Latest saved state (NEW!)</item>
+    </string-array>
+    <string-array name="update_target_vals">
+        <item>release</item>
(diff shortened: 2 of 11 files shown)


amstone326:
Note to self to make confluence docs page visible once this is released: https://confluence.dimagi.com/display/commcarepublic/Update+Target+Options

amstone326:
Release Note: I don't think this actually warrants a release note but using this as a way to remind us to toggle the visibility of [this confluence page](https://confluence.dimagi.com/display/commcarepublic/Update+Target+Options) once this is released.

orangejenny:
> What are your thoughts on the "NEW!" text I added next to the saved state option?

I don't think it has much purpose, because I doubt many seasoned users spend much time in settings, so I doubt it'll be noticed. This is generally a hard-to-discover feature (which is fine, it isn't applicable to many users, I just don't think pointing out that it's new makes much difference). But as long as it isn't there forever, I don't strongly object.

Agree that it would be cool for mobile to be able to link to more info - which seems possible, it could link to a confluence page that opened in a browser - but could also be confusing, so let's not do that (at least not without giving it more thought).

orangejenny:
Actually, after thinking for another 3 seconds, this seems like something that could/should go into the product update and maybe the HQ bell announcement (might be too small a feature for the latter). Which is to say, I'm thinking of this as an HQ feature, not a mobile feature, since it's aimed at app builders, and should be announced through the usual HQ channels.

PR #1634, opened 2017-02-10
Show notification message for all update errors

amstone326:
Somehow the error handling for updates had gotten set up such that we were only creating a notification message for certain types of failures, even though the primary error message would always say "(See notification for details)". By moving the exception-handling up the chain from `AndroidResourceManager.checkAndPrepareUpgradeResources` to `UpdateTask.doInBackground`, we can retain access to the exception error messages in order to use them in a notification. 

(There's no ticket for this, Mike O just reported the behavior to me)

### DIFF
--- a/app/unit-tests/src/org/commcare/android/tests/activities/UpdateActivityTest.java
+++ b/app/unit-tests/src/org/commcare/android/tests/activities/UpdateActivityTest.java
@@ -50,7 +50,8 @@ public class UpdateActivityTest {
      */
     @Test
     public void invalidUpdateTest() {
-        String invalidUpdateReference = "jr://resource/commcare-apps/update_tests/invalid_suite_update/profile.ccpr";
+        String invalidUpdateReference =
+                "jr://resource/commcare-apps/update_tests/invalid_suite_update/profile.ccpr";
 
         // start the update activity
         Intent updateActivityIntent =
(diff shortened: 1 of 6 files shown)


app/unit-tests/src/org/commcare/android/tests/activities/UpdateActivityTest.java
```diff
@@ -64,27 +65,33 @@ public void invalidUpdateTest() {
...
         shadowActivity.clickMenuItem(UpdateActivity.MENU_UPDATE_FROM_CCZ);
 
-        // make sure there are no pinned notifications
+        // Make sure there are no pinned notifications before we start
         NotificationManager notificationManager =
-                (NotificationManager)RuntimeEnvironment.application.getSystemService(Context.NOTIFICATION_SERVICE);
-        Notification notification = Shadows.shadowOf(notificationManager).getNotification(CommCareNoficationManager.MESSAGE_NOTIFICATION);
+                (NotificationManager)RuntimeEnvironment.application
+                        .getSystemService(Context.NOTIFICATION_SERVICE);
+        notificationManager.cancel(CommCareNoficationManager.MESSAGE_NOTIFICATION);
```
amstone326: I had to add this line to prevent this test from failing at line 75. I think this is fine because all this seems to be trying to do is to make sure that the notifications bar is clear before we start, so that when we test for `assertNotNull(notification)` at line 88, we know that it represents the new notification we just caused and not something that was pre-existing. My assumption is that the change I made in this PR caused a test that runs prior to this one to spawn a notification where it wasn't before, and that's why the behavior changed. Let me know if anyone sees a problem with this assessment, though.

wpride: Nice catch, sounds very unfortunate - usually I'd expect the test runner to give you a clean sandbox for each run, though maybe the DVM spin-up makes that unrealistic for Roboelectric. If we run into this again maybe we should genericize what you've done to run between each test

PR #1635, opened 2017-02-14
Force full-screen graphs to landscape mode

orangejenny:
@amstone326

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -208,7 +208,8 @@
             android:windowSoftInputMode="stateUnchanged|adjustResize">
         </activity>
         <activity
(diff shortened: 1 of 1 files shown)


amstone326:
@orangejenny Just a note that you can also do this programmatically, if you need it to be conditional in any way: https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/activities/EntitySelectActivity.java#L171. But seems right to do it in the manifest here if you want it to always be that way.

orangejenny:
Cool, good to know.

orangejenny:
updated, thanks for the help: https://github.com/dimagi/commcare-android/pull/1659

PR #1636, opened 2017-02-16
Error correctly and remove record when files missing

wpride:
Previously we failed quietly and did nothing when a FormRecord's files were missing, blocking submission. This removes the files and gives a descriptive error. Addresses http://manage.dimagi.com/default.asp?246572. 

@ctsims @amstone326 I wasn't sure whether quarantining or removing was more appropriate. This seems to fall more under the category of "malformed data" which we remove for, but perhaps keeping the metadata around would be useful. LMK what you think.


amstone326:
@wpride  you said "This removes the files and gives a descriptive error." I don't see where you're removing anything, it looks like the only thing that changed here is showing a notification message. Is there code missing or am I misunderstanding something?

app/src/org/commcare/utils/FormUploadUtil.java
```diff
@@ -129,6 +129,7 @@ public static FormUploadResult sendInstance(int submissionNumber, File folder,
         if (files.length == 0) {
             Log.e(TAG, "no files to upload");
             listener.cancel(true);
```
amstone326: Doesn't this mean that we're still canceling the `ProcessAndSendTask` when this error is encountered? Is that what we want? (Possible that it is, I just couldn't tell for sure from the discussion).

app/src/org/commcare/utils/FormUploadUtil.java
```diff
@@ -129,6 +129,7 @@ public static FormUploadResult sendInstance(int submissionNumber, File folder,
...
+            throw new FileNotFoundException("Folder at path " + folder.getAbsolutePath() + " had no files.");
```
wpride: @amstone326 throwing this exception results in the file being deleted here: https://github.com/dimagi/commcare-android/blob/20fbbe174fb9b9d47729665dec858de96ce37563/app/src/org/commcare/tasks/ProcessAndSendTask.java#L288-L288

app/src/org/commcare/utils/FormUploadUtil.java
```diff
@@ -129,6 +129,7 @@ public static FormUploadResult sendInstance(int submissionNumber, File folder,
         if (files.length == 0) {
             Log.e(TAG, "no files to upload");
```
amstone326: Might be nice to make this log message more specific

amstone326:
@wpride @ctsims Clearly agree that this PR represents a better state than what we're currently in, but my high-level concern is what happens from a data sensibility standpoint when we blindly wipe this form record and proceed with submitting the rest. What if, for instance, the record we wiped was the registration form for a case that has a bunch of queued update forms? Are we just accepting that this is the best we can do and that it will least be apparent in the case data on HQ that something went wrong when the form is missing?

wpride:
@amstone326 Agreed that this state is terrible but I think submitting what we have to HQ as we currently do is about the best we can do. If the form has been deleted or corrupted then there's nothing to be done in terms of saving the original data. HQ seems like a much better place to attempt to patch together what happened and re-create any lost data.

I do think it'd be useful to have much better and more immediate visibility when we do something so dire/irreversible as wipe a form record. It seems wrong that these sorts of exceptions are logged in the same place and manner as, say, a connection timeout. I'd be interested in adding some logging flag that causes mobile devs to receive emails (similar to HQ 500's) when the logs are processed on HQ.

amstone326:
@wpride I strongly agree with that last point. I was thinking about this recently because I've been combing through device logs a lot for a combination of this ICDS bug and the weird missing registration form bug, and thinking that it is really, really not good that it's not easier to separate out the logs that mean something bad happened from the bazillion other logs that are just recording normal procedural behaviors (I'm not even sure why we log those things). And doing something assertive like an email about especially bad things also seems really useful. I would be in favor of resourcing someone to do something about this asap, I think the ROI will be huge in terms of time saved in debugging weird scary issues like these.

amstone326:
**NOTE**: This needs to go into LTS as well

app/src/org/commcare/tasks/ProcessAndSendTask.java
```diff
@@ -289,10 +289,13 @@ private void sendForms(FormRecord[] records) {
                         if (CommCareApplication.instance().isStorageAvailable()) {
                             //If storage is available generally, this is a bug in the app design
                             Logger.log(AndroidLogger.TYPE_ERROR_DESIGN, "Removing form record because file was missing|" + getExceptionText(e));
```
amstone326: @wpride Once your new logging PR is merged, can we change this log tag to be the new FORM_SUBMISSION one? It would be really helpful to have these all show up under 1 filter.

amstone326:
@ctsims Hoping to get your input on a question from above: Is it ok for us to just blindly wipe a form record when this happens, and then still proceed with submitting the rest? It seems like that could lead to the sort of bugs we've been encountering lately where there's a case on HQ that has update forms but is missing a registration form (for example).

app/src/org/commcare/tasks/ProcessAndSendTask.java
amstone326: (fyi just did this)

ctsims:
@wspride It looks like this is actually reflective of the original intent of the system anyway, so seems fine for us to do it for now.

Can we make one big change, though? Can we add a Soft Assert for this, and can we add the form/record ID if it exists to the message? That will let us match up the missing record with the new "saved form" record from the other PR.

wpride:
@amstone326 note minor change to log content to include record Id

PR #1637, opened 2017-02-17
Better logging around form submission and completion

wpride:
Re: http://manage.dimagi.com/default.asp?247810#1289890

cc @snopoke

### DIFF
--- a/app/src/org/commcare/provider/InstanceProvider.java
+++ b/app/src/org/commcare/provider/InstanceProvider.java
@@ -520,7 +520,8 @@ public class InstanceProvider extends ContentProvider {
             return;
         }
 
-        Logger.log(AndroidLogger.TYPE_FORM_ENTRY, "Form Entry Completed");
+        Logger.log(AndroidLogger.TYPE_FORM_ENTRY,
+                String.format("Form Entry Completed for record with id %s", current.getInstanceID()));
 
         // The form is either ready for processing, or not, depending on how it was saved
         if (complete) {
--- a/app/src/org/commcare/tasks/ProcessAndSendTask.java
+++ b/app/src/org/commcare/tasks/ProcessAndSendTask.java
@@ -272,6 +272,8 @@ public abstract class ProcessAndSendTask<R> extends CommCareTask<FormRecord, Lon
                             }
                             results[i] = FormUploadUtil.sendInstance(i, folder, new SecretKeySpec(record.getAesKey(), "AES"), url, this, mUser);
                             if (results[i] == FormUploadResult.FULL_SUCCESS) {
+                                Logger.log(AndroidLogger.TYPE_FORM_SUBMISSION,
+                                        String.format("Successfully submitted form with id %s", record.getInstanceID()));
                                 break;
                             } else {
                                 attemptsMade++;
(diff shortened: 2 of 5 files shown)


snopoke:
Nice, thanks @wpride

amstone326:
@wpride I think it would be helpful to add logs for unsuccessful responses to form submission attempts as well. Otherwise the logs don't have any record of a failure, other than the lack of a success I guess. But seems like it would be useful to be more explicit that it failed and what the response code was.

wpride:
@amstone326 I think unsuccessful attempts are already logged in the submission code (at least something like that appears in the device logs)

app/src/org/commcare/utils/StorageUtils.java
```diff
@@ -110,8 +110,6 @@ public int compare(Integer lhs, Integer rhs) {
                 dateAsSeconds = Long.valueOf(dateAsString);
             } catch (NumberFormatException e) {
                 // Go with the next best ordering for now
-                Logger.log(AndroidLogger.TYPE_ERROR_ASSERTION,
```
ctsims: Hey ,

This definitely stay as a soft assert now, since we've fixed the problem as far as we know.

app/src/org/commcare/tasks/ProcessAndSendTask.java
```diff
@@ -272,6 +272,8 @@ private void sendForms(FormRecord[] records) {
                             }
                             results[i] = FormUploadUtil.sendInstance(i, folder, new SecretKeySpec(record.getAesKey(), "AES"), url, this, mUser);
                             if (results[i] == FormUploadResult.FULL_SUCCESS) {
+                                Logger.log(AndroidLogger.TYPE_FORM_SUBMISSION,
```
ctsims: Flagging that this is going to be _really_ verbose overall. Might be unavoidable, but seems a bit unfortunate.

app/src/org/commcare/provider/InstanceProvider.java
```diff
@@ -520,7 +520,8 @@ private void linkToSessionFormRecord(Uri instanceUri) {
             return;
         }
 
-        Logger.log(AndroidLogger.TYPE_FORM_ENTRY, "Form Entry Completed");
+        Logger.log(AndroidLogger.TYPE_FORM_ENTRY,
+                String.format("Form Entry Completed for record with id %s", current.getInstanceID()));
```
ctsims: Is the instanceID here _our_ instance Id, or is it the ID of the content provider?

app/src/org/commcare/tasks/ProcessAndSendTask.java
snopoke: We already have logs for starting and ending form entry. Won't be more verbose than that will it?

app/src/org/commcare/provider/InstanceProvider.java
wpride: Looks like what we want: 

screen shot 2017-02-22 at 11 55 44 am


02-22 11:53:52.346 14502-14927/org.commcare.dalvik W/System.err: logger> form-entry: Form Entry Completed for record with id 9b2efabe-4ad3-47b2-8835-5b76adbdfd36
02-22 11:53:57.254 14502-14729/org.commcare.dalvik W/System.err: logger> form-submission: Successfully submitted form with id 9b2efabe-4ad3-47b2-8835-5b76adbdfd36

PR #1639, opened 2017-02-17
Prevent updateFormRelevancies() from calling itself downstream

amstone326:
Fix for http://manage.dimagi.com/default.asp?247841. 

This was a fun one. The root of this ended up being the fact that `ComboboxWidget` calls `widgetEntryChanged()` when it loses focus: https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/views/widgets/ComboboxWidget.java#L95. `QuestionWidget.widgetEntryChanged()` triggers `FormEntryActivity.updateFormRelevancies()` downstream, which in turn calls `questionsView.removeQuestionsFromIndex()`. In this case, `removeQuestionsFromIndex()` rightfully removes the `ComboboxWidget` for district choice (it will then get added back later in the method), but removing the widget triggers its `onFocusChangedListener` and thus results in `updateFormRelevancies()` effectively recursing in the middle of its first execution. The method is definitely not designed to handle that, because when it returns from the recursion, the `promptsLeftInView` list that has been built up in the first call is now stale and inaccurate, but it proceeds with using it anyway. Thus, even though the 2 combobox widgets have already been added back to the view once by the recursive call, they get added a second time.

`ComboboxWidget` is the only widget we have which calls `widgetEntryChanged()` from a place like this, which is why this has never come up before. 

@czue I'm assuming this needs a hotfix, let me know if that's not necessarily true.

### DIFF
--- a/app/src/org/commcare/activities/FormEntryActivityUIController.java
+++ b/app/src/org/commcare/activities/FormEntryActivityUIController.java
@@ -75,6 +75,8 @@ public class FormEntryActivityUIController implements CommCareActivityUIControll
     private int indexOfLastChangedWidget = -1;
     private BlockingActionsManager blockingActionsManager;
 
+    private boolean formRelevanicesUpdateInProgress = false;
+
     private static final String KEY_LAST_CHANGED_WIDGET = "index-of-last-changed-widget";
 
     enum AnimationType {
@@ -694,7 +696,13 @@ public class FormEntryActivityUIController implements CommCareActivityUIControll
         activity.findViewById(R.id.multiple_intent_dispatch_button).setVisibility(View.GONE);
     }
 
-    protected void updateFormRelevancies() {
+    protected synchronized void updateFormRelevancies() {
+        if (formRelevanicesUpdateInProgress) {
+            // Don't allow this method to call itself downstream accidentally
+            return;
+        }
+        formRelevanicesUpdateInProgress = true;
+
         ArrayList<QuestionWidget> oldWidgets = questionsView.getWidgets();
         // These 2 calls need to be made here, rather than in the for loop below, because at that
         // point the widgets will have already started being updated to the values for the new view
@@ -751,5 +759,7 @@ public class FormEntryActivityUIController implements CommCareActivityUIControll
             }
         }
         updateCompoundIntentButtonVisibility();
+
+        formRelevanicesUpdateInProgress = false;
     }
 }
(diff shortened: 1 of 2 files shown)


czue:
nice, thanks for the super fast fix! :racing_car: 

confirming it would be great if we can hotfix on 2.33.

czue:
> confirming it would be great if we can hotfix on 2.33.

Although probably makes sense to bundle this with [the calendar bug](http://manage.dimagi.com/default.asp?247868) as well if that looks like it will be fixable.

wpride:
Nice catch, one question

amstone326:
@wpride I don't think whatever question you had came through on github

wpride:
Weird, just sent it

amstone326:
@czue Heads up that we just chatted about release timelines and determined that it should be sufficient for enikshay's needs to have these fixes released in 2.34, instead of doing a hotfix. We're code freezing for 2.34 this Wednesday, so based on how long QA has been taking lately (now that most of it is automated) that should get released before March 6th, easily. Let me know if you see any issues with that plan, though.

app/src/org/commcare/activities/FormEntryActivityUIController.java
```diff
@@ -694,7 +696,13 @@ protected void hideCompoundIntentCalloutButton() {
         activity.findViewById(R.id.multiple_intent_dispatch_button).setVisibility(View.GONE);
     }
 
-    protected void updateFormRelevancies() {
+    protected synchronized void updateFormRelevancies() {
+        if (formRelevanciesUpdateInProgress) {
```
amstone326: Hey, really glad you caught that, the `synchronized` keyword shouldn't be there anymore. That was leftover from debugging attempts when I thought that the duplicate calls might have been coming from another thread, which turned out not to be the case. Removing it now!

amstone326:
@ctsims @wpride One high-level question: When implementing this, I reasoned through the usage of `updateFormRelevancies()` and was able to convince myself with reasonable confidence that there are no instances where the correct behavior of that method _relies_ on being able to call itself downstream. Can either of you confirm that you agree with that? I'm just a little paranoid because this is a pretty crucial method that I'm not sure we have amazing test coverage for.

czue:
> @czue Heads up that we just chatted about release timelines and determined that it should be sufficient for enikshay's needs to have these fixes released in 2.34, instead of doing a hotfix. We're code freezing for 2.34 this Wednesday, so based on how long QA has been taking lately (now that most of it is automated) that should get released before March 6th, easily. Let me know if you see any issues with that plan, though.

thanks for the heads up! this sounds fine to me.

amstone326:
@wpride This should be good to go now, although if you have any thoughts on my question above that would be good too.

wpride:
@amstone326 I agree with your interpretation - updateFormRelevancies looks like it should only be triggered by user triggered events (IE question answers changing) and should not be called by anything downstream of itself

amstone326:
@wpride Ok great, can you merge then?

PR #1641, opened 2017-02-20
Don't request focus when just getting answer

wpride:
Fix for http://manage.dimagi.com/default.asp?247868

getAnswer() shouldn't require getting focus (its frequently called when it should not have focus)


czue:
woop

ctsims:
@wpride I think the reason that setfocus was being called was to update some values on the widget when it was in a bad state (see https://github.com/dimagi/commcare-android/blame/2447ddb4c96bca2c494068f9532cd63d730ab82d/app/src/org/commcare/views/widgets/GregorianDateWidget.java#L210), if we're going to remove it from the getAnswer chain default behavior, we should likely re-add it to the incomplete/invalid answer code paths.

PR #1644, opened 2017-02-22
re-add setFocus to invalid conditions

wpride:
@ctsims updated here. Confirmed this still fixes the initial error reported by Cory but not clear what these setFocuses are meant to address - let me know if there's a specific condition I should be checking.


ctsims:
happy to pull these back out later if we can, I think I am not 100% clear on their necessity, but seems low risk to revert tot he original behavior until we do.

PR #1646, opened 2017-02-24
Move query planning classes into core from android

wpride:
Moving some classes into core for usage in formplayer

cross-request: https://github.com/dimagi/commcare-core/pull/526


wpride:

PR #1648, opened 2017-02-27
Debug Tool: End-User Offline Data Restore 

ctsims:
Created a developer option to allow restoring a custom XML file to make it easier to debug/demonstrate certain issues on device without munging files at the APK level.

Adds an option to Developer Options that lets you pick an XML file from the file system and then runs that restore as if it had come from the server.

### DIFF
--- a/app/res/xml/preferences_developer.xml
+++ b/app/res/xml/preferences_developer.xml
@@ -113,5 +113,9 @@
         android:entryValues="@array/pref_enabled_vals"
         android:key="cc-update-to-latest-saved"
         android:title="Enable Updating to Latest Saved Version"/>
+    <Preference
+        android:title="Restore a custom user XML data file"
+        android:key="cc-custom-restore-doc-location"/>
+
 
 </PreferenceScreen>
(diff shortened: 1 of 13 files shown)


ctsims:
@amstone326 I'm with ya! That was my initial plan

Unfortunately those classes are actually "LocalReferencePullResponse", and they assume you're passing in a CommCare reference (IE: jr://file/something). 

I couldn't unify because the other one is useful for non file references (it's actually used mostly for asset files, not filesystem files), so I couldn't change it to be based on files, and there's no super straightforward way to wire in a file system reference that isn't in our protected area and starts in the (jr://) format.

ctsims:
@amstone326 Could still wire it to re-use more code if needed, though. It didn't seem like it was going to be possible without keeping most of the "volume" of shared code similar (since most of the text is declarations and stuff)

amstone326:
@ctsims Ah no I gotcha, I don't think it's really worth the effort of forcing them to share code then, at least not right now. I just changed the names of the 2 existing classes to what you suggested (`LocalReferencePullResponse`) so that it's clearer why both implementations have to exist. Will merge this now

PR #1650, opened 2017-03-01
Actually use the display text in fillInPreviousAnswer

amstone326:
Addresses https://manage.dimagi.com/default.asp?248507, which was caused by https://github.com/dimagi/commcare-android/pull/1647. I had been using the select choice's _value_ when I meant to be using its display text; I clearly did not test this sufficiently, I'm sorry about that. Also going to give Nick a heads up because the change I made for him won't work properly at all until this is merged.

### DIFF
--- a/app/src/org/commcare/views/widgets/ComboboxWidget.java
+++ b/app/src/org/commcare/views/widgets/ComboboxWidget.java
@@ -18,6 +18,7 @@ import org.javarosa.core.model.data.InvalidData;
 import org.javarosa.core.model.data.SelectOneData;
 import org.javarosa.core.model.data.helper.Selection;
 import org.javarosa.form.api.FormEntryPrompt;
+import org.javarosa.xpath.XPathTypeMismatchException;
 
 import java.util.Vector;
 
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/views/widgets/ComboboxWidget.java
```diff
@@ -115,9 +116,17 @@ public void afterTextChanged(Editable s) {
...
-            if (comboBox.isValidUserEntry(previousAnswerValue)) {
-                comboBox.setText(previousAnswerValue);
+            Selection priorSelection = (Selection)prompt.getAnswerValue().getValue();
+            String previousEnteredText;
+            try {
+                previousEnteredText = prompt.getSelectItemText(priorSelection);
+            } catch (XPathTypeMismatchException e) {
+                // Means that a default value was provided that is not a precise answer choice,
+                // but we can still grab the entered text this way and check if it's a valid entry
+                previousEnteredText = priorSelection.xmlValue;
```
ctsims: Will this ever be expected to be a valid selection? xmlValue here is always the <value> choice, right? I don't think we want to be searching for the <value> body if it's not in the current list of valid choices.

You could imagine a scenerio where the label for a value matched the semantic value of another element, even, in which case this would overwrite that answer.

I think it'd be good to coordinate this logic to be consistent with the other select options if possible, seems odd to have different codepaths for each one.

PR #1652, opened 2017-03-02
fix for when graph Y is null

wpride:
FYI @orangejenny


orangejenny:
Curious what the use case here is. Did something come up for mLabor?

wpride:
@orangejenny sorry should have added a note, this is in reference to https://manage.dimagi.com/default.asp?248652

orangejenny:
Oh, I see. I'd limit this to only run for bar graphs, don't think it makes sense to also do for xy-based graphs. Otherwise this looks reasonable.

PR #1653, opened 2017-03-02
Add new consumer apps to google-services template

amstone326:


amstone326:
@ctsims This is good to go now (and has to be merged before we release 2.34)

PR #1654, opened 2017-03-03
match_parent

wpride:
Fixes https://manage.dimagi.com/default.asp?248138

I *think* this will always do the right thing, still reasoning out all the logic


wpride:
@amstone326 to confirm this plays nicely with badges

amstone326:
Confirmed that this doesn't mess with the look of menu list icons that have a badge, so merging! Thank you so much @wpride for finding the fix!

PR #1661, opened 2017-03-09
Log all app updates 

amstone326:
Prompted by https://manage.dimagi.com/default.asp?248119, where Ray was requesting to be able to see when users have updated to new app versions via the device logs. Seemed harmless to add it so just went for it.

### DIFF
--- a/app/src/org/commcare/activities/UpdateActivity.java
+++ b/app/src/org/commcare/activities/UpdateActivity.java
@@ -422,14 +422,15 @@ public class UpdateActivity extends CommCareActivity<UpdateActivity>
     }
 
     private static void reportAppUpdate() {
+        String updateLogMessage = "Update to app version " + ReportingUtils.getAppBuildNumber();
         try {
             String username = CommCareApplication.instance().getRecordForCurrentUser().getUsername();
-            String updateLogMessage = "User " + username + " updated to app version " +
-                    ReportingUtils.getAppBuildNumber();
-            Logger.log(AndroidLogger.TYPE_USER, updateLogMessage);
+            updateLogMessage += " by user " + username;
         } catch (SessionUnavailableException e) {
             // Must be updating from the app manager, in which case we don't have a current user
         }
+        Logger.log(AndroidLogger.TYPE_RESOURCES, updateLogMessage);
+
     }
 
     private void logoutOnSuccessfulUpdate() {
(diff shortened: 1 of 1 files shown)


amstone326:
@wpride @ctsims Can you think of any reason we shouldn't be logging this to the device logs? Just want to double check before I report back to Ray that I've added this.

wpride:
Seems fine to me. One thing that we should discuss at some point is how we want to use tags. In my opinion updating would seem useful to have under its own tag, but maybe people use these differently?

amstone326:
@wpride Agreed that our tagging system for the device logs is very haphazard, but I think I'm going to leave this PR as-is for now, because I have ambitions to do an overhaul of how the device logs are organized asap anyway.

app/src/org/commcare/activities/UpdateActivity.java
```diff
@@ -416,6 +421,17 @@ private static CustomProgressDialog generateNormalUpdateInstallDialog(int taskId
...
     }
 
+    private static void reportAppUpdate() {
+        try {
+            String username = CommCareApplication.instance().getRecordForCurrentUser().getUsername();
+            String updateLogMessage = "User " + username + " updated to app version " +
+                    ReportingUtils.getAppBuildNumber();
+            Logger.log(AndroidLogger.TYPE_USER, updateLogMessage);
+        } catch (SessionUnavailableException e) {
+            // Must be updating from the app manager, in which case we don't have a current user
```
ctsims: Wouldn't it still be worth reporting in this case, just without the username?

We'll still see it by device ID in the reports

app/src/org/commcare/activities/UpdateActivity.java
```diff
@@ -416,6 +421,17 @@ private static CustomProgressDialog generateNormalUpdateInstallDialog(int taskId
         return dialog;
     }
 
+    private static void reportAppUpdate() {
+        try {
+            String username = CommCareApplication.instance().getRecordForCurrentUser().getUsername();
+            String updateLogMessage = "User " + username + " updated to app version " +
+                    ReportingUtils.getAppBuildNumber();
+            Logger.log(AndroidLogger.TYPE_USER, updateLogMessage);
```
ctsims: Traditionally we use "TYPE_RESOURCES" for logs related to app installs/updates/etc.

PR #1662, opened 2017-03-09
Graph printing + Print from a case detail

amstone326:
cross-request: https://github.com/dimagi/commcare-core/pull/527


amstone326:

app/build.gradle
```diff
@@ -191,6 +192,7 @@ android {
     defaultConfig {
         minSdkVersion 9
         targetSdkVersion 25
+        multiDexEnabled true
```
ctsims: pretty big change to move to multiDex. This is 100% needed for tihs work, right?

app/src/org/commcare/graph/model/GraphData.java
```diff
@@ -63,4 +73,91 @@ public String getConfiguration(String key, String defaultValue) {
...
+            String chartHTML = "<div id='chart'></div>";
+            html.append(
+                    "\n} catch (e) { displayError(e); }</script>" +
+                            "<script type='text/javascript' src='file:///android_asset/graphing/graph." + affix + ".js'></script>" +
+                            "</head>" +
+                            "<body>" + titleHTML + errorHTML + chartHTML + "</body>" +
+                            "</html>");
+        } catch (JSONException e) {
+            e.printStackTrace();
+            throw new RuntimeException(e);
```
ctsims: Shouldn't this be a GraphException?

app/src/org/commcare/print/PrintableDetailField.java
```diff
@@ -0,0 +1,100 @@
...
+            String graphTitle = field.getHeader().evaluate();
+            if ("".equals(graphTitle) || graphTitle == null) {
+                // DO NOT CHANGE THIS -- Workaround to address that fact that a graph will not
+                // render properly if the 'chart-title' <div> is empty
+                graphTitle = " ";
+            }
+            String fullGraphHtml = ((GraphData)evaluatedField).getGraphHTML(graphTitle);
+            this.valueString = fullGraphHtml;
+        } catch (GraphException e) {
+            this.valueString = "";
```
ctsims: This seems like it's going to make this pretty hard to debug if there are any issues

app/src/org/commcare/activities/EntityDetailActivity.java
```diff
@@ -251,4 +263,42 @@ public void onBackPressed() {
...
+                printDetail();
+                return true;
+        }
+        return super.onOptionsItemSelected(item);
+    }
+
+    private void printDetail() {
+        Intent i = new Intent(this, TemplatePrinterActivity.class);
+        i.putExtra(TemplatePrinterActivity.PRINT_TEMPLATE_REF_STRING,
+                detail.getDerivedPrintTemplatePath());
```
ctsims: So for clarity in a tabbed detail view, only the "Parent" detail is currently capable of printing right now, right?

app/build.gradle
amstone326: Yea, the app won't run without it. It seems like adding the jsoup library just took us over the limit. We could try to go through and see if there are any libraries we can remove, but unless we removed a lot, we'd probably just hit this again sometime soon.

app/src/org/commcare/graph/model/GraphData.java
amstone326: Not sure; I just copied this code over from one place to another, so this is how it already was, but that seems right to me so I'll change it.

app/src/org/commcare/activities/EntityDetailActivity.java
amstone326: Yep. mLabour was very clear that the only thing they were interested in was being able to print the entire detail at once, so I just went with implementing that for now.

amstone326:
@ctsims So I'm tempted to disagree about the global template thing. In the short term it actually has to stay in there because the HQ-side work for detail printing isn't done yet, so there's currently no other way to set the template there while we're testing. But there have also been a couple of times in semi-recent memory that that HQ document upload stopped working in some way or another, and it was really nice to have the global template to fall back on. But if you're definitely still opposed to it, I'll make a note to get rid of it once the HQ side is done and we've confirmed that that method of template upload is working consistently.

Other than that, addressed all of your comments here.

app/src/org/commcare/activities/EntityDetailActivity.java
ctsims: cool

app/src/org/commcare/print/PrintableDetailField.java
```diff
@@ -0,0 +1,100 @@
...
+            String graphTitle = field.getHeader().evaluate();
+            if ("".equals(graphTitle) || graphTitle == null) {
+                // DO NOT CHANGE THIS -- Workaround to address the fact that a graph will not
+                // render properly if the 'chart-title' <div> is empty
+                graphTitle = " ";
+            }
+            String fullGraphHtml = ((GraphData)evaluatedField).getGraphHTML(graphTitle);
+            this.valueString = fullGraphHtml;
+        } catch (GraphException e) {
+            this.valueString = "<p>ERROR ENCOUNTERED WHILE TRYING TO PRINT GRAPH</p>";
```
ctsims: We should put the error text in this message, right? So people can figure out what failed?

PR #1663, opened 2017-03-09
Order form submissions by global counter

amstone326:
Changes our form submission ordering strategy to use a simple per-app counter. A `FormRecord` is marked with the next sequential `submissionOrderingNumber` only after it has been marked as completed, thus ensuring that forms are submitted in the order that they were completed by the user in. 

The ordering number for a record is also now included in a device log message for both attempting to submit the form and successfully submitting the form, which should make it easy to tell when a form has gone missing (since the submission numbers will always be sequential with no missing numbers).

I haven't yet been able to verify the results of this in the device logs or submitted forms report on HQ due to the current reporting issues, but I did verify that CommCare attempted to submit forms in the proper order after these changes.

Background here: http://manage.dimagi.com/default.asp?247383

FYI @snopoke

### DIFF
--- a/app/src/org/commcare/activities/HomeScreenBaseActivity.java
+++ b/app/src/org/commcare/activities/HomeScreenBaseActivity.java
@@ -655,10 +655,8 @@ public abstract class HomeScreenBaseActivity<T> extends SyncCapableCommCareActiv
 
             // The form is either ready for processing, or not, depending on how it was saved
             if (complete) {
-                // We're honoring in order submissions, now, so trigger a full
-                // submission cycle
+                current.setFormNumberForSubmissionOrdering();
                 checkAndStartUnsentFormsTask(false, false);
-
                 refreshUI();
 
                 if (wasExternal) {
@@ -667,8 +665,8 @@ public abstract class HomeScreenBaseActivity<T> extends SyncCapableCommCareActiv
                     return false;
                 }
 
-                // Before we can terminate the session, we need to know that the form has been processed
-                // in case there is state that depends on it.
+                // Before we can terminate the session, we need to know that the form has been
+                // processed in case there is state that depends on it.
                 boolean terminateSuccessful;
                 try {
                     terminateSuccessful = currentState.terminateSession();
--- a/app/src/org/commcare/activities/HomeScreenBaseActivity.java
+++ b/app/src/org/commcare/activities/HomeScreenBaseActivity.java
@@ -656,6 +656,7 @@ public abstract class HomeScreenBaseActivity<T> extends SyncCapableCommCareActiv
             // The form is either ready for processing, or not, depending on how it was saved
             if (complete) {
                 current.setFormNumberForSubmissionOrdering();
+                CommCareApplication.instance().getUserStorage(FormRecord.class).write(current);
                 checkAndStartUnsentFormsTask(false, false);
                 refreshUI();
 
--- a/app/src/org/commcare/activities/HomeScreenBaseActivity.java
+++ b/app/src/org/commcare/activities/HomeScreenBaseActivity.java
@@ -56,6 +56,7 @@ import org.commcare.utils.ChangeLocaleUtil;
 import org.commcare.utils.EntityDetailUtils;
 import org.commcare.utils.GlobalConstants;
 import org.commcare.utils.SessionUnavailableException;
+import org.commcare.utils.StorageUtils;
 import org.commcare.utils.UriToFilePath;
 import org.commcare.views.UserfacingErrorHandling;
 import org.commcare.views.dialogs.CommCareAlertDialog;
@@ -657,7 +658,7 @@ public abstract class HomeScreenBaseActivity<T> extends SyncCapableCommCareActiv
             if (complete) {
                 // Now that we know this form is completed, we can give it the next available
                 // submission ordering number
-                current.setFormNumberForSubmissionOrdering();
+                current.setFormNumberForSubmissionOrdering(StorageUtils.getNextFormSubmissionNumber());
                 CommCareApplication.instance().getUserStorage(FormRecord.class).write(current);
                 checkAndStartUnsentFormsTask(false, false);
                 refreshUI();
--- a/app/src/org/commcare/activities/HomeScreenBaseActivity.java
+++ b/app/src/org/commcare/activities/HomeScreenBaseActivity.java
@@ -655,6 +655,8 @@ public abstract class HomeScreenBaseActivity<T> extends SyncCapableCommCareActiv
 
             // The form is either ready for processing, or not, depending on how it was saved
             if (complete) {
+                // Now that we know this form is completed, we can give it the next available
+                // submission ordering number
                 current.setFormNumberForSubmissionOrdering();
                 CommCareApplication.instance().getUserStorage(FormRecord.class).write(current);
                 checkAndStartUnsentFormsTask(false, false);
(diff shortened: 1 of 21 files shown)


app/src/org/commcare/models/database/SqlStorage.java
```diff
@@ -106,18 +103,14 @@ public SqlStorage(String table, Class<? extends T> ctype, AndroidDbHelper helper
         return fillIdWindow(c, DatabaseHelper.ID_COL, returnSet);
     }
 
-    public static Vector<Integer> fillIdWindow(Cursor c, String columnName) {
```
amstone326: unused

app/src/org/commcare/models/database/user/UserDbUpgradeUtils.java
```diff
@@ -0,0 +1,161 @@
...
+                Logger.log(AndroidLogger.TYPE_ERROR_ASSERTION,
+                        "Invalid date in last modified value: " + dateAsString);
+                idToDateIndex.put(id, (long)id);
+                continue;
+            }
+            idToDateIndex.put(id, dateAsSeconds);
+        }
+        return idToDateIndex;
+    }
+}
```
snopoke: nice refactor

app/src/org/commcare/utils/StorageUtils.java
```diff
@@ -63,9 +62,10 @@ public static int getNumIncompleteForms() {
             return new FormRecord[0];
         }
 
-        // Order ids so they're submitted to and processed by the server in
-        // the correct order.
-        sortRecordsByDate(ids, storage);
+        if (sortRequired) {
+            // Order ids so they're submitted to and processed by the server in the correct order.
+            sortRecordsBySubmissionOrderingNumber(ids, storage);
```
snopoke: What the reason behind fetching and sorting the ID's first before actually fetching the records?

amstone326: Just so that in the for loop below, by reading out the records from storage in order of the sorted IDs, we're making sure that the `FormRecord[]` that this method returns is also in that same sorted order.

snopoke: I mean why not fetch all the records first and then sort them? Currently there are 2 storage operations per form:
* get ordering number
* get form record

Just wondering if that could be reduced to 1 if the form records were fetched up front. 

Perhaps just my understanding of how the storage works is wrong.

amstone326: Ah gotcha, I see now! You're right, I have no idea why it was like that, I was just blindly maintaining the code structure that had been there when we were sorting by last modified date, but it does seem silly. I just changed this in the most recent commit (2df14d2). @ctsims Let me know if there was a reason for it being done the old way, but I'm guessing it was just a result of what util methods already existed when someone went to write that code, and nothing more than that.

app/src/org/commcare/models/database/user/UserDbUpgradeUtils.java
ctsims: 👍

app/src/org/commcare/tasks/ProcessAndSendTask.java
```diff
@@ -332,6 +331,22 @@ private void sendForms(FormRecord[] records) {
         }
     }
 
+    private static void logSubmissionAttempt(FormRecord record) {
+        String attemptMesssage = String.format(
+                "Attempting to submit form with id %1$s and submission ordering number %2$s",
```
ctsims: the submission order here feels a little odd to include, is it primarily for debugging the dev work? might not be needed long term (and might introduce some amount of confusion).

app/src/org/commcare/utils/StorageUtils.java
```diff
@@ -49,74 +63,47 @@ public static int getNumIncompleteForms() {
                 new String[]{FormRecord.STATUS_INCOMPLETE, currentAppId}).size();
     }
 
-    public static FormRecord[] getUnsentRecords(SqlStorage<FormRecord> storage) {
+    public static FormRecord[] getUnsentRecordsForCurrentApp(SqlStorage<FormRecord> storage,
```
ctsims: In the past we were only sorting the records conditionally because of how expensive it was to parse out all of the date strings. Given that the sort data is now super simple and included in the record pull already, I'd be super supportive of just always sorting rather than having multiple ways to pull the list.

app/src/org/commcare/activities/HomeScreenBaseActivity.java
```diff
@@ -655,10 +655,11 @@ private boolean processReturnFromFormEntry(int resultCode, Intent intent) {
 
             // The form is either ready for processing, or not, depending on how it was saved
             if (complete) {
-                // We're honoring in order submissions, now, so trigger a full
-                // submission cycle
+                // Now that we know this form is completed, we can give it the next available
+                // submission ordering number
+                current.setFormNumberForSubmissionOrdering();
```
ctsims: If this is the only place we use this, this is where i'd replace this with something like

current.setFormNumberForSubmissionOrdering(DbUtil.getLargestMetaDataValue(FormRecordStorage, FormRecord.META_SUBMISSION_ORDERING_NUMBER)

which feels like it might isolate the dependency a bit more

app/src/org/commcare/android/database/user/models/FormRecord.java
```diff
@@ -67,24 +71,34 @@
...
     @MetaField(META_LAST_MODIFIED)
     private Date lastModified;
+
     @Persisting(7)
     @MetaField(META_APP_ID)
     private String appId;
 
+    @Persisting(value = 8, nullable = true)
+    @MetaField(META_SUBMISSION_ORDERING_NUMBER)
+    private String submissionOrderingNumber;
```
ctsims: why is this being stored as a string which is turned into a number, btw? is that a limitation of the metadata framework?

app/src/org/commcare/tasks/ProcessAndSendTask.java
amstone326: Yea I actually think this is super useful for debugging because the idea is that when logging at this series of logs for any app, the submissions IDs should always be sequential, and if 1 is missing then we know that something went wrong with a form submission.

app/src/org/commcare/utils/StorageUtils.java
amstone326: Cool, I removed the `sortRequired` param

app/src/org/commcare/android/database/user/models/FormRecord.java
amstone326: yea it is

app/src/org/commcare/activities/HomeScreenBaseActivity.java
amstone326: @ctsims Hey, so I just switched to this strategy in the latest commit, but I'm a little hesitant about it. Agreed that it's definitely a cleaner way to make sure we're always using the right value, but it comes at the cost of making each call to `setFormNumberForSubmissionOrdering()` more expensive, since we're now iterating through all the form records in the db each time. I'm holding off on merging this just to hear your thoughts/confirmation on that being a worthwhile trade-off.

ctsims:
@amstone326 Looks great! Also, really appreciate the commit-by-commit changes after feedback, super, super clean and easy to review!

PR #1664, opened 2017-03-10
Graphing Fix

wpride:
Fixes https://manage.dimagi.com/default.asp?248652 and another bug where bar graphs with no entries would not appear

@orangejenny one proposed solution

### DIFF
--- a/app/src/org/commcare/suite/model/graph/Graph.java
+++ b/app/src/org/commcare/suite/model/graph/Graph.java
@@ -174,7 +174,7 @@ public class Graph implements Externalizable, DetailTemplate, Configurable {
                     }
                     String x = s.evaluateX(refContext);
                     String y = s.evaluateY(refContext);
-                    if (x != null && (y == null || graphData.getType().equals(GraphUtil.TYPE_BAR))) {
+                    if (x != null && (y != null || graphData.getType().equals(GraphUtil.TYPE_BAR))) {
                         if (y == null) {
                             y = "0";
                         }
--- a/app/src/org/commcare/suite/model/graph/Graph.java
+++ b/app/src/org/commcare/suite/model/graph/Graph.java
@@ -136,8 +136,6 @@ public class Graph implements Externalizable, DetailTemplate, Configurable {
         }
         for (String key : nonvariables) {
             String value = template.getConfiguration(key).evaluate(context);
-            System.out.println("139 Evaluated key " + key + " configuration "
-                    + template.getConfiguration(key) + " to value " + value);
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/suite/model/graph/Graph.java
```diff
@@ -173,7 +174,10 @@ private void evaluateSeries(GraphData graphData, EvaluationContext context) {
                     }
                     String x = s.evaluateX(refContext);
                     String y = s.evaluateY(refContext);
-                    if (x != null && y != null) {
+                    if (x != null && graphData.getType().equals(GraphUtil.TYPE_BAR)) {
```
orangejenny: This looks like it'll never be true for xy graphs, and therefore never draw any points on xy graphs. I'd expect it to be `x != null && (y == null || graphData.getType().equals(GraphUtil.TYPE_BAR))`

app/src/org/commcare/suite/model/graph/XYSeries.java
```diff
@@ -76,23 +76,35 @@ public void setY(String y) {
...
 
+    public void removeConfiguration(String key) {
+        mConfiguration.remove(key);
+    }
+
     @Override
     public void setConfiguration(String key, Text value) {
-        mConfiguration.put(key, value);
+        if (mPointConfiguration.keySet().contains(key)) {
+            mPointConfiguration.remove(key);
```
orangejenny: Minor, but I think this is redundant; the `put` below will overwrite the current value.

app/src/org/commcare/suite/model/graph/Graph.java
```diff
@@ -149,9 +149,10 @@ private void evaluateSeries(GraphData graphData, EvaluationContext context) {
                 Hashtable<String, Text> pointConfiguration = new Hashtable<>();
                 for (Enumeration e = s.getPointConfigurationKeys(); e.hasMoreElements();) {
                     String key = (String) e.nextElement();
-                    Text value = s.getConfiguration(key);
+                    Text value = s.getPointConfiguration(key);
                     if (value != null) {
                         pointConfiguration.put(key, value);
+                        s.removeConfiguration(key);
```
orangejenny: Not quite following why the key needs to be removed from the configuration...?

amstone326:
@wpride Are you trying to get this into 2.35?

wpride:
@orangejenny comments addressed, would be good to get this into 2.35

orangejenny:
Fixed conditional because I made the wrong suggestion. Good to go.

PR #1665, opened 2017-03-14
Allow app to submit to a separate logging endpoint

ctsims:
https://manage.dimagi.com/default.asp?249286

### DIFF
--- a/app/src/org/commcare/utils/CommCareUtil.java
+++ b/app/src/org/commcare/utils/CommCareUtil.java
@@ -82,8 +82,7 @@ public class CommCareUtil {
     }
 
     public static void triggerLogSubmission(Context c) {
-        SharedPreferences settings = CommCareApplication.instance().getCurrentApp().getAppPreferences();
-        String url = settings.getString(CommCareServerPreferences.PREFS_SUBMISSION_URL_KEY, null);
+        String url = LogSubmissionTask.getSubmissionUrl(CommCareApplication.instance().getCurrentApp().getAppPreferences());
 
         if (url == null) {
             //This is mostly for dev purposes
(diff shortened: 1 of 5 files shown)


app/src/org/commcare/utils/CommCareUtil.java
```diff
@@ -82,8 +82,7 @@ public static int countPreds(TreeReference reference) {
     }
 
     public static void triggerLogSubmission(Context c) {
-        SharedPreferences settings = CommCareApplication.instance().getCurrentApp().getAppPreferences();
-        String url = settings.getString(CommCareServerPreferences.PREFS_SUBMISSION_URL_KEY, null);
+        String url = LogSubmissionTask.getSubmissionUrl(CommCareApplication.instance().getCurrentApp().getAppPreferences());
```
wpride: Looks like this method is always called with `CommCareApplication.instance().getCurrentApp().getAppPreferences()` as the argument - could move into the getSubmissionUrl

ctsims: I feel like it's an antipattern to increase the number of places in the code where a static singleton is requested from a static context, so I felt like it was wrong to do so. If you feel strongly will move it.

wpride:
Interesting, hadn't thought about it that way. Merging

app/src/org/commcare/utils/CommCareUtil.java
amstone326: @ctsims Can you elaborate on that? Definitely not something I currently avoid, would be good to understand why

ctsims: Generally speaking the use of singletons in Java specifically is mostly a way to prevent incredibly annoying and verbose injection of state variables. For us "Application" is more or less persistent across each non-static method call in the app, and if every method had it passed in, things would work normally.

Static methods _wouldn't_ have that dependency, IE: "CommCareApplication._()" wouldn't have existed in their scope before, so to them is an injection of state they didn't have before. It introduces a "secret" dependency that that static method is actually stateful, rather than that it performs some entirely-internal logic (which is what static methods generally can be expected to do).

amstone326: Cool, thanks for explaining, that makes sense.

PR #1666, opened 2017-03-14
Enable Bulk Processing Restore Data

ctsims:
reduces amount of time to process 20k transactions on my phone from around 3 minutes to about 30 seconds.

cross-request: https://github.com/dimagi/commcare-core/pull/537

### DIFF
--- a/app/src/org/commcare/xml/AndroidBulkCaseXmlParser.java
+++ b/app/src/org/commcare/xml/AndroidBulkCaseXmlParser.java
@@ -33,6 +33,7 @@ import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
 import java.util.SortedMap;
@@ -50,12 +51,20 @@ public class AndroidBulkCaseXmlParser extends BulkProcessingCaseXmlParser {
     private final SqlStorage<ACase> storage;
 
     public AndroidBulkCaseXmlParser(KXmlParser parser,
+                                         SqlStorage<ACase> storage,
+                                         HttpRequestEndpoints generator) {
+        this(parser, storage, new EntityStorageCache("case"), new AndroidCaseIndexTable(), generator);
+    }
+
+    public AndroidBulkCaseXmlParser(KXmlParser parser,
                                     SqlStorage<ACase> storage,
+                                    EntityStorageCache entityStorageCache,
+                                    AndroidCaseIndexTable indexTable,
                                     HttpRequestEndpoints generator) {
         super(parser);
         this.generator = generator;
-        mEntityCache = new EntityStorageCache("case");
-        mCaseIndexTable = new AndroidCaseIndexTable();
+        mEntityCache = entityStorageCache;
+        mCaseIndexTable = indexTable;
         this.storage = storage;
     }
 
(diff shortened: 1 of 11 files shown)


app/src/org/commcare/xml/AndroidBulkCaseXmlParser.java
```diff
@@ -0,0 +1,109 @@
...
+                currentOperatingSet.put(c.getCaseId(), c);
+            }
+            db.setTransactionSuccessful();
+        } finally {
+            db.endTransaction();
+        }
+    }
+
+    @Override
+    protected void performBulkWrite(SortedMap<String, Case> writeLog) throws IOException {
```
wpride: Why is this a Map when we only use the values? Performance?

app/src/org/commcare/xml/AndroidTransactionParserFactory.java
```diff
@@ -45,9 +49,10 @@
      * A mapping from an installed form's namespace its install path.
      */
     private Hashtable<String, String> formInstanceNamespaces;
+    private TransactionParserFactory defaultCaseParser;
```
wpride: unused

app/src/org/commcare/xml/AndroidBulkCaseXmlParser.java
ctsims: We want this list to only have one model in it per-write, but it's hard to guarantee that two objects which represent the same model actually will match with equality, so it seemed helpful to be super explicit about there needing to be a unique qualifier for it.

It's kind of awkward, but I think it will prevent errors long-term.

app/src/org/commcare/xml/AndroidBulkCaseXmlParser.java
```diff
@@ -0,0 +1,120 @@
...
+    protected void performBulkWrite(LinkedHashMap<String, Case> writeLog) throws IOException {
+        SQLiteDatabase db;
+        db = getDbHandle();
+        db.beginTransaction();
+        ArrayList<Integer> recordIdsToWipe = new ArrayList<>();
+        try {
+            for(String cid : writeLog.keySet()) {
+                Case c = writeLog.get(cid);
+                storage.write(c);
+                recordIdsToWipe.add(c.getID());
```
amstone326: Can you explain why these record IDs are getting wiped? I don't think I'm fully following what the reason for the 2 new methods below are (`invalidateCaches` and `clearCaseIndices`)

app/src/org/commcare/xml/AndroidBulkCaseXmlParser.java
```diff
@@ -0,0 +1,120 @@
...
+    private final AndroidCaseIndexTable mCaseIndexTable;
+    private final SqlStorage<ACase> storage;
+
+    public AndroidBulkCaseXmlParser(KXmlParser parser,
+                                         SqlStorage<ACase> storage,
+                                         HttpRequestEndpoints generator) {
+        this(parser, storage, new EntityStorageCache("case"), new AndroidCaseIndexTable(), generator);
+    }
+
+    public AndroidBulkCaseXmlParser(KXmlParser parser,
```
ctsims: It's used needed by tests. Still updating the framework to use them (the same way I did in the last one)

app/src/org/commcare/xml/AndroidBulkCaseXmlParser.java
ctsims: They are getting "Wiped" from the case index and entity cache tables in order to get rebuilt. We don't bother "updating" index rows, we just delete all rows for a case and add new ones each time ("update if different" is just annoyingly complicated and not worth the performance gains since it involves a read)

In order to do that with maximal efficiency we do the wipe as one DB operation by putting them in a list.

amstone326: Cool makes sense, thanks!

app/src/org/commcare/xml/AndroidBulkCaseXmlParser.java
amstone326: Ah ok, so should I be holding off on merging this then?

ctsims: Added test coverage

PR #1667, opened 2017-03-15
Don't allow form entry to proceed if uncommitted combobox change is present

amstone326:
Addresses http://manage.dimagi.com/default.asp?249276. Because `widgetEntryChanged()` for the `ComboboxWidget` is NOT called on every keystroke (which would overload the app with constant changes), but instead is only called when focus to the widget changes or the user touches a selection in the dropdown, Kriti was encountering a strange but valid edge case where:

-There were 2 combobox widgets on 1 screen, with the answer choices provided for the 2nd depending on the answer to the 1st
-If you selected answer choices for both, and then changed the answer to the first by pressing backspace but NOT clicking away, AND that change resulted in _another_ valid answer to that question (rather than an invalid one), and then navigated forward, CommCare would allow you to proceed, but the 2nd answer would now have become invalid (since the valid answer choices have changed), but the user would never see that, and would think they had answered the 2nd question, when in fact it will now be blank when the form is submitted.

It may be a little hard to understand what was going on here just from reading that, so I'd recommend following Kriti's repro steps in the ticket (which are really quick) if you want to better understand the situation I was trying to address.

The resolution to this feels a bit hacky, but was the best I could come up with without having to call `widgetEntryChanged()` more frequently. Instead, when saving the answers for the current screen, CommCare checks if there were "uncommitted changes" to any combobox widgets before doing so, and if there were, it calls `updateFormRelevances()` at the end of that method (to "commit" the previously uncommitted change), and also prevents the screen from advancing (so that the user will be able to see that the 2nd combobox widget has now become empty).

### DIFF
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -542,6 +542,11 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
         // group
         boolean success = true;
         if (isEventQuestionOrListGroup()) {
+
+            if (evaluateConstraints) {
+                uiController.questionsView.lockComboboxAnswers();
+            }
+
             HashMap<FormIndex, IAnswerData> answers =
                     uiController.questionsView.getAnswers();
 
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -580,6 +580,7 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
             }
         }
         if (uncommittedChangesWerePresent) {
+            // Lock in the uncommitted changes
             uiController.updateFormRelevancies();
         }
         return success && !uncommittedChangesWerePresent;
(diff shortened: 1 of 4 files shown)


app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -707,10 +712,10 @@ private void saveDataToDisk(boolean exit, boolean complete, String updatedSaveNa
             }
             return;
         }
-        // save current answer; if headless, don't evaluate the constraints
-        // before doing so.
+        // Save current answer; if headless, don't evaluate the constraints before doing so.
+        boolean evaluateConstraints = !headless;
         boolean wasScreenSaved =
-                saveAnswersForCurrentScreen(FormEntryConstants.DO_NOT_EVALUATE_CONSTRAINTS, complete, headless);
+                saveAnswersForCurrentScreen(evaluateConstraints, complete, headless);
```
amstone326: IMPORTANT: This change is separate from the rest of the PR, and I'm not sure whether it's right or not, although it seems like it should be... Previously, it looks like we were never evaluating constraints on `saveAnswersForCurrentScreen()` when actually saving the data to disk. That seemed odd to me, and then when reading the comment, also seemed contrary to the sentiment expressed there. So I changed this to reflect the comment instead. If anyone sees a potential issue with this, please flag.

amstone326:
Closing this for now after chatting with Kriti; gonna keep the branch around in case we decide at a later date that it's necessary

PR #1672, opened 2017-03-22
Fix pause behavior of expanded audio widget

amstone326:
Fix for http://manage.dimagi.com/default.asp?249822.

So I'm honestly not sure how this was ever working, because the progress bar animation was just not being paused when the user pressed the pause button. I'm guessing Phillip was trying to work around having to use `ValueAnimator.pause()` in some way because it requires API level 19, but using `cancel()` instead was just not working at all. So I've made it so that an `ExpandedAudioPlaybackView` can only be instantiated on devices running KitKat or higher, which I think is fine since this is just a nice-to-have UI enhancement thing. On earlier OS versions, we just include the normal audio button instead.


ctsims:

PR #1675, opened 2017-03-29
Don't set bar-color values to null

orangejenny:
http://manage.dimagi.com/default.asp?250307

[Earlier bugfix](https://github.com/dimagi/commcare-android/pull/1664/files#diff-5707293f4ecca257903ea6fad4243d73R54) started using an empty Text to represent "no value", but [the code that evaluates bar-color](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/suite/model/graph/Graph.java#L170-L174) was still attempting to evaluate that Text, resulting in `bar-color` getting set to the string `"null"` for each bar, which then died when later code tried to parse `"null"` as a color.

@amstone326

### DIFF
--- a/app/src/org/commcare/suite/model/graph/Graph.java
+++ b/app/src/org/commcare/suite/model/graph/Graph.java
@@ -170,7 +170,9 @@ public class Graph implements Externalizable, DetailTemplate, Configurable {
                     for (Enumeration e = pointConfiguration.keys(); e.hasMoreElements();) {
                         String key = (String) e.nextElement();
                         String value = pointConfiguration.get(key).evaluate(refContext);
-                        expandedConfiguration.get(key).addElement(value);
+                        if (value != null) {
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/suite/model/graph/Graph.java
```diff
@@ -170,7 +170,9 @@ private void evaluateSeries(GraphData graphData, EvaluationContext context) {
                     for (Enumeration e = pointConfiguration.keys(); e.hasMoreElements();) {
                         String key = (String) e.nextElement();
                         String value = pointConfiguration.get(key).evaluate(refContext);
-                        expandedConfiguration.get(key).addElement(value);
+                        if (value != null) {
```
orangejenny: It'd be cleaner to undo this change and update line 165 above to not add the key to `expandedConfiguration` in the first place if it's just an "empty" Text, but I haven't found a good way to do that (check for "emptiness").

PR #1680, opened 2017-04-04
Increase max heap size

wpride:
These are required for commcare now (so far as I can tell)


amstone326:
@wpride What's the reason for this? Jenkins build seems to be unhappy about it

wpride:
hmm this is a change I have to make locally to get this to build, but looks like that's not universal (and Jenkins doesn't have the memory to accommodate). Closing.

PR #1684, opened 2017-04-12
Use Developer Preferences to control visibility of Update Options setting item

amstone326:
Addresses https://manage.dimagi.com/default.asp?251402

Updated documentation in https://confluence.dimagi.com/display/commcarepublic/Update+Target+Options to instruction users that they will need to enable this Developer Option in CommCare 2.36 and above.

NOTE: This change means that we are explicitly deciding to make the Developer Options something that is public to non-Dimagi users (I plan to clean up [this page](https://confluence.dimagi.com/display/internal/CommCare+Android+Developer+Options) and then move it to the public wiki, as well as link to it from the public wiki page above). @ctsims and I briefly discussed that we may want to start using the mobile privileges framework to allow us to hide certain more-dangerous settings within the Developer Options.

### DIFF
--- a/app/src/org/commcare/preferences/CommCarePreferences.java
+++ b/app/src/org/commcare/preferences/CommCarePreferences.java
@@ -152,8 +152,6 @@ public class CommCarePreferences
 
         GoogleAnalyticsUtils.createPreferenceOnClickListeners(prefMgr, prefKeyToAnalyticsEvent,
                 GoogleAnalyticsFields.CATEGORY_CC_PREFS);
-        // Override the default listener for the print pref key b/c it has extra behavior
-        createPrintPrefOnClickListener(prefMgr);
     }
 
     private void setupUI() {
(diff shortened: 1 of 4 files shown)


app/src/org/commcare/preferences/CommCarePreferences.java
```diff
@@ -227,6 +225,45 @@ public boolean onPreferenceClick(Preference preference) {
         } else {
             getPreferenceScreen().removePreference(analyticsButton);
         }
+
+        Preference printTemplateSetting = findPreference(PREFS_PRINT_DOC_LOCATION);
```
wpride: What was the reason for pulling out this print template functionality from its own method? Was this meant to be included here?

amstone326: Yea it is - once I needed to be calling `setupButtons()` from within `setVisibilityOfUpdateOptionsPref()`, it made me realize that it makes more sense to have the functionality for setting the onClick method for the print template be in that method as well, instead of having it be randomly separate.

amstone326:
Release Note: The "Update Options" setting item in both the main settings menu and Update Activity options menu will now only be visible if you have enabled the "Show Update Options Item" developer preference first, in order to prevent users in the field from changing it when they shouldn't.

PR #1687, opened 2017-04-18
Improvements to install-from-list feature

amstone326:
Improvements made:
- switch to storing retrieved available apps in a the global db, instead of in a SharedPreferences object
- fix spacing of app list items and improve error message for failed request
- add analytics for feature usage

cross-request: https://github.com/dimagi/commcare-core/pull/551


amstone326:
@wpride This PR switches the install-from-list functionality over to storing the retrieved apps in a db instead of SharedPreferences, like you suggested

app/src/org/commcare/activities/InstallFromListActivity.java
```diff
@@ -422,4 +414,22 @@ public boolean onOptionsItemSelected(MenuItem item) {
         return super.onOptionsItemSelected(item);
     }
 
+    private void loadPreviouslyRetrievedAvailableApps() {
+        for (AppAvailableToInstall availableApp : storage()) {
```
wpride: Wow this is cool! Embarrassing, I never thought to use an inline iterator with storage() before.

Will this throw a NullPointerException if storage() has no records?

app/src/org/commcare/android/database/global/models/AppAvailableToInstall.java
```diff
@@ -0,0 +1,70 @@
...
+ * Represents an app that exists on CommCare HQ (either india or prod) and is accessible for
+ * the currently-authenticated user (either mobile or web) to install.
+ *
+ * Created by amstone326 on 2/3/17.
+ */
+public class AppAvailableToInstall extends Persisted {
+
+    public static final String STORAGE_KEY = "available_apps";
+
+    @Persisting(1)
```
wpride: Did you move this into `commcare-android` for the usage of these annotations?

app/src/org/commcare/activities/InstallFromListActivity.java
```diff
@@ -341,21 +333,21 @@ public void handleIOException(IOException exception) {
     }
 
     private void handleRequestError(int responseCode) {
-        System.out.println(responseCode);
+        System.out.println("Request to " + urlCurrentlyRequestingFrom + " had error code response: " + responseCode);
```
wpride: Should this be logging instead of Sysout?

app/src/org/commcare/activities/InstallFromListActivity.java
```diff
@@ -243,9 +234,9 @@ protected void onPostExecute(Void result) {
             };
 
             task.connect((CommCareTaskConnector)this);
-            task.executeParallel();
             task.setResponseProcessor(this);
             setAttemptedRequestFlag();
+            task.executeParallel();
```
wpride: confirming this reordering is intentional

amstone326: yea it was; I don't think it matters much but seemed more right this way

app/src/org/commcare/activities/InstallFromListActivity.java
amstone326: Yea, I wasn't sure.. I've been trying lately to lean on the side of not cluttering the logs with things we probably won't ever act upon, but I may have gone too far in the other direction. This should probably be a log.

app/src/org/commcare/activities/InstallFromListActivity.java
amstone326: Hm interesting thought but apparently not, because when you open this activity for the first time, storage() will always have no records, and that works just fine

app/src/org/commcare/android/database/global/models/AppAvailableToInstall.java
amstone326: I moved it into Android because the `Persisted` class is in Android, and this extends that

app/src/org/commcare/activities/InstallFromListActivity.java
amstone326: Ah, did you mean log as in `Logger.log()` or as in `Log.e()`?

wpride:
@amstone326 is this missing a release note?

PR #1689, opened 2017-04-19
"Heartbeat" comms infrastructure (including prompted updates)

amstone326:
This PR adds a "Heartbeat Manager" which is tied to the session service and will, as long as the session remains active, periodically (currently 1x per day) send a request to the server for a "heartbeat" for that app. 

The request from mobile can include any fixed-length pieces of information that are relevant to communicating a "heartbeat" about the device's current state. [This doc](https://docs.google.com/a/dimagi.com/document/d/1aw_m4fsCcZOho2qZsuKeT6DhhPE4Ol0gb4jvPoTMpK0/edit?usp=sharing) outlines the few things that I'm already including, as well as a list of several other things we could potentially include. @ctsims, we should either chat through the things on that list, or you can just comment with thoughts/additions.

The response from HQ should include any information that is useful toward that app maintaining a "healthy" state on the current device (right now, the only thing I'm checking for is if there is an .apk and/or .ccz updates that the user should be prompted/forced to do).

**Notes/Questions:**
- The HQ work for this has not been done or even resourced yet, so right now I've been parsing a test response to test everything. The code to make the request is in there but not getting called yet. Once the HQ side is actually done, I can swap out the call to ` parseTestHeartbeatResponse()` in favor of `requestHeartbeat(currentUser)`, and test it.
- The current UI and conceptual model for the update piece (i.e. how I distinguish between forced vs. prompted updates, etc.) is just my best first guess. We should do some UAT with interested parties to get feedback on how it seems to them. (Update: I'm waiting on hearing back from Saket from this)
- Right now, the screen to prompt/force an update on the user is only shown once after a login, if it is appropriate to do so, and then not again until the user logs in again. It should probably be shown more often than this, but I wasn't sure on a good rule, so I haven't implemented anything for this yet. @ctsims, let me know if you had anything in mind for this. Otherwise I think this is a good question for UAT, as well. 

**Screenshots below -->**


amstone326:
This screen will appear upon login if there is currently an .apk or. ccz update to prompt:
screenshot_20170420-153547

(and if only 1 is present, then only that shows):
screenshot_20170425-145159

If 1 of the updates is "forced" (meaning a force-by date was set and it is past that date), then that one is emphasized:
screenshot_20170425-145130

amstone326:

wpride:
@amstone326 conflicts

app/src/org/commcare/UpdateToPrompt.java
```diff
@@ -0,0 +1,158 @@
...
+
+    public boolean isNewerThanCurrentVersion(CommCareApp currentApp) {
+        if (isApkUpdate) {
+            try {
+                Context c = CommCareApplication.instance();
+                PackageInfo pi = c.getPackageManager().getPackageInfo(c.getPackageName(), 0);
+                ApkVersion currentVersion = new ApkVersion(pi.versionName);
+                return currentVersion.compareTo(this.apkVersion) < 0;
+            } catch (PackageManager.NameNotFoundException e) {
+                System.out.println("Couldn't get current .apk version to compare with in " +
```
wpride: Should be logged?

app/src/org/commcare/UpdateToPrompt.java
```diff
@@ -0,0 +1,158 @@
...
+            try {
+                ExtUtil.write(serializedStream, this);
+                String serializedString = Base64.encodeToString(baos.toByteArray(), Base64.DEFAULT);
+                prefs.edit().putString(
+                        isApkUpdate ? KEY_APK_UPDATE_TO_PROMPT : KEY_CCZ_UPDATE_TO_PROMPT,
+                        serializedString).commit();
+                baos.close();
+                serializedStream.close();
+            } catch (IOException e) {
+                System.out.println("IO error encountered while serializing UpdateToPrompt: "
```
wpride: Worth logging?

app/src/org/commcare/UpdateToPrompt.java
```diff
@@ -0,0 +1,158 @@
...
+            }
+        }
+    }
+
+    @Override
+    public void readExternal(DataInputStream in, PrototypeFactory pf) throws IOException, DeserializationException {
+        this.versionString = ExtUtil.readString(in);
+        this.isApkUpdate = ExtUtil.readBool(in);
+        this.hasForceByDate = ExtUtil.readBool(in);
+        if (hasForceByDate) {
```
wpride: The usual pattern here is using an `ExtWrapNullable` - any reason you decided to do the checking manually?

app/src/org/commcare/UpdateToPrompt.java
```diff
@@ -0,0 +1,158 @@
...
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
+import java.util.Date;
+
+/**
+ * Created by amstone326 on 4/13/17.
+ */
+
+public class UpdateToPrompt implements Externalizable {
```
wpride: Did you consider using two separate classes for apkVersion and ccVersion, possibly with a shared super? Not certain that it'd be cleaner but it does seem like there's a good amount of `if`/`else` checking and not that many shared paths.

app/src/org/commcare/UpdateToPrompt.java
```diff
@@ -0,0 +1,158 @@
...
+
+    public static final String KEY_CCZ_UPDATE_TO_PROMPT = "ccz-update-to-prompt";
+    public static final String KEY_APK_UPDATE_TO_PROMPT = "apk-update-to-prompt";
+
+    private String versionString;
+    private int cczVersion;
```
wpride: Total NIT, but the `Z` in `ccz` just stands for "zip file" so I think ccVersion would be more accurate (though completely feel free to ignore)

app/src/org/commcare/UpdateToPrompt.java
```diff
@@ -0,0 +1,158 @@
...
+    private ApkVersion apkVersion;
+    private Date forceByDate;
+    private boolean hasForceByDate;
```
wpride: Seems like you could get rid of the `hasFroceByDate` variable and instead check for `forceByDate` being null

app/src/org/commcare/activities/HomeScreenBaseActivity.java
```diff
@@ -94,16 +96,16 @@
...
     /**
      * Request code for launching FormEntryActivity
      */
     protected static final int MODEL_RESULT = 4;
-    public static final int GET_INCOMPLETE_FORM = 16;
-    protected static final int PREFERENCES_ACTIVITY = 512;
-    protected static final int ADVANCED_ACTIONS_ACTIVITY = 1024;
-    protected static final int CREATE_PIN = 16384;
-    protected static final int AUTHENTICATION_FOR_PIN = 32768;
+    protected static final int MAKE_REMOTE_POST = 5;
```
wpride: ✅ on the ridiculous int cleanup

app/src/org/commcare/activities/PromptUpdateActivity.java
```diff
@@ -0,0 +1,185 @@
...
+
+import org.commcare.CommCareHeartbeatManager;
+import org.commcare.UpdateToPrompt;
+import org.commcare.dalvik.R;
+import org.commcare.views.ManagedUi;
+import org.commcare.views.UiElement;
+import org.javarosa.core.services.locale.Localization;
+
+/**
+ * Created by amstone326 on 4/19/17.
```
wpride: Big enough class to merit a comment IMO

app/src/org/commcare/CommCareHeartbeatManager.java
```diff
@@ -0,0 +1,287 @@
...
+     * @return an UpdateToPrompt that has been stored in SharedPreferences and is still relevant
+     * (i.e. the user hasn't updated to or past this version since we stored it)
+     */
+    public static UpdateToPrompt getCurrentUpdateToPrompt(boolean forApkUpdate) {
+        CommCareApp currentApp = CommCareApplication.instance().getCurrentApp();
+        if (currentApp != null) {
+            String prefsKey = forApkUpdate ?
+                    UpdateToPrompt.KEY_APK_UPDATE_TO_PROMPT : UpdateToPrompt.KEY_CCZ_UPDATE_TO_PROMPT;
+            String serializedUpdate = currentApp.getAppPreferences().getString(prefsKey, "");
+            if (serializedUpdate != null && !"".equals(serializedUpdate)) {
```
wpride: lint tells me `serializedUpdate != null` will always be `true`

app/src/org/commcare/CommCareHeartbeatManager.java
```diff
@@ -0,0 +1,287 @@
...
+import java.io.InputStream;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.Timer;
+import java.util.TimerTask;
+
+/**
+ * Created by amstone326 on 4/13/17.
```
wpride: Could use a comment

wpride: Otherwise, seems like this class is still anticipated some changes from HQ until its finished up so will wait until this happens before digging in.

app/src/org/commcare/UpdateToPrompt.java
amstone326: Lol. Fair but definitely not going to change it because ccVersion could easily be misinterpreted to mean "CommCare version" which is not what that value represents.

app/src/org/commcare/UpdateToPrompt.java
amstone326: I actually didn't consider it at all. I think you're right that it would be marginally cleaner, but I think it probably doesn't have enough of a clear advantage to be worth the refactor right now. Let me know if you disagree, though.

app/src/org/commcare/CommCareHeartbeatManager.java
amstone326: right it is

app/src/org/commcare/CommCareHeartbeatManager.java
amstone326: Makes sense, although I don't anticipate many changes to this (besides actually calling the real request method instead of parsing the test response). The code that's there obviously needs to be tested, but I'm fairly confident it's not going to need a lot of changes, if any.

app/src/org/commcare/UpdateToPrompt.java
amstone326: Oh, I guess just because I've never used that and wasn't super aware of it. Will adjust.

app/src/org/commcare/CommCareHeartbeatManager.java
```diff
@@ -0,0 +1,309 @@
...
+
+        @Override
+        public void handleIOException(IOException exception) {
+            Logger.log(AndroidLogger.TYPE_ERROR_SERVER_COMMS,
+                    "Encountered IOException while getting response stream for heartbeat response: "
+                            + exception.getMessage());
+        }
+
+        private void processErrorResponse(int responseCode) {
+            Logger.log(AndroidLogger.TYPE_ERROR_SERVER_COMMS,
```
ctsims: We may end up processing a lot of these, so I think we should be cautious to avoid logging errors to our forensics lots unless we can explicitly rate limit them.

app/src/org/commcare/UpdateToPrompt.java
```diff
@@ -0,0 +1,154 @@
...
+            }
+        } else {
+            int currentVersion = currentApp.getCommCarePlatform().getCurrentProfile().getVersion();
+            return currentVersion < this.cczVersion;
+        }
+    }
+
+    private void writeToPrefsObject(SharedPreferences prefs) {
+        ByteArrayOutputStream baos = null;
+        DataOutputStream serializedStream = null;
```
ctsims: I think there's an existing Utility method for "Pass in an externalizable, get back an byte[]", I'd try to re-use that as the error handling here is quite unfortunate to end up replicating

app/src/org/commcare/UpdateToPrompt.java
```diff
@@ -0,0 +1,154 @@
...
+        try {
+            baos = new ByteArrayOutputStream();
+            serializedStream = new DataOutputStream(baos);
+            try {
+                ExtUtil.write(serializedStream, this);
+                String serializedString = Base64.encodeToString(baos.toByteArray(), Base64.DEFAULT);
+                prefs.edit().putString(
+                        isApkUpdate ? KEY_APK_UPDATE_TO_PROMPT : KEY_CCZ_UPDATE_TO_PROMPT,
+                        serializedString).commit();
+                baos.close();
```
ctsims: You don't need to put these here if you have them in the finally() block below

app/src/org/commcare/CommCareHeartbeatManager.java
```diff
@@ -0,0 +1,309 @@
...
+        if (currentApp != null) {
+            String prefsKey = forApkUpdate ?
+                    UpdateToPrompt.KEY_APK_UPDATE_TO_PROMPT : UpdateToPrompt.KEY_CCZ_UPDATE_TO_PROMPT;
+            String serializedUpdate = currentApp.getAppPreferences().getString(prefsKey, "");
+            if (!"".equals(serializedUpdate)) {
+                try {
+                    byte[] updateBytes = Base64.decode(serializedUpdate, Base64.DEFAULT);
+                    DataInputStream stream = new DataInputStream(new ByteArrayInputStream(updateBytes));
+                    UpdateToPrompt update = (UpdateToPrompt)
+                            ExtUtil.read(stream, UpdateToPrompt.class, ExtUtil.defaultPrototypes());
```
ctsims: We have a different "master prototypes factory" set in Android that you can get from the context.

app/src/org/commcare/CommCareHeartbeatManager.java
```diff
@@ -0,0 +1,309 @@
...
+                    if (update.isNewerThanCurrentVersion(currentApp)) {
+                        return update;
+                    } else {
+                        // The update we had stored is no longer relevant, so wipe it and return nothing
+                        wipeStoredUpdate(forApkUpdate);
+                        return null;
+                    }
+                } catch (Exception e) {
+                    // Something went wrong, so clear out whatever is there
+                    Logger.log(AndroidLogger.TYPE_ERROR_WORKFLOW,
```
ctsims: I agree that crashing here in general if you can't read the update prompt is the right plan, but I'd wrap this call over the extutil.read() call specifically, rather than generically catching all errors

app/src/org/commcare/CommCareApplication.java
```diff
@@ -712,6 +712,7 @@ public void onServiceConnected(ComponentName className, IBinder service) {
                         }
                     }
 
+                    CommCareHeartbeatManager.instance().startHeartbeatCommunications();
```
ctsims: I think the HeartbeatManager's lifecycle can be managed entirely within the CommCareService, which would remove the need for the singleton implementation. The CommCareService already manages its lifecycle cleanly, so if it handled these methods inside of the service ("one Heartbeat per session service") it would be equivalent

app/src/org/commcare/CommCareHeartbeatManager.java
```diff
@@ -0,0 +1,309 @@
...
+                    parseTestHeartbeatResponse();
+                } catch (SessionUnavailableException e) {
+                    // Means the session has ended, so we should stop these requests
+                    stopHeartbeatCommunications();
+                }
+            }
+        };
+
+        heartbeatTimer = new Timer();
+        heartbeatTimer.schedule(heartbeatTimerTask, new Date(), ONE_DAY_IN_MS);
```
ctsims: The scheduling doesn't quite match my expectations. based on what we'd discussed previously, I think I would expect this process to be

[Manager] -> [One Thread Ever]

Manager starts the thread and wraps its events/lifecycle.

one- thread-ever is in a while() loop, that exits If and When [The thread succeeds] OR it reads a "stopPolling" flag 

stopHeartBeat -> one-thread-ever.stopPolling = true;

at the end of the while loop, if I/O was unsuccesful, the thead sleeps for a while(60 seconds)

When the thread wakes up it checks a flag that says "you should try again right now", if that's set to true, it tries again, if not it keeps sleeping in 60 second increments until some longer backoff (minutes/hours)

When a form push succeeds,the app can say "session.tellTheHeartbeatToTry->Manager->the-one-thread.tryagainflag = true"

app/src/org/commcare/CommCareHeartbeatManager.java
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    public void stopHeartbeatCommunications() {
+        if (heartbeatTimer != null) {
+            heartbeatTimer.cancel();
```
ctsims: The "inside" of the heardbeat should likely have a flag set to cancel the poll, but we should also disconnect the "listener" from the heartbeat. Right now, the heartbeat thread will always be dependent on this class as long as it is running, which could result in a memory leak over it. This won't be needed if all of the state for the heartbeat processing can live inside of the internal heartbeat class, rather than the manger.

app/src/org/commcare/CommCareHeartbeatManager.java
```diff
@@ -0,0 +1,309 @@
...
+ * Currently, the primary content of the server's response to the heartbeat request will be
+ * information about potential binary or app updates that the app should prompt users to conduct.
+ *
+ * Created by amstone326 on 4/13/17.
+ */
+public class CommCareHeartbeatManager {
+
+    private static final long ONE_DAY_IN_MS = 24 * 60 * 60 * 1000;
+
+    private static final String TEST_RESPONSE =
```
amstone326: Where else would it go? It's a test response for the request that this class makes? (I'm also going to delete it once I'm able to test making the actual request and it works)

app/assets/locales/android_translatable_strings.txt
```diff
@@ -19,6 +19,16 @@ updates.pinned.download=Downloading updates
 updates.pinned.progress=${0}/${1}
 install.current.version=Current version: ${0}
 
+updates.available.title=Updates Available
+update.required.title=Update Required
+update.later.option=I'll update later.
+prompted.apk.update.info=There is a newer version of CommCare available, and it is recommended that you update now.
+forced.apk.update.info=There is a newer version of CommCare available. You must update before continuing to use CommCare.
```
amstone326: Is what on our radar? Right now I just didn't design the non-forced updates to have any date associated with them, since I wasn't sure how we'd use it. I can add that if we want it just as something to display to the user in this message, but I was envisioning that by the time app builders would usually be setting an update prompt, they would most likely just want it to be done asap.

app/src/org/commcare/CommCareHeartbeatManager.java
amstone326: why would we expect to get a lot of these?

app/src/org/commcare/CommCareHeartbeatManager.java
```diff
@@ -0,0 +1,309 @@
...
+    public void startHeartbeatCommunications() {
+        TimerTask heartbeatTimerTask = new TimerTask() {
+            @Override
+            public void run() {
+                try {
+                    User currentUser = CommCareApplication.instance().getSession().getLoggedInUser();
+                    //simulateRequestGettingStuck();
+                    //requestHeartbeat(currentUser);
+                    parseTestHeartbeatResponse();
+                } catch (SessionUnavailableException e) {
```
amstone326: added

app/src/org/commcare/UpdateToPrompt.java
amstone326: Ah cool, that's much better

app/src/org/commcare/CommCareApplication.java
amstone326: "The CommCareService already manages its lifecycle cleanly," --> But this `onServiceConnected` method is clearly responsible for initializing a lot of things that are supposed to be tied to the start of the session. What makes this different?

app/src/org/commcare/CommCareHeartbeatManager.java
```diff
@@ -0,0 +1,309 @@
...
+ * with a "heartbeat" request, and then handle the response. The lifecycle of the TimerTask is
+ * tied to that of the CommCareSessionService; it should be started whenever a session service is
+ * started, and ended whenever a session service is ended for any reason.
+ *
+ * Currently, the primary content of the server's response to the heartbeat request will be
+ * information about potential binary or app updates that the app should prompt users to conduct.
+ *
+ * Created by amstone326 on 4/13/17.
+ */
+public class CommCareHeartbeatManager {
```
amstone326: done

app/src/org/commcare/CommCareHeartbeatManager.java
amstone326: Not following this so we should discuss in our sync.

app/src/org/commcare/UpdateToPrompt.java
amstone326: removed

app/src/org/commcare/CommCareHeartbeatManager.java
amstone326: Why? No matter what, we know that we’ve failed to read out whatever is currently stored in prefs, so we may as well clear it out or else we’re just going to encounter the same error over and over

app/src/org/commcare/CommCareHeartbeatManager.java
```diff
@@ -0,0 +1,309 @@
...
+                    //requestHeartbeat(currentUser);
+                    parseTestHeartbeatResponse();
+                } catch (SessionUnavailableException e) {
+                    // Means the session has ended, so we should stop these requests
+                    stopHeartbeatCommunications();
+                }
+            }
+        };
+
+        heartbeatTimer = new Timer();
```
amstone326: Addressed by always canceling the existing Timer, if it’s non-null, before instantiating a new one

app/src/org/commcare/CommCareHeartbeatManager.java
amstone326: (no longer relevant since I'm using the `SerializationUtil` method for this

app/src/org/commcare/CommCareHeartbeatManager.java
amstone326: Hey, I'm a little confused about this comment. In the minimal discussions we had about how to set this up, you had explicitly said that you weren't sure whether it was preferable to tie the heartbeat's lifecycle to being triggered post-form submissions, vs. just set it up to happen on a timer, and that I should go with whichever choice ended up seeming easier/better to me. We also then later discussed that I was planning on using the timer strategy because I mentioned to you that I had found discussions online where people were wary of using a `TimerTask`, and you checked those out and then we agreed that it seemed like the concerns weren't relevant to us, and you agreed that that seemed like a good option. So I'm confused why now you're surprised that I went with the timer option over the tied-to-form-submission option.

app/src/org/commcare/CommCareHeartbeatManager.java
amstone326: (this is now `HeartbeatLifecycleManager` and `HeartbeatRequester`)

app/src/org/commcare/CommCareHeartbeatManager.java
ctsims: I hadn't realized when I commented that we weren't doing the retry yet, can chat over voice. I think we expect one of these each time we fail to process a response, which will likely occur more often if we retry more proactively.

app/src/org/commcare/CommCareHeartbeatManager.java
ctsims: Ah, I think I was unclear in this comment, the issue was't around why deserialization failed, it was that there was more code wrapped in the try/catch than just the read/deserialize, so in theory there could be other bugs in the code that instead of being properly thrown would get obfuscated as deserialization errors.

ctsims:
Hey,

This is looking good based on what we talked about. I think one last judgement call we might need to make is how often to poll for doing the update, and I wonder whether the server team might have guidance on how often might be too often in high-load situations (IE: If they are rate limiting requests, retrying with higher frequency will keep the server overloaded).

amstone326:
@ctsims Cool, so things I've fixed and added since you looked last night (encompassed in the most recent 6 commits, although 2 of those were a commit-and-revert):

- The place in `onServiceConnected` from which I was calling `initHeartbeatLifecycle()` was slightly off; should only be called when we have a user
- The way I was constructing a string from the response bytes was incorrect (discovered this while testing with the actual response from HQ; once that was fixed, the response was otherwise working as expected)
- Added a call to `initHeartbeatLifecycle()` immediately after form submission succeeds
- `initHeartbeatLifecycle()` was incorrectly always instantiating a new lifecycle manager object when called, which would have orphaned an existing one if there was one. Changed this to only instantiate the object if it's null.
- Decreased the frequency with which the TimerTask makes the request from once every 5 min to once every hour. I'll check in with Cal about the concern you raised, but this seemed like a good way to address it in the mean time. Especially now that we make the attempt after form submission (which is most likely to be when we succeed anyway), it seems fine to significantly reduce the rate of the random attempts.
- Added the IMEI and user id as default query params in `AndroidModernHttpRequester` (previously these were only being added to `HttpRequestGenerator.makeCaseFetchRequest`)

amstone326:
Self-merging this so that we can code freeze. 

@ctsims if there are any changes you want based on my last comment, let me know later and they can go in during QA. I'm also planning on adding tests, and possibly UI changes based on feedback I get from Saket tomorrow.

PR #1695, opened 2017-04-25
Grid menu density fix

ctsims:
Fix issue where on grid menus the initial scaledown that gets applied to images is the same as for non-grid icons. Made images look terrible.

Before:
scale_before

After
scale_after

### DIFF
--- a/app/res/values/dimens.xml
+++ b/app/res/values/dimens.xml
@@ -33,6 +33,7 @@
     <dimen name="list_content_with_space_for_badge">58dp</dimen>
     <dimen name="list_icon_size">48dp</dimen>
     <dimen name="list_icon_bounding_dimen">60dp</dimen>
+    <dimen name="list_grid_bounding_dimension">180dp</dimen>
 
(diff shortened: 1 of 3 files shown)


amstone326:
@ctsims is this ready to go? Just doesn't have any tags so wasn't sure.

ctsims:
Hey!

@amstone326 

I think so. quick question for you: I noticed that there weren't different values for the icon expansion bounding boxes before (IE: For "small" / "medium" or "HDPI", etc). I know we recently switched to fixed-size module screen icons for non-grid view, is it likely we should try to see whether there are failure modes for those either?

ctsims:
Release Note:

Fix for bug where "Grid" menu images showed at lower resolution than intended.

amstone326:
@ctsims Release notes for 2.36 have already been compiled since that happens before QA starts, so any additions should get added directly to https://confluence.dimagi.com/display/MD/Mobile+Release+Notes+Archives

ctsims:
@amstone326 ah, should have realized that. Updated the doc. thanks!

PR #1696, opened 2017-04-27
Log potential form loss causes

amstone326:
Add logs to any place in the code where we knowingly delete a `FormRecord` from the db (under a new log tag called `form-deletion`)


wpride:
@amstone326 what's the question?

amstone326:
@wpride @ctsims Since I've determined that whatever is causing the elusive missing forms bug is not currently getting logged, I want to make sure we add to 2.36 logging for any other potential causes we can think of. So I've gone through and added a log for every time that we knowingly delete a form record from the db, for whatever reason. Can you think of any other types of situations that we should add logging for as well?

app/src/org/commcare/android/database/user/models/FormRecord.java
```diff
@@ -219,4 +221,14 @@ public void setFormNumberForSubmissionOrdering(int num) {
         this.submissionOrderingNumber = ""+num;
     }
 
+    public void logPendingDeletion(String classTag, String reason) {
+        String logMessage = String.format(
+                "Wiping form record with id %1$s and submission ordering number %2$s " +
```
wpride: Think you can use just %s instead of %[*]$s for these

app/src/org/commcare/tasks/WipeTask.java
```diff
@@ -38,6 +38,7 @@ protected Boolean doTaskBackground(String... params) {
 
         Log.d(TAG, "Wiping sent form records");
         for (FormRecord record : records) {
+            record.logPendingDeletion(TAG, "we are about to initiate a wifi direct transfer");
```
wpride: This would be after Wi-Fi Direct completes (not before initiation)

wpride:
@amstone326 SD Card transfer deletes forms here https://github.com/dimagi/commcare-android/blob/3d34b715579c0a9adee962cec01dc0979d82f0b4/app/src/org/commcare/tasks/SendTask.java#L120-L120

wpride:
@amstone326 Looks great! Just a couple comments.

amstone326:
@wpride Re:your last comment- I think I'm comfortable with not logging that, because I already added a log for when the form records associated with those files get deleted after the files are dumped: https://github.com/dimagi/commcare-android/pull/1696/files#diff-495847570d842daad554b457c9aa5b8dR252. So this should just be a duplicate indicator of the same event, right?

amstone326:

PR #1697, opened 2017-04-28
Hide dangerous dev options behind multiple apps privilege (for now)

amstone326:
Since we need to make the documentation for Developer Preferences public in 2.36 (due to putting the visibility of the Update Options settings menus behind a developer pref), we also need to add an extra layer of hiding to those Developer Preferences that would be dangerous for non-Dimagi users to encounter and mess with. 

This PR is a stop-gap measure for doing so, by tying the visibility of the more dangerous developer prefs to the "unlimited multiple apps" privilege that comes from scanning the "Enable Mobile Privileges" barcode on HQ. This will ensure that those prefs are only visible to Dimagi users for now. Going forward, we plan to make some changes to the mobile privileges framework that will allow us to create a flag for this permission specifically, instead of piggy-backing off of the multiple apps one.


amstone326:
Ready to merge pending @ctsims' approval of this strategy

ctsims:
@amstone326 
Two quick updates here:

1)
Can we add the name of the second privilege to this now, so once we have enabled it, it will start working automatically

2) 
I just noticed that there are actually a few other app properties I think we should hide this way:

All of the server URI properties in-use right now are quite dangerous. there are multiple crashes on ICDS that are pretty clearly people changing those properties. Can we hide the "Server SEttings" preference area as well?

amstone326:
@ctsims Made the following changes:
- Changed the privilege check that we're currently doing to check for _either_ the multiple apps flag or a new `advanced_settings_access` flag that doesn't exist on HQ yet, but will in the future
- Placed the visibility of the option that opens the server settings screen behind this same privilege
- Changed the developer preference screen to use a whitelisting strategy. I added 2 other settings to the whitelist that you hadn't mentioned: `cc-superuser-enabled` (which just controls whether or not the developer options themselves are enabled) and `cc-alternate-question-text-format` (which was the feature I added for ICDS to show images above question text instead of below them, which seems harmless and could potentially be interesting to other projects)

app/src/org/commcare/preferences/DeveloperPreferences.java
```diff
@@ -378,4 +394,30 @@ public static boolean shouldShowUpdateOptionsSetting() {
                 CommCarePreferences.YES) || BuildConfig.DEBUG;
     }
 
+    private void hideDangerousDeveloperPrefsIfNeeded() {
+        if (!GlobalPrivilegesManager.isAdvancedSettingsAccessEnabled() && !BuildConfig.DEBUG) {
+            // Dangerous privileges should not be showing
+            PreferenceScreen prefScreen = getPreferenceScreen();
```
ctsims: We could iterate over this with
prefScreen.getPreferenceCount -> prefScreen.getPReference(cout)

and just hide anything that doesn't show up in the whitelisted keys, which would be a bit less verbose (and less error prone) than building both lists and diffint

amstone326: Ah, thank you! I was trying to find a way to iterate through all prefs on the screen and was clearly looking in the wrong place.

amstone326:
Release Note: Users can now gain access to a short list of advanced settings called "Developer Preferences" that allow configuration of some specialized on-device behaviors 

Announce to field-dev: By default when you turn on the Developer Options, the list will now only include those that have been publicly documented here: https://confluence.dimagi.com/display/commcarepublic/CommCare+Developer+Options. In order to re-gain access to the full list, you can follow the instructions [here](https://confluence.dimagi.com/display/internal/CommCare+Android+Developer+Options+--+Internal#CommCareAndroidDeveloperOptions--Internal-HowtogainaccesstoALLdeveloperoptions(relevantfor2.36andabove)).

PR #1706, opened 2017-05-09
Don't re-derive print template path

orangejenny:
https://manage.dimagi.com/default.asp?252703

This was throwing an `InvalidReferenceException` because the path had already been derived, looks like that happens [here](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/commcare/suite/model/Detail.java#L469).

@amstone326


amstone326:
closing in favor of https://github.com/dimagi/commcare-core/pull/561

PR #1711, opened 2017-05-11
Robolectric tests for heartbeat request and forced updates

amstone326:
@ctsims @wpride, 1 Question on this-- The reason I'm only testing the .ccz updates here is because I wasn't sure how to reliably know what robolectric would see as being the current version of CommCare during test runs. For some reason when I tested it, it was coming up with 2.32.1 as the current version. Do either of you have any idea why that would be?


amstone326:

ctsims:
this looks awesome, good to go!

PR #1713, opened 2017-05-12
More install-from-list improvements

amstone326:
- Start app install automatically after an app is selected from the list
- Check for connectivity before making the request, and show a useful error message if not


amstone326:

PR #1717, opened 2017-05-22
Preserve params when deriving JavaHttpRoot

wpride:
fixes https://manage.dimagi.com/default.asp?254166

We were dropping all parameters when deriving a root from its parent since we were just taking the substring up to the file name. This checks for params in the context root and appends them to the derived path if they're present.

I think this change has enough potential side effects, and the affected feature internal enough, that we should wait until 2.37 instead of trying to get into 2.36, but open to discussion.


ctsims:
Agree that this feels like a 2.37 fix. 

@wpride Can we add tests to this which cover our expected behaviors? This seems like a good place for them, as it's tough to understand the code on its own.

PR #1718, opened 2017-05-22
Fix for image file enlargement on Android 7

ctsims:
Fix for: https://manage.dimagi.com/default.asp?253951

Images on Android 7 would crash when double clicked, due to lack of file system permissions

@amstone326 Copied your technique for this from the video work.


ctsims:
Release Note: Fix for viewing enlarged image files on Android 7 devices

PR #1721, opened 2017-05-24
Quarantine forms that errored on HQ

amstone326:
~~Don't merge until the following questions have been answered and relevant code changes have been made:~~ Both resolved:

~~1. Whether HQ will always be able to send a useful, short-ish, pre-determined error string for each case, or if there will be some generic failure cases that result in sending back a longer stacktrace. If it is the latter, Aliza will need to make some changes to the expectations surrounding the message that is received back from HQ in the 422 response.~~ **Sravan has confirmed this**
~~2. Whether it will be possible to create a form submission in a calabash test that will _always_ trigger this response, thus allowing this functionality to be tested in calabash. If not, Aliza will need to revisit writing a robolectric test.~~ **test here: https://github.com/dimagi/commcare-ui-tests/pull/93**

**Changes this PR makes:**

- Mobile now expects HQ to send back a 422 if a form submission could not be processed, along with a concise string explaining the reason why in body of the `<message>` element in the response. When mobile receives this response, it will quarantine the FormRecord in question. Hence, there is now a conceptual distinction between whether a form record was quarantined due to a local issue with the record, or due to an HQ processing failure

- Quarantined forms in the Saved Forms list now include a line for the reason they were quarantined (it will either say `Quarantined due to: Server Processing Error` or `Quarantined due to: Local Record Issue`:
screenshot_20170615-132504

- I also added an option in the long-click menu for "View Full Reason for Quarantine". If selected, a dialog will appear with a longer message for why the form was quarantined (If it was quarantined due to an HQ processing failure, this is where the reason for failure that HQ sends back will be displayed)
screenshot_20170615-164623
screenshot_20170615-164629


amstone326:

amstone326:
This is ready to be reviewed, but don't want it to be pulled yet since HQ side has not been figured out yet

app/unit-tests/src/org/commcare/android/tests/processing/QuarantinedFormsTest.java
```diff
@@ -0,0 +1,31 @@
...
+import org.junit.Test;
+
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+
+/**
+ * Created by amstone326 on 6/14/17.
+ */
+
+public class QuarantinedFormsTest {
```
amstone326: I took a stab at writing a robolectric test for this, with a mocked `ProcessAndSendTask` object, etc., but it was a mess and didn't seem like it was super valuable based on the amount of stuff I needed to fake, so I am planning to write a Calabash test instead, once the end-to-end work is done

app/src/org/commcare/utils/FormUploadResult.java
```diff
@@ -10,38 +10,51 @@
...
      */
     FAILURE(1),
 
     /**
-     * The server returned an authentication error
+     * There is a problem with this record that prevented submission success for this record
+     * specifically, but others will be allowed to continue trying
      */
-    AUTH_FAILURE(2),
+    RECORD_FAILURE(2),
```
amstone326: @ctsims I reordered these a little bit because the behavior of `FormUploadResult.getWorstResult` implies that the worse a result is, the higher the `orderVal` should be. The ordering that previously existed here did not, in my understanding of things, accurately reflect that, but I'm not sure if the one I settled on is correct either. Would appreciate your input, to the extent that you think it's important to get this right.

ctsims: your new order seems correct to me

app/unit-tests/src/org/commcare/android/tests/processing/QuarantinedFormsTest.java
ctsims: We have mock response endpoints for data restores, it might be worth seeing how hard it would be to add mock endpoints for submissions.

Here's the restore Mock: https://github.com/dimagi/commcare-android/blob/master/app/unit-tests/src/org/commcare/android/mocks/ModernHttpRequesterMock.java

I'm a bit concerned that a calabash test her might be pretty fragile, but since you've worked here more recently than I mean I could be wrong.

ctsims:
@amstone326 - One thing I'd mention: The response from HQ is quite likely to be long, and potentially include a stack trace or other whitespace sensitive value.

Have we tested a long error response and seen what it looked like? It might be worth putting the response data in a smaller test field, or one with fixed width fonts.

app/unit-tests/src/org/commcare/android/tests/processing/QuarantinedFormsTest.java
amstone326: @ctsims That's essentially what I tried to do (if you look at the full commit history of this PR you'll see that I iterated on that for a while and then ultimately removed it). My decision was based on the fact that a) The test was failing in a way that I couldn't figure out the cause of after trying for quite a while, and b) That I was mocking a large portion of the behavior, which made the usefulness of the test feel really limited. Based on those 2 things, I decided a calabash test would be both easier (or at least take less of my time) and a "better" test. 

On the flip side, the challenge of writing a calabash test for this is having a form submission that will consistently trigger this response from HQ. In order to determine the feasibility of that, I would need to talk with Ben or whoever works on this for HQ to determine what situations will trigger this response, and then think about which (if any) is something that we can reliably recreate on mobile. So I think it makes sense to wait on deciding about this until I can have that converesation. If reproducing in a calabash test is feasible, I do think I prefer that, for the reasons I outlined above.

amstone326:
@ctsims I was actually thinking of specifically telling HQ _not_ to include a stack trace, and instead just having a specific, longer-but-not-too-long error message pre-determined for each class of error that would trigger this (again, the specifics of that are pending discussion with whoever does this). But my thinking was that it's unlikely for an HQ stacktrace to be very meaningful to whoever is looking at the quarantined form on the phone, even if it is a mobile dev, whereas a targeted error message may be more helpful. Do you disagree with not including the stacktrace?

ctsims:
If HQ can always guarantee that they can send back a "good" message, that's great! 

I think that if they have a generic error and don't know what went wrong it would still be preferable to send a stack trace (or raw error) rather than nothing, so being ready to receive a fairly long error might still be helpful.

Thanks for thinking ahead on that, though, in general I think a targeted error seems way more helpful. Definitely not blocking if there will always be an actionable error.

amstone326:
@ctsims OK, I will make sure to have that conversation with whoever works on this and then make a determination based on that

amstone326:
Bumping the milestone for this to 2.39 since the HQ side isn't done yet, and I don't feel comfortable releasing this until it's tested

amstone326:
@ctsims This is ready to go for real now. Simon's side is merged and I did some manual testing with it, and then wrote [this calabash test](https://github.com/dimagi/commcare-ui-tests/pull/93) to replicate my manual testing. I also tested that on prior versions of CommCare, HQ does _not_ send the 422 response, and the form submission succeeds as it always has previously.

app/src/org/commcare/android/database/user/models/FormRecord.java
```diff
@@ -75,10 +75,13 @@
     public static final String STATUS_JUST_DELETED = "just-deleted";
 
     public static final String QuarantineReason_LOCAL_PROCESSING_ERROR = "local-processing-error";
+    public static final String QuarantineReason_SERVER_PROCESSING_ERROR = "server-processing-error";
     public static final String QuarantineReason_RECORD_ERROR = "record-error";
     public static final String QuarantineReason_MANUAL = "manual-quarantine";
     public static final String QuarantineReason_FILE_NOT_FOUND = "file-not-found";
 
+    private static final String QUARANTINE_REASON_AND_DETAIL_SEPARATOR = "@@SEP@@";
```
ctsims: I assume this is just to prevent the need for a migration which would add a new field?

app/src/org/commcare/network/CommcareRequestGenerator.java
```diff
@@ -131,7 +131,7 @@ public static CommcareRequestGenerator buildNoAuthGenerator() {
 
     public static HashMap getHeaders(String lastToken) {
         HashMap<String, String> headers = new HashMap<>();
-        headers.put("X-OpenRosa-Version", "2.1");
+        headers.put("X-OpenRosa-Version", "3.0");
```
ctsims: Is there a place where we keep track of what's included in these? I thought there was but now I can't track it down

app/src/org/commcare/tasks/ProcessAndSendTask.java
```diff
@@ -273,13 +273,10 @@ private boolean blockUntilTopOfQueue() throws TaskCancelledException {
...
         for (int i = 0; i < records.length; ++i) {
-            //See whether we are OK to proceed based on the last form. We're now guaranteeing
-            //that forms are sent in order, so we won't proceed unless we succeed. We'll also permit
-            //proceeding if there was a local problem with a record, since we'll just move on from that
-            //processing.
-            if (i > 0 && !(results[i - 1] == FormUploadResult.FULL_SUCCESS || results[i - 1] == FormUploadResult.RECORD_FAILURE)) {
-                //Something went wrong with the last form, so we need to cancel this whole shebang
-                Logger.log(LogTypes.TYPE_WARNING_NETWORK, "Cancelling submission due to network errors. " + (i - 1) + " forms succesfully sent.");
+
+            if (previousFailurePredictsFutureFailures(results, i)) {
```
ctsims: I think this might be slightly misleading as a documenting-method-name. 

It's not that the previous failure implies the next form will fail, it's that submitting the previous form was mandatory for consistency reasons (IE: We couldn't send a create form, so we need to not try to submit the followup form)

app/src/org/commcare/utils/FormUploadUtil.java
```diff
@@ -399,4 +415,29 @@ private static boolean isAudioVisualMimeType(String filename) {
...
+                    if ("processing_failure".equals(natureOfResponse)) {
+                        return parser.nextText();
+                    }
+                    return "";
+                }
+            };
+            return responseParser.parse();
+        } catch (IOException | InvalidStructureException | XmlPullParserException |
+                UnfullfilledRequirementsException e) {
+            Log.e(TAG, "Error while parsing form upload response");
```
ctsims: I think we should use Logger.exception() for this so we can track why we wren't able to process hte response.

Also: I thought we explicitly required that the response both be a 422 *and* be a valid openrosa response payload in order for us to quarantine the result, so swallowing this silently feels dangerous to me. am I misremembering?

app/src/org/commcare/utils/QuarantineUtil.java
```diff
@@ -0,0 +1,69 @@
...
+import org.javarosa.core.services.locale.Localization;
+
+/**
+ * Created by amstone326 on 9/25/17.
+ */
+
+public class QuarantineUtil {
+
+    public static String getQuarantineReasonDisplayString(FormRecord r, boolean includeDetail) {
+        String displayString = "";
```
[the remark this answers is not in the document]
ctsims: FWIW: Java is going to compile this into a string builder at compile time anyway, so I don't think there's a specific reason to use one, given taht this seems more readable.

app/src/org/commcare/android/database/user/models/FormRecord.java
amstone326: Yes.. There have been so many damn migrations in this class and I just couldn't bring myself to do another one for a not-very-essential String field...

app/src/org/commcare/network/CommcareRequestGenerator.java
amstone326: I assume you mean like a history of when it changed and what the change was for? If so, I don't know of one existing.

app/src/org/commcare/tasks/ProcessAndSendTask.java
amstone326: @ctsims That's not at all what this method checks for. Go look at its implementation (which is not something I wrote; I just refactored the code that was already there to be in its own method). All we have ever done here is to check what type of error caused the last submission to fail, and stop the process if it was an error type that indicates future submissions are also likely to fail for the same reason. There's nothing in our code that checks the type of the form that failed and employs any logic based on that.

app/src/org/commcare/network/CommcareRequestGenerator.java
ctsims: Aha, found it at the bottom of
https://github.com/dimagi/commcare-core/wiki/SyncAndRestoreAPI

which defintely doesn't seem like the right place. I'll move it over

ctsims: updated  here: https://github.com/dimagi/commcare-core/wiki/OpenRosaRequest

does this sound right?

app/src/org/commcare/tasks/ProcessAndSendTask.java
ctsims: Hey, sorry, wasn't trying to make any assertions about type.

I was trying to say that the the name seems slightly off. The word "Predicts" implies that the reason we should stop submitting is because the next submission will fail, but the reason we should stop submitting is that the protocol requires that we send form N before we send form N+1 (the create/followup example was just meant to illustrate that, sorry for confusing.

amstone326:
Product Note: HQ has changed the response code that is sent when it receives a form, but then encounters an error while processing it in such a way that the form did not take effect in the database. When mobile receives this response code, it will now quarantine the form and record with that quarantined record the reason why.

PR #1722, opened 2017-05-24
Treat InputIOExceptions during form submission as a record failure

amstone326:
fyi @ctsims

### DIFF
--- a/app/src/org/commcare/utils/FormUploadUtil.java
+++ b/app/src/org/commcare/utils/FormUploadUtil.java
@@ -169,9 +169,19 @@ public class FormUploadUtil {
             return FormUploadResult.TRANSPORT_FAILURE;
         } catch (IOException | IllegalStateException e) {
             e.printStackTrace();
-            Logger.log(AndroidLogger.TYPE_ERROR_STORAGE,
-                    "Error reading form during submission: " + e.getMessage());
-            return FormUploadResult.TRANSPORT_FAILURE;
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/utils/FormUploadUtil.java
```diff
@@ -169,9 +169,19 @@ private static FormUploadResult submitEntity(MultipartEntity entity, String url,
             return FormUploadResult.TRANSPORT_FAILURE;
         } catch (IOException | IllegalStateException e) {
             e.printStackTrace();
-            Logger.log(AndroidLogger.TYPE_ERROR_STORAGE,
-                    "Error reading form during submission: " + e.getMessage());
-            return FormUploadResult.TRANSPORT_FAILURE;
+            if (e instanceof InputIOException) {
```
ctsims: Is this caught above by the  } catch (InputIOException ioe) { line?

ctsims:
Confirming that the issue appears to be that when we moved from writeFromInputToOutput to writeFromInputToOutputNew in StreamsUtil  (in submitEntity from FormUploadUtil), the new method didn't implement the directional stream exceptions, We can target 2.37 for that fix.

amstone326:
@ctsims Can you open a ticket explaining what the issue is in a little bit more detail and assign it to the mobile backlog?

ctsims:
done http://manage.dimagi.com/default.asp?255101

PR #1723, opened 2017-05-24
Enable detail nodeset sorting

amstone326:
cross-request: https://github.com/dimagi/commcare-core/pull/566

I've tested/confirmed that this ensures: 
a) Proper sorting behavior in `EntitySelectActivity` is preserved
b) If a `<sort>` element is added to a `<field>` within a sub-detail that uses a nodeset, the elements of the nodeset get sorted by that field 

~~Holding off on merge until the HQ side has been implemented and I can test with that.~~
Since the timeline and person responsible for the HQ side of this is currently unclear, Jenny and I decided it makes sense to consider this ready and have it merged now. I've tested that it works as long as HQ provides the correct suite formatting, which has been specified to them.

### DIFF
--- a/app/src/org/commcare/activities/EntitySelectActivity.java
+++ b/app/src/org/commcare/activities/EntitySelectActivity.java
@@ -790,14 +790,6 @@ public class EntitySelectActivity extends SaveSessionCommCareActivity
                                   NodeEntityFactory factory, int focusTargetIndex) {
         loader = null;
 
-        int[] order = shortSelect.getSortOrder();
-        for (int i = 0; i < shortSelect.getFields().length; ++i) {
-            String header = shortSelect.getFields()[i].getHeader().evaluate();
-            if (order.length == 0 && !"".equals(header)) {
-                order = new int[]{i};
-            }
-        }
-
         AdapterView visibleView;
         if (shortSelect.shouldBeLaidOutInGrid()) {
             visibleView = ((GridView)this.findViewById(R.id.screen_entity_select_grid));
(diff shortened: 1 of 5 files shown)


app/src/org/commcare/activities/EntitySelectActivity.java
```diff
@@ -790,14 +790,6 @@ public void deliverLoadResult(List<Entity<TreeReference>> entities,
                                   NodeEntityFactory factory, int focusTargetIndex) {
         loader = null;
 
-        int[] order = shortSelect.getSortOrder();
```
amstone326: this was moved into `SortableEntityAdapter`

app/src/org/commcare/adapters/EntityListAdapter.java
```diff
@@ -44,11 +45,10 @@
  * @author ctsims
  * @author wspride
  */
-public class EntityListAdapter implements ListAdapter, EntitySortNotificationInterface {
+public class EntityListAdapter extends SortableEntityAdapter implements ListAdapter {
```
amstone326: All of the code deleted from this class was just moved into `SortableEntityAdapter`

app/src/org/commcare/interfaces/SortableEntityAdapter.java
```diff
@@ -0,0 +1,80 @@
...
+import org.commcare.views.notifications.NotificationMessageFactory;
+import org.javarosa.core.model.instance.TreeReference;
+
+import java.util.List;
+
+/**
+ * Created by amstone326 on 5/24/17.
+ */
+
+public abstract class SortableEntityAdapter implements EntitySortNotificationInterface {
```
wpride: Would be great if this class could live in `commcare-core` - looks like the `notifyBadFilter` override is the only android dependency

wpride:
@amstone326 approved, nice cleanup and clean implementation. Feel free to merge if you don't want to make that change (think I'll just have to later for Formplayer)

amstone326:
@wpride Hey yea sorry I'm not seeing a trivial way to do it so going to leave for now.

wpride:
@amstone326 One thing I missed - why is `SortableEntityAdapter` abstract?

amstone326:
@wpride Because it should never be instantiated on its own. It only has a very small fraction of the code necessary to be an actual working entity adapter, it just provides extra functionality to an existing one. Same concept as why something like `SessionAwareCommcareActivity` is abstract

PR #1724, opened 2017-05-26
2.36 hotfix - Several fixes to the new date widget

amstone326:
- Fixes part 1 of https://manage.dimagi.com/default.asp?253349 -- The max year for the years dropdown in the calendar widget was being restricted improperly
- Fixes https://manage.dimagi.com/default.asp?254923 -- The calendar fragment was not always using the year that had been set in the main widget UI
- Linting
- Fixes part 2 of https://manage.dimagi.com/default.asp?253349 + general UI improvements -- Made a bunch of changes to the layout files so that things stopped getting cut off, and also to just make the calendar fragment look better (it looked sort of janky before, before/after screenshots below). One of the things I ultimately decided to do in this process was to set up a way for widgets to request that the orientation of form entry be restricted to portrait while they are showing, and applied this to the new date widget. I think I found a decently clean way to do this, and it avoids what would otherwise be a good deal more work to make the calendar fragment look OK in landscape mode.

Calendar fragment before changes:
screenshot_20170529-203048

Calendar fragment after changes:
screenshot_20170529-202457


app/src/org/commcare/views/widgets/GregorianDateWidget.java
```diff
@@ -370,7 +366,9 @@ private void refreshDisplay() {
 
     private void openCalendar() {
         setFocus(getContext());
-        timeBeforeCalendarOpened = calendar.getTimeInMillis();
+        timeBeforeCalendarOpened = getCurrentMillis();
```
amstone326: Fix number 2: The `timeBeforeCalendarOpened` should be set based on the current UI of the main widget, and the calendar that underlies the calendar fragment should be updated to use the same

app/src/org/commcare/views/widgets/CalendarFragment.java
```diff
@@ -164,9 +164,15 @@ public void onNothingSelected(AdapterView<?> parent) {}
         });
     }
 
+    private int getMaxYearForCalendar() {
```
amstone326: Fix number 1: the max year for the year dropdown in the calendar fragment should be set based on the actual current year (like the current phone's date), not the currently-selected year

app/src/org/commcare/views/widgets/GregorianDateWidget.java
amstone326: Fix number 2: The `timeBeforeCalendarOpened` should be set based on the current UI of the main widget, and the calendar that underlies the calendar fragment should be updated to use the same

app/src/org/commcare/views/widgets/CalendarFragment.java
```diff
@@ -46,45 +47,44 @@
...
         setWindowSize();
 
         return layout;
     }
 
-    private void initWeekDays(){
+    @Override
+    public void onViewStateRestored(@Nullable Bundle savedInstanceState) {
+        super.onViewStateRestored(savedInstanceState);
+        refresh();
```
amstone326: Fix number 3: Don't call refresh() until after the prior state of the fragment has settled, or else that process overrides the new values

app/src/org/commcare/views/widgets/GregorianDateWidget.java
```diff
@@ -46,21 +46,20 @@
...
-    private final int maxYear;
     private final long todaysDateInMillis;
     private long timeBeforeCalendarOpened;
 
     private final CalendarFragment myCalendarFragment;
     private final FragmentManager fm;
 
-    public static final int MINYEAR = 1900;
+    public static final int MIN_YEAR = 1900;
+    public static final int MAX_YEAR = 2100;
```
amstone326: Fix number 1: Instead of setting the max year based upon an offset from the currently-selected year, just use the same hard-coded max that the default Android date-picker uses

PR #1727, opened 2017-05-31
2.36 icds hotfix

amstone326:
if domain is icds, set custom callout to assume plain text values unless told otherwise

cross-request: https://github.com/dimagi/commcare-core/pull/571

### DIFF
--- a/app/src/org/commcare/activities/EntitySelectActivity.java
+++ b/app/src/org/commcare/activities/EntitySelectActivity.java
@@ -28,6 +28,7 @@ import org.commcare.CommCareApplication;
 import org.commcare.activities.components.EntitySelectCalloutSetup;
 import org.commcare.activities.components.EntitySelectViewSetup;
 import org.commcare.adapters.EntityListAdapter;
+import org.commcare.android.logging.ReportingUtils;
 import org.commcare.cases.entity.Entity;
 import org.commcare.cases.entity.NodeEntityFactory;
 import org.commcare.dalvik.R;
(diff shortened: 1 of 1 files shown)


amstone326:

amstone326:

wpride:
This hackiness seems pretty unfortunate though I'm assuming you guys had good reason - curious how you came to this decision? cc @ctsims

amstone326:
@wpride So the high-level biggest-blocker that we (eventually) realized we have is that, because there is currently no way for the ICDS team to ensure that all users get a .ccz update before they start updating to 2.36, any solution we choose has to work without making _any_ changes to the ICDS app. With that constraint in mind, we were considering 2 options:

1. (my original plan) Try to parse all values as XPath, and if they fail, assume that they were intended to be plain text values. This felt shitty to me because it meant that if someone messed up in writing something that they actually intended to be an XPath expression, we would just silently accept it.

2. (the plan we chose) Assume, for ICDS only, that all values are plain text, unless the key explicitly says that they are xpath. We decided on this because it satisfies the constraint above (in that the existing state of the app will work), but also gives them the flexibility to use xpath values later on.

@ctsims Having written both of those out, it does occur to me that we could have gone with a sort of hybrid, where we do option 1, but restrict it to the ICDS domain only in the same way we have for option 2. It does seem like a less complicated approach in that it doesn't involve the ICDS team remembering to do anything extra if/when they do want to use an xpath value in the future. Thoughts on that?

ctsims:
@amstone326 I'm fine signing the ICDS team up for a more complex transition in the future at the cost of not treating an xpath parse failure as a flat value, I think the risk of the latter is much higher.

My guess is that the ICDS team is unlikely to need much from this functionality in the future, so the transition is likely to occur when when force them onto 2.36 and can remove the gross code from future releases v. the ICDS team intentionally doing something on their side and needing XPath.

amstone326:
@ctsims ok sounds good to me. this is ready then.

PR #1738, opened 2017-06-16
Make back button collapse an expanded persistent case tile

amstone326:
Fix for https://manage.dimagi.com/default.asp?256304

This overrides `onBackPressed` in `CommCareActivity` to first check whether the persistent case tile is currently expanded. If it is, we will collapse the tile instead of moving back a screen.

fyi @sheelio


wpride:
👍 awesome

PR #1740, opened 2017-06-16
Log Graphing actions

wpride:
For Neal and Rashmi's CHW Performance Graphing analysis


amstone326:
Interesting, why did they want this if we have it in Google Analytics already?

amstone326:
@wpride Does that mean we should delete the GA logging for graphing? Or is there still utility in that as well?

wpride:
@amstone326 I think they wanted to keep that too

PR #1744, opened 2017-06-22
2.36 changes to Prompted Update feature

amstone326:
This PR implements the changes to the Prompted Update feature requested by the ICDS team based on field testing. The spec for those changes is [here](https://docs.google.com/document/d/1LtjJmsKAQdHRgjY_9c9_JO1Yx0CkuZ12CWss9zlExXo/edit). They included mockups at the end, which I have matched as closely as possible. (The only thing I changed was the ordering of the cue image and the "update later" button because I thought it made a bit more sense, but I'm going to confirm with Saket/Srikant that this is OK).

screenshot_20170711-221724
screenshot_20170711-221728

This PR also includes 1 other small change not specifically requested by the ICDS team. The feedback from the ICDS team regarding the "force" component of this feature was that they would not ever want to use it because it would be too difficult/confusing for users who frequently operate in low-connectivity environments. Since the "forced" setting is configurable anyway, I have not removed the capability from the code entirely. But based on this feedback in conjunction with a UX discussion I had with Jenny about making the HQ configurability as easily as possible, I have decided to change the manner of configuration for the "forced" component from expecting a date value (`"force-by-date"`), to instead expecting just a boolean value (`"force"`).


app/src/org/commcare/services/CommCareSessionService.java
```diff
@@ -109,6 +109,9 @@
     private HeartbeatLifecycleManager heartbeatManager;
     private boolean heartbeatSucceededForSession;
 
+    public boolean cczUpdatePromptWasShown;
```
ctsims: These should be private and have getters

PR #1745, opened 2017-06-22
Add support for markdown templates in case detail

shubham1g5:
Case url: https://manage.dimagi.com/default.asp?255261#1352893

Added setUpMarkdown function to support Markdown templates.  Also extracted out the code for simple text template in a different function called setUpText for better modularity.

### DIFF
--- a/app/src/org/commcare/views/EntityDetailView.java
+++ b/app/src/org/commcare/views/EntityDetailView.java
@@ -115,13 +115,13 @@ public class EntityDetailView extends FrameLayout {
                             int index, int detailNumber) {
         super(context);
 
-        detailRow = (LinearLayout) View.inflate(context, R.layout.component_entity_detail_item, null);
-        label = (TextView) detailRow.findViewById(R.id.detail_type_text);
-        spacer = (TextView) detailRow.findViewById(R.id.entity_detail_spacer);
-        data = (TextView) detailRow.findViewById(R.id.detail_value_text);
+        detailRow = (LinearLayout)View.inflate(context, R.layout.component_entity_detail_item, null);
+        label = (TextView)detailRow.findViewById(R.id.detail_type_text);
+        spacer = (TextView)detailRow.findViewById(R.id.entity_detail_spacer);
+        data = (TextView)detailRow.findViewById(R.id.detail_value_text);
         currentView = data;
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/views/EntityDetailView.java
```diff
@@ -213,6 +214,24 @@ public void setParams(Detail d, Entity e, int index, int detailNumber) {
         }
     }
 
+    private boolean setUpText(String textField) {
+        data.setText((textField));
+        updateCurrentView(TEXT, data);
+        return isTextVeryLong(textField);
+    }
+
+    private boolean isTextVeryLong(String textField) {
```
amstone326: :+1: :+1:

app/src/org/commcare/views/EntityDetailView.java
```diff
@@ -111,26 +115,26 @@ public EntityDetailView(Context context, Detail d, Entity e,
                             int index, int detailNumber) {
         super(context);
 
-        detailRow = (LinearLayout)View.inflate(context, R.layout.component_entity_detail_item, null);
-        label = (TextView)detailRow.findViewById(R.id.detail_type_text);
-        spacer = (TextView)detailRow.findViewById(R.id.entity_detail_spacer);
-        data = (TextView)detailRow.findViewById(R.id.detail_value_text);
+        detailRow = (LinearLayout) View.inflate(context, R.layout.component_entity_detail_item, null);
```
amstone326: Same note on casts as the other PR. Will show you how to make this stop happening as an interim measure

shubham1g5: Thanks. Was able to figure this out. Reformatted the code with cast spacing removed.

amstone326:
@ctsims this is good-to-go from my perspective, but you had 1 comment in the ticket that I'm not sure what to make of:

```Note that a Hyperlink text might look weird in the detail page due to color matching, so we should likely apply some sort of styling to hyperlinks in that context so they will look good against that background.```

I tried this on a phone and it looks fine to me, since the background will only ever be white or gray and the URL is blue. Is there something more you want done styling-wise?

ctsims:
@amstone326 If it looks fine in both configurations that sounds good to me! Merging

amstone326:
Release Note: Add support for markdown in case details

PR #1746, opened 2017-06-22
Memory Tuning for Optimizations

ctsims:
See cross requested PR: Implements hooks which switch async entity list loads to _not_ use bulk optimizations inside of the entity load (since they load one by one no matter what), and allow different amounts of heap memory to be used based on device.

cross-request: https://github.com/dimagi/commcare-core/pull/596

### DIFF
--- a/app/src/org/commcare/engine/cases/AndroidCaseInstanceTreeElement.java
+++ b/app/src/org/commcare/engine/cases/AndroidCaseInstanceTreeElement.java
@@ -16,6 +16,7 @@ import org.commcare.models.database.SqlStorage;
 import org.commcare.models.database.user.models.AndroidCaseIndexTable;
 import org.commcare.modern.engine.cases.CaseGroupResultCache;
 import org.commcare.modern.engine.cases.CaseIndexQuerySetTransform;
+import org.commcare.modern.engine.cases.CaseObjectCache;
 import org.commcare.modern.engine.cases.query.CaseIndexPrefetchHandler;
 import org.commcare.modern.util.PerformanceTuningUtil;
 import org.javarosa.core.model.instance.AbstractTreeElement;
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/engine/cases/AndroidCaseInstanceTreeElement.java
```diff
@@ -278,23 +281,47 @@ public boolean isReferencePatternCachable(TreeReference ref) {
...
-        if(caseGroupCache != null && caseGroupCache.hasMatchingCaseSet(recordId)) {
-            if(!caseGroupCache.isLoaded(recordId)) {
-                EvaluationTrace loadTrace = new EvaluationTrace("Bulk Case Load");
+
+        CaseObjectCache caseObjectCache = getCaseObjectCacheIfRelevant(context);
+
+        if (caseObjectCache != null &&
+                (caseObjectCache.isLoaded(recordId) || canLoadCaseFromGroup(caseGroupCache, recordId))) {
+
+            if (!caseObjectCache.isLoaded(recordId)) {
```
amstone326: I think this can be simplified by adding this into to the above `if` statement

ctsims:
@amstone326 All PR Feedback Integrated

ctsims:
Pulling since PR feedback was integrated. 

@wspride Note that these changes will require minor updates to formplayer to match. Specifically 

https://github.com/dimagi/commcare-android/pull/1746/files#diff-1bb1940c04c07c8c092b902410858656R283

Will need to be updated. My guess is your implementation more or less is a copy/paste of this method anyway, so probably fine to re-copy/paste (should still be quite portable).

This should let formplayer run this stuff even faster though, since it'll let it have more memory for optimization.

wpride:
thanks @ctsims

PR #1748, opened 2017-06-25
Generalize text for entity list exception

shubham1g5:
Case URL: https://manage.dimagi.com/default.asp?255119#1353059
cross-request: https://github.com/dimagi/commcare-core/pull/600

Since the exception can occur due to sort expression as well saying 'Filtering your list failed with a bad filter expression' is a bit misleading'.


wpride:
Good stuff. Man, the usage of XPathSyntaxException is a disaster - looks like its almost always either re-thrown as a Runtime exception or swallowed and ignored entirely. Definite good future refactor candidate.

PR #1749, opened 2017-06-27
Migrate all instances of PreferenceActivity to use PrferenceFragmentCompat

shubham1g5:
Case Url: https://manage.dimagi.com/default.asp?253212#1354692

- All preferences now use the common CommCarePreferenceFragment which serves as a super class and handles the common functionalities.
- We have a single activity called CommCarePreferenceActivity which initialises the respective preference fragment type depending on the intent extra 'extra_pref_type'.
- SessionAwarePreferenceActivity just adds the session listeners on top of CommCarePreferenceActivity and should be used if the preference screen is accessible after login.

### DIFF
--- a/app/src/org/commcare/fragments/CommCarePreferenceFragment.java
+++ b/app/src/org/commcare/fragments/CommCarePreferenceFragment.java
@@ -28,15 +28,15 @@ public abstract class CommCarePreferenceFragment extends PreferenceFragmentCompa
 
     @Override
     public void onCreatePreferences(Bundle savedInstanceState, String rootKey) {
-        GoogleAnalyticsUtils.reportPrefActivityEntry(GoogleAnalyticsFields.CATEGORY_CC_PREFS);
-        setupUI();
-        initPrefs();
+        GoogleAnalyticsUtils.reportPrefActivityEntry(getAnalyticsCategory());
+        setTitle();
+        initPrefsFile();
         loadPrefs();
     }
 
     @CallSuper
-    protected void initPrefs() {
-        if(isAppLevelPreference()) {
+    protected void initPrefsFile() {
+        if (isPersistentAppPreference()) {
             PreferenceManager prefMgr = getPreferenceManager();
             prefMgr.setSharedPreferencesName((CommCareApplication.instance().getCurrentApp().getPreferencesFilename()));
         }
@@ -54,7 +54,7 @@ public abstract class CommCarePreferenceFragment extends PreferenceFragmentCompa
     }
 
     @CallSuper
-    protected void setupUI() {
+    protected void setTitle() {
         if (getActivity() != null) {
             getActivity().setTitle(getTitle());
         }
@@ -78,17 +78,19 @@ public abstract class CommCarePreferenceFragment extends PreferenceFragmentCompa
         }
     }
 
+
+
     @Override
-    public void onStart() {
-        super.onStart();
+    public void onResume() {
+        super.onResume();
         // register the preference change listener
         getPreferenceScreen().getSharedPreferences()
                 .registerOnSharedPreferenceChangeListener(this);
     }
 
     @Override
-    public void onStop() {
-        super.onStop();
+    public void onPause() {
+        super.onPause();
         // unregister the preference change listener
         getPreferenceScreen().getSharedPreferences()
                 .unregisterOnSharedPreferenceChangeListener(this);
@@ -104,16 +106,18 @@ public abstract class CommCarePreferenceFragment extends PreferenceFragmentCompa
     }
 
     /**
-     *
      * @return whether the preference should be stored in app specific preference file.
      */
-    protected boolean isAppLevelPreference() {
+    protected boolean isPersistentAppPreference() {
         return false;
     }
 
 
     @NonNull
     protected abstract String getTitle();
+
+    @NonNull
+    protected abstract String getAnalyticsCategory();
 
     @Nullable
     protected abstract Map<String, String> getPrefKeyAnalyticsEventMap();
@@ -123,7 +127,6 @@ public abstract class CommCarePreferenceFragment extends PreferenceFragmentCompa
     @Nullable
     protected abstract Map<String, String> getPrefKeyTitleMap();
 
-    @NonNull
     protected abstract int getPreferencesResource();
 
     @Override
--- a/app/src/org/commcare/fragments/CommCarePreferenceFragment.java
+++ b/app/src/org/commcare/fragments/CommCarePreferenceFragment.java
@@ -1,0 +1,118 @@
+package org.commcare.fragments;
+
+import android.content.SharedPreferences;
+import android.os.Bundle;
+import android.support.annotation.CallSuper;
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+import android.support.v7.preference.PreferenceFragmentCompat;
+import android.support.v7.preference.PreferenceManager;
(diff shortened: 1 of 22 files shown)


app/src/org/commcare/fragments/CommCarePreferenceFragment.java
```diff
@@ -0,0 +1,131 @@
...
+    protected void reset() {
+        getPreferenceScreen().removeAll();
+        loadPrefs();
+    }
+
+    /**
+     *
+     * @return whether the preference should be stored in app specific preference file.
+     */
+    protected boolean isAppLevelPreference() {
```
amstone326: Can we call this `isPersistentAppPreference()` instead? There are some preference fragments that relate to app-level actions/settings, but return `false` for this method because we don't want to store them in a persistent prefs file

app/src/org/commcare/preferences/CommCarePreferences.java
```diff
@@ -167,40 +151,39 @@ private void hideServerPrefsIfNeeded() {
...
-    private void setupButtons() {
+    @Override
+    protected boolean isAppLevelPreference() {
+        return true;
+    }
+
+    @Override
+    public void onStart() {
+        super.onStart();
+        setVisibilityOfUpdateOptionsPref();
```
amstone326: Have the same question regarding moving this from `onResume()` to `onStart()` as I commented on `CommCarePreferenceFragment`; feel free to answer it in just once place, I just want to flag this to remember to come back to it if we want it to change

app/src/org/commcare/preferences/CommCareServerPreferences.java
```diff
@@ -45,32 +48,43 @@
...
+    @Override
+    protected int getPreferencesResource() {
+        return R.xml.server_preferences;
+    }
 
-        GoogleAnalyticsUtils.createPreferenceOnClickListeners(prefMgr, prefKeyToAnalyticsEvent,
-                GoogleAnalyticsFields.CATEGORY_SERVER_PREFS);
+    @Override
+    public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, String key) {
+        // No listeners
```
amstone326: Can you add this same comment to the other empty implementations of this method? I think it's useful to make it clear that it's intentional

app/src/org/commcare/preferences/DeveloperPreferences.java
```diff
@@ -72,45 +75,87 @@
...
 
+    @NonNull
     @Override
-    protected void onCreate(Bundle savedInstanceState) {
-        super.onCreate(savedInstanceState);
-        GoogleAnalyticsUtils.reportPrefActivityEntry(GoogleAnalyticsFields.CATEGORY_DEV_PREFS);
-        populatePrefKeyToEventLabelMapping();
-        initAllPrefs();
+    protected String getTitle() {
+        return "Developer Options";
```
amstone326: I realize that this was like this before, but can you make this title localizable like all of the others?

app/src/org/commcare/preferences/DeveloperPreferences.java
```diff
@@ -72,45 +75,87 @@
...
+
+    @Override
+    protected boolean isAppLevelPreference() {
+        return true;
+    }
+
+    @Override
+    public void onStart() {
+        super.onStart();
+        hideOrShowDangerousSettings();
```
amstone326: same question here about the move from `onResume` to `onStart`

app/src/org/commcare/fragments/CommCarePreferenceFragment.java
```diff
@@ -0,0 +1,131 @@
...
+ * Common PreferenceFragment class from which all other preferences extend.
+ */
+public abstract class CommCarePreferenceFragment extends PreferenceFragmentCompat
+        implements SharedPreferences.OnSharedPreferenceChangeListener {
+
+    private static final String TAG = CommCarePreferenceFragment.class.getSimpleName();
+
+    @Override
+    public void onCreatePreferences(Bundle savedInstanceState, String rootKey) {
+        GoogleAnalyticsUtils.reportPrefActivityEntry(GoogleAnalyticsFields.CATEGORY_CC_PREFS);
```
amstone326: If you look at all of the calls to `reportPrefActivityEntry()` that were removed in this PR, you'll notice that each preference activity used to invoke this call with its own specific category. Can you switch it back to doing that? Probably makes sense to do it in the same way that you've implemented setting a different title for each fragment

shubham1g5: Yeah. Missed it somehow. Great catch 👍

app/src/org/commcare/preferences/AdvancedActionsPreferences.java
```diff
@@ -71,26 +78,38 @@
...
+    @NonNull
     @Override
-    protected void onCreate(Bundle savedInstanceState) {
-        super.onCreate(savedInstanceState);
+    protected String getTitle() {
+        return Localization.get("settings.advanced.title");
+    }
 
-        addPreferencesFromResource(R.xml.advanced_actions);
+    @Nullable
```
shubham1g5: Yeah. Android Studio does add all annotations from base function automatically and gives a warning if we omit them. So it definitely seems like annotations doesn't get inherited automatically in Android Studio.

app/src/org/commcare/fragments/CommCarePreferenceFragment.java
```diff
@@ -0,0 +1,131 @@
...
+                }
+            }
+        }
+    }
+
+    @Override
+    public void onStart() {
+        super.onStart();
+        // register the preference change listener
+        getPreferenceScreen().getSharedPreferences()
```
shubham1g5: No. It's a mistake on my part. User can only interact with the UI after onResume been called while he can see it at the onStart stage. So will move all listeners calls to register/unregister  back to onResume/onPause.

app/src/org/commcare/preferences/CommCarePreferences.java
shubham1g5: For this I think onStart() is a better place as it's related to visibility rather than any ui interaction and as UI becomes visible in OnStart() we should not defer it to onResume.

app/src/org/commcare/preferences/DeveloperPreferences.java
shubham1g5: Again would prefer onStart() for it as we are setting up what prefs to show here which should be done as soon as UI becomes visible.

shubham1g5:
@amstone326 Made the requested changes.

app/src/org/commcare/preferences/CommCarePreferences.java
amstone326: Ah cool good reasoning, agreed

amstone326:
@shubham1g5 Awesome, thanks for the changes. 1 last thing-- If you haven't already, can you spend a few minutes testing the persistence/functionality of at least 1 setting in each preference fragment, just to make sure nothing went awry during the refactor? Let me know once that's confirmed and then I'm ready to merge.

shubham1g5:
@amstone326 I have done some preliminary testing and can confirm that preferences are working on Android 7.0 and Android 5.1. Though I wanted to test them on pre 4.0 devices just for assurance though it's quite hard to get hold of one.

amstone326:
Hey, I have a 4.2 device which I just tested on and everything looks good, so going to pull this!

PR #1751, opened 2017-06-29
Add stop feature to upload sync

shubham1g5:
Case url : https://manage.dimagi.com/default.asp?252457#1355469

- Add Stop button to send forms dialog
- Checks while sending forms if task is cancelled, if yes throws the TaskCancelledException to return FormUploadResult.FAILURE as result

### DIFF
--- a/app/src/org/commcare/activities/SyncCapableCommCareActivity.java
+++ b/app/src/org/commcare/activities/SyncCapableCommCareActivity.java
@@ -356,12 +356,14 @@ public abstract class SyncCapableCommCareActivity<T> extends SessionAwareCommCar
                 title = Localization.get("sync.progress.submitting.title");
                 message = Localization.get("sync.progress.submitting");
                 dialog = CustomProgressDialog.newInstance(title, message, taskId);
+                dialog.addCancelButton();
                 break;
             case ProcessAndSendTask.PROCESSING_PHASE_ID:
                 title = Localization.get("form.entry.processing.title");
                 message = Localization.get("form.entry.processing");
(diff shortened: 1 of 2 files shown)


amstone326:
@shubham1g5 Sweet, so fast! Did it turn out not to be necessary from a UI perspective to use a different `FormUploadResult` for cancellations?

shubham1g5:
@amstone326 It did not. There was no toast appearing on cancelling the task. So I just used the current FormUploadResult.FAILURE . Though we were assuming that a cancellation will only happen in case of a logout or when a session expires and were displaying a notification to the user saying the process interrupted due to a logout and will resume on login. So I added a check for login before showing the user here - https://github.com/dimagi/commcare-android/blob/d355f5d9fccc9ef993ec5667d5cebfd0a46a4847/app/src/org/commcare/tasks/ProcessAndSendTask.java#L512 . Though I am not sure if it's the right way to check if user is logged in or not!

amstone326:
@shubham1g5 Cool sounds good, and the way you did that looks right to me!

1 small request -- Can we add a toast that gets shown after manual cancellation that tells the user how many forms did get successfully sent before they cancelled?

shubham1g5:
@amstone326 Yeah. We can do that. We are only logging that number right now. Should I restore this branch ?

amstone326:
@shubham1g5 Fine to do it on this branch or a new one, whatever you prefer

amstone326:
Announce: There are some projects (mainly TDH) who will be really excited to hear about this. Should include in some sort of internal announcement after 2.38 is released

PR #1752, opened 2017-06-30
Managed Configurations

wpride:
Enables Google EMM to configure CommCare installation and login. 

Targeting for 2.38, ~~I'll need to release this on the Play Store (under some other package) in order to test~~ testing done and confirmed working


amstone326:
"I'll need to release this on the Play Store (under some other package) in order to test" @wpride Have you already done this? If not, is it safe to include this code in 2.38 before you have?

app/build.gradle
```diff
@@ -293,6 +299,11 @@ android {
             // use the settings from defaultConfig
         }
 
+        managed {
```
amstone326: What does this build flavor represent? I'm not following why this implementation would need something like this

app/src/org/commcare/activities/CommCareSetupActivity.java
```diff
@@ -933,4 +941,24 @@ public static String getAnalyticsActionFromInstallMode(int installModeCode) {
...
+            RestrictionsManager restrictionsManager =
+                    (RestrictionsManager) getSystemService(Context.RESTRICTIONS_SERVICE);
+            Bundle appRestrictions = restrictionsManager.getApplicationRestrictions();
+            if (appRestrictions.containsKey("profileUrl")) {
+                Toast.makeText(this,  "Found managed configuration install URL "
+                        + appRestrictions.getString("profileUrl"), Toast.LENGTH_LONG).show();
+                Log.d(TAG, "Found managed configuration install URL "
+                        + appRestrictions.getString("profileUrl"));
+                incomingRef = appRestrictions.getString("profileUrl");
+                uiState = UiState.READY_TO_INSTALL;
```
amstone326: Can you create a new install mode value for managed installations (i.e. like `INSTALL_MODE_BARCODE` and `INSTALL_MODE_URL`), and then add a line here that sets the value of `lastInstallMode` to that?

app/src/org/commcare/activities/CommCareSetupActivity.java
```diff
@@ -933,4 +941,24 @@ public static String getAnalyticsActionFromInstallMode(int installModeCode) {
...
+
+    private void checkManagedConfiguration() {
+        Log.d(TAG, "Checking managed configuration");
+        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.LOLLIPOP) {
+            // Check for managed configuration
+            RestrictionsManager restrictionsManager =
+                    (RestrictionsManager) getSystemService(Context.RESTRICTIONS_SERVICE);
+            Bundle appRestrictions = restrictionsManager.getApplicationRestrictions();
+            if (appRestrictions.containsKey("profileUrl")) {
+                Toast.makeText(this,  "Found managed configuration install URL "
```
amstone326: Is there a reason to leave this toast in?

app/res/xml/app_restrictions.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<restrictions xmlns:android="http://schemas.android.com/apk/res/android">
```
amstone326: Where/how do the values for a restriction get set?

app/build.gradle
wpride: I needed to release to the Play Store because sending down the configuration `yaml` file only can happen through Play Store. Of course I couldn't do this on our production build so I added this flavor. Could remove now, but I'm sure I'll need to test new features in the future so I opted to leave in

app/res/xml/app_restrictions.xml
wpride: They're set via the EMM console that I've been working on - so by Google itself, essentially. Can read about them [here](https://developer.android.com/work/managed-configurations.html)

PR #1753, opened 2017-06-30
Db Migration for INDDEX work

amstone326:
cross-request: https://github.com/dimagi/commcare-core/pull/594

### DIFF
--- a/app/src/org/commcare/models/database/user/UserDatabaseUpgrader.java
+++ b/app/src/org/commcare/models/database/user/UserDatabaseUpgrader.java
@@ -531,7 +531,7 @@ class UserDatabaseUpgrader {
                         storageForThisFixture.iterate().nextRecord();
                 IndexedFixturePathUtils.buildFixtureIndices(db, tableName, new HashSet<>(exampleChildElement.getIndices()));
             }
-
+            db.setTransactionSuccessful();
             return true;
         } finally {
             db.endTransaction();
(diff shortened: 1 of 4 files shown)


app/src/org/commcare/models/database/user/UserDatabaseUpgrader.java
```diff
@@ -512,6 +518,29 @@ private boolean upgradeSeventeenEighteen(SQLiteDatabase db) {
...
+
+            // Step 2: Recreate table
+            IndexedFixturePathUtils.createStorageBackedFixtureIndexTable(db);
+
+            // Step 3: Re-parse the indexes for all fixtures in storage
+            IStorageUtilityIndexed<FormInstance> userFixtureStorage =
+                    CommCareApplication.instance().getUserStorage("fixture", FormInstance.class);
+            for (IStorageIterator<FormInstance> userFixtures = userFixtureStorage.iterate(); userFixtures.hasMore();) {
+                FormInstance fi = userFixtures.nextRecord();
+                // TODO: what now??
```
amstone326: Hey @ctsims- I ran into a bit of a blocker today in writing this migration; hoping that you have some insight that I'm missing. The idea here as I understand from our conversation is to wipe out the old fixture indexes table entirely, and then re-create it by re-parsing the old fixture format which we still have stored. From my reading of `[IndexedFixtureXmlParser](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/commcare/xml/IndexedFixtureXmlParser.java)`, it needs 3 essential pieces of information in order to parse (or re-parse) the fixture's indexes: the `TreeElement` representing the root of the fixture's xml tree, the name of the fixture, and the parsed `FixtureIndexSchema`. And while I can see how to get the first 2 things from the `FormInstance` that is stored in old fixture storage, I don't see anywhere that we are saving the `FixtureIndexSchema` in a way that it can be retrieved. Do you know if this is being saved somewhere that I'm not seeing?

ctsims:

PR #1755, opened 2017-07-03
Show forms sent on stopping the upload sync

shubham1g5:
Case Url: https://manage.dimagi.com/default.asp?252457#1356296

- Feedback from https://github.com/dimagi/commcare-android/pull/1751
- Shows a toast indicating no of forms sent before user presses stop to cancel the task.

### DIFF
--- a/app/assets/locales/android_translatable_strings.txt
+++ b/app/assets/locales/android_translatable_strings.txt
@@ -93,7 +93,7 @@ menu.print.detail=Print Case Detail
 menu.enable.privileges=Mobile Privileges
 menu.privilege.claim.disable=Disable Privileges
 
-sync.success.sent.singular=Form sent to server!
+sync.success.sent.singular=${0} form sent to server!
 sync.success.sent=${0} forms sent to server!
 sync.fail.unsent=Having issues communicating with the server to send forms. Will try again later.
 sync.fail.timeout=CommCare didn't receive a response from the remote service in time, it may be busy! Please wait a while and try your request again later.
@@ -329,6 +329,7 @@ title.datum.wrapper=[${0}]
 activity.task.error.generic=Unexpected error! Please try again.
 activity.task.cancelling.title=Cancelling: ${0}
 activity.task.cancelling.message=Cancelling...
+activity.task.cancelled.message=Task Cancelled.
 
 connection.test.prompt=This will try to fix connection problems that you may be having.
 connection.test.run=Run Connection Test
--- a/app/assets/locales/android_translatable_strings.txt
+++ b/app/assets/locales/android_translatable_strings.txt
@@ -93,7 +93,8 @@ menu.print.detail=Print Case Detail
 menu.enable.privileges=Mobile Privileges
 menu.privilege.claim.disable=Disable Privileges
 
-sync.success.sent.singular=${0} form sent to server!
+sync.success.sent.none=No forms sent to server!
+sync.success.sent.singular=1 form sent to server!
 sync.success.sent=${0} forms sent to server!
 sync.fail.unsent=Having issues communicating with the server to send forms. Will try again later.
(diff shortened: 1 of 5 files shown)


app/assets/locales/android_translatable_strings.txt
```diff
@@ -93,7 +93,7 @@ menu.print.detail=Print Case Detail
 menu.enable.privileges=Mobile Privileges
 menu.privilege.claim.disable=Disable Privileges
 
-sync.success.sent.singular=Form sent to server!
+sync.success.sent.singular=${0} form sent to server!
```
amstone326: Hey, I'm confused as to why anything needed to change here. The idea previously was that `sync.success.sent.singular` should be used if only 1 form was sent to the server (in which case no parameter for # of forms is needed since it's always just 1), and then `sync.success.sent` should be used if more than 1 form was sent.

app/src/org/commcare/tasks/ProcessAndSendTask.java
```diff
@@ -466,6 +467,18 @@ protected int getSuccessfulSends() {
         return successes;
     }
 
+    protected String getLabelForFormsSent() {
```
amstone326: Ah I see, you were just pulling out the code for doing this that was previously in `FormAndDataSyncer`. So in that original implementation, the mistake there was actually that there is no reason to pass a numeric parameter in the case that you are using `sync.success.sent.singular` key. So you should just remove the parameter in that case, and change that key back to what it was before

app/assets/locales/android_translatable_strings.txt
```diff
@@ -329,6 +329,7 @@ title.datum.wrapper=[${0}]
 activity.task.error.generic=Unexpected error! Please try again.
 activity.task.cancelling.title=Cancelling: ${0}
 activity.task.cancelling.message=Cancelling...
+activity.task.cancelled.message=Task Cancelled.
```
amstone326: Would prefer this to be a more user-friendly message, like "Form sending cancelled."

app/src/org/commcare/tasks/ProcessAndSendTask.java
shubham1g5: @amstone326 Ok. I was trying to be more explicit with a numeral value rather than letting user identify the singular 'Form', kind of more on lines of Android Plurals. But yeah can revert it back. Also what should be the case if user cancelled the task before any form was uploaded (in the form processing phase itself) - Right now we say '0 forms sent to the server'. We can change it to 'No forms sent to the server' if that sounds more correct.

shubham1g5:
@amstone326  Also here the reason behind why the toast message associated with FormUpload.Failure is not showing is because on cancelling a task onPostExecute never gets called and only onCancelled gets called. Another better way to handle cancellation of task here could be to just return `null` from different places in code on checking `isCancelled()` instead of throwing and catching the TaskCancelledException as what we return from doInBackground doesn't really get used in case of task getting cancelled.

app/assets/locales/android_translatable_strings.txt
shubham1g5: Agree. Will change.

app/src/org/commcare/tasks/ProcessAndSendTask.java
amstone326: Gotcha. If you want to change the static value of the message associated with `sync.success.sent.singular` to be "1 Form sent to the server!", that's definitely fine with me, but it doesn't make sense to be treating it like we're passing a dynamic value. 

Also agree that "No forms sent to the server" is slightly better than saying 0, so feel free to change that

shubham1g5: Agree. Made the change.

shubham1g5:
@amstone326 Made all requested changes. Please review again.

PR #1758, opened 2017-07-07
Set authority for profile install based on install mode

amstone326:
Fixes https://manage.dimagi.com/default.asp?257770. Previously we were blindly setting the resource authority for the Profile to always be local, instead of setting it intelligently based on the install mode being used. Somehow this wasn't causing any issues previously, but in the case of the demo user restore feature, which relies upon application profiles, it was resulting in a far-down-the-chain bug where we were incorrectly trying to contextualize a local relative ref against a remote parent ref, which resulted in installing the suite file from an incorrect URL that did not include the reference for the current application profile

cross-request: https://github.com/dimagi/commcare-core/pull/606


amstone326:

wpride:

wpride:

amstone326:

amstone326:

wpride:

app/src/org/commcare/activities/CommCareSetupActivity.java
```diff
@@ -508,6 +509,11 @@ protected void deliverError(CommCareSetupActivity receiver,
         }
     }
 
+    private int determineAuthorityForInstall() {
+        return lastInstallMode == INSTALL_MODE_OFFLINE ?
```
ctsims: Hm, strictly speaking the authority should be more about the reference connector than the specific method here. You can use the URL install or barcode install to direct to a local file, for instance.

I don't think it's important enough to drag everything out now, but would be worth at least documenting that limitation here as a cue for someone doing a bugfix in the future.

PR #1764, opened 2017-07-14
Fix Case tiles in 2.37 Release

ctsims:
Fixes bugs introduced in:
https://github.com/dimagi/commcare-android/pull/1738

where all case tiles (regardless of inline/persistent/etc) wouldn't attach correctly and come out empty:

pre (all tiles look like this):
image

post:
image

post:
image


wpride:
@ctsims is there a separate PR for master?

ctsims:
@wpride should get addressed in backmerge, 2.37 didn't get released due to this, figuring out plan right now.

PR #1766, opened 2017-07-20
Normalize DB Transaction Formats

ctsims:
Many of our DB writes were not using safe transaction formatting.

Most of these are trivial changes, but the HybridFileBackedSqlStorage update should get a deeper look, as its transaction format was very dangerously configured.

### DIFF
--- a/app/src/org/commcare/models/database/HybridFileBackedSqlStorage.java
+++ b/app/src/org/commcare/models/database/HybridFileBackedSqlStorage.java
@@ -296,7 +296,9 @@ public class HybridFileBackedSqlStorage<T extends Persistable> extends SqlStorag
             if (dataFilePath != null) {
                 HybridFileBackedSqlHelpers.unsetFileAsOrphan(db, dataFilePath);
             }
-            db.setTransactionSuccessful();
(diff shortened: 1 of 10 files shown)


app/src/org/commcare/models/database/HybridFileBackedSqlStorage.java
```diff
@@ -292,11 +296,15 @@ public void write(Persistable persistable) {
             if (dataFilePath != null) {
                 HybridFileBackedSqlHelpers.unsetFileAsOrphan(db, dataFilePath);
             }
-            db.setTransactionSuccessful();
+            if(startedTransaction) {
```
shubham1g5: We can use db.inTransaction() instead of maintaining a boolean ourselves.

ctsims: We'll actually be in a transaction either way, quite likely, since there is often a transaction wrapping this method call. This would be evaluating the status of a secondary transaction.

The reason we need to use a transaction in this method independent of the outer transaction is that there's a specific secondary (non-database call) side effect being tracked from the Orphan handlers.

PR #1768, opened 2017-07-24
Query Fallback Tests

ctsims:
Tests for:
https://github.com/dimagi/commcare-core/pull/618

Can't be fully set up right now due to lack of instrumentation, but should guarantee execution of all codepaths. 

cross-request: https://github.com/dimagi/commcare-core/pull/618


ctsims:

amstone326:
@ctsims I'm not seeing where the new code is actually being tested. Is your PR comment meant to say that it currently doesn't?

ctsims:
@amstone326 The code isn't currently "tested" in the sense that the code doesn't explicitly require the correct optimizations to be used (IE: there are no asserts that the right optimizations were run).

The tests _does_ however ensure that the code paths that are written will be executed. Until I can write a query analyser, this is a good middle ground to ensure that at the very least a bug can't sneak in which crashes the optimization.

PR #1769, opened 2017-07-25
Replaces Apache HTTP with Retrofit

shubham1g5:
Case: https://manage.dimagi.com/default.asp?257212

- `HTTPRequestGenerator` now uses `ModernHTTPRequestor` (which is now retrofit based) to make all requests
- There are no processRedirection callbacks now as we always follow valid redirects at the HTTP client level and therefore never gets a 301 from the http client which would have triggered this callback.

cross-request: https://github.com/dimagi/commcare-core/pull/619/files


shubham1g5:

shubham1g5:

amstone326:
@shubham1g5 This has conflicts with master now; can you resolve them and then I'll jump into reviewing this? Probably good to back-merge master into the commcare-core part of this PR as well

shubham1g5:

shubham1g5:

shubham1g5:

amstone326:
@shubham1g5 Would you mind separating out the "Moves LogTypes from AndroidLogger to Commcare Core" part of this into another PR? It seems unfortunate to make this diff harder to read when it's an important/large change and needs to be reviewed carefully.

shubham1g5:

app/src/org/commcare/network/HttpUtils.java
```diff
@@ -20,11 +47,25 @@
...
-            SharedPreferences prefs = CommCareApplication.instance().getCurrentApp().getAppPreferences();
-            if (prefs.contains(HttpRequestGenerator.USER_DOMAIN_SUFFIX)) {
-                domain = prefs.getString(HttpRequestGenerator.USER_DOMAIN_SUFFIX, null);
-            }
+            domain = CommCarePreferences.getUserDomain();
         }
         return new Pair<>(user, domain);
     }
+
+    public static String buildAppPassword(String password) {
```
amstone326: can be private

app/src/org/commcare/network/HttpUtils.java
```diff
@@ -20,11 +47,25 @@
...
+        if (DeveloperPreferences.useObfuscatedPassword()) {
+            return CredentialUtil.wrap(password);
+        }
+        return password;
+    }
+
+    public static String buildDomainUser(String username) {
```
amstone326: can be private

app/src/org/commcare/preferences/CommCareServerPreferences.java
```diff
@@ -88,6 +88,10 @@ public static String getSupportEmailAddress() {
         return getServerProperty(PREFS_SUPPORT_ADDRESS_KEY, CommCareApplication.instance().getString(R.string.support_email_address_default)) ;
     }
 
+    public static String getDataServerKey() {
```
amstone326: This can potentially be removed, depending on the answer to my question in `HttpRequestGenerator` about why this is getting used.

app/src/org/commcare/tasks/ModernHttpTask.java
```diff
@@ -5,17 +5,20 @@
...
+import org.commcare.core.network.HTTPMethod;
 import org.commcare.core.network.ModernHttpRequester;
 import org.commcare.modern.util.Pair;
-import org.commcare.network.AndroidModernHttpRequester;
 import org.commcare.tasks.templates.CommCareTask;
-import org.commcare.utils.AndroidCacheDirSetup;
 
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
```
amstone326: unused

app/src/org/commcare/utils/FormUploadUtil.java
```diff
@@ -208,10 +208,10 @@ private static void logResponse(int responseCode, String responseString) {
 
     /**
      * Validate the content body of the XML submission file.
-     *
+     * <p>
```
amstone326: Guessing these `<p>` additions were auto-generated somehow?

app/unit-tests/src/org/commcare/android/mocks/ModernHttpRequesterMock.java
```diff
@@ -60,17 +78,21 @@ public static void setRequestPayloads(String[] payloadReferences) {
     }
 
     @Override
-    protected HttpURLConnection setupConnection(URL builtUrl) throws IOException {
+    public Response<ResponseBody> makeRequest() throws IOException {
+        if(isAuthenticated && !new URL(url).getProtocol().contentEquals("https")){
```
amstone326: Is this necessary in the mock requester?

app/unit-tests/src/org/commcare/android/mocks/ModernHttpRequesterMock.java
```diff
@@ -60,17 +78,21 @@ public static void setRequestPayloads(String[] payloadReferences) {
...
         }
 
         if (requestPayloadStack.isEmpty()) {
-            return HttpURLConnectionMock.mockWithEmptyStream(builtUrl, responseCodeStack.remove(0));
+            return OkHTTPResponseMock.createResponse(responseCodeStack.remove(0));
         } else {
             String payloadReference = requestPayloadStack.remove(0);
             if (payloadReference == null) {
-                return HttpURLConnectionMock.mockWithErroringStream(builtUrl, responseCodeStack.remove(0));
+                throw new IOException(ioErrorMessage);
```
amstone326: I think this line is the wrong replacement for the line above it?

app/unit-tests/src/org/commcare/android/mocks/ModernHttpRequesterMock.java
```diff
@@ -46,6 +60,10 @@ public static void setResponseCodes(Integer[] responseCodes) {
         Collections.addAll(responseCodeStack, responseCodes);
     }
 
+    public static void setAuthenticated(boolean authenticated) {
```
amstone326: This method is unused

app/unit-tests/src/org/commcare/android/mocks/ModernHttpRequesterMock.java
```diff
@@ -79,11 +101,19 @@ protected HttpURLConnection setupConnection(URL builtUrl) throws IOException {
                 } catch (InvalidReferenceException ire) {
                     throw new IOException("No payload available at " + payloadReference);
                 }
-                return HttpURLConnectionMock.mockWithStream(builtUrl, responseCodeStack.remove(0), payloadStream);
+                return OkHTTPResponseMock.createResponse(responseCodeStack.remove(0),payloadStream);
```
amstone326: Space between params should be added back

app/unit-tests/src/org/commcare/android/mocks/ModernHttpRequesterMock.java
```diff
@@ -79,11 +101,19 @@ protected HttpURLConnection setupConnection(URL builtUrl) throws IOException {
...
             }
         }
     }
 
+    private URL buildUrlWithParams() throws MalformedURLException {
```
amstone326: Since the only usage of this method immediately calls `toString()` on the returned result, might as well make the return type `String` instead

app/unit-tests/src/org/commcare/android/tests/DataPullTaskTest.java
```diff
@@ -126,13 +126,14 @@ public void dataPullSelfIndexingCaseTest() {
...
         installAndUseLocalKeys();
+        int initialCount = LocalReferencePullResponseFactory.getNumRequestsMade();
         runDataPullWithAsyncRestore();
 
         Assert.assertEquals(DataPullTask.PullTaskResult.DOWNLOAD_SUCCESS, dataPullResult.data);
         
         // Indicates that the task executed all of the retries we indicated, and then successfully
         // parsed the final success response
-        Assert.assertEquals(4, LocalReferencePullResponseFactory.getNumRequestsMade());
+        Assert.assertEquals(4, LocalReferencePullResponseFactory.getNumRequestsMade() - initialCount);
```
amstone326: @shubham1g5 Ahhh, is this why this test fails sometimes??

app/src/org/commcare/network/HttpUtils.java
```diff
@@ -3,15 +3,42 @@
...
+
 /**
  * @author Phillip Mates (pmates@dimagi.com)
  */
 public class HttpUtils {
-    public static Pair<User, String> getUserAndDomain(boolean isAuthenticatedRequest) {
+
+    public static String getCredential(@Nullable Pair<String, String> usernameAndPasswordToAuthWith) {
+        String credential;
+        if (usernameAndPasswordToAuthWith == null) {
```
shubham1g5: `getCredential(String username, String password)` is now doing this check.

app/src/org/commcare/activities/PostRequestActivity.java
```diff
@@ -108,13 +112,8 @@ private void performSync() {
...
         if (!hasTaskLaunched && !inErrorState) {
-            ModernHttpTask postTask;
-            try {
-                postTask = new ModernHttpTask(this, url, params, true, null);
-            } catch (ModernHttpRequester.PlainTextPasswordException e) {
-                enterErrorState(Localization.get("post.not.using.https", url.toString()));
-                return;
-            }
+            RequestBody requestBody = ModernHttpRequester.getPostBody(params);
+            ModernHttpTask postTask = new ModernHttpTask(this, url.toString(), new HashMap(), requestBody, HTTPMethod.POST, null);
```
shubham1g5: I don't think that it should be the responsibility of `ModernHttpRequester` to create the Request body as in general different  POST requests would always have different request bodies. Though to make it a bit cleaner we can make an overloaded constructor for `ModernHttpRequester`

app/src/org/commcare/network/HttpUtils.java
amstone326: Ah gotcha, makes sense. It looks like that method hasn't made its way into this class though, btw

app/src/org/commcare/activities/PostRequestActivity.java
amstone326: Cool that seems like a good solution

app/unit-tests/src/org/commcare/android/mocks/ModernHttpRequesterMock.java
shubham1g5: This is added here to  to make the mockRequester behaves the same as the original one, since code that throws this exception moved to okHTTP interceptor which only gets called when we make an actual request.

app/unit-tests/src/org/commcare/android/mocks/ModernHttpRequesterMock.java
shubham1g5: In non retrofit code what `HttpURLConnectionMock` does is returns an `IOException` when we call `getInputStream`. Now the `makeRequest()` function we are in is responsible for executing the request and returning the response unlike earlier when `ModernHttpRequesterMock` was  returning the `HTTPConnection` which in turn was throwing this exception when `getInputStream()` used to call. So since the responsibility of `ModernHttpRequesterMock` has changed from just returning the connection to execute and returning the response, I think the above change is justified.

app/unit-tests/src/org/commcare/android/tests/DataPullTaskTest.java
shubham1g5: Yeah!

app/src/org/commcare/activities/PostRequestActivity.java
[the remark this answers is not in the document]
shubham1g5: Now I don't think it's a good solution as doing that will cause another overloaded constructor in `CommCareApplication` and yet another in `ModernHTTPTask` which doesn't seem to be worth for removing an argument. Another not so good approach would be to have a setter for `requestBody` in `ModernHttpRequester` instead of passing it as an argument to constructor. I am not sure making `requestBody`mutable is a good idea and hence of the opinion to keep it same as now.

shubham1g5:

app/src/org/commcare/activities/PostRequestActivity.java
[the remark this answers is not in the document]
amstone326: Hmm.. I would probably be slightly in favor of having a setter method for the request body, but understand your hesitation. Will leave it up to you.

app/unit-tests/src/org/commcare/android/tests/DataPullTaskTest.java
amstone326: amazing find, thank you

amstone326:
@shubham1g5 Cool, thanks for the clean-up. The 2 things remaining from my feedback now are:

1. My comment about drawing a clearer conceptual distinction between `HttpRequestEndpoints` and `CommCareNetworkService`

2. My general question about more tests

shubham1g5:
@amstone326 

I am not sure if removing the `HttpRequestEndpoints` interface is a better approach as currently different implementations of `DataPullRequester` ex. `LocalFilePullResponseFactory` are coded to this interface instead of concrete implementations of it. I am all in favour of changing the naming conventions around it though. 

I also just realised that now `SimpleGetTask` doesn't make any sense as it's essentially doing the same thing as `ModernHTTPTask`. If we decide to retain the `HTTPRequestGenerator`, I think it would be a good idea to delete SimpleGetTask and have a simplePost method similar to the simpleGet in `HTTPRequestGenerator` and have the ModernHttpTask utilise  `HTTPRequestGenerator` methods (SimpleGet, SimplePost, postMultipart) internally depending on the request method. This way we can centralise all different request endpoints into `HTTPRequestGenerator`.

Regarding tests, we seems to already have tests for these network paths -

ModernHTTPTask -> PostRequestActivity -> PostRequestActivityTest
ModernHTTPTask -> QueryRequestActivity -> QueryRequestActivityTest
makeCaseFetchRequest -> DataPullRequester -> DataPullTaskTest
makeKeyFetchRequest -> ManageKeyRecordTask -> KeyRecordTest
HearBeatRequester -> HeartbeatAndPromptedUpdateTests

While for these paths I was not able to find any Robo tests - 

postMultipart -> ProcessAndSendTask,
postMultipart -> SendTask,
postMultipart -> ForceCloseLogger,
postMultipart -> LogSubmissionTask

SimpleGet -> MicroNode::getAvailableApplications,
SimpleGet -> JavaHTTPReference::getStream,
SimpleGet -> InstallFromListActivity.requestAppList

I am not sure yet how to proceed on writing tests for these as I am feeling a bit without context regarding how exactly we are interpreting the request/response body for these paths specially for SimpleGet paths. Though I am looking into these and trying to debug through paths through which we are making these calls.

amstone326:
@shubham1g5 

1. Totally agree that `SimpleGetTask` doesn't make sense anymore, good catch. I created that when we still had 2 different implementations of HTTP libs and I specifically didn't want to use the version that `ModernHttpTask` was using. But since it's all one and the same now, you should definitely go ahead and remove it.

2. I'm fine with just renaming those classes along the lines of how I suggested, instead of deleting `HttpRequestEndpoints`, if you think that's better. 

3. As for tests:

- For `ProcessAndSendTask` and `SendTask`, the use of `postMultipart()` is for sending forms. This is already tested in the calabash tests and I think that is the best place for those tests, so this can be considered covered.
- For `ForceCloseLogger` and `LogSubmissionTask`, the use of `postMultipart()` is for sending logs. I actually do not think we have a calabash test for this behavior, so this would be a great opportunity to add one if you're up for that. I imagine the test would be to trigger some behavior that we know creates some device logs, and then trigger the log-send process (there is a button in advanced settings to force log submission), and confirm that we get a success response. Even better would be to use a script to confirm that the log submission is actually present on HQ, but I'm not sure how tricky that would be.
- For `InstallFromListActivity.requestAppList()`, there is already a calabash test for this, so that is sufficient
- `JavaHTTPReference::getStream` is used for installing app resources from a remote authority, so this is tested by virtue of OTA app installs. This is tested in the calabash tests in `basic_test_1.feature`, so that is sufficient
- For `MicroNode::getAvailableApplications`, I'm going to defer to Clayton. @ctsims, is there any existing test coverage for the MicroNode functionality? If not, what would be the best place to set one up?

Having said that a lot of our test coverage for this can come from Calabash, I also wanted to double check that you've run the full calabash test suite agains this PR? If not, that should definitely get done before this is merged.

shubham1g5:
@amstone326 Thanks for breaking it up for me. Super useful. I have managed to run the complete calabash test once, though once we make these changes I will run them again to test for any regressions. 

Action items for now - 

* Get rid of `SimpleGetTask`
* Improve naming around `HttpRequestEndpoints`
* Calabash test for `ForceCloseLogger` and `LogSubmissionTask`

amstone326:
@shubham1g5 Great, that sounds like the right list. I would add 1 more thing to that potentially, which is to add a test for the MicroNode behavior, depending on Clayton's response. (We can talk more about what that feature actually is if a test is in fact needed).

shubham1g5:

ctsims:
Hey, no current tests for micronode. I could come up with some specific mock files to parse with the manifest formats, but it's inherently an IPC channel, so it's super hard to test as a regression. 

I can manually test it for now since it's an R&D thing. No need to block the PR on it, I can fix before it makes it out if needed.

app/src/org/commcare/activities/InstallFromListActivity.java
```diff
@@ -363,9 +361,13 @@ public void processOther(int responseCode) {
     }
 
     @Override
-    public void handleIOException(IOException exception) {
-        Logger.log(LogTypes.TYPE_ERROR_SERVER_COMMS,
-                "An IOException was encountered during get available apps request: " + exception.getMessage());
+    public void handleException(Exception exception) {
```
ctsims: This explains the fuzzing of the exception type earlier. I think it would be better to type the plaintext password exception as an IOEXception that we specifically catch and log rather than moving the processing to take generic exceptions. (We already do similar things with HTTPS bad certificate responses).

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -251,13 +248,6 @@ private static boolean shouldGenerateFirstKey() {
             e.printStackTrace();
             Logger.log(LogTypes.TYPE_WARNING_NETWORK, "Timed out listening to receive data during sync");
             responseError = PullTaskResult.CONNECTION_TIMEOUT;
-        } catch (ConnectTimeoutException e) {
```
ctsims: Is this no longer a valid exception for some reason? There's definitely still a difference between SocketTimeouts and Connection timeouts in retrofit, right? Do we need to add the new exception type for retrofit's verison?

app/src/org/commcare/utils/FormUploadUtil.java
```diff
@@ -316,9 +309,9 @@ private static boolean buildMultipartEntity(MultipartEntity entity,
         return true;
     }
 
-    private static int addPartToEntity(MultipartEntity entity, File f, ContentBody fb) {
-        if (fb.getContentLength() <= MAX_BYTES) {
-            entity.addPart(f.getName(), fb);
+    private static int addPartToEntity(List<MultipartBody.Part> parts, File f, String contentType) {
+        if (f.length() <= MAX_BYTES) {
```
ctsims: Informational (IE: If you see this and it makes you think of something, good, but I think it's fine it not):

Slightly pedantic, but this was (I believe) intentionally targeting the content length previously, think? I think the equivalent in this case would likely be 
```
part = createFilePart(name,file,type)
if(part.contentLength < MAX_BYTES)
  parts.add(part)
...
```
but I don't have a specific scenario in my head where that's likely to affect things unless the body itself ends up being bigger/smaller by a significant amount, which could occur in very small part due to metadata (but not enough for us to care).

app/assets/locales/android_translatable_strings.txt
```diff
@@ -164,7 +164,6 @@ post.io.error=Error reading server response: ${0}
 post.unknown.response=Received unknown resonse code (${0}) from server
 post.client.error=Client-side error (code ${0}) received from network request.
 post.server.error=Server-side error (code ${0}) received from network request.
-post.redirection.error=Network request resulted in redirection (code ${0}). Please use a non-redirecting url.
 post.gone.error=Case is not present on server.
```
ctsims: Shouldn't there be a new string here for the new error case? (Attempt to auth over HTTP)?

app/src/org/commcare/preferences/CommCareServerPreferences.java
shubham1g5: It's actually getting utilized now. I crated the function earlier while forgetting to use this in actual implementation.

app/assets/locales/android_translatable_strings.txt
shubham1g5: Currently the error is only getting logged into logs and not shown to user and hence we don't have a corresponding error here. Though just realized there are few other places where PlainTextPasswordException should get caught and logged. Will add it them once we decide on whether to retain the exception or not.

app/src/org/commcare/utils/FormUploadUtil.java
shubham1g5: So I checked the implementation for `FileBody:: getContentLength()` and it just returns `file.length()`. So I think we should be good here.

app/assets/locales/android_translatable_strings.txt
ctsims: I think it's possible that the other requester was just never triggering the error. I think it's important that the user be able to see the specific error if it occurs (since otherwise it will be impossible for an end-user to understand why their network requests are failing), but since the error would actually only ever occur when someone technical is setting up a new environment it actually probably doesn't need to be in the translation set (IE: It's fine if the error isn't localizable and the message is in code, as long as it's clear that the error is "the server you are connecting to is insecure (HTTP), so commcare can't connect with authentication."

app/src/org/commcare/utils/FormUploadUtil.java
ctsims: Great! Checked off.

app/src/org/commcare/network/RemoteDataPullResponse.java
```diff
@@ -111,31 +112,14 @@ public void notifyCurrentCount(long bytesRead) {
     }
 
     protected InputStream getInputStream() throws IOException {
-        return AndroidHttpClient.getUngzippedContent(response.getEntity());
-    }
-
-    public String getShortBody() throws IOException {
-        return new String(StreamsUtil.inputStreamToByteArray(AndroidHttpClient.getUngzippedContent(response.getEntity())));
+        return response.body().byteStream();
```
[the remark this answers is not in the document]
shubham1g5: So I had confirmed that the request does contain the header for gzip encoding which basically means that there should not be any change in response. I am not sure though how can I test it with the returned HTTP response. Should I check for a size decrease or something like that!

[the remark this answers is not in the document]
ctsims: Checking the response for the `Content-Encoding: gzip` header should be enough, i'd think, but we shouldn't block on it if you already validated that we're sending up the gzip request, since there's no reason to believe the server would respond differently. (We can double check next week to ensure we're getting a gzip'd response content encoding)

app/src/org/commcare/tasks/DataPullTask.java
shubham1g5: I removed it earlier since it was deprecated. and OkHTTP sends the same SocketTimeOutException in both cases currently. Though on looking at this https://github.com/square/okhttp/issues/3318 it seems like something like `ConnectTimeoutException` might get added in newer releases.

app/src/org/commcare/network/RemoteDataPullResponse.java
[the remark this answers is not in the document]
shubham1g5: Verified `Content-Encoding: gzip` in response as well.

app/src/org/commcare/network/HttpCalloutTask.java
```diff
@@ -78,7 +81,7 @@ protected HttpCalloutOutcomes doTaskBackground(Object... params) {
                 } else {
                     outcome = doResponseOther(response);
                 }
-            } catch (ClientProtocolException | UnknownHostException e) {
+            } catch (UnknownHostException e) {
                 outcome = HttpCalloutOutcomes.NetworkFailure;
             } catch (SSLPeerUnverifiedException e) {
```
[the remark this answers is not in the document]
shubham1g5: OkHTP uses the same`java.net.ssl.SSLPeerUnverifiedException`

shubham1g5:

shubham1g5:
@ctsims I think all comments are addressed now along with Secure Endpoint Validation. Please have a look at it again. Also ran complete test suite again successfully as this seems near final now to me.

ctsims:
@shubham1g5 Congrats on finishing this up! Is a huge milestone for us and I'm excited to see the impact.

PR #1770, opened 2017-07-25
Logging and better handling for negative period value

amstone326:
@ctsims After looking at this a bit more, it seems to me that there are 2 possible causes for the "Non-positive period" issue:

1. HQ is actually returning a negative value for the retry period. This PR makes it so that if this is in fact what is happening, a special `InvalidWaitTimeException` will be thrown, whose message will communicate what the returned retry-after value was 

2. The interval of time that passes between when the retry-after value is parsed, and when `startReportingServerProgress()` is called, is somehow longer than the duration of the retry-after value itself. It occurred to me that this is vaguely possible now that HQ changed the retry-after period from 30 seconds to 5 seconds. It still seems unlikely to me that it could take CommCare 5 seconds to go through that process, but it's possible. I've added code in this PR which will handle this case by logging the occurrence and instead of starting the timer, just doing 1 isolated call to `reportServerProgress()`

### DIFF
--- a/app/src/org/commcare/tasks/AsyncRestoreHelper.java
+++ b/app/src/org/commcare/tasks/AsyncRestoreHelper.java
@@ -38,6 +38,11 @@ public class AsyncRestoreHelper {
         }
         try {
             long waitTimeInMilliseconds = Integer.parseInt(retryHeader.getValue()) * 1000;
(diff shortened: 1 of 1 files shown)


ctsims:
agree with the new approach, thanks for taking a look at this.

PR #1771, opened 2017-07-25
Make recovery sync handle retry response

amstone326:


app/unit-tests/src/org/commcare/android/tests/DataPullTaskTest.java
```diff
@@ -80,6 +80,13 @@ public void dataPullRecoverFailTest() {
     }
 
     @Test
+    public void dataPullRecoverWithRetryTest() {
+        installLoginAndUseLocalKeys();
+        runDataPull(new Integer[]{412, 202, 200}, new String[]{GOOD_RESTORE, RETRY_RESPONSE, GOOD_RESTORE});
```
ctsims: Oh man this rules

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -465,26 +467,39 @@ private static void recordSuccessfulSyncTime(String username) {
 
     //TODO: This and the normal sync share a ton of code. It's hard to really... figure out the right way to 
     private Pair<Integer, String> recover(HttpRequestEndpoints requestor, AndroidTransactionParserFactory factory) {
-        this.publishProgress(PROGRESS_RECOVERY_NEEDED);
-
-        Logger.log(AndroidLogger.TYPE_USER, "Sync Recovery Triggered");
+        while (asyncRestoreHelper.retryWaitPeriodInProgress()) {
```
ctsims: There should be a sleep() after this call so this thread doesn't stay redlined while we wait.

app/unit-tests/src/org/commcare/android/tests/DataPullTaskTest.java
amstone326: Ha yea I was so pleased with how easy it was to write this test

app/src/org/commcare/tasks/DataPullTask.java
amstone326: For how long?

ctsims: It will be full seconds (many of them) for the retry period, so probably 500ms or so?

-Clayton

amstone326: Updated, and I also added a sleep to the parallel while loop in `getRequestResultOrRetry()`

PR #1772, opened 2017-07-25
Hotfix for 412's not cleaning storage properly

ctsims:
Fixes issue where 412 syncs don't clear case index table, and rebuild existing index table to ensure any old stale state is corrected

### DIFF
--- a/app/src/org/commcare/models/database/user/models/AndroidCaseIndexTable.java
+++ b/app/src/org/commcare/models/database/user/models/AndroidCaseIndexTable.java
@@ -126,6 +126,18 @@ public class AndroidCaseIndexTable implements CaseIndexTable {
         }
     }
 
+    /**
+     * Removes all records from the case index table
+     */
+    public void wipeTable() {
+        db.beginTransaction();
+        try {
+            db.delete(TABLE_NAME, null, null);
+            db.setTransactionSuccessful();
+        } finally {
+            db.endTransaction();
+        }
+    }
 
     /**
      * Get a list of Case Record id's for cases which index a provided value.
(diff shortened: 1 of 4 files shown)


app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -504,7 +506,7 @@ private static void recordSuccessfulSyncTime(String username) {
 
         //Wipe storage
         //TODO: move table instead. Should be straightforward with sandboxed db's
```
amstone326: Is this comment still relevant?

app/src/org/commcare/models/database/user/models/AndroidCaseIndexTable.java
```diff
@@ -279,4 +291,17 @@ public static String getArgumentBasedVariableSet(int number) {
         return sb.toString();
     }
 
+    public void reIndexAllCases(SqlStorage<ACase> caseStorage) {
+        db.beginTransaction();
+        try {
+            wipeTable();
+            for(ACase c : caseStorage) {
+                indexCase(c);
```
amstone326: If `db.insert()` is called on a table name that doesn't yet exist, does it create the table? Otherwise this seems like it would fail (since `wipeTable()` deletes the table entirely, rather than just wiping its contents)

ctsims: Yeah, the table is guaranteed to exist. If it's not created yet it gets created in the 5->6 migration. I'll double test against a fresh install as well, for confidence.

ctsims: Oh, sorry,t o be clear: wipeTable() does delete rows, not the full table. db.delete(TABLE_NAME, null, null); is "delete all rows", not "Drop table"

app/src/org/commcare/tasks/DataPullTask.java
ctsims: It is, the better strategy here would basically be to stage an entirely new sandbox and restore into that, it's just much more complex.

app/src/org/commcare/models/database/user/models/AndroidCaseIndexTable.java
amstone326: ah gotcha, thanks

PR #1774, opened 2017-07-27
Don't show save option in read only mode

amstone326:
Fixes an issue that made it possible for already-sent/saved forms to get into an "Unsent" state and thus be re-submittable: http://manage.dimagi.com/default.asp?232353

Also, Clayton and I both tested and confirmed that if you resubmit a form that is set to "Unsent" in this way, it's form submission ordering # is set to -1, which explains the troubling logs we were seeing


app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -128,6 +127,7 @@
     public static AndroidFormController mFormController;
 
     private boolean mIncompleteEnabled = true;
+    private boolean instanceIsReadOnly = false;
```
ctsims: @amstone326 This needs to be added to the list of serialized case properties save and recovered when the activity cycle restarts or the state won't persist.

amstone326:
@ctsims Good catch, thanks. updated

PR #1777, opened 2017-08-01
Include app version and cc version as params in heartbeat request

amstone326:


app/src/org/commcare/android/logging/ReportingUtils.java
```diff
@@ -91,4 +94,14 @@ public static String getVersion() {
         }
     }
 
+    public static String getCommCareVersionString() {
```
wpride: is this doing anything different from https://github.com/dimagi/commcare-android/blob/bbabdbc72fb8ecd04d8582f432f6634d3c7e1b6a/app/src/org/commcare/AppUtils.java#L140-L140 ?

amstone326: @wpride Yea it is, if you look at the implementation of that method, it returns a lot more information than just the version itself.

PR #1778, opened 2017-08-01
Don't attempt remote user key management AFTER data pull success

amstone326:
I think this will fix the current failure of the demo_user_restore_offline test on AWS.

The demo user work was already configured to not attempt remote user key management on the _first_ run of `ManageKeyRecordTask`, but apparently not on the one that happens _after_ a successful data pull. This wasn't causing any issues in live usage of the feature, because on that 2nd run of the task, `calloutSuccessRequired` is set to false since there is already a record present at that point (https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/tasks/ManageKeyRecordTask.java#L249). I'm not sure what about AWS exactly is making it fail here where nothing else did, but I think Clayton seems right that it has to do with it "failing" by waiting for a long time and then timing out.


ctsims:

ctsims:

ctsims:

ctsims:
Oof, things definitely seem much more consistent now, although it's pretty clear that it's a bit of a rough setup.

Pulling. I wonder if we should make a tech debt backlog for mobile. I think the amount of different flags and state in restore/sync/etc would make a good candidate for an entry in that...

PR #1780, opened 2017-08-03
Dev Feature: Load forms from a local file

shubham1g5:
- Adds a feature to load forms from a local xml file containing Form Payload. Similar to https://github.com/dimagi/commcare-android/pull/1102/files which is loading forms from a server url.  Sample Payload file - https://gist.github.com/shubham1g5/f22cfd99decc68067190d37a63e0de6b

- No longer removing submission url on saving form payload status from Dev Preferences to allow for sync after we change the payload status

- Adds a custom preference FilePreference to have the ability to trigger file manager along with the file path editText

### DIFF
--- a/app/res/values/attrs.xml
+++ b/app/res/values/attrs.xml
@@ -51,6 +51,7 @@
 
     <declare-styleable name="FilePreference">
         <attr name="file_type" format="string"/>
+        <attr name="error_dialog_title" format="string"/>
     </declare-styleable>
 
 </resources>
--- a/app/res/values/attrs.xml
+++ b/app/res/values/attrs.xml
@@ -49,4 +49,8 @@
         <attr name="clip_height" format="reference|float"/>
     </declare-styleable>
 
+    <declare-styleable name="FilePreference">
+        <attr name="file_type" format="string"/>
+    </declare-styleable>
+
 </resources>
--- a/app/res/values/attrs.xml
+++ b/app/res/values/attrs.xml
@@ -51,7 +51,6 @@
 
     <declare-styleable name="FilePreference">
         <attr name="file_type" format="string"/>
-        <attr name="error_dialog_title" format="string"/>
     </declare-styleable>
 
 </resources>
--- a/app/res/xml/preferences_developer.xml
+++ b/app/res/xml/preferences_developer.xml
@@ -88,8 +88,8 @@
     <EditTextPreference
         android:defaultValue=""
         android:enabled="true"
-        android:key="cc-form-record-file-path"
-        android:title="Form Record File"/>
+        android:key="cc-local-form-payload-file-path"
+        android:title="Local Form Payload File"/>
     <ListPreference
         android:defaultValue="saved"
         android:enabled="true"
--- a/app/res/xml/preferences_developer.xml
+++ b/app/res/xml/preferences_developer.xml
@@ -90,6 +90,11 @@
         android:enabled="true"
         android:key="cc-local-form-payload-file-path"
         android:title="Local Form Payload File"/>
+    <EditTextPreference
+        android:defaultValue="@string/remote_form_payload_url"
+        android:enabled="true"
+        android:key="remote-form-payload-url"
+        android:title="Form Payload Server"/>
     <ListPreference
         android:defaultValue="saved"
         android:enabled="true"
--- a/app/res/xml/preferences_developer.xml
+++ b/app/res/xml/preferences_developer.xml
@@ -98,6 +98,7 @@
         android:enabled="true"
         android:key="cc-local-form-payload-file-path"
         android:title="Local Form Payload File"
+        app:error_dialog_title="cannot.set.payload"
         app:file_type="xml"/>
     <EditTextPreference
         android:defaultValue="@string/remote_form_payload_url"
@@ -132,9 +133,13 @@
         android:entryValues="@array/pref_enabled_vals"
         android:key="cc-use-pw-obfuscation"
         android:title="Use Password Obfuscation"/>
-    <Preference
+    <org.commcare.preferences.FilePreference
+        android:defaultValue=""
+        android:enabled="true"
         android:key="cc-custom-restore-doc-location"
-        android:title="Restore a custom user XML data file"/>
+        android:title="Restore a custom user XML data file"
+        app:error_dialog_title="cannot.restore.xml"
+        app:file_type="xml"/>
     <ListPreference
         android:defaultValue="no"
         android:enabled="true"
--- a/app/res/xml/preferences_developer.xml
+++ b/app/res/xml/preferences_developer.xml
@@ -98,7 +98,6 @@
         android:enabled="true"
         android:key="cc-local-form-payload-file-path"
         android:title="Local Form Payload File"
-        app:error_dialog_title="cannot.set.payload"
         app:file_type="xml"/>
     <EditTextPreference
         android:defaultValue="@string/remote_form_payload_url"
@@ -138,7 +137,6 @@
         android:enabled="true"
         android:key="cc-custom-restore-doc-location"
         android:title="Restore a custom user XML data file"
-        app:error_dialog_title="cannot.restore.xml"
         app:file_type="xml"/>
     <ListPreference
         android:defaultValue="no"
--- a/app/res/xml/preferences_developer.xml
+++ b/app/res/xml/preferences_developer.xml
@@ -6,8 +6,8 @@
         android:key="cc-superuser-enabled"
         android:title="Developer Mode Enabled"/>
     <ListPreference
-        android:enabled="true"
         android:defaultValue="no"
+        android:enabled="true"
         android:entries="@array/pref_enabled_labels"
         android:entryValues="@array/pref_enabled_vals"
         android:key="cc-show-update-target-options"
@@ -85,7 +85,14 @@
         android:entryValues="@array/pref_enabled_vals"
         android:key="cc-auto-purge"
         android:title="Detect and Trigger Purge on Form Save"/>
-    <EditTextPreference
+    <ListPreference
+        android:defaultValue="saved"
+        android:enabled="true"
+        android:entries="@array/pref_load_form_payload_as_labels"
+        android:entryValues="@array/pref_load_form_payload_as_vals"
+        android:key="cc-form-payload-status"
+        android:title="Load form payload status"/>
+    <org.commcare.preferences.FilePreference
         android:defaultValue=""
         android:enabled="true"
         android:key="cc-local-form-payload-file-path"
@@ -95,13 +102,6 @@
         android:enabled="true"
         android:key="remote-form-payload-url"
         android:title="Form Payload Server"/>
-    <ListPreference
-        android:defaultValue="saved"
-        android:enabled="true"
-        android:entries="@array/pref_load_form_payload_as_labels"
-        android:entryValues="@array/pref_load_form_payload_as_vals"
-        android:key="cc-form-payload-status"
-        android:title="Load form payload status"/>
     <ListPreference
         android:defaultValue="yes"
         android:enabled="true"
@@ -124,18 +124,18 @@
         android:key="cc-show-entity-trace-outputs"
         android:title="Output Entity List Trace Data to ADB"/>
     <ListPreference
-        android:enabled="true"
         android:defaultValue="no"
+        android:enabled="true"
         android:entries="@array/pref_enabled_labels"
         android:entryValues="@array/pref_enabled_vals"
         android:key="cc-use-pw-obfuscation"
         android:title="Use Password Obfuscation"/>
     <Preference
-        android:title="Restore a custom user XML data file"
-        android:key="cc-custom-restore-doc-location"/>
+        android:key="cc-custom-restore-doc-location"
+        android:title="Restore a custom user XML data file"/>
     <ListPreference
-        android:enabled="true"
         android:defaultValue="no"
+        android:enabled="true"
         android:entries="@array/pref_enabled_labels"
         android:entryValues="@array/pref_enabled_vals"
         android:key="cc-enable-bulk-performance"
--- a/app/res/xml/preferences_developer.xml
+++ b/app/res/xml/preferences_developer.xml
@@ -1,4 +1,5 @@
-<PreferenceScreen xmlns:android="http://schemas.android.com/apk/res/android">
+<PreferenceScreen xmlns:android="http://schemas.android.com/apk/res/android"
+                  xmlns:app="http://schemas.android.com/apk/res-auto">
     <ListPreference
         android:enabled="true"
         android:entries="@array/pref_enabled_labels"
@@ -96,7 +97,8 @@
         android:defaultValue=""
         android:enabled="true"
         android:key="cc-local-form-payload-file-path"
-        android:title="Local Form Payload File"/>
+        android:title="Local Form Payload File"
+        app:file_type="xml"/>
     <EditTextPreference
         android:defaultValue="@string/remote_form_payload_url"
         android:enabled="true"
--- a/app/res/xml/preferences_developer.xml
+++ b/app/res/xml/preferences_developer.xml
@@ -85,6 +85,11 @@
         android:entryValues="@array/pref_enabled_vals"
         android:key="cc-auto-purge"
         android:title="Detect and Trigger Purge on Form Save"/>
+    <EditTextPreference
+        android:defaultValue=""
+        android:enabled="true"
+        android:key="cc-form-record-file-path"
+        android:title="Form Record File"/>
     <ListPreference
         android:defaultValue="saved"
         android:enabled="true"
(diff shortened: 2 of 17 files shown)


app/res/xml/preferences_developer.xml
```diff
@@ -85,6 +85,11 @@
         android:entryValues="@array/pref_enabled_vals"
         android:key="cc-auto-purge"
         android:title="Detect and Trigger Purge on Form Save"/>
+    <EditTextPreference
+        android:defaultValue=""
+        android:enabled="true"
+        android:key="cc-form-record-file-path"
+        android:title="Form Record File"/>
```
amstone326: Same comment about renaming (something like `"Local Form Payload File"`).

amstone326: Ah, I just realized you were copying this naming pattern from what Philip was using for the remote version, which makes sense. I think that one is confusing too, and would be in favor of renaming the preference title for that similarly (something like `"Form Payload Server"`)

app/res/xml/preferences_developer.xml
```diff
@@ -85,6 +85,11 @@
         android:entryValues="@array/pref_enabled_vals"
         android:key="cc-auto-purge"
         android:title="Detect and Trigger Purge on Form Save"/>
+    <EditTextPreference
```
amstone326: Instead of being an `EditTextPreference`, this should be a preference that when clicked, opens a file browser for the user to select a local file which contains a payload. Then when the user selects a file, CommCare saves the file path to the preference to be retrieved later. It can also do some validation to check whether you've selected a valid file type (a .xml file), and can reject the selection if it isn't. [The preference for `print-doc-location`](https://github.com/dimagi/commcare-android/blob/c3602a949e2fc753e53fa46a9390576fa3259154/app/src/org/commcare/preferences/CommCarePreferences.java#L83-L83) follows this exact model, so you can use that as an example.

app/src/org/commcare/logic/ArchivedFormRemoteRestore.java
```diff
@@ -25,11 +29,30 @@
...
+    public static void pullArchivedFormsFromFile(String filePath,
+                                                 final FormRecordListActivity activity,
+                                                 final CommCarePlatform platform) {
+
+        File file = new File(filePath);
+        if (file != null && file.exists()) {
+            LocalFilePullResponseFactory.setRequestPayloads(new File[]{file});
+            requestForms(activity, platform, "fake-server-that-is-never-used", LocalFilePullResponseFactory.INSTANCE, true);
+        } else {
+            Toast.makeText(activity, "Restore file doesn't exist", Toast.LENGTH_LONG).show();
```
amstone326: Small, but message should say `"Payload file doesn't exist"` instead of `"Restore file"`

app/res/xml/preferences_developer.xml
shubham1g5: I can do that, but my worry is how will we configure the UI tests to choose the file from a file manager considering different file managers every phone has!

app/src/org/commcare/preferences/CommCareServerPreferences.java
```diff
@@ -96,6 +96,10 @@ public static String getSupportEmailAddress() {
         return getServerProperty(PREFS_SUPPORT_ADDRESS_KEY, CommCareApplication.instance().getString(R.string.support_email_address_default)) ;
     }
 
+    public static String getFormRecordKey() {
```
shubham1g5: I will change the nomenclature around it. Though today our prefs get methods are arranged by the screen they belong to. So since remote form payload url pref is available on ServerPreferences I think the getter should be defined here as well.

shubham1g5: Though I think we can move the remote payload url pref to Developer Prefs as well as it's meant to be a dev feature!

app/res/xml/preferences_developer.xml
amstone326: That is a fair concern, but I don't think we want to get in the habit of implementing features sub-optimally just to make it easier to test them. I would much rather just add this to the manual test plan if it's really not feasible to test in calabash.

app/src/org/commcare/preferences/CommCareServerPreferences.java
amstone326: Ah yea that's a fair point, but I think you're right that these should both just live and be made visible through Developer Preferences only. While the remote payload URL is technically a server pref in that it represents a URL for server communications, I think it's more important that it be hidden from regular users.

app/res/xml/preferences_developer.xml
shubham1g5: I agree but the main use case of this pref is to enable UI tests benchmark different app functionalities on a a bulk load of forms (Upload Sync for example). So dropping calabash support for this is not really an option.

app/src/org/commcare/preferences/CommCareServerPreferences.java
shubham1g5: Yeah. I agree as well. I am going to proceed with moving them both to Developer Prefs.

app/res/xml/preferences_developer.xml
amstone326: Ha, right, I forgot that key point. I still think it's important to make this more usable in this way though, and actually don't think it will be a problem in calabash from thinking a bit more about it. If you look at how the step for `"I Install the .ccz file at {some_file_path}"` is implemented in our calabash steps, that also relies on the behavior of selecting a file from the file system, and that has been quite reliable thus far.

shubham1g5: Cool. If that's the case I think we should be good. Will change it to a file manager implementation.

shubham1g5: So I took a look at the .ccz installation code. Actually for .ccz installation we have both options , to write a file path manually and a  folder icon to choose it from file manager. Calabash tests never actually utilise the file manager option and directly put the path in EditText. I am going to try to do the same here and build a custom preference FilePreference for such prefs.

shubham1g5:
@amstone326 made the required changes. Please take a look again.

amstone326:
Hey @shubham1g5, I'm a little confused on a high-level as to why this whole refactor was needed, instead of just implementing the file browser opening for a preference in the same way that was already being done for the print template? 

Or conversely, if there was a clear reason for changing the strategy, why didn't you change it for the print template and custom restore file prefs as well?

shubham1g5:
@amstone326 I wanted to avoid using a file manager in Calabash tests and therefore wanted to provide an EditText for manually entering the file path as well a button which opens the file manager to let user choose the file. It's exactly like what we do for offline ccz install (Button + EditText). 

We can definitely change the behaviour of print template and custom restore file prefs as well but since they don't have any Calabash tests for them right now, I didn't touch them.

app/src/org/commcare/preferences/DeveloperPreferences.java
```diff
@@ -236,15 +227,6 @@ public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, Strin
                     DevSessionRestorer.clearSession();
                 }
                 setSessionEditText();
-
-                break;
-            case LOAD_FORM_PAYLOAD_AS:
```
amstone326: What's the reason for deleting this block of code?

app/src/org/commcare/preferences/FilePreferenceDialogFragmentCompat.java
```diff
@@ -0,0 +1,61 @@
...
+            CommCarePreferences.startFileBrowser(FilePreferenceDialogFragmentCompat.this, REQUEST_FILE, "cannot.set.payload");
+        });
+    }
+
+    @Override
+    public void onActivityResult(int requestCode, int resultCode, Intent intent) {
+        if (requestCode == REQUEST_FILE) {
+            if (resultCode == RESULT_OK && intent != null) {
+                Uri uri = intent.getData();
+                String filePath = UriToFilePath.getPathFromUri(CommCareApplication.instance(), uri);
```
amstone326: Can you restrict this to being a `.xml` file, since that's the only file type that will work?

app/src/org/commcare/preferences/DeveloperPreferences.java
```diff
@@ -167,26 +172,12 @@ public boolean onPreferenceClick(Preference preference) {
                 GoogleAnalyticsUtils.reportPrefItemClick(
                         GoogleAnalyticsFields.CATEGORY_DEV_PREFS,
                         GoogleAnalyticsFields.LABEL_CUSTOM_RESTORE);
-                startFileBrowser();
+                startFileBrowser(DeveloperPreferences.this, REQUEST_SYNC_FILE, "cannot.restore.xml");
```
amstone326: Nevermind ignore this comment; I didn't realize that this triggers the settings activities to finish and do the restore immediately

app/src/org/commcare/preferences/DeveloperPreferences.java
shubham1g5: Earlier we were blocking the sync on changing the load form saved status to Unsent/Incomplete by clearing out the submission url so that user doesn't end up submitting lots of form to HQ. Now since we are using this to test the form upload, we want to be able to sync forms after we load forms as unsent/incomplete.

app/res/xml/preferences_developer.xml
```diff
@@ -6,8 +6,8 @@
         android:key="cc-superuser-enabled"
         android:title="Developer Mode Enabled"/>
     <ListPreference
```
shubham1g5: My current AS styling settings arranges the xml attributes alphabetically. Do we follow that practice or should I turn it off.

amstone326: Ah no I don't think we do, can turn it off.

app/src/org/commcare/preferences/DeveloperPreferences.java
amstone326: Ah right right

amstone326:
@shubham1g5 Last thing-- can you actually refactor the "restore from custom file" developer preference to use the new dialog setup you have here? I literally just came upon a scenario where I'd like to use it in a calabash test as well haha, so I think it makes sense. (The printer one it makes sense to leave as-is, because that setting is user-facing, so better to keep the UX as intuitive as possible).

app/res/xml/preferences_developer.xml
shubham1g5: Hey just noticed all other pref files seems to follow the alphabetical order. Would you be able to recheck it in your editor. I do think now that we are ordering alphabetically.

amstone326: Hm, I'm not seeing that in most of the prefs .xml files, but also honestly isn't a big deal so fine to just leave!

app/res/values/attrs.xml
```diff
@@ -49,4 +49,9 @@
         <attr name="clip_height" format="reference|float"/>
     </declare-styleable>
 
+    <declare-styleable name="FilePreference">
+        <attr name="file_type" format="string"/>
+        <attr name="error_dialog_title" format="string"/>
```
amstone326: Hey, so while this seems like a clean solution for customizing the dialog title, I'm still not sure why it's necessary, and this seems like a lot of setup for something that we could just not do. Was there a specific reason you saw that made it seem more important to customize the title, as opposed to just going with something like "No File Browser Available"?

app/src/org/commcare/activities/HomeScreenBaseActivity.java
```diff
@@ -374,16 +374,23 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
                         finish();
                     } else if (resultCode == DeveloperPreferences.RESULT_SYNC_CUSTOM) {
                         try {
-                            Uri uri = intent.getData();
-                            String filePath = UriToFilePath.getPathFromUri(CommCareApplication.instance(), uri);
-                            if (filePath != null) {
-                                File f = new File(filePath);
-                                if (f != null && f.exists()) {
-                                    formAndDataSyncer.performCustomRestoreFromFile(this, f);
+                            String filePath = DeveloperPreferences.getCustomRestoreDocLocation();
```
amstone326: Would be nice to pull this chunk of code out into its own method, since there's now a lot going on here

app/src/org/commcare/logic/ArchivedFormRemoteRestore.java
```diff
@@ -18,7 +19,7 @@
 
 /**
  * Only used for developer debugging.
- *
+ * <p>
```
amstone326: no-op?

app/src/org/commcare/activities/HomeScreenBaseActivity.java
```diff
@@ -374,16 +374,23 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
+                            if (filePath != null && !filePath.isEmpty()) {
+                                if(FilenameUtils.getExtension(filePath).contentEquals("xml")) {
+                                    File f = new File(filePath);
+                                    if (f.exists()) {
+                                        formAndDataSyncer.performCustomRestoreFromFile(this, f);
+                                    } else {
+                                        Toast.makeText(this, Localization.get("custom.restore.file.not.exist"), Toast.LENGTH_LONG).show();
+                                    }
+                                }else {
```
amstone326: space needed

app/src/org/commcare/activities/HomeScreenBaseActivity.java
```diff
@@ -374,16 +374,23 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
                         try {
-                            Uri uri = intent.getData();
-                            String filePath = UriToFilePath.getPathFromUri(CommCareApplication.instance(), uri);
-                            if (filePath != null) {
-                                File f = new File(filePath);
-                                if (f != null && f.exists()) {
-                                    formAndDataSyncer.performCustomRestoreFromFile(this, f);
+                            String filePath = DeveloperPreferences.getCustomRestoreDocLocation();
+                            if (filePath != null && !filePath.isEmpty()) {
+                                if(FilenameUtils.getExtension(filePath).contentEquals("xml")) {
```
amstone326: space needed

app/src/org/commcare/preferences/DeveloperPreferences.java
```diff
@@ -236,14 +196,16 @@ public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, Strin
...
                 setSessionEditText();
-
                 break;
-            case LOAD_FORM_PAYLOAD_AS:
-                if (!formLoadPayloadStatus().equals(FormRecord.STATUS_SAVED)) {
-                    // clear submission server so that 'unsent' forms that are loaded don't get sent to HQ
-                    CommCareApplication.instance().getCurrentApp().getAppPreferences().edit()
-                            .putString(CommCareServerPreferences.PREFS_SUBMISSION_URL_KEY, "")
-                            .apply();
+            case PREFS_CUSTOM_RESTORE_DOC_LOCATION:
```
amstone326: nice, great that we can move the handling of this here instead of `onActivityResult`

app/src/org/commcare/preferences/FilePreferenceDialogFragmentCompat.java
```diff
@@ -0,0 +1,71 @@
...
+                String fileType = ((FilePreference)getPreference()).getFileType();
```
amstone326: Since you're checking for the proper file type as soon as it is selected here, I think all of the other places downstream from here where you've added checks for file type matching are unnecessary. Let me know if I'm missing something in that analysis

amstone326:
@shubham1g5 Ok, left one last round of comments. Sorry for all the back-and-forth, but should be really close now.

app/res/values/attrs.xml
shubham1g5: I would definitely prefer standardising it for all dialogues, though I was just trying to keep the earlier behaviour of showing dialog title specific to the feature requesting the file.

app/res/xml/preferences_developer.xml
shubham1g5: Alright, though for future purposes would you be just able to confirm it by putting the enabled attr above defaultValue and formatting the xml file and see whether AS rearranges the enables after defaultValue. This style file  https://github.com/dimagi/commcare-android/blob/master/.android_studio_settings/codestyles/CommCare%20Coding%20Style.xml#L177 seems to indicate ordering as well but I am not really sure.

app/src/org/commcare/preferences/FilePreferenceDialogFragmentCompat.java
shubham1g5: The problem is user can still change the filePath in the editText right now. I will see if I can validate the editText as well here itself.

app/res/values/attrs.xml
amstone326: I think it's fine and good to just change this

app/src/org/commcare/preferences/FilePreferenceDialogFragmentCompat.java
amstone326: Ah yea I forgot about that. Could you just do the validation when the users presses "OK" on the dialog?

PR #1781, opened 2017-08-03
Prevent deleting unsent forms

amstone326:
This PR makes the following changes to prevent users from deleting unsent form records:

- The option to delete a form record on long-click doesn't appear for unsent forms:
screenshot

- In place of the delete option, an option to quarantine a form has been added to the end of the record integrity scan result dialog:
screenshot_20170803-125311

@ctsims As it turns out, the ability to move a quarantined form back into the unsent queue already exists:
screenshot_20170803-125557

### DIFF
--- a/app/res/layout/custom_alert_dialog.xml
+++ b/app/res/layout/custom_alert_dialog.xml
@@ -83,7 +83,7 @@
 
                     android:layout_toLeftOf="@id/neutral_button"
 
-                    android:layout_marginRight="@dimen/standard_spacer"
+                    android:layout_marginRight="@dimen/standard_spacer_double"
                     android:layout_marginLeft="@dimen/standard_spacer"
 
                     android:background="@android:color/transparent"
--- a/app/res/layout/custom_alert_dialog.xml
+++ b/app/res/layout/custom_alert_dialog.xml
@@ -83,7 +83,7 @@
 
                     android:layout_toLeftOf="@id/neutral_button"
 
-                    android:layout_marginRight="@dimen/standard_spacer_double"
+                    android:layout_marginRight="@dimen/standard_spacer"
                     android:layout_marginLeft="@dimen/standard_spacer"
 
                     android:background="@android:color/transparent"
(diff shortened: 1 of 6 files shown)


app/res/layout/custom_alert_dialog.xml
```diff
@@ -83,7 +83,7 @@
 
                     android:layout_toLeftOf="@id/neutral_button"
 
-                    android:layout_marginRight="@dimen/standard_spacer"
+                    android:layout_marginRight="@dimen/standard_spacer_double"
```
ctsims: Hm, quick note that this is a broad-cutting change which could introduce UI changes which were unintentional. I'd be more comfortable pulling it with confirmation of whether dialogs with three elements look fine, still, since the neutral button will now have asymetric space between it and the other buttons

app/src/org/commcare/activities/FormRecordListActivity.java
```diff
@@ -436,16 +443,38 @@ public void run() {
...
+                quarantineRecord(record);
+                dismissAlertDialog();
+            }
+        });
+
+        showAlertDialog(dialog);
+    }
+
+    private void quarantineRecord(FormRecord record) {
+        this.formRecordProcessor.updateRecordStatus(record, FormRecord.STATUS_QUARANTINED);
```
ctsims: This is a suggestion-not-requirement comment, but would it be useful to record in the same place as we do with the "quarantined from web" records a message that denotes "Quarantined manually by user" so there's no confusion/ambiguity?

app/res/layout/custom_alert_dialog.xml
amstone326: I actually didn't mean to leave this change in, because it didn't make any noticeable difference visually. Just reverted

app/src/org/commcare/activities/FormRecordListActivity.java
amstone326: By "in the same place", I assume you mean under the same log tag? Quarantines due to a local record issue are [currently logged under the tag `"error-storage"`](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/tasks/ProcessAndSendTask.java#L317) which I don't think is a semantically correct place to log this new type of quarantine. I have just updated the log message for the old log though, to specify the cause as being "due to local problem with record"

ctsims:
@amstone326 I think this is a good candidate for either a UI test or a robolectric test

amstone326:
@ctsims What would we want to test? Just that deleting an unsent form isn't possible?

ctsims:
@amstone326 Oh, I was thinking just the full end to end workflow. IE: Create One form, send it, go into airplane mode, create two forms, go here, quarantine both forms, go to quarantine, delete one, move one back to normal disable airplane mode, make sure form sends

PR #1783, opened 2017-08-08
Prompted update fixes

amstone326:


ctsims:
Nice fix! I think this is actually more readable in general as well. Let me know when it's ready.

amstone326:
@ctsims Ok this is good to go for now. The ICDS team responded that they _do_ want me to do the hotfix, but also that they want me to wait to do it until they've had a chance to do their own testing of the feature, I guess in case they find anything else that they'd like to change. So we should merge this for now, and if they do find anything else later on, I'll submit a new PR against the QA branch.

PR #1784, opened 2017-08-08
Expand Signed Privilege System to Allow for Multiple Privileges

ctsims:
V2 of privilege granting
tests for privilege blocks
tools for generating test signatures more easily
split out test and usage signatures for clarity

Still needs some testing, but review can start.


app/src/org/commcare/activities/GlobalPrivilegeClaimingActivity.java
```diff
@@ -101,72 +105,45 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
...
-                        } else {
-                            privilegeClaimFailed();
+                    String scanResult = data.getStringExtra("SCAN_RESULT");
+                    try {
+                        Pair<String, String[]> activatedPrivileges =
+                                new PrivilegesUtility(GlobalConstants.TRUSTED_SOURCE_PUBLIC_KEY).
+                                        processPrivilegePayloadForActivatedPrivileges(scanResult);
+
+                        for(String p : activatedPrivileges.second) {
+                            if (!GlobalPrivilegesManager.allGlobalPrivilegesList.contains(p)) {
```
amstone326: I know that you just copied this behavior from the code I had previously, but I'm now wondering why this is necessary and whether it might be harmful. Why would we want to prevent earlier versions of CommCare from being able to activate privileges that we hadn't yet conceived of at that point? This seems like it is severely limiting

app/src/org/commcare/activities/GlobalPrivilegeClaimingActivity.java
```diff
@@ -101,72 +105,45 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+                    } catch (PrivilegesUtility.PrivilagePayloadException e) {
+                        e.printStackTrace();
+                        privilegeClaimFailed();
                     }
                 }
         }
     }
 
-    private void privilegeClaimFailed() {
+    private void privilegeClaimTooNew() {
```
amstone326: Can you call this something like `privilegePayloadVersionTooNew()` instead? `privilegeClaimTooNew` sounds more to me like the situation where the device is trying to claim a privilege that it doesn't know exists yet

app/src/org/commcare/utils/PrivilegesUtility.java
```diff
@@ -0,0 +1,185 @@
...
+
+    public PrivilegesUtility(String authorityPublicKeyString) {
+        this.authorityPublicKeyString = authorityPublicKeyString;
+    }
+
+    public Pair<String, String[]> processPrivilegePayloadForActivatedPrivileges
+            (String privilegePayloadJson) throws PrivilagePayloadException{
+        try {
+            JSONObject obj = new JSONObject(privilegePayloadJson);
+            int version;
```
amstone326: Can lines 27-36 be a separate `getPayloadVersion()` method?

app/src/org/commcare/utils/SigningUtil.java
```diff
@@ -116,6 +116,20 @@ private static void assertWhitelistedUrlHost(String urlString) {
...
     public static String verifyMessageAndBytes(String message, byte[] signature) throws Exception {
         String keyString = GlobalConstants.TRUSTED_SOURCE_PUBLIC_KEY;
+        return verifyMessageAndBytes(keyString, message, signature);
+    }
+
+
+    /**
+     * Given the link and signature, verify the link using the public key
+     *
+     * @param message   the download link
```
amstone326: this javadoc comment seems wrong to me; `message` isn't a download link, it's just a string

app/src/org/commcare/utils/PrivilegesUtility.java
```diff
@@ -0,0 +1,185 @@
...
+    private Pair<String, String[]> processAndValidateV1Payload(JSONObject obj) throws JSONException, PrivilagePayloadException {
+        String[] payload = processV1Payload(obj);
+        if(validateV1PayloadSignature(payload[0], payload[1], payload[2])) {
+            return new Pair(payload[1], new String[] {payload[0]});
+        } else {
+            throw new InvalidPrivilegeSignatureException("Signatures don't match");
+        }
+    }
+
+    private boolean validateV1PayloadSignature(String flag, String username, String signature) throws JSONException, PrivilagePayloadException {
```
amstone326: This method is identical to `validateV2PayloadSignature()` except for how `expectedUnsignedValue` is generated. Can you make that variable an argument to the method, and then the rest of the code need not be duplicated?

app/src/org/commcare/utils/PrivilegesUtility.java
```diff
@@ -0,0 +1,185 @@
...
+
+import java.util.ArrayList;
+
+/**
+ * Helper methods for privilege data
+ *
+ * Created by ctsims on 8/7/2017.
+ */
+
+public class PrivilegesUtility {
```
amstone326: Can you add javadoc comments with @returns declarations to all of the methods in here? The fact that they're all just returning different configurations of `Pair`s and `String[]`s is really confusing without that

app/src/org/commcare/utils/PrivilegesUtility.java
```diff
@@ -0,0 +1,185 @@
...
+                default:
+                    throw new UnrecognizedPayloadVersionException();
+            }
+
+        } catch (JSONException e) {
+            throw new MalformedPayloadException(e);
+        }
+    }
+
+    private Pair<String, String[]> processAndValidateV2Payload(JSONObject obj) throws JSONException, PrivilagePayloadException {
```
amstone326: Even though it seems a little repetitive, I think the readability of this method and `procesesAndValidateV1Payload()` would be helped enormously if after calling `processV1Payload()` and `processV2Payload()`, respectively, you created named variables for each of the results, instead of just referencing them by `payload.first`, `payload[0]`, etc.

app/unit-tests/src/org/commcare/utils/PrivilegesUtilTest.java
```diff
@@ -0,0 +1,163 @@
...
+                        "}"
+        );
+    }
+
+    @Test(expected = PrivilegesUtility.InvalidPrivilegeSignatureException.class)
+    public void errorMalformedSignature() throws Exception {
+        new PrivilegesUtility(TEST_CERT_STRING).processPrivilegePayloadForActivatedPrivileges(
+                "{username:'privilege_test'," +
+                        "flags:['multiple_apps_unlimited']," +
+                        "multiple_flags_signature:'Ceci n`est pas une signature',"+
```
amstone326: lol

app/src/org/commcare/activities/GlobalPrivilegeClaimingActivity.java
ctsims: Honestly no strong feelings about it specifically, other than that coding in behavior we don't actually use could have unintended consequences. Happy to pull this out and just let the phone process ones it doesn't understand.

The only argument for a whitelist would be that in theory it kind of makes sense for the phone to wipe all *not* present privileges, which it can't do if it doesn't know the keys for all of the ones that have been activated. 

Anyway, happy to go either direction

app/src/org/commcare/utils/SigningUtil.java
ctsims: Yeah, just copied this from above, but will fix

ctsims:
@amstone326 Thanks for the review. Agreed will all comments and changed code to reflect all of your changes.

Only remaining question is whether to not process "not understood" privileges just yet. No strong feelings, happy for you to make the call on that. Let me know.

app/src/org/commcare/activities/GlobalPrivilegeClaimingActivity.java
amstone326: Mm ok thought through it a more and decided it doesn't really change much, since all it would mean is that someone could theoretically have a new privileges working automatically as soon as they performed an upgrade to a version of CommCare containing that privilege (if they had previously scanned a barcode containing it). But I don't know that there's much value in that, and could maybe result in weird/unexpected happenings. So I think fine to leave.

amstone326:
@ctsims Great, 2 remaining questions then:

1. Should we add the `advanced_settings` privilege in this PR? I'm not sure what the decision ended up being in terms of when that work is getting done on HQ, but if it might be soon, then I think we should go ahead and add it now. We could also do an intermediary step of 1 release where the implementation of `isAdvancedSettingsAccessEnabled()` relies on _either_ privilege being enabled, before we switch it over completely.
2. This PR is still waiting on you finishing all the tests, right?

ctsims:
... right, I totally forgot to actually turn off the whole reason I made the PR. I'll go ahead and do that.

confirmed that I still have a couple of tests to run

ctsims:
@amstone326 Tests are run and everything is working. Do you want me to nix the whitelist for new permissions, or no?

amstone326:
@ctsims No let's keep the whitelisting. This looks all set to me now except that the build is still failing.

ctsims:

PR #1792, opened 2017-08-17
Trigger a sync upon login after app update

amstone326:
Prompted by https://trello.com/c/is0IKxfA/140-trigger-sync-on-app-update


app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -462,15 +462,16 @@ public void tryAbort() {
...
         CommCareApplication.instance().getCurrentApp().getAppPreferences().edit()
-                .putLong("last-ota-restore", new Date().getTime()).commit();
+                .putLong(CommCarePreferences.LAST_SYNC_ATTEMPT, new Date().getTime()).apply();
+        CommCarePreferences.setPostUpdateSyncNeeded(false);
     }
 
     private static void recordSuccessfulSyncTime(String username) {
         CommCareApplication.instance().getCurrentApp().getAppPreferences().edit()
-                .putLong(SyncDetailCalculations.getLastSyncKey(username), new Date().getTime()).commit();
+                .putLong(SyncDetailCalculations.getLastSyncKey(username), new Date().getTime()).apply();
```
wpride: nice call

app/src/org/commcare/activities/HomeScreenBaseActivity.java
```diff
@@ -1010,13 +1010,14 @@ private void formEntry(Uri formUri, FormRecord r, String headerTitle) {
...
 
-    /**
-     * Triggered when an automatic sync is pending
-     */
-    private void handlePendingSync() {
-        long lastSync = CommCareApplication.instance().getCurrentApp().getAppPreferences().getLong("last-ota-restore", 0);
-        String footer = lastSync == 0 ? "never" : SimpleDateFormat.getDateTimeInstance().format(lastSync);
-        Logger.log(LogTypes.TYPE_USER, "autosync triggered. Last Sync|" + footer);
+    private void triggerSync(boolean triggeredByAutoSyncPending) {
+        if (triggeredByAutoSyncPending) {
```
wpride: Is the thinking here that logging this sync attempt will be redundant when we've just updated (as we know that we'll now always attempt syncing in this event)?

amstone326: No I just did this because the message that's being logged here is about an auto-sync specifically, and supposedly what's meant to be conveyed by this log message is that we're triggering one now, along with when we last triggered one.

PR #1793, opened 2017-08-18
Side by Side 'Compact' Numeric Entry

shubham1g5:
Case - https://manage.dimagi.com/default.asp?259573#1379573

- Passes a new parameter inCompactGroup to QuestionWidgets and sets layout properties based on it

Release Note: Adds ability to put question label and input side by side for Numeric inputs.

cross-request: https://github.com/dimagi/commcare-core/pull/630


shubham1g5:

shubham1g5:

app/src/org/commcare/views/widgets/StringWidget.java
```diff
@@ -80,9 +85,20 @@ public StringWidget(Context context, FormEntryPrompt prompt, boolean secret) {
             }
         }
 
+        if(isInCompactMode()){
```
shubham1g5: Can be part of xml

shubham1g5:

shubham1g5:

ctsims:

app/src/org/commcare/views/widgets/StringWidget.java
```diff
@@ -80,7 +85,18 @@ public StringWidget(Context context, FormEntryPrompt prompt, boolean secret) {
             }
         }
 
-        addView(mAnswer);
+        if(isInCompactMode()){
```
amstone326: space needed here and on line 90

app/src/org/commcare/views/widgets/DecimalWidget.java
```diff
@@ -37,18 +37,15 @@
...
+    public DecimalWidget(Context context, FormEntryPrompt prompt, boolean secret, boolean compact) {
+        super(context, prompt, secret, compact);
 
         // formatting
         mAnswer.setTextSize(TypedValue.COMPLEX_UNIT_DIP, mAnswerFontSize);
         mAnswer.setImeOptions(EditorInfo.IME_FLAG_NO_EXTRACT_UI | EditorInfo.IME_ACTION_NEXT);
 
         // needed to make long readonly text scroll
         mAnswer.setHorizontallyScrolling(false);
-        if (!secret) {
```
amstone326: Why did this get deleted (both here and in `IntegerWidget`)?

app/res/layout/question_widget_text_compact.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<TextView xmlns:android="http://schemas.android.com/apk/res/android"
+          android:layout_width="0dp"
+          android:layout_weight="1"
```
amstone326: My understanding is that `layout_weight` only does something useful when that view's parent has defined the property `weightSum`. I'm not seeing that anywhere in the PR, so what does this do?

app/res/layout/question_widget_text_compact.xml
```diff
@@ -0,0 +1,9 @@
...
+          android:layout_height="match_parent"
+          android:ellipsize="end"
+          android:maxLines="1"
```
amstone326: Is this actually what we want? I wasn't imagining that a requirement for this would be limiting the question text to a single line

app/res/layout/question_widget_text_compact.xml
```diff
@@ -0,0 +1,9 @@
...
+          android:scrollHorizontally="false"
+          android:text="QuestionWidget placeholder text: there should be a really big question text here, greater than the screen&apos;s size"/>
```
amstone326: Do you know what having this here does? I know it's present in `question_widget_text.xml` as well, but I don't know why

app/res/values-xlarge/dimens.xml
```diff
@@ -12,4 +12,5 @@
     <dimen name="row_padding_vertical">8sp</dimen>
     <dimen name="row_padding_horizontal">12sp</dimen>
 
+    <dimen name="edit_text_question_widget_compact_width">120dp</dimen>
```
amstone326: If this needs to be overridden in `values-xlarge` and `values-large`, should it be overridden for other sizes as well?

app/res/layout/question_widget_text_compact.xml
shubham1g5: If no weightSum is specified, it is computed as the sum of all layout_weight values. We can set it explicitly as well if it's more clear.

app/res/layout/question_widget_text_compact.xml
shubham1g5: Hmm. The spec [here](https://docs.google.com/document/d/19FqLRCWQcXkMIgsR5-Isu4tHHTGx5El2cXKuEwh882I/edit#heading=h.j1x6lwgn57a) do expect us to ellipsize it at end and fit it in a single line.

app/res/layout/question_widget_text_compact.xml
shubham1g5: It's just a placeholder. Doesn't really do anything. Should remove it from both places or use `tools:text` instead of `android:text` which would make it appear only in Android Studio xml preview.

app/res/values-xlarge/dimens.xml
shubham1g5: Yeah. I think `values-sw720dp-land` will correspond to `xlarge` bucket as well. So going to add it there. Apart from that it looks good to me.

shubham1g5: Actually no, since android will automatically use the best matching dimension value depending on the screen size so we don't need  to specify it in `values-sw720dp-land` as well

app/src/org/commcare/views/widgets/DecimalWidget.java
shubham1g5: This is already there in `StringWidget` [here](https://github.com/dimagi/commcare-android/blob/compactAppearance/app/src/org/commcare/views/widgets/StringWidget.java#L67). So no point including them twice.

shubham1g5:
@amstone326 regarding the formatting can you check if AS auto format doesn't cause these space diffs for you for `QuestionsView` class. I really do want to sort out my style file if there is indeed some issue with it.

app/src/org/commcare/views/widgets/DecimalWidget.java
amstone326: 👍

app/res/layout/question_widget_text_compact.xml
amstone326: Hmm but this is the only place that `layout_weight` is being used, so wouldn't that make the effect of this that this view will take up the whole screen? I was assuming the goal was for it to take up half of it, but maybe I'm missing something. (That also makes me realizing that including some screenshots in the PR description would be helpful; would you mind doing that now?)

app/res/layout/question_widget_text_compact.xml
amstone326: Ok let's do that

app/res/layout/question_widget_text_compact.xml
amstone326: Hmm yea I think that's my takeaway from reading the spec as well. I just want to be sure because that does sound fairly limiting to me -- @ctsims Is it intended that the _question text_ (as well as the answer) should be limited to 1 line only? The reason I'm asking is because if the screen width is already being split in half between the question text and answer text, it seems like _very_ little text would fit on 1 line.

amstone326:
@shubham1g5 Regarding the auto-formatting -- To be completely honest I'm not very familiar with how this works in Android Studio, but it definitely seems like your style file is making more aggressive auto-changes than mine. I don't get these sorts of automatic whitespace changes when I edit files. I think this is something we should address separately, but either way it's best to make sure that lots of formatting no-ops don't get included in functional PRs, because it makes the diffs harder to read. I would be all for making a separate push to get our formatting standardized, if you're interested in taking some time to spearhead that.

shubham1g5:

app/res/layout/question_widget_text_compact.xml
shubham1g5: Added some screenshots. So we want the question text to take all of the space available apart from the input edit text plus some padding which is what exactly `layout_weight` 1 is doing in this case.

app/res/layout/question_widget_text_compact.xml
ctsims: I think it's fairly important for the utility of the widget that the app compacts things. 1 line is very little text, but generally speaking we'd expect people to be using a label above to contextualize the entry points, like

Current Stock Levels:
AZT:                     ____
Combinex:           ____
...

so I think it's fine to set this limitation for now.

amstone326: Makes sense, thanks for the explanation.

app/res/layout/question_widget_text_compact.xml
amstone326: Ok cool, the screenshots help for sure. Looks good to me.

PR #1796, opened 2017-08-21
Replaces ACRA with Crashlytics

shubham1g5:
Case: https://manage.dimagi.com/default.asp?258564


shubham1g5:

app/src/org/commcare/utils/CrashUtil.java
```diff
@@ -0,0 +1,56 @@
...
+import org.commcare.CommCareApplication;
+import org.commcare.android.logging.ReportingUtils;
+import org.commcare.dalvik.BuildConfig;
+
+import io.fabric.sdk.android.Fabric;
+
+/**
+ * Contains constants and methods used in Crashlytics reporting.
+ *
+ * Created by wpride1 on 3/3/15.
```
amstone326: Should change this to your name!

app/src/org/commcare/activities/DispatchActivity.java
```diff
@@ -63,7 +64,7 @@
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-
+        CrashUtil.init(this);
```
amstone326: What was the reason for moving this from `CommCareApplication` to here?

app/src/org/commcare/utils/CrashUtil.java
```diff
@@ -0,0 +1,56 @@
...
+public class CrashUtil {
+
+    private static final String APP_VERSION = "app_version";
+    private static final String APP_NAME = "app_name";
+    private static final String DOMAIN = "domain";
+    private static final String DEVICE_ID = "device_id";
+
+    private static boolean crashlyticsEnabled = BuildConfig.USE_CRASHLYTICS;
+
+    public static void reportException(Exception e) {
```
amstone326: I know you were just copying the prior behavior of `AcraUtil`, but I'm confused as to why the only place this is being called is when a `CommCareTask` encounters an exception. @wpride Can you weigh in is at why that was the case for ACRA? I would have thought we would also be reporting things to ACRA in lots of other places, namely in `CommCareExceptionHandler`.

app/src/org/commcare/utils/CrashUtil.java
```diff
@@ -0,0 +1,56 @@
...
+        if (crashlyticsEnabled) {
+            Crashlytics.logException(e);
+        }
+    }
+
+    public static void init(Context context) {
+        if(crashlyticsEnabled) {
```
amstone326: space needed between `if` and `(`

app/build.gradle
```diff
@@ -112,6 +113,23 @@ afterEvaluate {
...
+        initFabricPropertiesIfNeeded()
+    }
+}
+
+def initFabricPropertiesIfNeeded() {
+    def propertiesFile = file('fabric.properties')
+    if (!propertiesFile.exists()) {
+        def commentMessage = "This is autogenerated fabric property from system environment to prevent key to be committed to source control."
+        ant.propertyfile(file: "fabric.properties", comment: commentMessage) {
+            entry(key: "apiSecret", value: FABRIC_API_SECRET)
```
amstone326: where do these come from?

shubham1g5: I have added these on Jenkins machine gradle.properties file

app/src/org/commcare/activities/DispatchActivity.java
shubham1g5: Since this the only entry point for the application, I added it here.  It can go in `CommCareApplication` as well.

app/src/org/commcare/utils/CrashUtil.java
wpride: @amstone326 kind of spit balling here since I wrote this a while ago and don't remember original intent well, but all actual *crashes* were caught and reported through the '@ReportsCrashes` annotation on `CommCareApplication` removed [here]( https://github.com/dimagi/commcare-android/pull/1796/files#diff-6702335a8e0e670e374872510c46a8b5L111)

I believe this call was necessary only in cases where we were catching and displaying errors to the user *without crashing* but we still wanted to crash to get the ACRA. Its possible there are other places where we should be doing this, or that we should remove this (IE maybe if we think users can debug the issue then we don't need it on our servers).

wpride: @shubham1g5 should confirm that something like the `ReportsCrashes` annotation exists here that automatically reports all hard crashes

app/src/org/commcare/android/logging/ReportingUtils.java
```diff
@@ -104,4 +104,15 @@ public static String getCommCareVersionString() {
         }
     }
 
+    public static int getAppVersion() {
+        return CommCareApplication.instance().getCurrentApp().getAppRecord().getVersionNumber();
```
wpride: I think these will crash if there isn't an app successfully seated yet. Usually good to wrap these reporting helpers in a `try`/`catch` so that we guarantee we get the report

wpride: Also any reason not to use `getVersion` above? Think it gives more diagnostic information

wpride:
Couple comments but looks great! I think this will be a game changer

app/src/org/commcare/utils/CrashUtil.java
shubham1g5: So all the hard crashes get automatically intercepted. I have confirmed it by triggering a crash manually. Though the exceptions we are catching won't make it to the Crashlytics without calling this `reportException` explicitly. I am in favour of reporting all exceptions to Crashlytics but since we already send those to HQ not sure if we would want to do that.

wpride: Cool, sounds fine to me.

app/src/org/commcare/android/logging/ReportingUtils.java
shubham1g5: Yeah will wrap it in try catch. And from the `getCommCareVersionString` we get an output like `CommCare Android, version \"2.38.1\"(439996). App v3490. CommCare Version 2.38. Build 439996, built on: 2017-08-25` .  Crashlytics automatically adds the Commcare version so All I wanted to get is ccz version to make errors searchable on that. Other information doesn't really seem useful.

wpride: (yes)

app/src/org/commcare/utils/CrashUtil.java
amstone326: @shubham1g5 If I'm understanding what you said correctly, that means that any crashes which are caught and stopped [here](https://github.com/dimagi/commcare-android/blob/97bba52eabe2d6b47c606f28beea67f5ecc289b9/app/src/org/commcare/utils/CommCareExceptionHandler.java#L36-L39) won't end up on Crashlytics as of now. If that's correct, can we just add an explicit call to `reportException()` within that if statement as well?

app/src/org/commcare/activities/DispatchActivity.java
amstone326: Hm, I imagine it shouldn't go in 2 places, I just wasn't sure why it was necessary to move where this happens, since `CommCareApplication.onCreate()` will also always get called when the app starts up.

PR #1798, opened 2017-08-21
2.36.3 - Re-add the logic to hide 'Report Problem'

amstone326:
(I'm only adding this cross request comment so that the PR builder will run with a version of CC that is compatible; the code in these 2 PRs is technically totally independent)

cross-request: https://github.com/dimagi/commcare-core/pull/633

### DIFF
--- a/app/res/xml/preferences_developer.xml
+++ b/app/res/xml/preferences_developer.xml
@@ -130,5 +130,10 @@
         android:entryValues="@array/pref_enabled_vals"
         android:key="cc-enable-bulk-performance"
         android:title="Enable cutting-edge performance improvements"/>
-
+    <ListPreference
+        android:enabled="true"
+        android:entries="@array/pref_enabled_labels"
+        android:entryValues="@array/pref_enabled_vals"
+        android:key="cc-hide-issue-report"
+        android:title="Hide the 'Report an Issue' Menu Option"/>
 </PreferenceScreen>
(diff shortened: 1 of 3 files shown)


app/res/xml/preferences_developer.xml
```diff
@@ -130,5 +130,10 @@
         android:entryValues="@array/pref_enabled_vals"
         android:key="cc-enable-bulk-performance"
         android:title="Enable cutting-edge performance improvements"/>
-
+    <ListPreference
```
amstone326: @ctsims The original implementation of this that got overriden had the `defaultValue` for this `ListPreference` set to `"yes"`. This seemed wrong to me, particularly since the `shouldHideReportIssue()` method was implemented to call `doesPropertyMatch()` with `NO` as the default value, so I've removed that in this version. But I just want to make sure I'm not missing something.

ctsims: Hm. I'm trying to remember whether we snuck in this being on by default for LTS only in 2.27, which is why it didn't get ported over.

As ported currently this implementation looks correct.

amstone326:

app/res/xml/preferences_developer.xml
ctsims: @amstone326 ok, validated that the project has the flag set correctly in their app preferences, so this looks good to go. Good catch.

PR #1802, opened 2017-08-23
Fix for being able to process forms with errors half-way

amstone326:
SO, we have finally discovered the primary cause of the missing reg. forms mystery. Previously, when CommCare encountered an exception while processing a form instance, it would wipe the form record from the db, but would ALSO allow any local db transactions that had already taken place before the exception to remain intact. This is how it was possible for a user to submit update forms against a case that the server had no notion of ever being created. It also explains why users would often fill out a duplicate registration form immediately afterward, because the user is given a very prominent error message immediately upon form completion when this type of error happens. Having now seen this happen in action (Clayton helpfully created a form to reproduce the behavior [here](https://www.commcarehq.org/a/ctsims/apps/view/3bff10662205c18fcd1bbf5430a62aa3/modules-1/forms-0/source/)), it makes a ton of sense why users would respond by filling out the same form again. 

This PR makes the following changes to fix this flaw going forward:
- Wrap form processing (both on form-save and pre-form-sending) in a transaction, so that if any part of the processing fails, the effects of the entire form on the local user db are rolled back
- In `handleExceptionFromFormProcessing()`, quarantine the problematic record instead of deleting it. (Although I'm actually now wavering on whether this change is right, or if we should in fact just be deleting the record, since there is no chance of ever recovering it in this case. @ctsims or @wpride Do you have an opinion on this?)
- Add handling in `sendForms()` that expects some records to have status of `STATUS_QUARANTINED` (due to the change above)

Some other small changes that aren't directly related to fixing this situation:
- Show a specific error message for when `FormUploadResult.RECORD_FAILURE` is returned by `ProcessAndSendTask`
- I came across 2  existing things we were doing in `ProcessAndSendTask` that looked not-right to me, so I've made the change I think is appropriate, and commented with questions below to get confirmation


app/src/org/commcare/tasks/ProcessAndSendTask.java
```diff
@@ -334,6 +328,7 @@ private void sendForms(FormRecord[] records) throws TaskCancelledException {
                             CommCareApplication.notificationManager().reportNotificationMessage(
                                     NotificationMessageFactory.message(ProcessIssues.RecordFilesMissing), true);
                             FormRecordCleanupTask.wipeRecord(c, record);
+                            results[i] = FormUploadResult.RECORD_FAILURE;
```
amstone326: Question1: When a form file is missing when we try to submit, I think we should be counting this is as a `RECORD_FAILURE`, right? With the code as-is, the result for this form will be `FAILURE` (since that's what all values in the `results` array are initialized to to begin with), which is technically incorrect.

app/src/org/commcare/tasks/ProcessAndSendTask.java
```diff
@@ -188,6 +188,7 @@ private boolean checkFormRecordStatus(FormRecord[] records)
                         record.logPendingDeletion(TAG,
                                 "the xml submission file associated with the record could not be found");
                         FormRecordCleanupTask.wipeRecord(c, record);
+                        records[i] = FormRecord.StandInForDeletedRecord();
```
amstone326: Question 2: What was happening here previously seems really shady to me, because the form gets wiped from the db, but then there is still a stale handle to that `FormRecord` object in the `records` array, which is what `sendForms()` goes on to use after this method call finishes. So it looks to me like we would have still been trying to send that stale record when we get to it in `sendForms()`, with unclear results. 

I'm proposing this as an alternate strategy, where we create a stand-in record with its own status type (`STATUS_JUST_DELETED`), and then check for this status explicitly in each iteration of `sendForms()` and consider the result for a record with this status to be `RECORD_FAILURE`

app/src/org/commcare/tasks/ProcessAndSendTask.java
```diff
@@ -174,60 +177,65 @@ private boolean checkFormRecordStatus(FormRecord[] records)
...
+                SQLiteDatabase db =
+                        CommCareApplication.instance().getUserDbHandle();
+                db.beginTransaction();
                 try {
                     records[i] = processor.process(record);
+                    db.setTransactionSuccessful();
                 } catch (InvalidStructureException | XmlPullParserException |
                         UnfullfilledRequirementsException e) {
-                    handleExceptionFromFormProcessing(record, e);
+                    db.endTransaction();
```
amstone326: The reason all of these calls to `endTransaction()` need to be separated out (instead of being called just once in a `finally` block) is that the transaction needs to be ended _before_ the catch blocks execute their responses to the exceptions, which in some cases make db modifications that we do NOT want to be rolled back

app/src/org/commcare/provider/InstanceProvider.java
```diff
@@ -69,7 +69,7 @@ public String getAppId() {
         }
 
         @Override
-        public void onCreate(SQLiteDatabase db) {
+        public void onCreate(android.database.sqlite.SQLiteDatabase db) {
```
wpride: Why the full path instead of importing?

app/src/org/commcare/provider/InstanceProvider.java
```diff
@@ -542,10 +542,14 @@ private void linkToSessionFormRecord(Uri instanceUri) {
             // before moving on. We'll catch any errors here and just eat them
             // (since the task will also try the process and fail if it does.
             if (FormRecord.STATUS_COMPLETE.equals(current.getStatus())) {
+                net.sqlcipher.database.SQLiteDatabase db =
```
wpride: same question

app/src/org/commcare/android/database/user/models/FormRecord.java
```diff
@@ -99,6 +112,9 @@
     @MetaField(META_SUBMISSION_ORDERING_NUMBER)
     private String submissionOrderingNumber;
 
+    @Persisting(value = 9, nullable = true)
```
wpride: Confirming: a migration isn't necessary because `nullable` is true

app/src/org/commcare/android/database/user/models/FormRecord.java
```diff
@@ -133,6 +149,12 @@ public FormRecord updateInstanceAndStatus(String instanceURI, String newStatus)
         return fr;
     }
 
+    public static FormRecord StandInForDeletedRecord() {
```
wpride: Why is the first letter capitalized?

app/src/org/commcare/tasks/ProcessAndSendTask.java
wpride: Another option would be to have the `wipeRecord` method change the status of the original record object to `STATUS_JUST_DELETED`. No strong feelings here, agree that the previous behavior feels dangerous.

app/src/org/commcare/tasks/ProcessAndSendTask.java
wpride: Makes sense. Another option would be to use a `finally` block that uses `inTransaction` to check whether a previous `catch` has closed the transaction already and only closes if not. Probably doesn't make this much cleaner

app/src/org/commcare/tasks/ProcessAndSendTask.java
```diff
@@ -174,60 +177,65 @@ private boolean checkFormRecordStatus(FormRecord[] records)
...
                                 NotificationMessageFactory.message(ProcessIssues.StorageRemoved), true);
                         //Otherwise, the SD card just got removed, and we need to bail anyway.
                         throw e;
                     }
                 } catch (IOException e) {
+                    db.endTransaction();
                     Logger.log(LogTypes.TYPE_ERROR_WORKFLOW, "IO Issues processing a form. " +
                             "Tentatively not removing in case they are resolvable|" + getExceptionText(e));
+                } catch (Exception e) {
+                    db.endTransaction();
```
wpride: Why are we adding this? To ensure that we close the database handle regardless? If so seems like we should log this rather than fail silently on any non-caught exception

app/src/org/commcare/activities/FormRecordListActivity.java
```diff
@@ -397,8 +397,12 @@ public void onCreateContextMenu(ContextMenu menu, View v, ContextMenuInfo menuIn
             menu.add(Menu.NONE, DELETE_RECORD, DELETE_RECORD, Localization.get("app.workflow.forms.delete"));
         }
 
-        if (FormRecord.STATUS_QUARANTINED.equals(value.getStatus())) {
-            menu.add(Menu.NONE, RESTORE_RECORD, RESTORE_RECORD, Localization.get("app.workflow.forms.restore"));
+        if (FormRecord.STATUS_QUARANTINED.equals(value.getStatus()) &&
+                !FormRecord.QuarantineReason_LOCAL_PROCESSING_ERROR.equals(value.getReasonForQuarantine())) {
```
wpride: It seems a bit weird that we're comparing what are essentially two human-readable error messages for equality here (rather than using, say, an Enum or String key). From what I can tell the semantic meaning of these strings is never used (IE the error messages saved under "reason" are never displayed to the user). Is there any reason not to turn these into an enum or simple keys?

app/src/org/commcare/tasks/ProcessAndSendTask.java
amstone326: Hm, I think I prefer this way because I think it's cleaner to represent the record with something separate, rather than to try to hang onto a stale pointer to a deleted record.

app/src/org/commcare/activities/FormRecordListActivity.java
amstone326: Ha yea fair point, so this looks a little weirder than it actually was in my mind because I implemented this PR with an eye to the fact that https://github.com/dimagi/commcare-android/pull/1721 will hopefully get merged sometime soon-ish (waiting on HQ work with no clear timeline). But I essentially copied over a bit of what was already in that PR to here, and there are further changes in that PR that would actually be using this string as something to display to the user. 

As far as why not to make this an enum, my reason was that the `reasonForQuarantine` field is set as a `@Persisting` value in `FormRecord`, which I believe means it has to be a primitive (although I could be wrong about that?). So once I was making these variables Strings anyway, I figured it didn't actually matter whether the string values were short strings or long strings, and so they could double as identifiers and messages. 

Let me know if you still think this is a bad idea given those explanations.

app/src/org/commcare/android/database/user/models/FormRecord.java
amstone326: Right, the `nullable` makes it not necessary, and there is no other reason to do one, since we would have to make a random guess as to the reason for pre-existing records.

app/src/org/commcare/android/database/user/models/FormRecord.java
amstone326: I thought that was the java convention for static methods that serve as constructors. If you think I'm making that up, let me know...

app/src/org/commcare/provider/InstanceProvider.java
amstone326: Because I'm referencing both `android.database.sqlite.SQLiteDatabase` and `net.sqlcipher.database.SQLiteDatabase` in the same class. I think the easiest way to that is to not have an import statement for either one, and just reference them both by their full path

app/src/org/commcare/tasks/ProcessAndSendTask.java
amstone326: Ah wow, good catch. I think what I should probably do is just preserve the old behavior precisely, by re-throwing the exception after ending the transaction. Does that sound right to you?

wpride: :+1:

amstone326:
@wpride Think I responded to all the actionable questions, and made 1 resulting code change. There are a couple of lingering questions from me, if you can address those.

@ctsims I agree that I'd like your eyes on this as well before it gets merged.

app/src/org/commcare/activities/FormRecordListActivity.java
ctsims: Generally agreed that it doesn't seem quite right for this to serve double-duty, and it also has the disadvantage that if we change the error message it will (quite unexpectedly) result in the forms being restorable.

One alternative is to simply store the "status before quarantine", rather than flagging as "unrecoverable" essentially, so that records which are quarantined from "unprocessed" get returned to "unprocessed" rather than "unsent"

app/src/org/commcare/android/database/user/models/FormRecord.java
ctsims: I actually didn't realize that it was true that setting nullable made these not require migration, and find it kind of surprising. How does the processor know to stop reading values? If someone added a tenth value after this one and the phone jumped from 2.36 -> That Version I'm not sure I understand technically how it would know to figure out when to stop reading the stream... (Or is it that you can now extend perisistables all you want as long as all values are nullable after the last hard migraiton?)

app/src/org/commcare/provider/InstanceProvider.java
ctsims: It's fine to keep the import for the common one as long as you keep the fully qualified reference to the sqlcipher DB here. Not sure if that causes Lint issues or anyhting

ctsims: I'd rename this reference to "userDb" or something so that someone searching the file doesn't confuse which "db" object is being manipulated here (since most of hte time it's not this one, but they share the same name)

app/src/org/commcare/tasks/ProcessAndSendTask.java
ctsims: @wpride That pattern is actually quite dangerous because it wont' be able to identify a nested transaction (so you could still end up firing two "ends" for one "begin"

app/src/org/commcare/provider/InstanceProvider.java
```diff
@@ -542,10 +542,14 @@ private void linkToSessionFormRecord(Uri instanceUri) {
...
+                        CommCareApplication.instance().getUserDbHandle();
+                db.beginTransaction();
                 try {
                     new FormRecordProcessor(getContext()).process(current);
```
ctsims: I wonder if it would make more sense to move the transactionality here inside of the process event rather than here? It doesn't seem like the side effects of what happens in this method will change, and generally speaking it is ~always the right decision for an individual form to process transactionally, so there's no reason why we wouldn't want to change how process() works to protect the ability to do a non-transactional form process

app/src/org/commcare/tasks/ProcessAndSendTask.java
```diff
@@ -174,60 +177,68 @@ private boolean checkFormRecordStatus(FormRecord[] records)
...
+                SQLiteDatabase db =
+                        CommCareApplication.instance().getUserDbHandle();
+                db.beginTransaction();
                 try {
                     records[i] = processor.process(record);
+                    db.setTransactionSuccessful();
                 } catch (InvalidStructureException | XmlPullParserException |
                         UnfullfilledRequirementsException e) {
-                    handleExceptionFromFormProcessing(record, e);
+                    db.endTransaction();
```
ctsims: One other option for this pattern would be 

```
try{
  toRethrow = null
  beginTransaction
  try{
    stuff
    transactionSuccesful
  } catch(Exception e) {
    toRethrow = e
  } finally {
    endtransaction
  }
  if(rethrow != null) { throw rethrow }

} catch(actualerrorhandlinghere) {
}

````

File: app/src/org/commcare/tasks/ProcessAndSendTask.java (hunk 2 above)
Comment (ctsims) [reply app/src/org/commcare/tasks/ProcessAndSendTask.java]: StandIn approach makes sense to me. Agree that keeping the handle to the deleted record around feels quite dangerous if we truly think we should abandon it.

File: app/src/org/commcare/tasks/ProcessAndSendTask.java [hunk 12]
```diff
@@ -174,60 +177,68 @@ private boolean checkFormRecordStatus(FormRecord[] records)
...
                         CommCareApplication.notificationManager().reportNotificationMessage(
                                 NotificationMessageFactory.message(ProcessIssues.StorageRemoved), true);
                         //Otherwise, the SD card just got removed, and we need to bail anyway.
                         throw e;
                     }
                 } catch (IOException e) {
+                    db.endTransaction();
                     Logger.log(LogTypes.TYPE_ERROR_WORKFLOW, "IO Issues processing a form. " +
                             "Tentatively not removing in case they are resolvable|" + getExceptionText(e));
+                } catch (Exception e) {
```
ctsims: If you want this to execute for ~all code paths this should be a `Throwable`, since this will miss `Error` throws. I put an alternative possibility above that I think might be a bit cleaner (and less risky that someone forgets to end a transaction in a new catch())

app/src/org/commcare/tasks/ProcessAndSendTask.java
ctsims: That sounds correct to me, but I'm not as fresh on the return codes.

app/src/org/commcare/android/database/user/models/FormRecord.java
amstone326: Errr... I was making an assumption, which now that you say things that way, I think must be wrong. I'll add in a migration. sorry about that :/

app/src/org/commcare/tasks/ProcessAndSendTask.java
amstone326: @ctsims I liked how this looked in theory and just tried to implement it, but there's actually a compilation problem with it. In order to do this, we would have to change the method signature to include `throws Exception` since the compiler now interprets this method as being capable of throwing any type of Exception at all. I'm not convinced that that's a good trade-off, so I'm going to leave this the way I had it for now.

app/src/org/commcare/provider/InstanceProvider.java
amstone326: @ctsims I'm not sure how that would work. In order to ensure that `endTransaction()` gets called regardless of any exceptions encountered, that call _has_ to be made outside of the `process()` method itself (as is happening in the `finally` block here). While I theoretically could move the `beginTransaction()` and `setTransactionSuccessful` calls to _within_ `process()`, it seems worse to me to have different portions of the transaction-related code in different places; I would much rather it be getting clearly handled in one place. Let me know if I'm missing some other way of acheiving this.

app/src/org/commcare/activities/FormRecordListActivity.java
amstone326: Fixed this

app/src/org/commcare/tasks/ProcessAndSendTask.java
amstone326: Good point, changed

app/src/org/commcare/tasks/ProcessAndSendTask.java
ctsims: Hey, 

I was actually making that way more complex than it needed to be.

All you actually need to do here is
```
try {
  startTransaction()
  try {
    doCode();
    transactionSuccess();
  } finally {
    endTransaction();
  } 
} catch(TypedException e) {
  handle stuff
}
```

The interior finally() block will execute as soon as java leaves the scope of the interior try{} block before the exterior block takes the  control flow

app/src/org/commcare/android/database/user/models/FormRecordV3.java
```diff
@@ -0,0 +1,93 @@
+package org.commcare.android.database.user.models;
```
ctsims: This class needs to replace all references to the previous "FormRecord" class in the UserDatabaseUpgrader that existed, in addition to being used in the current upgrader.

`upgradeSixteenSeventeen() `, for example, should actually now be upgrading from `FormRecordV2` to `FormRecordV3`

amstone326: Ugh yea sorry, fixed

amstone326: By the way, as far as I can tell, `upgradeSixteenSeventeen()` is the only place this needed to happen. The only other reference to `FormRecord` is in `markSenseIncompleteUnsent()`, but that is being used by really old migrations, so I assume something must be different about that, or it would have had to change a long time ago

ctsims:
Awesome, looks great. Will merge after checks pass

PR #1807, opened 2017-08-29
Don't allow manual quarantine of Saved or already-quarantined forms

amstone326:
We shouldn't be exposing the button to manually quarantine a form after doing a record integrity scan when the record in question is already sent or already quarantined


wpride:
Merged but then had another thought - we probably only want to show this option when the form is `STATUS_COMPLETE` or `STATUS_UNSENT`, right?

shubham1g5:
Just curious, in what scenarios does the user uses this quarantine option. Or in general what use case does a manual quarantine serve ?

amstone326:
@shubham1g5 The ability to manually quarantine an unsent form is important to have as a backup option for when something unexpected has gone wrong with a form that is causing it to block submission of all other forms on the device. In theory, CommCare expects and handles all of the ways that this might happen, but it has come up before that we have missed something, and users have been left in a state where their phone is bricked and can't send anything until they get that one form out of the way.

@wpride I think you're right, I'll open a new PR to change it

PR #1808, opened 2017-08-29
Barcode scan using ZXing lib

wpride:


wpride:
Release Note: Removes the need for installing external barcode application for install

amstone326:
@wpride Is this still trying to get into 2.39? If so, merge conflicts need to be resolved

wpride:
@amstone326 this is good to go now

app/src/org/commcare/fragments/SelectInstallModeFragment.java
```diff
@@ -74,9 +75,8 @@ public void onClick(View v) {
                     if (currentActivity instanceof CommCareSetupActivity) {
                         ((CommCareSetupActivity)currentActivity).clearErrorMessage();
                     }
-                    Intent i = new Intent("com.google.zxing.client.android.SCAN");
-                    i.putExtra("SCAN_FORMATS", "QR_CODE");
-                    getActivity().startActivityForResult(i, CommCareSetupActivity.BARCODE_CAPTURE);
+                    Intent intent = new IntentIntegrator(getActivity()).createScanIntent();
```
amstone326: Are there any potential failure points of this call? Like only working above a certain Android API level or something? It seems worth testing this on a few different devices if you haven't already (or at least flagging to the QA team that they should be sure to do so themselves)

app/src/org/commcare/activities/CommCareSetupActivity.java
```diff
@@ -374,15 +374,12 @@ protected void onSaveInstanceState(Bundle outState) {
     @Override
     protected void onActivityResult(int requestCode, int resultCode, Intent data) {
         super.onActivityResult(requestCode, resultCode, data);
-
         String result = null;
         switch (requestCode) {
             case BARCODE_CAPTURE:
                 if (resultCode == Activity.RESULT_OK) {
-                    result = data.getStringExtra("SCAN_RESULT");
-                    String dbg = "Got url from barcode scanner: " + result;
```
ctsims: This output is actually _quite_ helpful sometimes and I'd like to keep it in,

wpride: Good call

wpride:
@amstone326 good call, replaced except for `BarcodeWidget` - that class is pretty paired with the `IntentCallout`/`IntentWidget` paradigm at the moment and don't want to pull it out this late

wpride:
I confirmed working on a 5.1 but couldn't find any older device

amstone326:
@wpride Proposing that we shift this back to 2.40 since there's nothing hinging on it and there hasn't been anything added to the QA plan for this

wpride:
That's fine but I don't think we really need to add anything? All these
usages should already be covered. If they can do the tests on a 4.X device
should be set

ctsims:
I agree on holding off until 2.40 for a different reason, which is that this is going to make whatever release it goes into not able to auto-update, since it extends a new permission (I think Camera is in that set), right?

amstone326:
@ctsims Oof I forgot about that. How exactly does a new permission change the update process?

ctsims:
@amstone326 Changes to the permissions set which change specific "High Risk" permissions make it the case that the user isn't able to receive an automatic update with the change (at least pre-android 7. This might be different with the new permissions framework, since you approve individually).

amstone326:
@ctsims @wpride In that case, is this even worth doing? That seems like a pretty huge hit to take for a nice-to-have change. Maybe we should wait until we have a reason that we _have_ to update permissions anyway, and include this in that release?

wpride:
@amstone326 @ctsims I believe that in 6.0+ we can change this to request the permission only when needed and not at install/update [link](https://developer.android.com/training/permissions/requesting.html)

If we decided this is too big a hit for pre-6.0 devices then I'm fine closing this out

ctsims:
Let's leave this open for now and make a call on a future release on when we don't think it will be disruptive to internalize the feature by requiring the permission (and blocking auto-update for that release).

amstone326:
@ctsims @wpride So funnily enough, it turns out that this PR would actually fix a crashlytics crash: https://fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/59cfdad6be077a4dcc8831b9?time=last-thirty-days. 

Background: https://stackoverflow.com/questions/11388450/using-zxing-barcode-scanner-causes-securityexception helped me understand that this crash will happen if 2 things are true:
1. A device has an application installed which claims in its manifest to handle barcode scanning by having an activity with an intent filter for `com.google.zxing.client.android.SCAN`, but also doesn't export that activity. 
2. The device does _not_ have a valid standard barcode scanner app installed.

The combination of these 2 things results in the intent routing system trying to use the barcode scanning activity of 1 of those random applications, and then getting that Permission Denial since it's not actually allowed to launch that activity. You can see in the Crashlytics stack traces for specific sessions of this crash the package identifiers of some of the various apps that can cause this (`com.alcatel.mcrm`, `com.deltecs.wipro`, etc.), and I was able to replicate the crash by installing one of them, uninstalling my normal barcode scanning apps, and then trying to install an app in CommCare via barcode scanning.

Using the `IntentIntegrator` paradigm was the recommended solution for this problem anyway, so this seems like a good reason for us to push this PR over the hump and release it.

wpride:
👍 sounds good to me. probably want the go ahead from @ctsims given the auto-update implications

ctsims:
Hey,

So now that the next release isn't critical for anyone I'm OK introducing a new manual-accept update.

amstone326:
Sweet. @wpride I updated this PR with the refactor needed to do the switch for the `BarcodeWidget` as well. Now that this is actually fixing a bug, it seemed undesirable to leave it out. Want to take a look at what I did before we merge?

PR #1810, opened 2017-09-01
Makes CommCareActivity implement DetailCalloutListener

shubham1g5:
Fixes: https://manage.dimagi.com/default.asp?260852

Not sure if we should add this implementation to the base `CommCareActivity` instead as thats where we are creating `BreadcrumbBarFragment` which has an instance of `TabbedDetailView`  and thus `EntityDetailPagerAdapter` and thus `EntityDetailFragment` which potentially can contain phone numbers etc. as layout fields. So it seems to me from code that every `CommCareActivity` should potentially handle these callouts.


amstone326:
Hey @shubham1g5 yea, I think what you are proposing is the correct way to do this. Because the inline case detail can show up on several different screens, this implementation should be added in more places. Adding it to `CommCareActivity` sounds like a good catch-all solution to me.

PR #1812, opened 2017-09-03
Async Badges fix

shubham1g5:
Case: https://manage.dimagi.com/default.asp?261071#1386375

* Refactors MenuAdapter to use view holder pattern
*  implement cache for badges so that we don't end up doing repeated calculation on listview scrolling. 

cross-request: https://github.com/dimagi/commcare-core/pull/637

### DIFF
--- a/app/src/org/commcare/adapters/MenuAdapter.java
+++ b/app/src/org/commcare/adapters/MenuAdapter.java
@@ -180,7 +180,6 @@ public class MenuAdapter extends BaseAdapter {
         } else {
             if (audioPlaybackButton != null) {
                 audioPlaybackButton.modifyButtonForNewView(viewId, audioURI, false);
-                audioPlaybackButton.setVisibility(View.GONE);
             }
         }
     }
--- a/app/src/org/commcare/adapters/MenuAdapter.java
+++ b/app/src/org/commcare/adapters/MenuAdapter.java
@@ -3,12 +3,12 @@ package org.commcare.adapters;
 import android.database.DataSetObserver;
 import android.graphics.Bitmap;
 import android.util.Log;
+import android.util.SparseArray;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.BaseAdapter;
 import android.widget.ImageView;
-import android.widget.LinearLayout;
 import android.widget.TextView;
 
 import org.commcare.CommCareApplication;
@@ -55,10 +55,11 @@ public class MenuAdapter extends BaseAdapter {
     private Exception loadError;
     private String errorMessage = "";
     final CommCareActivity context;
-    final MenuDisplayable[] displayableData;
+    private final MenuDisplayable[] displayableData;
     private final CompositeDisposable compositeDisposable = new CompositeDisposable();
+    private SparseArray<String> badgeCache = new SparseArray<>();
 
-    class MenuLogger implements LoggerInterface {
+    private class MenuLogger implements LoggerInterface {
 
         @Override
         public void logError(String message, Exception cause) {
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/adapters/MenuAdapter.java
```diff
@@ -213,25 +210,37 @@ public void setupImageView(ImageView mIconView, MenuDisplayable menuDisplayable,
...
+        if (badgeCache.get(position) != null) {
+            updateBadgeView(badgeView, badgeCache.get(position));
+        } else {
+            compositeDisposable.add(
+                    menuDisplayable.getTextForBadge(asw.getEvaluationContext(menuDisplayable.getCommandID()))
+                            .subscribeOn(Schedulers.computation())
+                            .observeOn(AndroidSchedulers.mainThread())
+                            .subscribe(badgeText -> {
+                                        // Make sure that badgeView corresponds to the right position and update it
+                                        if (((int)badgeView.getTag()) == position) {
```
amstone326: @shubham1g5 what makes this check necessary?

shubham1g5: If we don't do this and scroll through listview, we can get the result of the async request after that particular view has been recycled for another position. For ex. I made a request for badgeView which was at position 5 and immediately scroll downwards which will cause that particular badgeview to be recycled and get used for some other position say 15. So when we get back the result from the async request for data at position 5, we will update the badgeview for position 15 with position 5 data until we finally get the result for position 15 data.

app/src/org/commcare/adapters/MenuAdapter.java
```diff
@@ -178,12 +180,12 @@ private void setupAudioButton(int rowId, AudioPlaybackButton audioPlaybackButton
         } else {
             if (audioPlaybackButton != null) {
                 audioPlaybackButton.modifyButtonForNewView(viewId, audioURI, false);
-                ((LinearLayout)audioPlaybackButton.getParent()).removeView(audioPlaybackButton);
+                audioPlaybackButton.setVisibility(View.GONE);
```
shubham1g5: @amstone326 Where in HQ do we add audio for a menu ?  I want to test once for that this change won't cause any regressions!

wpride: If you select the options "Gear" icon on a menu the audio upload widget should be right there:

screen shot 2017-09-06 at 3 13 35 pm

app/src/org/commcare/adapters/MenuAdapter.java
amstone326: Ah cool makes perfect sense

app/src/org/commcare/adapters/MenuAdapter.java
amstone326: @shubham1g5 Let me know when you're done testing this; I'm good to merge once you have

shubham1g5:
So things got slightly bit more interesting for me after adding audio files to menu. In the production code currently scrolling through a menu list causes a NPE as we were removing the `audioPlaybackButton` from the parent view. With the earlier change in this PR of setting the visibility to`GONE` instead, there was no crash but an audio button in `Playing` state will go invisible if we scroll it off the screen and then scroll back to it. So the latest commit on this PR fixes that problem.

amstone326:
Hey @shubham1g5, thanks a lot for the thoroughness on that. The latest commit makes total sense to me, but can you explain a bit more as to why the previous behavior on master of removing the audio button from the parent view was causing an exception? I want to understand if this is something we should consider hotfixing (and also try to understand why we haven't gotten any bug reports from users).

shubham1g5:
@amstone326 So the crash only happens when we scroll to a menu item with audio which has been off screen to start with. I think we don't have a project on commcare where we have a long list of menus with audio uris set to them and thus it has not come in notice till now. 

 On why it's happening, when we remove the `audioPlaybackButton` from parent view and when this view gets recycled for some other position which has an audio set , [this line](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/adapters/MenuAdapter.java#L150) will return the `audioPlaybackButton` to null and cause a NPE [here](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/adapters/MenuAdapter.java#L177)

amstone326:
@shubham1g5 Got it! Good explanations all around. Merging

PR #1813, opened 2017-09-04
Bye Bye expanded audio widget

shubham1g5:
Release Note: Expanded audio widget functionality can no longer be used for audio playback


amstone326:
Hey, shouldn't `ExpandedAudioPlaybackView` have gotten deleted here as well?

shubham1g5:
@amstone326 we decided to do that as a second step and for now just disable the switch to the `ExpandedAudioPlaybackView`

PR #1814, opened 2017-09-05
NFC support

amstone326:
Support for writing to or reading from an NFC tag in CommCare in 2 ways:
- Capability to write to or read from an NFC tag within an app callout question type in a form
- Capability to use the value read from an NFC tag to search within a case list (parallel to searching the case list via barcode or fingerprint scanning)

Detailed documentation has been added here: https://confluence.dimagi.com/display/internal/Integrating+CommCare+with+NFC-enabled+devices


app/src/org/commcare/android/nfc/NfcActivity.java
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        try {
+            nfcManager.checkForNFCSupport();
+            if (requiredFieldsMissing()) {
+                return;
+            }
+            setReadyToHandleTag();
+        } catch (NfcManager.NfcNotEnabledException e) {
+            finishWithErrorToast("nfc.not.enabled");
```
wpride: In this case we could prompt them to turn on NFC and listen for them to activate it, no?

app/src/org/commcare/android/nfc/NfcReadActivity.java
```diff
@@ -0,0 +1,119 @@
...
+        Ndef ndefObject = Ndef.get(tag);
+        try {
+            ndefObject.connect();
+            NdefMessage msg = ndefObject.getNdefMessage();
+            if (msg == null) {
+                finishWithErrorToast("nfc.read.no.data");
+                return;
+            }
+            NdefRecord firstRecord = msg.getRecords()[0];
+            Pair<String, Boolean> resultAndSuccess =
```
wpride: Might be worth mentioning in a method comment for `readValueFromRecord` that the first result is either the successfully read value *or* the error message locale key

app/src/org/commcare/android/nfc/NfcReadActivity.java
```diff
@@ -0,0 +1,119 @@
...
+            } catch (IOException e) {
+                // nothing we can do
+            }
+        }
+    }
+
+    @Override
+    protected void setResultExtrasBundle(Intent i, boolean success) {
+        Bundle responses = new Bundle();
+        responses.putString("nfc_read_result", success ? "success" : "failure");
```
wpride: Is this used anywhere? Pattern seems slightly awkward

app/src/org/commcare/android/nfc/NfcReadActivity.java
```diff
@@ -0,0 +1,119 @@
...
+ * CommCare recognizes (as determined by the filters in NfcActivity::setReadyToHandleTag) will be
+ * consumed by this activity. If the tag is of the specified/expected type(s), then this activity
+ * will attempt to read the tag. If the read is successful, the value read will be stored as an
+ * extra in the activity's result intent under the key "odk_intent_data", where it can be accessed
+ * by other activities. Otherwise, an appropriate error toast will be shown.
+ *
+ * @author Aliza Stone
+ */
+@TargetApi(Build.VERSION_CODES.JELLY_BEAN)
+public class NfcReadActivity extends NfcActivity {
```
wpride: The inheritence structure with this and `NfcWriteActivity` is very elegant

app/src/org/commcare/android/nfc/NfcReadActivity.java
```diff
@@ -0,0 +1,119 @@
...
+ * When this activity is in the foreground, any NFC tag scanned by the device that is of a format
+ * CommCare recognizes (as determined by the filters in NfcActivity::setReadyToHandleTag) will be
+ * consumed by this activity. If the tag is of the specified/expected type(s), then this activity
+ * will attempt to read the tag. If the read is successful, the value read will be stored as an
+ * extra in the activity's result intent under the key "odk_intent_data", where it can be accessed
+ * by other activities. Otherwise, an appropriate error toast will be shown.
+ *
+ * @author Aliza Stone
+ */
+@TargetApi(Build.VERSION_CODES.JELLY_BEAN)
```
wpride: Can this annotation be moved into the super class?

app/src/org/commcare/android/nfc/NfcWriteActivity.java
```diff
@@ -0,0 +1,103 @@
...
+    protected void setResultExtrasBundle(Intent i, boolean success) {
+        Bundle responses = new Bundle();
+        responses.putString("nfc_write_result", success ? "success" : "failure");
+        i.putExtra(IntentCallout.INTENT_RESULT_EXTRAS_BUNDLE, responses);
+    }
+
+    @Override
+    protected void setResultValue(Intent i, boolean success) {
+        if (success) {
+            i.putExtra(IntentCallout.INTENT_RESULT_VALUE, "Wrote value: " + payloadToWrite);
```
wpride: Why are you appending `Wrote value: ` here? That seems dangerous/wrong (IE if I were parsing this intent response I would not expect that header)

app/src/org/commcare/android/nfc/NdefRecordUtil.java
```diff
@@ -0,0 +1,133 @@
...
+            return createWellKnownTypeRecord(userSpecifiedType, payloadToWrite);
+        } else {
+            return createExternalRecord(userSpecifiedType, userSpecifiedDomain, payloadToWrite);
+        }
+    }
+
+    protected static boolean isCommCareSupportedWellKnownType(String type) {
+        // For now, the only "well known type" we're supporting is NdefRecord.RTD_TEXT, which
+        // users should encode in their configuration by specifying type "text"
+        return "text".equals(type);
```
wpride: We've discussed how WTO might come back and ask for some encrypted/encoded format, but all of this `type` checking has me thinking YAGNI.

app/src/org/commcare/android/nfc/NdefRecordUtil.java
```diff
@@ -0,0 +1,133 @@
...
+                                                                 String domainForExternalTypes) {
+        String typeAsString = new String(record.getType(), UTF8_CHARSET);
+        if (recordTypeIsExpected(typeAsString, userSpecifiedTypes, domainForExternalTypes)) {
+            return new Pair<>(new String(record.getPayload(), UTF8_CHARSET), true);
+        } else {
+            return new Pair<>(READ_ERROR_MISMATCH_KEY, false);
+        }
+    }
+
+    private static boolean recordTypeIsExpected(String recordTypeAsString,
```
wpride: Unclear what this function is doing

app/src/org/commcare/android/nfc/NdefRecordUtil.java
```diff
@@ -0,0 +1,133 @@
...
+        int lengthOfPrefix = langBytesLength + 1; // add 1 for the status byte itself
+        byte[] payloadWithoutLang = Arrays.copyOfRange(fullPayload, lengthOfPrefix, fullPayload.length);
+        return new String(payloadWithoutLang, UTF8_CHARSET);
+    }
+
+    protected static NdefRecord createNdefRecord(String userSpecifiedType,
+                                                 String userSpecifiedDomain,
+                                                 String payloadToWrite) {
+        if (isCommCareSupportedWellKnownType(userSpecifiedType)) {
+            return createWellKnownTypeRecord(userSpecifiedType, payloadToWrite);
```
wpride: Can you explain (maybe comment) the effective difference between a "well known" and "external" type record?

app/src/org/commcare/android/nfc/NfcActivity.java
```diff
@@ -0,0 +1,150 @@
...
+    protected static final String NFC_DOMAIN_ARG = "domain";
+
+    protected NfcManager nfcManager;
+    protected PendingIntent pendingNfcIntent;
+    protected String domainForType;
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) {
```
wpride: else... return?

app/src/org/commcare/android/nfc/NfcActivity.java
```diff
@@ -0,0 +1,150 @@
...
+     * The intent filters being passed to enableForegroundDispatch() here are intentionally very
+     * broad, on the assumption that if CommCare's NfcActivity is in the foreground while a user
+     * tries to scan an NFC tag, they were intending to scan something that CommCare would
+     * recognize. So if the user scans a tag of a type that we aren't expecting, instead of
+     * filtering it out (which would result in the device's default NFC handler consuming the tag,
+     * resulting in confusing on-screen behavior for the user), we'll consume it ourselves and
+     * then show a useful error message.
+     */
+    private void setReadyToHandleTag() {
+        IntentFilter ndefDiscoveredFilter = new IntentFilter(NfcAdapter.ACTION_NDEF_DISCOVERED);
```
wpride: Is this not handled by the additions to the `AndroidManifest`? Or does this guarantee us higher priority?

app/src/org/commcare/android/nfc/NdefRecordUtil.java
amstone326: The confluence documentation that I linked to explains this. Specifically the "Choosing and Defining Payload Types" section, although reading the whole page would probably be useful context.

app/src/org/commcare/android/nfc/NdefRecordUtil.java
amstone326: Are you referring to not supporting the custom record types at all? If so, I would really rather not do that. I think there is a high probability that 1 of the next few projects that want to use this will require something like the custom types (and the work is already done to include it in a way that I feel does not add an unreasonable complexity burden). I also think there is a non-trivial chance that 2 months from now, World Vision will decide there is some reason they actually do need this.

app/src/org/commcare/android/nfc/NdefRecordUtil.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    protected static boolean isCommCareSupportedWellKnownType(String type) {
+        // For now, the only "well known type" we're supporting is NdefRecord.RTD_TEXT, which
+        // users should encode in their configuration by specifying type "text"
+        return "text".equals(type);
+    }
+
+    private static NdefRecord createWellKnownTypeRecord(String type, String payload) {
+        if (type.equals("text")) {
```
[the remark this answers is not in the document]
amstone326: Right now the implementation is the same, yes, but the intended meaning of the method (as evidenced by its name) is different. If we were to start supporting any other well-known type, that would get added to the body of `isCommCareSupportedWellKnownType()` as well. Whereas here, I need to very specifically be checking for 'text' type only. So I don't agree that I should switch to calling that method, if that was what you meant.

app/src/org/commcare/android/nfc/NdefRecordUtil.java
```diff
@@ -0,0 +1,133 @@
...
+            return createTextRecord(payload);
+        } else {
+            throw new IllegalArgumentException();
+        }
+    }
+
+    private static NdefRecord createTextRecord(String payload) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+            return NdefRecord.createTextRecord(null, payload);
```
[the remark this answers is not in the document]
amstone326: I don't know how to evaluate whether they are "identical" per say, but I have tested that both paths create an Ndef Record of type text that is readable by both CommCare and the Android device's default NFC processing system. So that seems sufficient to me. Android's NFC implementation guidelines strongly suggest using their built-in NdefRecord create methods where possible, and we also follow this sort of pattern quite often (where we follow 2 different paths depending on Android version so as to use the newer/better standard where possible).

app/src/org/commcare/android/nfc/NfcActivity.java
amstone326: I'm not sure what you're getting at; what would returning at the end of `onCreate` achieve? `onResume` takes care of finishing the activity if this condition isn't met, since finishing within `onCreate` is incorrect practice.

app/src/org/commcare/android/nfc/NfcActivity.java
```diff
@@ -0,0 +1,150 @@
...
+                    setText(Localization.get(getInstructionsTextKey()));
+        }
+    }
+
+    protected void initFields() {
+        this.nfcManager = new NfcManager(this);
+
+        this.domainForType = getIntent().getStringExtra(NFC_DOMAIN_ARG);
+        if (this.domainForType == null) {
+            // Usually there is no reason for the user to provide this themselves, and we will just
```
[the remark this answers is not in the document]
amstone326: Mm this may be a little bit YAGNI, but my thinking was that if any app ever needs to read a value that was originally written by an app from a different domain, then it would be important to have this override option.

app/src/org/commcare/android/nfc/NfcActivity.java
amstone326: Mm good call that I should add some instructions to activate it to the error message, but I think listening for activation goes beyond what is reasonably needed, and also wouldn't really fit into the control flow. Throwing this exception will finish the NFC activity, and then they can simply re-press the button that triggers the activity to open once they've gone into settings and activated it.

app/src/org/commcare/android/nfc/NfcActivity.java
amstone326: This is the setup Android outlines for making it so that an activity has highest priority for handling an NFC tag when, and _only_ when, that activity is in the foreground. So this way we only try to pick up NFC tags when the user has explicitly launched `NfcActivity` (either by clicking the Intent callout button in a callout question in a form, or by clicking the search callout button in a case list).

app/src/org/commcare/android/nfc/NfcReadActivity.java
amstone326: oo yes good call, this got missed out when I did some refactoring

amstone326: Actually, I take this back, this should get removed entirely because nothing is actually using/respecting it anymore. I'm now taking care of this via the version code checks in `onCreate` and `onResume`

app/src/org/commcare/android/nfc/NfcReadActivity.java
amstone326: Yes definitely

app/src/org/commcare/android/nfc/NfcWriteActivity.java
amstone326: Yea you're right, will remove.

app/src/org/commcare/android/nfc/NfcReadActivity.java
amstone326: Ok final update-- the correct thing here was to move these targets to the method in each class that actually does the work, so `readFromNfcTag()` and `writeMessageToNfcTag()`, since those are the ones that actually need to be protected from being called by lower versions (and are being protected by the check in `onResume`)

app/src/org/commcare/android/nfc/NfcReadActivity.java
amstone326: Mm I guess for the read it may be a little unnecessary since you would just know by whether or not you received a value. But it definitely serves a real purpose for the write actions, since there's otherwise no way to know in the model whether or not it worked. Do you think it would be better if I removed this for reads and only had it for writes?

amstone326:
@wpride Thanks for the review! Either responded to or pushed changes for all your feedback.

app/src/org/commcare/android/nfc/NfcReadActivity.java
wpride: I was thinking why not use a boolean?

amstone326: (offlined about this)

PR #1818, opened 2017-09-12
Sort based on edit-distance when fuzzy

wpride:
fixes https://manage.dimagi.com/default.asp?261687

Not sure if this was a bug or intentional behavior that I'm missing but this change makes more sense to me. Previously, if we weren't using `asyncMode` (which I think we mostly aren't?) then we'd never sort filtered entities based on the edit distance from the search field (IE `score` was never used). This changes the behavior to re-sort the entities whenever we're in fuzzy mode.

### DIFF
--- a/app/src/org/commcare/adapters/EntityStringFilterer.java
+++ b/app/src/org/commcare/adapters/EntityStringFilterer.java
@@ -121,7 +121,8 @@ public class EntityStringFilterer extends EntityFiltererBase {
                     matchScores.add(Pair.create(index, score));
                 }
             }
-            if (isAsyncMode) {
+            // If fuzzy search is enabled need to re-sort based on edit distance
+            if (isFuzzySearchEnabled) {
                 Collections.sort(matchScores, new Comparator<Pair<Integer, Integer>>() {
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/adapters/EntityStringFilterer.java
```diff
@@ -121,7 +121,8 @@ private void buildMatchList() {
                     matchScores.add(Pair.create(index, score));
                 }
             }
-            if (isAsyncMode) {
+            // If fuzzy search is enabled need to re-sort based on edit distance
+            if (isFuzzySearchEnabled) {
```
amstone326: @wpride Why would we only want to do this when fuzzy search is enabled? It looks like the `matchScores` object gets populated whether or not we're using fuzzy search, so wouldn't we just always want to sort by those scores?

wpride: Because if fuzzy search isn't enabled then all we have is the set of entries matching the search fields exactly, so I think in that case keeping the default sort is correct

PR #1819, opened 2017-09-12
Hack migration fix

wpride:
Hacky fix for https://manage.dimagi.com/default.asp?260824

Attempts to perform the missing `owner_id` migration once the user logs in. Meant to be updated to by the affected users so that they can submit forms, then immediately deleted.

Tested by:

1. Installing 2.35.1
2. Filling out form, saving as incomplete
3. Updating to this version and *commenting out the seventeen > eighteen migration*
4. Logging in and confirming that the migration runs
5. Submitting forms successfully 

DO NOT MERGE - just here as an FYI, going to send a custom APK to users

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -3,7 +3,7 @@
           xmlns:tools="http://schemas.android.com/tools"
           package="org.commcare.dalvik"
           android:versionCode="106"
-          android:versionName="2.37.2">
+          android:versionName="Custom Fix 2.37.2">
 
     <uses-permission android:name="android.permission.NFC"/>
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION"/>
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -3,7 +3,7 @@
           xmlns:tools="http://schemas.android.com/tools"
           package="org.commcare.dalvik"
           android:versionCode="106"
-          android:versionName="Custom Fix 2.37.2">
+          android:versionName="2.37.99">
 
     <uses-permission android:name="android.permission.NFC"/>
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION"/>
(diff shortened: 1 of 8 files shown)


app/AndroidManifest.xml
```diff
@@ -3,7 +3,7 @@
           xmlns:tools="http://schemas.android.com/tools"
           package="org.commcare.dalvik"
           android:versionCode="106"
-          android:versionName="2.37.1">
+          android:versionName="2.37.2">
```
ctsims: This version is (I believe) just a string with no semantic value, right? I think we should make it say "Custom Migration Fix 2.37.2"

app/src/org/commcare/CommCareApplication.java
```diff
@@ -1051,4 +1056,39 @@ public static CommCareNoficationManager notificationManager() {
         return app.noficationManager;
     }
 
+    public static boolean applyMissedMigration(SQLiteDatabase db, Context c) {
```
ctsims: You probably want this to happen on user input somewhere, not on login. Otherwise this is going to fire over and over and if they fail to send after the first time it'll crash and they'll be locked out again

wpride: @ctsims its wrapped in a try/catch so it'll fail silently

app/AndroidManifest.xml
```diff
@@ -3,7 +3,7 @@
           xmlns:tools="http://schemas.android.com/tools"
           package="org.commcare.dalvik"
           android:versionCode="106"
-          android:versionName="2.37.1">
```
shubham1g5: Seems like I am missing something obvious though how is it that diff here says versionName as 2.37.1 when it's 2.39 in master currently!

amstone326: ...good question

wpride: This isn't against master, its against 2.37.2 (which they're on). Retargeting for clarity

shubham1g5: Yeah. Not important. I think it's the conflicts though because of which github diff is getting confused and showing 2.37.1 as the removed version!

amstone326:
@wpride Did this end up working for the users in question? Slash can the PR be closed now?

wpride:
Haven't heard - that project was in Haiti and last I heard they were bracing for Irma 😞 closing

amstone326:
Sad :(

PR #1823, opened 2017-09-20
Quarantining forms instead of wiping on File not Found

shubham1g5:
Workaround for this - https://manage.dimagi.com/default.asp?261911#1394223

Right now when in recovery mode if for one of the form we could not find any instances, it blocks the whole sending forms task and user gets stuck with the blocking recovery mode screen. Hence in such a case we are now quarantining that form so that forms upload can proceed without sending the buggy form and user can get out of recovery mode.


shubham1g5:

PR #1824, opened 2017-09-20
Quarantining forms instead of wiping on File not Found

shubham1g5:
Workaround for this - https://manage.dimagi.com/default.asp?261911#1394223

Right now when in recovery mode if for one of the form we could not find any instances, it blocks the whole sending forms task and user gets stuck with the blocking recovery mode screen. Hence in such a case we are now quarantining that form so that forms upload can proceed without sending the buggy form and user can get out of recovery mode.

### DIFF
--- a/app/src/org/commcare/tasks/ProcessAndSendTask.java
+++ b/app/src/org/commcare/tasks/ProcessAndSendTask.java
@@ -296,7 +296,17 @@ public abstract class ProcessAndSendTask<R> extends CommCareTask<FormRecord, Lon
                     //Good!
                     //Time to Send!
                     try {
-                        folder = new File(record.getPath(c)).getCanonicalFile().getParentFile();
+                        try {
+                            folder = new File(record.getPath(c)).getCanonicalFile().getParentFile();
+                        } catch (FileNotFoundException e) {
+                            //This will put us in the same "Missing Form" handling path as below
+                            throw e;
+                        } catch (IOException e) {
+                            // Unexpected/Unknown IO Error path from cannonical file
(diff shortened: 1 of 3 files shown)


shubham1g5:

app/src/org/commcare/tasks/ProcessAndSendTask.java
```diff
@@ -347,6 +343,10 @@ private void sendForms(FormRecord[] records) throws TaskCancelledException {
                             break;
                         }
                         continue;
+                    } catch (IOException e) {
```
ctsims: Hey,

I don't think this is quite safe enough as a processing path to be comfortable. There's a lot of IO that gets handled in this method, and it would be very easy for someone to introduce and I/O issue between lines 298 and 356 that they don't realize is actually unhandled. 

I'm gonna change this around just a bit and propose a different pattern

app/src/org/commcare/tasks/ProcessAndSendTask.java
```diff
@@ -292,16 +292,21 @@ private void sendForms(FormRecord[] records) throws TaskCancelledException {
...
                     //Good!
                     //Time to Send!
                     try {
+                        try {
+                            folder = new File(record.getPath(c)).getCanonicalFile().getParentFile();
+                        } catch (FileNotFoundException e) {
+                            //This will put us in the same "Missing Form" handling path as below
+                            throw e;
+                        } catch (IOException e) {
+                            // Unexpected/Unknown IO Error path from cannonical file
```
shubham1g5: Makes sense. Thanks for the correction 👍

shubham1g5:
@ctsims Added display message for number of quarantine forms, though think it's a bit confusing since it occurs as a text for 'Send Data' button and quarantine forms never actually get sent.

ctsims:
I think it's fine that they never get sent, better to know than not know.

PR #1826, opened 2017-09-20
Fix for Badge Processing Erroneously loading Instances

ctsims:
The badge text processor wasn't processing what instances were needed for displaying menus, which was causing delays in loading menus for un-needed data sources.

cross-request: https://github.com/dimagi/commcare-core/pull/644


ctsims:

PR #1828, opened 2017-09-21
Cleanups for Asyncronous setup in badge processing

ctsims:
Most of the actual fix is in commcare-core, but I cleaned up the processing a bit to unify where disposables are tracked.

One specific fix here is that I moved the execution of the badge lambdas into `Schedulers.single()`, since they end up unrolling a ton of state and it's _really_ easy for them to OOM the VM. At some point we may want to revisit whether we can set them up to process concurrently, but it'll take some thinking on how to make sure we're managing the memory pool.

Currently there are still some ~shaky semantics about the scheduler that I'm a bit concerned by (I'm pretty worried about leaving the Observables around and executing until onDestroy if, say, onStop has been fired, but our scope on using the new execution pattern is pretty limited.

cross-request: https://github.com/dimagi/commcare-core/pull/645

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -46,6 +46,7 @@ dependencies {
     compile(name: 'htmlspanner-custom', ext: 'aar')
     compile(name: 'android-zebra-interface-1.1', ext: 'aar')
     compile 'com.simprints:LibSimprints:1.0.10'
+    compile 'com.android.support:multidex:1.0.1'
     compile 'com.android.support:cardview-v7:25.0.1'
     compile 'com.android.support:recyclerview-v7:25.0.1'
     compile 'com.android.support:support-v4:25.0.1'
--- a/app/src/org/commcare/activities/CommCareActivity.java
+++ b/app/src/org/commcare/activities/CommCareActivity.java
@@ -57,6 +57,10 @@ import org.javarosa.core.services.Logger;
 import org.javarosa.core.services.locale.Localization;
 import org.javarosa.core.util.NoLocalizedTextException;
 
+import io.reactivex.Observable;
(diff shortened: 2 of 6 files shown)


shubham1g5:
Looks good to me. Feel free to merge once you approve of additional changes in core. 

Also thanks for the cleanup. I was worried if we will use the Rx pattern elsewhere and hence was weary of adding the Disposables functions to `CommCareActivity`.

`Schedulers.single()` does make sense for this. Though we can also restrict the thread pool size in case we do want the badges to load faster and still consume low memory. 

I don't see any specific problems around disposing Disposables in onDestory and it seems to be a standard practice. Though will think about it a bit more to see if there is a better way we can do this.

ctsims:
The only problem that jumps out to me with onDestroy is that it's pretty hard to decide whether the processing should keep going until then.

In the AsyncTaskPattern we use in CommCareActivity one thing we do is basically to "unsusbscribe" the listener from the activity after onPause() because it lets us reconnect the task to either the original activity or a new activity (if, say, rotation happens), which lets us start an async action in one activity and then hand it off to the next. It also means that if the task response is stateful and dangerous to accept the result of it if you are in an a paused state (IE: triggering view changes might not work).

I _hate_ async tasks and would be super open to aligning the pattern in CommCareActivity around this toolkit instead, long term. I did some reading to better understand reactivex and I really appreciate how much lighter weight it is. 

I think the best way to think about things moving forward would be to use the CommCareTask pattern for actions which have long-term stateful consequences (like submitting forms) where you care a lot about reconnecting the task, and the Disposable Observerable pattern for async actions which are straightforward and it's ok if the result gets dropped. Lowering the barrier to making small actions asynchronous would be a huge win for our UX

PR #1833, opened 2017-09-26
Check for pin length when keyboard action buttons are used

amstone326:
Fix for https://manage.dimagi.com/default.asp?262413

### DIFF
--- a/app/src/org/commcare/activities/CreatePinActivity.java
+++ b/app/src/org/commcare/activities/CreatePinActivity.java
@@ -103,11 +103,11 @@ public class CreatePinActivity extends SessionAwareCommCareActivity<CreatePinAct
             public boolean onEditorAction(TextView v, int actionId, KeyEvent event) {
                 // processes the done/next keyboard action
                 if (actionId == EditorInfo.IME_ACTION_DONE || actionId == EditorInfo.IME_ACTION_NEXT) {
-                    if (!pinLengthIsValid(enterPinBox.getText())) {
+                    if (pinLengthIsValid(enterPinBox.getText())) {
+                        continueButton.performClick();
+                    } else {
                         Toast.makeText(CreatePinActivity.this, Localization.get("pin.length.error"),
                                 Toast.LENGTH_LONG).show();
-                    } else {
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/activities/CreatePinActivity.java
```diff
@@ -103,7 +103,12 @@ private void setListeners() {
             public boolean onEditorAction(TextView v, int actionId, KeyEvent event) {
                 // processes the done/next keyboard action
                 if (actionId == EditorInfo.IME_ACTION_DONE || actionId == EditorInfo.IME_ACTION_NEXT) {
-                    continueButton.performClick();
+                    if (enterPinBox.getText().toString().length() != 4) {
```
ctsims: @amstone326 I think maybe I'm just missing something, but it seems confusing to me that the validation logic would be replicated here, v. unified in whatever happens when continueButton.performClick() is called. Is there a reason that pattern is hard here?

amstone326: @ctsims The reason this is implemented separately here is because for `continueButton` (i.e. the button I created on-screen, rather than the built-in keyboard action), I handled this by making that button completely disabled until the input length reaches 4 (see the implementation of `afterTextChanged` in `CreatepinActivity.getPinTextWatcher()`). This pattern is not an option for the keyboard action buttons because there is no concept of disabling them; the only method of intervention available is to override `onEditorAction` and potentially intercept what happens when it's pressed. This is also why I missed preventing this in the original implementation of this feature (this isn't a regression; it's always been this way). I could have instead placed this new code in the `onClick` methods for `continueButton` and `enterButton`, respectively, but I see no advantage to doing it that way.

ctsims: aha! That certainly clarifies things. Thanks for explaining, it was a bit confusing to tell at a glance from the code.

I think it would be useful to separate the logic such that the button press being possible and the enter button being pressed follow the same code path for validation (so that if a future constraint was added or "4" was changed it wouldn't need to be changed in both places, but Approved to pull with or without that change.

PR #1834, opened 2017-09-27
Submitting logs on hitting Recovery Mode

shubham1g5:
Triggers logs submission on Entering Recovery to debug corrupt storage error issues.

Case: https://manage.dimagi.com/default.asp?262385

Targeting 2.39 since there are increasing reports for corrupt-storage issue and users can't do log submission right now after encountering corrupt storage dialogue and it's pretty dark without logs when it comes to corrupt storage bugs. 

Though let me know if this is too late for 2.39 and need to target master instead.

### DIFF
--- a/app/src/org/commcare/CommCareApp.java
+++ b/app/src/org/commcare/CommCareApp.java
@@ -233,7 +233,7 @@ public class CommCareApp implements AppFilePathBuilder {
         }
 
         String failureReason = profile == null ? "profle being null" : "profile status value " + String.valueOf(profile.getStatus());
-        Logger.log(LogTypes.TYPE_ERROR_STORAGE, "Initializing application failed because of " + failureReason);
+        Logger.log(LogTypes.TYPE_RESOURCES, "Initializing application failed because of " + failureReason);
         return false;
     }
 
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/CommCareApp.java
```diff
@@ -231,6 +231,9 @@ private boolean initializeApplicationHelper() {
             }
             return true;
         }
+
+        String failureReason = profile == null ? "profle being null" : "profile status value " + String.valueOf(profile.getStatus());
+        Logger.log(LogTypes.TYPE_ERROR_STORAGE, "Initializing application failed because of " + failureReason);
```
amstone326: I think it makes more sense to make this `TYPE_RESOURCES`

app/src/org/commcare/activities/RecoveryActivity.java
```diff
@@ -48,7 +49,7 @@
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-
+        CommCareUtil.triggerLogSubmission(this);
```
amstone326: This should be within the `if` statement immediately below, so that we only do it once

PR #1837, opened 2017-09-28
Only trigger sync after update for release builds

amstone326:
This behavior is annoying and pointless on debug builds, so limit it to release builds only


amstone326:

PR #1838, opened 2017-09-28
Killable Queries 

ctsims:
Android requirements for the cross-request

cross-request: https://github.com/dimagi/commcare-core/pull/654

### DIFF
--- a/app/src/org/commcare/models/database/SqlStorage.java
+++ b/app/src/org/commcare/models/database/SqlStorage.java
@@ -14,6 +14,8 @@ import org.commcare.modern.models.EncryptedModel;
 import org.commcare.modern.util.Pair;
 import org.commcare.util.LogTypes;
 import org.commcare.utils.SessionUnavailableException;
+import org.javarosa.core.model.condition.Abandonable;
+import org.javarosa.core.model.condition.LifecycleSignaler;
 import org.javarosa.core.services.Logger;
 import org.javarosa.core.services.storage.EntityFilter;
 import org.javarosa.core.services.storage.IStorageIterator;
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/models/database/SqlStorage.java
```diff
@@ -626,6 +633,7 @@ public void bulkRead(LinkedHashSet cuedCases, HashMap recordMap) {
                     c.moveToFirst();
                     int index = c.getColumnIndexOrThrow(DatabaseHelper.DATA_COL);
                     while (!c.isAfterLast()) {
+                        LifecycleSignaler.AssertNotAbandoned(abandonable);
```
amstone326: So just making sure I'm understanding-- this is the line that would stop the background thread for the badge calculations if its context has been marked as abandoned? If so, I'm definitely missing something. In the case that this throws an exception.. won't the program just crash? There's nothing in the code that catches/handles a `RequestAbandonedException`

ctsims: The new async layer that we are using handles errors for "Disposed" threads gracefully.

For right now this system basically isn't allowed to _guarantee_ that the exception that is thrown from this is a RequestAbandonedException, since the request abandon may end up clearing out memory which results in another error. Part of the semantics of an abandoned query are that once it's abandoned, it can throw any exceptions, since it's hard to know how it will fail.

amstone326: @ctsims But if this were called by anything other than the async framework that's handling the badge calculation, it would need real error handling, right?

amstone326: The reason I ask is because if so, I think this method declaration in `IStorageUtilityIndexed` should warn in some way that exceptions should be expected

ctsims: For sure. I'll update the docs to reflect that.

PR #1839, opened 2017-09-28
Fix indexed fixture paths table

amstone326:
Fixes a bug where the indexed fixture paths table would contain duplicate rows for the same fixture if CommCare processed a transaction containing the same fixture multiple times

cross-request: https://github.com/dimagi/commcare-core/pull/655


app/unit-tests/src/org/commcare/android/tests/database/SqlStorageBaseTests.java
```diff
@@ -29,4 +24,5 @@
         TestUtils.initializeStaticTestStorage();
         return TestUtils.getStorage("ShoeStorage", Shoe.class);
     }
+
```
wpride: ?

PR #1842, opened 2017-10-02
Fixes NPE for Grid Menu when there is audioFile for a menu

shubham1g5:
In 2.39, while refactoring for Menu Badges, I also refactored MenuAdapter and GridAdapter to use same getView code and what I didn't realise is there is not audio button in grid menu list item layout. This is causing a crash when we have anyway added a audio file for a Grid menu in HQ since right now mobile code is assuming that if there is a audio file there should also be a audioButton for it. 

https://fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/59cf74adbe077a4dcc831241?time=last-ninety-days

This is impacting only the ICDS AWW app right now and it's causing a crash on clicking Start.

### DIFF
--- a/app/src/org/commcare/adapters/MenuAdapter.java
+++ b/app/src/org/commcare/adapters/MenuAdapter.java
@@ -163,24 +163,24 @@ public class MenuAdapter extends BaseAdapter {
     }
 
     private void setupAudioButton(int rowId, AudioPlaybackButton audioPlaybackButton, MenuDisplayable menuDisplayable) {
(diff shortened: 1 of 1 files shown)


shubham1g5:
This is going to have mass impact and since it's a one liner, I am self merging it.

shubham1g5:
Actually just realised that ICDS is on LTS , so I am not sure why some users are on 2.39. Nevertheless since most users are on LTS, impact for this should be limited and I am now waiting on the merge.

PR #1845, opened 2017-10-03
2.39.2 - Universal timeout increase on upgrades

amstone326:
Fixes https://fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/59bab3f7be077a4dcca0087b?time=last-seven-days.

The stack trace in the above crash doesn't reveal what was really going on here (thanks @ctsims for getting to the bottom of it). The bug was that for the v20 --> v21 user db migration (which reindexes all cases), the migration was taking longer than 5 seconds, which is the default timeout set in `CommCareApplication` (`MAX_BIND_TIMEOUT`) for binding to the session service. Previously, we would selectively increase this timeout in certain migrations that we expected to take a long time, but this one was missed and is causing issues for a few users (thankfully only 3 so far, all in the same app, so seems quite rare). But since there's really not much harm in increasing this timeout even when it's not strictly necessary, we've decided to simplify matters and prevent this in the future by having a blanket increase in the timeout for all user db upgrades.

### DIFF
--- a/app/src/org/commcare/models/database/migration/FixtureSerializationMigration.java
+++ b/app/src/org/commcare/models/database/migration/FixtureSerializationMigration.java
@@ -48,9 +48,6 @@ public class FixtureSerializationMigration {
     public static boolean migrateFixtureDbBytes(SQLiteDatabase db, Context c,
                                                 String directoryName,
                                                 byte[] fileMigrationKeySeed) {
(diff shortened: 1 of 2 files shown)


ctsims:

amstone326:
Release Note: Fixes a crash that could occur during some database migrations if upgrading by several CommCare versions

PR #1847, opened 2017-10-03
Wi-Fi Direct Bug Fix

wpride:
A line of the Wi-Fi Direct copy file code could result in the copy going through as successful (and the sender deleting their forms) but also triggering the receiver to delete the received payload (even though it would be, in fact, in tact).

This fix changes that by:

Catching the socket close IOException (which should really be ignored) outside the normal catch
Not deleting the forms in the event of an error (because the folder name is time-based we don't really need to worry about cleaning these up)

### DIFF
--- a/app/src/org/commcare/fragments/FileServerFragment.java
+++ b/app/src/org/commcare/fragments/FileServerFragment.java
@@ -72,9 +72,7 @@ public class FileServerFragment extends Fragment {
         Logger.log(TAG, "File Server starting...");
 
         mStatusText.setText("Starting server");
-
         mView.setVisibility(View.VISIBLE);
-
         if (mFileServer != null) {
             mFileServer.cancel(true);
         }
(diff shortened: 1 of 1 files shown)


wpride:

ctsims:
@wpride is this good to go? It doesn't have any tags applied to it

wpride:
@ctsims thanks, added

amstone326:
@wpride @ctsims Do you guys think this is definitely safe to go into a hotfix? I just don't have any context on this feature so don't know how to judge.

ctsims:
I think this is safe to hotfix. @amstone326 is 2.39.2 still open? Can we work this in there?

amstone326:
@ctsims yes it is, that's why this is targeting the 2.39 branch. If this looks good to you and you can merge it now that would be great. I'd like to do the 2.39.2 release asap today

ctsims:
@amstone326 Done and g2g

amstone326:
Release note: Fixes a bug that could cause a crash during a WiFi Direct transfer to result in data loss

PR #1850, opened 2017-10-05
2.36.3 - Set up ability to require PIN for dev options access

amstone326:
Addresses concerns raised in https://manage.dimagi.com/default.asp?262779

### DIFF
--- a/app/res/xml/preferences_developer_pin.xml
+++ b/app/res/xml/preferences_developer_pin.xml
@@ -1,7 +1,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<PreferenceScreen xmlns:android="http://schemas.android.com/apk/res/android">
-    <EditTextPreference
-        android:dialogTitle="Enter Access Pin"
-        android:key="cc-dev-prefs-user-entered-pin"
-        android:title="Developer Options Access Pin"/>
-</PreferenceScreen>
--- a/app/res/xml/preferences_developer_pin.xml
+++ b/app/res/xml/preferences_developer_pin.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<PreferenceScreen xmlns:android="http://schemas.android.com/apk/res/android">
+    <EditTextPreference
+        android:dialogTitle="Enter Access Pin"
+        android:key="cc-dev-prefs-user-entered-pin"
+        android:title="Developer Options Access Pin"/>
+</PreferenceScreen>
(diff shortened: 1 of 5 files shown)


app/src/org/commcare/preferences/DeveloperPreferences.java
```diff
@@ -94,14 +97,31 @@ protected void onCreate(Bundle savedInstanceState) {
...
+
+        if (userAccessPinNeeded()) {
+            addPreferencesFromResource(R.xml.preferences_developer_pin);
+        } else {
+            initAllPrefs();
+        }
+    }
+
+    private static boolean userAccessPinNeeded() {
+        SharedPreferences prefs = CommCareApplication.instance().getCurrentApp().getAppPreferences();
```
ctsims: @amstone326 We should set this to not require a pin if the user is a SuperAdmin (or whatever the barcode scan/APK_IN_DEBUG check is) as well, so we have a way to access developer options without needing project specific data.

app/res/xml/preferences_developer_pin.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<PreferenceScreen xmlns:android="http://schemas.android.com/apk/res/android">
+    <EditTextPreference
+        android:dialogTitle="Enter Access Pin"
```
ctsims: My guess is that Saket may suggest that this PIN use PW bubbles (since admins enter it in front of users). No strong preference on whether we should oblige on that,

app/src/org/commcare/preferences/DeveloperPreferences.java
```diff
@@ -215,9 +235,33 @@ public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, Strin
...
     }
 
+    private static boolean userEnteredAccessPinMatches() {
+        SharedPreferences prefs = CommCareApplication.instance().getCurrentApp().getAppPreferences();
+        String projectAccessPin = prefs.getString(PROJECT_SET_ACCESS_PIN, null);
+        String userAccessPin = prefs.getString(USER_ENTERED_ACCESS_PIN, null);
+        return userAccessPin == null ? false : userAccessPin.equals(projectAccessPin);
+    }
+
+    private static void clearUserEnteredAccessPin() {
```
ctsims: We should fire this event when the user de-activates developer mode as well, I think. Otherwise the PIN will still be set and will allow the keystroke pattern to re-active developer mode.

amstone326: And by "de-activate developer mode", I'm assuming you mean editing the first setting available in Developer Preferences? Otherwise I don't know what other mechanism there is for this. But I never really imagined anyone actually using that setting to turn _off_ access.

ctsims: You'd want to disable to return the device to "production" settings.

It's definitely part of the workflow if you used something from developer settings in the field.

 * Admin is in the field with the phone and user
 * Turns on Developer mode to access / change something
 * makes the change
 * Turns off developer mode afterward so they can give the device back without granting access to those features.

amstone326:
ok @ctsims, should be ready

PR #1853, opened 2017-10-05
Prevent NPE when EntitySelectActivity has no command

amstone326:
I believe this should fix https://fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/59caaf16be077a4dcc4b044e?time=last-seven-days, though I would like someone else to think through if they agree with my logic.

The NPE is clearly happening because `entitySelectSearchUI` doesn't get initialized if `session.getCommand()` is null. What's unclear to me is the exact right way to handle it. I did some digging into the history of `EntitySelectActivity` and the comments surrounding what to do when there's no session command. My conclusion from that investigation is that the original intent here was for us to finish in onResume when in this state, but that wasn't actually happening. So I've added that, in addition to protecting `onCreateOptionsMenu` from doing anything if we're about to finish, because I _think_ there are instances in which onCreateOptionsMenu can get called before onResume.

### DIFF
--- a/app/src/org/commcare/activities/EntitySelectActivity.java
+++ b/app/src/org/commcare/activities/EntitySelectActivity.java
@@ -349,6 +349,10 @@ public class EntitySelectActivity extends SaveSessionCommCareActivity
 
     @Override
     protected void onResumeSessionSafe() {
+        if (session.getCommand() == null) {
+            setResult(RESULT_CANCELED);
+            this.finish();
+        }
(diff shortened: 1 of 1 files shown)


amstone326:

amstone326:

amstone326:

amstone326:

amstone326:

amstone326:
@wpride I'm closing this because I think I figured out what's going on and it's unrelated

PR #1855, opened 2017-10-12
Use Standard Image inflate for Callout Menu Icons 

ctsims:
Super small change to use the normal dynamic image inflation strategy for callout icons


amstone326:
@ctsims 2 questions: 
1. Did something prompt this change?
2. Did you test this with a couple of live apps to make sure the bounding dimensions don't cause anything funny to happen?

ctsims:
1. I was getting funny behavior with Icon inflation as a result of the fact that the icons have limited layout control properties (you can't set their scale behavior in the imageview since it's in a menu). 

2) I did. Made this to fix some inconsistencies in how it was inflating an image for an NFC demo I made for myself

PR #1856, opened 2017-10-12
disable analytics on LTS

amstone326:


app/src/org/commcare/google/services/analytics/GoogleAnalyticsUtils.java
```diff
@@ -400,7 +400,7 @@ private static Tracker getTracker() {
     }
 
     private static boolean analyticsDisabled() {
-        return !CommCarePreferences.isAnalyticsEnabled();
+        return true;
```
ctsims: hm, just so this doesn't come back and bite us with a bad backmerge later: is there a way to just tag this by not using the core package id?

PR #1858, opened 2017-10-12
Async entity list updates

shubham1g5:
- Wipe cache on ccz update or firstlogin after update and 412's
- Introduces a blocking async mode for case when all sort fields are not marked as cache and index, when in blocking mode we wait for showing the entity list untill we prime and caches the data i.e. instead of priming the thread on another thread we do it on same thread as `EntityLoaderTask`. This allows us to still sort the entity list by params not marked as cache and index. 

Wiki: https://confluence.dimagi.com/pages/viewpage.action?pageId=41484944

cross-request: https://github.com/dimagi/commcare-core/pull/666

### DIFF
--- a/app/src/org/commcare/CommCareApplication.java
+++ b/app/src/org/commcare/CommCareApplication.java
@@ -3,7 +3,6 @@ package org.commcare;
 import android.Manifest;
 import android.annotation.SuppressLint;
 import android.app.ActivityManager;
-import android.app.Application;
 import android.content.ComponentName;
 import android.content.Context;
 import android.content.Intent;
@@ -36,6 +35,7 @@ import org.commcare.android.database.global.models.ApplicationRecord;
 import org.commcare.android.javarosa.AndroidLogEntry;
 import org.commcare.android.logging.ForceCloseLogEntry;
 import org.commcare.android.logging.ForceCloseLogger;
+import org.commcare.android.logging.ReportingUtils;
 import org.commcare.core.interfaces.HttpResponseProcessor;
 import org.commcare.core.network.CommCareNetworkServiceGenerator;
 import org.commcare.core.network.HTTPMethod;
@@ -64,6 +64,7 @@ import org.commcare.models.database.SqlStorage;
 import org.commcare.models.database.app.DatabaseAppOpenHelper;
 import org.commcare.models.database.global.DatabaseGlobalOpenHelper;
 import org.commcare.models.database.user.DatabaseUserOpenHelper;
+import org.commcare.models.database.user.models.EntityStorageCache;
 import org.commcare.models.legacy.LegacyInstallUtils;
 import org.commcare.modern.database.Table;
 import org.commcare.modern.util.Pair;
@@ -728,6 +729,10 @@ public class CommCareApplication extends MultiDexApplication {
                             HybridFileBackedSqlHelpers.removeOrphanedFiles(mBoundService.getUserDbHandle());
 
                             PurgeStaleArchivedFormsTask.launchPurgeTask();
+                        }
+
+                        if (EntityStorageCache.getEntityCacheWipedPref(user.getUsername()) < ReportingUtils.getAppVersion()) {
+                            EntityStorageCache.tryWipeCache();
                         }
                     }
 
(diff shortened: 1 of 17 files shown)


shubham1g5:

shubham1g5:

app/src/org/commcare/CommCareApplication.java
```diff
@@ -729,6 +730,10 @@ public void onServiceConnected(ComponentName className, IBinder service) {
 
                             PurgeStaleArchivedFormsTask.launchPurgeTask();
                         }
+
+                        if (EntityStorageCache.getEntityCacheWipedPref(user.getUsername()) < ReportingUtils.getAppVersion()) {
```
ctsims: Not blocking, but technically I believe we should be using the GUID/internal user ID for this. 

I **think** we also have to require the app ID as part of this key, because technically user storage isn't app isolated right now.

app/src/org/commcare/activities/FormAndDataSyncer.java
```diff
@@ -163,7 +163,6 @@ public void syncDataForLoggedInUser(final SyncCapableCommCareActivity activity,
     }
 
     public void performOtaRestore(LoginActivity context, String username, String password) {
-        SharedPreferences prefs = CommCareApplication.instance().getCurrentApp().getAppPreferences();
```
ctsims: Thanks for doing this cleanup along the way!

app/src/org/commcare/interfaces/AndroidSortableEntityAdapter.java
```diff
@@ -17,7 +17,8 @@
 public abstract class AndroidSortableEntityAdapter extends SortableEntityAdapter {
 
     public AndroidSortableEntityAdapter(List<Entity<TreeReference>> entityList, Detail detail, NodeEntityFactory factory) {
-        super(entityList, detail, factory instanceof AsyncNodeEntityFactory);
+        super(entityList, detail,
+                !(factory instanceof AsyncNodeEntityFactory) || ((AsyncNodeEntityFactory)factory).isBlockingAsyncMode());
```
ctsims: Would it be worth shifting the "isBlockingAsyncMode()" to the normal node entity factory and just having that return false?

app/src/org/commcare/models/database/user/models/EntityStorageCache.java
```diff
@@ -130,4 +132,32 @@ public static int getSortFieldIdFromCacheKey(String detailId, String cacheKey) {
             return -1;
         }
     }
+
+    public static void tryWipeCache() {
```
ctsims: Is the expected behavior of this function either that it wipes the cache immediately *or* that it triggers a setting that will required the cache to wipe at a later time?

Right now i'm not comfortable with the guaranteed execution of that process given the Session exception path here. Let's chat over voice on that.

app/src/org/commcare/services/CommCareSessionService.java
```diff
@@ -438,6 +438,10 @@ public void closeServiceResources() {
         }
     }
 
+    public void invalidateUserDbHandle() {
```
ctsims: This is the result of a 412, right? I think we need to document exactly what state this creates for the current session and articulate what expectations should hold after it is executed.

app/src/org/commcare/interfaces/AndroidSortableEntityAdapter.java
shubham1g5: It's kind of contradicting to have a `isBlockingAsyncMode()` for a non async factory and I think will just pollute the Node Entity factory contract

app/src/org/commcare/tasks/InstallStagedUpdateTask.java
```diff
@@ -5,6 +5,7 @@
 import org.commcare.engine.resource.AndroidResourceManager;
 import org.commcare.engine.resource.AppInstallStatus;
 import org.commcare.engine.resource.ResourceInstallUtils;
+import org.commcare.models.database.user.models.EntityStorageCache;
```
ctsims: The changes to this file look like no-ops at this point, can we clean those out just for clarity?

PR #1860, opened 2017-10-13
Handle possibility of onCreateOptionsMenu getting called before onCreate finishes

amstone326:
Apparently if the action bar is being used in an activity, it is possible for `onCreateOptionsMenu` to get called before `onCreate` has finished executing (see https://stackoverflow.com/questions/7705927/android-when-is-oncreateoptionsmenu-called-during-activity-lifecycle/7706597#7706597). Although this definitely doesn't happen on all phones, it's the best explanation I can find for [this crash](https://fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/59caaf16be077a4dcc4b044e/sessions/59E0B141028000014ABF51F37354211A_f3dbf274b01111e7919a56847afe9799_0_v2?) (after determining that the cause I was proposing in https://github.com/dimagi/commcare-android/pull/1853 doesn't seem possible).

### DIFF
--- a/app/src/org/commcare/activities/EntitySelectActivity.java
+++ b/app/src/org/commcare/activities/EntitySelectActivity.java
@@ -189,6 +189,13 @@ public class EntitySelectActivity extends SaveSessionCommCareActivity
 
             boolean isOrientationChange = savedInstanceState != null;
             setupUI(isOrientationChange);
+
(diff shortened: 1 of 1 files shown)


wpride:
Wow, nice catch

PR #1863, opened 2017-10-25
Implementation of in-form http callouts for the android platform

ctsims:
Implements in-form http callouts for android.

cross-request: https://github.com/dimagi/commcare-core/pull/667


wpride:
@ctsims looks like the signature of the `CommCareTestApp` method are now incorrect

ctsims:
@wpride thanks! fixed.

PR #1865, opened 2017-10-27
Prevent null form controller after rotate during form load

amstone326:
I _think_ this is the fix for: https://fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/59c6c7cabe077a4dcc21be3b?time=last-seven-days.

Logically, it makes perfect sense that this is what would have been causing the issue; if you rotated while the form was still loading, then `hasFormLoadBeenTriggered` would have been true in the call to `onCreate()`, causing us to call `uiController.refreshView()`. But `mFormController` doesn't get initialized until `handleFormLoadCompletion()`, hence it would be null. The only problem with this theory is that I couldn't actually reproduce it when I tried opening a form that took a while to load, and rotated it a couple of times before loading finished. Somewhat inexplicably to me, it seemed that the call to `onCreate` that would be triggered by the rotation was queued up in some way by the OS, and not called until `FormLoaderTask` finished (this I surmised from using the debugger). Does this make any sense to anyone? Given how odd this behavior does seem, combined with how much sense this hypothesis makes from a codepath perspective, I'm guessing that the behavior on rotation when a task is in progress is inconsistent between devices in some way, and still think this is likely to be the correct bug and the correct fix. But let me know if others know more that might discredit this.

### DIFF
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -116,6 +116,7 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
     // rotation (or similar)
     private static final String KEY_FORM_LOAD_HAS_TRIGGERED = "newform";
     private static final String KEY_FORM_LOAD_FAILED = "form-failed";
+    private static final String KEY_FORM_LOAD_COMPLETED = "form-load-completed";
     private static final String KEY_LOC_ERROR = "location-not-enabled";
     private static final String KEY_LOC_ERROR_PATH = "location-based-xpath-error";
     private static final String KEY_IS_READ_ONLY = "instance-is-read-only";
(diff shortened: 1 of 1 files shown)


shubham1g5:
Might be unrelated but this seems like a problem @ctsims and me encountered while working on improving badge calculations where 2 threads were using the same lock causing them to wait for each other. I think it would be worth to see which phase of activity lifecycle waits for `FormLoaderTask` completion (for eg - onDestory vs onCreate) and see if `FormLoaderTask'  is locking a resource that activity is waiting on.

amstone326:
@shubham1g5 interesting thought. what were you and Clayton using to examine the threads/locks? I'm not familiar with how to do that.

shubham1g5:
@amstone326 I asked Clayton the same thing earlier and he was just using the regular AS debugger and monitoring threads to see if main thread is waiting on something while the async thread is executing. It can get hard to figure out what exactly is happening in this case though figuring out in which phase of activity lifecycle is the main thread getting stuck in waiting should narrow the problem down (you can add breakpoints in different lifecycle methods to see which lifecycle methods gets called and gets completed before completion of the async task)

amstone326:
@shubham1g5 @wpride Closing this because Clayton helped me determine that this is definitely not the cause. The flaw in my logic was that the `if` clause [here](https://github.com/dimagi/commcare-android/blob/f2a53ef8371e28287524cece30f1e1938f7f0c0b/app/src/org/commcare/activities/FormEntryActivity.java#L190-L192) prevents that `else if` clause from ever being called in the situation that I thought it could be. Opening a new PR for the actual cause, which we were also able to find.

PR #1867, opened 2017-10-27
onCreate --> onCreateSessionSafe

amstone326:
Should fix https://fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/59c3fc75be077a4dcc02c8f0?time=last-seven-days. `FormRecordListActivity` was implementing `SessionAwareCommCareActivity`, but not actually utilizing `onCreateSessionSafe`, which defeats the whole point.


amstone326:
Closing because somehow the link i have here and the fix I implemented aren't aligning.. Will reopen elsewhere with a couple of fixes combined

PR #1873, opened 2017-11-01
Try to handle boolean and int prefs

wpride:
The UI tests caught [this bug](https://gist.github.com/wpride/bc47750867cd87ef0ad9208d77898478) on opening Recovery Mode

One of Aliza's recent refactors expected `SharedPreference`s to only be of String type:

https://github.com/dimagi/commcare-android/commit/8f92adfae4226100083e1739e2cbce3c81c2d054#diff-0f317061966a57e0ec9feda1db245a00R174

Meanwhile one of Shubham's refactors added a boolean pref:

https://github.com/dimagi/commcare-android/commit/fd687a256e18acd26a1421e210c19c82fded1395#diff-48d5e471c4e5e4492d24f712bf5d37e9R571

I expect that we want the `onSharedPreferenceUpdated` code to be robust against all setting types,  but unfortunately [there doesn't seem to be a clean way to do this](https://stackoverflow.com/questions/29615920/android-check-sharedpreferences-for-value-type)

I don't like this PR but it might be the cleanest way to go

### DIFF
--- a/app/src/org/commcare/fragments/CommCarePreferenceFragment.java
+++ b/app/src/org/commcare/fragments/CommCarePreferenceFragment.java
@@ -171,25 +171,13 @@ public abstract class CommCarePreferenceFragment extends PreferenceFragmentCompa
 
     @Override
     public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, String key) {
-        String prefValue = getPrefValueAsString2(sharedPreferences, key);
+        String prefValue = getPrefValueAsString(sharedPreferences, key);
         if (prefValue != null) {
             FirebaseAnalyticsUtil.reportEditPreferenceItem(key, prefValue);
         }
     }
 
     private static String getPrefValueAsString(SharedPreferences sharedPreferences, String key) {
-        try {
-            return sharedPreferences.getString(key, null);
-        } catch (ClassCastException e) {
-            try {
-                return String.valueOf(sharedPreferences.getBoolean(key, false));
-            } catch (ClassCastException e2) {
-                return String.valueOf(sharedPreferences.getInt(key, -1));
-            }
-        }
-    }
-
-    private static String getPrefValueAsString2(SharedPreferences sharedPreferences, String key) {
         return String.valueOf(sharedPreferences.getAll().get(key));
     }
 }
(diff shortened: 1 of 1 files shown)


ctsims:
@wpride - I think we should also change the preference Shubham introduced into a string, and actually just reference that specific value in the polymorphic checker for now, rather than generically being able to reference arbitrary types, if that's the only one.

The methods that reference the value can translate it into/out of a boolean (boolean.toString() is fine), but I think there's not a clear value for us to allow the data in the preferences table to be non-homogenous, there's a big downside (constantly having to perform checks like this everywhere).

wpride:
@ctsims there are a few places we use boolean preferences:

https://github.com/dimagi/commcare-android/blob/0ae66c4e5011e7e86dc8cda9d3af68cf82ac7f00/app/src/org/commcare/views/widgets/QuestionWidget.java#L495-L495
https://github.com/dimagi/commcare-android/blob/8aca8e0c47b0ca89f4d7ce6c17fab5321973503a/app/src/org/commcare/models/legacy/LegacyCommCareUpgrader.java#L81-L81
https://github.com/dimagi/commcare-android/blob/8f92adfae4226100083e1739e2cbce3c81c2d054/app/src/org/commcare/preferences/CommCarePreferences.java#L547-L547
https://github.com/dimagi/commcare-android/blob/8f92adfae4226100083e1739e2cbce3c81c2d054/app/src/org/commcare/preferences/CommCarePreferences.java#L472-L472

Looks like the first is unused (could probably remove that code?) and the second is just in a migration, but the latter two are used (and I just confirmed that opting out of analytics crashes CommCare as well). I think we'd have to write a migration for these if we wanted to turn them into `String`s going forward.

app/src/org/commcare/fragments/CommCarePreferenceFragment.java
```diff
@@ -171,9 +171,25 @@ protected boolean isPersistentAppPreference() {
...
+            try {
+                return String.valueOf(sharedPreferences.getBoolean(key, false));
+            } catch (ClassCastException e2) {
+                return String.valueOf(sharedPreferences.getInt(key, -1));
+            }
+        }
+    }
+
+    private static String getPrefValueAsString2(SharedPreferences sharedPreferences, String key) {
+        return String.valueOf(sharedPreferences.getAll().get(key));
```
amstone326: @wpride @ctsims What about this instead? (I just tested that it works with both a String and non-String pref value)

wpride: Looks great, nice call

PR #1876, opened 2017-11-03
Adds appId to Entity Cache

shubham1g5:
* Adds appId to entity_cache and modifies the current queries to query using the current appID


app/src/org/commcare/models/database/user/models/EntityStorageCache.java
```diff
@@ -115,7 +115,9 @@ public void invalidateCaches(Collection<Integer> recordIds) {
         List<Pair<String, String[]>> whereParamList = TableBuilder.sqlList(recordIds);
         int removed = 0;
         for (Pair<String, String[]> querySet : whereParamList) {
-            removed += db.delete(TABLE_NAME, COL_CACHE_NAME + " = '" + this.mCacheName + "' AND " + COL_ENTITY_KEY + " IN " + querySet.first, querySet.second);
+            removed += db.delete(TABLE_NAME, COL_APP_ID + " = '" + mAppId + "' AND " +
```
ctsims: Somewhat counterintuitively, I don't think this specific call actually should be app restricted.

If the case model changes, it should invalidate caches around that case model regardless of which app produced them.

app/src/org/commcare/models/database/user/models/EntityStorageCache.java
```diff
@@ -133,11 +135,17 @@ public static int getSortFieldIdFromCacheKey(String detailId, String cacheKey) {
...
 
-    public static void tryWipeCache() {
+    public static void wipeCacheForCurrentApp() {
         SQLiteDatabase userDb = CommCareApplication.instance().getUserDbHandle();
-        SqlStorage.wipeTable(userDb, TABLE_NAME);
-        String uuid = CommCareApplication.instance().getSession().getLoggedInUser().getUniqueId();
-        setEntityCacheWipedPref(uuid, CommCareApplication.instance().getCurrentApp().getAppRecord().getVersionNumber());
+        userDb.beginTransaction();
+        try {
+            userDb.delete(TABLE_NAME, COL_APP_ID + " = ?", new String[]{ReportingUtils.getAppId()});
```
ctsims: I'm pretty uncomfortable with getting the current app ID from ReportingUtils. I think we should move that somewhere more semantically "important", since I wouldn't anticipate someone making changes to ReportingUtils assuming they would break core app behavior if they made changes.

app/unit-tests/src/org/commcare/android/util/TestUtils.java
```diff
@@ -103,15 +104,15 @@ public TransactionParser getParser(KXmlParser parser) {
                     //there's no good lifecycle to manage the bulk processor in, but at least
                     //this will validate that the bulk processor works.
                     if(bulkProcessingEnabled)  {
-                        return new AndroidBulkCaseXmlParser(parser, getCaseStorage(db), new EntityStorageCache("case", db), new AndroidCaseIndexTable(db)) {
+                        return new AndroidBulkCaseXmlParser(parser, getCaseStorage(db), new EntityStorageCache("case", db, ReportingUtils.getAppId()), new AndroidCaseIndexTable(db)) {
```
ctsims: Same comment as above WRT reportingUtils. Very important case-in-point: ReportingUtils fails by producing a blank string. We want to fail with a real exception here that bubbles up, since these methods won't work without a real id.

app/src/org/commcare/models/database/user/models/EntityStorageCache.java
shubham1g5: What about the respective singular method - `invalidateCache` . Is it fine to incorporate appId in that. I can see that whie doing bulk processing we use `invalidateCaches` instead of `invalidateCache` but couldn't really find any code which indicate that `invalidateCaches` gets called in case of case model changes. Also a bit unclear what case model refers to in this case specifically.

ctsims: I think in general any time we are invalidating a cache due to a change to a case record we should *not* restrict by appId. 

I was under the impression that this gets called after a case is changed, and that the recordId is the row Id of a case in the db, is that not correct?

shubham1g5: Hmm. I agree that we don't need to restrict these calls by appID as we are querying on case id which would be a sufficient condition to get to a specific row.  Though this gets called from `BulkElementParser::parse` and I am not sure whether it's specically related to case model change.

shubham1g5:

app/unit-tests/src/org/commcare/android/util/TestUtils.java
shubham1g5: @ctsims After making this change, `processResourceTransaction(String resourcePath)` is failing due to `java.lang.RuntimeException: No App installed!!!` exception while getting appID at a bunch of places. Is this expected and we need to fail silently here or we need to change the tests to install an app first before calling `processResourceTransaction`.

ctsims: Hm, my guess is that the fundamental data type (EntityStoreCache) should be allowed to get created with no App ID, and if it's going to fail that it should fail laziy, or that we should not load the entity cache if the app ID isn't available.

there are operations which are possible without the app ID, it might be the case that we should allow it to be loaded in a way where those functions work, but anything that relies on an appID fails if you try to use them in that mode.

shubham1g5: I am now passing entityCache as a Nullable and checking to see if it's exist to get past these errors. Though not sure if that's the right way to go based on your above comment.

ctsims: I think this is essentially equivalent to what we were doing previously, so this looks good to me for now. 

I'm a bit confused though, I thought that the appID being null was what was causing those issues. Was it actually that the entityCache itself the source of problems?

ctsims: oh, whoops, misread that. Re-reviewing now, I think there's one change to make.

app/unit-tests/src/org/commcare/android/util/TestUtils.java
```diff
@@ -97,21 +99,27 @@ public TransactionParser getParser(KXmlParser parser) {
...
 
                     //Note - this isn't even actually bulk processing. since this class is static
                     //there's no good lifecycle to manage the bulk processor in, but at least
                     //this will validate that the bulk processor works.
-                    if(bulkProcessingEnabled)  {
-                        return new AndroidBulkCaseXmlParser(parser, getCaseStorage(db), new EntityStorageCache("case", db), new AndroidCaseIndexTable(db)) {
+                    EntityStorageCache entityStorageCache;
+                    try {
+                        entityStorageCache = new EntityStorageCache("case", db, AppUtils.getCurrentAppId());
+                    } catch (Exception e) {
```
ctsims: If the only reason this ever occurs is the null currentAppId we should check for that explicitly, rather than catch a generic exception here, which will hide other errors that we aren't anticipating.

shubham1g5: Agree. Updated.

PR #1877, opened 2017-11-03
Handle returning to form entry/hierarchy after process is killed in background

amstone326:
Clayton helped me determine that https://fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/59c6c7cabe077a4dcc21be3b?time=last-thirty-days happens when the following sequence of events occurs:

1. The user was last in `FormEntryActivity` before CommCare was sent to the background for any reason
2. While CommCare is in the background, the process gets killed 
3. The user reopens CommCare, and thus CommCare tries to restart `FormEntryActivity` fresh, and is in the odd state of _having_ a `savedInstanceState` (hence why `hasFormLoadBeenTriggered` is true), but no longer has the `mFormController` because it is not stored in the same way. There is no way for us to recover from this situation, so the best thing to do is just exit gracefully, which I'm doing via the session-aware activity harnessing. 

**UPDATE**: the same issue was causing a crash in `FormHierarchyActivity`: https://fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/59ddea56be077a4dcc1a0dce?time=last-thirty-days. If we return directly to `FormHierarchyActivity` after the process was killed while CommCare was in the background, we encounter a NPE when trying to access `FormEntryActivity.mFormController`. The latest commit addresses this.


phillipm:
Wait, does this fix the mysterious `mFormController` is `null` exception that has been plaguing error logs forever?!! :clap: :clap:

amstone326:
@phillipm You are absurd. But yes :slightly_smiling_face:

amstone326:

amstone326:

amstone326:
@ctsims See last commit. Unfortunately it took a bit more refactoring than it might have seemed at first glance, but I think the end result is an improvement for sure. Thanks for the suggestion. (I know you approved already but since this is a decently-sized refactor, want to pass it by you again before merging)

wpride:
@amstone326 nice refactor, very clean inheritance structure

PR #1883, opened 2017-11-10
Return to form entry after session expiration via SSD

amstone326:
Addresses https://manage.dimagi.com/default.asp?247211 by using the stored `SessionStateDescriptor` to re-load an incomplete form after it is interrupted by session expiration.

@ctsims: This is a slightly different way of doing this than we discussed, but I think it's much cleaner and more reliable. When I started trying to utilize the session-saving infrastructure for this, the workflows I had to use felt both messy and brittle. I then noticed the paradigm that was being used in `AndroidSessionWrapper.getExistingIncompleteCaseDescriptor()` to detect when the session for which the user is trying to load a form already has an incomplete form in storage, and realized that a similar strategy could be used here. The strategy I adopted does a check upon login for a stored `SessionStateDescriptor` that has been marked as "interrupted" by a session expiration, and re-loads it and its corresponding `FormRecord` if so.

1 question I have: By doing it this way, the default result is that the user gets dumped into the `FormHierarchyActivity` for their saved form, rather than directly into form entry. It is easy enough to change that if that seems undesirable, but I didn't have a strong opinion either way. Let me know if you do.


amstone326:
Product Note: This will make it so that when form entry is interrupted by the automatic user session expiration, the user will be entered directly back into that form when they log back in, rather than having to figure out what happened and navigate to re-opening the saved form themselves.

amstone326:

amstone326:
@wpride Thanks for the heads up, resolved conflicts. @ctsims Would still like you to take a last look at this before it's merged.

app/src/org/commcare/models/database/user/UserDbUpgradeUtils.java
```diff
@@ -198,4 +200,28 @@ protected static void migrateFormRecordsToV3(Context c, SQLiteDatabase db) {
...
+                SessionStateDescriptor.STORAGE_KEY,
+                SessionStateDescriptorV1.class,
+                new ConcreteAndroidDbHelper(c, db));
+
+        SqlStorage<SessionStateDescriptor> newStorage = new SqlStorage<>(
+                SessionStateDescriptor.STORAGE_KEY,
+                SessionStateDescriptor.class,
+                new ConcreteAndroidDbHelper(c, db));
+
+        for (SessionStateDescriptorV1 ssd : oldStorage) {
```
ctsims: @amstone326 

Important note: Within this file. _all SSD references_ before this migration occurs (which is all of them for now) need to be on SSDV1, otherwise a Db update from, say, v5 to v22 could try to load the SSD it's manipulating as an SSDV2 and crash.

amstone326: @ctsims Very good catch, thank you.

amstone326: ..Ah, I just realized your other suggestion makes this irrelevant. I think that is a better way to go so I'll do that refactor now

amstone326:
ok @ctsims went with your suggestion of storing the interrupted SSD's id in prefs instead of using the flag. Definitely simpler, and also has the advantage of allowing us to be more precise about which SSD we intend to retrieve

amstone326:

app/unit-tests/src/org/commcare/android/tests/processing/FormStorageTest.java
```diff
@@ -304,6 +304,7 @@
             , "org.javarosa.core.model.actions.SendAction"
 
             // Added in 2.41
+            , "org.commcare.android.database.user.models.SessionStateDescriptorV1"
```
ctsims: This no longer exists and is causing a test failure now

app/src/org/commcare/models/AndroidSessionWrapper.java
```diff
@@ -152,6 +141,47 @@ public SessionStateDescriptor getExistingIncompleteCaseDescriptor() {
...
+     * @return
+     */
+    public static SessionStateDescriptor getFormStateForInterruptedUserSession() {
+        int idOfInterrupted = HiddenPreferences.getIdOfInterruptedSSD();
+        if (idOfInterrupted == -1) {
+            return null;
+        }
+        SqlStorage<SessionStateDescriptor> sessionStorage =
+                CommCareApplication.instance().getUserStorage(SessionStateDescriptor.class);
+        SessionStateDescriptor interrupted = sessionStorage.read(idOfInterrupted);
```
ctsims: Two quick things:

1)  I don't know _precisely_ why this could fail after fixing #2 below (no record exists), but I think that if this record isn't in session storage we shouldn't fail.
2) Shared preferences are per-app, not per-user, so right now this can bleed across userdbs (which could cause #1, but could also just cause bizarre behavior). Easiest fix is to append the user ID to the shared preferences key in the HiddenPreferences changes

PR #1884, opened 2017-11-13
Don't override hierarchy view back-behavior

amstone326:
Confirmed with substantial testing that this fixes https://fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/59d37644be077a4dccac473f?time=last-thirty-days, while maintaining the behavior that was the goal of https://github.com/dimagi/commcare-android/pull/640 (which is what introduced this bug). The reason this PR was causing the crash is as follows: If a user went to the hierarchy view from form entry while on a question that was within any type of group, pressing back would result in calling `goUpLevel()`, which modifies the form index without any provisions for re-setting it afterward. Thus when the user pressed back _again_ and ended up back in `FormEntryActivity`, their form index would be screwed up.

~~The change I made here is that we should only override the back button behavior in `FormHierarchyActivity` if we came from the list of saved/incomplete forms. The reason I solved it this way rather than trying to store and re-set the form index before returning to form entry is that I think this is potentially undesirable behavior to employ when the user has come from form entry regardless. For example, in the example I described above, the user would expect pressing back once to take them back to form entry, since it's irrelevant to them that they happen to have been within a nested group before coming to the hierarchy view. In other words, it doesn't make any sense that the behavior of back in this case would vary depending on what type of question the user was on in form entry beforehand. I think this consideration strongly outweighs the value of having `back` take the user up a level if they do happen to start navigating around within the hierarchy view after coming from form entry.~~

This PR makes it so that pressing back in hierarchy activity no longer does anything but cancel the activity. All that has been changed now is to make sure that we first re-set the form index to what it was when the activity started, to avoid ending up in a totally wacko state.

**Product Note** _(this is incredibly minor and almost certainly not worth announcing, but could be useful for support to know in case anyone writes in complaining that the behavior changed)_: Makes it so that pressing the 'back' button while in the form hierarchy (the view that shows the outline of a form and allows you to jump to a question) no longer does anything but cancel the activity. Previously, pressing back from there would _sometimes_ result in doing navigation within the hierarchy view instead, but this has been removed in favor of universally-predictable behavior.


ctsims:
@amstone326 After your latest change, would we still need the behavior where you can't use 'back' to navigate up when coming from a form entry session? It seems likely that your latest change would address that issue structurally without limiting the user from pressing 'back' to navigate up.

It's likely that it's still the right behavior in-form (IE: "One hits "go to prompt" sees the screen, doesn't want to be there, then hits 'back' to go to the previous screen and instead navigates internally in form heirarchy), just wanted to double check that we were taking advantage of the new code, if possible.

amstone326:

amstone326:
@ctsims Updated

ctsims:
@amstone326 Signed off on the code change.

Want to flag that this is not actually invisible (pr flag wise), since it will change the expected user experience for mobile users with no special privileges. 

Probably not worth communicating proactively to people, but almost certainly worth communicating to the Support Team so they will know that any bug reports matching this ("I press back and it does something different from before") are expected and not bugs.

PR #1885, opened 2017-11-14
Remove server settings config screen + Split out hidden settings from configurable settings

amstone326:
This makes 2 major organizational changes to CommCare settings:

1. Get rid of the server settings configuration screen entirely, since we've deemed this dangerous and not necessary. A simple static class `CCServerUrls` now exists to manage the relevant keys and provide read methods for these values.

2. We had up until now been utilizing the `CommCarePreferences` class for a bizarre dual purpose -- it was both a `PreferenceFragment` that presented configurable settings to the user, and a catch-all place for managing many other settings that were _not_ configurable in the mobile UI. I have now split out these 2 roles into 2 different classes. The `PreferenceFragment` with user-configurable prefs is called `MainConfigurablePreferences`, and I created a new class called `HiddenCommCarePreferences` to server the role of managing all preferences that are not configurable in the mobile UI.

I also did bits of linting/clean-up, including changing usages of `commit()` to `apply()` where I saw them, because this is the Android-recommended method.
 
**Also** this is the ui test that should be merged in conjunction with this PR: https://github.com/dimagi/commcare-ui-tests/pull/120


amstone326:

app/src/org/commcare/preferences/CCPrefValues.java
```diff
@@ -0,0 +1,26 @@
+package org.commcare.preferences;
+
+import android.util.DisplayMetrics;
+
+/**
+ * Created by amstone326 on 11/14/17.
+ */
+public class CCPrefValues {
```
wpride: What do you think about naming these `PrefValues` and `ServerUrls`

app/src/org/commcare/preferences/HiddenCommCarePreferences.java
```diff
@@ -0,0 +1,197 @@
...
+
+/**
+ * Provides hooks for reading and writing all preferences that are not configurable by the user in
+ * the mobile UI. There are 2 classes of settings that fall into this category:
+ * 1) Those whose values are sent down from HQ via the app profile
+ * 2) Those whose values are set based upon actions that a user takes during their usage of CommCare
+ *
+ * Created by amstone326 on 11/14/17.
+ */
+public class HiddenCommCarePreferences {
```
wpride: This is a great, great refactor

wpride:
One small question and one comment inline.

Only big question: don't moving these preferences around require a migration?

app/src/org/commcare/preferences/CCPrefValues.java
amstone326: sure

amstone326:
@wpride What about this would require a migration? I haven't renamed any of the preference keys with which the values are stored in `SharedPreferences`; I've just changed which static class some of those keys live in.

wpride:
Ah I see the underlying `SharedPreferences` hasn't changed

PR #1887, opened 2017-11-14
Session aware infrastructure improvements

amstone326:
A lot of the crashes that have been showing up in Crashlytics are related to unhandled instances of `SessionUnavailableException`, despite the fact that we have a framework in place that is supposed to address this, because it is currently incompletely implemented. This PR fills in the remaining gaps by doing the following:

- Introducing the notion of `onActivityResultSessionSafe()`, and applying it to all classes that extend `SessionAwareCommCareActivity`. Since `onActivityResult()` gets called before `onResume()`, if `onActivityResult()` requires the session in any way, then we would never reach our protections in `onResume()` and instead just crash there. This looks to be what was causing https://fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/59c3fc75be077a4dcc02c8f0?time=last-thirty-days
- Making sure that _all_ classes that extend `SessionAwareCommCareActivity` implement `onCreateSessionSafe` instead of `onCreate`. This had been done very incompletely for some reason. This looks to be what was causing https://fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/59f979c161b02d480d8ff90d?time=last-thirty-days


app/src/org/commcare/activities/FormRecordListActivity.java
```diff
@@ -139,8 +139,8 @@ public String getMessage() {
     }
 
     @Override
-    protected void onCreate(Bundle savedInstanceState) {
-        super.onCreate(savedInstanceState);
+    protected void onCreateSessionSafe(Bundle savedInstanceState) {
```
amstone326: Should fix https://fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/59f979c161b02d480d8ff90d?time=last-thirty-days

app/src/org/commcare/activities/HomeScreenBaseActivity.java
```diff
@@ -358,7 +358,7 @@ protected void onSaveInstanceState(Bundle outState) {
     }
 
     @Override
-    protected void onActivityResult(int requestCode, int resultCode, Intent intent) {
+    protected void onActivityResultSessionSafe(int requestCode, int resultCode, Intent intent) {
```
amstone326: Should fix https://fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/59c3fc75be077a4dcc02c8f0?time=last-thirty-days

ctsims:
This is awesome! Thanks for going through these.

PR #1888, opened 2017-11-16
Amaze fix

shubham1g5:
Case: https://manage.dimagi.com/default.asp?262091#1419259

Fixes: https://fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/59e776f361b02d480dc9bd46?time=last-ninety-days

Apologies for not a clean PR but I had to refactor some things so that I don't have to make same change at 2 places. Though The crux of the PR is when we fail to get the filePath from file provider we need to either do all the post processing by opening a inputStream and work with that or make a copy of the file internally using that inputStream and treat that file's path as the new file path. 

Product Note: Fixes a bug on Android N upwards where choosing a file through Amaze file provider anywhere in Commcare will crash the app.


shubham1g5:

shubham1g5:

shubham1g5:

app/src/org/commcare/activities/InstallArchiveActivity.java
```diff
@@ -149,11 +151,28 @@ private void onUnzipSuccessful() {
...
+            // result
+            Toast.makeText(context,
+                    Localization.get("mult.install.state.invalid.path"),
+                    Toast.LENGTH_SHORT).show();
+        } else {
+            String filePath;
+            try {
+                filePath = UriToFilePath.getPathFromUri(CommCareApplication.instance(), uriPath);
+            } catch (UriToFilePath.NoDataColumnForUriException e) {
+                filePath = intent.getData().toString();
```
shubham1g5: Saving the filePath as content uri itself in case of an Exception, which gets displayed in edittext for filepath and later gets processed as Content uri

app/src/org/commcare/activities/InstallArchiveActivity.java
```diff
@@ -176,7 +194,7 @@ private void evalState() {
             return;
         }
 
-        if (!(new File(location)).exists()) {
+        if (!(location.startsWith("content://") || (new File(location)).exists())) {
```
shubham1g5: Not doing the check for file presence in case of content uri here

app/src/org/commcare/activities/components/ImageCaptureProcessing.java
```diff
@@ -117,10 +136,49 @@ public static void processImageChooserResponse(FormEntryActivity activity,
...
+            showInvalidImageMessage(activity);
+            return;
+        }
+
+        try {
+            String imagePath = UriToFilePath.getPathFromUri(activity, selectedImage);
+            processImageGivenFilePath(activity, instanceFolder, imagePath);
+        } catch (UriToFilePath.NoDataColumnForUriException e) {
+            // Can't get file path from Uri, so need to work with uri instead
+            processImageGivenFileUri(activity, instanceFolder, selectedImage);
```
shubham1g5: Handling uri differently than filepath (If we can't get gile path, we are using the uri to first make a copy of the file internally and then passing that new file's path to the `processImageGivenFilePath`)

app/src/org/commcare/preferences/MainConfigurablePreferences.java
```diff
@@ -101,25 +99,13 @@ protected void setupPrefClickListeners() {
...
-                }
+            analyticsButton.setOnPreferenceClickListener(preference -> {
+                showAnalyticsOptOutDialog();
+                return true;
             });
         } else {
             getPreferenceScreen().removePreference(analyticsButton);
         }
-
-        Preference printTemplateSetting = findPreference(PREFS_PRINT_DOC_LOCATION);
```
shubham1g5: made print preference use `FilePreference` construct, so we don't need to handle this explicitly here.

app/src/org/commcare/preferences/MainConfigurablePreferences.java
```diff
@@ -172,14 +158,7 @@ public void onClick(DialogInterface dialog, int which) {
 
     @Override
     public void onActivityResult(int requestCode, int resultCode, Intent data) {
-        if (requestCode == REQUEST_TEMPLATE) {
```
shubham1g5: Handled in `FilePreference`

app/src/org/commcare/preferences/MainConfigurablePreferences.java
```diff
@@ -190,22 +169,6 @@ else if (resultCode == DeveloperPreferences.RESULT_DEV_OPTIONS_DISABLED) {
         }
     }
 
-    private void tryToSetPrintTemplateLocation(Intent data) {
```
shubham1g5: Handled in `FilePreference`

app/src/org/commcare/tasks/MultimediaInflaterTask.java
```diff
@@ -14,79 +13,10 @@
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
 
-public abstract class MultimediaInflaterTask<R> extends CommCareTask<String, String, Boolean, R> {
+public abstract class MultimediaInflaterTask<R> extends UnzipTask<R> {
 
     @Override
-    protected Boolean doTaskBackground(String... params) {
```
shubham1g5: MultimediaInflaterTask was completely same as `UnzipTask`, so exteding from `UnzipTask` instead and removing the duplicate code

app/src/org/commcare/tasks/UnzipTask.java
```diff
@@ -1,109 +1,137 @@
...
         this.taskId = UNZIP_TASK_ID;
-
         TAG = UnzipTask.class.getSimpleName();
     }
 
     @Override
     protected Integer doTaskBackground(String... params) {
-        File archive = new File(params[0]);
-        File destination = new File(params[1]);
+        if (params[0].startsWith("content://")) {
```
shubham1g5: Making distinction between filePath and uri here

app/src/org/commcare/tasks/UnzipTask.java
```diff
@@ -1,109 +1,137 @@
...
+            return unZipFromContentUri(params[0], params[1]);
+        } else {
+            return unZipFromFilePath(params[0], params[1]);
+        }
+    }
 
-        Log.d(TAG, "Unzipping archive '" + archive + "' to  '" + destination + "'");
+    private Integer unZipFromFilePath(String filePath, String destinationPath) {
```
shubham1g5: If we have a filepath already, get an inputSteam from it and pass it to `unZipFromStream`

app/src/org/commcare/tasks/UnzipTask.java
```diff
@@ -1,109 +1,137 @@
...
+            publishProgress(getInvalidZipFileErrorMessage());
+            return -1;
+        }
+        return unZipFromStream(zis, destinationPath);
+    }
 
-        int count = 0;
-        ZipFile zipfile;
-        //From stackexchange
+    private Integer unZipFromContentUri(String uriString, String destinationPath) {
```
shubham1g5: If we don't have a filePath but a content uri, get an inputStream for the file using uri and then ultimately pass the stream to `unZipFromStream`

app/src/org/commcare/tasks/UnzipTask.java
```diff
@@ -1,109 +1,137 @@
...
+    private Integer unZipFromStream(ZipInputStream zis, String destinationPath) {
+        File destination = new File(destinationPath);
+        int count = 0;
+        ZipEntry entry = null;
+        try {
+            while ((entry = zis.getNextEntry()) != null) {
+                publishProgress(Localization.get("mult.install.progress", new String[]{String.valueOf(count)}));
+                count++;
 
-            File outputFile = new File(destination, entry.getName());
```
shubham1g5: While unzipping, Instead of operating on file, we are now operating on a stream now

app/src/org/commcare/utils/UriToFilePath.java
```diff
@@ -72,6 +77,9 @@ public static String getPathFromUri(final Context context, final Uri uri) {
                 return getDataColumn(context, contentUri, selection, selectionArgs);
             }
         } else if ("content".equalsIgnoreCase(uri.getScheme())) {
+            // Return the remote address
```
shubham1g5: This is result of merging `FileUtil. getPathFromUri` in here.

app/src/org/commcare/views/widgets/AudioWidget.java
```diff
@@ -32,18 +23,13 @@
  * @author Yaw Anokwa (yanokwa@gmail.com)
  */
 public class AudioWidget extends MediaWidget {
```
shubham1g5: Moved a lot of common functionalities from Audio and Video Widgets to MediaWidget

app/src/org/commcare/views/widgets/MediaWidget.java
```diff
@@ -162,4 +180,117 @@ public void setFocus(Context context) {
...
+     */
+    protected String createFilePath(Object binaryuri) {
+        String path = "";
+        destMediaPath = "";
+        if (binaryuri instanceof Uri) {
+            try {
+                path = UriToFilePath.getPathFromUri(CommCareApplication.instance(),
+                        (Uri)binaryuri);
+            } catch (UriToFilePath.NoDataColumnForUriException e) {
+                // Need to make a copy of file using uri, so might as well copy to final destination path directly
```
shubham1g5: This is the major change here where we are making a internal copy first of file using uri and then passing it forward.

app/src/org/commcare/views/widgets/AudioWidget.java
shubham1g5: Moved a lot of common functionalities from Audio and Video Widgets to MediaWidget

app/src/org/commcare/views/widgets/MediaWidget.java
```diff
@@ -162,4 +180,117 @@ public void setFocus(Context context) {
                 (InputMethodManager)context.getSystemService(Context.INPUT_METHOD_SERVICE);
         inputManager.hideSoftInputFromWindow(this.getWindowToken(), 0);
     }
+
+    @Override
+    public void setBinaryData(Object binaryuri) {
```
shubham1g5: `setBinaryData` is almost same as the earlier implementation apart from the fact that we now might have already made a copy of file at the destination in `createFilePath` (in case we were only able to get the uri from file provider and not filepath). So we are now checking if the file already exists at `destMediaPath` and only copy the flle if it doesn't (which would be the case when we would have successfully got filePath from file provider)

shubham1g5:
@amstone326 Updated with comments. Hopefully it would make more sense now.

app/src/org/commcare/tasks/MultimediaInflaterTask.java
amstone326: nice

app/src/org/commcare/utils/UriToFilePath.java
```diff
@@ -102,8 +110,12 @@ private static String getDataColumn(Context context, Uri uri, String selection,
             cursor = context.getContentResolver().query(uri, projection,
                     selection, selectionArgs, null);
             if (cursor != null && cursor.moveToFirst()) {
-                final int column_index = cursor.getColumnIndexOrThrow(column);
-                return cursor.getString(column_index);
+                try {
+                    final int column_index = cursor.getColumnIndexOrThrow(column);
+                    return cursor.getString(column_index);
+                } catch (IllegalArgumentException e) {
+                    throw new NoDataColumnForUriException();
```
amstone326: This method should declare that it throws this exception so that all usages are forced to handle it

app/src/org/commcare/utils/UriToFilePath.java
```diff
@@ -135,4 +147,23 @@ private static boolean isDownloadsDocument(Uri uri) {
...
+
+    /**
+     * @param uri The Uri to check.
+     * @return Whether the Uri authority is Google Photos.
+     */
+    private static boolean isGooglePhotosUri(Uri uri) {
+        return "com.google.android.apps.photos.content".equals(uri.getAuthority());
+    }
+
+    public static void grantPermissionForUri(Context context,Intent i, Uri uri, int flags) {
```
amstone326: Can you add a comment explaining what this does? Also I don't think this file is the best place for this method to live

app/src/org/commcare/views/widgets/MediaWidget.java
```diff
@@ -162,4 +180,117 @@ public void setFocus(Context context) {
...
+        i.setDataAndType(mediaUri, mediaType);
+
+        UriToFilePath.grantPermissionForUri(getContext(), i, mediaUri, Intent.FLAG_GRANT_READ_URI_PERMISSION);
+        try {
+            getContext().startActivity(i);
+        } catch (ActivityNotFoundException e) {
+            Toast.makeText(getContext(),
+                    StringUtils.getStringSpannableRobust(getContext(),
+                            R.string.activity_not_found,
+                            "play media"),
```
amstone326: I think it's valuable to specify which type of media activity we weren't able to find.

shubham1g5:

app/src/org/commcare/preferences/FilePreferenceDialogFragmentCompat.java
```diff
@@ -56,6 +66,26 @@ public void onActivityResult(int requestCode, int resultCode, Intent intent) {
...
+        // we need to copy the file internally and pass the filepath of the new file
+        String filePath = getContext().getFilesDir().getAbsolutePath() + "/temp/" + FileUtil.getFileName(uri.toString());
+        File file = new File(filePath);
+        try {
+            InputStream fileStream = getContext().getContentResolver().openInputStream(uri);
+            FileUtil.ensureFilePathExists(file);
+            FileUtil.copyFile(fileStream, file);
+        } catch (IOException e1) {
+            e1.printStackTrace();
+            Logger.log(LogTypes.TYPE_MAINTENANCE, "Getting new file path from uri failed due to " + e1.getMessage());
```
wpride: Should call `Logger.exception` to log full stack trace

app/src/org/commcare/preferences/FilePreferenceDialogFragmentCompat.java
```diff
@@ -56,6 +66,26 @@ public void onActivityResult(int requestCode, int resultCode, Intent intent) {
...
+        // since we are going to use this path later on in some other
+        // activity stack which will not have permissions for this uri
+        // we need to copy the file internally and pass the filepath of the new file
+        String filePath = getContext().getFilesDir().getAbsolutePath() + "/temp/" + FileUtil.getFileName(uri.toString());
+        File file = new File(filePath);
+        try {
+            InputStream fileStream = getContext().getContentResolver().openInputStream(uri);
+            FileUtil.ensureFilePathExists(file);
+            FileUtil.copyFile(fileStream, file);
+        } catch (IOException e1) {
```
wpride: Any reason this is `e1` instead of `e`?

app/src/org/commcare/preferences/FilePreferenceDialogFragmentCompat.java
```diff
@@ -47,7 +52,12 @@ public void onActivityResult(int requestCode, int resultCode, Intent intent) {
...
             if (resultCode == RESULT_OK && intent != null) {
                 Uri uri = intent.getData();
-                String filePath = UriToFilePath.getPathFromUri(CommCareApplication.instance(), uri);
+                String filePath;
+                try {
+                    filePath = UriToFilePath.getPathFromUri(CommCareApplication.instance(), uri);
+                } catch (UriToFilePath.NoDataColumnForUriException e) {
+                    filePath = getNewPathFromUri(uri);
+                }
                 validateFile(filePath);
```
wpride: if `getNewPathFromUri` above returns null we'll call `Toast.makeText()` twice in quick succession here https://github.com/dimagi/commcare-android/blob/8b4982199f3d9b45f795996ca783c249f554e956/app/src/org/commcare/preferences/FilePreferenceDialogFragmentCompat.java#L59-L59 and then here https://github.com/dimagi/commcare-android/blob/8b4982199f3d9b45f795996ca783c249f554e956/app/src/org/commcare/preferences/FilePreferenceDialogFragmentCompat.java#L119-L119

Might be better to throw an exception from `getNewPathFromUri` and handle that rather than returning null

app/src/org/commcare/preferences/MainConfigurablePreferences.java
```diff
@@ -101,25 +99,13 @@ protected void setupPrefClickListeners() {
 
         Preference analyticsButton = findPreference(DISABLE_ANALYTICS);
         if (MainConfigurablePreferences.isAnalyticsEnabled()) {
-            analyticsButton.setOnPreferenceClickListener(new Preference.OnPreferenceClickListener() {
-                @Override
-                public boolean onPreferenceClick(Preference preference) {
-                    showAnalyticsOptOutDialog();
-                    return true;
-                }
+            analyticsButton.setOnPreferenceClickListener(preference -> {
```
wpride: Oh look out its Lambda guy!

app/src/org/commcare/tasks/UnzipTask.java
```diff
@@ -1,109 +1,137 @@
...
+            return unZipFromFilePath(params[0], params[1]);
+        }
+    }
 
-        Log.d(TAG, "Unzipping archive '" + archive + "' to  '" + destination + "'");
+    private Integer unZipFromFilePath(String filePath, String destinationPath) {
+        File archive = new File(filePath);
+        ZipInputStream zis;
+        try {
+            zis = new ZipInputStream(new FileInputStream(archive));
```
wpride: NIT but these two functions feel un-DRY (wet?) Once you have the `inputStream` the try/catch blocks and handling re identical

app/src/org/commcare/tasks/UnzipTask.java
```diff
@@ -1,109 +1,137 @@
...
                 }
-            } finally {
-                try {
-                    outputStream.close();
-                } catch (IOException ioe) {
+                if (outputFile.exists()) {
+                    //Try to overwrite if we can
+                    if (!outputFile.delete()) {
+                        //If we couldn't, just skip for now
+                        continue;
```
wpride: Why is this OK? Couldn't this potentially result in an entry from the zip file not being copied (and then lost once the zip is deleted) silently?

app/src/org/commcare/utils/FileUtil.java
```diff
@@ -449,6 +449,13 @@ public static String getExtension(String filePath) {
         return "";
     }
 
+    public static String getFileName(String filePath) {
+        if (filePath.contains("/")) {
+            return last(filePath.split("/"));
+        }
+        return "";
```
wpride: This doesn't seem right... wouldn't this result in the file's path being the file path of the parent directory (and hence overwriting that directory)?

app/src/org/commcare/utils/UriToFilePath.java
```diff
@@ -72,6 +77,10 @@ public static String getPathFromUri(final Context context, final Uri uri) {
                 return getDataColumn(context, contentUri, selection, selectionArgs);
             }
         } else if ("content".equalsIgnoreCase(uri.getScheme())) {
+            // Return the remote address
+            if (isGooglePhotosUri(uri)) {
```
wpride: Is this from SO? Or can you link to docs about why this is handled specially?

app/src/org/commcare/views/widgets/MediaWidget.java
```diff
@@ -162,4 +180,117 @@ public void setFocus(Context context) {
...
+                        (Uri)binaryuri);
+            } catch (UriToFilePath.NoDataColumnForUriException e) {
+                // Need to make a copy of file using uri, so might as well copy to final destination path directly
+                InputStream inputStream;
+                try {
+                    inputStream = getContext().getContentResolver().openInputStream((Uri)binaryuri);
+                } catch (FileNotFoundException e1) {
+                    showToast("form.attachment.notfound");
+                    e1.printStackTrace();
+                    return "";
```
wpride: This and line 267 seem incorrect/potentially dangerous to me. Execution will proceed with "" as a path which in the worst case could over-write or delete a parent directory IE here https://github.com/dimagi/commcare-android/blob/7a2de5a271c461e56cd792ac0e2d368e423a73ed/app/src/org/commcare/views/widgets/MediaWidget.java#L190-L190. These exceptions should be logged and handled explicitly.

app/src/org/commcare/activities/components/ImageCaptureProcessing.java
ctsims: When choosing a file from the file system this code looks like it ends up trying to delete the selected image from its original location, which doesn't seem like the right approach to me. If someone just _Took_ a picture it's fine (desirable actually) for it to not exist elsewhere on the device. If they chose one, they'll likely be annoyed if it gets deleted afterwards.

app/src/org/commcare/views/widgets/MediaWidget.java
```diff
@@ -162,4 +180,117 @@ public void setFocus(Context context) {
...
+                    e1.printStackTrace();
+                    showToast("form.attachment.copy.fail");
+                    return "";
+                }
+                path = destMediaPath;
+            }
+            customFileTag = "";
+        } else {
+            path = (String)binaryuri;
+            customFileTag = CUSTOM_TAG;
```
wpride: What purpose is `customFileTag` serving?

app/src/org/commcare/preferences/FilePreferenceDialogFragmentCompat.java
```diff
@@ -56,6 +66,26 @@ public void onActivityResult(int requestCode, int resultCode, Intent intent) {
...
+        File file = new File(filePath);
+        try {
+            InputStream fileStream = getContext().getContentResolver().openInputStream(uri);
+            FileUtil.ensureFilePathExists(file);
+            FileUtil.copyFile(fileStream, file);
+        } catch (IOException e1) {
+            e1.printStackTrace();
+            Logger.log(LogTypes.TYPE_MAINTENANCE, "Getting new file path from uri failed due to " + e1.getMessage());
+            Toast.makeText(getActivity(), Localization.get("file.selection.failed"), Toast.LENGTH_LONG).show();
+            return null;
```
ctsims: This seems like  kind of an odd return path to me v. throwing an I/O exception, since this is the fallback case. 

Do we really expect code to be able to function effectively if this method fails?

app/src/org/commcare/views/widgets/MediaWidget.java
ctsims: Same question as above WRT whether we shouldn't be more deliberate here with Exceptions rather than potentially ambiguous return types.

app/src/org/commcare/utils/UriToFilePath.java
shubham1g5: This is result of merging FileUtil::getPath in here. Corresponding code https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/utils/FileUtil.java#L622

app/src/org/commcare/tasks/UnzipTask.java
shubham1g5: I am not sure as well, I took it from master equivalent over here - https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/tasks/UnzipTask.java#L67 . Maybe @ctsims know more about why are we doing this ?

app/src/org/commcare/views/widgets/MediaWidget.java
shubham1g5: I guess we are trying to make a distinction between files recorded vs files chosen from file provider. Though don't really think if it's useful and can remove if it's not serving any pupose right now.

app/src/org/commcare/preferences/FilePreferenceDialogFragmentCompat.java
```diff
@@ -52,11 +52,15 @@ public void onActivityResult(int requestCode, int resultCode, Intent intent) {
...
-                String filePath;
+                String filePath = null;
                 try {
                     filePath = UriToFilePath.getPathFromUri(CommCareApplication.instance(), uri);
                 } catch (UriToFilePath.NoDataColumnForUriException e) {
-                    filePath = getNewPathFromUri(uri);
+                    try {
+                        filePath = getNewPathFromUri(uri);
+                    } catch (IOException ioe) {
+                        Logger.exception(LogTypes.TYPE_MAINTENANCE, e);
```
wpride: We just want to return here, right?

app/src/org/commcare/tasks/UnzipTask.java
wpride: Given that this is inherited doesn't need to block, but we should throw an Exception here and halt the process

wpride:
@ctsims I'm merging so we can cut the release and get QA started. I confirmed your comments were addressed so let us know if you had further.

PR #1889, opened 2017-11-16
Go back to getting fixture names as a Set for the migration

amstone326:
This was a really really dumb regression on my part :( When I fixed the issue with the indexed fixture paths table allowing duplicates [here](https://github.com/dimagi/commcare-android/pull/1839), I somehow decided that that meant I could also change the migration method that originally revealed this bug to go back to fetching the names as a list. The problem of course is that people who are upgrading still have the old version and so still potentially have duplicate rows in that table.

@ctsims It's possible we want to hotfix this? Crashlytics shows it as only affecting 14 users so I'm not sure how to weight that: https://fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/59d2e849be077a4dcca6ac46?time=last-thirty-days

### DIFF
--- a/app/src/org/commcare/models/database/IndexedFixturePathUtils.java
+++ b/app/src/org/commcare/models/database/IndexedFixturePathUtils.java
@@ -61,6 +61,12 @@ public class IndexedFixturePathUtils {
         }
     }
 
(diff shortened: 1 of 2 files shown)


amstone326:
Obviated by https://github.com/dimagi/commcare-android/pull/1892

PR #1890, opened 2017-11-20
Fixes NPE in AlertDialog after process kill

shubham1g5:
Case: https://manage.dimagi.com/default.asp?265664#1422233
Crash: https://fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/59d3443bbe077a4dcca9ffff?time=last-thirty-days

@amstone326 The reason behind this crash wasn't configuration change as it was working properly on rotaion but Android killing the process(Both Activity and Fragment) in case of low memory conditions. I was able to reproduce it by following @ctsims protip on turning the 'Don't keep Activities' developer options on and then leaving the activity when AlertDialog is open and returning to it.  

I am now dismissing these dialogues safely when such a case surfaces so that user can atleast access the corresponding activity. Though if we see value in retaining these dialogues in such cases we will need a way to recreate these dialogues whenever it happens.

### DIFF
--- a/app/src/org/commcare/views/dialogs/AlertDialogFragment.java
+++ b/app/src/org/commcare/views/dialogs/AlertDialogFragment.java
@@ -1,10 +1,12 @@
 package org.commcare.views.dialogs;
 
 import android.app.Dialog;
+import android.content.Context;
 import android.content.DialogInterface;
 import android.os.Bundle;
 import android.support.annotation.NonNull;
 import android.support.v4.app.DialogFragment;
+import android.util.Log;
 
 /**
  * Wrapper for CommCareAlertDialogs that allows them to persist across screen orientation changes,
@@ -34,6 +36,9 @@ public class AlertDialogFragment extends DialogFragment {
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setRetainInstance(true);
+        if (underlyingDialog == null) {
+            dismiss();
+        }
     }
 
     @Override
@@ -51,8 +56,13 @@ public class AlertDialogFragment extends DialogFragment {
     @Override
     @NonNull
     public Dialog onCreateDialog(Bundle savedInstanceState) {
+        if (underlyingDialog == null) {
+            return super.onCreateDialog(savedInstanceState);
+        }
         return underlyingDialog.getDialog();
     }
+
+
 
     @Override
     public void onDestroyView() {
--- a/app/src/org/commcare/views/dialogs/AlertDialogFragment.java
+++ b/app/src/org/commcare/views/dialogs/AlertDialogFragment.java
@@ -60,8 +60,6 @@ public class AlertDialogFragment extends DialogFragment {
         return underlyingDialog.getDialog();
     }
 
-
-
     @Override
     public void onDestroyView() {
         // Ohh, you know, just a 5 year old Android bug ol' G hasn't fixed yet
--- a/app/src/org/commcare/views/dialogs/AlertDialogFragment.java
+++ b/app/src/org/commcare/views/dialogs/AlertDialogFragment.java
@@ -34,9 +34,6 @@ public class AlertDialogFragment extends DialogFragment {
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setRetainInstance(true);
-        if (underlyingDialog == null) {
-            dismiss();
-        }
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/views/dialogs/AlertDialogFragment.java
```diff
@@ -51,6 +54,9 @@ public void onDismiss(DialogInterface dialog) {
     @Override
     @NonNull
     public Dialog onCreateDialog(Bundle savedInstanceState) {
+        if (underlyingDialog == null) {
+            return super.onCreateDialog(savedInstanceState);
```
amstone326: what makes this the right thing to do here?

amstone326: hat makes this the right thing to do here?

shubham1g5: This just creates an empty dialogue which gets dismissed in onCreate. I can't return null from here since DialogFragment assumes that this method always returns a Non null value, so to get past the NPE I couldn't think of any other solution.  Also I can't dismiss the dialogue from here since it's not been created yet.

amstone326: Hm, what's confusing me then is that [the docs](https://developer.android.com/reference/android/support/v4/app/DialogFragment.html#onCreateDialog(android.os.Bundle)) say that this method gets called AFTER `onCreate()`. I guess the issue is just that calling `dismiss()` in `onCreate()` doesn't stop `onCreateDialog()` from getting called?

shubham1g5: And yeah, you are right. `onCreateDialog()` gets called even if we dismiss the dialog in `onCreate()` and results in NPE. I am going to move even `dismiss()` inside onCreateDialog() just to make it bit less confusing.

amstone326:
Ah cool ok so this does make more sense to me, but realized 1 other concern now-- If `dismiss()` gets called _before_ you return a dialog from `onCreateDialog()`, will that not reverse the dismissal somehow? I think the answer to that is probably no, but wanted to double check that you've physically tested this new version of it and things still work fine?

ctsims:
@amstone326 The process getting cleaned up in the background is definitely a different lifecycle than the process just getting killed.

The background cleanup actually does get signaling and such (IE: gets an onStop and an onDestroy), it just goes straight to the activities rather than taking down the process deliberately.

@shubham1g5 Good catch, one thing: when this happens, we should in fact get onStops/onDestroys which may be a better place to do some cleanup than here. The state we're testing for here (underlying dialog has just disappeared and is null) could have resulted from any number of different states, rather than the specific one that you are looking for here, so it might trigger unexpected behavior later.

amstone326:
@ctsims Sorry I should have been more specific -- what I tested was pressing the button in the device monitor to kill the process _while_ CommCare was in the background. should that not amount to the same thing?

shubham1g5:
@amstone326  For me the crash was happening on returning back to the killed process i.e. when the activity after being recreated was trying to create the fragment from it's old state. Exact steps I was following after turing on the 'Don't keep Activities' developer option were like this - 

1. Open sort options dialoge on case list
2. Press home button so that we exit the activities and our activity and fragment both gets killed
2. Go back to the Commcare from recent apps screen and observe the crash

I am not sure how would we simulate the returning back part by using the stop process from device monitor. 

@ctsims Agree that underlying dialog can become null due to more scenarios though not really sure what cleanup can be done in OnStop to prevent this from happening. The stop/destroy methods do get called for this fragment but the poblem only happens when this fragment gets recreated.

ctsims:
@amstone326 Confirming that the process your'e describing is slightly different. The "Activity killed in background" action is part of the lifecycle of the activity, rather than the CommCare process as a whole. The activity can be hard-backgrounded while the CommCare process is still running, which is what Shubham's test replicates. Killing the process entirely results in a different execution path.

@shubham1g5 I was assuming we'd handle it by calling "dismiss()" at the time the activity is backgrounded rather than when it started back up (or something similar), which would ideally result in this all being handled without needing a check, since the previous fragment body detached the state.

shubham1g5:
@ctsims Still not sure how can it be achieved. The problem is combination of these 2 facts - 

1. We can't do any fragment transaction after Activity's `onPause()` since that's the last point at which activity will remember any fragment transaction (by calling `onSaveInstanceState`)

2.  We don't wanna dismiss the dialoge fragment if Activitiy is just getting destoyed because of some configuration change since in this case our fragment would still be alive and just detached with the activity. 

Because of point2, the first point in Fragmnet's lifecycle where the cleanup can be done is Fragment's 'onDestroy' which is too late in consideration with the point1.

PR #1893, opened 2017-11-21
Send last sync time parameter for heartbeat in ISO8601 format

amstone326:
fyi @snopoke

### DIFF
--- a/app/src/org/commcare/heartbeat/HeartbeatRequester.java
+++ b/app/src/org/commcare/heartbeat/HeartbeatRequester.java
@@ -123,9 +123,14 @@ public class HeartbeatRequester {
     }
 
     private static String getISO8601FormattedLastSyncTime() {
-        DateFormat df = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm'Z'");
-        df.setTimeZone(TimeZone.getTimeZone("UTC"));
-        return df.format(SyncDetailCalculations.getLastSyncTime());
+        long lastSyncTime = SyncDetailCalculations.getLastSyncTime();
+        if (lastSyncTime == 0) {
+            return "";
+        } else {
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/heartbeat/HeartbeatRequester.java
```diff
@@ -116,10 +118,16 @@ protected void requestHeartbeat() {
...
         params.put(UNSENT_FORMS_PARAM, "" + StorageUtils.getNumUnsentForms());
-        params.put(LAST_SYNC_TIME_PARAM, new Date(SyncDetailCalculations.getLastSyncTime()).toString());
+        params.put(LAST_SYNC_TIME_PARAM, getISO8601FormattedLastSyncTime());
         return params;
     }
 
+    private static String getISO8601FormattedLastSyncTime() {
+        DateFormat df = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm'Z'");
+        df.setTimeZone(TimeZone.getTimeZone("UTC"));
+        return df.format(SyncDetailCalculations.getLastSyncTime());
```
wpride: is `getLastSyncTime()` guaranteed to return non-null?

amstone326: `getLastSyncTime()` returns a `long` so it is guaranteed to be non-null, but you're probably right that I should be sending something other than 1/1/1970 if we haven't recorded one yet (although theoretically that should never happen anymore).

@snopoke Is it ok for me to send something like "Never" if we don't have a last sync time recorded?

snopoke:
An empty string or leaving the param out completely would be better.

amstone326:
ok @snopoke I went with sending a blank string since that seems like it will make parsing on the other end slightly smoother than leaving a param out entirely

PR #1897, opened 2017-11-27
Prevent NPE stemming from onRetainCustomNonConfigurationInstance()

amstone326:
I believe this will address https://fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/59db367fbe077a4dccfd06ba?time=last-ninety-days, which I think is just a timing issue in form entry, where if a rotation of the device triggers a call to `onRetainCustomNonConfigurationInstance()` after form loading completes, but before the first view is actually rendered (a very short but presumably real window of time), then we would try to call `saveAnswersForCurrentScreen()` with a null `questionsView`. I wasn't able to replicate the crash based on this guess, but I think it's worth putting in this fix and seeing if that addresses it in the next release before digging much deeper.

### DIFF
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -632,7 +632,7 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
             return mSaveToDiskTask;
 
         // mFormEntryController is static so we don't need to pass it.
-        if (mFormController != null && currentPromptIsQuestion()) {
+        if (mFormController != null && currentPromptIsQuestion() && uiController.questionsView != null) {
             saveAnswersForCurrentScreen(FormEntryConstants.DO_NOT_EVALUATE_CONSTRAINTS);
         }
         return null;


app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -632,7 +632,7 @@ public Object onRetainCustomNonConfigurationInstance() {
             return mSaveToDiskTask;
 
         // mFormEntryController is static so we don't need to pass it.
-        if (mFormController != null && currentPromptIsQuestion()) {
+        if (mFormController != null && currentPromptIsQuestion() && uiController.questionsView != null) {
             saveAnswersForCurrentScreen(FormEntryConstants.DO_NOT_EVALUATE_CONSTRAINTS);
```
wpride: Hmm, isn't this call pretty important? Is it ok for this not to be called, or can we be sure that this will be called again once `uiController.questionsView` isn't null?

amstone326: My reasoning was that if `questionsView` is null, then there is no current screen with answers that needs to be saved

wpride: Ah OK understood, that seems totally sound.

PR #1899, opened 2017-11-27
Add 'data' evaluation as XPath in IntentCallout

wpride:
Small feature for Ismaïla and UNCDF. Allows the `data` field in `IntentCallouts` to be evaluated as an XPath. They need this to preload phone numbers for making calls.

Currently there's no way to set the `data` field in an `<intent>` block in HQ besides editing the Form XML directly so I don't think that there're likely to be existing users. However, this should continue to work gracefully if anyone was using this as a constant field. 

Not needed until February so can wait for the next release.


amstone326:
@wpride Won't this break all existing usages of intent callouts in existing apps, since they've been configured to pass a string without quotes and expect it to be treated as a string? We've had a very similar problem to this before.

amstone326:
@wpride was the last commit meant to address my concern? If so I'm not following how it does

ctsims:
==Aliza I'm not following how this won't break existing callouts

Previously if you had 
```
<data>FOO</data>
```
that would have been parsed as a string and forwarded along.

Now you'd get a parse error, since that's not a string.

wpride:
@ctsims @amstone326 Honestly not sure I'm still understanding your concern exactly, but added four tests demonstrating why I believe this is still safe. These cover:

1. Intent with XPath `data` attribute
2. Intent with no `data` attribute
3. Intent with empty ("") `data` attribute
4. Intent with hard coded `data` attribute

These intents still parse and behave fine. Am I missing something?

amstone326:
(offlined briefly with Will but there are failing tests so I think that should help clarify)

amstone326:
@wpride Can this be closed at this point or is it still needed by UNCDF?

wpride:

app/unit-tests/src/org/commcare/android/tests/formentry/IntentCalloutTests.java
```diff
@@ -0,0 +1,117 @@
...
+                "jr://resource/commcare-apps/phone_call/profile.ccpr",
+                "test", "123");
+        ShadowEnvironment.setExternalStorageState(Environment.MEDIA_MOUNTED);
+    }
+
+    /**
+     * Test different behaviors for possibly grouped intent callout views
+     */
+    @Test
+    public void testIntentCalloutWithData() {
```
amstone326: Can you rename this test to make it more obvious that this is the one testing the XPath format?

app/unit-tests/resources/commcare-apps/phone_call/default/app_strings.txt
```diff
@@ -0,0 +1,26 @@
+app.display.name=1Scratchwork
```
amstone326: Not positive but I think you can get rid of the app_strings resources and have the tests still work fine

app/unit-tests/resources/commcare-apps/phone_call/media_suite.xml
```diff
@@ -0,0 +1,2 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
amstone326: Think you can also get rid of this

wpride:
@amstone326 thanks for the feedback, updated. will add docs

PR #1900, opened 2017-11-28
Include graph title in graph logs

amstone326:
Adding on Jeremy/Rashmi's request.


amstone326:
@orangejenny Hm, my concerns with using detail ID would be a) it's not very descriptive, and b) they can change if modules get reordered. But if you think title is likely to be missing pretty often, then maybe that's still preferable?

orangejenny:
Ugh, I forgot about module reordering. I do still think it'd be good to include detail id, though. Depending solely on title that forces Rashmi/Jeremy to display a title on the graph, which they might want to avoid (or to experiment with).

amstone326:
@orangejenny Ok this is ready again. Didn't implement the detail id logging for now, for the reasons we discussed

PR #1901, opened 2017-11-29
Transactional Recoveries

wpride:
Fixes the issue in https://manage.dimagi.com/default.asp?265905. Specifically, crashes during recovery restore parse could result in the case database being left empty. Accomplish this by doing the wipe and restore parse in one transaction that we simply do not commit unless we fully succeed. A couple supporting refactors.

I tested this by throwing a `RuntimeException` right before the `readInput` call; previously this resulted in an empty user db, now this is successfully rolled back.

I think we should get this into 2.41.

### DIFF
--- a/app/src/org/commcare/models/database/user/models/EntityStorageCache.java
+++ b/app/src/org/commcare/models/database/user/models/EntityStorageCache.java
@@ -22,7 +22,7 @@ import java.util.List;
  */
 public class EntityStorageCache {
     private static final String TAG = EntityStorageCache.class.getSimpleName();
-    private static final String TABLE_NAME = "entity_cache";
+    public static final String TABLE_NAME = "entity_cache";
 
     private static final String COL_CACHE_NAME = "cache_name";
     private static final String COL_ENTITY_KEY = "entity_key";
@@ -133,11 +133,15 @@ public class EntityStorageCache {
         }
     }
 
+    public static void setCachedWipedPref() {
+        String uuid = CommCareApplication.instance().getSession().getLoggedInUser().getUniqueId();
+        setEntityCacheWipedPref(uuid, CommCareApplication.instance().getCurrentApp().getAppRecord().getVersionNumber());
+    }
+
     public static void tryWipeCache() {
         SQLiteDatabase userDb = CommCareApplication.instance().getUserDbHandle();
         SqlStorage.wipeTable(userDb, TABLE_NAME);
-        String uuid = CommCareApplication.instance().getSession().getLoggedInUser().getUniqueId();
-        setEntityCacheWipedPref(uuid, CommCareApplication.instance().getCurrentApp().getAppRecord().getVersionNumber());
+        setCachedWipedPref();
     }
 
     public static void setEntityCacheWipedPref(String username, int version) {
--- a/app/src/org/commcare/models/database/user/models/EntityStorageCache.java
+++ b/app/src/org/commcare/models/database/user/models/EntityStorageCache.java
@@ -133,20 +133,17 @@ public class EntityStorageCache {
         }
     }
 
-    public static void setCachedWipedPref() {
-        String uuid = CommCareApplication.instance().getSession().getLoggedInUser().getUniqueId();
-        setEntityCacheWipedPref(uuid, CommCareApplication.instance().getCurrentApp().getAppRecord().getVersionNumber());
-    }
-
     public static void tryWipeCache() {
         SQLiteDatabase userDb = CommCareApplication.instance().getUserDbHandle();
         SqlStorage.wipeTable(userDb, TABLE_NAME);
-        setCachedWipedPref();
+        setEntityCacheWipedPref();
     }
 
-    public static void setEntityCacheWipedPref(String username, int version) {
+    public static void setEntityCacheWipedPref() {
+        String uuid = CommCareApplication.instance().getSession().getLoggedInUser().getUniqueId();
+        int versionNumber = CommCareApplication.instance().getCurrentApp().getAppRecord().getVersionNumber();
         CommCareApplication.instance().getCurrentApp().getAppPreferences().edit()
-                .putInt(username + "_" + ENTITY_CACHE_WIPED_PREF_SUFFIX, version).apply();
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -531,41 +531,36 @@ private static void recordSuccessfulSyncTime(String username) {
...
-            //Get new data
-            String syncToken = readInput(cache.retrieveCache(), factory);
+            String syncToken = readInputWithoutCommit(cache.retrieveCache(), factory);
             updateUserSyncToken(syncToken);
             Logger.log(LogTypes.TYPE_USER, "Sync Recovery Successful");
+            userDb.setTransactionSuccessful();
             return new Pair<>(PROGRESS_DONE, "");
-        } catch (ActionableInvalidStructureException e) {
-            e.printStackTrace();
-            failureReason = e.getLocalizedMessage();
```
wpride: Condense these two `catch` blocks; the only difference was `getLocalizedMessage()` vs `getMessage()`, but `getLocalizedMessage()` defaults to calling `getMessage()` unless its been overridden, in which case presumably we want to use the localized version anyways

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -531,41 +531,35 @@ private static void recordSuccessfulSyncTime(String username) {
...
         return new Pair<>(PROGRESS_RECOVERY_FAIL_BAD, failureReason);
     }
 
-    private void wipeStorageForFourTwelveSync() {
-        CommCareApplication.instance().getUserStorage(ACase.STORAGE_KEY, ACase.class).removeAll();
-        new AndroidCaseIndexTable().wipeTable();
-        CommCareApplication.instance().getUserStorage(Ledger.STORAGE_KEY, Ledger.class).removeAll();
-        EntityStorageCache.tryWipeCache();
+    private void wipeStorageForFourTwelveSync(SQLiteDatabase userDb) {
+        userDb.beginTransaction();
```
amstone326: I know it's functionally equivalent but can you move this call to the line before `wipeStorageForFourTwelveSync()` is called instead? It just improves the readability of clearly seeing the transaction start and end in a linear manner

app/src/org/commcare/models/database/user/models/EntityStorageCache.java
```diff
@@ -133,11 +133,15 @@ public static int getSortFieldIdFromCacheKey(String detailId, String cacheKey) {
         }
     }
 
+    public static void setCachedWipedPref() {
```
amstone326: I think it would be a little cleaner if this was an overloaded implementation of `setEntityCacheWipedPref()` that took no arguments (and then can also move it to be next to that method below)

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -531,41 +531,35 @@ private static void recordSuccessfulSyncTime(String username) {
...
             Logger.log(LogTypes.TYPE_USER, "Sync Recovery Successful");
+            userDb.setTransactionSuccessful();
             return new Pair<>(PROGRESS_DONE, "");
-        } catch (ActionableInvalidStructureException e) {
-            e.printStackTrace();
-            failureReason = e.getLocalizedMessage();
         } catch (InvalidStructureException | XmlPullParserException
-                | UnfullfilledRequirementsException | SessionUnavailableException | IOException e) {
-            e.printStackTrace();
-            failureReason = e.getMessage();
```
amstone326: You can't delete the setting of this variable because it gets returned at the end of the method

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -531,41 +531,35 @@ private static void recordSuccessfulSyncTime(String username) {
...
+                | UnfullfilledRequirementsException | SessionUnavailableException
+                | IOException e) {
+            Logger.exception("Sync recovery failed|" + e.getLocalizedMessage(), e);
         } finally {
             //destroy temp file
```
amstone326: can you move this comment to be back above the `cache.release()` line, since that's what it describes?

amstone326:
@wpride What made you pick the internal-or-flagged category? I would have thought of this as invisible since it's just a bug fix without any UI component to it

wpride:
@amstone326 There was a discussion about these flags w/r/t bug fixes on an email thread today, quoting Sheel "if there is a user facing change, even a bug fix, I'd prefer that it is tagged all-users." - I was thinking this would be good to flag for our FMs who might've encountered some of the symptoms of this. Could all see it being invisible though

app/src/org/commcare/tasks/DataPullTask.java
wpride: Ah, nice catch. Can actually move the return in here though, I think its cleaner

app/src/org/commcare/tasks/DataPullTask.java
wpride: Nice catch, thanks

app/src/org/commcare/tasks/DataPullTask.java
wpride: Yep, that's nicer

app/src/org/commcare/models/database/user/models/EntityStorageCache.java
wpride: Looks like the method with arguments was only being used by the method without, so just consolidated

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -531,41 +531,34 @@ private static void recordSuccessfulSyncTime(String username) {
         //this is the temporary implementation of everything past this point
 
         //Wipe storage
-        //TODO: move table instead. Should be straightforward with sandboxed db's
-        wipeStorageForFourTwelveSync();
+        SQLiteDatabase userDb = CommCareApplication.instance().getUserDbHandle();
+        wipeStorageForFourTwelveSync(userDb);
+        userDb.beginTransaction();
```
amstone326: this needs to come before the call to wipeStorage, not after

wpride: 😳

amstone326:
@wpride FYI merge conflicts

PR #1904, opened 2017-12-01
Shows a better error message instead of a NPE

shubham1g5:
Trying to display a better error on getting a 500 from Server during app update. Earlier app was failing with this error message - 
"Attempt to invoke virtual method 'java.io/InputStream okhttp3.ResponseBody.byteStream()' on a null object reference".

### DIFF
--- a/app/src/org/commcare/engine/references/JavaHttpReference.java
+++ b/app/src/org/commcare/engine/references/JavaHttpReference.java
@@ -3,10 +3,14 @@ package org.commcare.engine.references;
 import org.commcare.interfaces.CommcareRequestEndpoints;
 import org.commcare.network.CommcareRequestGenerator;
 import org.javarosa.core.reference.Reference;
+import org.javarosa.core.services.locale.Localization;
 
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+
+import okhttp3.ResponseBody;
+import retrofit2.Response;
 
 /**
  * @author ctsims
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/engine/references/JavaHttpReference.java
```diff
@@ -35,7 +39,12 @@ public OutputStream getOutputStream() throws IOException {
 
     @Override
     public InputStream getStream() throws IOException {
-        return generator.simpleGet(uri).body().byteStream();
+        Response<ResponseBody> response = generator.simpleGet(uri);
+        if (response.isSuccessful()) {
+            return response.body().byteStream();
+        } else {
+            throw new RuntimeException(Localization.get("install.fail.error", Integer.toString(response.code())));
```
wpride: Granted that we expect HTTP errors to happen fairly regularly is throwing a `RuntimeException` the best behavior? Can we throw this as an `IOException` and catch that ? 

Also, won't this crash the application?

shubham1g5: We are catching the generic Exception [here](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/tasks/UpdateTask.java#L125) so it doesn't crash the app though can definitely wrap it as `IOException` if it's more clean.

wpride: Ah gotcha, nice. Yeah I think adding a block for `IOException` there handling this case would be preferable, but could be just my taste

shubham1g5:

PR #1910, opened 2017-12-13
Add '.' in MediaWidget

wpride:
Cause of the bug here https://manage.dimagi.com/default.asp?267070

For certain URIs (seems exclusive to 7.0 and certain media apps) we would set the path here, which resulted in the bug exposed here and thus the media not being recognized as valid on submission

### DIFF
--- a/app/src/org/commcare/views/widgets/MediaWidget.java
+++ b/app/src/org/commcare/views/widgets/MediaWidget.java
@@ -210,7 +210,7 @@ public abstract class MediaWidget extends QuestionWidget {
             newMedia = new File(destMediaPath);
         } else {
             recordedFileName = FileUtil.getFileName(binaryPath);
-            destMediaPath = mInstanceFolder + System.currentTimeMillis() + customFileTag + FileUtil.getExtension(binaryPath);
+            destMediaPath = mInstanceFolder + System.currentTimeMillis() + customFileTag + "." + FileUtil.getExtension(binaryPath);
 
             // Copy to destMediaPath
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/views/widgets/MediaWidget.java
```diff
@@ -210,7 +210,7 @@ public void setBinaryData(Object binaryuri) {
             newMedia = new File(destMediaPath);
         } else {
             recordedFileName = FileUtil.getFileName(binaryPath);
-            destMediaPath = mInstanceFolder + System.currentTimeMillis() + customFileTag + FileUtil.getExtension(binaryPath);
+            destMediaPath = mInstanceFolder + System.currentTimeMillis() + customFileTag + "." + FileUtil.getExtension(binaryPath);
```
ctsims: @wpride FileUtil can return a blank extension if the path doesn't have one, which could result in files like
```
something.
```

We should make sure we only add the dot if the extension is getting added to a non-blank value.

wpride: Thanks Clayton, nice catch. Seems like we're a bit fast and loose with the path and extension handling and much of this code could be centralize.

My fix implements the same assertion as [here](https://github.com/dimagi/commcare-android/blob/d270afa3b30b10a92c0c73a9bffecad48aeeb54c/app/src/org/commcare/activities/FormEntryActivity.java#L356-L356) - ensuring that the media type is one that we can actually upload to the server

PR #1917, opened 2017-12-15
Add meta instanceId to log submissions to prevent duplicates

amstone326:
With help from @emord I confirmed that HQ does treat device log xforms the same way as normal forms in that it checks for a unique id included in a `<meta>` block on the form. Mobile was previously not including any `<meta>` block in device log submissions, so adding one now will prevent duplicate logs from being included in device reports on HQ in the future. 

I was also able to confirm this theory via manual testing by observing that duplicate logs would previously show up on HQ if we prevented the deletion of a DeviceReportRecord after it was submitted, and observing that this PR fixes that behavior.

### DIFF
--- a/app/src/org/commcare/models/database/user/UserDatabaseUpgrader.java
+++ b/app/src/org/commcare/models/database/user/UserDatabaseUpgrader.java
@@ -8,6 +8,8 @@ import net.sqlcipher.database.SQLiteDatabase;
 import org.commcare.CommCareApplication;
 import org.commcare.android.database.user.models.FormRecordV2;
 import org.commcare.android.database.user.models.FormRecordV3;
+import org.commcare.android.javarosa.DeviceReportRecord;
+import org.commcare.android.javarosa.DeviceReportRecordV1;
 import org.commcare.android.logging.ForceCloseLogEntry;
 import org.commcare.android.javarosa.AndroidLogEntry;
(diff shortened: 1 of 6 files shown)


app/src/org/commcare/models/database/user/UserDatabaseUpgrader.java
```diff
@@ -580,6 +586,20 @@ private boolean upgradeTwentyOneTwentyTwo(SQLiteDatabase db) {
         }
     }
 
+    private boolean upgradeTwentyTwoTwentyThree(SQLiteDatabase db) {
+        db.beginTransaction();
+        try {
+            SqlStorage<DeviceReportRecord> existingStorage =
+                    new SqlStorage<>(DeviceReportRecord.STORAGE_KEY, DeviceReportRecordV1.class,
+                            new ConcreteAndroidDbHelper(c, db));
+            updateModels(existingStorage);
```
ctsims: @amstone326 The result of this is that the models have a uuid but it's empty, right? I'm a bit surprised that doesn't cause an issue, and it might be prudent to ensure that they get some value in the field if for no other reason than I don't think it's clear from the model that the field can be (blank? null?)

amstone326:
@ctsims Oh wow, I hadn't fully registered that it worked that way. Going to close this in favor of https://github.com/dimagi/commcare-android/pull/1923 since it's a much simpler change in that case

PR #1918, opened 2017-12-15
Markdown of constraint text

wpride:
Addresses https://manage.dimagi.com/default.asp?266984

The only downside of this approach is if people have constraint text that uses markdown syntax but they do not want it parsed as markdown (for example, they just want asterisks around a word) then this will turn that into mark down.

The alternative is a heaver touch refactor that checks whether there's a markdown form of this text (and we'd have to build that setup mechanism in HQ too). I'm not too worried about the downside here but open for opinion


amstone326:
@wpride What's the status of this?

wpride:
@amstone326 Given that this will be harder than I originally thought I think this should go to roadmap to decide how much we care and resourcing

PR #1926, opened 2018-01-09
Instantiate BarcodeWidget intent for quick appearance

wpride:
addresses https://manage.dimagi.com/default.asp?268210

Looks like the [refactor](https://github.com/dimagi/commcare-android/commit/f1f8204bc51b85947c51bb00272adcd586967aa2#diff-42be99734c63ffc769073deecabd62ff) to make the `BarcodeWidget` use the ZXing library postponed the instantiation of the `intent` class variable (see comment on line 32) such that this is still null when the `quick` appearance attribute is used. Unfortunate edge case.


amstone326:
Ah shoot sorry. Nice fix, thanks

PR #1927, opened 2018-01-10
Send users to the proper location settings page for the situation

amstone326:
@Ben-Talbot pointed out that when location services are missing, we were previously not always sending users to the proper settings page for them to fix it. This is because there are 2 different situations that can cause location services to be unavailable in CommCare: one is if location services are turned off globally for the device, and one is if they have been disabled (or never granted in the first place) for CommCare specifically. The settings pages on which the user can toggle these 2 things are different, so we should detect which situation the user is in and send them to the proper screen.

### DIFF
--- a/app/src/org/commcare/utils/GeoUtils.java
+++ b/app/src/org/commcare/utils/GeoUtils.java
@@ -149,12 +149,16 @@ public class GeoUtils {
         boolean gpsEnabled = false, networkEnabled = false;
         try {
             gpsEnabled = lm.isProviderEnabled(LocationManager.GPS_PROVIDER);
-        } catch (Exception ex){
+        } catch (Exception ex) {
+            // Prior to API level 21, this will throw a SecurityException if the location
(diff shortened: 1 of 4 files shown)


app/src/org/commcare/utils/GeoUtils.java
```diff
@@ -131,6 +133,32 @@ private static StandardAlertDialog setupAlertFactory(Context context,
...
+            intent = new Intent(android.provider.Settings.ACTION_LOCATION_SOURCE_SETTINGS);
+        }
+        context.startActivity(intent);
+    }
+
+    private static boolean locationServicesEnabledGlobally(LocationManager lm) {
+        boolean gpsEnabled = false, networkEnabled = false;
+        try {
+            gpsEnabled = lm.isProviderEnabled(LocationManager.GPS_PROVIDER);
+        } catch (Exception ex){
```
wpride: A comment explaining why an exception is expected/ignorable here would be useful

PR #1930, opened 2018-01-12
Don't use dynamic action id intent filters that will be set/called from outside CommCare

amstone326:
https://manage.dimagi.com/default.asp?268446 made me realize that when we added the paradigm of using the `${applicationId}` placeholder in `AndroidManifest`, we went overboard with the switch. There are some places where we don't want to make the action ID for an intent filter dynamic, because they're referenced by some place outside of CommCare that has no notion of switching that request based on which flavor of CommCare it's calling out to. This is definitely true of `TemplatePrinterActivity` because HQ injects the intent ID into the XForm, so it has to be a static value. Note that this means printing has never worked on LTS.

@wpride and @ctsims Aside from printing, I've changed this in all of the other places that I _think_ it should be changed. I confirmed that for all of these, there is nothing _within_ CommCare that references them via the current package ID (and therefore requires them to use the dynamic filter ID). (And actually the `ccc` script is still relying on the static `org.commcare.dalvik` filter ID for all of the `DebugControlsReceiver` actions). But I'm not sure if there is anything set up externally to rely upon this dynamic system. An example of what I mean would be that I have already changed the CommCare Toolkit to allow you to change which CommCare package ID you're targetting (https://github.com/dimagi/commcare-developer-toolkit/blob/master/app/src/main/java/dalvik/commcare/org/commcaretoolkit/activities/RefreshToLatestBuildActivity.java#L31), but I can easily change that back. Can you think of any other issues that would be caused by any of these changes?

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -387,7 +387,7 @@
             android:name="org.commcare.print.TemplatePrinterActivity"
             android:theme="@style/Theme.Dialog.NoTitle">
             <intent-filter>
-                <action android:name="${applicationId}.action.PRINT"/>
+                <action android:name="org.commcare.dalvik.action.PRINT"/>
                 <category android:name="android.intent.category.DEFAULT"/>
             </intent-filter>
         </activity>
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -140,7 +140,7 @@
         </activity>
         <activity android:name="org.commcare.activities.DotsEntryActivity">
             <intent-filter>
-                <action android:name="${applicationId}.action.DotsEntry"/>
+                <action android:name="org.commcare.dalvik.action.DotsEntry"/>
 
                 <category android:name="android.intent.category.DEFAULT"/>
             </intent-filter>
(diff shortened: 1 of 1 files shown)


amstone326:
@wpride No actually the [version of the `ccc` script](https://github.com/dimagi/commcare-ui-tests/blob/master/scripts/ccc) that lives in the `commcare-ui-tests` repo uses the `org.commcare.dalvik` ID for all its commands, which means that currently if we had tried running the UI tests on an .apk with a different package ID, they wouldn't have worked. So this will fix that potential issue (although apparently it has yet to come up as an actual problem).

amstone326:
@ctsims Going to wait for your approval on this since it definitely falls in the bucket of "could affect obscure use cases that only you know about"

shubham1g5:
A bit late on this but would support this change. An external app should not care what flavour of Commcare is installed on the device and should be able to choose between different flavours via Intent Chooser if multiple variants of Commcare are installed on the same device. Speaking of which I am curious why do we allow multiple Commcare flavours to be installed on same device anyway ? Is this because it's convinient to devs or do organisations actually want to run different apps on different flavours 😮

amstone326:
@ctsims Bumping this, see my comment above re:waiting on you approval

PR #1933, opened 2018-01-18
Answer questions and update cues after showNext

wpride:
Fixes https://manage.dimagi.com/default.asp?268210#edit_1_268210

After getting the next view for questions and groups, save answers (needed to catch the Android-level default value of the DateWidget) and update the navigation bar.

This is a pretty heavy-handed solution because when the DateWidget is created we're so far up the chain from the FormDef/TreeElement level that updating that node really isn't tenable from the Widget and this code path (`saveAnswersForCurrentScreen`) is the only extant one that allows you to set low-level values from Android. 

Hold off on merge because I have an idea for how this could be addressed on the HQ side.


wpride:

PR #1934, opened 2018-01-18
Turn off lint for gradle build

wpride:
As suggested in https://android.jlelse.eu/speeding-up-gradle-builds-619c442113cb


shubham1g5:
I am not really sure if turning off lint permanently is a good idea. Specially since we are supporting till api level 9, it's pretty easy to use some newer APIs not meant for older Android versions and not get warned about it in case we have lint disabled.

wpride:
@shubham1g5 you are the man!

PR #1935, opened 2018-01-18
Revert "gradle.properties changes for build speed"

amstone326:
Reverts dimagi/commcare-android#1932


shubham1g5:
Why revert 😢

amstone326:
@shubham1g5 It was causing the build on jenkins to fail so I need to figure out why

shubham1g5:
It might have been because of some gradle cache.. since the PR builder job was building successfully I think. Deleting the workspace and rebuilding might resolve the issue.

amstone326:
Good thought, I'll give that a shot

PR #1937, opened 2018-01-23
Corrects Graph legend colors

shubham1g5:
Case: https://manage.dimagi.com/default.asp?268818#1448115

Product Note: Fixes a bug where legend rectangle color won't match the respective bar's color in mobile bar charts.


wpride:
nice!

app/assets/graphing/graph.max.js
```diff
@@ -170,6 +170,7 @@ var intervalID = setInterval(function() {
                 if (legend.length) {
                     legend = legend[0];
                     legend.style.opacity = data.lineOpacities[yID];
+                    legend.style.stroke =  data.barColors[yID][0];
```
orangejenny: [This](https://github.com/dimagi/commcare-android/blob/master/app/assets/graphing/graph.max.js#L147-L149) makes me think `barColors` isn't always defined and you should check for it.

app/assets/graphing/graph.min.js
```diff
@@ -1 +1 @@
...
...r c=d3.selectAll(".c3-areas-"+t+" path")[0][0];c&&(c.style.fill=data.areaColors[t],c.style.opacity=data.areaOpacities[t])}var d=0,s=setInterval(function(){d+=retryFrequency,window.Android?(clearInterval(s),Android.run()):d>maxDelay&&clearInterval(s)})},c3.generate(config)}catch(t){displayError(t)}}else delay>maxDelay&&clearInterval(intervalID)},retryFrequency);function applyPointShape(t,a){if("bar"!==type&&"bubble"!==type){var e=d3.selectAll(".c3-circles-"+t),i=e.selectAll("circle")[0];if(i)for(var r=0;r<i.length;r++)i[r].style.opacity=0,appendSymbol(e,i[r].cx.baseVal.value,i[r].cy.baseVal.value,a,i[r].style.fill)}}function applyLegendShape(t,a){if("none"!==a){var e=d3.selectAll(".c3-legend-item-"+t),i=e.selectAll("line");if(!i||!i.length)return;(i=i[0][0]).style.opacity=0,appendSymbol(e,i.x1.baseVal.value+5,i.y1.baseVal.value,a,i.style.stroke)}}function appendSymbol(t,a,e,i,r){"none"!==i&&t.append("path").attr("transform",function(t){return"translate("+a+", "+e+")"}).attr("class","symbol").attr("d",d3.svg.symbol().type(i).size(50)).style("fill",r).style("stroke",r)}function displayError(t){console.log(t);var a=document.getElementById("error");a.innerHTML=t,a.style.display="block"}
```
orangejenny: I'm surprised this file has changed so much. Verifying that you tested it? Also curious what you used to generate this file. It might be good to always use the same tool so we don't get non-meaningful changes due to differences in minifiers.

orangejenny:
Nice, thanks for taking this!

app/assets/graphing/graph.min.js
shubham1g5: I used a random online minifier. I tested this though agree that good to use the same tool to minimize this diff. Is there a standard tool that we use across Dimagi, would prefer to re-minify this using that.

orangejenny: I don't think there's a standard - on HQ we use Django's compressor, which isn't practical for the mobile team to use here. I've also just used random online minifiers - I think usually https://www.cleancss.com/css-minify/ but not for any particular reason.

Could you you pick your favorite online minifier and document it, probably [here](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/commcare/core/graph/model/GraphData.java#L101-L103)? And maybe also where that code is duplicated in [formplayer](https://github.com/dimagi/formplayer/blob/master/src/main/java/util/FormplayerGraphUtil.java#L44-L46). Then we'll have a standard.

amstone326:
@shubham1g5 Is there still more work to be done on this or can it be merged? I'm assuming this should get in before code freeze

shubham1g5:
@orangejenny  Thanks. Updated the PR with feedback and reminified the file using the tool you mentioned (Retested with new min files as well).

Smalls PRs for docs  - https://github.com/dimagi/commcare-core/pull/743 and https://github.com/dimagi/formplayer/pull/608

PR #1938, opened 2018-01-24
Updates sql-cipher

shubham1g5:
Case: https://manage.dimagi.com/default.asp?268970#1449115

Updates sql-cipher to include fix for https://github.com/sqlcipher/android-database-sqlcipher/issues/313

Product Note: Fixes a crash happening during some DB operations on Commcare caused by a bug in one of the supporting Library Commcare uses.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -58,7 +58,7 @@ dependencies {
     compile 'com.google.android.gms:play-services-analytics:10.0.1'
     compile 'joda-time:joda-time:2.9.4'
     compile 'net.sf.kxml:kxml2:2.3.0'
-    compile 'net.zetetic:android-database-sqlcipher:3.5.4@aar'
+    compile 'net.zetetic:android-database-sqlcipher:3.5.7@aar'
     compile 'org.apache.james:apache-mime4j:0.7.2'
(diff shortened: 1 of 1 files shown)


shubham1g5:

shubham1g5:
@amstone326 Now when I look back at more crash reports, it's not always true though [the crashes for HKI user](https://fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/59bab3f7be077a4dcca0087b/sessions/5A66BCB2010E00010A02A6115058EDA2_DNE_0_v2?) suggests that this was always happening when the DB migration from v20 to v21 was undergoing on main thread (May be related to large no of forms being migrated for that user).

amstone326:
@shubham1g5 I think it's risky from a liability standpoint with users to be overly specific with release/product notes if we're not quite sure on the scope, so would rather us just say something like "Fixes a bug where CommCare could crash on login for some users after updating CommCare versions, due to a bug in one of our supporting libraries"

shubham1g5:
@amstone326  Makes sense. Though since this crash can occur anywhere and not just on login, updated the note to this - "Fixes a crash happening during some DB operations on Commcare caused by a bug in one of the supporting Library Commcare uses."

PR #1940, opened 2018-01-31
Merge Form Provider and Insance Provider to App DB

shubham1g5:
Summary:
* Replaces `FormProvider` with `FormDefRecord`
* Merges `InstanceProvider` into `FormRecord`
* Refactors to replace all URIs dependent code to work on FormDefRecord sql ids and in process creates new versions for XFormAndroidInstaller and FormRecord
* Takes care of migration from old user/app DBs to their newer versions


ctsims:
@shubham1g5 Hey, Shubham. Maybe easiest for us to chat over voice, I think my main point of confusion is that it's still not obvious to me why we need to maintain FormRecord _and_ InstanceRecord as different models, rather than unifying them. Can you clarify that specific point at a high level?

app/src/org/commcare/android/database/app/models/FormDefRecord.java
```diff
@@ -0,0 +1,287 @@
...
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Vector;
+
+@Table(FormDefRecord.STORAGE_KEY)
+public class FormDefRecord extends Persisted {
+
+    public static final String STORAGE_KEY = "form_def";
+
+    public static final String META_DISPLAY_NAME = "displayName";
```
ctsims: Did we audit whether we need to retain all of these? I think many/most of these properties can actually be eliminated, but i could be wrong about that

app/src/org/commcare/models/database/app/AppDatabaseUpgrader.java
```diff
@@ -1,20 +1,31 @@
...
 import org.commcare.models.database.SqlStorage;
-import org.commcare.android.database.app.models.UserKeyRecord;
-import org.commcare.android.database.app.models.UserKeyRecordV1;
 import org.commcare.models.database.migration.FixtureSerializationMigration;
-import org.commcare.android.storage.framework.Persisted;
+import org.commcare.modern.database.TableBuilder;
+import org.commcare.provider.FormsProviderAPI;
+import org.commcare.provider.InstanceProviderAPI;
 import org.commcare.resources.model.Resource;
+import org.javarosa.core.util.externalizable.PrototypeFactory;
```
ctsims: I think that this file needs to have all of the upgradeXtoY methods from 1 to 8 audited to make sure their records use the old version of the classes. We need to load old Resources with the V1 installer for XForms, for example, right?

app/src/org/commcare/models/database/app/AppDatabaseUpgrader.java
```diff
@@ -200,6 +217,131 @@ private boolean upgradeSevenEight(SQLiteDatabase db) {
         }
     }
 
+    // Migrate records form FormProvider and InstanceProvider to new FormDefRecord and InstanceRecord respectively
+    private boolean upgradeEightNine(SQLiteDatabase db) {
+        db.beginTransaction();
+        android.database.Cursor cursor = null;
```
ctsims: Strictly speaking no lines of code should occur after db.begin until the try starts.

app/src/org/commcare/android/database/app/models/InstanceRecord.java
```diff
@@ -0,0 +1,448 @@
...
+    }
+
+    private static SqlStorage<InstanceRecord> getInstanceRecordStorage() {
+        if (sInstanceRecordStorage == null) {
+            sInstanceRecordStorage = CommCareApplication.instance().getAppStorage(InstanceRecord.class);
+        }
+        return sInstanceRecordStorage;
+    }
+
+    public static void setinstanceRecordStorage(SqlStorage<InstanceRecord> instanceRecordStorage) {
```
ctsims: I'm not OK with managing this state statically. I know it's annoying, but I think we need to refactor to move all of the static functionality in this class to being called through some other pattern. 

It would be OK in my mind for it to be linked through the platform through some context object, but it has only ever caused a lot of headache/tech debt to allow static maintenance of the storage sandboxes, so we should figure out how to make this something that comes from the platform.

app/src/org/commcare/android/database/app/models/FormDefRecord.java
shubham1g5: Yeah, Some of these are actually not getting used right now. I think I was being overly cautious to not create unnecessary changes as part of this migration. Can remove these now though.

app/src/org/commcare/android/database/app/models/InstanceRecord.java
shubham1g5: will do, though I would be curious to know what kind of problems have we seen in past arising from storage sadboxes being singletons.

shubham1g5:
@ctsims I did not think that we would like to merge the two tables. I was confused myself in utility of having them as separate tables though assumed that we might want to keep the separation as it is even in the new world. Will work on merging them now!

ctsims:
@shubham1g5 The reason they were split was because for a while we were trying to assume we'd treat ODK as a separate service on the phone (when android looked like it was going to be able to work that way), so  we had to add our metadata somewhere else (which is where we made FormRecord). Now that we're scrapping the form entry modularity I think it just adds overhead. Thanks for the merge!

app/src/org/commcare/android/database/app/models/InstanceRecord.java
ctsims: Two big things that have come up a lot:

1) We already have a *lot* of lifecycle objects around the code (java lifecycle, app lifecycle, user lifecycle) and it's a lot of overhead to maintain registering and unregistering data from them.

2) In many cases we are on a track to being able to decouple static state from apps entirely, which would give us pretty incredible capabilities like the ability to update apps in the background without interfering with seated applications. Adding new statically managed holders (rather than coupling them into another lifecycle object) significantly complicates those efforts by making a full refactor necessary and requiring us to inject the new lifecycle object in lots of places (this came up a ton with things like ReferenceManager, and it's a pattern we want to move away from)

shubham1g5:
@ctsims Is there a philosophy behind what things go to UserDb vs what goes to AppDb. When I created InstanceRecord and FormDefRecord, they both seemed app specific to me and so I added them to AppDB but now when I see the FormRecord it's part of UserDb. So now I am confused where should I put the merged version of FormRecord and InstanceRecord!

app/src/org/commcare/android/database/app/models/InstanceRecord.java
shubham1g5: Thanks! Currently we have different versions of `getUserStorage` and `getAppStorage` in Application class [here](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/CommCareApplication.java#L556). I am thinking of using these same functions for FormDefRecord and InstanceRecord to replace the static counterparts. Does that sound right or are we looking for something alternate/better here ?

ctsims: That should be fine. Passing the storage as a constructor to this (and/or wrapping that operation in a static getter from the Application) is a fine solution, since that puts us in the same state.

As a best-practice pattern I try to only call the singleton application methods from lifecycle objects (like the Activity classes), because it makes it easier to manage the outcome if the lifecycle is in a bad state (say, the user has been logged out).

ctsims:
@shubham1g5 FormRecords (and InstanceRecords) are UserDB Objects. They are encrypted by the user's keys and are protected assets for the user storage sandbox.

FormDefRecords are part of the App sandbox because they are a shared resource that is accessed by any user who is logged into the app and stored unencrypted.

shubham1g5:
@ctsims  Thanks. Makes sense now!

app/src/org/commcare/android/database/app/models/FormDefRecord.java
shubham1g5: @ctsims Here is a list of all fields that are not getting used currently - 

Instance Provider/ Instance Record

* submissionUri
* date (last status change date)
* displaySubtext

Form Def 

* description
* submissionUri
* displaySubtext
* md5Hash
* date
* jrcacheFilePath
* language

Just want to make sure we are fine with removing all of these.

app/src/org/commcare/models/database/app/AppDatabaseUpgrader.java
shubham1g5: Good point though none of these methods seem to be loading Resources i.e. performing a read on Resource table!

shubham1g5:

shubham1g5:
@ctsims Ready to review again

app/src/org/commcare/android/database/app/models/FormDefRecord.java
```diff
@@ -0,0 +1,183 @@
...
+
+    public Integer getModelVersion() {
+        return mModelVersion;
+    }
+
+    public Integer getUiVersion() {
+        return mUiVersion;
+    }
+
+    public String getBase64RsaPublicKey() {
```
ctsims: Are these for the migration only? If so it would be useful to visible separate them out, or to use a subclass exposing these to allow them to be read only for the migration or some other technique to prevent confusion

app/src/org/commcare/android/database/user/models/FormRecord.java
```diff
@@ -73,6 +91,12 @@
      * object representation of to reference in the short-term
      */
     public static final String STATUS_JUST_DELETED = "just-deleted";
+    private static final String TAG = FormRecord.class.getName();
+
+    @StringDef({STATUS_UNSTARTED, STATUS_INCOMPLETE, STATUS_COMPLETE, STATUS_UNSENT, STATUS_SAVED, STATUS_QUARANTINED, STATUS_UNINDEXED, STATUS_JUST_DELETED})
```
ctsims: This is a really cool pattern, thanks for sharing it!

app/src/org/commcare/models/database/app/AppDatabaseUpgrader.java
```diff
@@ -200,6 +215,99 @@ private boolean upgradeSevenEight(SQLiteDatabase db) {
...
+                    FormDefRecord formDefRecord = new FormDefRecord(cursor);
+                    formDefRecord.save(formDefRecordStorage);
+                }
+            }
+        } finally {
+            safeCloseCursor(cursor);
+        }
+
+        // Delete migrated entries
+        context.getContentResolver().delete(FormsProviderAPI.FormsColumns.CONTENT_URI, null, null);
```
ctsims: So one quick point:

Errors in this operation can't be rolled back, and also aren't material to the future functionality of the app, since this CR will be ignored moving forward.

I think for those reasons that moving this op _outside_ of the v8->v9 Migrater makes a bit more sense. 

Right now, it's:

```
transaction[
1) add table
2) move data
3) delete old data
]
```
and if "delete old data" deletes _half_ of the data before failing, the first two ops will be undone, but the delete old data op _won't be_, which will put the phone in an irrecoverable state. If instead #3 was treated as a "best effort" op

```
transaction[
1) add table
2) move data
]
try{
3) delete old data
} catch(){
//oops... do nothing.
}
```
"Half of a failure" in step three doesn't result in an irrecoverable app.

app/src/org/commcare/models/database/user/UserDbUpgradeUtils.java
```diff
@@ -198,4 +196,54 @@ protected static void migrateFormRecordsToV3(Context c, SQLiteDatabase db) {
...
+                    oldRecord.getFormNamespace(),
+                    oldRecord.getAesKey(),
+                    oldRecord.getInstanceID(),
+                    oldRecord.lastModified(),
+                    oldRecord.getAppId());
+
+            // Merge Record's instance
+            String instanceUri = oldRecord.getInstanceURIString();
+            Cursor cursor = c.getContentResolver().query(Uri.parse(instanceUri), null, null, null, null);
+            if (cursor != null && cursor.getCount() > 0) {
```
ctsims: quick note that technically cursor operations should be wrapped in try{}closes

app/src/org/commcare/models/database/user/UserDbUpgradeUtils.java
```diff
@@ -198,4 +196,54 @@ protected static void migrateFormRecordsToV3(Context c, SQLiteDatabase db) {
...
+        db.execSQL(builder.getTableCreateString());
+
+        // Write to the new table
+        SqlStorage<FormRecord> newStorage = getFormRecordStorage(c, db, FormRecord.class);
+        for (FormRecord newRecord : newRecords) {
+            newStorage.write(newRecord);
+        }
+
+        // delete all instance provider entries
+        c.getContentResolver().delete(InstanceProviderAPI.InstanceColumns.CONTENT_URI, null, null);
```
ctsims: Same note as in the other migration that I think it's a better practice for us to track "success" for the transaction of this operation **before** this delete and not after, since it inherently can't be a part of the transaction

app/src/org/commcare/models/legacy/LegacyInstallUtils.java
```diff
@@ -231,38 +225,18 @@ public static void checkForLegacyInstall(Context c, SqlStorage<ApplicationRecord
         Logger.log(LogTypes.TYPE_MAINTENANCE, "Legacy| Files moved. Updating Handles");
 
         //We also need to tell the XForm Provider that any/all of its forms have been moved
-
-        ArrayList<Pair<Uri, String>> toReplace = new ArrayList<>();
```
ctsims: **Important Note**: -This code will no longer work after this change. This legacy installer should continue assuming that a form provider exists (using hard-coded strings or whatever else is retained)  **or** we should kill the legacy updater code.

app/src/org/commcare/android/database/app/models/FormDefRecord.java
shubham1g5: Reg: "last status change date": You are right. We had 2 date fields one in FormRecord and other in InstacneRecord. Now in the merged FormRecord I have retained the original FormRecord field which also gets used for form ordering. 

Reg: "private key" : There is no private key currently in FormDefRecord. So guessing we already removed it earlier. Though we do use the public key [here](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/utils/EncryptionUtils.java#L131)

app/src/org/commcare/android/database/app/models/FormDefRecord.java
```diff
@@ -0,0 +1,183 @@
...
+            mDisplayName = form.getName();
+        }
+
+        if (StringUtils.isEmpty(mFormMediaPath)) {
+            mFormMediaPath = getMediaPath(mFormFilePath);
+        }
+
+        formDefRecordStorage.write(this);
+
+        if (recordId == -1) {
```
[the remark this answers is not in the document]
shubham1g5: It can definitely return -1 which means the insertion eventually failed due to unknown reasons. I think we are safe though as there is not anything stateful that happens in SqlStorage.write().

app/src/org/commcare/android/database/app/models/FormDefRecord.java
shubham1g5: It's getting used in Encryption currently. Migration doesn't use any of these getters.

app/src/org/commcare/android/database/user/models/FormRecord.java
shubham1g5: Yeah. It also saves memory overhead caused by Enums as these annotations are compile time only.

app/src/org/commcare/models/database/app/AppDatabaseUpgrader.java
shubham1g5: Ahh never realised deletes are not atomic. I think one other way to do this is wrap the delete call [here](https://github.com/dimagi/commcare-android/blob/mergeFormInsanceProviders/app/src/org/commcare/provider/InstanceProvider.java#L163) in a transaction. What do you think ?

app/src/org/commcare/models/legacy/LegacyInstallUtils.java
shubham1g5: Yeah. Didn't realize this block of code runs before DB update. We can defnitely keep the Content Provider code if it's worth it. What all versions of CommCare does legacy actually represents ?

app/src/org/commcare/android/database/app/models/FormDefRecord.java
ctsims: I am 99% that EncryptedFormInformation stuff is only used by ODK. That class is only called while reading forms currently (and we've never written any that way), if you trace the code back, so I think we can remove it without concern.

app/src/org/commcare/android/database/app/models/FormDefRecord.java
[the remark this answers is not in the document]
ctsims: Cool, gotcha. Killing this thread then

app/src/org/commcare/android/database/app/models/FormDefRecord.java
ctsims: I think this goes along with my comment above. If this is part of the EncryptedFormInformation code, i don't think we use this

app/src/org/commcare/models/database/app/AppDatabaseUpgrader.java
ctsims: I *think* that would make everything here work reliably, as long as the transaction for this code is halted _immediately_ after the delete succeeds. I'm still a bit hesitant to introduce semantics where the ContentProvider.delete() call can succeed (in whole or in part) while the migrate transaction is still open, just as a matter of most-importance.

Like, if the line of code that says "commit the ContentProvider DB delete transaction" executes and then the phone battery dies, the next time the app spins up, our DB will roll back the migrate transaction and the content provider data will be gone. 

If we are super concerned with the ContentProvider delete occurring, I'd rather we separate those transactional semantics from the semantics of getting the data **into** this db. It's ok for the app to crash if the delete is failing (if that's a real bug that we should be catching and fixing), but I want to separate that condition from this transaction, since there's no way to use SQL transactions in a way which will Atomically link loading the new DB with the ContentProvider data, and deleting the data in the content provider DB. We'd need that to come from a separate source.

app/src/org/commcare/models/database/user/UserDbUpgradeUtils.java
ctsims: We don't need to catch any exceptions, we just need a try{}finally{} block where the cursor gets closed.

app/src/org/commcare/models/legacy/LegacyInstallUtils.java
ctsims: Confirmed over slack, but the Legacy db represents CommCare 2.3.0 and below, so we can move forward with the assumption that we are removing legacy migration code.

app/src/org/commcare/models/database/user/UserDbUpgradeUtils.java
shubham1g5: Gotcha!

app/src/org/commcare/android/database/app/models/FormDefRecord.java
shubham1g5: Cool. Going to remove the `EncryptionUtils.getEncryptedFormInformation` as well since that will always return null in case of no publick key

app/src/org/commcare/models/database/app/AppDatabaseUpgrader.java
shubham1g5: Hmm I see the concerns. Will make the changes accordingly.

amstone326:

shubham1g5:
@ctsims This should be now ready for another pass.

ctsims:
@shubham1g5 I think this covers everything I was concerned about other than one lingering comment above ("This seems like a bit of an odd time to fail on this to me. Can't we validate that this path isn't empty when it is being set, rather than when it is being used?").

shubham1g5:
@ctsims I made a mistake in incuding that change as part of a merge commit which I am guessing is why that change is not getting reflected here. But if you go [to this commit](https://github.com/dimagi/commcare-android/commit/0bc948ffde48425ce2b2f461b839865a6bd222cb) and search for `FormDefRecord.java` you will be able to see the required changes for pre-validating `filePath`.

ctsims:
Great, I'm all set, then. Thanks for going through the 3-4 rounds of random testing for multiple apps, multiple users, etc! Ready to merge once we've cut 2.34

shubham1g5:
Merging this now in anticipation to find early bugs!

wpride:
🎉 🎉 🎉 Congrats @shubham1g5 ! Looked like a slog

PR #1941, opened 2018-02-05
Adds Bypass aar from maven repo to include the .so files for 64 bit processors

shubham1g5:
Case: https://manage.dimagi.com/default.asp?269866#1454597
Crash: https://www.fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/5a7820de8cb3c2fa6356160f?time=last-thirty-days

Markdown code is currently crashing on the phones with 64 bit processors because we didn't have the required .so files for 64 bit architectures.  I have deleted the local aar file and jni files to use the remote aar from maven repo which contains the missing files.


wpride:
@shubham1g5 nicee

PR #1943, opened 2018-02-12
Fixes NPE on rotation for Update Prompts

shubham1g5:
Case: https://manage.dimagi.com/default.asp?270185
Crash: https://fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/5a300d4a8cb3c2fa63b92579/sessions/latest?build=69952000

On Rotation, before we could refresh the `updateToPrompt` in `onCreateSessionSafe ` `isBackEnabled`  gets called from `BreadcrumbBarFragment.configureSimpleNav` which results in this NPE.  I am now refreshing the `updateToPrompt` object again if it's null and have added the isNull check in onCreateSessionSafe to avoid duplicate initialization. 

Product Note: Fixes Commcare crash on rotation on the update prompt screen.

### DIFF
--- a/app/src/org/commcare/activities/PromptUpdateActivity.java
+++ b/app/src/org/commcare/activities/PromptUpdateActivity.java
@@ -36,7 +36,9 @@ public abstract class PromptUpdateActivity extends SessionAwareCommCareActivity 
     @Override
     public void onCreateSessionSafe(Bundle savedInstanceState) {
         super.onCreateSessionSafe(savedInstanceState);
-        refreshUpdateToPromptObject();
+        if (updateToPrompt == null) {
+            refreshUpdateToPromptObject();
+        }
         setupUI();
     }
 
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/activities/PromptUpdateActivity.java
```diff
@@ -36,7 +36,9 @@
     @Override
     public void onCreateSessionSafe(Bundle savedInstanceState) {
         super.onCreateSessionSafe(savedInstanceState);
-        refreshUpdateToPromptObject();
+        if (updateToPrompt == null) {
```
amstone326: this null check seems unnecessary to me because `updateToPrompt` will always be null here

amstone326: Ah nevermind I understand-- in the instance where `isBackEnabled` gets called before `onCreate`, it won't be null and we'd be doing the refresh twice. Sorry for slowness on the uptake..

PR #1945, opened 2018-02-13
Vary update prompt frequency

amstone326:
cross-request: https://github.com/dimagi/commcare-core/pull/748


app/src/org/commcare/heartbeat/UpdatePromptShowHistory.java
```diff
@@ -0,0 +1,89 @@
...
+     * that means we need to show the prompt once every 4 logins, which means that on any given
+     * login, we need to know about the prior 3 logins in order to decide whether or not to show it.
+     */
+    private boolean isFull() {
+        return showHistory.size() == UpdateToPrompt.getReducedShowFrequency()-1;
+    }
+
+    @Override
+    public void readExternal(DataInputStream in, PrototypeFactory pf) throws IOException, DeserializationException {
+        showHistory = (LinkedList<Boolean>)ExtUtil.read(in, new ExtWrapList(Boolean.class, LinkedList.class), pf);
```
ctsims: Since this is the only thing being written, I would actually probably move the read/write directly into the other object, and have it pass that to this object as a helper. Keeping more externalizable classes around adds a small amount of weight that doesn't seem like it's quite worth it to me.

amstone326:
@ctsims updated

ctsims:
Looks great, thanks for the update!

PR #1946, opened 2018-02-14
Improve UX when Installs/Updates Fail with a Notification

ctsims:
Previously users had to figure out how to load notifications from the top bar in order to see issues with install resource sets.

This allows them to click the message to view the notifications.

Old Behavior
screenshot-1518650956384

New Behavior
screenshot-1518714935578


app/assets/locales/android_translatable_strings.txt
```diff
@@ -218,6 +218,7 @@ cchq.case=Case
 select.placeholder.message=Please select a ${0}
 notification.for.details.wrapper=${0} (See notification for details)
 notification.for.details.setup.wrapper=${0} (Press "See More" or see notification for details)
```
shubham1g5: Not specifically related to this PR but this is not used anymore.

app/src/org/commcare/fragments/SelectInstallModeFragment.java
```diff
@@ -18,8 +18,10 @@
 
 import com.google.zxing.integration.android.IntentIntegrator;
 
+import org.commcare.CommCareNoficationManager;
```
shubham1g5: Unused ?

app/src/org/commcare/fragments/SelectInstallModeFragment.java
```diff
@@ -118,6 +120,13 @@ public void onClick(View v) {
         });
 
         mErrorMessageView = (TextView)view.findViewById(R.id.install_error_text);
+        mErrorMessageView.setClickable(true);
```
shubham1g5: Minor but If it's always clickable, should this go into layout xml file ?

shubham1g5:
@ctsims Shouldn't this behaviour get implemented for LoginActivity as well ?

orangejenny:
Picky ux feedback: can this use a button so it's more obviously a click target, without needing to read?

ctsims:
@orangejenny @shubham1g5 So much for light refactors 😛 ! 

Re-introduced rectangular buttons and made a button that launches things instead. I didn't re-implement for other views because it's a fair amount of work now, but I think it stands alone fine as an improvement.

orangejenny:
@ctsims New screenshot looks much nicer! Scope creep is fun, eh?

amstone326:
@ctsims This should get added to the roadmap to have the work finished off if you're not going to implement the change everywhere that it's relevant right now. It seems odd to me that we'd leave this fixed in only 1 place for too long.

ctsims:
@amstone326 

Implemented for Login
screenshot-1518732023421

and update

screenshot-1518741882928

Should be fine to include now.

app/src/org/commcare/activities/LoginActivityUIController.java
```diff
@@ -377,13 +405,13 @@ private void setStyleDefault() {
...
-            // don't hide error box when showing permission error
-            errorBox.setVisibility(View.GONE);
+            clearErrorMessage();
         }
     }
 
     protected void clearErrorMessage() {
-        errorBox.setVisibility(View.GONE);
+        errorContainer.setVisibility(View.GONE);
+        notificationButtonView.setVisibility(View.GONE);
```
shubham1g5: Shouldn't `errorContainer.setVisibility(View.GONE)` also takes care of `notificationButtonView` since it's a child of errorContainer (Not putting this for other use cases, but should get corrected for install and update as well if we do it here)

app/src/org/commcare/activities/LoginActivityUIController.java
```diff
@@ -347,16 +372,19 @@ protected LoginMode getLoginMode() {
...
         loginButton.setBackgroundColor(getResources().getColor(R.color.cc_attention_negative_bg));
         loginButton.setTextColor(getResources().getColor(R.color.cc_attention_negative_text));
 
-        errorBox.setVisibility(View.VISIBLE);
-        errorBox.setText(message);
+        errorContainer.setVisibility(View.VISIBLE);
+        errorTextView.setText(message);
+        if(showNotificationButton) {
+            notificationButtonView.setVisibility(View.VISIBLE);
+        }
```
shubham1g5: `notificationButtonView.setVisibility(showNotificationButton  ? View.VISIBLE : View.GONE)` seems more safe to me in case `setErrorMessageUI` gets called twice with different values for `showNotificationButton`

app/src/org/commcare/fragments/SelectInstallModeFragment.java
```diff
@@ -128,6 +145,11 @@ public void onClick(View v) {
         return view;
     }
 
+    private void onErrorViewButtonClicked() {
```
shubham1g5: Can we make it a public static and put it somewhere where all 3 use-cases utilize it, may be in MessageActivity itself.

app/src/org/commcare/activities/LoginActivityUIController.java
ctsims: For login specifically, the error container sometimes contains only an error message. When things pop back up, if we were previously displaying an error message and the button, the next time it may only contain an error message, so we need both to have their own logic.

app/src/org/commcare/activities/LoginActivityUIController.java
ctsims: Agreed, that would also simplify the other logic. I'll make this change

app/src/org/commcare/fragments/SelectInstallModeFragment.java
ctsims: Good call. Done

ctsims:
@shubham1g5 Made the changes you requested. Also made one final change which makes the button disappear if the notifications are gone, so the user isn't being directed to an empty screen.

Ready for review again.

amstone326:
Release Note: Improved UI/UX for users to view the error message when CommCare app updates or installs fail

PR #1947, opened 2018-02-15
Return RESULT_OK for external callouts

wpride:
Ran into this while dog fooding, we are always returning `RESULT_CANCELED` for external callouts into CommCare, even when we succeed. Conveniently we already have a boolean corresponding to whether the event succeeded or failing, so we should set the result based on that.


amstone326:
@wpride I'm not seeing how this code path is related to external callouts; can you explain the connection a bit more?

wpride:
@amstone326 If you look where the `wasExternal` flag is set:

https://github.com/dimagi/commcare-android/blob/ecb4f1009209da448f83ccfe43297587ecc859af/app/src/org/commcare/activities/HomeScreenBaseActivity.java#L164-L164

This signifies that we launched from an external application using the [form entry API](https://github.com/dimagi/commcare-android/wiki/Form-Entry-Workflow)

PR #1948, opened 2018-02-15
More Crashes

shubham1g5:
Adding some crashes that were getting left out to Crashlytics.


ctsims:
@shubham1g5 , just to be clear what we'd really like from our programmers is _fewer_ crashes, not more of them!!!!

😝

PR #1951, opened 2018-02-22
Updates landscape layout for update activity with recent changes

shubham1g5:
Crash: https://www.fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/5a8dad6d8cb3c2fa6387e5a0/sessions/5A8ED8FF00BC00012D0FE83CC2E3B097_DNE_0_v2?select_key=user

Regression from https://github.com/dimagi/commcare-android/pull/1946, we missed out on adding the new error view to landscape layout causing a NPE in landscape mode. 

Side Note: I stumbled on this while working on integration tests. This was appearing as the infamous `HTTPClient::KeepAliveDisconnected: (HTTPClient::KeepAliveDisconnected` error on device farm with no crash logs in logcat logs. So it seems like logcat provided by device farm is not always complete/accurate.

### DIFF
--- a/app/res/layout/update_activity.xml
+++ b/app/res/layout/update_activity.xml
@@ -1,11 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
-<ScrollView xmlns:SquareButtonWithText="http://schemas.android.com/apk/res-auto"
-    xmlns:android="http://schemas.android.com/apk/res/android"
-    android:layout_width="fill_parent"
-    android:layout_height="fill_parent"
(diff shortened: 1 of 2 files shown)


app/res/layout/update_activity.xml
```diff
@@ -1,11 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
```
shubham1g5: All changes in this file are due to reformatting

amstone326:
good catch, thanks Shubham

ctsims:
Oh, wow, totally missed this. thanks so much for the quick fix, and sorry to have missed this entirely!

PR #1955, opened 2018-02-26
Make properties of type String so values from HQ are compatible

amstone326:


wpride:
One question: do we need to worry about migrations at all? IE if someone has saved a Preference as an `int`, updates, then the code tries to pull it with `Preferences.getString` will that fail?

amstone326:
@wpride Oops, sorry about that. 

As for your question, the existence of these properties has not been released yet; they were only first added in a PR targeting 2.43: https://github.com/dimagi/commcare-android/pull/1945. It's also not possible for users to specify custom properties in HQ as integers; they are automatically sent as strings, which is why I needed to make this change.

amstone326:

amstone326:

PR #1957, opened 2018-03-07
 Adds mechanism to set image path for Image widget through a broadcast

shubham1g5:
Adds ability to upload an image in ImageWidget using a broadcast with 'file_path' as an extra with it. 

For: https://github.com/dimagi/commcare-ui-tests/pull/130


app/src/org/commcare/activities/components/ImageCaptureProcessing.java
```diff
@@ -50,19 +53,20 @@ private static File moveAndScaleImage(File originalImage, boolean shouldScale,
...
         }
-
         if (!savedScaledImage) {
             // If we didn't create a scaled image and save it to the final path, then relocate the
             // original image from the temp filepath to our final path
             File finalFile = new File(finalFilePath);
 
-            if (!originalImage.renameTo(finalFile)) {
+            try {
+                FileUtil.copyFile(originalImage, finalFile);
```
amstone326: @shubham1g5 was there any specific reason for this change (and the identical one below)?

shubham1g5: @amstone326 the AWS device was throwing an IOException when using the 'renameTo' function though I was not able to reproduce that on my device.

PR #1958, opened 2018-03-07
Revert "Sort completed forms in form record list by submission ordering number"

amstone326:
Reverts dimagi/commcare-android#1939. I realized thanks to https://manage.dimagi.com/default.asp?271707 that this change makes no sense because the submission ordering number resets once forms are submitted successfully, so it can't be relied upon as a global comparator for already-sent forms. Sorting based on ordering number is theoretically still valid for _unsent_ forms, but I don't think it makes sense to make the sort behavior for that one filter different, so let's just revert to the way things were. 

Waiting for input from @ctsims to decide whether this is also hotfix-worthy.

**Product Note**: Undoes the change in dimagi/commcare-android#1939, meaning that we're back to sorting unsent/saved forms by their last-modified date.

### DIFF
--- a/app/src/org/commcare/adapters/IncompleteFormListAdapter.java
+++ b/app/src/org/commcare/adapters/IncompleteFormListAdapter.java
@@ -23,6 +23,7 @@ import org.commcare.views.IncompleteFormRecordView;
 
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


shubham1g5:
Maybe a long term solution here is to give user sort options by multiple params!

PR #1960, opened 2018-03-12
Removes stranded setTransactionSuccesful

shubham1g5:
Crash: https://www.fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/5aa268088cb3c2fa63a75788/sessions/5AA44BF800E40001275D784C70E54F29_DNE_0_v2?select_key=user

I am not able to repro this though this seems to happen intermittently on device farm causing @CaseListSearch to fail.

### DIFF
--- a/app/src/org/commcare/adapters/EntityStringFilterer.java
+++ b/app/src/org/commcare/adapters/EntityStringFilterer.java
@@ -72,7 +72,6 @@ public class EntityStringFilterer extends EntityFiltererBase {
         }
         Entity<TreeReference> e = fullEntityList.get(index);
         if (isCancelled()) {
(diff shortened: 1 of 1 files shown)


amstone326:
@shubham1g5 Just double checking that I understand the rationale here-- you're assuming that the crash at `setTransactionSuccessful()` at the end of `buildMatchList()` is happening because it has already been called in `getEntityAtIndex()`. So instead we can just always wait and let the other call to `setTransactionSuccessful()` take care of it?

shubham1g5:
@amstone326 that's right. We are immediately returning if 'isCancelled()' is true and there is nothing really happenig between the two setTransactionSuccesful calls in this case.

PR #1966, opened 2018-04-02
Fixes Save in FormEntry after Resuming Session

shubham1g5:
FB: https://manage.dimagi.com/default.asp?272815

Instance path should not be re-initialized on resuming form entry after form Entry session. This results in a buggy behaviour(data loss) where the incomplete form from session expiration doesn't get updated with the data that user enters after resuming session. Moreover this incomplete instance gets uploaded to server since it gets marked as complete on save. 

@ctsims Just realizing that we might want to hotfix this instead. What do you think ?

Product Note: Bug Fix: When a user gets logged out of CommCare while entering data in a form, on login again he gets redirected to the Form he was previously in. This bug was causing any new data entered in this specific form to not get saved resulting in loss of data entered by user for this form.

### DIFF
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -964,9 +964,9 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
         boolean isRestartAfterSessionExpiration =
                 getIntent().getBooleanExtra(KEY_IS_RESTART_AFTER_EXPIRATION, false);
         // Set saved answer path
-        if (FormEntryInstanceState.mInstancePath == null || isRestartAfterSessionExpiration) {
+        if (FormEntryInstanceState.mInstancePath == null) {
             instanceState.initInstancePath();
-        } else {
+        } else if (!isRestartAfterSessionExpiration) {
             // we've just loaded a saved form, so start in the hierarchy view
             Intent i = new Intent(this, FormHierarchyActivity.class);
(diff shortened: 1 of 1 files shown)


ctsims:
Oh wow, that's a pretty serious issue.

I'm pro-hotfixing here, since even if this affects a small number of users, it also creates a situation which is _super_ hard to understand or spot, so it seems like it's high risk to continue leaving it live. We can't be sure how many problems are deriving from it.

amstone326:
@shubham1g5 @ctsims Hey, sorry for this and thanks for figuring out what was up. 

Just one thing I want to clarify and request a non-code change for: the Product Note implies that the bug here was the fact that the user gets re-entered into the form after session expiration. That part is intentional (which I think you realize because the code change in this PR retains that behavior); the bug was just that the instance path was getting re-initialized in that scenario when it shouldn't have been. Can we fix the Product Note to reflect that?

ctsims:
@amstone326 Good catch!

I think I'd actually reframe the product note entirely, since the details aren't relevant to the user, really.

I'd just make the note:

**Product Note:**
Corrected a bug where forms could contain invalid data when a user resumed the form after being automatically logged out.

amstone326:
@shubham1g5 I'm merging this but will you update the Product Note in the PR comment to be what Clayton posted above, so that Product sees the right thing if/when they come to view this PR?

shubham1g5:
@amstone326 Thanks. Just corrected the note. Sorry for framing it badly. Didn't really want to imply that resuming form entry is a bug! Also thanks for hotfixing 👍

PR #1967, opened 2018-04-05
Adds Data change logs along with Kotlin Support

shubham1g5:
* Implements data change log ([Spec Here](https://docs.google.com/document/d/1LJIaQMoex6SM90gAxJbLRsMw8FC_0FtJLbMbmxGE5dE/edit)) )

Screenshots:

Product Note: Adds additional mobile logs to allow for better debugging. These logs can be accessed by going to App Manager -> Advanced Settings -> Data Change Logs


shubham1g5:

shubham1g5:

shubham1g5:
@wpride Jenkins machine is constantly running out of memory while multidexing after adding Kotlin. I am wondering if it's possible for us to add SWAP space to the Jenkins mobile agent as mentioned here - https://stackoverflow.com/questions/31041512/jenkins-build-throwing-an-out-of-memory-error

shubham1g5:

app/src/org/commcare/logging/DataChangeLogger.kt
```diff
@@ -0,0 +1,156 @@
...
+ */
+class DataChangeLogger {
+
+    fun init(context: Context) {
+        initLogFiles(context)
+    }
+
+    private fun initLogFiles(context: Context) {
+        if (!isExternalStorageWritable()) {
+            Log.e(TAG, "External Storage unavialable to write logs")
```
amstone326: Let's log this information to the device logs instead. Logging to the console does ~nothing for us in situations of real app usage, and that way we'll be able to refer to the device logs to know if we shouldn't expect a user's device to have Data Change Logs on it.

app/src/org/commcare/logging/DataChangeLogger.kt
```diff
@@ -0,0 +1,156 @@
...
+import java.io.IOException
+import java.text.SimpleDateFormat
+import java.util.*
+
+/**
+ *  Used to log some crtitical platform events like DB migrations, Clear user data,
+ *  app installations etc. All these logs are backed by file storage and are accessible
+ *  from outside CommCare
+ */
+class DataChangeLogger {
```
amstone326: 100% for my learning/understanding: Can you explain the reasoning behind the general structure you used for this class, in terms of having a non-static init methods and then the `companion object` that houses everything else? From what I read, anything placed inside the `companion object` is basically equivalent to being a static method; is that right? If so, I'm still a little uncertain as to why the class fields go in there as well, and the reasoning behind keeping certain methods outside of that.

app/src/org/commcare/logging/DataChangeLogger.kt
```diff
@@ -0,0 +1,155 @@
...
+
+    private fun initLogFiles(context: Context) {
+        if (!isExternalStorageWritable()) {
+            Log.e(TAG, "External Storage unavialable to write logs")
+            return
+        }
+
+        primaryFile = initFile(context, PRIMARY_LOG_FILE_NAME)
+        if (primaryFile!!.exists() && primaryFile!!.length() > MAX_FILE_SIZE) {
+            // When primary file gets full -> Delete secondary and move logs in primary to secondary
```
amstone326: Were you able to test this behavior manually? (the easiest way is probably to reduce the max file size)

app/src/org/commcare/logging/DataChangeLogger.kt
```diff
@@ -0,0 +1,155 @@
...
+        /**
+         * Logs a given message to the fileSystem
+         */
+        @JvmStatic
+        fun log(message: String) {
+            // Include this info as part of any crash reports
+            CrashUtil.log(message)
+
+            // Write to local storage
+            if (primaryFile != null && primaryFile!!.exists()) {
```
amstone326: Given that you've already checked for `primaryFile != null`, what purpose does using the `!!` serve?

app/src/org/commcare/logging/DataChangeLogger.kt
```diff
@@ -0,0 +1,155 @@
...
+        }
+        return file
+    }
+
+    companion object {
+
+        private const val PRIMARY_LOG_FILE_NAME = "CommCare Data Change Logs.txt"
+        private const val SECONDARY_LOG_FILE_NAME = "CommCare Data Change Logs1.txt"
+        private const val DATE_FORMAT = "d MMM yy HH:mm:ss z"
+        private const val MAX_FILE_SIZE = (40 * 1000).toLong() // ~40KB
```
amstone326: Curious how you picked this?

app/src/org/commcare/logging/DataChangeLogger.kt
```diff
@@ -0,0 +1,155 @@
...
+        if (!isExternalStorageWritable()) {
+            Log.e(TAG, "External Storage unavialable to write logs")
+            return
+        }
+
+        primaryFile = initFile(context, PRIMARY_LOG_FILE_NAME)
+        if (primaryFile!!.exists() && primaryFile!!.length() > MAX_FILE_SIZE) {
+            // When primary file gets full -> Delete secondary and move logs in primary to secondary
+            secondaryFile = initFile(context, SECONDARY_LOG_FILE_NAME)
+            secondaryFile!!.delete()
```
amstone326: Why are you deleting the file right after initializing it?

app/src/org/commcare/logging/DataChangeLogger.kt
shubham1g5: Yup. It works as expected.

app/src/org/commcare/logging/DataChangeLogger.kt
shubham1g5: It's a bit confusing for me too. But as I understand since `primaryFile` is a mutable nullable property, the kotlin compiler throws an error when I just do `primaryFile.exists` since it's possible that some other process would have changed it's value back to `null` after the null check. So adding !! is just asserting that it's still not null.

In contrast we don't need any such checks for local properties - [Example here](https://github.com/dimagi/commcare-android/blob/dataChangeLogs/app/src/org/commcare/logging/DataChangeLogger.kt#L139)

app/src/org/commcare/logging/DataChangeLogger.kt
shubham1g5: I did a basic byte calculation for 200 logs assuming each to be 100 characters long.

app/src/org/commcare/logging/DataChangeLogger.kt
shubham1g5: Since I wanted all constants and both file accessors to be static, I have put them in `companion` . Though I get your confusion for not making `init` static which I see no harm in. I think since we only ever want one instance of `DataChangeLogger` we can instead make this whole class a Kotlin object which will also remove the need for the `companion`.

app/src/org/commcare/logging/DataChangeLogger.kt
```diff
@@ -0,0 +1,156 @@
...
+        private var primaryFile: File? = null
+        private var secondryFile: File? = null
+
+        /**
+         * Logs a given message to the fileSystem
+         */
+        @JvmStatic
+        fun log(message: String) {
+            // Include this info as part of any crash reports
+            CrashUtil.log(message)
```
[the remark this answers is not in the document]
amstone326: Whoa!! I had no idea Crashlytics allowed us to add informational logs to crashes. What I'm not clear on is how Crashlytics decides which crashes a given log will get attached to?

app/src/org/commcare/logging/DataChangeLogger.kt
amstone326: Got it ok yea, obviously I'm super unfamiliar with Kotlin coding conventions, but that does sound a little more intuitive to me than the current setup.

app/src/org/commcare/logging/DataChangeLogger.kt
amstone326: Ah I see, so Kotlin automatically treats all function args as `val`s rather than `var`s? Do you know why that is?

app/src/org/commcare/logging/DataChangeLogger.kt
amstone326: Makes sense, though we may want to bump up the number of logs we plan to keep around. If we expect each file to fit ~200 logs, that gives us a history of 400 total. I think the number Clayton had proposed was closer to 1,000 logs, so maybe we can do 500 per file?

app/src/org/commcare/logging/DataChangeLogger.kt
amstone326: Ah of course. Thanks.

app/src/org/commcare/logging/DataChangeLogger.kt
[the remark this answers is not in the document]
shubham1g5: My assumption is we will only see the logs logged in the same session as that of crash. So I think from the time the app starts Fabric should be maintaining a list of logs which gets attached to any crash that happens in that session and this log list should be getting reset on every session start.

app/src/org/commcare/logging/DataChangeLogger.kt
shubham1g5: No, they both are `var`s but since Kotlin like Java follows pass-by-value semantics while calling a function, each function call has it's own local copy and it's not possible for any other process to change it.

amstone326: Haha ok that does make sense- thanks for the explanation!

amstone326:
@shubham1g5 Just want to make sure you know that this has conflicts and some pending review requests from me.

shubham1g5:
@amstone326 Ready for review again.

app/src/org/commcare/views/CommCareShareActionProvider.java
```diff
@@ -0,0 +1,49 @@
...
+
+    private final Context mContext;
+
+    public CommCareShareActionProvider(Context context) {
+        super(context);
+        mContext = context;
+    }
+
+    @Override
+    public View onCreateActionView() {
```
amstone326: So I just did a little bit of hunting around on Stack Overflow because I was surprised at how much work it took just to customize the icon, and found this suggestion: https://stackoverflow.com/a/22247524/3873658 (just read the "UPDATE" line at the top, not the rest of the post). I just confirmed locally that his/her suggestion works; do you have any opposition to switching to that?

app/src/org/commcare/activities/DataChangeLogsActivity.java
```diff
@@ -0,0 +1,57 @@
...
+        mShareActionProvider = (CommCareShareActionProvider)item.getActionProvider();
+        setShareIntent();
+    }
+
+    // Call to update the share intent
+    @RequiresApi(api = Build.VERSION_CODES.ICE_CREAM_SANDWICH)
+    private void setShareIntent() {
+        if (mShareActionProvider != null) {
+            Intent shareIntent = new Intent(Intent.ACTION_SEND);
+            shareIntent.setType("text/plain");
```
amstone326: I'm sorry I didn't notice/think of this on my first review, but I just realized that I was envisioning us sharing/sending the file itself, rather than sending the log contents as the body of a plain text message. In the event that the logs have gotten quite long, I think that would cleaner. Does that switch make sense to you as well, or was there another reason you decided to do it this way?

app/src/org/commcare/views/CommCareShareActionProvider.java
shubham1g5: Nice. This one looks much cleaner! Thanks 👍

app/src/org/commcare/activities/DataChangeLogsActivity.java
shubham1g5: No particular reason. Changed to files.

app/src/org/commcare/google/services/analytics/AnalyticsParamValue.java
```diff
@@ -89,7 +89,7 @@
     public static final String MANAGE_SD = "manage_sd";
     public static final String WIFI_DIRECT = "wifi_direct";
     public static final String CONNECTION_TEST = "connection_test";
-    public static final String CLEAR_USER_DATA = "clear_user_data";
+    public static final String CLEAR_USER_DATA = "ClearUserData";
```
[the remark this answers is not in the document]
shubham1g5: Reverted

amstone326:

amstone326:
@shubham1g5 build is failing after your most recent changes, and also has conflicts with master now (must be from merging Will's markdown PR)

shubham1g5:
@amstone326 Thanks. Looking into this now

shubham1g5:
@amstone326 Resolved

amstone326:
Release Note: Adds a class of custom logs viewable on-device from the CommCare App Manager that record important "data change" actions which occur in CommCare (installs, updates, migrations, etc.)

amstone326:
@shubham1g5 I just added a release note for this since I think this is worthy of a public announcement.  Can you add some documentation on what these are and how to access them to [this page](https://confluence.dimagi.com/display/commcarepublic/CommCare+Android+Troubleshooting)

amstone326:
Hm actually, changing my mind-- I think it makes the most sense to document this feature at the bottom of [this page](https://confluence.dimagi.com/display/commcarepublic/Force+Log+Submissions), with the directive being something like "If you are unable to get to the point required for forcing log submissions because your app won't start up or let you log in, it may be helpful to submit the following to support instead". 

I've just added a link from the Android Troubleshooting page to the log submission page, so that people will now come across both of these things as 1 option for troubleshooting.

shubham1g5:
@amstone326 Yeah. Will add some info soon.

PR #1968, opened 2018-04-06
Enables Change Language Option in Practice Mode

shubham1g5:
Did this as an one off build earlier. Though since it's now a permanent requirement for ICDS, PRing it to master. 

Product Note: Adds ability to change language in Practice mode.


app/unit-tests/src/org/commcare/android/tests/DemoUserRestoreTest.java
```diff
@@ -41,16 +42,6 @@
     private final static String REF_BASE_DIR =
             "jr://resource/commcare-apps/demo_user_restore/";
 
-    @Test
-    public void loginUsingDemoUserWithoutRestore() {
```
shubham1g5: Removed this since same steps as in this test are executed in `demoUserRestoreAndUpdateTest` as well

PR #1969, opened 2018-04-06
Make noun project credit translatable (for CAS)

amstone326:
Note that I'm not 100% sure how the `update_translations` script in the release process is going to react to a multi-line translations value like this. If we're concerned I could try testing it ahead of time, but I didn't do so right now because it's a pain to run that in isolation without all of the surrounding harnessing.


shubham1g5:

PR #1972, opened 2018-04-12
Fix for Pre v17 Upgrade Path on Case Model

ctsims:
Fix for pre V6 Case Model schema to not index owner id, fixes an issue where the updater can't perform an in-place upgrade for very old db's


amstone326:
@ctsims do you think this should get a release note?

ctsims:
Release Note:

"Fixed bugs which were blocking logins after upgrades from older app versions."

PR #1974, opened 2018-04-13
Target specific flavour: Show Link to Open/Install App on Error Screen

shubham1g5:
Spec: https://docs.google.com/document/d/13GBlrrSQ2qu8u8sOeE4-kz3rydnyZiWfOaXMB31NqXA/edit

New screenshots:

1. Error page if LTS is not installed

2. Error page if LTS is already installed

Product Note: Adds a check on mobile to not allow apps targeting LTS to be installed on non LTS Commcare flavour and vice versa.( See the spec linked above for details)


app/res/values/strings.xml
```diff
@@ -352,4 +352,17 @@ Some images in CommCare are graciously used under Creative Commons Licensing fro
     <string name="sign_button" cc:translatable="true">Gather Signature</string>
     <string name="markup_image" cc:translatable="true">Markup Image</string>
     <string name="draw_image" cc:translatable="true">Sketch Image</string>
+
+    <!--Incorrect Target Screen-->
+    <string name="target_mismatch_install_help" cc:translatable="true">If you have any issues installing, please contact your Helpdesk</string>
```
amstone326: unused

app/res/values/strings.xml
```diff
@@ -352,4 +352,17 @@ Some images in CommCare are graciously used under Creative Commons Licensing fro
     <string name="sign_button" cc:translatable="true">Gather Signature</string>
     <string name="markup_image" cc:translatable="true">Markup Image</string>
     <string name="draw_image" cc:translatable="true">Sketch Image</string>
+
+    <!--Incorrect Target Screen-->
```
amstone326: Since none of these are being referenced from a layout file, I would have a slight preference for moving them to `android_translatable_strings.txt`. I think it's good for us to stick to some rule about which strings we put in which files.

app/res/values/strings.xml
```diff
@@ -352,4 +352,17 @@ Some images in CommCare are graciously used under Creative Commons Licensing fro
...
+    <string name="target_mismatch_install_help" cc:translatable="true">If you have any issues installing, please contact your Helpdesk</string>
+    <string name="target_mismatch_lts_required" cc:translatable="true">App Requires CommCare LTS</string>
+    <string name="target_mismatch_lts_required_install" cc:translatable="true">Install CommCare LTS</string>
+    <string name="target_mismatch_lts_open" cc:translatable="true">Open CommCare LTS</string>
+    <string name="target_mismatch_lts_required_info" cc:translatable="true">In order to install CommCare LTS, Please click on install button below. Then try installing your app again.</string>
```
amstone326: Slight modification suggestion (and the same for the parallel non-LTS string):
`Follow the link below to install CommCare LTS from the Play Store, and then try installing your app again.`

app/res/values/strings.xml
```diff
@@ -352,4 +352,17 @@ Some images in CommCare are graciously used under Creative Commons Licensing fro
...
+    <string name="target_mismatch_lts_available_info" cc:translatable="true">You can only install your app in CommCare LTS, Please click on open button below and try installing your app again.</string>
```
amstone326: Slight modification suggestion (and the same for the parallel non-LTS string):
`This app requires CommCare LTS. Click the button below to open CommCare LTS, and then try installing your app again.`

app/res/values/strings.xml
shubham1g5: I put these in here since it results in better IDE/Gradle support in comparison to `android_translatable_strings.txt` . For ex. If one of these strings get deleted by mistake the build will break and we will get indication of something is wrong. I am of opinion that we should try to avoid `android_translatable_strings.txt`  for this reason unless it's absolutely necessary.

amstone326: Hadn't thought about it that way, that's fair.

PR #1979, opened 2018-04-17
Device log cleaning

amstone326:
All of the changes made in this PR were done toward 1 of 4 goals:

1. Making sure all log messages can be traced back to the location in-code that triggered them, by ensuring the message contains some static content (e.g. the log should not be just the content of an exception's message).
2. Carrying out some of the proposed deletions that I made in [this doc](https://docs.google.com/document/d/1hL5S_x_0wImIL5z5Io2Cp8Z3NK3gkQb1Q_uawX3O_ks/edit?usp=sharing) a while back. For each item in the doc, I've added a message in purple explaining what I ultimately decided to do and why. 
3. Fixing log messages that were generally wrong or confusing.
4. Making sure a log message is being recorded under a log type that both makes sense and is 1 of the official types in the `LogTypes` class.

cross-request: https://github.com/dimagi/commcare-core/pull/788


app/src/org/commcare/tasks/FormRecordCleanupTask.java
```diff
@@ -345,9 +345,7 @@ private static void wipeRecord(Context context, int sessionId,
                             "Inconsistent formRecordId's in session storage");
                 }
             } catch (Exception e) {
-                Logger.exception(e);
-                Logger.log(LogTypes.TYPE_ERROR_ASSERTION,
-                        "Session ID exists, but with no record (or broken record)");
+                Logger.exception("Session ID exists, but with no record (or broken record)", e);
```
shubham1g5: I feel like ideally we should be  tagging such logs with both `TYPE_ERROR_ASSERTION` and `exception` for such cases. Is it possible to provide multiple tags to a log with SUMO ?

app/src/org/commcare/tasks/FormLoaderTask.java
```diff
@@ -102,7 +102,7 @@ protected FECWrapper doTaskBackground(Uri... form) {
                     " from cached file: " + formBin.getAbsolutePath());
             fd = deserializeFormDef((Context)activity, formBin);
             if (fd == null) {
-                Logger.log(LogTypes.SOFT_ASSERT,
+                Logger.log(LogTypes.TYPE_RESOURCES,
```
shubham1g5: I would be in support of `LogTypes.SOFT_ASSERT` here since we are assuming that this line of code should never get called and `SOFT_ASSERT` is a way for us to validate that assumption.  I think with SUMO we should have some kind of alert setup  based on `SOFT_ASSERT` so that we get to know if a new `SOFT_ASSERT` is getting logged which most probably requires dev attention/action.

shubham1g5:

app/src/org/commcare/tasks/FormLoaderTask.java
amstone326: mm that's a good point, I was still thinking along the lines of the `SOFT_ASSERT` tag not really being useful to us

amstone326: Sorry, I should amend that slightly-- that's a good point in general and I'm going to revert some of the other places in this PR where I removed a usage of `SOFT_ASSERT`. But this particular instance is one in which `SOFT_ASSERT` actually isn't right, because we do expect this to happen sometimes. (As the lines below indicate, we sometimes expect the form cache to be gone or invalid, and to have to re-load it from the file)

app/src/org/commcare/tasks/FormRecordCleanupTask.java
amstone326: I haven't currently set it up to make that possible, no. we'd have to just send the same log twice with different tags

amstone326:
@shubham1g5 this is now actually ready for review

app/src/org/commcare/utils/FormUploadUtil.java
```diff
@@ -233,8 +233,11 @@ private static void logResponse(int responseCode, String responseString) {
         if (!(responseCode >= 200 && responseCode < 300)) {
             Logger.log(LogTypes.TYPE_WARNING_NETWORK, responseCodeMessage);
             Logger.log(LogTypes.TYPE_FORM_SUBMISSION, responseCodeMessage);
-            Logger.log(LogTypes.TYPE_FORM_SUBMISSION,
-                    "Response string to failed form submission attempt was: " + responseString);
+            if (!responseString.startsWith("<html>")) {
```
shubham1g5: Don't we still want to log just the response code if not the whole failure message ?

amstone326: Yea we do, that's why I left it outside of this new `if` block. Or are you saying you think we shouldn't? We definitely still want to log the response code whenever submission fails, even if the response _message_ isn't useful.

shubham1g5: Ahh sorry. Missed that we are already loggin the code.

PR #1981, opened 2018-04-20
Wsp/markdown media

wpride:
Addresses https://manage.dimagi.com/default.asp?274545

The Markdown library needs to use our custom `FileProvider` class in order to view images on Android 7+ devices. Fortunately our new markdown library allows us to override this behavior.

Temporarily don't merge because I need to think through how this will work on older versions/file paths.


shubham1g5:
@wpride I didn't get your concern about older versions! The error occurs only on Android 7+ devices which you are handling this using FileProvider shareable uri.

PR #1985, opened 2018-04-24
Adds test for DataChangeLogger

shubham1g5:
* Adds JUnit tests for `DataChangeLogger`

### DIFF
--- a/app/unit-tests/src/org/commcare/logging/DataChangeLoggerTest.java
+++ b/app/unit-tests/src/org/commcare/logging/DataChangeLoggerTest.java
@@ -1,0 +1,46 @@
+package org.commcare.logging;
+
+import android.net.Uri;
+
+import org.commcare.CommCareTestApplication;
+import org.commcare.android.CommCareTestRunner;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.robolectric.annotation.Config;
+
+import java.util.ArrayList;
+
+import static junit.framework.Assert.assertEquals;
+import static junit.framework.Assert.assertTrue;
+
+@Config(application = CommCareTestApplication.class)
+@RunWith(CommCareTestRunner.class)
+public class DataChangeLoggerTest {
+
+    @Before
+    public void setupTests() {
+        DataChangeLogger.init(CommCareTestApplication.instance());
+    }
+
+    @Test
+    public void getLogs_shouldContainLoggedMessages() {
+        assertTrue(DataChangeLogger.getLogs().contains(new DataChangeLog.CommCareInstall().getMessage()));
+
+        DataChangeLog clearUserDataLog = new DataChangeLog.ClearUserData();
+        DataChangeLogger.log(clearUserDataLog);
+        assertTrue(DataChangeLogger.getLogs().contains(clearUserDataLog.getMessage()));
+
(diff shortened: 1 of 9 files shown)


app/unit-tests/src/org/commcare/logging/DataChangeLoggerTest.java
```diff
@@ -0,0 +1,46 @@
...
+    @Before
+    public void setupTests() {
+        DataChangeLogger.init(CommCareTestApplication.instance());
+    }
+
+    @Test
+    public void getLogs_shouldContainLoggedMessages() {
+        assertTrue(DataChangeLogger.getLogs().contains(new DataChangeLog.CommCareInstall().getMessage()));
+
+        DataChangeLog clearUserDataLog = new DataChangeLog.ClearUserData();
```
amstone326: @shubham1g5 I think these tests are nice to have in here, but it feels a bit circular to me to count as a fully-sufficient test of the functionality. All we're doing here is directly executing the method that writes the log, and then checking that the log was written. To me that only tests half of the feature, and not the more complicated half; we're not doing any testing for whether the expected actions/workflows correctly trigger the logs being written in the first place. If that seems tricky to do in robolectric, I think a Calabash test makes sense (that's what I was originally envisioning for this).

shubham1g5: Agree that the point of these unit tests is only to verify the functionalities of `DataChangeLogger`. I think these 2 tests gives us around 100% test coverage for `DataChangeLogger` since they call all public methods for DataChangeLogger and hence goes through all `DataChangeLogger` code. 

You are right that rest half of actually calling `DataChangeLoggger` in various places is still not tested though I also think since UI tests cause a lot of overhead we should use them for only critical/core app functionalities.

amstone326: Hm, I don't think I agree with that. In our pre-Calabash QA world, this would have qualified as something worthy of adding a manual QA test for. I really don't think the fact that we've moved to an automated testing infrastructure should cause us to leave things out of QA that we would have previously included. If you think there's a reason to specifically do this in manual QA rather than Calabash, that also seems totally fine to me. But I think it's a bad precedent for us to start leaving things out of our QA plan just because Calabash is clunky.

shubham1g5: That does make sense to me. Can I just add this to manual QA plan here - https://docs.google.com/spreadsheets/d/1tbpfaxcKxI1YG957GCtbELwr6DGo0M1ul-NgrQUnEF8/edit?ts=5a7477d8#gid=876347402

amstone326: This is the right sheet actually: https://docs.google.com/spreadsheets/d/11OSH0pI4HbvbHiDWNrIrXvDwVzVnhcBibAshOIWmgng/edit#gid=760951263. You can add the test there directly if you want, or you can submit a QA request form to the QA team. If you do it yourself, I usually ask either Clark or Ben to do a quick review of the test when I'm done, to make sure they think the instructions are completely clear.

shubham1g5: @amstone326 These are now in here - https://docs.google.com/spreadsheets/d/11OSH0pI4HbvbHiDWNrIrXvDwVzVnhcBibAshOIWmgng/edit#gid=876347402

PR #1986, opened 2018-04-24
2.43 - Don't exit before adding url param

amstone326:
Remake of https://github.com/dimagi/commcare-android/pull/1984 into 2.43


amstone326:
The build failure is just because the jenkins configuration for the PR builder job has already been changed to accommodate the new gradle structure

PR #1987, opened 2018-04-24
Hide home screen training button if its menus are hidden

amstone326:
**Question for everyone**: I've created this PR as the easiest way to get input on the soundness of this -- the ICDS team would like to make it so that if an app contains 1 or more training menus, _but_ they all have display conditions which evaluate to false, then the home screen training button does not show up. I explained (in a simplified way) that this is not strictly doable within CommCare's model because at the time the home screen buttons are rendered, the proper evaluation context for the menus isn't present. However, I've realized that technically I can reconstruct it, as I've done in this PR. My question is whether anyone sees any risks associated with doing this that I'm missing, or if it just seems too gross to allow ourselves to do. (I've tested that it works from a desired functionality standpoint).

### DIFF
--- a/app/src/org/commcare/CommCareApp.java
+++ b/app/src/org/commcare/CommCareApp.java
@@ -300,8 +300,11 @@ public class CommCareApp implements AppFilePathBuilder {
     }
 
     public boolean hasVisibleTrainingMenu() {
+        // This is the same eval context that would be used to evaluate the relevancy conditions of
+        // these menus when they're actually loaded
         EvaluationContext ec =
-                CommCareApplication.instance().getCurrentSessionWrapper().getEvaluationContext();
+                CommCareApplication.instance().getCurrentSessionWrapper().getEvaluationContext(Menu.TRAINING_MENU_ROOT);
+
         for (Suite s : platform.getInstalledSuites()) {
             List<Menu> trainingMenus = s.getMenusWithRoot(Menu.TRAINING_MENU_ROOT);
             if (trainingMenus != null) {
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/CommCareApp.java
```diff
@@ -296,11 +299,28 @@ public AndroidCommCarePlatform getCommCarePlatform() {
...
+            List<Menu> trainingMenus = s.getMenusWithRoot(Menu.TRAINING_MENU_ROOT);
             if (trainingMenus != null) {
-                return true;
+                for (Menu m : trainingMenus) {
+                    try {
+                        if (m.getMenuRelevance() == null ||
+                                FunctionUtils.toBoolean(m.getMenuRelevance().eval(ec))) {
+                            return true;
+                        }
+                    } catch (XPathSyntaxException e) {
```
ctsims: I'm not sure I'm aligned with the logic here. feels a little bit like if there's an invalid xpath expression here it's useful to fail as early as possible. Is it out of the question to let this bubble up to the point that we trigger an unrecoverable error (like we do when there's an error later in the app)? If it makes things significantly muckier we can skip it, but especially Syntax errors are "the app can't work as designed" kind of problems, so failing to load entirely seems like the most fail-fast behavior.

amstone326: I'm going to merge this as-is so it's in before QA, but if you disagree still let's discuss further.

amstone326:
Merging w/ build failure because it's only failing due to the new configuration for the assets path

app/src/org/commcare/CommCareApp.java
ctsims: Quick note: None of this is blocking (which is why I marked as approved)

I think the 'be visible down the road' assumption is a bit dangerous. In my mind errors which are structural (IE: XPath syntax is just wrong) are always better to reveal earlier rather than later, since they are unrecoverable from the end-user (IE: CHW, not app builder) perspective, so revealing them in a 'blocking' way forces the app builder to not confuse a broken app with a functional one. 

Maybe the best way to frame this thought is "What is the value of allowing an app builder who has defined a broken condition to continue using the application while a portion of it is broken?" I think the downside of being blocked is lower from the downside of potentially not clicking the button and realizing that the app is broken.

PR #1991, opened 2018-04-30
2.43 - Check for entries in determining training button visibility

amstone326:
Extension of what was done in https://github.com/dimagi/commcare-android/pull/1987 to also factor in any entries that are direct children of the `training-root` menu (which would happen if a training menu has 'Display Only Forms' enabled).

### DIFF
--- a/app/src/org/commcare/CommCareApp.java
+++ b/app/src/org/commcare/CommCareApp.java
@@ -39,6 +39,7 @@ import org.javarosa.core.services.locale.Localization;
 import org.javarosa.core.services.storage.Persistable;
 import org.javarosa.core.util.UnregisteredLocaleException;
 import org.javarosa.xpath.expr.FunctionUtils;
+import org.javarosa.xpath.expr.XPathExpression;
 import org.javarosa.xpath.parser.XPathSyntaxException;
 
 import java.io.File;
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/CommCareApp.java
```diff
@@ -299,28 +300,61 @@ public AndroidCommCarePlatform getCommCarePlatform() {
...
-                    try {
-                        if (m.getMenuRelevance() == null ||
-                                FunctionUtils.toBoolean(m.getMenuRelevance().eval(ec))) {
-                            return true;
-                        }
-                    } catch (XPathSyntaxException e) {
-                        // Now is the wrong time to show the user an error about this since they
-                        // haven't actually navigated to the menu. To be safe, just assume that this
-                        // menu is visible, and then if they navigate to it they'll see the XPath error there
+            if (visibleMenusInTrainingRoot(s, ec) || visibleEntriesInTrainingRoot(s, ec)) {
```
shubham1g5: Both these methods can be static. We might also want to extract them out to a helper class.

amstone326: @shubham1g5 Thanks! I made them static but after looking through our existing helper classes for a bit, i didn't find any that seemed fitting, and I think it's a little overkill to create a new helper class for just 2 methods.

amstone326:
Build failure is just the asset path change issue

PR #2000, opened 2018-05-16
Actually get the translated version of the string if it exists

amstone326:


shubham1g5:

PR #2004, opened 2018-05-17
Remove concept of being able to edit a form after it's completed

amstone326:
This is a stale concept, and one we don't want/need to continue supporting. Once a form is marked as completed, it is never again editable.

cross-request: https://github.com/dimagi/commcare-core/pull/804

### DIFF
--- a/app/src/org/commcare/provider/InstanceProviderAPI.java
+++ b/app/src/org/commcare/provider/InstanceProviderAPI.java
@@ -18,7 +18,6 @@ public final class InstanceProviderAPI {
         private InstanceColumns() {
         }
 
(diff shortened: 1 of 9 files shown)


app/src/org/commcare/provider/InstanceProviderAPI.java
```diff
@@ -18,7 +18,6 @@ private InstanceProviderAPI() {
         private InstanceColumns() {
         }
 
-        public static final Uri CONTENT_URI = Uri.parse("content://" + AUTHORITY + "/instances");
```
amstone326: unused

PR #2005, opened 2018-05-17
Set submission ordering # at same time we mark a form as completed

amstone326:
**Includes https://github.com/dimagi/commcare-android/pull/2004, review + merge that first.**

This will fix http://manage.dimagi.com/default.asp?276261. I realized what was happening here from examining the device logs surrounding submission of the forms that Derek linked in that ticket. Because of where we were previously calling `setFormNumberForSubmissionOrdering()`, a crash that occurred at the end of form entry would result in the `FormRecord` being successfully saved as complete, but never having its submission ordering number set (and therefore having a value of -1). This PR fixes that by setting the submission order number at the same time we update a `FormRecord`'s status to `STATUS_COMPLETE`. (The crashes themselves that were occurring post-form entry were varied and none of them seemed particularly alarming, so I chose to focus on fixing the fact that any sort of crash would cause this malfunction).

### DIFF
--- a/app/src/org/commcare/tasks/SaveToDiskTask.java
+++ b/app/src/org/commcare/tasks/SaveToDiskTask.java
@@ -13,6 +13,7 @@ import org.commcare.models.database.SqlStorage;
 import org.commcare.models.encryption.EncryptionIO;
 import org.commcare.tasks.templates.CommCareTask;
 import org.commcare.util.LogTypes;
+import org.commcare.utils.StorageUtils;
 import org.javarosa.core.io.StreamsUtil;
 import org.javarosa.core.model.FormDef;
 import org.javarosa.core.model.FormIndex;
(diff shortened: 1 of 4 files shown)


app/src/org/commcare/tasks/SaveToDiskTask.java
```diff
@@ -149,36 +149,29 @@ private void updateFormRecord(SqlStorage<FormRecord> formRecordStorage, boolean
...
                 recordName = formDefRecord.getDisplayname();
             }
+        }
 
-            FormRecord formRecord = CommCareApplication.instance().getCurrentSessionWrapper().getFormRecord();
-            formRecord.setFilePath(mFormRecordPath);
-            formRecord.setDisplayName(recordName);
-            formRecord.setCanEditWhenComplete(canEditWhenComplete);
-            formRecord.setStatus(status);
-            formRecord.update(formRecordStorage);
```
shubham1g5: `FormRecord::update` can be made private now after this change.

amstone326:

PR #2006, opened 2018-05-21
Incorrect Time Notification

shubham1g5:
* Shows alert dialog to correct time when the custom property `incorrect_time_warning_enabled` is set. 
* Adds `drift` in CommCare Session
* Sends `current_drfit` and `max_drfit_since_last_heartbeat` in heartbeat request. 

Screenshot - 

Product Note: On login into CommCare, user sees an alert dialog notifying user to correct time in case device time is wrong by more than an hour. ([incorrect_time_warning_enabled](https://confluence.dimagi.com/display/internal/CommCare+Android+Developer+Options+--+Internal#CommCareAndroidDeveloperOptions--Internal-ShowIncorrectTimeWarning) needs to be set in Custom Properties section in Advanced Settings for utilizing this functionality)

cross-request: https://github.com/dimagi/commcare-core/pull/807


shubham1g5:

shubham1g5:

shubham1g5:
Tested with HQ changes and everything looks good.

app/src/org/commcare/activities/DrfitHelper.java
```diff
@@ -0,0 +1,73 @@
...
+        driftDialog.setPositiveButton(context.getResources().getString(R.string.incorrect_time_dialog_correct_time), (dialog, which) -> {
+            context.startActivity(new Intent(android.provider.Settings.ACTION_DATE_SETTINGS));
+            dialog.dismiss();
+        });
+        driftDialog.setNegativeButton(context.getResources().getString(R.string.incorrect_time_dialog_cancel), ((dialog, which) -> dialog.dismiss()));
+        return driftDialog;
+    }
+
+    private static boolean isWarningEnabled() {
+        return DeveloperPreferences.doesPropertyMatch(INCORRECT_TIME_WARNING_ENABLED, PrefValues.NO, PrefValues.YES);
```
amstone326: Is a preference for this going to get added on HQ? Or is the expectation that people will just add it as a custom preference?

app/src/org/commcare/activities/DrfitHelper.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    private static SharedPreferences getAppPreferences() {
+        return PreferenceManager.getDefaultSharedPreferences(CommCareApplication.instance());
```
amstone326: Can we switch all of the calls to `CommCarePreferenceManagerFactory.getCommCarePreferenceManager()` in this file to use `getAppPreferences()` instead? It seems weird to mix calls to both of those things in a single file, since they do the same thing but would appear to someone new reading the code as different things, and therefore be confusing. More generally, I would prefer it if nothing in `android` referred to `CommCarePreferenceManagerFactory` (except for the call to `init()` in `CommCareApplication`) _unless_ we switch _all_ references to the default SharedPrefs object to use that pathway. Basically we should pick one and stick with it.

app/src/org/commcare/activities/DrfitHelper.java
```diff
@@ -0,0 +1,73 @@
...
+
+public class DrfitHelper {
+
+    private static final String INCORRECT_TIME_WARNING_ENABLED = "incorrect_time_warning_enabled";
+    private static final String LAST_INCORRECT_TIME_WARNING_AT = "last_incorrect_time_warning_at";
+
+    static boolean shouldShowDriftWarning() {
+        if (isWarningEnabled()) {
+            long lastIncorrrectTimeWarningAt = getAppPreferences().getLong(LAST_INCORRECT_TIME_WARNING_AT, -1);
+            if (new Date().getTime() - lastIncorrrectTimeWarningAt > DAY_IN_MS) {
```
amstone326: Should this threshold be configurable at all? How did we decide on a day?

app/src/org/commcare/activities/DrfitHelper.java
shubham1g5: Yup, it's getting enabled through a custom property in Advanced app setiings on HQ.

app/src/org/commcare/activities/DrfitHelper.java
shubham1g5: I just took it from Clayton's initital spec of the feature and Derek didn't seem to have any issues with this. It just looks like a good duration to start with.

app/src/org/commcare/activities/DrfitHelper.java
shubham1g5: Nice catch.

app/src/org/commcare/activities/DrfitHelper.java
shubham1g5: Actually I am not aware of any other method to achive this other than custom property . What do you mean when you say `Is a preference for this going to get added on HQ?`

app/src/org/commcare/activities/DrfitHelper.java
amstone326: Fair

app/src/org/commcare/activities/DrfitHelper.java
amstone326: I'm referring to all of the named settings that are available if you go to the Advanced Settings tab for an app on HQ.

app/src/org/commcare/activities/DriftHelper.java
```diff
@@ -0,0 +1,71 @@
...
+        driftDialog.setNegativeButton(context.getResources().getString(R.string.incorrect_time_dialog_cancel), ((dialog, which) -> dialog.dismiss()));
+        return driftDialog;
+    }
+
+    private static boolean isWarningEnabled() {
+        return DeveloperPreferences.doesPropertyMatch(INCORRECT_TIME_WARNING_ENABLED, PrefValues.NO, PrefValues.YES);
+    }
+
+    private static ICommCarePreferenceManager getPreferenceManager() {
+        return CommCarePreferenceManagerFactory.getCommCarePreferenceManager();
```
amstone326: Sorry I think I didn't phrase my original comment as clearly as I could have, but I think it would be preferable to _not_ use this abstraction in Android code right now and just make the call to get the `SharedPreferences` object directly instead. Otherwise it seems pretty arbitrary that just this one place in Android code is doing it this way, but nowhere else is .

shubham1g5: got it

app/src/org/commcare/activities/DrfitHelper.java
shubham1g5: I see. Is the idea behind this is add the most common preference settings as a UI element but also allow people to add custom properties for not so common configs ?

amstone326: Yes exactly

PR #2009, opened 2018-06-08
Add more logs to debug missing app files issue

shubham1g5:
I saw some instances on Sumo that suggests that this issue starts occurring either after an app auto update. Furthermore, Sachit from Community health got back to us saying that there are users in the field who are facing storage corrupt issue after a playstore update. I am hoping sending data change logs to device logs will help us in determining the exact sequence of events for that particular update and would help us narrow down this bug. 

* Sends data change logs to device logs
* Logs result for auto update. 

cross-request: https://github.com/dimagi/commcare-core/pull/811

### DIFF
--- a/app/src/org/commcare/activities/LoginActivity.java
+++ b/app/src/org/commcare/activities/LoginActivity.java
@@ -31,6 +31,8 @@ import org.commcare.interfaces.RuntimePermissionRequester;
 import org.commcare.interfaces.WithUIController;
 import org.commcare.android.database.app.models.UserKeyRecord;
 import org.commcare.android.database.global.models.ApplicationRecord;
+import org.commcare.logging.DataChangeLog;
+import org.commcare.logging.DataChangeLogger;
 import org.commcare.models.database.user.DemoUserBuilder;
 import org.commcare.preferences.DevSessionRestorer;
 import org.commcare.suite.model.OfflineUserRestore;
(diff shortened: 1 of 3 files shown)


shubham1g5:

shubham1g5:

app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -542,6 +546,7 @@ protected void deliverResult(LoginActivity receiver,
                             Toast.makeText(receiver,
                                     Localization.get("login.update.install.success"),
                                     Toast.LENGTH_LONG).show();
+                            DataChangeLogger.log(new DataChangeLog.CommCareAppUpdated());
```
amstone326: @shubham1g5 This actually isn't the code path for an auto update; this is for the demo user restore feature. Are you trying to just log whenever an app update completes? If so, you should be able to do that somewhere in `UpdateActivity` or `UpdateTask`

shubham1g5: @amstone326 Hey. Would you be able to verify this again. The `deliverResult` here is part of `InstallStagedUpdateTask`. Is not this the task resposible for auto update ?

amstone326: Oh wow I absolutely read this wrong and too quickly, I think because of the way Github collapses the file, it looked like this was coming after line 207/211 above. Sorry about that!!

amstone326: However, this is still not the only place you want to be adding this logic. `InstallStagedUpdateTask` is also launched from `UpdateActivity::launchUpdateInstallTask`

shubham1g5: That [code path](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/activities/UpdateActivity.java#L439) we already covered while implementing data change log, though this auto update one got left behind.

PR #2010, opened 2018-06-19
Add external_id as indexable case property

wpride:
Enables indexed searches by an `external_id` property to be used by (you guessed it) external systems like SimPrints. The implementation uses the existing `external_id` case property as the value if it exists since this is already used by HQ in the same way. 

cross-request: https://github.com/dimagi/commcare-core/pull/814

### DIFF
--- a/app/src/org/commcare/models/database/app/AppDatabaseUpgrader.java
+++ b/app/src/org/commcare/models/database/app/AppDatabaseUpgrader.java
@@ -28,7 +28,6 @@ import org.javarosa.core.util.externalizable.PrototypeFactory;
 
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.HashMap;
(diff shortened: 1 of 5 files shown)


wpride:

app/src/org/commcare/models/database/app/AppDatabaseUpgrader.java
```diff
@@ -28,7 +28,6 @@
 
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.Comparator;
```
ctsims: no-op

ctsims: whoops, nevermind.

ctsims:
This looks good to go once the version bump happens and you've had a chance to test the migration

wpride:
Release Note: Make `external_id` property indexed

PR #2015, opened 2018-07-03
Re-download missing Resources from Recovery Screen 

shubham1g5:
* Implements 'Recover App' functionality in `RecoveryActivity` to re-download any missing resources
* `ResourceTable` now maintains a vector for `missingAppResources` which gets set while initializing an app

Screenshots - 

1. Storage Corrupt Warning - @ctsims I don't see any particular benefits of this screen since all this asks is whether to enter recovery screen or exit the app. Instead we can directly make user enter the recovery screen and user can always exit the app by pressing back. 

2. Recovery Screen

3. Recovery in progress

4. Recovery Success

cross-request: https://github.com/dimagi/commcare-core/pull/817

Product Note: Adds functionality to re-download missing app resources from Recovery Screen.

### DIFF
--- a/app/src/org/commcare/android/resource/installers/XFormAndroidInstaller.java
+++ b/app/src/org/commcare/android/resource/installers/XFormAndroidInstaller.java
@@ -70,6 +70,7 @@ public class XFormAndroidInstaller extends FileSystemInstaller {
     public boolean initialize(AndroidCommCarePlatform platform, boolean isUpgrade) throws
             IOException, InvalidReferenceException, InvalidStructureException,
             XmlPullParserException, UnfullfilledRequirementsException {
+        super.initialize(platform, isUpgrade);
         platform.registerXmlns(namespace, formDefId);
         return true;
     }
--- a/app/src/org/commcare/tasks/ProcessAndSendTask.java
+++ b/app/src/org/commcare/tasks/ProcessAndSendTask.java
@@ -44,7 +44,6 @@ import javax.crypto.spec.SecretKeySpec;
  */
 public abstract class ProcessAndSendTask<R> extends CommCareTask<FormRecord, Long, FormUploadResult, R> implements DataSubmissionListener {
 
-    private Context c;
     private String url;
     private FormUploadResult[] results;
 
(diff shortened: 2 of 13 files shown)


app/src/org/commcare/tasks/ResourceRecoveryTask.java
```diff
@@ -0,0 +1,76 @@
...
+            throw new RuntimeException(e);
+        } finally {
+            unsetTableListeners(global);
+        }
+        return success;
+    }
+
+    private String getProfileReference() {
+        String profileRef = ResourceInstallUtils.getDefaultProfileRef();
+        return profileRef.split("\\?")[0];
```
wpride: This looks a little magic-y; this is splitting the base URL and params right? Maybe a comment or more descriptive method title would be helpful?

app/src/org/commcare/tasks/ProcessAndSendTask.java
```diff
@@ -81,7 +80,6 @@ public ProcessAndSendTask(Context c, String url) {
      * @param inSyncMode blocks the user with a sync dialog
      */
     public ProcessAndSendTask(Context c, String url, boolean inSyncMode) {
-        this.c = c;
```
wpride: :+1:

app/src/org/commcare/android/resource/installers/XFormAndroidInstaller.java
```diff
@@ -70,6 +70,7 @@ public XFormAndroidInstaller(String localLocation, String localDestination, Stri
     public boolean initialize(AndroidCommCarePlatform platform, boolean isUpgrade) throws
             IOException, InvalidReferenceException, InvalidStructureException,
             XmlPullParserException, UnfullfilledRequirementsException {
+        super.initialize(platform, isUpgrade);
```
wpride: Interesting - OOC why were we able to get away with *not* calling `super` for so long but we need to now?

wpride:
@shubham1g5 Looks good - couple of questions on the `commcare-core` PR then good to merge

app/src/org/commcare/android/resource/installers/XFormAndroidInstaller.java
shubham1g5: Super class `FileSystemInstaller` earlier had `initialize` as an `abstract` method and I just  converted it to a `non abstract` one in this PR to be able to add common functionality to check for missing files before we proceed.

PR #2018, opened 2018-07-06
Adds crash log for Unique form id constraint failure

shubham1g5:
Adds more debugging info for https://www.fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/59e2076c61b02d480d8ee18e

### DIFF
--- a/app/src/org/commcare/models/AndroidSessionWrapper.java
+++ b/app/src/org/commcare/models/AndroidSessionWrapper.java
@@ -18,9 +18,12 @@ import org.commcare.suite.model.FormEntry;
 import org.commcare.suite.model.SessionDatum;
 import org.commcare.suite.model.StackOperation;
 import org.commcare.util.CommCarePlatform;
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/models/AndroidSessionWrapper.java
```diff
@@ -198,7 +199,12 @@ public void commitStub() {
         setFormRecordId(r.getID());
 
         SessionStateDescriptor ssd = SessionStateDescriptor.buildFromSessionWrapper(this);
-        sessionStorage.write(ssd);
+        try {
+            sessionStorage.write(ssd);
+        } catch (Exception e) {
+            CrashUtil.log("SessionStateDescriptor form id: " + ssd.getFormRecordId());
```
ctsims: Just a heads up that if there's ~any chance SSD can be null here that this will actually obfuscate the underlying error by resulting in a null pointer instead.

PR #2021, opened 2018-07-16
Fixes Crash due to SSD records getting detached in FormProvider Migration

shubham1g5:
CL: https://www.fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/59e2076c61b02d480d8ee18e?time=last-thirty-days

Repro: 
1. Install CommCare 2.43
2. Create 2 Forms A and B in that sequence
3. Delete A
4. Update to CommCare 2.44 or 2.44.1
5. Observe a crash on creating any new form

While migrating FormRecords in v23 migration, we delete the old table and rewrite all entries into a new FormRecord table which results in `id` of FormRecords getting changed to new values. Since `SessionStateDescriptor` records hold references to these ids, these ssd records need to be migrated as well. In a general case new form ids would be same as old causing this bug not to surface except when user or some process deletes a form record from the device which is not the latest. For example, I am highlighting the state of formRecord and SSD storage below after every step in repro above - 
```
Step		FormStorage Id              SSD Form Record Ids
2			1, 2			        1, 2 
3			2				2
4			1    			        2
5                       1, 2                           2, 2 (Unique Constraint Failure)

```

For devices which are already on 2.44, 2.44.1 I am checking if there are any ssd records who holds a form id not corresponding to any forms on the device to validate if there were forms deleted from device before updating to 2.44.0(1). In that case I am dropping the whole table since it means that records in SSD table holds references to either incorrect or non existent form records entries.

@ctsims Let me know if you can think of better strategies around this.

### DIFF
--- a/app/src/org/commcare/models/database/user/UserDatabaseUpgrader.java
+++ b/app/src/org/commcare/models/database/user/UserDatabaseUpgrader.java
@@ -695,15 +695,9 @@ class UserDatabaseUpgrader {
                 // Since we have wiped out SSD records, we won't be able to resume
                 // incomplete forms with their earlier session state. Therfore we are
                 // going to delete all incomplete form records as well
-                Vector<Integer> incompleteRecords = new Vector();
-                for (FormRecord formRecord : formRecordStorage) {
-                    if (formRecord.getStatus().contentEquals(FormRecord.STATUS_INCOMPLETE)) {
-                        incompleteRecords.add(formRecord.getID());
-                    }
-                }
-
-                for (int incompleteRecordId : incompleteRecords) {
-                    formRecordStorage.remove(incompleteRecordId);
+                Vector<FormRecord> incompleteRecords = formRecordStorage.getRecordsForValue(FormRecord.META_STATUS, FormRecord.STATUS_INCOMPLETE);
+                for (FormRecord incompleteRecord : incompleteRecords) {
+                    formRecordStorage.remove(incompleteRecord);
                 }
             }
             db.setTransactionSuccessful();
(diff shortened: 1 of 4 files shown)


shubham1g5:

ctsims:
@shubham1g5 Can you explain the implications of exactly what happens in each scenerio here for real users? Does this affect saved forms, in-progress forms, unsent forms?

shubham1g5:
@ctsims Sorry I didn't think it through last time. It's affecting the incomplete forms since we won't be able to get the session state for those forms after deleting the SSD table. So resuming incomplere form will likely result in bad form state and weird errors. 

Though I can't think of any way to recover from this situation apart from deleting all incomplete forms as well from device which is quite unfortunate.

app/src/org/commcare/models/database/user/UserDatabaseUpgrader.java
```diff
@@ -648,6 +664,50 @@ private boolean upgradeTwentyThreeTwentyFour(SQLiteDatabase db) {
...
+
+            if (strandedRecordObserved) {
+                SqlStorage.wipeTable(db, SessionStateDescriptor.STORAGE_KEY);
+
+                // Since we have wiped out SSD records, we won't be able to resume
+                // incomplete forms with their earlier session state. Therfore we are
+                // going to delete all incomplete form records as well
+                for (FormRecord formRecord : formRecordStorage) {
+                    if (formRecord.getStatus().contentEquals(FormRecord.STATUS_INCOMPLETE)) {
+                        formRecordStorage.remove(formRecord);
```
ctsims: So I generally think of deleting in-place like this (while still using an iterator) as being a dangerous operation, and would assume we'd need to make a "todelete" list and then apply the deletions afterwards. Am I incorrect about that WRT our iterators?

PR #2023, opened 2018-07-23
Fixes bug in updating File path for an existing Form with new file path

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/browse/MOB-17

This was silly on my end. Form record path wasn't  getting updated to the  `install` path from `upgrade` path since `updateFilePath` in `FormDefRecord.java` was not actually setting the new filepath value for the form before updating the form record.

This bug doesn't surface in normal cases since generally form records filepath references doesn't change. The bug only gets exposed in cases when namespace of a form changes in between 2 updates since that's the only case where a form record with `upgrade` path gets created and then we later wanna change it to `install` file path.  For installations who have already encountered this scenario, we are applying an one time update which checks for non existent `upgrade` file references in FormDef Records and moves the path to correposnding `install` path.

### DIFF
--- a/app/src/org/commcare/android/database/app/models/FormDefRecord.java
+++ b/app/src/org/commcare/android/database/app/models/FormDefRecord.java
@@ -138,6 +138,7 @@ public class FormDefRecord extends Persisted {
         }
 
         // Set new values now
+        mFormFilePath = newFilePath;
         mFormMediaPath = getMediaPath(newFilePath);
         formDefRecordStorage.write(this);
     }
--- a/app/src/org/commcare/models/database/app/AppDatabaseUpgrader.java
+++ b/app/src/org/commcare/models/database/app/AppDatabaseUpgrader.java
@@ -3,6 +3,7 @@ package org.commcare.models.database.app;
 import android.content.Context;
 import android.database.Cursor;
 import android.net.Uri;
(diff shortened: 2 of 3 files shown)


ctsims:
@shubham1g5 Will kick off a review.

For tickets like these, btw, where something we were doing resulted in a bad state being created, it would be really helpful upfront in the PR to state:
 * What we were doing wrong (which you do here)
 * How the PR prevents this from happening for phones which haven't ended up in the bad state
 * Whether the PR also fixes devices which have ended up in the bad state, and whether it's by changing state (IE: "applies a one-time change to XXX which corrects devices which have updated already") or whether it's simply by changing some behavior (IE: "We were looking in the wrong place before, but now we are looking in the right place")

ctsims:
@shubham1g5 Looks good to me.

Can you confirm that you've tested both relevant paths here?

IE: 
1) Brick a phone as Clark did, apply this update, make sure form works
2) Apply Clark's steps _after_ updating to the new hotfix, make sure bad state is never created.

shubham1g5:
@ctsims  Makes sense. Added some extra info in description around what PR is doing to correct for corrupted state. Though in this case the first two PR description points kind of converges to the same thing. 

And yeah, have tested both these scenarios and are working good with this change.

PR #2024, opened 2018-07-25
Removes Storage Corrupt Dialog Warning

shubham1g5:
* Directly Navigates user to Recovery Screen instead of first showing him the storage corrupt warning
* Add "Go to App Manager" menu on revovery screen

### DIFF
--- a/app/res/values/strings.xml
+++ b/app/res/values/strings.xml
@@ -364,6 +364,7 @@
     <string name="incorrect_time_dialog_cancel" cc:translatable="true">Cancel</string>
 
     <!--Recovery Activity-->
+    <string name="recovery_mode_title">CommCare Recovery Mode</string>
     <string name="recovery_mode_detail">CommCare has entered recovery mode due to some problem with your app. Please make sure you have network access and attempt to recover the app.</string>
     <string name="recovery_mode_send_data_btn">Send Data</string>
     <string name="recovery_mode_attempt_recovery_btn">Attempt Recovery</string>
--- a/app/res/values/strings.xml
+++ b/app/res/values/strings.xml
@@ -364,7 +364,7 @@
     <string name="incorrect_time_dialog_cancel" cc:translatable="true">Cancel</string>
 
     <!--Recovery Activity-->
-    <string name="recovery_mode_title">CommCare has entered Recovery Mode due to a problem with the local device</string>
+    <string name="recovery_mode_detail">CommCare has entered recovery mode due to some problem with your app. Please make sure you have network access and attempt to recover the app.</string>
     <string name="recovery_mode_send_data_btn">Send Data</string>
     <string name="recovery_mode_attempt_recovery_btn">Attempt Recovery</string>
     <string name="recovering_resources_title">Recovering Missing Resources</string>
--- a/app/src/org/commcare/activities/RecoveryActivity.java
+++ b/app/src/org/commcare/activities/RecoveryActivity.java
@@ -157,13 +157,17 @@ public class RecoveryActivity extends SessionAwareCommCareActivity<RecoveryActiv
 
     @Override
     public boolean onCreateOptionsMenu(Menu menu) {
+        setUpActionBar();
+        return super.onCreateOptionsMenu(menu);
+    }
+
+    private void setUpActionBar() {
         ActionBar actionBar = getActionBar();
         if (actionBar != null) {
             actionBar.setHomeButtonEnabled(false);
             actionBar.setDisplayHomeAsUpEnabled(false);
-//            actionBar.setDisplayShowHomeEnabled(false);
+            actionBar.setTitle(getActivityTitle());
         }
-        return super.onCreateOptionsMenu(menu);
     }
 
     private boolean isNetworkAvaialable() {
@@ -265,7 +269,7 @@ public class RecoveryActivity extends SessionAwareCommCareActivity<RecoveryActiv
         // If app is no longer corrupt, launch home screen again
         if (!isAppCorrupt()) {
             Intent i = new Intent(this, DispatchActivity.class);
-            i.setFlags(Intent.FLAG_ACTIVITY_CLEAR_TASK | Intent.FLAG_ACTIVITY_NEW_TASK);
+            i.setFlags(Intent.FLAG_ACTIVITY_CLEAR_TASK | Intent.FLAG_ACTIVITY_CLEAR_TOP);
             startActivity(i);
         } else {
             super.onBackPressed();
@@ -315,5 +319,10 @@ public class RecoveryActivity extends SessionAwareCommCareActivity<RecoveryActiv
             return dialog;
         }
         return null;
+    }
+
+    @Override
+    protected String getActivityTitle() {
+        return StringUtils.getStringRobust(this, R.string.recovery_mode_title);
     }
 }
--- a/app/src/org/commcare/activities/RecoveryActivity.java
+++ b/app/src/org/commcare/activities/RecoveryActivity.java
@@ -6,6 +6,7 @@ import android.content.Intent;
 import android.content.SharedPreferences;
 import android.os.Bundle;
 import android.view.Menu;
+import android.view.MenuItem;
 import android.view.View;
 import android.view.View.OnClickListener;
 import android.widget.Button;
@@ -24,12 +25,14 @@ import org.commcare.utils.AndroidCommCarePlatform;
 import org.commcare.utils.CommCareUtil;
 import org.commcare.utils.ConnectivityStatus;
 import org.commcare.utils.FormUploadResult;
+import org.commcare.utils.Permissions;
 import org.commcare.utils.SessionUnavailableException;
 import org.commcare.utils.StorageUtils;
 import org.commcare.utils.StringUtils;
 import org.commcare.views.ManagedUi;
 import org.commcare.views.UiElement;
 import org.commcare.views.dialogs.CustomProgressDialog;
+import org.commcare.views.dialogs.DialogCreationHelpers;
 import org.javarosa.core.services.Logger;
 import org.javarosa.core.services.locale.Localization;
 
@@ -40,6 +43,8 @@ import org.javarosa.core.services.locale.Localization;
 public class RecoveryActivity extends SessionAwareCommCareActivity<RecoveryActivity> {
 
     private static final int RECOVERY_TASK = 10000;
+    private static final int MENU_APP_MANAGER = Menu.FIRST;
+
     @UiElement(R.id.screen_recovery_unsent_message)
     TextView txtUnsentAndQuarantineForms;
 
@@ -158,6 +163,7 @@ public class RecoveryActivity extends SessionAwareCommCareActivity<RecoveryActiv
     @Override
     public boolean onCreateOptionsMenu(Menu menu) {
         setUpActionBar();
+        menu.add(0, MENU_APP_MANAGER, 1, Localization.get("login.menu.app.manager"));
         return super.onCreateOptionsMenu(menu);
     }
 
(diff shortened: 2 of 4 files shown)


app/src/org/commcare/activities/RecoveryActivity.java
```diff
@@ -37,132 +37,141 @@
...
-                                } else if (result == FormUploadResult.AUTH_OVER_HTTP) {
-                                    receiver.displayMessage(Localization.get("auth.over.http"));
-                                }
-                            }
+                                            R.string.recovery_forms_send_failure) :
+                                    StringUtils.getStringRobust(
+                                            RecoveryActivity.this,
+                                            R.string.recovery_forms_send_parital_success,
+                                            String.valueOf(successfulSends));
+                            updateStatus(failureMessage);
```
ctsims: I believe all of these should be 

receiver.updateStatus() 

to handle rotation/re-attachment of tasks

app/src/org/commcare/activities/RecoveryActivity.java
```diff
@@ -37,132 +37,141 @@
...
 
-                            @Override
-                            protected void deliverError(RecoveryActivity receiver, Exception e) {
-                                Logger.exception("Error in recovery form send: " + ForceCloseLogger.getStackTrace(e), e);
-                                receiver.displayMessage(StringUtils.getStringRobust(receiver, R.string.recovery_forms_send_error) + ": " + e.getMessage());
-                            }
+                    @Override
+                    protected void deliverError(RecoveryActivity receiver, Exception e) {
+                        Logger.exception("Error in recovery form send: " + ForceCloseLogger.getStackTrace(e), e);
+                        updateStatus(StringUtils.getStringRobust(receiver, R.string.recovery_forms_send_error) + ": " + e.getMessage());
```
ctsims: same comment as above

app/src/org/commcare/activities/RecoveryActivity.java
```diff
@@ -37,132 +37,141 @@
...
+                        loadingIndicator.setVisibility(View.INVISIBLE);
+                    }
 
-                        };
+                };
 
-                mProcess.addSubmissionListener(CommCareApplication.instance().getSession().getListenerForSubmissionNotification());
+        mProcess.addSubmissionListener(CommCareApplication.instance().getSession().getListenerForSubmissionNotification());
+        mProcess.connect(RecoveryActivity.this);
```
ctsims: Hm, isn't this (and executeParallel below) now executable through a common mechanism rather than manually?

app/src/org/commcare/activities/RecoveryActivity.java
```diff
@@ -176,120 +185,111 @@ protected void deliverResult(RecoveryActivity recoveryActivity, Boolean success)
                                 // Try Reinitializing and attemptRecovery again to correct against further issues
                                 CommCareApplication.instance().initializeAppResources(CommCareApplication.instance().getCurrentApp());
                                 attemptRecovery();
+                                loadingIndicator.setVisibility(View.INVISIBLE);
                             } else {
-                                recoveryActivity.displayMessage(StringUtils.getStringRobust(recoveryActivity, R.string.recovery_error));
+                                onRecoveryFailure();
```
ctsims: Now that recovery is, well, a "real" process, I think it would be useful to bubble up more information about this if we can. If nothing else, knowing that recovery is failing due to network issues would be super useful.

app/res/values/strings.xml
```diff
@@ -364,28 +364,20 @@
...
-    <string name="recovery_mode_send_data_btn">Send Data</string>
-    <string name="recovery_mode_attempt_recovery_btn">Attempt Recovery</string>
-    <string name="recovering_resources_title">Recovering Missing Resources</string>
-    <string name="recovering_resources_progress">Installing missing resources...</string>
+    <string name="recovery_mode_title">CommCare Recovery Mode</string>
+    <string name="recovery_mode_detail">CommCare has entered recovery mode due to a problem with your app. We are trying to recover your app. Please wait...</string>
     <string name="recovery_resource_progress">Recovered %1$s out of %2$s resources</string>
-    <string name="recovery_error">We were not able to recover your app. Please make sure all forms are submitted and reinstall your app</string>
+    <string name="recovery_error">We were not able to recover your app. Please reinstall your app by going to App Manager</string>
     <string name="recovery_success">App recovery has been successful. Get back to work!</string>
```
ctsims: This is an old message, right? I think we should take out the cheeky "Back to work" part of it (I'm reminded of the cringe-y faux-casual messages Windows gives you when it dies losing all of your work)

shubham1g5: lol. No that was me trying to be cheeky and trying to prompt user to press back. I now think if recovery is successful we should instead show a toast and launch the DispatchActivity again instead of waiting for the user action ?

app/src/org/commcare/activities/RecoveryActivity.java
shubham1g5: Spend some time looking at our configuration retention code and had an Aha moment! Guess all accessors to the activity's methods and variables need to be modified, and not just the `updateStatus` calls.

app/res/values/strings.xml
ctsims: I think it is useful for users to see that the app recovery worked, actually (feedback during recovery operations is a good UX best practice) so I'd default to leaving the screen rather than auto-directing back. I do think it's probably useful to tell the user to press 'back' or provide a UI element to resume normal functioning).

And I'm definitely fine with cheeky messages elsewhere, but we've gotten into trouble in the past with light-hearted stuff being shown to users around errors or problems, since the tone feels really off when/if people are frustrated.

app/src/org/commcare/activities/RecoveryActivity.java
ctsims: Yeah, this is one of those frustrating Android things. The nice part about updateStatus type wrapper methods is that it prevents us from having to make repeated calls that are prefixed with the `receiver.` accessor, but basically we should treat the inside of deliverResult methods as if they were static and `receiver` is the only access it has to the surrounding context.

app/src/org/commcare/activities/RecoveryActivity.java
ctsims: Got it, sorry! Mis-remembered on my part.

shubham1g5:
@ctsims Did some testing with rotation and updated the PR with more changes -
1. Treats resources with `unitialized` status as missing to retry installing them during recovery. This is to protect against failures during install process which would leave resources as `unitialized`
2. Prevents duplication of tasks on rotation by only launching them once when `savedInstanceState` is `null`

shubham1g5:

shubham1g5:

shubham1g5:

ctsims:
Hi Shubham. Regarding your second point:

"Prevents duplication of tasks on rotation by only launching them once when savedInstanceState is null"

The task 'reconnect' framework should definitely be managing this for you without needing to manually track it. Is the issue that there are multiple potential tasks associated with the activity?

Similarly, I think 'savedInstanceState' isn't actually a safe enough signal to detect whether an activity is resumed. The API docs are full of warnings about the fragility of the related hooks.

We use the TaskConnectorFragment in CommCareActivity to provide a more reliable 'drop and reattach' consistency. It would be more helpful I think to shift any logic around background task management into the core framework and extending it if there are limitations to how it's handling this case.

shubham1g5:
@ctsims  The problem is even though the drop and reattach works fine, on rotation a duplicate task gets created since we are not checking if a task already exists before creating the task. 

If you think checking for `savedInstanceState==null` is not a strong enough signal that we are launching the activity first time and not getting recreated from a configuration change, other approach is to check whether there is a `currentTask` with specific taskId which already exists in `TaskConnectorFragment` . Though this check would need to be done in the top level activity code before we are creating the instance of the task and am a bit out of ideas how can we move it to the core framework.

ctsims:
Hey, makes sense.

So I think there are a few approaches that could make sense, here.

The more I think about it the more it feels like it's a bit sketchy to rely solely on an activity lifecycle to perform this check (making sure two things aren't running). It would be super bad for two of these tasks to ever be running at the same time (IE: manipulating the core resource table), and you could imagine this occurring even if the activity isn't rotated by, say, starting a task, swiping the activity left in the Android manager, then reloading CommCare. 

Phillip followed a slightly different pattern with the _update task_ which is more of a true 'background' task.

https://github.com/dimagi/commcare-android/blob/fd63a5b2471a1e33ac6bf1ce77ac82daf558b08a/app/src/org/commcare/tasks/UpdateTask.java

We may not need to follow that rigid of a structure, exactly, but we could use a similar locking pattern to accomplish similar things. IE: One option might actually be to just put a static lock around the inside of this async task.

Alternatively, I think it would be fine for CommCareActivity to support something like a conditional Task Create function that would let onCreate on onResume know whetehr to start a task based on the TaskConnector's current state.

shubham1g5:
@ctsims `ResourceRecoveryTask` uses a singleton pattern now. I tried managing it through `TaskConnector` but the problem is it's currently designed to hold a reference to only one single task whatever is currently running.  So in this case, since we run `RecoveryTask` after `sendFormsTask`, when we check whether another instance of `RecoveryTask` exists the `TaskConnectorFragment` already loses the `RecoveryTask` instance and holds the `sendFormsTask` reference instead.

app/src/org/commcare/activities/RecoveryActivity.java
```diff
@@ -37,259 +37,241 @@
...
-
-                        @Override
-                        protected void deliverError(RecoveryActivity recoveryActivity, Exception e) {
-                            Logger.exception("Error while recovering missing resources " + ForceCloseLogger.getStackTrace(e), e);
-                            recoveryActivity.displayMessage(StringUtils.getStringRobust(recoveryActivity, R.string.recovery_error));
-                        }
-                    };
+            ResourceRecoveryTask task = ResourceRecoveryTask.getInstance();
             task.connect(this);
             task.executeParallel();
```
ctsims: Does executeParallel crash out if the task is already running and/or handle it gracefully? 

Crashing out is still better than letting two execute at the same time, but flagging that if the getInstance returns an already existing instance it'll cause issues here.

I think the "Safest" pattern would be something like

```
task = ResourceRecoveryTask.getInstance();
task.connect()
if(taskstate = unstarted) { task.execute}

ResourceRecoveryTask.getInstanceAndConnect(connector) {
    lock {
          //since the lock code right now happens on the UI thread, the task can actually finish before it's disconnected
          if  ( exists.state is cancelled or complete) { exists = null}
          if  ( exists = null) { new exists() }
          return exists;
    }
}
```

does that make sense?

app/src/org/commcare/tasks/ResourceRecoveryTask.java
```diff
@@ -38,8 +53,55 @@ protected Boolean doTaskBackground(Void... voids) {
         return success;
     }
 
+    @Override
+    protected void onPostExecute(Boolean aBoolean) {
+        super.onPostExecute(aBoolean);
+        clearTaskInstance();
```
ctsims: Abstractly, I think this (and the call below) should be in a finally() {}, but all of the locking code right now is in areas that'd only execute on the UI thread, so it's hard for me to come up with a clear narrative where the onX() callbacks can crash and the code continues executing.

It may make more sense (based on the reasoning I mentioned above) for the clearing to actually happen in a finaly() at the end of execution, rather than in the callback. That would prevent the state of the task from changing in-between connection. You'd have to put the Activity attach() inside of the getNewIsntance() for that to work.

app/src/org/commcare/activities/RecoveryActivity.java
shubham1g5: Updated the `getNewInstance` to use the throw exception pattern Philip used for `UpdateTask`

app/src/org/commcare/tasks/ResourceRecoveryTask.java
shubham1g5: Just for context I was trying to use the same pattern `SingletonTask` [uses](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/tasks/SingletonTask.java#L38) . When you say we should move clearing at `end of execution` do you mean something like this - 

```
    @Override
    protected Boolean doTaskBackground(Void... voids) {
        try {
         // do work 
        } finally {
        clearTaskInstance();
       }
        return success;
    }
```

app/src/org/commcare/activities/RecoveryActivity.java
ctsims: That looks good to me, thanks for jumping back into that

app/src/org/commcare/tasks/ResourceRecoveryTask.java
ctsims: Ah, didn't realize that pattern was involved elsewhere. Let's stay consistent and manage it as outlined.

PR #2026, opened 2018-08-01
Removes incorrectly shown form lost warning on saving incomplete forms

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/browse/MOB-19

Something that I overlooked during FormProvider refactoring. This warning message was earlier shown when `FormEntryActivity` didn't return any form instance uri. Though since now FormProvider is merged into FormRecord, don't think there is any valid scenario for this error anymore.

### DIFF
--- a/app/src/org/commcare/activities/HomeScreenBaseActivity.java
+++ b/app/src/org/commcare/activities/HomeScreenBaseActivity.java
@@ -760,14 +760,6 @@ public abstract class HomeScreenBaseActivity<T> extends SyncCapableCommCareActiv
                 // Otherwise, we want to keep proceeding in order
                 // to keep running the workflow
             } else {
-                CommCareApplication.notificationManager().reportNotificationMessage(
-                        NotificationMessageFactory.message(
-                                NotificationMessageFactory.StockMessages.FormEntry_Unretrievable));
-                Toast.makeText(this,
-                        "Error while trying to read the form! See the notification",
-                        Toast.LENGTH_LONG).show();
-                Logger.log(LogTypes.TYPE_ERROR_WORKFLOW,
-                        "Form Entry did not return a form");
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/activities/HomeScreenBaseActivity.java
```diff
@@ -760,14 +760,6 @@ private boolean processReturnFromFormEntry(int resultCode, Intent intent) {
...
                 // to keep running the workflow
             } else {
-                CommCareApplication.notificationManager().reportNotificationMessage(
-                        NotificationMessageFactory.message(
-                                NotificationMessageFactory.StockMessages.FormEntry_Unretrievable));
-                Toast.makeText(this,
-                        "Error while trying to read the form! See the notification",
-                        Toast.LENGTH_LONG).show();
-                Logger.log(LogTypes.TYPE_ERROR_WORKFLOW,
-                        "Form Entry did not return a form");
```
ctsims: Seems odd to me to remove the error message but retain the code route. 

Is there a valid reason for execution to still hit this path without it being an error worth mentioning to the user? If not, I think we'd want to move the shift up the conditional chain and remove the if/else fork entirely (with the assumption that a hard error would happen somewhere else)

shubham1g5: Yeah, this code path is definitely valid since it gets hit if user saves a form as incomplete. This error message got mistakenly inroduced at this code path in Form Provider refactoring when it was originally meant for the case when `FormEntyActivity` didn't return a form instance content uri to Home Activity. Now after merging the `FormProvider` into `FormRecord`, we don't rely on `FormEntryActivity` to return any reference for Form Instance since we can reference that through the `session` in form of `FormRecord`

PR #2027, opened 2018-08-14
Use RGB_565 config for Signature Bitmaps

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/browse/MOB-18
 CL: https://www.fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/59caaf6bbe077a4dcc4b084c?time=last-ninety-days 
https://www.fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/59fa5c6561b02d480d9b5247?time=last-ninety-days

Problem: We are hitting OOMs frequently for some users on signature questions which is becoming a operational problem for them. 

Optimisation: RGB_565 uses 2 byte per pixel in comparison to 4 bytes per pixel used by ARG_8888. This results in less memory consumption when creating these bitmaps. In terms of quality, I couldn't see any difference from naked eye when comparing the 2 configs. Maybe since we are only using black and white colors it doesn't make much of a difference.

### DIFF
--- a/app/src/org/commcare/activities/DrawActivity.java
+++ b/app/src/org/commcare/activities/DrawActivity.java
@@ -221,7 +221,7 @@ public class DrawActivity extends Activity {
             FileOutputStream fos;
             fos = new FileOutputStream(f);
             Bitmap bitmap = Bitmap.createBitmap(drawView.getWidth(),
-                    drawView.getHeight(), Bitmap.Config.ARGB_8888);
+                    drawView.getHeight(), Bitmap.Config.RGB_565);
             Canvas canvas = new Canvas(bitmap);
             drawView.draw(canvas);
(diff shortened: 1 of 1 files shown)


ctsims:
This is definitely good to go (merged). I think one big reason this is likely coming up is that screens are super high-res now. I wonder if we could just set a maximum 'effective' size for the bitmap object and then the canvas will write it in as sampled?

One heads up: This widget came straight over from Opendatakit Collect, they may have a newer version of the widget that we can move over that has corrected for this issue.

shubham1g5:
@ctsims Checked out the ODK code and they have not changed anything when it comes to Bitmap handling. I agree that if it continues to be a problem we should incorporate image sampling to restrict the size.

PR #2029, opened 2018-08-21
Unifies Password Strategy to incorporate password obuscation property

shubham1g5:


app/src/org/commcare/network/HttpUtils.java
```diff
@@ -18,29 +18,23 @@
     public static String getCredential(@Nullable Pair<String, String> usernameAndPasswordToAuthWith) {
         String credential;
         if (usernameAndPasswordToAuthWith == null) {
```
shubham1g5: Earlier when `usernameAndPasswordToAuthWith` was passed as `null` we weren't using `buildAppPassword` to obfuscate the password.

shubham1g5:
@ctsims I am planning to build this as a hotfix (2.44.4) but not actually push the apk on playstore for the non LTS version.

app/src/org/commcare/network/HttpUtils.java
```diff
@@ -60,22 +54,11 @@ private static String buildDomainUser(String username) {
...
-        final String password = user.getCachedPwd();
-        return getCredential(username, password);
-    }
-
     private static String getCredential(String username, String password) {
         if (username == null || password == null) {
             return null;
         } else {
-            return okhttp3.Credentials.basic(username, password);
+            return okhttp3.Credentials.basic(buildDomainUser(username), buildAppPassword(password));
```
ctsims: Hey, 

So it was definitely intentional before that there was a version of this function which *didn't* try to append domain credentials. If you've audited and it's clear that we never used the domain-less version, fine to remove it, although I'd change the name of this function to something that makes it clear that it's doing double duty (IE: Creating the basic auth credential string *AND* mutating the username to attach domain data). 

I think in the context it's used, it wouldn't necessarily be unambiguous that the username you pass in here isn't the one that's going to come back out, and am worried it'll be a source of future errors.

shubham1g5: @ctsims Agree. Moved out the `buildDomainUser` out again since I now think it's better to not modify `username` inside the `getCredential` and let it just handle password obfuscation.

PR #2033, opened 2018-08-27
Session expiration crash fixes

shubham1g5:
CL: 
[1](https://www.fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/5a8e74bc8cb3c2fa63919665?time=last-ninety-days) , [2](https://www.fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/5b35f0126007d59fcd255e73?time=last-ninety-days), [3](https://www.fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/5a38128c8cb3c2fa632ab59f?time=last-ninety-days)

Better to catch these then crashing the app. This should ideally be a checked expression so that all code paths are required to handle this.

### DIFF
--- a/app/src/org/commcare/activities/StandardHomeActivity.java
+++ b/app/src/org/commcare/activities/StandardHomeActivity.java
@@ -5,6 +5,7 @@ import android.os.Bundle;
 import android.util.Log;
 import android.view.Menu;
 import android.view.MenuItem;
+import android.widget.Toast;
 
 import org.commcare.CommCareApplication;
 import org.commcare.google.services.analytics.AnalyticsParamValue;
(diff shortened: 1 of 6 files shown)


ctsims:
Hey Shubham,

We have a "SessionSafe" Activity pattern which is meant to catch these when they bubble up to the user level (rather than catching them deeper in the code, when we won't know how to react to them). Is there a reason that pattern isn't the best fit here? I'm really adverse to catching "Lifecycle" type exceptions anywhere other than at the top part of the lifecycle, IE: In activity v. in worker code.

shubham1g5:
Closing it for now to implement a better fix in future.

PR #2036, opened 2018-10-08
Quarantine Forms with Empty Paths

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/browse/MOB-31

Throwing FNF for forms with empty paths to eventually quarantine them. Earlier we were doing the same thing when a instance URI was empty for a FormInstance and hence this change is in line with the behaviour before the FormProvider refactor.


shubham1g5:
@ctsims Yeah agree with adding that check as well in another PR. My assumption right now is that this situation is only arising from migrating a bad FormInstance (validated from logs) but will be good to put some checks neverthless.

PR #2041, opened 2018-10-30
Disable the finish button after first click

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/projects/MOB/issues/MOB-38

There are 3 forms across 3 different users for which users were getting an error like - 
`Error during task execution: XML Syntax Error at Line: 1, Column: 251!]` on opening these forms from Saved Forms and these forms were also causing syncs to fail as well with 500 response code. 

Itnerpreting logs, there exists a loose pattern with all these 3 forms. For two users, Form Entry successfully completes for the respective forms but soon after (~2 mins) we get an error arising from form save code stack. The exact error being - 

`Error during task execution: Attempt to invoke virtual method 'void org.commcare.android.database.user.models.FormRecord.setFilePath(java.lang.String)' on a null object reference]`

This suggests that the saveTask got triggered twice one after the other and the second one failed since we had already wiped the formRecord from session as a result of successful completion of first instance. 

For another user, form entry again completes successfully but soon after (~20 secs) there is another log for successful form entry completion again suggesting that the save task got triggered twice and both of them succedded.

I am not exactly sure how duplicate instances of saveTask are corrupting the form and causing the 'XML Syntax Error' when user tries to open that form through Saved forms. 

After doing some testing the only reason that I can see SaveTask would have got triggered twice is user pressing the Finish button again at the same time when first task finishes. Since we don't disable the finish button after user presses it first time there is a small timeframe when user has the opportunity to click the finish button again at the same time when the Save Form dialog finishes. I was able to repro this and create another instane of Save Form Task which always fails for me wih the same NPE on `FormRecord.setFilePath` . 

Though I was not able to the repro the xml error arising because of this behaviour , think that it's arising as some consequence of duplication of save task.

### DIFF
--- a/app/src/org/commcare/tasks/SaveToDiskTask.java
+++ b/app/src/org/commcare/tasks/SaveToDiskTask.java
@@ -97,7 +97,6 @@ public class SaveToDiskTask extends
             return new ResultAndError<>(SaveStatus.SAVE_ERROR, cleanedMessage);
         }
 
-        FormEntryActivity.mFormController.setFormSaveTriggered(true);
         FormEntryActivity.mFormController.postProcessInstance();
 
         try {
@@ -124,6 +123,7 @@ public class SaveToDiskTask extends
             return new ResultAndError<>(SaveStatus.SAVE_ERROR, cleanedMessage);
         }
 
+        FormEntryActivity.mFormController.markFormSaveProcessComplete(true);
         if (exitAfterSave) {
             FormRecord saved = CommCareApplication.instance().getCurrentSessionWrapper().getFormRecord();
             Logger.log(LogTypes.TYPE_FORM_ENTRY,
--- a/app/src/org/commcare/tasks/SaveToDiskTask.java
+++ b/app/src/org/commcare/tasks/SaveToDiskTask.java
@@ -94,10 +94,10 @@ public class SaveToDiskTask extends
         } catch (XPathException xpe) {
             String cleanedMessage = "An error in your form prevented it from saving: \n" +
                     xpe.getMessage();
-
             return new ResultAndError<>(SaveStatus.SAVE_ERROR, cleanedMessage);
         }
 
+        FormEntryActivity.mFormController.setFormSaveTriggered(true);
         FormEntryActivity.mFormController.postProcessInstance();
 
         try {
@@ -216,7 +216,7 @@ public class SaveToDiskTask extends
             if (!instanceXml.delete()) {
                 Log.e(TAG,
                         "Error deleting " + instanceXml.getAbsolutePath()
-                        + " prior to renaming submission.xml");
+                                + " prior to renaming submission.xml");
                 return;
             }
 
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -713,6 +713,7 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
             }
             return;
         }
+
         // save current answer; if headless, don't evaluate the constraints
         // before doing so.
         boolean wasScreenSaved =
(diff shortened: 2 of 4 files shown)


ctsims:
Woah! Shubham! Awesome debugging, this is a super important find.

What happens here if the finish button fails? 

IE: You skip over a required question in the middle of the form with Go to Prompt and then hit finish and it jumps back?

Following our general paradigm of "UI shouldn't control safety" a slightly different change here could be to gate the SaveToDisk task more effectively within the formentryactivity.

We could, for example, save some variable that says "the form has definitely finished" and circumvent the second launch.

If you are having trouble replicating, can I suggest adding just a big sleep() loop right after whatever unblocks the SaveToDisk task from launching? I think this sounds like it might be a huge bug (given our recent issues with form processing records), and would be worth us getting deep in.

shubham1g5:
@ctsims 

> IE: You skip over a required question in the middle of the form with Go to Prompt and then hit finish and it jumps back?
Nice catch. I didn't really think of it. 

> If you are having trouble replicating, can I suggest adding just a big sleep() loop right after whatever unblocks the SaveToDisk task from launching?
SaveToDisk is triggered by the clickListener on Finish button. Can you explain a bit what unblocking SaveToDiskTask means here ?

ctsims:
@shubham1g5 

> > If you are having trouble replicating, can I suggest adding just a big sleep() loop right after whatever unblocks the SaveToDisk task from launching?
> 
> SaveToDisk is triggered by the clickListener on Finish button. Can you explain a bit what unblocking SaveToDiskTask means here ?

What I mean is that if there is a sequence of events that can result in finish() being clicked twice it will be evidenced by a period where something asynchronous is responsible for forcing the screen to become blocking. 

IE:

UI Thead: Kick-off task
Task Thread: Tell UI Thread to pop a modal
UI Thread: receive modal event and block the UI

The Ui Thread in between kicking off the task and polling and receiving the modal event will let the user click the button twice. Adding a Thead.sleep() into the task thread right before it tells the UI to block will let you trigger it more reliably 

It's also possible for the window to occur in the other direction if there are multiple callbacks to return control (IE: "Tell the UiThread to stop blocking" followed by "Tell the UiThread that save is complete and to move on" rather than those two events being handled in the same callback)

shubham1g5:
@ctsims So I am definitely able to repro 2 instances of SaveToDiskTask even without adding sleep by 
1. Cicking Finish very quickly one after the other
2. Clicking Finish at same time as the Blocking pop up finishes. 

What I am not able to repro is the XML error that occurs for those users as a consequence of two saves. I am not experimenting on my own domain though so it might be something speicifc to `enveritas`.

shubham1g5:
@ctsims 

> IE: You skip over a required question in the middle of the form with Go to Prompt and then hit finish and it jumps back?

This still works since Finish only become visible once form is complete. I agree that we should implement a UI decoupled change to handle this though do we wanna safe guard using UI as well by retaining the disable on finish button ?

shubham1g5:
@ctsims Made a different change for this. Please have a look.

app/src/org/commcare/tasks/SaveToDiskTask.java
```diff
@@ -94,10 +94,10 @@ public SaveToDiskTask(int formRecordId, int formDefId, String formRecordPath, Bo
         } catch (XPathException xpe) {
             String cleanedMessage = "An error in your form prevented it from saving: \n" +
                     xpe.getMessage();
-
             return new ResultAndError<>(SaveStatus.SAVE_ERROR, cleanedMessage);
         }
 
+        FormEntryActivity.mFormController.setFormSaveTriggered(true);
```
ctsims: is there a reason we set this here, rather than after the form is fully processed?

shubham1g5: @ctsims  Thought it will also provide protection from 2 instances of this task running together which potentially can result in 2 saves on the form.

ctsims: Oh, I meant just inserting this same call, but at the bottom of the method, since in theory the form could fail to save below? Or is that unrecoverable after the initial save passes above?

shubham1g5: Hmm that makes sense. I was more thinking about protecting the form save from two parallel SaveTask threads but looking at the code it doesn't look possible. Made the change.

PR #2042, opened 2018-10-31
Refactors ModernHttpTask to handle response stream

shubham1g5:
https://github.com/dimagi/commcare-android/pull/2038 introduced a bug caused by ModernHttpTask operating on a closed stream. This happened because `ModernHttpTask` was trying to read from the stream in `deliverResult` after the stream got processed and closed in `doTaskBackground` . This PR refactors the `ModernHttpTask` to not process the response in `doTaskBackground` but only in `deliverResult`. 

cross-request: https://github.com/dimagi/commcare-core/pull/827

### DIFF
--- a/app/src/org/commcare/tasks/ModernHttpTask.java
+++ b/app/src/org/commcare/tasks/ModernHttpTask.java
@@ -17,6 +17,8 @@ import java.util.HashMap;
 import javax.annotation.Nullable;
 
 import okhttp3.RequestBody;
+import okhttp3.ResponseBody;
+import retrofit2.Response;
 
 /**
  * Makes get/post request and delegates http response to receiver on completion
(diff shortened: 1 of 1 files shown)


shubham1g5:

ctsims:
@shubham1g5 One big question from my side: I thought it was the case that these requests wouldn't actually be handled until a response was triggered. IE: If the server takes 90 seconds to *produce* a response, won't this change force the Task to handle that I/O wait time in the synchronous part (deliverResult) rather than the asynchronous part (doTaskBackground)?

shubham1g5:
@ctsims the call to `makeRequest()` synchronously send the request and return its response. The processing in `deliverResult` is only the post processing of the response.

ctsims:
Got it. No objections on my side then, although I'm going to be 100% upfront with you that I'm not sure I can anticipate all of the implications of this change.

shubham1g5:
@ctsims  On reflecting back, think it makes more sense to have the IO to and from cache in the background method. Changed PR to reflect the same.

PR #2043, opened 2018-10-31
Don't allow Required DateWidget to be skipped without user changing them

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/browse/MOB-41

It's steming from the above issue but fixes a rather important anamoly where we were allowing users to skip a date question without them changing them. 

Product Note: Bug Fix: Not allows users to skip a required date question.


ctsims:
Hi Shubham,

I think this might lead to an odd user experience in the event that the date shown by default on this widget is actually already correct.

This basically makes the user interact with the widget even to set the date to today, right?

shubham1g5:
@ctsims that's right. Taking the conversation over Jira.

PR #2044, opened 2018-10-31
Update Gradle dependencies for latest AndroidStudio builds

ctsims:
Hey,

Had to make some changes to the gradle config to be compatible with the latest Android Studio setup before my builds could get going

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,6 +1,8 @@
 import org.apache.tools.ant.filters.ReplaceTokens
 import com.android.builder.core.DefaultManifestParser
 
+import java.util.function.BooleanSupplier
+
 ////////////
 // README //
 ////////////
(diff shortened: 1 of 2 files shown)


shubham1g5:
@ctsims Upgrading to AGP 3.2.1 is causing Robo tests to fail with our current robolectric version. On building with 3.2.1 , assets are put on a new path that the one Robolectric try to find them at and results in a FNF - `build/intermediates/bundles/commcare/debug/assets/locales/android_translatable_strings.txt (No such file or directory)` 

we have 2 choices here - either to spend some time in upgrading robolectric version to Robo 4 or revert these changes and include a single PR for both of them.

ctsims:
Let's revert for now and I'll re-make the price with the fixes in place.
Thanks for raising

PR #2049, opened 2018-11-12
Show Finish button on advancing to end of form

shubham1g5:
Refreshes view to show Finish button on advancing to end of form instead of triggering auto-save. 

Jira: https://dimagi-dev.atlassian.net/browse/MOB-41

Product Note: If there is a required date widget on the last screen in the form, earlier swiping ahead would auto-save the form. We have now changed this behaviour to instead show Finish button on swiping ahead and user needs to click on the Finish button to trigger save on the form.

### DIFF
--- a/app/src/org/commcare/activities/FormEntryActivityUIController.java
+++ b/app/src/org/commcare/activities/FormEntryActivityUIController.java
@@ -418,7 +418,7 @@ public class FormEntryActivityUIController implements CommCareActivityUIControll
                             break group_skip;
                         case FormEntryController.EVENT_END_OF_FORM:
                             // auto-advance questions might advance past the last form quesion
-                            activity.triggerUserFormComplete();
+                            refreshCurrentView();
                             break group_skip;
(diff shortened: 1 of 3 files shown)


shubham1g5:
@ctsims If you think we should look into preserving Auto-Save for Auto-Advance setups, can you please set up a quick form displaying Auto-Advance functionality.

ctsims:
Hi Shubham!

I made a form that should have demonstrated this behavior here:
https://www.commcarehq.org/a/swat/apps/view/4f507cb183d1562ad59513fc79a8ddd7/releases/

but on my device it looks like it already does what you outlined (Finish button pops up and you have to click it), so I think this is a safe change in behavior.

shubham1g5:

shubham1g5:
@ctsims Thanks! Didn't realize it's just the `quick` appearance attribute.  I was thinking that it's more a form level setting which gets applied to every question in the form.

PR #2050, opened 2018-11-14
Set all notifications to not make sound

shubham1g5:
`IMPORTANCE_DEFAULT` makes noise which I don't think we want. So setting priority for all notifications as `IMPORTANCE_LOW`


ctsims:
Hi Shubham!

On many devices I believe "low" priority also makes the notification collapsed by default, which I think would make the error dialogs even less visible to users. Can we set that particular notification back to default? I think it's fine if there's noise when the user runs across those erros since they are generally 'show-stopping'

shubham1g5:
@ctsims Makes sense. made that change.

PR #2054, opened 2018-11-27
Makes the AndroidPrototypeFactoryV1 Singleton to Improve App Upgrade time

shubham1g5:


shubham1g5:
@ctsims  No, this is a temporary factory which is only used while upgrading the DB. So it's only going to affect the update performance.

PR #2059, opened 2018-12-19
Error Message Improvement for Form Uploads for 406 reesponse code

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/projects/MOB/issues/MOB-47

Passes through HQ error messages to user in case of actionable failures (406) during form upload.

### DIFF
--- a/app/src/org/commcare/utils/FormUploadUtil.java
+++ b/app/src/org/commcare/utils/FormUploadUtil.java
@@ -15,9 +15,12 @@ import org.javarosa.core.io.StreamsUtil;
 import org.javarosa.core.io.StreamsUtil.InputIOException;
 import org.javarosa.core.model.User;
 import org.javarosa.core.services.Logger;
+import org.javarosa.core.services.locale.Localization;
 import org.javarosa.xml.ElementParser;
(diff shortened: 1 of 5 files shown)


app/src/org/commcare/utils/FormUploadUtil.java
```diff
@@ -204,17 +206,35 @@ private static FormUploadResult submitEntity(List<MultipartBody.Part> parts, Str
...
         } else {
             return FormUploadResult.FAILURE;
         }
     }
 
+    private static FormUploadResult processActionableFaiure(Response<ResponseBody> response) {
+        String message;
+        try {
+            JSONObject errorKeyAndDefault = new JSONObject(response.errorBody().string());
+            message = Localization.getWithDefault(
```
ctsims: I'm a bit confused by this line. Does HQ send down a _locale key_ with the error? That seems... really weird to me as a mechanism.

shubham1g5: @ctsims Yeah thats whats happening currently. HQ sends a locale key along with a default message.

PR #2060, opened 2018-12-24
Throws better error on resource file not found during install

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/browse/MOB-45

cross-request: https://github.com/dimagi/commcare-core/pull/834

### DIFF
--- a/app/src/org/commcare/android/resource/installers/FileSystemInstaller.java
+++ b/app/src/org/commcare/android/resource/installers/FileSystemInstaller.java
@@ -20,6 +20,7 @@ import org.javarosa.core.reference.InvalidReferenceException;
 import org.javarosa.core.reference.Reference;
 import org.javarosa.core.reference.ReferenceManager;
 import org.javarosa.core.services.Logger;
+import org.javarosa.core.services.locale.Localization;
 import org.javarosa.core.util.externalizable.DeserializationException;
 import org.javarosa.core.util.externalizable.ExtUtil;
 import org.javarosa.core.util.externalizable.PrototypeFactory;
@@ -87,7 +88,7 @@ abstract class FileSystemInstaller implements ResourceInstaller<AndroidCommCareP
                 inputFileStream = ref.getStream();
             } catch (FileNotFoundException e) {
                 // Means the reference wasn't valid so let it keep iterating through options.
-                return false;
+                throw new UnresolvedResourceException(r, Localization.get("install.error.file.not.found", r.getDescriptor()), true);
             }
 
             File tempFile = new File(CommCareApplication.instance().getTempFilePath());
--- a/app/src/org/commcare/android/resource/installers/FileSystemInstaller.java
+++ b/app/src/org/commcare/android/resource/installers/FileSystemInstaller.java
@@ -2,7 +2,9 @@ package org.commcare.android.resource.installers;
 
 import android.support.v4.util.Pair;
 
+import org.commcare.CommCareApp;
 import org.commcare.CommCareApplication;
+import org.commcare.dalvik.R;
 import org.commcare.engine.resource.installers.LocalStorageUnavailableException;
 import org.commcare.resources.model.MissingMediaException;
 import org.commcare.resources.model.Resource;
@@ -15,6 +17,7 @@ import org.commcare.util.CommCarePlatform;
 import org.commcare.util.LogTypes;
 import org.commcare.utils.AndroidCommCarePlatform;
 import org.commcare.utils.FileUtil;
+import org.commcare.utils.StringUtils;
 import org.javarosa.core.io.StreamsUtil;
 import org.javarosa.core.reference.InvalidReferenceException;
 import org.javarosa.core.reference.Reference;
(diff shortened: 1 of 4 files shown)


app/src/org/commcare/android/resource/installers/FileSystemInstaller.java
```diff
@@ -87,7 +88,7 @@ public boolean install(Resource r, ResourceLocation location,
                 inputFileStream = ref.getStream();
             } catch (FileNotFoundException e) {
                 // Means the reference wasn't valid so let it keep iterating through options.
-                return false;
+                throw new UnresolvedResourceException(r, Localization.get("install.error.file.not.found", r.getDescriptor()), true);
```
ctsims: This is actually a really big change to the process, since a URE is a high level exception rather than a low level one, and the file not found here is a procedural error (note the comment above that says the reference wasn't valid, so we should keep looping). 

This change may still be correct, but it's hard for me to judge based on this PR without knowing how else this code path might get hit.

I think this change would be unecessary if 

https://github.com/dimagi/commcare-core/pull/834/files#diff-a38548887050e5bdfb20728655fee4c0R432

included the r.getDescriptor() call in its text somewhere, since that will be the eventual upward call, and this throw doesn't include any context that isn't present in there other than "file not found" which isn't really a user-actionable issue (the JIRA ticket is more around "what resource is missing" rather than "what is the exact nature of the failure pertaining to that resource).

I think it would be a safer change to retain your change to ArchiveFileReference (IE: throwing a FNFE), letting this continue to return 'false') and then updating https://github.com/dimagi/commcare-core/pull/834/files#diff-a38548887050e5bdfb20728655fee4c0L425 to include more context.

ctsims: Oh, also: This part of the code gets executed when there often aren't seated Localization strings, so we should avoid translating errors with locale keys.

shubham1g5: @ctsims 
Today while installing the resource remotely, in case phone doesn't have proper internet or Multimedia file is missing, an `UnreliableSourceException` gets thrown which we don't consider a user facing exception and just show a general error saying "Application can't be found".  

That triggred me to throw a more high level exception here as URE which can take over the invisible `UnreliableSourceException` thrown during remote install. [This](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/commcare/resources/model/ResourceTable.java#L425) code path doesn't gets triggered today because of `UnreliableSourceException` being not null (thrown from remote install).

ctsims: I follow your process, I just think the URE throw part is unnecessary Due to your addition of the FileNotFound error (rather than I/O) error, this 'return false' will now trigger. Is the issue that the phone is then trying *again* and running across the URE?

If so, I'm signed off. I think it was initially intentional that an individual resource install could 'fall back' to another installer location, so this is a functional change in addition to an "Error message" producing change. I can't think of a scenario where a file resource installer should be OK to be null anymore, but if we can produce the error message just by throwing FNFE rather than throwing FNFE and throwing URE, I think that's a "safer" change.

shubham1g5: > Is the issue that the phone is then trying again and running across the URE?
Yes. After returning false from here, on trying the install for another location (Mostly remote) when the alternate location install attempt also fails due to an exception we pass that latest exception as a failure cause to the user. 

> I think it was initially intentional that an individual resource install could 'fall back' to another installer location
We are still falling back to other resource locations since we are catching this URE over [here](https://github.com/dimagi/commcare-core/pull/834#discussion-diff-244360061R417) and looping the locations untill we run out of locations or succeed. 
So there is a functional change though it seems opposite of what you mentioned.  Earlier in case of a URE, we would not fall back to alternate locations since we were not catching the URE anywhere in `findResourceLocationAndInstall` . Though now since we have introduced the catch, we will still make a install attempt using other locations. Is that something that was intended earlier ?

shubham1g5: @ctsims It feels like we are still saying opposite things. Earlier in case of an URE we were exiting the location loop (no catch URE) and hence not looping through locations. But the corresponding core PR changes this behaviour so that even if we encounter an URE, we will continue looping through locations.

ctsims: @shubham1g5 Yeah, that makes sense, I phrased my response poorly.

What I meant to say is "Looping through locations is the expected intent of the code, even if it was broken and wasn't happening, so changing that expectation is a proactive decision." It might be the right decision, I'm just not confident in my ability to forecast any downstream affects of doing so because it affects the code's intent rather than its execution.

It seems fine to continue looping with the caught URE because a failure in the remote location is an unreliablesourceexception, which is thrown later in the followup stack here

https://github.com/dimagi/commcare-core/pull/834/files#diff-a38548887050e5bdfb20728655fee4c0R429

so the local file error would be reported first, correct?

shubham1g5: @ctsims yeah, thats right! Makes more sense now. Thanks.

PR #2065, opened 2019-01-31
Fixing Entity Detail View to not cut content

shubham1g5:
Jira:https://dimagi-dev.atlassian.net/projects/MOB/issues/MOB-65

Setting lisview height to `match_parent` to tackle variable text lengths. This also will make Android to handle listview's children variable heights more gracefully. Currently the `getView()` for `EntityDetailAdapter` ends up getting called multiple times per listview row.

### DIFF
--- a/app/res/layout/component_entity_detail_item.xml
+++ b/app/res/layout/component_entity_detail_item.xml
@@ -48,7 +48,6 @@
             android:layout_marginRight="12dp"
             android:drawableLeft="@drawable/sym_action_call"
             android:drawablePadding="@dimen/content_min_margin"
-            android:inputType="phone"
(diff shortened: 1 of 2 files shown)


app/res/layout/component_entity_detail_item.xml
```diff
@@ -48,7 +48,6 @@
             android:layout_marginRight="12dp"
             android:drawableLeft="@drawable/sym_action_call"
             android:drawablePadding="@dimen/content_min_margin"
-            android:inputType="phone"
```
ctsims: This is the core change, right? Does this have meaningful changes outside of removing the one-line limitation? Clicking it still initiates a call, correct?

PR #2067, opened 2019-02-04
Shows 406 error messages from HQ while App update

shubham1g5:
Spec:https://docs.google.com/document/d/1xfOaDsV0gllRwZM2zWigc4BEHarDjPKuwbGYp50-Xt0/edit#heading=h.ty5i7ie0b8sj

This enables the functionality for HQ to pass mobile an error like - "You need to be logged in to do an app update". We are doing this to restrict app updates when user is logged out for Location based App Updates feature. 

PR also moves the common 406 parsing code from various places to `HTTPUtils`.

### DIFF
--- a/app/src/org/commcare/network/HttpUtils.java
+++ b/app/src/org/commcare/network/HttpUtils.java
@@ -78,13 +78,15 @@ public class HttpUtils {
 
     public static String parseUserVisibleError(Response<ResponseBody> response) {
         String message;
+        String responseStr = null;
         try {
-            JSONObject errorKeyAndDefault = new JSONObject(response.errorBody().string());
+            responseStr = response.errorBody().string();
+            JSONObject errorKeyAndDefault = new JSONObject(responseStr);
             message = Localization.getWithDefault(
                     errorKeyAndDefault.getString("error"),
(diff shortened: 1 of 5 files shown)


app/src/org/commcare/network/HttpUtils.java
```diff
@@ -68,4 +75,17 @@ private static String getCredential(String username, String password) {
...
+
+    public static String parseUserVisibleError(Response<ResponseBody> response) {
+        String message;
+        try {
+            JSONObject errorKeyAndDefault = new JSONObject(response.errorBody().string());
+            message = Localization.getWithDefault(
+                    errorKeyAndDefault.getString("error"),
+                    errorKeyAndDefault.getString("default_response"));
+        } catch (JSONException | IOException e) {
+            message = "Unknown issue";
```
ctsims: I wonder whether it would be a net positive in this case to just publish the errorBody itself as the error? A 406 is specific enough that I feel like we'll be happy in the event of an issue to be able to dig into the error itself.

shubham1g5: Nice point. Will update.

PR #2071, opened 2019-02-04
Fix for RtL layouts with audio buttons

ctsims:
The AudioButton was not properly sizing itself in RtL previously on at least some devices, probably because our layout declaration was non-deterministic.

image

Fixes so that the button has the same layout configuration as in LtR mode

image

For consistency: this is what the app currently (not with this PR's code) lays out for RtL audio buttons
image

I'm not really sure whether the placement of the round button in the box model is an issue, but it's consistent now.

### DIFF
--- a/app/res/layout/small_audio_playback.xml
+++ b/app/res/layout/small_audio_playback.xml
@@ -1,26 +1,24 @@
 <?xml version="1.0" encoding="utf-8"?>
 
-<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
-                android:layout_width="match_parent"
-                android:layout_height="match_parent"
-                android:orientation="vertical">
+<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="55dp"
+    android:layout_height="55dp">
 
     <ImageButton
         android:id="@+id/play_button"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
-        android:layout_centerInParent="true"
+        android:layout_gravity="center"
         android:background="@null"
-        android:src="@drawable/audio_playback_selector"/>
+        android:src="@drawable/audio_playback_selector" />
 
     <ProgressBar
         android:id="@+id/circular_progress_bar"
         style="?android:attr/progressBarStyleHorizontal"
         android:layout_width="55dp"
         android:layout_height="55dp"
-        android:layout_centerInParent="true"
         android:max="500"
         android:progress="0"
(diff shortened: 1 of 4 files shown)


app/src/org/commcare/views/media/MediaLayout.java
```diff
@@ -185,41 +185,47 @@ private void addAudioVideoButtonsToView(RelativeLayout questionTextPane) {
         // Add the audioButton and videoButton (if applicable) and view
         // (containing text) to the relative layout.
         if (audioButton != null) {
-            audioParams.addRule(RelativeLayout.ALIGN_PARENT_RIGHT);
-            textParams.addRule(RelativeLayout.LEFT_OF, audioButton.getId());
             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
                 audioParams.addRule(RelativeLayout.ALIGN_PARENT_END);
                 textParams.addRule(RelativeLayout.START_OF, audioButton.getId());
+            } else {
```
ctsims: All of these "else" clause moves are no-ops that are here to be more explicit about the layout.

app/src/org/commcare/views/media/MediaLayout.java
```diff
@@ -312,11 +318,19 @@ private void addElementsToView(View mediaPane,
             if (viewText.getVisibility() == GONE) {
                 this.addView(questionTextPane, questionTextPaneParams);
                 if (audioButton != null) {
-                    mediaPaneParams.addRule(RelativeLayout.LEFT_OF, audioButton.getId());
+                    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
+                        mediaPaneParams.addRule(RelativeLayout.START_OF, audioButton.getId());
```
ctsims: These aren't related but looks like they were missed before.

app/res/layout/small_audio_playback.xml
```diff
@@ -1,26 +1,24 @@
...
         android:background="@null"
-        android:src="@drawable/audio_playback_selector"/>
+        android:src="@drawable/audio_playback_selector" />
 
     <ProgressBar
         android:id="@+id/circular_progress_bar"
         style="?android:attr/progressBarStyleHorizontal"
-        android:layout_width="55dp"
-        android:layout_height="55dp"
-        android:layout_centerInParent="true"
```
ctsims: this is the actual fix. Since the layout is defining its width by this component, it can't also declare the center.

PR #2072, opened 2019-02-06
Fix for CommCare not getting filePath for Files selected from File Explorer 'Recent' option

shubham1g5:
Issue: Noticed that on a Lennovo phone, `getDataColumn` returns null for files selected from 'Recent' option in stock file provider. 

Proposed Change: if we are not able to get a non null filePath from a file provider uri, returns the original uri for further processing (making a internal copy of the file using uri which seems to be the best optionto protect against such failures). 

Product Note: Fixes an issue where CommCare won't be able to detect a file selected from Recent folder of Android File Explorer.


ctsims:
Ooh, this might fix a lot of these that have come up in different explorers, thanks for catching!

PR #2074, opened 2019-02-11
Form Entry Menu Alignment Fix for RTL

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/browse/QA-171

In RTL, form entry menu was not getting aligned to the right on some devices. Moving the menu creation code to `onCreateOptionsMenu` from `onPrepareOptionsMenu` seems to fix the issue.

### DIFF
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -446,7 +446,7 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
     public boolean onCreateOptionsMenu(Menu menu) {
         if (CommCareApplication.instance().isConsumerApp()) {
             // Do not show options menu at all if this is a consumer app
-            return super.onPrepareOptionsMenu(menu);
+            return super.onCreateOptionsMenu(menu);
         }
 
         super.onCreateOptionsMenu(menu);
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -442,36 +442,42 @@ private boolean processIntentResponse(Intent response, boolean wasIntentCancelle
         return wasAnswerSet;
     }
 
-
     @Override
-    public boolean onPrepareOptionsMenu(Menu menu) {
+    public boolean onCreateOptionsMenu(Menu menu) {
         if (CommCareApplication.instance().isConsumerApp()) {
             // Do not show options menu at all if this is a consumer app
             return super.onPrepareOptionsMenu(menu);
```
ctsims: shouldn't this be onCreate now that it's moved?

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -442,36 +442,42 @@ private boolean processIntentResponse(Intent response, boolean wasIntentCancelle
         return wasAnswerSet;
     }
 
-
     @Override
-    public boolean onPrepareOptionsMenu(Menu menu) {
```
ctsims: Heads up that the reason this is in "onPrepare" is so that it gets rebuilt when langauges change. If it's moved to onCreate() those triggers no longer fire each time, so the langauges won't update

app/src/org/commcare/activities/FormEntryActivity.java
shubham1g5: Thanks for catching.

app/src/org/commcare/activities/FormEntryActivity.java
shubham1g5: Makes sense. Added a call to `invalidateOptionsMenu` on language change so that `onCreateOptionsMenu` gets called again.

PR #2075, opened 2019-02-11
Revert "menu list item now expands on whole screen"

shubham1g5:
This reverts [commit](https://github.com/dimagi/commcare-android/commit/a25a990c620bf4325bbd19dadaf661403be03898#diff-b7391c239a63a97341bf79e75211e6da).

This was done to align the audio button in the menu list. Though I would assume ShrinkingLayout was intentional here. I missed it during RTL review and noticed only now. So wants to get it cross-checked if it's alright to remove ShrinkingLayout here. 

cc: @rristovic

### DIFF
--- a/app/res/layout/menu_list_item_modern.xml
+++ b/app/res/layout/menu_list_item_modern.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<LinearLayout
+<org.commcare.views.ShrinkingLinearLayout
     xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     android:layout_width="match_parent"
@@ -49,8 +49,7 @@
         android:id="@+id/row_txt"
         style="@style/ListContentWithoutBadge"
         android:layout_height="@dimen/list_text_height"
-        android:layout_width="0dp"
-        android:gravity="start"
+        android:layout_width="wrap_content"
         android:layout_weight="1"
         android:ellipsize="end"
         android:maxLines="2"/>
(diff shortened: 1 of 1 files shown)


rristovic:
I believe Clayton had a different design in mind when he created ShrinkingLayout. Without it, everything behaves as requested.

shubham1g5:
@ctsims I don't see any issues with removing it as well.  So If you don't know of any legacy reasons behind this, I am going to close this PR.

ctsims:
@shubham1g5 Do you have a screenshot that illustrates this difference, btw? 

The shinking layout may be relied on by some folks, and I'm still not sure I have an impression of what's failing here.

These modules:
https://www.commcarehq.org/a/ccqa/apps/view/7ba9b150718448c442aa0ab3efd53995/module/26a335f56c62e12092fa76de4ff5ad3565303a32/

in the CCQA app demonstrate the old behavior  where the font would switch from the larger size to the smaller size to try to fit more on the screen

shubham1g5:
@ctsims  that menu appears exactly the same with and without Shrinking layout - 
screenshot_20190215-145113

ctsims:
Is that what it looks like on 2.44? In theory the ShrinkingTextView should be making the font smaller since the text is overrunning the bounds.

shubham1g5:
Hey yeah, the screenshot is for 2.44. Looking at [this condition](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/views/ShrinkingLinearLayout.java#L63) , it seems like it will only do anything when the text is actually smaller and not larger than the available space to probably make extra space available to other views.

ctsims:
oh, I must have been wrong about what this particular view's role is. Thanks for clarifying, i'm all signed off

PR #2077, opened 2019-02-12
Makes xforminstaller accounts for  form id changes during update

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/browse/ICDS-343

#### Problem

In case of an update containing an existing form with changed resource id, during staging the update table to replace master, we end up giving it the `formid` for the existing entry for existing form with same namespace in `FormDefStorage`. This existing form's xml file gets removed when we uninstall this existing form. Though since the new form is still linked to old form's path,  it results in the "Error Reading XForm" error on opening that form. 

#### Solution

Instead of linking the form resource in update to the existing master form, we wanna treat it as a completely new resource and hence give it a new row in FormDef Storage. So Irrespective of the fact that a form already exists with the same namespace, we add a new form def row (hence new `formid` ) in the FormDefStorag whenver `upgrade` is set to false i.e. we have a new resource id for that form. Also during uninstall of old resource no longer in the update, we remove the form def row for the existing form so that we don't end up with 2 forms with same namespace.

### DIFF
--- a/app/src/org/commcare/android/database/app/models/FormDefRecordV12.java
+++ b/app/src/org/commcare/android/database/app/models/FormDefRecordV12.java
@@ -24,7 +24,12 @@ import static org.commcare.android.database.app.models.FormDefRecord.META_JR_FOR
 import static org.commcare.android.database.app.models.FormDefRecord.META_MODEL_VERSION;
 import static org.commcare.android.database.app.models.FormDefRecord.META_UI_VERSION;
 
-// Model for Form Def Record pre App DB v13
+/**
+ * Represents the version of a FormDefRecord that exists on any devices running a pre-2.47
+ * version of CommCare, which was deprecated in app db version 13. This class is used to read a
+ * FormDefRecord that exists in such a database, in order to run a db upgrade.
+ *
+ */
 @Table(FormDefRecord.STORAGE_KEY)
 public class FormDefRecordV12  extends Persisted {
 
--- a/app/src/org/commcare/android/database/app/models/FormDefRecordV12.java
+++ b/app/src/org/commcare/android/database/app/models/FormDefRecordV12.java
@@ -1,0 +1,150 @@
+package org.commcare.android.database.app.models;
+
+import android.database.Cursor;
+import android.database.SQLException;
+
+import org.apache.commons.lang3.StringUtils;
+import org.commcare.android.storage.framework.Persisted;
+import org.commcare.models.database.SqlStorage;
+import org.commcare.models.framework.Persisting;
+import org.commcare.modern.database.Table;
+import org.commcare.modern.models.MetaField;
+import org.commcare.provider.FormsProviderAPI;
+import org.commcare.util.LogTypes;
+import org.commcare.utils.FileUtil;
+import org.javarosa.core.services.Logger;
+
+import java.io.File;
+import java.io.IOException;
+
+import static org.commcare.android.database.app.models.FormDefRecord.META_DISPLAY_NAME;
+import static org.commcare.android.database.app.models.FormDefRecord.META_FORM_FILE_PATH;
+import static org.commcare.android.database.app.models.FormDefRecord.META_FORM_MEDIA_PATH;
+import static org.commcare.android.database.app.models.FormDefRecord.META_JR_FORM_ID;
+import static org.commcare.android.database.app.models.FormDefRecord.META_MODEL_VERSION;
+import static org.commcare.android.database.app.models.FormDefRecord.META_UI_VERSION;
+
+// Model for Form Def Record pre App DB v13
+@Table(FormDefRecord.STORAGE_KEY)
+public class FormDefRecordV12  extends Persisted {
+
+    @Persisting(1)
+    @MetaField(META_DISPLAY_NAME)
+    private String mDisplayName;
+
+    @Persisting(2)
+    @MetaField(META_JR_FORM_ID)
+    private String mJrFormId;
+
+    @Persisting(3)
+    @MetaField(META_FORM_FILE_PATH)
+    private String mFormFilePath;
+
+    @Persisting(4)
+    @MetaField(META_FORM_MEDIA_PATH)
+    private String mFormMediaPath;
+
+    @Persisting(value = 5, nullable = true)
+    @MetaField(META_MODEL_VERSION)
+    private int mModelVersion = -1;
+
+    @Persisting(value = 6, nullable = true)
+    @MetaField(META_UI_VERSION)
+    private int mUiVersion = -1;
+
+    // Serialization Only!
+    public FormDefRecordV12() {
+    }
+
+    // Only for DB Migration
+    public FormDefRecordV12(Cursor cursor) {
+        mDisplayName = cursor.getString(cursor.getColumnIndex(FormsProviderAPI.FormsColumns.DISPLAY_NAME));
+        mJrFormId = cursor.getString(cursor.getColumnIndex(FormsProviderAPI.FormsColumns.JR_FORM_ID));
+        mModelVersion = cursor.getInt(cursor.getColumnIndex(FormsProviderAPI.FormsColumns.MODEL_VERSION));
+        mUiVersion = cursor.getInt(cursor.getColumnIndex(FormsProviderAPI.FormsColumns.UI_VERSION));
+        mFormMediaPath = cursor.getString(cursor.getColumnIndex(FormsProviderAPI.FormsColumns.FORM_MEDIA_PATH));
+        mFormFilePath = cursor.getString(cursor.getColumnIndex(FormsProviderAPI.FormsColumns.FORM_FILE_PATH));
+    }
+
+    public void updateFilePath(SqlStorage<FormDefRecordV12> formDefRecordStorage, String newFilePath) {
+        checkFilePath(newFilePath);
+        File newFormFile = new File(newFilePath);
+        try {
+            if (new File(mFormFilePath).getCanonicalPath().equals(newFormFile.getCanonicalPath())) {
+                // Files are the same, so we may have just copied over something we had already
+            } else {
+                // New file name. This probably won't ever happen, though.
+                FileUtil.deleteFileOrDir(mFormFilePath);
+            }
+        } catch (IOException ioe) {
+            //we only get here if we couldn't canonicalize, in which case we can't risk deleting the old file
+            //so don't do anything.
+        }
+
+        // Set new values now
+        mFormFilePath = newFilePath;
+        mFormMediaPath = getMediaPath(newFilePath);
+        formDefRecordStorage.write(this);
+    }
+
+    private static void checkFilePath(String formFilePath) {
+        if (StringUtils.isEmpty(formFilePath)) {
+            throw new IllegalArgumentException("formFilePath can't by null or empty");
+        }
+    }
+
+
+    public int save(SqlStorage<FormDefRecordV12> formDefRecordStorage) {
+        // if we don't have a path to the file, the rest are irrelevant.
+        // it should fail anyway because you can't have a null file path.
+        if (StringUtils.isEmpty(mFormFilePath)) {
+            Logger.log(LogTypes.SOFT_ASSERT, "Empty value for mFormFilePath while saving FormDefRecord");
+        }
+
+        // Make sure that the necessary fields are all set
+        File form = new File(mFormFilePath);
+        if (StringUtils.isEmpty(mDisplayName)) {
+            mDisplayName = form.getName();
+        }
+
+        if (StringUtils.isEmpty(mFormMediaPath)) {
+            mFormMediaPath = getMediaPath(mFormFilePath);
+        }
+
+        formDefRecordStorage.write(this);
+
+        if (recordId == -1) {
+            throw new SQLException("Failed to save the FormDefRecord " + toString());
+        }
+        return recordId;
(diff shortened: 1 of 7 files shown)


ctsims:
Hi Shubham,

Couple of questions

1) Does this change imply that there needs to be more work in the uninstall action for the old XForm, to ensure that it doesn't leave a stub record around (previously, the code would have simply overwritten said record, correct?)

Similarly - how does this interact with rollbacks?

2) I'm not sure I actually quite see why this works. In the event of an upgrade, I presume it's the case that the `upgrade` flag isn't actually set to `true` until we're doing the "Upgrade install" where the form from the upgrade table is seated?

Finally - I thought the issue with HQ was that the `resource_id` for the form resource changed in between versions, not the *schema* ID. Doesn't this process then result in two form records being written which have the same schema, or is the assumption that the process of removing the old form clears out the second XMLNS?

shubham1g5:
@ctsims 🤦‍♂️ Clearly I didn't think it through. Going to think of an alternate approach and PR it again.  Closing for now.

ctsims:
No worries, this is a tricky one. Definitely my fault for having made so many assumptions about this configuration with the installation paths a long time ago :/ .

shubham1g5:
@ctsims In retrospect, this seems to be the right change for this issue to me. It makes it so that the form with the changed resource id gets treated as a new form altogether. I am answering your earlier questions regarding this PR below - 

> Does this change imply that there needs to be more work in the uninstall action for the old XForm, to ensure that it doesn't leave a stub record around (previously, the code would have simply overwritten said record, correct?)
This is right. Just the change here will result in 2 form def records with same xmlns in the storage though since we update the xmlns to formdef mapping after installing the new form (form with changed resource id) , the stub doesn't create issues in normal form workflows. We should definitely do additional work to remove the stub record for the old Form which was something I overlook last time. 

> I'm not sure I actually quite see why this works. In the event of an upgrade, I presume it's the case that the upgrade flag isn't actually set to true until we're doing the "Upgrade install" where the form from the upgrade table is seated?
the `upgrade` flag here is just set on basis of a resource id match and is set to true if a resource with same id exists in the master table. You are right that it's only set to true during "Upgrade Install". 
Though how this solution works is that while "Upgrade Install" in case of a resource id unmatch, upgrade is set to false and therefore we treat the incoming form as a new form and makes a new Form Def entry for it irrespective of whether any other form with same form xmlns is present in storage. 

I can open this branch up again and continue with the remaining work of removing the stub from def entry if you think the overall approach is right here ?

ctsims:
Hi Shubham,

Got it, I think I was missing a few of the points in your explanation. Two quick questions for you on continuing this approach:

1) Is it true that the old approach would have caught the error of installing two XForms with the same XMLNS and not let the install proceed, and that this change will remove that consistency check? (It is likely still worth it, but wanted to verify my understanding of the purpose of this code)
2) I agree with you that it makes sense structurally to decouple the FormDef/XMLNS level registration from the Asset level (xml document/resource installer) element. Does the installer object have enough context that both XformInstallers (old form, new form) can properly manage those independently? 

It looks like your recent huge refactor of the database objects for the form definition records has made this whole process significantly more straightforward. Awesome job, and great to see it paying dividends into the future!

shubham1g5:
@ctsims I have updated the PR based on our conversation and it's ready to be reviewed now as a complete solution to this issue. I have updated the description as well to be more detailed. Let me know if there are any lingering questions on your mind that can pose some risks. Also I am not sure if your questions in last comment are still relevant after our chat.

app/src/org/commcare/android/resource/installers/XFormAndroidInstaller.java
```diff
@@ -91,21 +91,20 @@ protected int customInstall(Resource r, Reference local, boolean upgrade, Androi
         }
 
         Vector<Integer> existingforms = FormDefRecord.getFormDefIdsByJrFormId(platform.getFormDefStorage(), formDef.getMainInstance().schema);
-        if (existingforms != null && existingforms.size() > 0) {
```
shubham1g5: We are removing this check to decide whether we wanna create a new form def entry or not and making this decision based on the `upgrade` flag now which conveys whether we have an exisiting resource with the same id in the master table.

shubham1g5:

shubham1g5:

ctsims:
Hi Shubham,

No objections to the code changes, but would be good to get a bit of a feel for the testing approach here, as I think we established a couple of things we wanted to make sure were effective.

There are a few example project spaces/builds that replicate potential issues with this setup here that can be tested against 
https://docs.google.com/document/d/1PtIpLZsmUqVmkK_umcbI9c6CUCCXq2sYQSgsf14YJkg/edit

It would be good to validate that scenarios that broke before are now fixed, but also good to validate that nothing unexpected is occurring with, say, the scenarios that work today (IE: Updating from the 'broken chain' version to the new 'universally fixed' version)

shubham1g5:
@ctsims  Hey, during testing I realized that there is a flaw with removing the Form Def record in the uninstall approach. 

The `uninstall` gets called even for forms who are in both app versions but have got updated since currently installed version i.e if we have a Form F which got updated from version1  to version8 in the update, we call uninstall on F(v1) to clear the version1 form xml. Therefore we can't remove the Form Def Entry record for this form F in uninstall since doing so will remove the only Form Def record for F. 

**Side note** that I think there is a bigger flaw here that Form Def table doesn't tie with Resource table's corresponding form entries and therefore it's currently possible for these 2 tables to get out of sync with each other. For example, today when we install an app update in the UPGRADE table, we also make entries in the `FormDefRecord` table as part of this process though if this upgrade fails these entries form `FormDefRecord` doesn't get deleted. 

Coming back to this PR, I have reverted the removal of `FormDefRecord` entry in `uninstall` on this PR. The test scenarios you mentioned works well with this PR currently with the flaw that the Form Def table get duplicate entries for same xmlns whenever this issue happens which is not ideal. 

Wondering if you have any thoughts on doing a more laid-back cleanup of FormDefTable to avoid this duplication i.e. doing a pass on `FormDefRecord` Table on [app initialization](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/commcare/util/CommCarePlatform.java#L132) and deleting the older record in cases where there are 2 records with the same xmlns in the table.

ctsims:
One option would be to add some indicator in the FormDefRecord table which indicates the status of the record a bit more clearly. It sounds like the fundamental issue is mostly that when we do an upgrade right now for a FormDef, we re-use the FormDefRecord, rather than installing a second record and letting it supplant the original one by removing it.

I think there are two ways forward in that case
1) Revise the form installer to have one FormDefRecord per FormInstaller, rather than per Form Resource ID. This is hairy because we need a way to handle both the future case as well as any already-installed/updated FormInstallers that haven't been doing this.
2) Add metadata to the FormDefRecord which keeps track of some details like the installed version, or the linked resourceinstaller. This would let us do a check on uninstall like

* Read the FormDefRecord
* See if it has the same resourceID/version as the installer record
* If so, remove the record and form
* Otherwise (say, the installer version Id is 20 and the installed FormDefRecord is 50), presume the record is associated with a newer installer and leave it alone.

shubham1g5:
@ctsims Implemented the form def versioning changes here. The process now looks like this - 

1. During XformInstall, always insert a new record in the Form Def table with the corresponding resource version. 

2. During uninstall , check whether resource version is same as form def version and if so, remove the form def record. 

I tested that these 2 changes works for all scenarios except one - 

1. There is an app update out which changes the resource id of an existing form. 
2. User before taking CommCare update, stages this update but don't install the update. 
3. User update CommCare that includes this PR
4. User Install the update on login and go to the Form to experience the same error. 

Since update was already staged before user took CommCare update, that update would have overwritten the existing record in the form def table already and then updated uninstall code though would not remove the form def record entry, it still removes the file path for the form getting uninstalled. To incorporate this, does it make sense to skip the file removal for any old forms in uninstall more specifically change XFormAndroid uninstall `uninstall` to - 
```
// Only call super.uninstall (FileSystemInstaller) when it's a form that got added after CommCare update. 
if (formDefRecord.getResourceVersion() != -1) {
 super.uninstall(...)
}
```

I realize that this condition will cause some old form files to not get deleted from file system, but I could not think of a way to separate out the form for which resource id got changed.

ctsims:
@shubham1g5 My only concern about leaving a file on the filesystem is the potential overlap with incoming binaries. 

Essentially: If a new resource was installed later, is there a chance it would overlap with the FS resource, and if so does it handle that gracefully by, say, giving the new file a slightly different name. 

I'm definitely fine with leaving the file around in this edge case. XForm files themselves are small on disk compared to other resources, our approach already accepts some amount of 'leaving old records we'd rather have removed' as in the rows in the formdefrecord DB we couldn't find a way to clean, and it's better than the alternative of potentially bricking the installs.

shubham1g5:
@ctsims Today filenames for form xml are same as resource id for that form i.e a form with resource id 'x' will have a filename `x.xml`. I am not really sure how resource ids are generated and if there is a scenarios where they can be same for 2 resources.

ctsims:
The circumstance I'd be concerned about would be something like a rollback, IE:

V21 - Original
V22 - Form resource id X added
V23 - rollback to v21 (say, due to an accidental rollout), so Form resource X is removed from the local app install
V24 - Rollback to V22. Form X re-added to the app with the original resource ID.

For some resources I've seen the app react correctly if there's an overlap by, say, making a resource called X.xml.1, if that happens here it should be fine

shubham1g5:
@ctsims  Thanks, I tried that out and it seems that the form paths for same resource id does overlap though the new form file does replace the old one nevertheless on update.  Is that fine or we should still be safe here and change the file naming logic ?

ctsims:
Ok, that should work based on everything you've described, so I think that's a reasonable approach.

shubham1g5:
@ctsims  I take my last comment back. I got a bit delusional I think.  As you mentioned earlier What's happening indeed is we are indeed [adding a number to the original file name](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/android/resource/installers/FileSystemInstaller.java#L165) in case the file already exists. We then map this path correctly in the formDef record so I think it's still should not be an issue. 

Also the PR is now ready to be reviewed again. I have tested this with various permutations and it seems to correctly work without breaking the app.

app/src/org/commcare/android/database/app/models/FormDefRecordV12.java
```diff
@@ -0,0 +1,155 @@
...
+ * Represents the version of a FormDefRecord that exists on any devices running a pre-2.47
+ * version of CommCare, which was deprecated in app db version 13. This class is used to read a
+ * FormDefRecord that exists in such a database, in order to run a db upgrade.
+ *
+ */
+@Table(FormDefRecord.STORAGE_KEY)
+public class FormDefRecordV12  extends Persisted {
+
+    @Persisting(1)
+    @MetaField(META_DISPLAY_NAME)
```
ctsims: Quick note that it's generally better to inline these as raw strings (IE: META_DISPLAY_NAME -> "displayName") so that a change to the static definition won't affect the migration path.

shubham1g5: Good point 👍

shubham1g5:
@ctsims  Made the change to explicitly look for the form def with hightest resource version.

> I also remember us having a brief conversation about writing the version # of the resource back into the FormDefRecord during unstage/repair and the update application (since the resource is available there), but I don't remember whether we decided that caused issues.

Hey yeah, I think that was valid when we were overwriting the same form def record in the old solution. But now since we are always creating a new Form Def row, the resource version # for the original record never gets modified during an update making it unnecessary to rollback resource version. In case of an update failure, we call clear on the update table resources which should uninstall the new fomdefRecord row added during the update.

app/src/org/commcare/android/database/app/models/FormDefRecordV12.java
ctsims: nice, good idea!

app/src/org/commcare/android/resource/installers/XFormAndroidInstaller.java
```diff
@@ -71,10 +71,33 @@ public boolean initialize(AndroidCommCarePlatform platform, boolean isUpgrade) t
...
         super.initialize(platform, isUpgrade);
-        platform.registerXmlns(namespace, formDefId);
+        if (isLatestFormRecord(platform)) {
+            platform.registerXmlns(namespace, formDefId);
+        }
         return true;
     }
 
+    // Returns whether the associated formdefRecord is the one with highest form def resource version for our namespace
+    private boolean isLatestFormRecord(AndroidCommCarePlatform platform) {
```
ctsims: From a consistency / safety perspective, I think it's dangerous to make this a per-resource question like this, since it's hard in a closed-form way to guarantee that there *is* at least one form record per XMLNS.

From you and I's breakdown I think we determined that there's only ever two (or a fixed number) of forms that have the same XMLNS possible anyway, so I'd structure this method differently to return *a* valid form ID (probably the one below) and then check it against this one. 

It's a small, stylistic point, but I think right now it's possible for someone to make an easy mistake that results in there being no forms that think they are the latest (due to nested rules).

shubham1g5: @ctsims Made a change for this though not totally sure if that's what you meant. I am merging it anyway after testing, though feel free to review when you get back online.

PR #2079, opened 2019-02-18
Always format Decimal Widget as English 

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/browse/QA-186

When we format numbers in Arabic, they get converted to Arabic numbers which then throws an error on getting parsed as double. 

This PR sets the language to english for formatting the decimals. Moreover, it also removes the ew-conversion from formatted string to Double to avoid showing decimals in scientific/exponential notation.

### DIFF
--- a/app/src/org/commcare/views/widgets/DecimalWidget.java
+++ b/app/src/org/commcare/views/widgets/DecimalWidget.java
@@ -29,6 +29,7 @@ import org.javarosa.core.model.data.IAnswerData;
 import org.javarosa.form.api.FormEntryPrompt;
 
 import java.text.NumberFormat;
+import java.util.Locale;
 
 /**
  * A widget that restricts values to floating point numbers.
@@ -60,7 +61,7 @@ public class DecimalWidget extends StringWidget {
             d = (Double)getCurrentAnswer().getValue();
         }
 
-        NumberFormat nf = NumberFormat.getNumberInstance();
+        NumberFormat nf = NumberFormat.getNumberInstance(Locale.ENGLISH);
         nf.setMaximumFractionDigits(15);
         nf.setMaximumIntegerDigits(15);
         nf.setGroupingUsed(false);
--- a/app/src/org/commcare/views/widgets/DecimalWidget.java
+++ b/app/src/org/commcare/views/widgets/DecimalWidget.java
@@ -61,14 +61,7 @@ public class DecimalWidget extends StringWidget {
             d = (Double)getCurrentAnswer().getValue();
         }
 
-        NumberFormat nf = NumberFormat.getNumberInstance(Locale.ENGLISH);
-        nf.setMaximumFractionDigits(15);
-        nf.setMaximumIntegerDigits(15);
-        nf.setGroupingUsed(false);
         if (d != null) {
-            Double dAnswer = (Double)getCurrentAnswer().getValue();
-            String dString = nf.format(dAnswer);
-            d = Double.parseDouble(dString.replace(',', '.'));
             mAnswer.setText(d.toString());
         }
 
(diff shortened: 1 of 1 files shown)


ctsims:
Hm,

I'm not sure I think this is the right output.

Does this also affect things like "whether a 'decimal point' is a comma v. a period"? 

I think it's correct for us to treat

10,50
and
10.50

differently based on the user's locale, otherwise we could have serious data integrity issues in European locales.

Is the issue here that there's some assumption that if the locale of the input box is Arabic that if one types "1234" they get "٢١ ٣ ٤ " ? It does make sense for us to not have it automatically translate the unicode literals, I just wanna make sure it doesn't influence any other locale specific behavior.

shubham1g5:
@ctsims 
> Does this also affect things like "whether a 'decimal point' is a comma v. a period"?
Yes, though we are already replacing ',' with '.' in case of decimals [here](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/views/widgets/DecimalWidget.java#L70)

> Is the issue here that there's some assumption that if the locale of the input box is Arabic that if one types "1234" they get "٢١ ٣ ٤ " ?
Yes, that's what this PR is trying to address.

ctsims:
Ah, interesting. 

So it looks like as part of validating an input we allow someone to input "10,50" but we convert it to "10.50" before committing the answer. 

Would this change prevent them from inputting "10,50" at all, or just have the same behavior of updating it? If the latter, all seems well, if the former I think might warrant a bit more thought.

shubham1g5:
Hey it's definitely the latter i.e. we are not putting these restrictions on input. Interestingly we are only doing this formatting in the constructor for the `DecimalWidget` which means that these modifications will only happen if that question gets re-initialize which in turn will happen if the user returns to a question. This looks like this might cause data inconsistency for questions users return to vs which were only filled once by the user. For example if the language is set to Arabic, questions users return to will be formatted as Arabic rather than what user wrote (0-9,.). 

One more thing, we are setting a [key listener](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/views/widgets/DecimalWidget.java#L51) on the Decimal Input so that we only allow the numbers and `.`. So `,` will never be accepted as a valid input. We are also setting a [filter](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/views/widgets/DecimalWidget.java#L56) which only allows total of 15 digits. Given these restrictions I am not entirely sure why we are additionally applying the Number format formatting on decimal widget and it looks to me that it's not serving any additional purpose. I would be pro removing this formatting in entirety if you don't see any benefits of it as well.

ctsims:
Interesting. I think it's probably just a historic mismatch where there was one tool that was used initially, then another tool was added to address a more complex problem, and eventually the original tool no longer had any relevance.

If what you say is true, then I think that approach makes sense as the right one to follow.

shubham1g5:
@ctsims Thanks. I have updated the PR to reflect the same though one edge case that I can see is when a decimal question in a form aleady has some incorrect data. Since the restrictions we are applying are input filter based they stops user from inputting anything (using keypad) thats not  `{0-9,.}` but if that question have an incorrect value set already, it won't correct that input or throw an error. I am not sure though if such a situation can arise in the current setup and how big a deal it is.

ctsims:
I think I remembered what the number-format piece here is for

I believe that this prevents the system from turning doubles that can be 'reduced' into a smaller format into being compressed, ie:

.00000043 -> 43e-6

We should make sure that behavior isn't removed by this change

shubham1g5:
@ctsims That's a valid point. Currently if we input .00000043 it gets converted to 43e-6 if user returns to the question. Though that's the behaviour with number formatting present as well since we [again convert](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/views/widgets/DecimalWidget.java#L70) the formatted string to Double before setting the value to EditText. 

I have changed this PR now to retain number formatting with English locale so that the numbers are always english numbers - `0-9` and also have removed the re-conversion from formatted string to Double so that we don't show exponential numbers to users.

PR #2081, opened 2019-02-19
Handle Form resource Id changes during App Update

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/browse/ICDS-343

**Issue before this PR**

`upgrade` flag for `Resource.install` gets set based on the resource id match. If master table already has a resource with a given resource id, `upgrade` is set to true otherwise false.

For all update table resources who have a resource id match in master table and gets marked as `RESOURCE_STATUS_UPGRADE` during `install`, `Resource.upgrade` is called while applying the staged update. For a form resource `upgrade` is where we [update the filePath](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/android/resource/installers/XFormAndroidInstaller.java#L125) for an existing form in the FormDef storage to point to the newer form. 

This updation of path doesn't happen when a form changes the resource id  currently since both the conditions that are required for `upgrade` to get called doesn't meet - 
1. Such resources doesn't get marked as `RESOURCE_STATUS_UPGRADE` but as `RESOURCE_STATUS_INSTALLED` in `Resource.install` as due to resource id being different `upgrade` is set to false in [install](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/android/resource/installers/XFormAndroidInstaller.java#L113). 

2. No corresponding resource id match in master table which means `upgrade` [won't be called](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/commcare/resources/model/ResourceTable.java#L666) even if 1st condistion is met. 

To end the loop, we call `uninstall` for the master table resources who are no longer there in the update table. When a form changes the resource id, the old file path for the master form resource gets deleted [later](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/commcare/resources/model/ResourceTable.java#L723) during `uninstall` which results in the "Form Not found error" for such forms. 

**Solution**

In order to meet both conditions - 

1. Instead of using the upgrade flag to marke resources as `RESOURCE_STATUS_UPGRADE`, do it based on whether a form is already present with the given namespace.  (this is what this PR changes)

2. Call `upgrade` for all update resources marked as `RESOURCE_STATUS_UPGRADE` even if there is no resource id match in master for the update form resource (This is what the cross requested PR addresses)

cross-request: https://github.com/dimagi/commcare-core/pull/841

### DIFF
--- a/app/src/org/commcare/android/resource/installers/FileSystemUtils.java
+++ b/app/src/org/commcare/android/resource/installers/FileSystemUtils.java
@@ -31,6 +31,10 @@ public class FileSystemUtils {
         File newFile = new File(ReferenceManager.instance().DeriveReference(newLocation).getLocalURI());
         File oldFile = new File(ReferenceManager.instance().DeriveReference(oldLocation).getLocalURI());
 
+        if (oldFile.getAbsolutePath().contentEquals(newFile.getAbsolutePath())) {
+            return true;
+        }
+
         if (!oldFile.exists()) {
             //Nothing should be allowed to exist in the new location except for the incoming file
             //due to the staging rules. If there's a file there, it's this one.
--- a/app/src/org/commcare/android/resource/installers/XFormAndroidInstaller.java
+++ b/app/src/org/commcare/android/resource/installers/XFormAndroidInstaller.java
@@ -91,7 +91,8 @@ public class XFormAndroidInstaller extends FileSystemInstaller {
         }
 
         Vector<Integer> existingforms = FormDefRecord.getFormDefIdsByJrFormId(platform.getFormDefStorage(), formDef.getMainInstance().schema);
-        if (existingforms != null && existingforms.size() > 0) {
+        boolean formAlreadyExists = existingforms != null && existingforms.size() > 0;
+        if (formAlreadyExists) {
             //we already have one form. Hopefully this is during an upgrade...
             if (!upgrade) {
                 //Hm, error out?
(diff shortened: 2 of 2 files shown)


ctsims:
@shubham1g5 Hey. I'm really sorry to keep throwing out issues with these, but I'm not super comfortable with this approach. From my perspective the upgrade / install distinction is semantically important, and I don't think the resources should try to "infer" that they are an upgrade to another resource. 

I can't think of a distinct failure mode currently, but I think it's important for systems like these to behave predictably and consistently, and it could create huge confusing headaches down the line for someone who expects that something is only upgraded in-place if the resource Id's match.

For approach: 

When HQ screws up and loses track of which resources are which, I think the right behavior is for us to uninstall the old resource and replace it with the new one. To me the relevant issue is just the process of doing so wipes out the "incoming" file because we make an inappropriate assumption that the namespaces won't overlap. I don't think we need to treat the new files as "upgrades" to the old one, simply to allow for files with similar internals to be removed and installed on the same cycle.

It does occur to me that solving this on the phone side may not even work effectively for the team in this case, because if media files all get new ID's it's going to require hundreds of megs to get the update. I'll update the Jira ticket to see what the expectaiton is there

shubham1g5:
@ctsims  

> it could create huge confusing headaches down the line for someone who expects that something is only upgraded in-place if the resource Id's match

Completely agree that it's a confusing work around to solve the issue.

> To me the relevant issue is just the process of doing so wipes out the "incoming" file because we make an inappropriate assumption that the namespaces won't overlap

This is a great lead. Think I was over obsessed with upgrading the form that I didn't think of solving it this way. I am going to wait for the conversation on Jira before trying this out.

PR #2082, opened 2019-02-21
Restore search on getting back to the Case list

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/browse/MOB-73

Tries to restore the search after setting the adapter.

@ctsims It seems a bit urgent for REACH so getting it in 2.45 for next LTS release even though QA test run is complete. 

Product Note: Fixes a bug causing search results to not be retained on returning back to the case list.


shubham1g5:
@ctsims  Thanks. did the test and it's working 👍

PR #2085, opened 2019-02-27
Adds support for last_sync attribute for Indexed Fixtures

shubham1g5:
- Adds a column for root level attributes in IndexedFixtureIndex user DB table
- Adds `AndroidIndexedFixtureInstanceTreeElement` to support attributes lookup for root attibutes. 

cross-request: https://github.com/dimagi/commcare-core/pull/842

### DIFF
--- a/app/src/org/commcare/engine/cases/AndroidIndexedFixtureInstanceTreeElement.java
+++ b/app/src/org/commcare/engine/cases/AndroidIndexedFixtureInstanceTreeElement.java
@@ -1,0 +1,75 @@
+package org.commcare.engine.cases;
+
+import org.commcare.cases.instance.IndexedFixtureInstanceTreeElement;
+import org.commcare.cases.model.StorageIndexedTreeElementModel;
+import org.commcare.core.interfaces.UserSandbox;
+import org.commcare.utils.SerializationUtil;
+import org.javarosa.core.model.IndexedFixtureIdentifier;
+import org.javarosa.core.model.instance.AbstractTreeElement;
+import org.javarosa.core.model.instance.InstanceBase;
+import org.javarosa.core.model.instance.TreeElement;
+import org.javarosa.core.services.storage.IStorageUtilityIndexed;
+
+public class AndroidIndexedFixtureInstanceTreeElement extends IndexedFixtureInstanceTreeElement {
+
+    private TreeElement attributes;
+
+    public AndroidIndexedFixtureInstanceTreeElement(AbstractTreeElement instanceRoot,
+                                                    IStorageUtilityIndexed<StorageIndexedTreeElementModel> storage,
+                                                    IndexedFixtureIdentifier indexedFixtureIdentifier) {
+        super(instanceRoot, storage, indexedFixtureIdentifier);
+    }
+
+    public static IndexedFixtureInstanceTreeElement get(UserSandbox sandbox,
+                                                        String instanceName,
+                                                        InstanceBase instanceBase) {
+        IndexedFixtureIdentifier indexedFixtureIdentifier = sandbox.getIndexedFixtureIdentifier(instanceName);
+        if (indexedFixtureIdentifier == null) {
+            return null;
+        } else {
+            IStorageUtilityIndexed<StorageIndexedTreeElementModel> storage =
+                    sandbox.getIndexedFixtureStorage(instanceName);
+            return new AndroidIndexedFixtureInstanceTreeElement(instanceBase, storage, indexedFixtureIdentifier);
+        }
+    }
+
+    @Override
+    public int getAttributeCount() {
+        return getAttributes().getAttributeCount();
+    }
+
+    @Override
+    public String getAttributeNamespace(int index) {
+        return getAttributes().getAttributeNamespace(index);
+    }
+
+    @Override
+    public String getAttributeName(int index) {
+        return getAttributes().getAttributeName(index);
+    }
+
+    @Override
+    public String getAttributeValue(int index) {
+        return getAttributes().getAttributeValue(index);
+    }
+
+    @Override
+    public AbstractTreeElement getAttribute(String namespace, String name) {
+        TreeElement attr = getAttributes().getAttribute(namespace, name);
+        attr.setParent(this);
+        return attr;
+    }
+
+    @Override
+    public String getAttributeValue(String namespace, String name) {
+        return getAttributes().getAttributeValue(namespace, name);
+    }
+
+    private TreeElement getAttributes() {
+        if(attributes == null){
+            attributes = SerializationUtil.deserialize(attrHolder, TreeElement.class);
+        }
+        return attributes;
+    }
+
+}
--- a/app/src/org/commcare/engine/cases/AndroidIndexedFixtureInstanceTreeElement.java
+++ b/app/src/org/commcare/engine/cases/AndroidIndexedFixtureInstanceTreeElement.java
@@ -9,6 +9,7 @@ import org.javarosa.core.model.instance.AbstractTreeElement;
 import org.javarosa.core.model.instance.InstanceBase;
 import org.javarosa.core.model.instance.TreeElement;
 import org.javarosa.core.services.storage.IStorageUtilityIndexed;
+import org.javarosa.core.util.externalizable.DeserializationException;
 
 /**
  * The root element for an indexed fixture data instance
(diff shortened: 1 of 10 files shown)


ctsims:
Hi Shubham,

It seems a bit limiting for us to only be able to store a single attribute on the base element.

A lot of the dispatch downstream seems quite complex: https://github.com/dimagi/commcare-core/pull/842/files#diff-de4be0557361343fe019d6c9b25d8598R111

I'm wondering whether it would make more sense to instead serialize and store a single TreeElement which contains all of the core attributes. That element could be deserialized in whole, and then all of the *attribute* dispath methods in the element https://github.com/dimagi/commcare-core/pull/842/files#diff-de4be0557361343fe019d6c9b25d8598R95 could just be forwarded along, IE
```
    public String getAttributeName(int index) {
        return attributeHolderElement.getAttributeName(index)
    }
```

that would allow us to basically keep track of whatever other details need to be added in the future as well

shubham1g5:
@ctsims This is ready to be reviewed again.

app/src/org/commcare/models/database/user/UserDatabaseUpgrader.java
```diff
@@ -707,6 +716,21 @@ private boolean upgradeTwentyFourTwentyFive(SQLiteDatabase db) {
         }
     }
 
+    private boolean upgradeTwentyFiveTwentySix(SQLiteDatabase db) {
+        db.beginTransaction();
+        try {
+            db.execSQL(DbUtil.addColumnToTable(
+                    INDEXED_FIXTURE_PATHS_TABLE,
+                    INDEXED_FIXTURE_PATHS_COL_ATTRIBUTES,
+                    "TEXT"));
```
ctsims: Isn't this a BLOB?

app/src/org/commcare/models/database/IndexedFixturePathUtils.java
```diff
@@ -19,34 +24,65 @@
...
                 return null;
             } else {
                 c.moveToFirst();
-                return Pair.create(
-                        c.getString(c.getColumnIndexOrThrow(IndexedFixturePathsConstants.INDEXED_FIXTURE_PATHS_COL_BASE)),
-                        c.getString(c.getColumnIndexOrThrow(IndexedFixturePathsConstants.INDEXED_FIXTURE_PATHS_COL_CHILD)));
+                TreeElement attrsHolder = null;
+                byte[] attrsBlob = c.getBlob(c.getColumnIndexOrThrow(INDEXED_FIXTURE_PATHS_COL_ATTRIBUTES));
+                if (attrsBlob != null) {
+                    attrsHolder = SerializationUtil.deserialize(attrsBlob, TreeElement.class);
```
ctsims: I think we might need to have the prototype factory available here. I don't think it's relevant today, but it's best to avoid ever deserializing without it available.

ctsims: Additionally, I think we should have a concrete exception handling strategy for issues here, since deserialization exceptions are probably a very unexpected outcome.

It *might* be worth doing something like retrieving the byte[] here, and letting the TreeElement get deserialized at a later point (like in get() here: https://github.com/dimagi/commcare-core/pull/842/files#diff-de4be0557361343fe019d6c9b25d8598R38) when we are actively doing reads/deserializations from storage more broadly?

shubham1g5: the `deserialize` currently makes use of the prototype factory internally.

shubham1g5: > Additionally, I think we should have a concrete exception handling strategy for issues here, since deserialization exceptions are probably a very unexpected outcome.

Do you mean that we don't want the app to crash on such deserialization exceptions and instead show some sort of error message to the user ?

ctsims: Ah, got it. I thought this path used the "default" one, but it looks like it pulls it live from the session

ctsims: Mostly I mean that having the app crash here in a deserialization exception would be extremely confusing because it would be a symptom of an event that's out of the user's control and would be handled in a very strange part of the lifecycle.

Generally speaking I think it's very, very dangerous for "Utility" classes like this which are static and effectively stateless to throw errors that aren't strongly meaningful at the time. To the programmer, this method looks like it's just receiving a db hook and a string and doing some lookups. The caller wouldn't be expecting to need to be ready to handle mismatched serialized internal models based on the signature and expectations.

That **would** be their expectation later, however, when they are actually managing the resulting lookup table, so pushing the deserialization code there might be cleaner.

shubham1g5: @ctsims Thanks. I get your point now. Though I am struggling to find a better place for the deserialization to go since it requires access to the Prototype factory and hence needs to be in the `commcare-android`. The `get` function you linked above is part of `commcare-core`. The only other place I can forsee is 1 level up i.e. [getIndexedFixtureIdentifier](https://github.com/dimagi/commcare-android/blob/last_sync/app/src/org/commcare/models/database/AndroidSandbox.java#L67) but it again has the same semantics issue that doesn't convey to the caller to handle deserialization. 

Alternatively, we can decide to not catch `DeserializationException` in `SerializationUtil.deserialize` which would add a `throws DeserializationException` into function signature and hence will be much more evident to a caller function.

shubham1g5:

shubham1g5:

shubham1g5:

shubham1g5:

app/src/org/commcare/models/database/IndexedFixturePathUtils.java
ctsims: Hm, I see what you are saying. 

I can't think of a great workaround for now, and it is true that if we broke the TreeElement serialization, we'd need to be fixing the records from scratch anyway. 

Maybe the big point of confusion here is just that it doesn't feel right for the attribute holder to be in this record to begin with. Can we make a second version of this method that **doesn't** pull the attrsHolder so that if we needed to just get the base and child paths, that there's a path for that? Would be good for this version to also throw a typed error specific to this problem so the end user  (and us) can see very deliberately that it's an issue with the stored format for retrieving this.

I'm thinking about what pieces we'll need to have in place in a circumstance where we need  to migrate fixtures due to a TreeElement serialization format change, and right now this method would just be a blocker.

shubham1g5: @ctsims 

That makes a lot of sense though just to clarify are you suggesting that we create an unused function/ code path right now because we might need it in future ?

ctsims:
@shubham1g5 I think we are both aligned on potential ways forward here, and sorry for all of the back and forth. i'm fine with whatever approach you want to take based on our conversations

shubham1g5:

shubham1g5:

shubham1g5:
@ctsims This is ready to get reviewed again. Can you take another look.

app/src/org/commcare/utils/AndroidInstanceInitializer.java
```diff
@@ -56,6 +63,20 @@ protected AbstractTreeElement setupCaseData(ExternalDataInstance instance) {
         return casebase;
     }
 
+    @Override
+    protected AbstractTreeElement setupFixtureData(ExternalDataInstance instance) {
```
ctsims: @shubham1g5 We should make sure that we've also implemented anything we need to implement here for Formplayer. 

This may require, for example, require us to bump the version DB (which requires us to clear out Incomplete Forms). 

Unrelated to this PR structure, but useful for us to know whether we need to update the Formplayer DB Version # with the next commcare-core pull in that library before merging this one (let me know if this comment doesn't make sense).

app/src/org/commcare/engine/cases/AndroidIndexedFixtureInstanceTreeElement.java
```diff
@@ -0,0 +1,80 @@
...
+        attr.setParent(this);
+        return attr;
+    }
+
+    @Override
+    public String getAttributeValue(String namespace, String name) {
+        return getDeserializedAttributes().getAttributeValue(namespace, name);
+    }
+
+    private TreeElement getDeserializedAttributes() {
```
ctsims: This method needs to be synchronized in the same way as this one:
https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/commcare/cases/instance/StorageInstanceTreeElement.java#L103

to ensure that access to the attributes class is properly managed between threads.

We may want to consider a similar approach to that class where we have one "lazy load method" (there it's called loadElements()) that's called upfront before any methods, rather than specifically presuming that the lazy load is around the attributes.

app/src/org/commcare/engine/cases/AndroidIndexedFixtureInstanceTreeElement.java
```diff
@@ -0,0 +1,80 @@
...
+        if (attributes == null) {
+            attributes = SerializationUtil.deserialize(attrHolder, TreeElement.class);
```
ctsims: I think we may want to wrap serialization errors here, and specifically rethrow the error (still as a runtimeexception) with a message that says something like "Fixture Root Element Wrapper deserialization failed:" along with the real message, just so when this is seen and reported, we know exactly what is at error.

shubham1g5:
@ctsims I have not checked it on Formplayer yet though assume it will require some changes. 

1. Do we want to add this root level attribute lookup support on formplayer as well ? (This will require a similar implementation as`AndroidIndexedFixtureInstanceTreeElement` on formplayer level)

2. Is the concern here that the current core changes should not break existing functionality on Formplayer ? (This will require the DB upgrade for IndexedFixtureIndex table)

ctsims:
@shubham1g5 

1. I believe that we now should consider ourselves committed to making sure that core engine functionality such as this is implemented on both sides of the platform, since people primarily build and test their apps on Formplayer.
1. Yes. My concern is specifically that we no longer have a clear process for determining when we can push commcare-core changes to FormPlayer, so if I, say, wanted to pull https://github.com/dimagi/commcare-core/pull/851 into formplayer's commcare-core branch, we aren't tracking data model changes that will require a UserDB version upgrade.

On the #2 point there - We **don't** need to migrate formplayer's db. I am fine continuing to kill the UserDB entirely by updating the version, but I don't think we should know when we are making changes that result in commcare-core being unpullable in FormPlayer in case we needed to manage an urgent issue.

shubham1g5:
@ctsims Nice points. 
1. I think that also mean that we should push related changes for both Mobile and formplayer in the same mobile release cycle which essentially means parallel development for both. Though to support such process in long term, we will need to establish that everyone who works on mobile also is responsible for corresponding feature development on formplayer. 

2. You are right that today we need to be careful to examine any DB changes in the core PR before merging it in formplayer. I am not sure if it's an overkill to give some label to these kind of PRs to make it more evident.

app/src/org/commcare/engine/cases/AndroidIndexedFixtureInstanceTreeElement.java
shubham1g5: Made the method synchronized here though I am not sure of any particular benefits of doing deserialization ahead of time, specially since it requires us to be mindful of not accessing the `attributes` before we call the load on them. For ex. even though `loadElements()` is designed to be called once at the start we still end up calling it at couple other places.

shubham1g5:
@ctsims  bump to look at last set of changes here.

app/src/org/commcare/engine/cases/AndroidIndexedFixtureInstanceTreeElement.java
ctsims: Hi Shubham,

I think the lazy loading independent of attributes is just a stylistic thing, so fine for it to not be included.

-Clayton

ctsims:
good from my side!

PR #2086, opened 2019-02-27
Fixes CCZ Install during Recovery

shubham1g5:
Does 5 things- 
1.  Corrects the dispatch to offline install activity ([Jira](https://dimagi-dev.atlassian.net/browse/QA-203)). This was a silly mistake on my part and was affecting APP_REINSTALL_AND_UPDATE measure which allows user to choose between offline and online install. 

2. Handles the case when user returns to the Recovery Screen without selecting any file. 

3. Tries to handle the [NPE](https://www.fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/5c767784f8b88c29634aafe7/sessions/5C7684DA039E00010A59F1A16D7FEFE8_DNE_0_v2?). I have not been able to repro it so unsure why it's happening. 

4. Error handling: Catch any exceptions in async calls , Logs complete stack trace for an exception during install. 

5. In case we lost result of last recovery call and there is no task currently in progress, trigger the measures again. 

6. Fixes another silly mistake where update was not getting called for offline_reinstall_and_update_measure.


app/src/org/commcare/recovery/measures/ExecuteRecoveryMeasuresActivity.java
```diff
@@ -125,6 +125,8 @@ public void onActivityResult(int requestCode, int resultCode, Intent intent) {
             case OFFLINE_INSTALL_REQUEST:
                 if (resultCode == Activity.RESULT_OK) {
                     mPresenter.doOfflineAppInstall(intent.getStringExtra(InstallArchiveActivity.ARCHIVE_JR_REFERENCE));
+                } else {
+                    mPresenter.OnOfflineInstallCancelled();
```
shubham1g5: This is #2 in PR.

app/src/org/commcare/recovery/measures/ExecuteRecoveryMeasuresPresenter.java
```diff
@@ -256,9 +257,9 @@ private void onAsyncExecutionSuccess() {
 
     private void onAsyncExecutionFailure(String reason) {
         mLastExecutionStatus = STATUS_FAILED;
-        Logger.log(LogTypes.TYPE_MAINTENANCE, String.format(
-                "%s failed with %s for recovery measure %s", mCurrentMeasure.getType(), reason, mCurrentMeasure.getSequenceNumber()));
         if (mActivity != null) {
+            Logger.log(LogTypes.TYPE_MAINTENANCE, String.format(
```
shubham1g5: I am not sure how `mCurrentMeasure` is getting null though I am hoping it's because actvity is no longer alive and hence hoping moving the log under activity not null check might fix it.

ctsims: Just double chekcing, you mean how "mActivity is getting through null", right? not `mCurrentMeasure`? Or is `mCurrentMeasure` coming through null when mActivity is null?

shubham1g5: we are getting a crash report  because of `mCurrentMeasure` being null. Though I have not been able to repro it on my end.

app/src/org/commcare/recovery/measures/ExecuteRecoveryMeasuresPresenter.java
```diff
@@ -88,18 +89,22 @@
...
-        }
+        try {
+            if (cczSelectionEnabled) {
+                setCczSelectionVisibility(true);
+            }
 
-        if (mLastExecutionStatus == STATUS_FAILED) {
-            mActivity.enableRetry();
-            updateStatus(mLastDisplayStatus);
-        } else if (!(connectToUpdateTask() || mActivity.aTaskInProgress()) && mLastExecutionStatus != STATUS_WAITING) {
```
shubham1g5: Removed `mLastExecutionStatus != STATUS_WAITING` condition here since in case we don't have an exisiting task to connect to and mLastExecutionStatus is `STATUS_WAITING` means that we have lost result of the last async task. So the only resonable thing to do in this case is to trigger the measures again.

app/src/org/commcare/recovery/measures/ExecuteRecoveryMeasuresPresenter.java
```diff
@@ -138,53 +143,54 @@ private int executeMeasure() {
...
+
+        switch (mCurrentMeasure.getType()) {
+            case MEASURE_TYPE_APP_REINSTALL_AND_UPDATE:
+                if (AppUtils.notOnLatestAppVersion()) {
+                    showInstallMethodChooser();
+                    return STATUS_WAITING;
+                } else {
                     return STATUS_EXECUTED;
-            }
-        } catch (Exception e) {
```
shubham1g5: This is only change in L140 - L190. Moved this exception handling to the top level function `start`

app/src/org/commcare/recovery/measures/ExecuteRecoveryMeasuresPresenter.java
shubham1g5: I have now reverted this change and wrapped the whole thing inside a try catch block to not crash the app in case of any errors here but log the exception and show user a failure message.

shubham1g5:
@ctsims Made some more changes here specifically #4 and #5 in PR description. Can you take a re-look.

PR #2087, opened 2019-02-27
Install update as part of recovery measures

shubham1g5:
Before this PR, we were only staging the update as part of app update recovery measures and the actual update was getting applied when user was re-logging into the app. This PR changes that behaviour to apply/install the update as part of recovery itself without waiting for re-login.


app/src/org/commcare/activities/UpdateActivity.java
```diff
@@ -439,21 +439,27 @@ private static void reportAppUpdate() {
...
         } else {
+            final String upgradeFinishedText = Localization.get("updates.install.finished");
             Toast.makeText(this, upgradeFinishedText, Toast.LENGTH_LONG).show();
             setResult(RESULT_OK);
             this.finish();
         }
     }
 
+    // Helper function for common operations to do after an app update
+    public static void OnSuccessfulUpdate(boolean logout, boolean syncPostUpdate){
```
shubham1g5: Moved the common after update operations in here. I am not sure whether `syncPostUpdate` should be `true` when we do the update as part of a recovery measure.

ctsims: I don't think it should matter too much whether we do the sync or not fornoiw

PR #2088, opened 2019-03-01
Corrects eval context for menu relevance condition

shubham1g5:
https://dimagi-dev.atlassian.net/browse/MOB-76

we were using eval context for root menu to evaluate relevance condition for the child menu. This PR fixes that by using eval context for child menu.


shubham1g5:
@ctsims  This came from Saket through support. Emord might have been involved with that too.

PR #2089, opened 2019-03-01
Protects wipeTable call with a tableExists check

shubham1g5:
Jira: Jira: https://dimagi-dev.atlassian.net/browse/REACH-102

Refer core PR for context. 

cross-request: https://github.com/dimagi/commcare-core/pull/843

### DIFF
--- a/app/src/org/commcare/models/database/SqlStorage.java
+++ b/app/src/org/commcare/models/database/SqlStorage.java
@@ -506,9 +506,21 @@ public class SqlStorage<T extends Persistable> implements IStorageUtilityIndexed
         db.delete(table, null, null);
     }
 
-    @Override
-    public boolean exists() {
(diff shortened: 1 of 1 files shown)


shubham1g5:
@ctsims  PRing for 2.45 again since it seems important for REACH.

emord:
@shubham1g5 I don't think this is actually blocking REACH since there shouldn't be any users with 0 locations. I believe Nick just found this accidentally

shubham1g5:
@emord thats great to know. Nick mentioned  in the Jira issue that's it's high prio so I was assuming it must be satisfying some use case for REACH.  If that's not the case, I would prefer it to not PR for 2.45.

emord:
The fix I made for busting HQ's cache was high priority because it got around some issues around ICDS's consistent reversions. I don't htink this follow up is high enough priority to push into 2.45 if its at all risky/could be better tested

shubham1g5:
@emord Thanks for the context. I am going to de-prio for 2.45.

PR #2090, opened 2019-03-04
Don't update if we are already up to date

shubham1g5:


app/src/org/commcare/recovery/measures/ExecuteRecoveryMeasuresPresenter.java
```diff
@@ -523,7 +523,9 @@ public void handleTaskUpdate(Integer... updateVals) {
     @Override
     public void handleTaskCompletion(ResultAndError<AppInstallStatus> appInstallStatusResultAndError) {
         AppInstallStatus result = appInstallStatusResultAndError.data;
-        if (result == AppInstallStatus.UpdateStaged || result == AppInstallStatus.UpToDate) {
+        if (result == AppInstallStatus.UpToDate) {
```
shubham1g5: @ctsims  I am wondering if we should log the user out here (along with [other actions we do on app update](https://github.com/dimagi/commcare-android/blob/commcare_2.45/app/src/org/commcare/activities/UpdateActivity.java#L454)) as we do after an app update.

ctsims: That sounds totally right to me. The log out is especially important, but seems like we should just unify the call path to do exactly the same thing.

shubham1g5: Thanks. Added.

PR #2095, opened 2019-03-18
added rxBindings; added rx item click listener;

rristovic:
Added click filter on Entity list.
Jira ticket: https://dimagi-dev.atlassian.net/browse/MOB-77

### DIFF
--- a/app/src/org/commcare/activities/EntitySelectActivity.java
+++ b/app/src/org/commcare/activities/EntitySelectActivity.java
@@ -13,7 +13,6 @@ import android.view.Menu;
 import android.view.MenuItem;
 import android.view.View;
 import android.view.View.OnClickListener;
-import android.view.ViewGroup;
 import android.widget.AdapterView;
 import android.widget.AdapterView.OnItemClickListener;
 import android.widget.Button;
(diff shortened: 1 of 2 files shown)


damagatchi:

app/src/org/commcare/activities/EntitySelectActivity.java
```diff
@@ -263,7 +271,11 @@ private void setupUI(boolean isOrientationChange) {
             gridView.setVisibility(View.GONE);
             EntitySelectViewSetup.setupDivider(this, listView, shortSelect.usesEntityTileView());
         }
-        visibleView.setOnItemClickListener(this);
+        RxAdapterView.itemClickEvents(visibleView)
+                .subscribeOn(AndroidSchedulers.mainThread())
+                .throttleFirst(CLICK_DEBOUNCE_TIME, TimeUnit.MILLISECONDS)
+                .subscribe((Consumer<AdapterViewItemClickEvent>)clickEvent ->
```
shubham1g5: Hey, should we unsubscribe from these events when activity gets killed ?

rristovic: If I understand library author's explanation from this link https://github.com/JakeWharton/RxBinding/issues/282
we not need to unsubscribe, because in our case we do subscribe in activity where visibleView is part of Activity view. 
Correct me if I get this explanation wrong? 
Then again we could unsubscribe.
@shubham1g5 if you think we should unsubscribe, I will do make it.

shubham1g5: Hey sounds right! Thanks for looking into it.

PR #2097, opened 2019-03-20
Bug fix: Saving a form as Incomplete more than once doesn't work

shubham1g5:
Jira:https://dimagi-dev.atlassian.net/browse/MOB-86

Issue: When we rolled out [this change](https://github.com/dimagi/commcare-android/pull/2041/files) to allow a form to only be saved once we didn't consider that we currently allow incomplete forms to be saved more than once through the "Save" options menu. 

Proposed Change: Only mark forms as "saved" for complete forms so that a incomplete form's save don't get stuck at [the check here](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/activities/FormEntryActivity.java#L737).

Product Note: Fixes a bug due to which saving a form as Incomplete more than once doesn't work

### DIFF
--- a/app/src/org/commcare/tasks/SaveToDiskTask.java
+++ b/app/src/org/commcare/tasks/SaveToDiskTask.java
@@ -123,7 +123,10 @@ public class SaveToDiskTask extends
             return new ResultAndError<>(SaveStatus.SAVE_ERROR, cleanedMessage);
         }
 
-        FormEntryActivity.mFormController.markFormSaveProcessComplete(true);
+        if (mMarkCompleted) {
+            FormEntryActivity.mFormController.markCompleteFormAsSaved();
+        }
+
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/tasks/SaveToDiskTask.java
```diff
@@ -123,7 +123,10 @@ public SaveToDiskTask(int formRecordId, int formDefId, String formRecordPath, Bo
             return new ResultAndError<>(SaveStatus.SAVE_ERROR, cleanedMessage);
         }
 
-        FormEntryActivity.mFormController.markFormSaveProcessComplete(true);
+        if (mMarkCompleted) {
```
shubham1g5: core fix - checking if a form is complete before setting the flag.

ctsims:

PR #2099, opened 2019-03-25
added sorting for unsaved forms

rristovic:
Product note: When unsent filter selected on saved forms screen, forms are listed in submission order.

Jira task https://dimagi-dev.atlassian.net/browse/MOB-74

### DIFF
--- a/app/src/org/commcare/adapters/IncompleteFormListAdapter.java
+++ b/app/src/org/commcare/adapters/IncompleteFormListAdapter.java
@@ -168,15 +168,28 @@ public class IncompleteFormListAdapter extends BaseAdapter implements FormRecord
 
         // Sort FormRecords by modification time, most recent first.
         Collections.sort(records, (left, right) -> {
-            long leftModTime = left.lastModified().getTime();
-            long rightModTime = right.lastModified().getTime();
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/adapters/IncompleteFormListAdapter.java
```diff
@@ -168,15 +168,28 @@ public void resetRecords() {
...
+                if (leftModTime > rightModTime) {
+                    return -1;
+                } else if (leftModTime == rightModTime) {
+                    return 0;
+                } else {
+                    return 1;
+                }
             } else {
-                return 1;
+                int leftSubmissionOrderNum = left.getSubmissionOrderingNumber();
```
shubham1g5: We also do this [sorting](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/utils/StorageUtils.java#L95) when sending the forms to the server. It will be great if we can utilize the same code here.

rristovic: Used sorting from StorageUtils and also added to StorageUtils function for sorting by last modified time.

PR #2102, opened 2019-04-05
Updated app name in action bar.

rristovic:
Updated action bar to display correct application name depending on different build variant.
Task on Jira: https://dimagi-dev.atlassian.net/browse/MOB-82

Screen Shots
commcare

comcare2

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -36,7 +36,7 @@ configurations {
 dependencies {
 
     testImplementation 'junit:junit:4.12'
-    testImplementation ('org.robolectric:robolectric:3.2.2')
+    testImplementation('org.robolectric:robolectric:3.2.2')
     testImplementation 'org.robolectric:shadows-multidex:3.2.2'
     testImplementation 'org.robolectric:shadows-core:3.2.2'
     testImplementation 'org.json:json:20140107'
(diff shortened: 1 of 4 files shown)


app/build.gradle
```diff
@@ -83,7 +83,8 @@ dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
 
     // Markdown
-    implementation 'com.github.budsmile:bypass-android:1.0.5' //Needed for Ice Cream Sandwich (14&15)
+    implementation 'com.github.budsmile:bypass-android:1.0.5'
+    //Needed for Ice Cream Sandwich (14&15)
```
shubham1g5: we should revert formatting here. It can lead to someone believing the the comment is for `ru.noties:markwon:1.0.4` when it's actually for `com.github.budsmile:bypass-android:1.0.5`.

app/build.gradle
```diff
@@ -343,6 +348,9 @@ android {
 
             // set the app name
             resValue "string", "application_name", " CommCare LTS"
+            //setup action bar app name
+            buildConfigField "String", "APPLICATION_NAME", "\" CommCare LTS\""
```
shubham1g5: Is it possible to use the `application_name` defined above instead of creating a new field ?

app/build.gradle
rristovic: I missed this, thanks, it is reverted now.

app/build.gradle
rristovic: No it isn't because in ApplicationRecord.java class there is no context to get string.
But I did small code refactoring, I defined applicationName to use it for application_name, APPLICATION_NAME and  title_bar_name, this way is less chance to make mistake in some future changes.

rristovic: Unfortunately, I wasn't able to use BuildConfig fields in Manifest File. So the answer is no we cannot remove application_name.

shubham1g5: I see, I think for `ApplicationRecord` you should be able to use the application context (`CommCareApplication.instance()`) and hence able to use  `application_name` ?

rristovic: Yes it can use CommCareApplication.instance(), it is refactored now.

PR #2104, opened 2019-04-12
Fixes form submission in practice mode

shubham1g5:
Looks like practice mode form submission has been broken for a while (since retrofit migration). On submitting any data in practice mode we get 401 from server since we are no longer putting `params.put(AUTH_REQUEST_TYPE, AUTH_REQUEST_TYPE_NO_AUTH);` in the form submission request. 

I am not sure if it's worth a hotfix so just PRing to master. 

Product Note: Fixes form submission in practice mode.

### DIFF
--- a/app/src/org/commcare/network/CommcareRequestGenerator.java
+++ b/app/src/org/commcare/network/CommcareRequestGenerator.java
@@ -199,6 +199,7 @@ public class CommcareRequestGenerator implements CommcareRequestEndpoints {
 
         if (User.TYPE_DEMO.equals(userType)) {
             params.put(SUBMIT_MODE, SUBMIT_MODE_DEMO);
+            params.put(AUTH_REQUEST_TYPE, AUTH_REQUEST_TYPE_NO_AUTH);
         }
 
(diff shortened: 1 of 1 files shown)


ctsims:
@shubham1g5 My guess is that this line:
https://github.com/dimagi/commcare-android/pull/2104/files#diff-20bb24bd598ae6ae4f34b3bdb1cfbf2aR197

was what was actually triggering the noauth situation to submit before, btw. It might be worth seeing if that's supposed to be triggering up the chain.

shubham1g5:

shubham1g5:
@ctsims Earlier we were setting username and password to be null in this class in case of demo user which is why that line was sufficient. After retrofit migration, I mistakenly skipped that null setting for demo user which caused this bug. 

To fix it I thought it's better to not set username and password as null but instead be more explicit that we are setting `auth_request_type` to `noauth` for a demo user.

ctsims:
@shubham1g5 Perfect. Definitely agree that being more explicit is better, and glad that you made this change.

Thanks!

PR #2106, opened 2019-04-19
Improve Sync Request hashing performance

ctsims:
Significantly improves how fast we generate the state hash token for syncs (at least 100% faster with 54k cases, probably faster on slower devices) by using the metadata index coverage method rather than reading the case records raw.

Found this while investigating performance issues identified by Jessica on isth-production

cross-request: https://github.com/dimagi/commcare-core/pull/852

### DIFF
--- a/app/src/org/commcare/engine/cases/CaseUtils.java
+++ b/app/src/org/commcare/engine/cases/CaseUtils.java
@@ -25,10 +25,13 @@ import org.javarosa.core.services.Logger;
 import org.javarosa.core.services.storage.IStorageIterator;
 import org.javarosa.core.services.storage.IStorageUtilityIndexed;
 import org.javarosa.core.util.DAG;
+import org.javarosa.core.util.MD5;
 import org.javarosa.model.xform.XPathReference;
 
 import java.util.HashMap;
 import java.util.Vector;
+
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/engine/cases/CaseUtils.java
```diff
@@ -25,17 +25,51 @@
...
+            i.nextID();
+        }
+
+        //In the base case (with no cases), the case hash is empty
+        if (!casesExist) {
+            return "";
+        }
+
+        long timeEnd = System.currentTimeMillis();
+        Logger.log(LogTypes.TYPE_MAINTENANCE, String.format(
```
shubham1g5: Do we want to continue logging time here ? (Just flagging in case it was just meant for comparing the 2 methods)

ctsims: Good catch, thanks for bringing it up. I left it in because we also do this measurement in the Case Purge Process:
https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/engine/cases/CaseUtils.java#L113

and this actually runs less frequently than the purge does, but definitely good for us to keep track of.

PR #2109, opened 2019-04-23
Added check: if there is no question on the screen, save form and exit

rristovic:
Product note: "If there is no last question on the form, only hidden value, submit and exit the form will be triggered"

Jira issue: https://dimagi-dev.atlassian.net/browse/MOB-99

### DIFF
--- a/app/src/org/commcare/activities/FormEntryActivityUIController.java
+++ b/app/src/org/commcare/activities/FormEntryActivityUIController.java
@@ -418,7 +418,11 @@ public class FormEntryActivityUIController implements CommCareActivityUIControll
                             break group_skip;
                         case FormEntryController.EVENT_END_OF_FORM:
                             // auto-advance questions might advance past the last form quesion
-                            refreshCurrentView();
+                            if (questionsView != null) {
+                                refreshCurrentView();
+                            } else {
+                                activity.triggerUserFormComplete();
+                            }
                             break group_skip;
                         case FormEntryController.EVENT_PROMPT_NEW_REPEAT:
                             createRepeatDialog();
@@ -603,7 +607,7 @@ public class FormEntryActivityUIController implements CommCareActivityUIControll
             if (index.equals(q.getFormId())) {
 
                 if (q.getAnswer() instanceof InvalidData) {
-                    constraintText = ((InvalidData) q.getAnswer()).getErrorMessage();
+                    constraintText = ((InvalidData)q.getAnswer()).getErrorMessage();
                 }
 
                 q.notifyInvalid(constraintText, requestFocus);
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/activities/FormEntryActivityUIController.java
```diff
@@ -418,7 +418,11 @@ private void showNextView(boolean resuming) {
                             break group_skip;
                         case FormEntryController.EVENT_END_OF_FORM:
                             // auto-advance questions might advance past the last form quesion
-                            refreshCurrentView();
+                            if (questionsView != null) {
```
shubham1g5: Would be good to add a comment here something like -   `questionsView` null signifies that the form is empty and therefore we would want to auto-save the form.

rristovic: @shubham1g5 added, please check.

shubham1g5: Being nitpicky here though there is a typo in "special" and don't think the hidden value part is necessary since that's inherent in saving the form.

ctsims:
@rristovic 

Sorry I wasn't able to review earlier, but two points of feedback:

1) Where does the questionView = null state end up coming from? Given that it ends the form (which is a big, non-recoverable action), I'm a bit concerned about relying on an implicit state like this to trigger the move forward (in at least [one place](https://github.com/dimagi/commcare-android/pull/2109/files#diff-23f4ad0e03def61dfe0273e73b569da7R145) we reset it to null to get a clean state to rebuild, for example).

Can we add a specific validation check that actually [asks the form controller](https://github.com/dimagi/commcare-android/pull/2109/files#diff-23f4ad0e03def61dfe0273e73b569da7R171) specifically whether there are no interactable-questions in the form before triggering the action? (that should be a simple walk over events to identify that only Form Start and FormEnd exist)

2) Given that we've decided that this is now expected behavior, I think it would be worth adding a UI test for this if possible. I can't think of a less-expensive way to test the output and it seems likely that we'll break it at some point.

shubham1g5:
@ctsims Agree that it's better to have a explicit form empty check here. This fix just relies on the fact that if the `questionView` is null at the end of the form event that can only mean than the form is empty since otherwise the `questionView` would be set to a view. 

I guess we should revert it for 2.46 and then later make a more proper fix for it ?

ctsims:
@shubham1g5 I'm mixed on that, I think it's likely possible to reframe within the QA period if we think it's a straightforward check. I think the overall structure of this fix is right, just that it needs a new method on the form controller to confirm that there are no interact-able questions rather than just inferring it.

shubham1g5:
Makes sense. 

@rristovic  Would it be possible for you to prioritize this one. Would be good to have this in for 2.46.

PR #2112, opened 2019-04-26
Feature/update target version 28

rristovic:
Updated target API version to 28 (Android 9)
Jira ticket: https://dimagi-dev.atlassian.net/browse/MOB-91

Release Note: Adds support for Android 9

### DIFF
--- a/app/src/org/commcare/services/WiFiDirectBroadcastReceiver.java
+++ b/app/src/org/commcare/services/WiFiDirectBroadcastReceiver.java
@@ -38,15 +38,17 @@ public class WiFiDirectBroadcastReceiver extends BroadcastReceiver {
         super();
         this.manager = manager;
         this.activity = activity;
-        if (Build.VERSION.SDK_INT > Build.VERSION_CODES.O_MR1) {
-            connectivityManager = (ConnectivityManager)activity.getContext().getSystemService(Context.CONNECTIVITY_SERVICE);
-        }
     }
 
     @TargetApi(Build.VERSION_CODES.HONEYCOMB)
     @Override
     public void onReceive(Context context, Intent intent) {
         Log.d(TAG, "in on receive ");
+        
+        if (Build.VERSION.SDK_INT > Build.VERSION_CODES.O_MR1) {
+            connectivityManager = (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
+        }
+        
         String action = intent.getAction();
 
         Logger.log(LogTypes.TYPE_WIFI_DIRECT, "onReceive of WifiDirectBroadCastReceiver with action: " + action);
@@ -72,9 +74,7 @@ public class WiFiDirectBroadcastReceiver extends BroadcastReceiver {
             // asynchronous call and the calling activity is notified with a
             // callback on PeerListListener.onPeersAvailable()
             if (manager != null) {
-
                 activity.onPeersChanged();
-
             }
             Log.d(TAG, "P2P peers changed2");
         } else if (WifiP2pManager.WIFI_P2P_CONNECTION_CHANGED_ACTION.equals(action)) {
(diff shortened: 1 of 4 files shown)


app/src/org/commcare/services/WiFiDirectBroadcastReceiver.java
```diff
@@ -34,6 +38,9 @@ public WiFiDirectBroadcastReceiver(WifiP2pManager manager,
         super();
         this.manager = manager;
         this.activity = activity;
+        if (Build.VERSION.SDK_INT > Build.VERSION_CODES.O_MR1) {
+            connectivityManager = (ConnectivityManager)activity.getContext().getSystemService(Context.CONNECTIVITY_SERVICE);
```
shubham1g5: Is there a specific reason we are initializing `connectivityManager` in constructor instead of when we need to use it later in `onReceive`.

app/src/org/commcare/utils/Permissions.java
```diff
@@ -12,6 +12,12 @@
 import org.commcare.views.dialogs.DialogCreationHelpers;
 import org.javarosa.core.services.locale.Localization;
 
+import java.util.ArrayList;
```
shubham1g5: Thanks for making the code cleaner here, though just wanna confirm that no change in this class is relevant for Android 9 compatibility ?

rristovic:
> left some Minor comments, though wondering have you been able to test this on Android 9 either on phone or emulator ?

I've tested it only on emulator.

app/src/org/commcare/utils/Permissions.java
rristovic: Correct, just a code cleanup

app/src/org/commcare/services/WiFiDirectBroadcastReceiver.java
rristovic: no reason, i'll move it

app/src/org/commcare/services/WiFiDirectBroadcastReceiver.java
```diff
@@ -24,6 +27,7 @@
     private static final String TAG = WiFiDirectBroadcastReceiver.class.getSimpleName();
     private final WifiP2pManager manager;
     private final WiFiDirectManagementFragment activity;
+    private ConnectivityManager connectivityManager;
```
shubham1g5: Can be a local variable now.

shubham1g5: @rristovic bump on this in case you missed this.

app/src/org/commcare/services/WiFiDirectBroadcastReceiver.java
```diff
@@ -76,19 +83,25 @@ public void onReceive(Context context, Intent intent) {
...
 
-            NetworkInfo networkInfo = intent
-                    .getParcelableExtra(WifiP2pManager.EXTRA_NETWORK_INFO);
+            NetworkInfo networkInfo;
+            if (Build.VERSION.SDK_INT < Build.VERSION_CODES.P) {
+                networkInfo = intent
+                        .getParcelableExtra(WifiP2pManager.EXTRA_NETWORK_INFO);
+            } else {
+                //EXTRA_NETWORK_INFO is removed from Android API 28
+                networkInfo = connectivityManager.getActiveNetworkInfo();
```
shubham1g5: @rristovic  I looked a bit into this and it seems like `EXTRA_NETWORK_INFO` got removed for the broadcast `NETWORK_STATE_CHANGED_ACTION` and not for `WifiP2pManager.WIFI_P2P_CONNECTION_CHANGED_ACTION`. Also here `connectivityManager` will return whether the phone is connected to internet while the `networkInfo` earlier was supposed to convey the info regarding  wifi P2P connection for ex. if we have a P2P connection between 2 devices.

shubham1g5:
Closing this since I thinh changes relates to WifiP2P are unnecessary.

PR #2120, opened 2019-06-07
Clear focus from views to avoid form jumps on adding/removing views

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/browse/MOB-107
Also fixes: https://dimagi-dev.atlassian.net/browse/MOB-10 

Issue:
When after editing a editText, user shifts to a non editText field (like RadioButton), the last changed editText don't clear foucs by itself. This results into abrupt form jumps to this last focused EditText when soft keyboard gets hidden or if we add/remove any views from the Question List.

Solution:
Ensure that we remove focus from all views but the last changed one while updating Form Relevancies.

### DIFF
--- a/app/src/org/commcare/activities/FormEntryActivityUIController.java
+++ b/app/src/org/commcare/activities/FormEntryActivityUIController.java
@@ -753,7 +753,7 @@ public class FormEntryActivityUIController implements CommCareActivityUIControll
                 shouldRemoveFromView.add(i);
             }
 
-            // Finally if we have not changes this widget last, make sure that
+            // Finally if this is not the last widget user changed, make sure that
             // this widget is not in focus since it results in some absurd form
             // jumping issues.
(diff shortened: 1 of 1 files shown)


ctsims:
Unfortunately this is a very tricky one to review at the code level. The code itself looks fine

One big concern - clearing focus on some of the widgets (I'm thinking specifically of the date widgets) can trigger internal behaviors like revalidating that dates are valid. Might be worth making sure that his doesn't result in those widgets stealing back focus. 

Maybe it would be a good idea to work with QA to come up with a big page of confusing / potentially conflicting widgets and see if it would be a good test bed?

shubham1g5:
Noting down that this PR breaks the `ComboboxWidget` as it defines [some operations](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/views/widgets/ComboboxWidget.java#L83) based on Focus changes.

PR #2122, opened 2019-07-08
Moves Indexed Fixture attribute lookup methods to core

shubham1g5:
cross-request: https://github.com/dimagi/commcare-core/pull/858


shubham1g5:
@ctsims  that's right. Since QA is already complete for 2.47, I didn't wanna make more changes to 2.47. This PR is just some no-op refactoring to support formplayer changes and doesn't really have a functional impact. So I don't think it's required for 2.47.

PR #2124, opened 2019-07-09
Don't retry form submissions when rate limited

ctsims:
Currently when receiving a 503 from the server, the phone immediately tries again a few times. This updates the send task to short circuit back instead to cease further attempts to submit the forms automatically.

This is high priority as our rate limiting is currently getting hit with even higher submission counts after each form due to the automated retry.

### DIFF
--- a/app/src/org/commcare/tasks/ProcessAndSendTask.java
+++ b/app/src/org/commcare/tasks/ProcessAndSendTask.java
@@ -326,6 +326,9 @@ public abstract class ProcessAndSendTask<R> extends CommCareTask<FormRecord, Lon
                                 // A processing failure indicates that there there is no point in
                                 // trying that submission again immediately
                                 break;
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/tasks/ProcessAndSendTask.java
```diff
@@ -326,6 +326,9 @@ private void sendForms(FormRecord[] records) throws TaskCancelledException {
                                 // A processing failure indicates that there there is no point in
                                 // trying that submission again immediately
                                 break;
+                            } else if (results[i] == FormUploadResult.RATE_LIMITED) {
```
shubham1g5: @ctsims Where are we setting the result to `RATE_LIMITED` ?

ctsims: whoops, missed that delta in the commit, thanks. Added the callout here

PR #2126, opened 2019-07-24
Fixes for ProtoType Audio Widget

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/browse/MOB-122

The Prototyped widget was using `questionIndexText` for creating the filepath. This causes the widget to show up an already recorded file for all forms for that particular question. This PR fixes the issue by incorporating a more unique file identifier using datestamp.

### DIFF
--- a/app/src/org/commcare/views/widgets/RecordingFragment.java
+++ b/app/src/org/commcare/views/widgets/RecordingFragment.java
@@ -38,6 +38,8 @@ import java.io.IOException;
  */
 public class RecordingFragment extends android.support.v4.app.DialogFragment {
 
+    public static final String AUDIO_FILE_PATH_ARG_KEY = "audio_file_path";
+
     private String fileName;
     private static final String FILE_BASE =
             "/Android/data/org.commcare.dalvik/temp/Custom_Recording";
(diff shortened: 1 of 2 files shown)


shubham1g5:
@ctsims  PR'ed for master but planning to do a hotfix (2.47.2) once merged.

shubham1g5:

app/src/org/commcare/views/widgets/RecordingFragment.java
```diff
@@ -74,6 +84,10 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
         return layout;
     }
 
+    private void initAudioFile() {
+        fileName = Environment.getExternalStorageDirectory().getAbsolutePath() + FILE_BASE + listener.getFileUniqueIdentifier() + FILE_EXT;
```
ctsims: If the generic landing point here is going to be "CommCare's storage" I think we should point it to the same 'temp' directory where we unzip CCZ files, since that directory is regularly purged.

shubham1g5: @ctsims  Agree, the `FILE_BASE` here is actually a path with value `/Android/data/org.commcare.dalvik/temp/Custom_Recording` that takes care of this.

PR #2130, opened 2019-08-08
Check for task cancellation before making a form upload attempt

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/browse/MOB-118

On a poor network connection, there can be a gap of 2 mins when the request timeout and another attempt is made. So it only seems fair that we check for task cancellation in between the upload attempts.


shubham1g5:
Closing as including this as an hotfix - https://github.com/dimagi/commcare-android/pull/2133

PR #2132, opened 2019-08-09
[HOTFIX][2.47.3] Makes Auto Sync based on Last Upload Sync Time 

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/browse/MOB-48

Changes auto sync trigger calculation to use last upload sync time instead of last restore time.


app/src/org/commcare/preferences/HiddenPreferences.java
```diff
@@ -296,4 +297,18 @@ public static String getLastKnownCczLocation() {
         return PreferenceManager.getDefaultSharedPreferences(CommCareApplication.instance())
                 .getString(LAST_KNOWN_CCZ_LOCATION, null);
     }
+
+    public static void updateLastUploadAttemptTime() {
```
ctsims: Nitpick: everywhere else this is "Upload and Sync attempt," this one only mentions upload

app/src/org/commcare/preferences/HiddenPreferences.java
```diff
@@ -28,7 +29,7 @@
     public final static String POST_UPDATE_SYNC_NEEDED = "post-update-sync-needed";
     public final static String AUTO_UPDATE_IN_PROGRESS = "cc-trying-to-auto-update";
     public final static String LAST_UPDATE_ATTEMPT = "cc-last_up";
-    public final static String LAST_SYNC_ATTEMPT = "last-ota-restore";
+    public final static String LAST_UPLOAD_SYNC_ATTEMPT = "last-upload-sync";
```
ctsims: Small point: I'm not sure it's worth it to change the *value* of this key (compared to just the static variable name). It makes things more confusing, but changing the value will wipe any existing values for existing users.

In this case that's probably not a big deal, because anyone using this flag has autosync set up at max once a week, but worth considering user impact from the change.

app/src/org/commcare/activities/FormAndDataSyncer.java
```diff
@@ -46,7 +47,7 @@ public boolean checkAndStartUnsentFormsTask(SyncCapableCommCareActivity activity
                                                 boolean userTriggered) {
         SqlStorage<FormRecord> storage = CommCareApplication.instance().getUserStorage(FormRecord.class);
         FormRecord[] records = StorageUtils.getUnsentRecordsForCurrentApp(storage);
-
+        HiddenPreferences.updateLastUploadAttemptTime();
```
ctsims: I'm not sure we want this invocation:
Does this (or the one below) get triggered with **automatic** sends (the ones that happen in the background), or only on **blocking** sends (the ones that happen when the user presses 'sync', and/or the app triggers autosync)?

We don't want any attempt to send forms to invalidate the forced sync attempt, since the app tries to send forms every time the user completes one. We only want users to get 'credit' for a sync attempt with form submissions if they saw the blocking Modal Dialog come up, rather than having those submissions run in the background.

app/src/org/commcare/preferences/HiddenPreferences.java
shubham1g5: Hey agree with the concept in general though in this case we are also making a change where we are storing this info per user per app rather than just per app. And therefore the actual value is stored in a key "userid + "_" +  preferenceKey" . We can write a logic where it checks for the old preference key in case it doesn't find the new one but I am not sure if it's worth the lines of code that are going to stay in the codebase forever.

app/src/org/commcare/activities/FormAndDataSyncer.java
shubham1g5: This is the common path for both cases you described though this function gets passed an argument `syncAfterwards` which is always false for non-blocking sync. I am now using that to distinguish between the two cases

app/src/org/commcare/preferences/HiddenPreferences.java
ctsims: Aha, got it. I noticed the per-user part but failed to realize it was new.

Agree with you that it's not worth it to muddy the codebase based on that, and that a clean start is warranted

app/src/org/commcare/activities/FormAndDataSyncer.java
```diff
@@ -60,7 +66,7 @@ public void processAndSendForms(final SyncCapableCommCareActivity activity,
                                     FormRecord[] records,
                                     final boolean syncAfterwards,
                                     final boolean userTriggered) {
-
+        HiddenPreferences.updateLastUploadSyncAttemptTime();
```
ctsims: Just curious - Why isn't this also checking for the syncAfterwards flag? odd that it's passed in if it's always true.

If line 51 only ever results in a call to this method, i'd do the wrapping here, and ignore the edge case above, since it only occurs when there are no unsent forms and in that case a sync attempt would occur anyway if things were user triggered, which would naturally set the flag correctly

shubham1g5: @ctsims Hey, it's true that the only call to this method is from `checkAndStartUnsentFormsTask` . I have made this method `private` now to reflect the same and moved the `updateLastUploadSyncAttemptTime` call to the calling function `checkAndStartUnsentFormsTask` since we are no longer updating this time now from the restore sync and therefore we need to account for the edge case when there are no unsent forms as well.

PR #2134, opened 2019-08-16
Makes No. of Unsent Forms always visible

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/browse/ICDS-795

* Makes Unsent Forms Always Visible
* Adds a custom property "cc-show_unsent_forms_when_zero" to control whether to show unsent forms indicator when there are no unsent forms. 

Product Note: Makes No. of unsent forms always visible on the sync button.


shubham1g5:
@ctsims Modified this to show unsent forms when 0 based on a custom property. Can you take a look again.

PR #2137, opened 2019-09-03
Updates to Robolectric 4.3,  Gradle 5 , AGP 3.4

shubham1g5:
Robolectric Changes - 

* Turn off Strict mode death penalty in `CommCareTestApplication` and change all tests to build with `CommCareTestApplication` instead of `CommCareApplication`.  Otherwise various tests gets killed with message `StrictMode VmPolicy violation with POLICY_DEATH; shutting down.` because of leaked cursor/sqlite DB objects

* Run Tests  testing functions iincluding a call to NsdManager on sdk 18. There is a bug  on sdk 21 with robolectric that causes a NPE when accessing NsdManager functions

* API Changes for 4.3

cross-request: https://github.com/dimagi/commcare-core/pull/868

### DIFF
--- a/app/unit-tests/src/org/commcare/CommCareTestApplication.java
+++ b/app/unit-tests/src/org/commcare/CommCareTestApplication.java
@@ -129,9 +129,14 @@ public class CommCareTestApplication extends CommCareApplication implements Test
      */
     private static void initFactoryClassList() {
         if (factoryClassNames.isEmpty()) {
-            String baseODK = BuildConfig.BUILD_DIR + "/intermediates/javac/commcareDebug/compileCommcareDebugJavaWithJavac/classes/";
+            String[] baseODK = new String[]{BuildConfig.BUILD_DIR + "/intermediates/javac/commcareDebug/compileCommcareDebugJavaWithJavac/classes/"
+                        , BuildConfig.BUILD_DIR + "/intermediates/javac/commcareDebug/classes/"};
             String baseCC = BuildConfig.PROJECT_DIR + "/../../commcare-core/build/classes/java/main/";
-            addExternalizableClassesFromDir(baseODK.replace("/", File.separator), factoryClassNames);
+
+
+            for(String variant : baseODK) {
+                addExternalizableClassesFromDir(variant.replace("/", File.separator), factoryClassNames);
+            }
             addExternalizableClassesFromDir(baseCC.replace("/", File.separator), factoryClassNames);
         }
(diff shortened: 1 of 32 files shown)


app/unit-tests/src/org/commcare/CommCareTestApplication.java
```diff
@@ -125,7 +129,7 @@ public PrototypeFactory getPrototypeFactory(Context c) {
      */
     private static void initFactoryClassList() {
         if (factoryClassNames.isEmpty()) {
-            String baseODK = BuildConfig.BUILD_DIR + "/intermediates/classes/commcare/debug/";
+            String baseODK = BuildConfig.BUILD_DIR + "/intermediates/javac/commcareDebug/compileCommcareDebugJavaWithJavac/classes/";
```
ctsims: Is it possible to investigate whether there are better ways for us to estimate where these files come from? I'm totally fine if the answer is 'no' or 'it's a lot of work to even check,' but it worries me to keep relying on this hard coded of a setup

shubham1g5: I looked for it but could not find anything straightforward. Some people seems to define multiple paths in such cases hoping only one of them will exist at a time. Though I don't think it changes for a given version of AGP and therefore don't see value in relying on multiple paths either.

ctsims: Classic Java / Gradle - "Hyper generic until you need to hardcode 50 paths" 

this is awesome work, though, thanks for going through the deep slog of getting things compatible with new versions! 

Let me know when you think it's good-to-go by the way, and I'll spin up a test on my machine. Trying it out on a windows environment has illustrated some overfitting previously

shubham1g5: Hey yeah , It would be great if you can test it on your end. I have been able to run successful clean builds on Jenkins and on my own machine. So it's very much good to go from my side. I think you will need to update your AS thoughl. I am using AS 3.4 on my end though seems like AS 3.5 was just released.

ctsims: Sounds like as good a time to jump in as any. I'll update AS today and try this branch

shubham1g5:
@ctsims  Thanks for the changes. Even on AS 3.5, it builds file on the earlier path for me so we will need to keep both file Paths.

PR #2138, opened 2019-09-05
Android X Migration

shubham1g5:
All of these are just reference changes from `android.suuport` to corresponding `androidx` reference. 

The v1.0 Android X libraries are binary equivalent to android support 28.0.0. So we should not see any bugs because of this.


shubham1g5:

shubham1g5:

ctsims:
Recommendations around reviewing? If this was automated and non-semantic i'm fairly comfortable letting the tests do their job.

gradle.properties
```diff
@@ -17,4 +17,10 @@ org.gradle.parallel=true
...
 org.gradle.configureondemand=false
 
-android.enableUnitTestBinaryResources=true
\ No newline at end of file
+android.enableUnitTestBinaryResources=true
+
+# use the appropriate AndroidX library instead of a Support Library.
+android.useAndroidX=true
+
+# automatically migrates existing third-party libraries to use AndroidX by rewriting their binaries
```
ctsims: is this impact only around android support libraries or does it go further than that?

shubham1g5: Android X replaces only Support libs and it's API's. So the impact is limited to that.

shubham1g5:
@ctsims All of these are result of running Android X migration too in AS (Refactor -> Migrat to AndroidX) . So yeah I think build + test is good enough reviewing. I have gone through all the changes made by the tool and can cofirm they are all just references changes to androidX apart from `gradle.properties` and `build.gradle` ones.

shubham1g5:
@ctsims  Good to merge ?

PR #2141, opened 2019-09-24
On demand device logs

shubham1g5:
Spec: https://docs.google.com/document/d/1lkgq-w_olxPPaKBYb8_pbIHCOpSai3LVwoxcbvgzZbA/edit#heading=h.2bmw2e5u3lpt

* Implements a way to trigger device logs from HQ. 
* Implements a background task that runs weekly to delete any logs older than 6 months (triggered on login)

Product Note: Implements a way to trigger device logs submission for a user from HQ.  [Wiki here](https://confluence.dimagi.com/display/ICDS/On+Demand+Device+Logs).


app/src/org/commcare/activities/RecoveryActivity.java
```diff
@@ -56,7 +56,7 @@ public void onCreateSessionSafe(Bundle savedInstanceState) {
         super.onCreateSessionSafe(savedInstanceState);
         if (savedInstanceState == null) {
             // launching activity, not just changing orientation
-            CommCareUtil.triggerLogSubmission(this);
+            CommCareUtil.triggerLogSubmission(this, false);
```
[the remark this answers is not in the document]
shubham1g5: @ctsims  so this PR currently only using `force_logs` in log submission request when it's getting triggered from HQ. I agree that it's a good idea to submit logs in some of these cases specifically  1. Recovery 2. Manual Log submission by going to the menu. Though I wonder whether we will want to use the same `force_logs` flag for all of these cases.  HQ does further processing on getting a log submission request with force_logs set like looking for a DeviceLogRequest and deleting it. So It might be useful for us to differentiate  triggered log submission from all other log requests that we still need to go through even when log submission is diasabled.

app/src/org/commcare/network/CommcareRequestGenerator.java
```diff
@@ -190,24 +188,22 @@ private static String getDigest() {
...
     @Override
-    public Response<ResponseBody> postMultipart(String url, List<MultipartBody.Part> parts) throws IOException {
-
-        HashMap<String, String> params = new HashMap<>();
+    public Response<ResponseBody> postMultipart(String url, List<MultipartBody.Part> parts, HashMap<String, String> queryParams) throws IOException {
 
         //If we're going to try to post with no credentials, we need to be explicit about the fact that we're not ready
         if (username == null || password == null) {
-            params.put(AUTH_REQUEST_TYPE, AUTH_REQUEST_TYPE_NO_AUTH);
+            queryParams.put(AUTH_REQUEST_TYPE, AUTH_REQUEST_TYPE_NO_AUTH);
```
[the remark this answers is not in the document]
shubham1g5: Great Point. Thanks!

app/src/org/commcare/tasks/DeleteLogs.kt
```diff
@@ -0,0 +1,43 @@
...
+import androidx.work.Worker
+import androidx.work.WorkerParameters
+import org.commcare.CommCareApplication
+import org.commcare.CommCareApplication.MILLISECONDS_IN_A_DAY
+import org.commcare.android.javarosa.AndroidLogEntry
+import org.commcare.android.logging.ForceCloseLogEntry
+import org.commcare.logging.XPathErrorEntry
+import org.commcare.models.database.SqlStorage
+import org.commcare.preferences.HiddenPreferences
+import java.util.*
```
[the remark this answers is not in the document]
shubham1g5: Make sense. not sure why AS include it in entirety.

shubham1g5:
@ctsims  Great catch. I align with all the proposed changes and will revert soon with them.

app/src/org/commcare/activities/RecoveryActivity.java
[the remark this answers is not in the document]
ctsims: Got it, it definitely makes sense, and I missed that the force logs boolean is used both internally and sets a literal flag named `force_logs`, which makes sense.

If the boolean has the resulting behavior of triggering the HQ header as well as anything else, this seems totally consistent.

ctsims:
@shubham1g5 Awesome, thanks! I'm really, really excited for this feature and the approach you ended up taking. I think it'll be a huge asset to the project to have this kind of control for on-demand behavior in the app, and I hope it's something we can build on.

shubham1g5:
@ctsims  The PR is ready to review again.  I have shortened the 6 months period to 3 months since generally that's the threshold beyond which I have not required logs anytime. 

I also have another related Question - In the [doReportMaintenance](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/CommCareApplication.java#L757), we have another flag `force` that determines whether to serialize the logs or not. I am curious why do we need this flag and why don't we let serialization happen everytime ?

ctsims:
Hi @shubham1g5 

Reviewing now. 

To your question: The 'force' flag was previously used in the circumstances where there was an[error and we wanted to try to get logs off device](https://github.com/dimagi/commcare-android/blob/fd63a5b2471a1e33ac6bf1ce77ac82daf558b08a/app/src/org/commcare/activities/ReportProblemActivity.java#L39) or when we had [just completed a form submission and knew internet was good](https://github.com/dimagi/commcare-android/blob/e4532af658609d432c154c90422fbbeaa08dc719/app/src/org/commcare/tasks/ProcessAndSendTask.java#L164)

I do think that the code to enforce only doing the log submissions once a day on login is almost certainly an unnecessary consideration, since we rarely have more than one user login per day and in theory log submission is a non-blocking activity that happens in the BG. I think we started fairly conservative (mimicking logging behavior from j2me), and then later got more aggressive with the "After form submission try logs too" approach, which is why it's fairly uneven.

app/src/org/commcare/tasks/DeleteLogs.kt
```diff
@@ -0,0 +1,59 @@
...
+        purge(CommCareApplication.instance().getUserStorage(XPathErrorEntry.STORAGE_KEY, XPathErrorEntry::class.java), threeMonthsAgo)
+        purge(CommCareApplication.instance().getGlobalStorage(ForceCloseLogEntry.STORAGE_KEY, ForceCloseLogEntry::class.java), threeMonthsAgo)
+        purge(CommCareApplication.instance().getUserStorage(XPathErrorEntry.STORAGE_KEY, XPathErrorEntry::class.java), threeMonthsAgo)
+
+        // Clear logs older than 3 month which have been serialized to disk
+        val deviceReportRecordStorage = CommCareApplication.instance().getUserStorage(DeviceReportRecord::class.java)
+        val reportsToDelete = Vector<Int>()
+        for (deviceReportRecord in deviceReportRecordStorage) {
+            val reportFile = File(deviceReportRecord.filePath)
+            if (reportFile.lastModified() < threeMonthsAgo.time) {
```
ctsims: In the event that this code completes partially (IE: we make it 1/2 through the Storage loop, but never make it to line 43 below), there will be old historical DeviceReportRecords with references to Files that no longer exist. 

This line would need to include some provision for the file not existing to protect against that, but more generally there may be other uses of the DeviceReportRecord which will also need to be updated in the same way. 

I'm not sure how the records are cleared out in other contexts, since presumably they have the same transactional atomicity.

app/src/org/commcare/tasks/LogSubmissionTask.java
```diff
@@ -104,14 +108,16 @@ public static String getSubmissionUrl(SharedPreferences appPreferences) {
 
     @Override
     protected LogSubmitOutcomes doInBackground(Void... params) {
-        if (HiddenPreferences.isLoggingEnabled()) {
```
ctsims: Question: I think the reason we used to skip this class entirely if logging was disabled is that otherwise we create a DeviceLogReportRecord and a File that are empty (since the db had no logs to serialize) which we would never clean out since we used to only remove those records on log submission.

Is that outcome now not possible due to other structural changes? Otherwise I think we'd still need to skip serialization if Logging is totally disabled.

app/src/org/commcare/tasks/DeleteLogs.kt
shubham1g5: @ctsims  The `lastModified` returns 0 in case the file doesn't exist and the `delete` method doesn't throw an exception if file doesn't exist. So I think it's safe to leave it as it is. Verified the same at other places as well.

shubham1g5:
@ctsims  made the changes, Do you think it makes sense to remove that `force` flag and always let the log submission go through whenever `doReportMaintenance` is called ?

shubham1g5:
@ctsims Made those changes regarding doReportMaintenance and protection for file deletion. Can you do one more final check. Thanks!

app/src/org/commcare/tasks/DeleteLogs.kt
```diff
@@ -0,0 +1,59 @@
...
+        purge(CommCareApplication.instance().getUserStorage(XPathErrorEntry.STORAGE_KEY, XPathErrorEntry::class.java), threeMonthsAgo)
+
+        // Clear logs older than 3 month which have been serialized to disk
+        val deviceReportRecordStorage = CommCareApplication.instance().getUserStorage(DeviceReportRecord::class.java)
+        val reportsToDelete = Vector<Int>()
+        for (deviceReportRecord in deviceReportRecordStorage) {
+            val reportFile = File(deviceReportRecord.filePath)
+            if (reportFile.exists() && (reportFile.lastModified() < threeMonthsAgo.time)) {
+                reportFile.delete()
+                reportsToDelete.addElement(deviceReportRecord.id)
```
ctsims: This should occur whether or not the file exists, otherwise we'll never clear out the SQL entry for files 

(Error case would be "reportFile.delete()" executing successfully, then the process halting, then this code gets run again a second time, and since the file is gone we skip the removal for sql row)

shubham1g5: Right! Corrected it.

PR #2144, opened 2019-10-03
Streamlines App updates and user sync 

shubham1g5:
https://docs.google.com/document/d/1RYRPgmsvP0vUEkMJSJMHGtO0EprQK2P8kvwO0FKCSEA/edit

https://dimagi-dev.atlassian.net/browse/ICDS-737

### Summary of Changes

* Adds a custom property `pre-update-sync-needed` to turn on this feature
* Parses `x-commcarehq-appreleasedon` from response headers for app pofile requests and save it in preferences. If this header is not available, then guess this time as current time. 
* At the time of update install, compares the last sync time to app_release_time and blocks the update if required. 
* Put in a check on Home screen that auto triggers sync in case a blocked Staged update is present. 

The  gif below demostrates the whole UX for this - 

app_update_and_sync

Wiki: https://confluence.dimagi.com/pages/viewpage.action?spaceKey=ICDS&title=Pre+Update+Sync

cross-request: https://github.com/dimagi/commcare-core/pull/869


shubham1g5:

app/src/org/commcare/preferences/HiddenPreferences.java
```diff
@@ -320,4 +325,30 @@ public static boolean shouldShowUnsentFormsWhenZero() {
         SharedPreferences properties = CommCareApplication.instance().getCurrentApp().getAppPreferences();
         return properties.getString(SHOW_UNSENT_FORMS_WHEN_ZERO, PrefValues.NO).equals(PrefValues.YES);
     }
+
+    public static boolean preUpdateSyncNeeded() {
+        SharedPreferences properties = CommCareApplication.instance().getCurrentApp().getAppPreferences();
+        return properties.getString(PRE_UPDATE_SYNC_NEEDED, PrefValues.NO).equals(PrefValues.YES);
+    }
+
+    public static void setReleasedOnTimeForOngoingAppDownload(AndroidCommCarePlatform platform, long releasedOnTime) {
```
ctsims: There are a few places in the app where a pending update can be cleared, but I didn't see any way for this particular setter to be cleared out. Is there a situation where this gets zero'd because an update is abandoned?

app/src/org/commcare/preferences/HiddenPreferences.java
```diff
@@ -320,4 +325,30 @@ public static boolean shouldShowUnsentFormsWhenZero() {
...
+                .apply();
+    }
+
+    public static long geReleasedOnTimeForOngoingAppDownload() {
+        if (CommCareApplication.instance().getCurrentApp() == null) {
+            return 0;
+        }
+
+        return CommCareApplication.instance().getCurrentApp().getAppPreferences()
+                .getLong(RELEASED_ON_TIME_FOR_ONGOING_APP_DOWNLOAD, 0);
```
ctsims: I think we need a "sanity check" for this date/time. Once it's set, it is set pretty much forever and if, say, the phone had its date set wrong when the user downloaded the update (say, to the future), then failed to sync and then set the phone's time back to the correct time, this would constantly block them from updating.

Since the phone's date can change, I think we can only do that sanity checking when the value is *read* not when it is *written*, so it would seem to make sense to me to do it here.

app/src/org/commcare/android/resource/installers/ProfileAndroidInstaller.java
```diff
@@ -119,6 +128,29 @@ public boolean install(Resource r, ResourceLocation location, Reference ref,
         return false;
     }
 
+    private void storeReleasedTime(AndroidCommCarePlatform platform, Reference ref) {
+        long releasedOnTime = -1;
+        if (ref instanceof JavaHttpReference) {
+            Headers responseHeaders = ((JavaHttpReference)ref).getResponseHeaders();
```
ctsims: Slight nitpick / style point - It's a tradeoff of 'correctness' v. verbosity, but patterns like this kind of defeat the point of the interface/inheritance structure. 

I think the "Java" way to do this would be to make a new interface for References (like "MetadataSupportedReference") which instead of returning the headers raw would return the Released On Date directly, and then do (if ref instanceof MetadatasupportedReference) { ref.getReleasedOnDate()}

That keeps the HTTP-centric code in the HTTP centric class and lets another reference extend this same behavior without needing to modify the locations where the behavior is relied upon.

app/src/org/commcare/android/resource/installers/ProfileAndroidInstaller.java
```diff
@@ -119,6 +128,29 @@ public boolean install(Resource r, ResourceLocation location, Reference ref,
...
+            if (responseHeaders != null) {
+                String releasedOnStr = responseHeaders.get(HEADER_APP_RELEASED_ON);
+                try {
+                    if (releasedOnStr != null) {
+                        releasedOnTime =  new SimpleDateFormat("yyyy-MM-dd'T'HH:mm'Z'").parse(releasedOnStr).getTime();
+                    }
+                } catch (ParseException e) {
+                    e.printStackTrace();
```
ctsims: I'm not sure we actually want to eat this error silently. A parse error here means that HQ is setting the header, but in the wrong format, which means that the phones will all just silently fail to properly use the feature correctly, but without anyone's knowledge.

app/src/org/commcare/android/resource/installers/ProfileAndroidInstaller.java
```diff
@@ -79,6 +87,7 @@ public boolean install(Resource r, ResourceLocation location, Reference ref,
             throws UnresolvedResourceException, UnfullfilledRequirementsException {
         //First, make sure all the file stuff is managed.
         super.install(r, location, ref, table, platform, upgrade, recovery);
+        storeReleasedTime(platform, ref);
```
ctsims: I'm not sure the statefulness of all of these elements actually supports this workflow quite right.

In the current code workflow, 'ref' only actually retains the released time headers as long as the existing object is functional. Sometimes super.install will take multiple attempts across multiple runs to actually complete, but the profile will only actually get fetched remotely on the very first attempt, meaning that state will likely be gone by the time we reach here. 

I think all of this lifecycle management needs to be moved to

https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/engine/resource/AndroidResourceManager.java#L108

(the class, not that exact line of code)

where we actually semantically know what's happening with the update process. That'll make it clearer where we are in the process and what we are tracking about the update itself.

app/src/org/commcare/activities/HomeScreenBaseActivity.java
```diff
@@ -1130,6 +1134,9 @@ private boolean checkForPendingAppHealthActions() {
         if (RecoveryMeasuresHelper.recoveryMeasuresPending()) {
             finishWithExecutionIntent();
             kickedOff = true;
+        } else if (UpdateActivity.isUpdateBlockedOnSync()) {
```
ctsims: this method (checkForPendingAppHealthActions) triggers any time we get back to the home screen, not just any time the app starts up for the first time, right? 

I think that means that failures to sync will keep forcing the user to keep trying to sync after , say, every form, even if they can't currently sync. I think this needs some way to identify that the user has attempted to sync already in the current login, and only force a sync if one hasn't already happened, so we can trigger the right behavior (sync before update), but only once per session (or once per update attempt. If the user restarts the update and it's blocked on sync it should reset that flag back to force a sync)

app/src/org/commcare/android/resource/installers/ProfileAndroidInstaller.java
shubham1g5: This makes sense. Further Question though - Does this mean we should create another `ReferenceFactory` that instantiates this `MetadataSupportedReference` or use the current `JavaHttpRoot` that based on some condition (all profile urls ?) differentiate between `JavaHttpReference` and `MetadataSupportedReference` ?

app/src/org/commcare/activities/SyncCapableCommCareActivity.java
```diff
@@ -27,6 +25,9 @@
 import org.commcare.views.dialogs.CustomProgressDialog;
 import org.javarosa.core.services.locale.Localization;
 
+import androidx.annotation.AnimRes;
```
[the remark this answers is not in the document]
shubham1g5: They were present from before and diff is because of formatting. They are getting used as well.

app/src/org/commcare/preferences/HiddenPreferences.java
shubham1g5: Currently there is not, since any new update check overrides the previous written value for this pref. But agree it makes sense to still clean this to protect against any incorrect usage in future.

app/src/org/commcare/android/resource/installers/ProfileAndroidInstaller.java
ctsims: I don't think that we need to change anything other than the HttpReference to follow this pattern, the new metadata class would be something that's added directly to qualifying references, rather than something that changes how references work.

The old scenario was

`
interface ref {};
...
class href implements ref {}
...
if (r instanceof href) {}
`

the new one would be 

`
interface ref{};
...
interface ref_meta extends ref { getMetadata(); }
..
class href implements ref, ref_meta {
getMetadata() {};
}
...
if (r instanceof ref_meta) {}
`
So the only three files that change are the new Meta interface, the HttpReference itself, and this class

app/src/org/commcare/android/resource/installers/ProfileAndroidInstaller.java
shubham1g5: @ctsims Unfortunately I am having a hard time in imagining how this will work. Moving this code to `AndroidResourceManager` would mean that we will loose the access to the existing `ref` object with info about releasedOnTime. So are you suggesting that we store the `ref` object or make a static reference to`ref` from `AndroidResourceManager` ?

ctsims: Hey, you are right that this is harder than I thought. I forgot that we were still tracking the resource location in `AndroidResourceManager` as a string, and it only gets derived down to a reference through the IoC layer in `ResourceTable`. In order for my stated approach to make sense we'd have to change a bunch of code around so that the `ResourceLocation` was based on references (which I don't think is great).

After review, I think it's still not great that there's state sticking around inside of the reference object, but I can't think of a cleaner way to implement this than you have here, so let's keep the approach you are already using.

shubham1g5:
@ctsims  There was a suggestion from GS that in case auto-sync fails, we should make user navigate again to the Update page and show an error there stating that the update failed because sync could not succeed. We earlier decided to leave the user on Home Screen so that he can re-trigger sync manually though I think it makes sense to show the error to the user so that they get a clear sense that the update has failed and retry from update page itself. Thoughts ?

ctsims:
@shubham1g5 I agree that the usability would be improved if the sync failed for the phone to revert the user back to the update page, change the update icon to denote that pressing it will trigger a sync, and providing a message that says something akin to the fact that the sync attempt failed and a sync needs to happen before the update can apply itself.

I'd consider that work positive but optional, since on ICDS most active users are syncing very frequently. I might consider the following to be a reasonable middle ground

- When an update is "staged" (IE: fully downloaded and ready to be applied) but a sync is needed, as represented in your GIF by the "update to version XXX and log out" button, that the UI is different
- In the new UI the the icon is the "sync" icon and the text specifies that the app is ready to update but they need to sync first
- If the sync fails, the user is directed back to the update screen. This screen now by default would produce useful context to the user. If helpful new text can be added reflecting the sync issue, but I think in general that provides a fairly clear workflow even without it

ctsims:
@shubham1g5 One other thing I realized when responding - I can imagine a situation where a user is stuck in bad state where they need to apply an update in order to sync. I'm not sure if I'm overthinking, but I wonder if having some way to force the phone to clear the "released on" flag would be something we appreciated in the future...

ctsims:
@shubham1g5 Let me know when this is ready for a final look, by the way.

shubham1g5:
@ctsims  Yes, waiting for https://github.com/dimagi/commcare-hq/pull/25750 to be merged. Will update you once I have tested the feature after Jenny's PR gets deployed.

shubham1g5:
@ctsims  This is ready to be reviewed again, I have updated the GIF with new workflow in PR description.

app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -186,7 +187,7 @@ protected void initiateLoginAttempt(boolean restoreSession) {
             DevSessionRestorer.tryAutoLoginPasswordSave(uiController.getEnteredPasswordOrPin(), false);
         }
 
-        if (ResourceInstallUtils.isUpdateReadyToInstall()) {
+        if (ResourceInstallUtils.isUpdateReadyToInstall() && !UpdateActivity.isUpdateBlockedOnSync()) {
```
shubham1g5: @ctsims  Right now, we just restrict the auto updates here if a sync is needed and not auto trigger the sync afterwards the local login. Do you think it's the right behaviour or we should do a force sync once the user has logged in for these cases  ? Guessing we would not auto inititate the auto update (background) again anyway after login since the user is already inside the app.

ctsims: We do normally try to force syncs on Login, so it's not unusual for us to do so. Maybe the right behavior is to change the AutoSync behavior to `if ( autosync is pending OR update is blocked on sync)` which will mean that a sync attempt will trigger if needed every day but won't need to hijack the UX entirely the way that the pending update install will.?

One note - It occurs to me that we should also have some reporting mechanism that phones are stuck in this state - A device log event , and/or a

ctsims: (also fine to pursue these changes later, they definitely are not blocking)

app/src/org/commcare/android/resource/installers/ProfileAndroidInstaller.java
```diff
@@ -119,6 +130,21 @@ public boolean install(Resource r, ResourceLocation location, Reference ref,
...
 
+    private void storeReleasedTime(AndroidCommCarePlatform platform, Reference ref) throws ParseException {
+        long releasedOnTime = -1;
+
+        if (ref instanceof ReleasedOnTimeSupportedReference) {
+            releasedOnTime = ((ReleasedOnTimeSupportedReference)ref).getReleasedOnTime();
+        }
+
+        // If we don't have the release time, guess it as current time
+        if (releasedOnTime == -1) {
```
ctsims: I think this fallback isn't exactly right.

As-is, this change will immediately force every app to sync before update, which I don't think is quite the intent. 

I think this fallback is good if there is some other setting which controls whether the "Sync before Update" behavior triggers (like an app setting), which implies that the fallback for having a released on date should be syncing every time, but otherwise I think we should (until someone from the requirements team directs otherwise) not change the default behavior.

shubham1g5: @ctsims  We do have [this custom setting](https://github.com/dimagi/commcare-android/blob/appUpdateAndUserSync/app/src/org/commcare/preferences/HiddenPreferences.java#L363) `pre-update-sync-needed` which is needed currently to turn on the "Sync before update" workflow apart from the released on date.

shubham1g5:
@ctsims  When an user updates from App Manager, he gets the same "Sync to Update" button and get directed to `LoginActivity`. Once the user logs in, the sync->update workflow kicks in again similar to the normal use case.

app/src/org/commcare/activities/LoginActivity.java
shubham1g5: Makes sense, Made these changes in the last commit.

shubham1g5:
@ctsims  Bumping this as it's ready for a review again.

PR #2145, opened 2019-10-09
Fixes localization error message

shubham1g5:
In cases where the nolocatization exception is not the primary exception, the localization error gets hidden by the primary exception leaving user with a non concrete error message.  

Example Exception for which this PR fixes the error message - 

````
java.lang.RuntimeException: An error occurred while executing doInBackground()
        at android.os.AsyncTask$3.done(AsyncTask.java:353)
        at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:383)
        at java.util.concurrent.FutureTask.setException(FutureTask.java:252)
        at java.util.concurrent.FutureTask.run(FutureTask.java:271)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636)
        at java.lang.Thread.run(Thread.java:764)
     Caused by: org.javarosa.core.util.NoLocalizedTextException: The Localizer could not find a definition for ID: due_list_case_tile_mother_name in the 'kha' locale.
at org.javarosa.core.services.locale.Localizer.getText(Localizer.java:328)
        at org.javarosa.core.services.locale.Localization.get(Localization.java:21)
        at org.javarosa.core.services.locale.Localization.get(Localization.java:11)
````

### DIFF
--- a/app/src/org/commcare/utils/CommCareExceptionHandler.java
+++ b/app/src/org/commcare/utils/CommCareExceptionHandler.java
@@ -2,17 +2,18 @@ package org.commcare.utils;
 
 import android.content.Context;
 import android.content.Intent;
-import android.widget.Toast;
 
 import org.commcare.CommCareApplication;
 import org.commcare.activities.CrashWarningActivity;
-import org.commcare.activities.SessionAwareCommCareActivity;
 import org.commcare.android.logging.ForceCloseLogger;
 import org.commcare.recovery.measures.ExecuteRecoveryMeasuresActivity;
 import org.commcare.recovery.measures.RecoveryMeasuresHelper;
 import org.javarosa.core.util.NoLocalizedTextException;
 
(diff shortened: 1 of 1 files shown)


ctsims:
Hey,

I'm a bit worried that that this code takes something that was previously done twice via copy/paste (recursively inspecting the stack for a typed exception) and doing it a third time, which leaves the code vulnerable to subtle drifts in the paths each one of those methods takes.

The code itself seems fine, but if you have time I'd refactor a bit to do one step where you stackdive for a list of typed exceptions (if they exist) and then let the checks and message extraction happen directly on that pulled exception, rather than doing the stackdive on each action.

shubham1g5:
@ctsims  Nice catch. made some changes, let me know if they are not in alignment to what you were referring.

PR #2148, opened 2019-10-30
Error message improvements for 503 (Rate limiting) error

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/browse/ICDS-902

There was some discussion on Jira regarding wordings for this error. I am starting with being transparent first and open to iterations if we don't really wanna mention "servers unavailable/busy" to users encountering rate limiting errors. 

Product Note: Error message improvements for rate limiting errors encountered on login or sync.

### DIFF
--- a/app/assets/locales/android_translatable_strings.txt
+++ b/app/assets/locales/android_translatable_strings.txt
@@ -115,6 +115,7 @@ sync.fail.bad.data=Server provided improperly formatted data, please try again o
 sync.fail.bad.local=Your information was fetched, but a problem occurred with the log in. Please try again.
 sync.fail.bad.network=Couldn't contact server. Please make sure an internet connection is available or try again later.
 sync.fail.server.error=The server encountered an error processing your data. Please try again later. If the problem persists, contact CommCare Support.
+sync.fail.rate.limited.server.error=Our servers are unavailable at this time. Please try again later. If the problem persists, contact CommCare Support.
 sync.fail.unknown=An unexpected issue occurred during sync. Please try to sync again.
 sync.fail.unsent=Having issues communicating with the server to send forms. Will try again later.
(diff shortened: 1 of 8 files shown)


app/assets/locales/android_translatable_strings.txt
```diff
@@ -115,6 +115,7 @@ sync.fail.bad.data=Server provided improperly formatted data, please try again o
 sync.fail.bad.local=Your information was fetched, but a problem occurred with the log in. Please try again.
 sync.fail.bad.network=Couldn't contact server. Please make sure an internet connection is available or try again later.
 sync.fail.server.error=The server encountered an error processing your data. Please try again later. If the problem persists, contact CommCare Support.
+sync.fail.rate.limited.server.error=Our servers are unavailable at this time. Please try again later. If the problem persists, contact CommCare Support.
```
ctsims: I think we should remove the "contact CommCare Support" part of this message (and the one below).

Generally I think we should never in-code suggest to people that there is a "CommCare Support," but with this issue especially, a 503 error specifically is a message to the phone that there is nothing to report to support, and I don't want to increase the burden on an arbitrary support team if they are getting 503'd.

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -306,6 +306,8 @@ private static boolean shouldGenerateFirstKey() {
             return processErrorResponseWithMessage(pullResponse);
         } else if (responseCode == 500) {
             return handleServerError();
+        } else if (responseCode == 503) {
```
ctsims: I'm really confused, I thought I added this to a previous release. Did that change not get replicated to master?

shubham1g5: @ctsims  The change you made was for form submissions and not restore.

PR #2153, opened 2019-11-28
Handle null values in callout response keys

shubham1g5:
Fabric: https://www.fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/5a0cbe1d61b02d480d8329c3?time=last-ninety-days

### DIFF
--- a/app/src/org/commcare/android/javarosa/IntentCallout.java
+++ b/app/src/org/commcare/android/javarosa/IntentCallout.java
@@ -214,8 +214,8 @@ public class IntentCallout implements Externalizable {
 
                 // Get our response value
                 String responseValue = response.getString(key);
-                if (key == null) {
-                    key = "";
+                if (responseValue == null) {
+                    responseValue = "";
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/android/javarosa/IntentCallout.java
```diff
@@ -214,8 +214,8 @@ private boolean processOdkResponse(Intent intent, TreeReference intentQuestionRe
 
                 // Get our response value
                 String responseValue = response.getString(key);
-                if (key == null) {
```
shubham1g5: I don't see any point of resetting null keys to empty string. It looks like the intention here was to do this check on value instead of key and someone just made a mistake while coding.

ctsims: Agreed. In the original code that got pulled over the intent is a smidge unclear, but it's hard to believe that it was intended to function the way it was written.

PR #2154, opened 2019-11-29
Save as Incomplete Navigation fix for external launch

shubham1g5:
On saving form as incomplete, we end up relaunching the sesssion instance again instead of returning back to the calling application. This can be really confusing if CC was called with a session stack pointing to a form, on saving that form as incomplete, a new intance of the same form gets relaunched automatically. I think the incomplete form save here should follow the same navigation pattern as when we save a complete form. 

Related to: https://dimagi-dev.atlassian.net/browse/MOB-138

### DIFF
--- a/app/src/org/commcare/activities/HomeScreenBaseActivity.java
+++ b/app/src/org/commcare/activities/HomeScreenBaseActivity.java
@@ -803,11 +803,7 @@ public abstract class HomeScreenBaseActivity<T> extends SyncCapableCommCareActiv
     private void clearSessionAndExit(AndroidSessionWrapper currentState, boolean shouldWarnUser) {
         currentState.reset();
         if (wasExternal) {
-            if (shouldWarnUser) {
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/activities/HomeScreenBaseActivity.java
```diff
@@ -803,11 +803,7 @@ private boolean processReturnFromFormEntry(int resultCode, Intent intent) {
     private void clearSessionAndExit(AndroidSessionWrapper currentState, boolean shouldWarnUser) {
         currentState.reset();
         if (wasExternal) {
-            if (shouldWarnUser) {
```
shubham1g5: I am not sure why we are making a distinction here based on `shouldWarnUser`. We end up finishing the activity irrespective of value of `shouldWarnUser` below. So any warning message doesn't show up for external launches anyway.

PR #2158, opened 2019-12-23
Enable cleartext support by default

ShivamPokhriyal:
Problem: Starting with Android 9 (API level 28), cleartext support is disabled by default due to which on Android 9 devices, the installation using app code fails due to a security exception - "Cleartext communication to *** not permitted by network security policy."

Solution: Enable cleartext support by default in Manifest.


shubham1g5:

PR #2164, opened 2020-01-03
Use downsampled banner to prevent OOM Exceptions

ShivamPokhriyal:
https://dimagi-dev.atlassian.net/browse/MOB-136
This PR will use the downsampled banner image to prevent OOM Exceptions.


app/src/org/commcare/views/CustomBanner.java
```diff
@@ -55,4 +44,30 @@ public static boolean useCustomBanner(Context context,
         }
         return false;
     }
+
+    public enum Banner {
+        LOGIN, HOME;
```
shubham1g5: Good use of enums here 👍

PR #2169, opened 2020-01-14
Add days since last sync param to sync requests

ctsims:
When performing a sync request, includes the integer days since the last day the device has synced to the URL parameters of the request, ie:
`/a/no-feature-flags/phone/restore/07a82c079ce340a8ae705ef3d66089f7/?device_id=353522096503145&user_id=ce3719da49ef4b4ea8740b2e617ae6b5&days_since_last_sync=34&state=ccsh:ba14c90a2370ecd5778cbdf5d81370ff&version=2.0&items=true&since=2605adb21c2b11eabf720a5180e0c5bc`

The intent here is to allow HQ to create different pooled priorities for rate limiting based on how far behind a user is in syncing data, as such the data needs to be available to the proxy with no downstream processing. 

@shubham1g5 I am thinking that it makes the most sense for me to hotfix this as 2.47.5 as well, otherwise it won't make it out to users fast enough to be usable. Do you have reservations about that based on the code?

@snopoke This is what we were discussing this morning. HQ can ignore all other cases other than `days_since_last_sync=0`, which should be simple pattern matching. Any objections to that?

### DIFF
--- a/app/src/org/commcare/network/CommcareRequestGenerator.java
+++ b/app/src/org/commcare/network/CommcareRequestGenerator.java
@@ -11,6 +11,7 @@ import org.commcare.engine.cases.CaseUtils;
 import org.commcare.interfaces.CommcareRequestEndpoints;
 import org.commcare.models.database.SqlStorage;
 import org.commcare.provider.DebugControlsReceiver;
+import org.commcare.utils.SyncDetailCalculations;
 import org.javarosa.core.model.User;
 import org.javarosa.core.model.utils.DateUtils;
 
@@ -106,6 +107,11 @@ public class CommcareRequestGenerator implements CommcareRequestEndpoints {
             }
             if (digest != null) {
                 params.put("state", "ccsh:" + digest);
+            }
+
+            int getDaysSinceSync = SyncDetailCalculations.getDaysSinceLastSync();
+            if(getDaysSinceSync != -1) {
+                params.put("days_since_last_sync", Integer.toString(getDaysSinceSync));
             }
         }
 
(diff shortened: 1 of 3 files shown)


shubham1g5:
@ctsims That's a cool thought. Agree that we should hotfix it.  

I am going to initiate comms with ICDS team for the release since we generally get external approvals before a release and cc you on the thread.

app/src/org/commcare/network/CommcareRequestGenerator.java
```diff
@@ -107,6 +108,11 @@ public static CommcareRequestGenerator buildNoAuthGenerator() {
             if (digest != null) {
                 params.put("state", "ccsh:" + digest);
             }
+
+            int getDaysSinceSync = SyncDetailCalculations.getDaysSinceLastSync();
```
shubham1g5: Maybe use a header instead ? If not, just wanna double check since we are adding a new parameter, it won't result in any cache busting on HQ ?

snopoke: this won't impact caching. We did discuss a header vs a URL param. I don't think it makes any difference in terms of implementing this feature. We did think that it could be useful to have in the logs (headers don't appear in the logs) though we do also get this data from the heartbeat requests if we wanted to do more analysis

snopoke:
> HQ can ignore all other cases other than days_since_last_sync=0, which should be simple pattern matching. Any objections to that?

Looks good to me

app/src/org/commcare/network/CommcareRequestGenerator.java
shubham1g5: sounds good. Thanks!

snopoke:
@ctsims is it correct to assume that if there param will only exist for incremental syncs and not for first time syncs?

ctsims:
@snopoke That is correct, since the code is only injected when we are using the other sync flags.

That would be a good thing for us to add explicit testing for. I can see if there's an easy harness for me to inject that into.

ctsims:
@snopoke also - if you want to test anything, the code is live in the most recent build artifacts:
https://jenkins.dimagi.com/job/commcare-android-2.47/459160/
https://jenkins.dimagi.com/job/commcare-android/461382/

snopoke:
thanks

PR #2176, opened 2020-01-24
Update app to target Android 10 

ShivamPokhriyal:


app/src/org/commcare/CommCareApplication.java
```diff
@@ -410,20 +410,27 @@ public void initializeGlobalResources(CommCareApp app) {
     }
 
     @NonNull
+    @Deprecated
     public String getPhoneId() {
```
ShivamPokhriyal: Not Sure about the use-case of this particular method, but it won't work with android 10.
Android best practices suggest [See 3rd option](https://developer.android.com/training/articles/user-data-ids#best-practices-android-identifiers) using either a random UUID(we'll have to save it in preferences to make sure it doesn't change) or Firebase Instance ID([read here](https://firebase.google.com/docs/reference/android/com/google/firebase/iid/FirebaseInstanceId)). 

Though both of them have a drawback: if application is reinstalled or user clears app data then the id will be lost and a new id will be generated. 

Let me know, if this is something important and we should use one of the above approaches to get a id.

app/AndroidManifest.xml
```diff
@@ -64,6 +64,7 @@
         android:label="@string/application_name"
         android:hardwareAccelerated="true"
         android:usesCleartextTraffic="true"
+        android:requestLegacyExternalStorage="true"
```
ShivamPokhriyal: Read here for more information :: https://developer.android.com/training/data-storage#scoped-storage
If not enabled, this method throws error https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/models/ODKStorage.java#L22

ShivamPokhriyal: Also, https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/utils/MediaUtil.java#L500 BitmapFactory decodeFile fails if we try and decode the image chosen from gallery.

app/src/org/commcare/views/widgets/MediaWidget.java
```diff
@@ -237,21 +235,12 @@ public void setBinaryData(Object binaryuri) {
             }
         }
 
-        if (newMedia.exists()) {
-            // Add the copy to the content provider
```
ShivamPokhriyal: Though I'm not sure whether we should be adding all the files to the content provider. 
IMO, it makes sense for the video that are recorded using the record button in app, but for videos that are chosen from gallery, we just make a copy of that inside our app's directory so if we add it to content provider there'll be 2 copies of the same video shown to the user in the gallery app.

ShivamPokhriyal: PS: Though, this change will only prevent crashes and won't change the earlier behaviour.

ShivamPokhriyal:
@shubham1g5 @ctsims This is still in progress, but would like to know your opinions on the changes.

app/AndroidManifest.xml
shubham1g5: For Android 10 adding this sounds good. Though we should add Soft asserts log  statements to the code paths that we identified that were still accesing code paths. So that we know in future if they are indeed getting hit or not.

app/src/org/commcare/CommCareApplication.java
shubham1g5: So `deviceId` was useful to identify the 2 separate installations of the app. I think using Firebase Instance Id seems like a good workaround here to separate out the installations as we don't have any other id currently that we maintain that's unique per CC installation.

app/src/org/commcare/utils/MediaUtil.java
```diff
@@ -414,7 +414,7 @@ private static int getApproxScaleDownFactor(int newWidth, int originalWidth) {
         double widthScaleDownFactor =  (double)boundingWidth / originalWidth;
         // Choosing the larger of the scale down factors, so that the image still fills the entire
         // container
-        double dominantScaleDownFactor = Math.max(widthScaleDownFactor, heightScaleDownFactor);
+        double dominantScaleDownFactor = Math.min(widthScaleDownFactor, heightScaleDownFactor);
```
shubham1g5: Thanks for looking at it as part of this work. Though if it isn't realted to Android 10 work, would be nice to separate this out in another PR so that it's easy to revert in case of any edge cases.

app/src/org/commcare/views/widgets/MediaWidget.java
shubham1g5: Agree that we should not save it to content Providers. @ctsims we did remove it for Images sometime back in [here](https://github.com/dimagi/commcare-android/pull/2110/files), so guessing no objections in removing them for audio and videos as well.

app/src/org/commcare/views/widgets/MediaWidget.java
```diff
@@ -237,21 +235,12 @@ public void setBinaryData(Object binaryuri) {
...
-
-            Uri mediaUri =
-                    getContext().getContentResolver().insert(MediaStore.Audio.Media.EXTERNAL_CONTENT_URI, values);
-            String mediaUriString = mediaUri == null ? "null" : mediaUri.toString();
-            Log.i(TAG, "Inserting media returned uri = " + mediaUriString);
+        try {
+            FileUtil.addMediaToGallery(getContext(), newMedia);
             showToast("form.attachment.success");
-        } else {
+        } catch (Exception e) {
```
shubham1g5: We should let the app crash in case of exceptions here if we are going to retain this functionality.

app/AndroidManifest.xml
ctsims: We likely need to explicitly request permissions to the filepath beforehand the same way that we now do for things like camera widget capture. I think Shubham reviewed usage of File handles previously, but some of these places where raw paths are used likely didn't get caught.

ShivamPokhriyal:

app/src/org/commcare/utils/DeviceIdentifier.java
```diff
@@ -0,0 +1,41 @@
...
+     * Returns a unique(very highly likely) identifier for this device.
+     * This identifier is prefixed with "commcare_"
+     *
+     * This id may change when user re-installs the app or does a factory reset.
+     */
+    public static String getDeviceIdentifier(Context context) {
+        if (uuid == null) {
+            synchronized (DeviceIdentifier.class) {
+                if (uuid == null) {
+                    SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context);
```
ShivamPokhriyal: Using SharedPreference to store this unique id. We can also choose to create a text file and store UUID to that file. I'm not sure which option is more suitable of the two so using the sharedpreferences approach here.

app/src/org/commcare/activities/MultimediaInflaterActivity.java
```diff
@@ -162,6 +164,7 @@ private void searchForDefault() {
 
         if (Environment.MEDIA_MOUNTED.equals(Environment.getExternalStorageState()) ||
                 Environment.MEDIA_MOUNTED_READ_ONLY.equals(Environment.getExternalStorageState())) {
+            Logger.log(LogTypes.SOFT_ASSERT, "Access outside scoped storage in searching zip file");
```
shubham1g5: It seems like this path will definitely get hit if someone is using this functionality. So there is no point in adding soft assert here and we will likely need to disable this functionality in Android 11 if there is no other alternate provided in Android 11 for file searching.

app/src/org/commcare/recovery/measures/ScanCczTask.java
```diff
@@ -94,6 +96,7 @@ private void locateAppCcz(File[] files, int depth) {
         if (!StringUtils.isEmpty(lastKnownCczLocation)) {
             filePathsToScan.add(new File(lastKnownCczLocation));
         }
+        Logger.log(LogTypes.SOFT_ASSERT, "Access outside scoped storage in scanning cczs.");
```
shubham1g5: We can remove this soft assert as well as this path will definitely get hit for someone using this functionality,

app/src/org/commcare/utils/DeviceIdentifier.java
```diff
@@ -0,0 +1,41 @@
...
+    private static final String KEY_DEVICE_IDENTIFIER = "commcare_device_id";
+
+    /**
+     * Returns a unique(very highly likely) identifier for this device.
+     * This identifier is prefixed with "commcare_"
+     *
+     * This id may change when user re-installs the app or does a factory reset.
+     */
+    public static String getDeviceIdentifier(Context context) {
+        if (uuid == null) {
```
shubham1g5: there are 2 nested if here with `uuid == null` check. Is there a particular reason behind that ?

ShivamPokhriyal: Yeah, this pattern is also popular as _double-checked locking idiom_, used to reduce the overhead of acquiring a lock by first testing the locking criterion without actually acquiring the lock.

ShivamPokhriyal:

app/src/org/commcare/utils/DeviceIdentifier.java
shubham1g5: Interesting!

app/src/org/commcare/utils/DeviceIdentifier.java
```diff
@@ -0,0 +1,41 @@
...
+import android.content.Context;
+import android.content.SharedPreferences;
+import androidx.preference.PreferenceManager;
+import java.util.UUID;
+
+/**
+ * A Utility to get a unique identifier for app's installation.
+ * @author $|-|!˅@M
+ */
+public final class DeviceIdentifier {
```
shubham1g5: @ctsims  Goint to merge this to go ahead with code freeze though can you review this class and see the device identifier makes sense to you.

PR #2177, opened 2020-01-29
Change scaleDownFactor to make sure we don't load large image in imageView

ShivamPokhriyal:
In some cases the scaleDownFactor for a really large image comes out to be 1, so the bitmap is not scaled down at all and loading that bitmap in imageview crashes the app. 
This PR will fix it.

### DIFF
--- a/app/src/org/commcare/utils/MediaUtil.java
+++ b/app/src/org/commcare/utils/MediaUtil.java
@@ -414,7 +414,7 @@ public class MediaUtil {
         double widthScaleDownFactor =  (double)boundingWidth / originalWidth;
         // Choosing the larger of the scale down factors, so that the image still fills the entire
         // container
-        double dominantScaleDownFactor = Math.max(widthScaleDownFactor, heightScaleDownFactor);
+        double dominantScaleDownFactor = Math.min(widthScaleDownFactor, heightScaleDownFactor);
 
         int widthImposedByContainer = (int)Math.round(originalWidth * dominantScaleDownFactor);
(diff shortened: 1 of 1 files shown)


ctsims:
@ShivamPokhriyal i'm wondering whether we should be adding tests or other workflow code here to clarify the behavior of these functions. I believe this function is working as expected, and it's a different call that happens later which is intended to further reduce the size.

Specifically, the method signature describes that the intent of the function is to initially scale down the image explicitly to ensure that it still completely fills the passed in container in both dimensions, which means that the resulting scaled down image should not be shorter than the bounding box in either direction.

```
    /**
     * @return The smallest dimensions that both preserve the original aspect ratio, and mean
     * that the image still fills the container. It is unimportant to scale down exactly to the
     * container size; we just don't want to be creating a bitmap that is way bigger than necessary.
     */
```

That matches your described output from slack (and the behavior of 'max')

```
INPUT::
int originalHeight = 3335,
int originalWidth = 8000,
int boundingHeight = 1824,
int boundingWidth = 1080
OUTPUT::
width:: 4375 (this is wayyyy bigger than the bounding width)
height:: 1824
```
I'm definitely not trying to claim that it's the most useful output for a function to provide this behavior, but it does look like the function is behaving as written. I think we should be looking outside of this function for the buggy or incorrect behavior, which may be relying on this function instead of one that is more useful.

shubham1g5:
@ShivamPokhriyal bumping this for your response. If it's not relevant any more feel free to close.

PR #2168, opened 2020-01-08
Fix issue with mediacontroller visibility while navigation and scroll

ShivamPokhriyal:
https://dimagi-dev.atlassian.net/browse/MOB-117
**Problem**: Since [The MediaController will create a default set of controls and put them in a window floating above your application.](https://developer.android.com/reference/android/widget/MediaController), the mediacontroller will not respect parent view's animation or scroll. 

**Solution**: Add the mediaController to the view hierarchy to force it always conforms parent view's animation and scroll. 

**Extra**: Prevent logging inline video events if the video is never played.

PR #2178, opened 2020-01-29
Fix visibility issue of MediaController while navigation and scrolling

ShivamPokhriyal:
https://dimagi-dev.atlassian.net/browse/MOB-117
**Problem**: Since The MediaController will create a default set of controls and put them in a window floating above your application., the mediacontroller will not respect parent view's animation or scroll.

**Solution**: Add the mediaController to the view hierarchy to force it always conforms parent view's animation and scroll.

**Extra**: Prevent logging inline video events if the video is never played.

PS: A duplicate of this https://github.com/dimagi/commcare-android/pull/2168, since Jenkins builds gets really messy and produces some weird logs for failure even though there isn't one. 

Product Note: UI improvements for Inline Video

### DIFF
--- a/app/src/org/commcare/views/media/CommCareVideoView.java
+++ b/app/src/org/commcare/views/media/CommCareVideoView.java
@@ -39,6 +39,10 @@ public class CommCareVideoView extends VideoView {
     @Override
     protected void onDetachedFromWindow() {
         super.onDetachedFromWindow();
+        // Do not log events if the video is never played.
+        if (duration == 0) {
+            return;
+        }
         listener.onVideoDetached(duration);
     }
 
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/views/media/CommCareVideoView.java
```diff
@@ -39,6 +39,10 @@ public void start() {
     @Override
     protected void onDetachedFromWindow() {
         super.onDetachedFromWindow();
+        // Do not log events if the video is never played.
+        if (duration == 0) {
```
shubham1g5: Can we always fire the listener and do this check in the listener implementation. Would also be good to separate this out from this PR.

ShivamPokhriyal: Separated out this change here https://github.com/dimagi/commcare-android/pull/2179

ShivamPokhriyal:

ShivamPokhriyal:

PR #2180, opened 2020-01-30
Auto update scheduling

shubham1g5:
Summary of changes 
---------------------

* Introduces `UpdateWorker` which gets triggered by WorkManager periodically every couple hours to check for updates in background 

* Refactors `UpdateTask` to rely on `UpdateHandler` which is abstraction for common update related functionality between `UpdateTask` and `UpdateWorker`

* Removes old auto updates code 

* Introduces a new `AppInstallStatus`  `Cancelled` to denote cancellation of update tasks

* Binds the `UpdateActivity` UI to `UpdateWorker` when the worker task is in progress

Release Note: Implements a more robust background scheduling mechanism for Auto-updates


shubham1g5:
@ctsims bumping this.

ctsims:
thanks for the patience, reviewing today.

app/src/org/commcare/utils/AppLifecycleUtils.java
```diff
@@ -50,6 +52,9 @@ public static void uninstall(ApplicationRecord record) {
         record.setStatus(ApplicationRecord.STATUS_DELETE_REQUESTED);
         ccInstance.getGlobalStorage(ApplicationRecord.class).write(record);
 
+        // cancel all Workmanager tasks for this app
+        WorkManager.getInstance(CommCareApplication.instance()).cancelAllWorkByTag(record.getApplicationId());
```
ctsims: This needs to occur before line 45 above, since once the instance is unseated we would start to experience issues if the worker continued until line 56.

I'd actually recommend that we add this "cancelAllWorkByTag" action directly to the `CCInstance.unseat()`method in general

app/unit-tests/src/org/commcare/android/tests/application/AutoUpdateTest.java
```diff
@@ -38,92 +26,6 @@
...
-    private final static String REF_BASE_DIR = "jr://resource/commcare-apps/update_tests/";
-    private final static String username = "test";
-    private final static String password = "123";
-
-    /**
-     * Check that logging out after an auto-update check failed once should
-     * trigger the auto-update to resume.
-     */
-    @Test
-    public void testAppAutoUpdateLogoutRetry() {
```
ctsims: Is there a reason we aren't replicating a similar test? 

I think in general we should consider it a poor outcome to replace one process that was tested with another that had no tests.

app/src/org/commcare/CommCareApplication.java
```diff
@@ -755,6 +761,30 @@ public void onServiceDisconnected(ComponentName className) {
...
 
+    // Hand off an app update task to the Android WorkManager
+    private void scheduleAppUpdate() {
+        Constraints constraints = new Constraints.Builder()
+                .setRequiredNetworkType(NetworkType.CONNECTED)
+                .setRequiresBatteryNotLow(true)
+                .build();
+
+        PeriodicWorkRequest updateRequest =
+                new PeriodicWorkRequest.Builder(UpdateWorker.class, PERIODICITY_FOR_UPDATE_IN_HOURS, TimeUnit.HOURS)
```
ctsims: It looks like this more or less takes us to a context where we retry auto update every 2 hours, right?

Server load-wise I think that's quite often if we're presuming that we tried to contact the server, succeeded, and saw there was no update.

app/src/org/commcare/update/UpdateWorker.kt
```diff
@@ -0,0 +1,105 @@
...
+            updateHelper.OnUpdateCancelled()
+        }
+
+        updateHelper.OnUpdateComplete(updateResult)
+
+        cleanUp()
+
+        return when (updateResult.data.isUpdateInCompletedState) {
+            true -> Result.success()
+            else -> Result.retry()
```
ctsims: I'm not sure this resonates with me.

I think we only want to be applying the backoff-controlled-retries if there was a network reason for the update to fail. The UnknownError state, for example, doesn't feel like it should necessarily be triggering up to (5 min, 25 min, 10 hr) 3 retries over the course of the day before the user is logged out.

app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -238,11 +238,22 @@ protected boolean checkForSeatedAppChange() {
...
         return false;
     }
 
+
+    // cancels all worker tasks for previously seated app
+    private static void disableWorkForLastSeatedApp() {
+        SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(CommCareApplication.instance());
+        String lastSeatedId = prefs.getString(KEY_LAST_APP, "");
+        if(!lastSeatedId.isEmpty()) {
+            WorkManager.getInstance(CommCareApplication.instance()).cancelAllWorkByTag(lastSeatedId);
```
ctsims: Quick note that this request isn't synchronous. I'm not sure how killing the update tasks worked when we unseated apps, but if you haven't checked it would be worth confirming that the semantics are the same.

app/src/org/commcare/activities/UpdateActivity.java
```diff
@@ -128,22 +135,61 @@ private void loadSavedInstanceState(Bundle savedInstanceState) {
...
-            try {
-                updateTask.registerTaskListener(this);
-            } catch (TaskListenerRegistrationException e) {
-                Log.e(TAG, "Attempting to register a TaskListener to an already " +
-                        "registered task.");
-                uiController.errorUiState();
+        if (isAutoUpdateInProgress()) {
+            connectToUpdateWorker();
+        } else {
+            updateTask = UpdateTask.getRunningInstance();
```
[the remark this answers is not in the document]
shubham1g5: Right with the difference being `UpdateTask` is the reliable mechanism for foreground updates while `UpdateWorker` is only used for background ones. WorkManager is the replacement for traditional Android services kind of work and is not recommended for doing one off forground task that needs to happen on a user trigger as there is no way we can schedule a WorkManager task immediately.

app/src/org/commcare/activities/LoginActivity.java
shubham1g5: From the [note here](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/engine/resource/AndroidResourceManager.java#L280-L281), it looks like instead of cancelling the auto update for previous app, we were just firing the update request for the new seated app during the auto update process. We cancel all running tasks today [on logout](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/CommCareApplication.java#L322) which takes care of update request getting cancelled on logout and then when the user logins, we reinititate the task for the current app. We can follow the same semantics for cancelling the WorkManager tasks though since app updates can work fine even when the user is logged out, we can choose to just cancel work on app seating changes to keep the last update downloads running even when user is logged out.

app/src/org/commcare/CommCareApplication.java
shubham1g5: So the internal `UpdateWorker` [check here ](https://github.com/dimagi/commcare-android/blob/autoUpdateScheduling/app/src/org/commcare/update/UpdateWorker.kt#L63) will still respect the auto update frequency set for the app which should take care of server getting hit too often.

app/src/org/commcare/update/UpdateWorker.kt
shubham1g5: Makes sense. I was trying to immitate the earlier behaviour but agree that we should only retry on network errors.

app/unit-tests/src/org/commcare/android/tests/application/AutoUpdateTest.java
shubham1g5: this test checks whether a failed update resumes after the logout. Though since we are currently not cancelling the UpdateWorker on logout but seat change, I don't think this test serves any purpose. Plus earlier we were using our own code to schedule the app updates which made sense to test. Though now since WorkManager takes care of scheduling, it's not guaranteed for an update to resume right away after login.

app/src/org/commcare/activities/UpdateActivity.java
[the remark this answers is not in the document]
ctsims: Got it, that's frustrating in terms of the framework (especially since they have different reporting mechanisms back to the UI layer). The design makes sense given those constraints

app/src/org/commcare/update/UpdateWorker.kt
ctsims: Got it, that makes sense. Thanks for the revision.

app/unit-tests/src/org/commcare/android/tests/application/AutoUpdateTest.java
ctsims: Is there a best practice for how people test WorkManager related behaviors that Google advises? 

I agree that this wasn't a great test, but it did result in *some* code coverage of execution, so I'm a bit gun shy about that shift

app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -238,11 +238,22 @@ protected boolean checkForSeatedAppChange() {
...
             return true;
         }
         return false;
     }
 
+
+    // cancels all worker tasks for previously seated app
+    private static void disableWorkForLastSeatedApp() {
+        SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(CommCareApplication.instance());
+        String lastSeatedId = prefs.getString(KEY_LAST_APP, "");
```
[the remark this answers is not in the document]
shubham1g5: Moved it up [here](https://github.com/dimagi/commcare-android/blob/autoUpdateScheduling/app/src/org/commcare/activities/LoginActivity.java#L240).

app/unit-tests/src/org/commcare/android/tests/application/AutoUpdateTest.java
shubham1g5: @ctsims In terms of code coverage of our update code we do have [AppUpdateTest](https://github.com/dimagi/commcare-android/blob/master/app/unit-tests/src/org/commcare/android/tests/application/AppUpdateTest.java) that runs through most update scenarios. 

For testing WorkManager itself, Android has some [instrumentation tests APIs ](https://developer.android.com/topic/libraries/architecture/workmanager/how-to/testing). I am trying to currently figure out how to use them with Roblectric but facing an issue described [here](https://stackoverflow.com/questions/60222892/modifying-robolectrics-instrumentation-configuration-with-androidx-tests).

app/src/org/commcare/CommCareApplication.java
shubham1g5: @ctsims  Wondering if we should still respect the update frequency set in app properties for triggering the background update  worker. We might derive more value in terms of update uptake from ignoring this flag and instead have a more liberal frequency for update check that doesn't cause the server load to spike too much.

shubham1g5:
@ctsims  made some changes based to change periodicity based on the auto-update-frequency setting, would you be able to review the last commit https://github.com/dimagi/commcare-android/pull/2180/commits/3b928dd1acd5376fe7cbfd4cd99ca9b36b27b80a

shubham1g5:
@ctsims  bump on reviewing [last commit](https://github.com/dimagi/commcare-android/pull/2180/commits/3b928dd1acd5376fe7cbfd4cd99ca9b36b27b80a)

shubham1g5:
> If we think the code as written today for "Daily" is as often as we'd want to be doing the polling, no need for a new choice.
@ctsims  Sorry for dropping the conversation earlier. Since we are doing checks every 8 hours in Daily option, I do think that's as often as we should allow to have a good balance between overloading the server too much and keeping the app updates robust.

ctsims:
> > If we think the code as written today for "Daily" is as often as we'd want to be doing the polling, no need for a new choice.
> 
> @ctsims Sorry for dropping the conversation earlier. Since we are doing checks every 8 hours in Daily option, I do think that's as often as we should allow to have a good balance between overloading the server too much and keeping the app updates robust.

👍

shubham1g5:
@ctsims  can you re-approve this to be merged, last merge dismissed the last approval.

PR #2181, opened 2020-01-30
Port Android Pie changes for LTS

shubham1g5:
Unfortunately it skipped my mind that 2.47 didn't have Android pie changes. So porting them here to be able to make the release as we can no longer make releases without pie support

Original PR - https://github.com/dimagi/commcare-android/pull/2147


shubham1g5:
Self mergins since the original PR was already approved.

PR #2183, opened 2020-02-17
Upgrade to the Firebase Crashlytics SDK

ShivamPokhriyal:


app/src/org/commcare/utils/CrashUtil.java
```diff
@@ -29,35 +21,36 @@
...
+            FirebaseCrashlytics crashlytics = FirebaseCrashlytics.getInstance();
+            crashlytics.setCustomKey(DOMAIN, ReportingUtils.getDomain());
+            crashlytics.setCustomKey(APP_VERSION, ReportingUtils.getAppVersion());
+            crashlytics.setCustomKey(APP_NAME, ReportingUtils.getAppName());
         }
     }
 
     public static void registerUserData() {
         if (crashlyticsEnabled) {
-            Crashlytics.setUserName(ReportingUtils.getUser());
```
shubham1g5: Can we set the username as the custom key here ?

ShivamPokhriyal: Yeah, [setUserIdentifier is now setUserId. setUserName and setUserEmail are removed.](https://firebase.google.com/docs/crashlytics/upgrade-sdk?platform=android#setuseridentifier_is_now_setuserid_setusername_and_setuseremail_are_removed). I'll add the user name as a custom key.

shubham1g5: Unfortunately custom keys in firebase are not searchable and we definitely want to be able to search by username. Any ideas how can we achieve that with Firebase Crashlytics ?

ShivamPokhriyal: I guess this view(search using user name) may change when fabric is unavailable since firebasecrashlytics allows only 1 user identifier. We can though [link firebase to bigquery](https://support.google.com/firebase/answer/6318765#working_with_firebase_crashlytics_data_in_bigquery) which will allow us to search using custom keys as well.  

PS: Link requires project owner permissions in firebase so I cannot try it out.

shubham1g5: @ShivamPokhriyal  Gave you the required permissions. Would also be good to understand the cost implications of linking to bigquery.

shubham1g5:
@ShivamPokhriyal we should prioritize work on this for 2.49 as Fabric sdk is only going to receive crashes till November this year. Therefore we should be releasing a new update much before that so that most users update to the new version by November.

app/src/org/commcare/utils/CrashUtil.java
shubham1g5: In case the bigquery stuff doesn't work out, can we set the userIdentifier as username so that we can still retain search by username.

ShivamPokhriyal: It does work out farely well actually. You can see some queries using custom keys [here](https://console.cloud.google.com/bigquery?project=commcare-a57e4). Though, it sounds good to add username as userIdentifier, or perhaps a combination of userId and userName like(userId::userName) as userIdentifier?

shubham1g5: > It does work out fairly well actually. You can see some queries using custom keys `here.`
Awesome, is it free to use or are there any cost implications for us to use it ?

>  perhaps a combination of userId and userName like(userId::userName) as userIdentifier?

Does this allows us to search by any of the params or we will need to know the userId of the user to be able to search ?

ShivamPokhriyal: >Awesome, is it free to use or are there any cost implications for us to use it ?

It appears to me that using BigQuery over crashlytics data is completely [free](https://stackoverflow.com/a/54120786/6671572). 

> Does this allows us to search by any of the params or we will need to know the userId of the user to be able to search ?

Yes it does. You can check out the complete [schema here](https://console.cloud.google.com/bigquery?project=commcare-a57e4&p=commcare-a57e4&d=firebase_crashlytics&t=org_commcare_dalvik_ANDROID&page=table) and it has `user.name`, `user.email` and `user.id` fields there.

shubham1g5: > Yes it does. You can check out the complete schema here and it has user.name, user.email and user.id fields there.

I meant if we set the useridentifier as `userId::userName` will be able to search by any of the params in firebase ? If not, I think it's best to set it to username only to facilitate quick search from Crashlytics itself without needing BigQuery.

PR #2184, opened 2020-02-18
Handle captive portals in network requests

ShivamPokhriyal:
https://dimagi-dev.atlassian.net/browse/SAAS-10308

Fine grained network analysis before making sync/install/login requests.

### DIFF
--- a/app/assets/locales/android_translatable_strings.txt
+++ b/app/assets/locales/android_translatable_strings.txt
@@ -381,6 +381,15 @@ notification.sync.connections.title=No Connection
 notification.sync.connections.detail=No active data connections
 notification.sync.connections.action=You don't have any active data connections configured. Try connecting your phone to wifi or turning mobile data on.
 
+
+connection.captive_portal.title=Captive Portal Detected
+connection.captive_portal.detail=Cannot access internet
+connection.captive_portal.action=You are in a captive portal. Try authenticating into captive portal using browser or turning mobile data on.
+
+connection.commcare_blocked.title=CommCare APIs blocked
+connection.commcare_blocked.detail=Cannot access commcare
+connection.commcare_blocked.action=Commcare APIs are blocked. Try using a different wifi network or turning mobile data on.
+
 multiple.apps.unverified.title=Missing Multimedia for all Apps
 multiple.apps.unverified.message=All of your available apps are missing their associated multimedia resources. Please notify your supervisor so that he/she can fix this via the CommCare App Manager page.
 
--- a/app/assets/locales/android_translatable_strings.txt
+++ b/app/assets/locales/android_translatable_strings.txt
@@ -376,6 +376,8 @@ connection.task.success=No problems were detected.
 connection.task.report.commcare.popup=Your logs are being sent to CommCare!
 connection.task.unset.posturl=There has been an error on CommCare's end. Please report this problem to CommCare Support.
 
+connection.check.title=Checking Your Network Connection...
+connection.check.message=We are checking your network connection before proceeding with actual network call.
 
 notification.sync.connections.title=No Connection
 notification.sync.connections.detail=No active data connections
@@ -384,7 +386,7 @@ notification.sync.connections.action=You don't have any active data connections 
 
 connection.captive_portal.title=No Internet Detected
 connection.captive_portal.detail=Cannot access internet
-connection.captive_portal.action=You are in a wifi that requires sign-in from a browser. Try authenticating using browser or turning mobile data on.
+connection.captive_portal.action=You are connected to a network that requires sign-in from a browser. Try authenticating using browser or switching to mobile data
 
 connection.commcare_down.title=CommCare Servers Down
 connection.commcare_down.detail=Cannot access commcare
--- a/app/assets/locales/android_translatable_strings.txt
+++ b/app/assets/locales/android_translatable_strings.txt
@@ -382,13 +382,13 @@ notification.sync.connections.detail=No active data connections
 notification.sync.connections.action=You don't have any active data connections configured. Try connecting your phone to wifi or turning mobile data on.
 
 
-connection.captive_portal.title=Captive Portal Detected
+connection.captive_portal.title=No Internet Detected
 connection.captive_portal.detail=Cannot access internet
-connection.captive_portal.action=You are in a captive portal. Try authenticating into captive portal using browser or turning mobile data on.
+connection.captive_portal.action=You are in a wifi that requires sign-in from a browser. Try authenticating using browser or turning mobile data on.
 
-connection.commcare_blocked.title=CommCare APIs blocked
-connection.commcare_blocked.detail=Cannot access commcare
-connection.commcare_blocked.action=Commcare APIs are blocked. Try using a different wifi network or turning mobile data on.
+connection.commcare_down.title=CommCare Servers Down
+connection.commcare_down.detail=Cannot access commcare
+connection.commcare_down.action=There is an issue with CommCare's server. Please report this problem to CommCare.
 
 multiple.apps.unverified.title=Missing Multimedia for all Apps
 multiple.apps.unverified.message=All of your available apps are missing their associated multimedia resources. Please notify your supervisor so that he/she can fix this via the CommCare App Manager page.
--- a/app/src/org/commcare/activities/CommCareSetupActivity.java
+++ b/app/src/org/commcare/activities/CommCareSetupActivity.java
@@ -33,7 +33,6 @@ import org.commcare.fragments.InstallConfirmFragment;
 import org.commcare.fragments.InstallPermissionsFragment;
 import org.commcare.fragments.SelectInstallModeFragment;
 import org.commcare.fragments.SetupEnterURLFragment;
-import org.commcare.google.services.analytics.AnalyticsParamValue;
 import org.commcare.google.services.analytics.FirebaseAnalyticsUtil;
 import org.commcare.interfaces.RuntimePermissionRequester;
 import org.commcare.android.database.global.models.ApplicationRecord;
@@ -41,20 +40,19 @@ import org.commcare.logging.DataChangeLog;
 import org.commcare.logging.DataChangeLogger;
 import org.commcare.preferences.GlobalPrivilegesManager;
 import org.commcare.resources.model.InvalidResourceException;
-import org.commcare.resources.model.Resource;
 import org.commcare.resources.model.UnresolvedResourceException;
 import org.commcare.tasks.ConnectionDiagnosticTask;
 import org.commcare.tasks.ResourceEngineListener;
 import org.commcare.tasks.ResourceEngineTask;
 import org.commcare.tasks.RetrieveParseVerifyMessageListener;
 import org.commcare.tasks.RetrieveParseVerifyMessageTask;
+import org.commcare.utils.AppInstallationUtil;
 import org.commcare.utils.ConsumerAppsUtil;
 import org.commcare.utils.MultipleAppsUtil;
 import org.commcare.utils.Permissions;
 import org.commcare.views.ManagedUi;
 import org.commcare.views.dialogs.CustomProgressDialog;
 import org.commcare.views.dialogs.DialogCreationHelpers;
-import org.commcare.views.notifications.MessageTag;
 import org.commcare.views.notifications.NotificationMessage;
 import org.commcare.views.notifications.NotificationMessageFactory;
 import org.javarosa.core.reference.InvalidReferenceException;
@@ -65,8 +63,6 @@ import org.javarosa.core.util.PropertyUtils;
 import java.io.IOException;
 import java.security.SignatureException;
 import java.util.List;
-
-import static org.commcare.CommCareNoficationManager.AIRPLANE_MODE_CATEGORY;
 
 /**
  * Responsible for identifying the state of the application (uninstalled,
@@ -123,7 +119,7 @@ public class CommCareSetupActivity extends CommCareActivity<CommCareSetupActivit
     public static final int GET_APPS_FROM_HQ = 5;
 
     // dialog ID
-    private static final int DIALOG_INSTALL_PROGRESS = 4;
+    public static final int DIALOG_INSTALL_PROGRESS = 4;
 
     private boolean startAllowed = true;
     private String incomingRef;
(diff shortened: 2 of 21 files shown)


app/assets/locales/android_translatable_strings.txt
```diff
@@ -381,6 +381,15 @@ notification.sync.connections.title=No Connection
 notification.sync.connections.detail=No active data connections
 notification.sync.connections.action=You don't have any active data connections configured. Try connecting your phone to wifi or turning mobile data on.
 
+
+connection.captive_portal.title=Captive Portal Detected
```
shubham1g5: Our users won't know what a captive portal is. We should just say that No Internet Available and then explain in details that they are connected to a wifi they need to sign in into.

app/src/org/commcare/activities/StandardHomeActivity.java
```diff
@@ -98,9 +101,32 @@ void syncButtonPressed() {
                     AnalyticsParamValue.SYNC_MODE_SEND_FORMS,
                     AnalyticsParamValue.SYNC_FAIL_NO_CONNECTION);
             return;
+        } else {
```
shubham1g5: Since we are performing the granular task check here, the check above feels repeatetive. Is is possible to merge the two?

app/assets/locales/android_translatable_strings.txt
```diff
@@ -381,6 +381,15 @@ notification.sync.connections.title=No Connection
 notification.sync.connections.detail=No active data connections
 notification.sync.connections.action=You don't have any active data connections configured. Try connecting your phone to wifi or turning mobile data on.
 
+
+connection.captive_portal.title=Captive Portal Detected
+connection.captive_portal.detail=Cannot access internet
+connection.captive_portal.action=You are in a captive portal. Try authenticating into captive portal using browser or turning mobile data on.
+
+connection.commcare_blocked.title=CommCare APIs blocked
```
shubham1g5: This doesn't really mean that CC is blocked but more possibly CC servers are down, the string "connection.task.commcare.html.fail" provides a better message for the suer in these scenarios.

app/src/org/commcare/tasks/ConnectionDiagnosticTask.java
```diff
@@ -67,15 +81,19 @@ public ConnectionDiagnosticTask(Context c) {
...
+
     @Override
-    protected Test doTaskBackground(Void... params) {
-        Test out = null;
+    protected NetworkState doTaskBackground(Void... params) {
+        NetworkState out = NetworkState.CONNECTED;
         if (!isOnline(this.c)) {
-            out = Test.isOnline;
+            out = NetworkState.DISCONNECTED;
         } else if (!pingSuccess(googleURL)) {
```
shubham1g5: So not getting a 200 response from a url doesn't mean that it's a captive portal but just that network is unavailable. Similarly getting 200 response doesn't mean that user is connected but the response might be supplied through captive portal. We might wanna try https://stackoverflow.com/a/14030276

app/unit-tests/src/org/commcare/android/mocks/CommcareDiagnosticTaskMock.java
```diff
@@ -0,0 +1,30 @@
...
+import org.commcare.android.shadows.ShadowAsyncTaskNoExecutor;
+import org.commcare.tasks.ConnectionDiagnosticTask;
+import org.robolectric.annotation.Implementation;
+import org.robolectric.annotation.Implements;
+
+/**
+ * @author $|-|!˅@M
+ */
+@Implements(ConnectionDiagnosticTask.class)
+public class CommcareDiagnosticTaskMock extends ShadowAsyncTaskNoExecutor {
```
shubham1g5: nice!

app/unit-tests/src/org/commcare/android/CommCareTestRunner.java
```diff
@@ -33,6 +34,7 @@ public InstrumentationConfiguration createClassLoaderConfig(final FrameworkMetho
         InstrumentationConfiguration.Builder builder = new InstrumentationConfiguration.Builder(super.createClassLoaderConfig(method));
         builder.addInstrumentedPackage("net.sqlcipher.database.SQLiteDatabase");
         builder.addInstrumentedPackage("org.commcare.models.encryption");
+        builder.addInstrumentedClass(ConnectionDiagnosticTask.class.getName());
```
shubham1g5: Is just shadowing with the `CommcareDiagnosticTaskMock` class  not enough ?

app/src/org/commcare/network/HttpCalloutTask.java
```diff
@@ -116,6 +125,16 @@ protected HttpCalloutOutcomes doTaskBackground(Object... params) {
         return doPostCalloutTask(calloutFailed);
     }
 
+    protected boolean isCaptivePortal() throws IOException {
```
shubham1g5: This definitely is a better check for captive portal than what we do in `ConnectionDiagnosticTask`. We should abstract this function to  `ConnectivityStatus` and use it at both places.

app/unit-tests/src/org/commcare/android/CommCareTestRunner.java
ShivamPokhriyal: Hey, I'm not sure from where I got this, but it seems like the Shadow class won't be invoked while testing unless I add it to the build runner. The best possible reason I can find is [here](https://stackoverflow.com/a/29641926/6671572) that in order to shadow our own classes(originally we can only shadow android classes) we need to add the class in the custom test runner.

app/src/org/commcare/activities/CommCareSetupActivity.java
```diff
@@ -803,15 +771,23 @@ public void failTargetMismatch() {
...
-            return generateNormalInstallDialog(taskId);
+        switch (taskId) {
+            case DIALOG_INSTALL_PROGRESS:
+                if (isSingleAppBuild()) {
+                    return ConsumerAppsUtil.getGenericConsumerAppsProgressDialog(taskId, true);
+                } else {
+                    return generateNormalInstallDialog(taskId);
+                }
+            case ConnectionDiagnosticTask.CONNECTION_ID:
+                String title = Localization.get("connection.test.run.title");
```
shubham1g5: Nitpick: These strings made sense when user was explicity running a connection test by going into settings, but not so much when these connection test are part of another process. I would add new strings for this use case - 
title - "Checking Your Network Connection..."  , message - "We are checking your network connection to be able to proceed with installation"

app/src/org/commcare/utils/AppInstallationUtil.java
```diff
@@ -0,0 +1,123 @@
...
+ */
+public class AppInstallationUtil {
+
+    public static void startResourceInstall(Context context, String reference,
+                                            CommCareTaskConnector connector, CommCareApp ccApp,
+                                            boolean shouldSleep, int lastInstallMode) {
+        if (lastInstallMode == INSTALL_MODE_OFFLINE) {
+            startResourceInstall(reference, connector, ccApp, shouldSleep, lastInstallMode);
+        } else {
+            ConnectionDiagnosticTask<CommCareSetupActivity> task = new ConnectionDiagnosticTask(context);
```
shubham1g5: would be good to explicity typecast `new ConnectionDiagnosticTask(context)` as `new ConnectionDiagnosticTask<>(context)`

app/src/org/commcare/utils/AppInstallationUtil.java
```diff
@@ -0,0 +1,123 @@
...
+import static org.commcare.activities.CommCareSetupActivity.INSTALL_MODE_SMS;
+import static org.commcare.activities.CommCareSetupActivity.INSTALL_MODE_URL;
+
+/**
+ * @author $|-|!˅@M
+ */
+public class AppInstallationUtil {
+
+    public static void startResourceInstall(Context context, String reference,
+                                            CommCareTaskConnector connector, CommCareApp ccApp,
```
shubham1g5: should be typecasted as `CommCareTaskConnector<CommCareSetupActivity>`

app/src/org/commcare/utils/AppInstallationUtil.java
```diff
@@ -0,0 +1,123 @@
...
+            case INSTALL_MODE_FROM_LIST:
+                return AnalyticsParamValue.FROM_LIST_INSTALL;
+            case INSTALL_MODE_MANAGED_CONFIGURATION:
+                return AnalyticsParamValue.MANAGED_CONFIG_INSTALL;
+            default:
+                return "";
+        }
+    }
+
+    private static void startResourceInstall(String reference, CommCareTaskConnector connector,
```
shubham1g5: `CommCareTaskConnector<CommCareSetupActivity>`

app/src/org/commcare/utils/ConnectivityStatus.java
```diff
@@ -5,10 +5,36 @@
...
+import java.io.IOException;
+import java.util.HashMap;
+
+import okhttp3.ResponseBody;
+import retrofit2.Response;
+
 /**
  * @author Phillip Mates (pmates@dimagi.com)
  */
 public class ConnectivityStatus {
```
shubham1g5: Good job abstracting things here 👍

app/src/org/commcare/tasks/ConnectionDiagnosticTask.java
```diff
@@ -5,30 +5,25 @@
...
-public abstract class ConnectionDiagnosticTask<R> extends CommCareTask<Void, String, ConnectionDiagnosticTask.Test, R> {
+public class ConnectionDiagnosticTask<R> extends CommCareTask<Void, String, NetworkState, R> {
     private final Context c;
 
-    public enum Test {
-        isOnline,
-        googlePing,
-        commCarePing
-    }
+    private ConnectionDiagnosticListener listener;
```
shubham1g5: `ConnectionDiagnosticListener<R>`

app/assets/locales/android_translatable_strings.txt
```diff
@@ -381,6 +381,15 @@ notification.sync.connections.title=No Connection
 notification.sync.connections.detail=No active data connections
 notification.sync.connections.action=You don't have any active data connections configured. Try connecting your phone to wifi or turning mobile data on.
 
+
+connection.captive_portal.title=No Internet Detected
+connection.captive_portal.detail=Cannot access internet
+connection.captive_portal.action=You are in a wifi that requires sign-in from a browser. Try authenticating using browser or turning mobile data on.
```
shubham1g5: Nitpick: "You are connected to a network that requires sign-in from a browser. Try authenticating using browser or switching to mobile data"

app/src/org/commcare/activities/ConnectionDiagnosticActivity.java
```diff
@@ -55,31 +53,32 @@ protected void onCreate(Bundle savedInstanceState) {
...
+                                    receiver.txtInteractiveMessages.setText(Localization.get("connection.task.commcare.html.fail"));
+                                    receiver.txtInteractiveMessages.setVisibility(View.VISIBLE);
+                                    receiver.reportButton.setVisibility(View.VISIBLE);
+                                    break;
+                                case DISCONNECTED:
+                                case CAPTIVE_PORTAL:
+                                    //get the appropriate display message based on what the problem is
+                                    String displayMessage = failedTest == ConnectivityStatus.NetworkState.DISCONNECTED ?
+                                            Localization.get("connection.task.internet.fail")
+                                            : Localization.get("connection.task.remote.ping.fail");
```
shubham1g5: Instead of giving a general message like `connection.task.remote.ping.fail` , since we have pretty strong evidence that we are inside a captive portal , we should switch to giving a captive portal specific message to the user here.

app/assets/locales/android_translatable_strings.txt
```diff
@@ -376,11 +376,22 @@ connection.task.success=No problems were detected.
 connection.task.report.commcare.popup=Your logs are being sent to CommCare!
 connection.task.unset.posturl=There has been an error on CommCare's end. Please report this problem to CommCare Support.
 
+connection.check.title=Checking Your Network Connection...
+connection.check.message=We are checking your network connection before proceeding with actual network call.
```
shubham1g5: "actual netowrk call" seems a bit techy for users. Can we make these strings specific to the process they are related to like syncing, app install etc ?

shubham1g5:
@ctsims  Would you be able to give this a quick look, just wanna confirm the overall approach of preempting the actual network process with another connection checking task looks good to you.

app/assets/locales/android_translatable_strings.txt
ctsims: ==. It's fine to just say "We are checking the setup of your network connection" or something

ctsims:
Hey @shubham1g5 It's going to take me some time to have enough free time to dig through this.

Does this approach *always* do a first connection? If so that seems too negative to our users to me, since connection latency is their biggest issue.

shubham1g5:
> Does this approach always do a first connection? If so that seems too negative to our users to me, since connection latency is their biggest issue.

@ctsims That's right. Another approach on this can be more retroactive. For ex in case of sync, if user get a 200 response though the response body fails to parse, Do a captive portal check  and show error accordingly to the user ?

ctsims:
@shubham1g5 Yeah, that would make a bit more sense to me. The other option I can think of is identifying the captive portal in flight. I believe that a Captive Portal will attempt to issue an HTTP redirect to CommCare which the app should reject because it isn't able to get a valid HTTPS connection. Rather than wedging in-between all requests we could potentially capture that behavior, trigger a separate captive portal detection and then report the captive portal info back in a way that isn't direct.

app/src/org/commcare/utils/ConnectivityStatus.java
```diff
@@ -23,4 +49,21 @@ public static boolean isNetworkAvailable(Context context) {
         NetworkInfo activeNetworkInfo = connectivityManager.getActiveNetworkInfo();
         return activeNetworkInfo != null && activeNetworkInfo.isConnectedOrConnecting();
     }
+
+    public static NetworkState checkCaptivePortal() throws IOException {
+        // CommCare has its own URL for detecting Captive Portals.
+        // It must return an HTTP status code of 200 and a body containing "success".
+        String captivePortalURL = "http://www.commcarehq.org/serverup.txt";
```
ctsims: `www.commcarehq.org` isn't the only CommCare server on the internet, so we should be wary of reporting things like "CommCare Down." 

It's fine for us to use an "owned" URL to detect a network condition (since we are reporting something about the user's local connection), but if we are going to report on Server Status it should always be in reference to a Real Server that the user has attempted to communicate with, not our primary prod one.

shubham1g5: yeah, good catch.

shubham1g5:
>  I believe that a Captive Portal will attempt to issue an HTTP redirect to CommCare which the app should reject because it isn't able to get a valid HTTPS connection.

@ctsims  My sense was that a captive portal will block all requests to CC until user signs in for say in a password protected wifi portal. Any reasoning behind why you think it's going to redirect to CC ?

ctsims:
> > I believe that a Captive Portal will attempt to issue an HTTP redirect to CommCare which the app should reject because it isn't able to get a valid HTTPS connection.
> 
> @ctsims My sense was that a captive portal will block all requests to CC until user signs in for say in a password protected wifi portal. Any reasoning behind why you think it's going to redirect to CC ?

Oh, Sorry, I meant that the captive portal will attempt to redirect the user to the Captive Portal's URL..

Generally the way they work is that that they try to issue a redirect response to the HTTP request on behalf of the original request (to www.commcarehq.org, for instance), but that they can't do that for our requests because they fail the HTTPS exchange. We might get enough information in the failed request to identify that, though. I haven't inspected what a captive portal redirect failure looks like in a while.

ShivamPokhriyal:
Closing this in favour of #2206

PR #2185, opened 2020-02-25
Fixes a crash on opening form that's part of a pending update on phone

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/browse/ICDS-1264

Issue Repro: 

- Install an App and login
- Initiate an update , though don't install it
- Kill CC (to force app re-initialization)
- Open CC again 
- Go to a form that's part of the update 
- CC Crashes

Explaination: The crash happens when there is an update in progress or there is an update downloaded that’s not been installed and user tries to open a form that’s been updated in the pending update. When a form is installed as part of update, it c[reates formDefRecords](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/android/resource/installers/XFormAndroidInstaller.java#L107) into user storage. When an app which has a pending update gets initialized, we loop through forms to find the [latestFormId](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/android/resource/installers/XFormAndroidInstaller.java#L86) for the particular namespace. [This code](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/android/database/app/models/FormDefRecord.java#L161) was also looping through the forms that were part of the update causing `getLatestFormDefId` to return the formId of the updated form which in turn caused the right xform to not be mapped to namespace ever because of the check [here](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/android/resource/installers/XFormAndroidInstaller.java#L77).  When user opens a form that's part of a pending update, this causes a lookup for formId -1 into sql causing a crash. It's worth to note that this issue fixes itself once the pending update has been installed. 

Solution: This PR implements a resource version check in `getLatestFormDefId` to exclude any forms that has resource version greater than the resource version of the resource being initialized and thus excluding all the forms that are saved into DB because of the update. 

cross-request: https://github.com/dimagi/commcare-core/pull/880

### DIFF
--- a/app/src/org/commcare/activities/components/FormEntryInstanceState.java
+++ b/app/src/org/commcare/activities/components/FormEntryInstanceState.java
@@ -4,6 +4,8 @@ import android.content.Intent;
 import android.os.Bundle;
 import android.util.Pair;
 
+import org.commcare.CommCareApp;
+import org.commcare.CommCareApplication;
 import org.commcare.activities.FormEntryActivity;
 import org.commcare.android.database.app.models.FormDefRecord;
 import org.commcare.android.database.user.models.FormRecord;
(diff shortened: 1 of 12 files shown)


shubham1g5:
Closing in favour of https://github.com/dimagi/commcare-android/pull/2186/files

PR #2186, opened 2020-02-25
Fixes form load crash in case of an pending update

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/browse/ICDS-1264

Issue Repro:

- Install an App and login
- Initiate an update , though don't install it
- Kill CC (to force app re-initialization)
- Open CC again
- Go to a form that's part of the update
- CC Crashes

Explaination: The crash happens when there is an update in progress or there is an update downloaded that’s not been installed and user tries to open a form that’s been updated in the pending update. When a form is installed as part of update, it [creates formDefRecords](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/android/resource/installers/XFormAndroidInstaller.java#L107) into user storage. When an app which has a pending update gets initialized, we loop through forms to find the [latestFormId](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/android/resource/installers/XFormAndroidInstaller.java#L86) for the particular namespace. [This code](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/android/database/app/models/FormDefRecord.java#L161) was also looping through the forms that were part of the update causing getLatestFormDefId to return the formId of the updated form which in turn caused the right xform to not be mapped to namespace ever because of the check [here](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/android/resource/installers/XFormAndroidInstaller.java#L77). When user opens a form that's part of a pending update, this causes a lookup for formId -1 into sql causing a crash. It's worth to note that this issue fixes itself once the pending update has been installed.

Solution:  Instead of looking throgh the formDefRecords to decide whether to register a namespace or not this PR always registers a namespace if it doesn't exist and only overwrites in case the resource version of the new form is greater than the existing form.

### DIFF
--- a/app/src/org/commcare/android/resource/installers/XFormAndroidInstaller.java
+++ b/app/src/org/commcare/android/resource/installers/XFormAndroidInstaller.java
@@ -78,7 +78,7 @@ public class XFormAndroidInstaller extends FileSystemInstaller {
         if (platform.getFormDefId(namespace) == -1) {
             platform.registerXmlns(namespace, formDefId);
         } else {
-            // Only overwrite a form if the resourceVersion of the new form is higher
+            // Only overwrites the existing form if the resourceVersion of the new form is higher than the existing one
             FormDefRecord existingForm = FormDefRecord.getFormDef(platform.getFormDefStorage(), platform.getFormDefId(namespace));
             FormDefRecord newForm = FormDefRecord.getFormDef(platform.getFormDefStorage(), formDefId);
             if (newForm.getResourceVersion() > existingForm.getResourceVersion()) {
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/android/resource/installers/XFormAndroidInstaller.java
```diff
@@ -74,19 +74,21 @@ public boolean initialize(AndroidCommCarePlatform platform, boolean isUpgrade) t
...
         super.initialize(platform, isUpgrade);
-        if (isLatestFormRecord(platform)) {
+
+        if (platform.getFormDefId(namespace) == -1) {
             platform.registerXmlns(namespace, formDefId);
+        } else {
+            // Only overwrites the existing form if the resourceVersion of the new form is higher than the existing one
+            FormDefRecord existingForm = FormDefRecord.getFormDef(platform.getFormDefStorage(), platform.getFormDefId(namespace));
+            FormDefRecord newForm = FormDefRecord.getFormDef(platform.getFormDefStorage(), formDefId);
+            if (newForm.getResourceVersion() > existingForm.getResourceVersion()) {
```
ctsims: Small point that I think this should be `>=`. 

There could be a situation where the old resource and new resource have the same version and the old resource is on the way out. In that case the signal that this current resource is installed (because it is being initialized, which only happens to installed resources) should imply that this resource is valid, so with no other information about the installed resource we should presume that it may not be.

PR #2188, opened 2020-03-02
Backmerge Commcare 2.47.6

shubham1g5:


app/AndroidManifest.xml
```diff
@@ -3,8 +3,10 @@
           xmlns:tools="http://schemas.android.com/tools"
           package="org.commcare.dalvik"
           android:versionCode="106"
+          android:versionName="2.47.6">
```
ShivamPokhriyal: > `android:versionName="2.47.6"`

Will it create problem?

shubham1g5: yup! thanks for catching.

PR #2192, opened 2020-03-04
Fix mismatched attachments while editing a saved form 

ShivamPokhriyal:
Fix for: https://dimagi-dev.atlassian.net/browse/SAAS-10659

We will now save the form to disk as soon as there is a change in widgets containing attachments.

### DIFF
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -703,7 +703,7 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
         saveDataToDisk(FormEntryConstants.EXIT, false, null, true);
     }
 
-    private void updateFormMediaToDisk() {
+    protected void updateFormMediaToDisk() {
         // Works only when we are editing a saved form.
         FormRecord formRecord = FormRecord.getFormRecord(formRecordStorage, FormEntryInstanceState.mFormRecordPath);
         if (formRecord == null) {
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -352,7 +352,7 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
     public void saveImageWidgetAnswer(String imagePath) {
         uiController.questionsView.setBinaryData(imagePath, mFormController);
         saveAnswersForCurrentScreen(FormEntryConstants.DO_NOT_EVALUATE_CONSTRAINTS);
-        updateFormMediaToDisk();
+        onExternalAttachmentUpdated();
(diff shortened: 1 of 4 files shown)


app/src/org/commcare/activities/FormEntryActivityUIController.java
```diff
@@ -737,6 +738,12 @@ protected void updateFormRelevancies() {
                 shouldRemoveFromView.add(i);
                 continue;
             }
+
+            //If there was change(in particular image-remove) in an image widget, then update the form
+            if (oldWidgets.get(i) instanceof ImageWidget) {
+                activity.updateFormMediaToDisk();
```
ctsims: Shouldn't we check the value of the old and new widgets in this case and make sure they haven't changed? I think we call updateFormRelevencies in a number of situations beyond when media is discarded, and it could cause performance problems if we needed to save the form to disk frequently.

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -701,6 +703,18 @@ private void saveIncompleteFormToDisk() {
         saveDataToDisk(FormEntryConstants.EXIT, false, null, true);
     }
 
+    protected void updateFormMediaToDisk() {
```
ctsims: I think the name is a bit confusing. Can I suggest calling it something like `onFormMediaUpdated` or `onExternalAttachmentUpdated`? 

What the method is doing is somewhat clear from the code, but the context about why we are doing it might be confusing to the next programmer, so this might be a good place to document why the method is called and what should be done.

app/src/org/commcare/activities/FormEntryActivityUIController.java
```diff
@@ -716,6 +717,8 @@ protected void updateFormRelevancies() {
                 FormRelevancyUpdating.getOldSelectChoicesForEachWidget(oldWidgets);
         ArrayList<String> oldQuestionTexts =
                 FormRelevancyUpdating.getOldQuestionTextsForEachWidget(oldWidgets);
+        ArrayList<IAnswerData> oldAnswers =
```
ctsims: I'm really, really sorry for making this suggestion and then retracting it, but I'm almost positive that this change will have a much stronger runtime impact, since getting the answer from a widget can sometimes trigger deferred computation that results in other behavior. I think what you are doing here is the best way to do this check, though, so now that I see how we'd have to implement it, I think we should go back to making the save each time.

Sorry for the extra work.

shubham1g5: More of an FYI that this revert seems to be resulting in https://dimagi-dev.atlassian.net/browse/SUPPORT-18064?focusedCommentId=298192 when the image question is in a question list and we probably need to make a change here. @avazirna planning to debug more and explore this further.

PR #2193, opened 2020-03-05
[no op] Removes syncPending 

shubham1g5:
I didn't see any particular use case of maintaining this var `syncPending` and hence removing it to simplify some code.


shubham1g5:
@ctsims  bumping this

shubham1g5:
That's right. Thanks

PR #2195, opened 2020-03-05
Form submissions Scheduling

shubham1g5:
* Schedules FormSumbmission tasks using Work manager
* Refactors `ProcessAndSendTask` to user `FormSubmissionHelper`
* Removes the explicit trigger for sending forms on login as it should be triggered by Work Manager in timely fashion now


app/src/org/commcare/activities/FormAndDataSyncer.java
```diff
@@ -36,39 +36,29 @@
...
+    void startUnsentFormsTask(SyncCapableCommCareActivity activity,
+                              final boolean syncAfterwards,
+                              boolean userTriggered) {
 
         // We only want to update the last upload sync time when it's a blocking sync
         if (syncAfterwards) {
             HiddenPreferences.updateLastUploadSyncAttemptTime();
         }
 
-        if (records.length > 0) {
```
shubham1g5: made this code part of the [background task ](https://github.com/dimagi/commcare-android/blob/formSubmissionsScheduling/app/src/org/commcare/sync/FormSubmissionHelper.java#L79)

app/src/org/commcare/activities/FormAndDataSyncer.java
```diff
@@ -142,14 +132,9 @@ protected void handleCancellation(SyncCapableCommCareActivity receiver) {
         }
 
         processAndSendTask.connect(activity);
-        processAndSendTask.executeParallel(records);
+        processAndSendTask.executeParallel();
     }
 
-    private static String getFormPostURL(final Context context) {
```
shubham1g5: Moved to `FormSubmissionHelper`

app/src/org/commcare/activities/SyncCapableCommCareActivity.java
```diff
@@ -75,15 +75,12 @@ public void onAttachedToWindow() {
      * triggered after they are submitted. If no forms are sent, triggers a sync explicitly.
      */
     protected void sendFormsOrSync(boolean userTriggeredSync) {
-        boolean formsSentToServer = checkAndStartUnsentFormsTask(true, userTriggeredSync);
-        if (!formsSentToServer) {
```
shubham1g5: we now return a success result from upload forms task in case the # of forms to upload is 0 which in turn triggers the data pull.

app/src/org/commcare/utils/FormUploadUtil.java
```diff
@@ -134,13 +128,12 @@ public static FormUploadResult sendInstance(int submissionNumber, File folder,
...
 
-        if (hasListener) {
-            myListener.startSubmission(submissionNumber, bytes);
+        if (listener != null) {
+            listener.startSubmission(submissionNumber, bytes);
         }
 
         if (files.length == 0) {
             Log.e(TAG, "no files to upload");
-            listener.cancel(true);
```
shubham1g5: Removed this since we should not cancel upload for all forms because of an error with a single form and allow code to follow the [error handling here](https://github.com/dimagi/commcare-android/blob/formSubmissionsScheduling/app/src/org/commcare/sync/FormSubmissionHelper.java#L188-L202) for this case.

shubham1g5:
@ctsims  Should The worker tasks still respect the frequency set by `cc-auto-sync` flag even though the work submits the form in background. I see no harm in sendign forms in background frequently though it might cause server load to shoot up significantly. So wondering what's your take on this.

shubham1g5:
@ctsims  This now implements the heartbeat flag for disabling the work. Feel free to go ahead with review on it. 

With the backoffs, it seems like the retries keep backing off even beyond the periodic frequency with a max backoff of 5 hours. So with retries the next Run time can be greater than the normal periodic frequency. So I think it's safe to have the normal periodic frequency set to an hour or even less.

app/src/org/commcare/utils/FormUploadUtil.java
shubham1g5: > Specifically: There are lots of situations where we should cancel an upload because of an error with a single form. If there's a form we're having issues with that we don't understand, if we block submission we retain the synchronicity between the mobile and the server until we can correct it. If we let stuff through we risk letting the phone and server drift significantly.

@ctsims Agree that we might want to cancel the whole task in some specific error in a single form. Though talking specifically for this case, the [error handling](https://github.com/dimagi/commcare-android/blob/formSubmissionsScheduling/app/src/org/commcare/sync/FormSubmissionHelper.java#L316)  (I put a wrong link earlier) seems more robust as it takes into account whether the FNF is due to storage not being available or not and halt/continue the form upload for all forms depending on that which seems more robust to me than cancelling the task in all cases. Would we still want to cancel the upload for all forms even if the issue is with one form for the sake of mobile server parity ?

app/src/org/commcare/heartbeat/HeartbeatRequester.java
```diff
@@ -96,10 +98,20 @@ public void parseResponse(JSONObject responseAsJson) {
...
             attemptCczUpdateParse(responseAsJson);
             checkForForceLogs(responseAsJson);
+            checkForDisableBackgroundWork(responseAsJson);
         }
         DriftHelper.clearMaxDriftSinceLastHeartbeat();
     }
 
+    private void checkForDisableBackgroundWork(JSONObject responseAsJson) {
+        boolean disableBackgroundWork = responseAsJson.optBoolean("disable_background_work", false);
+        HiddenPreferences.setDisableBackgroundWork(disableBackgroundWork);
```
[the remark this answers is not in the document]
shubham1g5: Right, I think it makes sense then to enable these on next login in that case and then disable again if the flag in heartbeat still conveys that we should stop ?

app/src/org/commcare/utils/FormUploadUtil.java
shubham1g5: @ctsims  bump

app/src/org/commcare/heartbeat/HeartbeatRequester.java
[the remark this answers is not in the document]
shubham1g5: @ctsims  bump

app/src/org/commcare/utils/FormUploadUtil.java
ctsims: @shubham1g5 You are right that it looks like the general form processing loop will catch a FileNotfound error correctly if storage isn't available, so agreed that we can let it propagate and not catch it redundantly..

app/src/org/commcare/heartbeat/HeartbeatRequester.java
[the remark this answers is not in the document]
ctsims: I'm not sure how many phones successfully get a heartbeat every day (and when in the day they get it if they do, some CAS users get their heartbeat in the afternoon), so it might be worth waiting slightly longer than a day, but the pattern you describe sounds right.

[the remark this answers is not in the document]
shubham1g5: @ctsims [Changed the behaviour](https://github.com/dimagi/commcare-android/pull/2195/commits/32aaf0774baec7fcb676ceaef1b22c96dbd48ed0) for work to be disabled for 36 hours since the phone encounters the heartbeat flag to disable background work.

shubham1g5:
@ctsims  bumping it up as it's still awaiting approval to be merged. Last set of changes are [here](https://github.com/dimagi/commcare-android/pull/2195/commits/32aaf0774baec7fcb676ceaef1b22c96dbd48ed0)

ctsims:
Looks good to me, thanks for the long march on this one!

PR #2196, opened 2020-03-17
Commcare 2.48

ShivamPokhriyal:


app/AndroidManifest.xml
```diff
@@ -3,7 +3,7 @@
           xmlns:tools="http://schemas.android.com/tools"
           package="org.commcare.dalvik"
           android:versionCode="106"
-          android:versionName="2.49">
+          android:versionName="2.48.0">
```
shubham1g5, line 6: @ShivamPokhriyal Flagging that this should have been 2.49 (master points to the next major version)

PR #2197, opened 2020-03-19
Fix for navigation of saved form

ShivamPokhriyal:
https://forum.dimagi.com/t/a-validation-condition-in-the-preview/7333

In case of validation condition on a non-required question, navigation of submitted form gives constraint violation error for validation error

Product Note: Fixes a bug that causes validation errors to appear while navigating saved forms.

### DIFF
--- a/app/src/org/commcare/activities/FormEntryActivityUIController.java
+++ b/app/src/org/commcare/activities/FormEntryActivityUIController.java
@@ -285,7 +285,7 @@ public class FormEntryActivityUIController implements CommCareActivityUIControll
 
         // The answer is saved on a back swipe, but question constraints are ignored.
         if (activity.currentPromptIsQuestion()) {
-            activity.saveAnswersForCurrentScreen(FormEntryConstants.DO_NOT_EVALUATE_CONSTRAINTS);
+            activity.saveAnswersForCurrentScreen(false);
         }
 
         // Any info stored about the last changed widget is useless when we move to a new view
@@ -717,7 +717,7 @@ public class FormEntryActivityUIController implements CommCareActivityUIControll
         ArrayList<String> oldQuestionTexts =
                 FormRelevancyUpdating.getOldQuestionTextsForEachWidget(oldWidgets);
 
-        activity.saveAnswersForCurrentScreen(FormEntryConstants.DO_NOT_EVALUATE_CONSTRAINTS);
+        activity.saveAnswersForCurrentScreen(false);
 
         FormEntryPrompt[] newValidPrompts;
         try {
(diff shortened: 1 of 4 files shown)


app/src/org/commcare/activities/FormEntryActivityUIController.java
```diff
@@ -392,7 +392,7 @@ private void showNextView(boolean resuming) {
         }
 
         if (activity.currentPromptIsQuestion()) {
-            if (!activity.saveAnswersForCurrentScreen(FormEntryConstants.EVALUATE_CONSTRAINTS)) {
+            if (!activity.saveAnswersForCurrentScreen(!activity.mFormController.isFormReadOnly())) {
```
ctsims: What was the old flag set for? I would have guessed this would at the very least be changed to 

`FormEntryConstants.EVALUATE_CONSTRAINTS and !activity.mFormController.isFormReadOnly()`

ctsims:
Hm, I think this is kind of prime-time for us to be responsive to the global community, so I'd support a hotfix.

We should make sure to publicly respond to the forum note in addition to any tickets if/when the hotfix is released

app/src/org/commcare/activities/FormEntryActivityUIController.java
shubham1g5: @ctsims  `FormEntryConstants.EVALUATE_CONSTRAINTS` is a constant that equates to `true`, so it doesn't really make a difference.  Do you want to include it more for semantics ?

shubham1g5:
Makes sense. That will mean we need to do a hotfix 2.48.1 on top of 2.48 and instead of incrementally rolling out 2.48 (halt rollout), do the incremental rollout for 2.48.1. 

Alternatively we can let 2.48 fully rollout over next week and do a non-incremental 2.48.1 rollout afterwards. I think this is a better option since in both cases it will take us a week to propogate the update to all users.

shubham1g5:
thanks! @ShivamPokhriyal  Would you be able to open up a hotfix branch and create this PR against that. [This page](https://confluence.dimagi.com/pages/viewpage.action?pageId=35064144) describes all you should need though do reach out in case something breaks. I have bumped the release % on 2.48 to 50 today and On Tuesday instead of bumping 2.48 to 100 we can just make 2.48.1 release rolling it out to 100% of our users.

app/src/org/commcare/activities/FormEntryActivityUIController.java
ShivamPokhriyal: FYI @ctsims

ctsims: @shubham1g5 I think the best way to say it is that either the `FormEntryConstants.EVALUATE_CONSTRAINTS` does something (in which case we should leave it in the code) or that it's antiquated and we don't use it anymore, in which case we should remove the definition of the value in addition to its use here. I'm not comfortable with only doing one.

shubham1g5: That makes sense! I don't see any particular use of them and thought they were there just for better semantics. Definitely pro removing them. cc: @ShivamPokhriyal

PR #2199, opened 2020-03-24
Add forms with status complete to the unsent filter

ShivamPokhriyal:
This'll make forms with status COMPLETE and UNSENT appear same to the user.

### DIFF
--- a/app/src/org/commcare/activities/FormRecordListActivity.java
+++ b/app/src/org/commcare/activities/FormRecordListActivity.java
@@ -97,9 +97,9 @@ public class FormRecordListActivity extends SessionAwareCommCareActivity<FormRec
     public enum FormRecordFilter {
 
         /**
-         * Processed and Pending
+         * Submitted and Pending
          **/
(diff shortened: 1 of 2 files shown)


shubham1g5:
@ShivamPokhriyal  Think completed forms should only be added in the Submitted Forms filter to correctly represent the semantics here.

ShivamPokhriyal:
> @ShivamPokhriyal Think completed forms should only be added in the Submitted Forms filter to correctly represent the semantics here.

Not sure, since user can't quarantine submitted form. Moreover it represents forms that are sent to server. But completed are unprocessed forms and we should give user ability to quarantine these forms. Though user will get an option to delete. 
 
And if we add it to Submitted filter, then we'll have to add it to submitted and pending filter as well. 

@shubham1g5

shubham1g5:
@ShivamPokhriyal Ahh yes,  I got confused meaning of `STATUS_SAVED` . This looks good to me Thanks!

app/src/org/commcare/activities/FormRecordListActivity.java
```diff
@@ -99,7 +99,7 @@
         /**
          * Processed and Pending
```
shubham1g5: Can we change this comment to `Submitted and Pending` instead ?

PR #2201, opened 2020-03-26
Sets Pre update Sync needed from the staged update

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/browse/QA-1120

### DIFF
--- a/app/src/org/commcare/engine/resource/AndroidResourceManager.java
+++ b/app/src/org/commcare/engine/resource/AndroidResourceManager.java
@@ -9,6 +9,7 @@ import org.commcare.CommCareApplication;
 import org.commcare.google.services.analytics.FirebaseAnalyticsUtil;
 import org.commcare.logging.analytics.UpdateStats;
 import org.commcare.preferences.HiddenPreferences;
+import org.commcare.preferences.PrefValues;
 import org.commcare.resources.ResourceManager;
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/engine/resource/AndroidResourceManager.java
```diff
@@ -304,5 +305,6 @@ private long exponentionalRetryDelay(int numberOfRestarts) {
     public void clearUpgrade() {
         super.clearUpgrade();
         HiddenPreferences.setReleasedOnTimeForOngoingAppDownload((AndroidCommCarePlatform)platform, 0);
+        HiddenPreferences.setPreUpdateSyncNeeded(PrefValues.NO);
```
ctsims: Hm, only thing about this PR that I'm not sure about. This will unset the flag even if the current profile installed (not staged) had it set. 

That is likely totally fine (and likely difficult to work around), but wanted to quickly confirm with you that you don't see any problems with that behavior?

shubham1g5: Hey yeah, thanks for flagging. I do think that's fine since this flag is only relevant when there is an update staged.

ctsims:
@shubham1g5 Perfect, all changes look good to me!

PR #2205, opened 2020-04-03
Fix "Choose point on Map" functionality

ctsims:
The feature that lets you see the map while geopoints resolve was broken because it was looking for the old v1 play services element to enable it rather than the new one.

NOTE: This bug is blocking a user workflow on COVID-19 response so I think we should hotfix back into 2.48


shubham1g5:
[Cherry picked in 2.48](https://github.com/dimagi/commcare-android/commit/de313d3cc3815e55e22078950e468b41f80c76de) instead of creating another PR

PR #2208, opened 2020-04-06
GPS Maps: Adds Layer change Button

shubham1g5:
Adds a button to switch between layers on Google Maps

UX - https://drive.google.com/file/d/1VFv39Rn_UkowWpvMzj61s74hf5Qb4QXA/view?usp=sharing

@ctsims  Let me know if the UX doesn't make sense, though that seems to be more of a standard way to implement a very simple layer switching UI.

### DIFF
--- a/app/src/org/commcare/utils/GeoUtils.java
+++ b/app/src/org/commcare/utils/GeoUtils.java
@@ -9,6 +9,9 @@ import android.content.pm.PackageManager;
 import android.location.Location;
 import android.location.LocationManager;
 import android.net.Uri;
+
+import androidx.annotation.IntDef;
+import androidx.annotation.StringDef;
 import androidx.core.content.ContextCompat;
 
 import org.commcare.activities.CommCareActivity;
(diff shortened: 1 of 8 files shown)


shubham1g5:
@ctsims It's ready for review now.

app/src/org/commcare/utils/GeoUtils.java
```diff
@@ -9,6 +9,9 @@
 import android.location.Location;
 import android.location.LocationManager;
 import android.net.Uri;
+
+import androidx.annotation.IntDef;
```
ctsims, line 13: Flagging that imports need to get cleaned up here

app/src/org/commcare/utils/MapLayer.java
```diff
@@ -0,0 +1,19 @@
+package org.commcare.utils;
+
```
ctsims, line 2: Might be worth putting a link to where these come from and where to update them

PR #2220, opened 2020-04-13
Assist Crash reports with device logs

shubham1g5:
Crashlytics provide us a way to add logs to a particular crash report by adding the logs for the particular crash session to the crash report. Adding these device logs to crashlytics will help us in debugging some of the more trickier crashes.


app/src/org/commcare/logging/AndroidLogger.java
```diff
@@ -31,6 +32,7 @@ public AndroidLogger(SqlStorage<AndroidLogEntry> storage) {
     @Override
     public void log(String type, String message, Date logDate) {
         storage.write(new AndroidLogEntry(type, message, logDate));
+        CrashUtil.log(message);
```
ShivamPokhriyal, line 35: 👍

PR #2222, opened 2020-04-20
Fix NPE while building question label with asterisk

ShivamPokhriyal:
If there is a question without label, then `mPrompt.getLongText()` will return null and `builder.append(null);` will cause a NPE.

### DIFF
--- a/app/src/org/commcare/views/widgets/QuestionWidget.java
+++ b/app/src/org/commcare/views/widgets/QuestionWidget.java
@@ -388,6 +388,8 @@ public abstract class QuestionWidget extends LinearLayout implements QuestionExt
         SpannableStringBuilder builder = new SpannableStringBuilder();
         if (prompt.getMarkdownText() != null) {
             builder.append(forceMarkdown(prompt.getMarkdownText()));
+        } else if (mPrompt.getLongText() == null) {
+            return null;
         } else {
             builder.append(mPrompt.getLongText());
         }


app/src/org/commcare/views/widgets/QuestionWidget.java
```diff
@@ -388,6 +388,8 @@ private SpannableStringBuilder getTextFromPrompt(FormEntryPrompt prompt) {
         SpannableStringBuilder builder = new SpannableStringBuilder();
         if (prompt.getMarkdownText() != null) {
             builder.append(forceMarkdown(prompt.getMarkdownText()));
+        } else if (mPrompt.getLongText() == null) {
```
ShivamPokhriyal, line 391: In case the question is required, this won't build the `asterisk`, Since [we are hiding](https://github.com/dimagi/commcare-android/blob/539b3afeda46597d33fc7406f895fbd6a16043dd/app/src/org/commcare/views/widgets/QuestionWidget.java#L441) the label if there is no text.

shubham1g5: Thanks for catching this. We can just append an empty string to the builder in this case to solve for the asterisk.

ShivamPokhriyal:
> We can just append an empty string to the builder in this case to solve for the asterisk.

Though it will set the textView with `asterisk`, but the label will still be [hidden](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/views/widgets/QuestionWidget.java#L439). 
So, Shall we show label with just the asterisk in case of required question?

app/src/org/commcare/views/widgets/QuestionWidget.java
ShivamPokhriyal: @shubham1g5 Though it will set the textView with asterisk, but the label will still be hidden.
So, Shall we show label with just the asterisk in case of required question?

shubham1g5: ahh I think it's fine in that case for the whole quesiotn text to be hidden and to not show anything.

shubham1g5:
GTM once the build passes and PR attributes are added.

ShivamPokhriyal:

PR #2223, opened 2020-04-20
Fix unwanted onItemSelected callback in spinner

ShivamPokhriyal:
Jira: https://dimagi-dev.atlassian.net/browse/SAAS-10759

Fix a bug which removes spinner validation error UI when spinner question is the last question of the form.


ShivamPokhriyal:
Yeah it works fine!

PR #2227, opened 2020-04-27
Show PromptAPK activity for version mismatch error

ShivamPokhriyal:
Jira: https://dimagi-dev.atlassian.net/browse/SAAS-10748

Using this change, for version mismatch errors, we'll now launch PromptAPKUpdateActivity. 

Product Note: For Version Mismatch errors, CommCare will now launch a new screen with details about the required version and buttons to launch PlayStore to update the app.

### DIFF
--- a/app/src/org/commcare/activities/CommCareSetupActivity.java
+++ b/app/src/org/commcare/activities/CommCareSetupActivity.java
@@ -760,15 +760,10 @@ public class CommCareSetupActivity extends CommCareActivity<CommCareSetupActivit
     @Override
     public void failBadReqs(String versionRequired, String versionAvailable, boolean majorIsProblem) {
         String versionMismatch = Localization.get("install.version.mismatch", new String[]{versionRequired, versionAvailable});
-
-        String error;
-        if (majorIsProblem) {
-            error = Localization.get("install.major.mismatch");
-        } else {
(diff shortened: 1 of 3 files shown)


ShivamPokhriyal:

ShivamPokhriyal:

shubham1g5:
@ShivamPokhriyal  Some late feedback. Wodndering in case an in-app update is available, will user also get shown the in app update dialogue on top of the prompt ? If so, the update prompt should not be forced in this case since user might not be in a proper network area and an immediate update will block him for long without allowing him to work on the older app version.

ShivamPokhriyal:
Yeah it makes sense if they're installing app locally. Will make the
changes.

ShivamPokhriyal:
> @ShivamPokhriyal Some late feedback. Wodndering in case an in-app update is available, will user also get shown the in app update dialogue on top of the prompt ? If so, the update prompt should not be forced in this case since user might not be in a proper network area and an immediate update will block him for long without allowing him to work on the older app version.

On second thought, users can always kill the app and then get back to what they wanna do after opening it again, so I think forced update will work.

shubham1g5:
I still think it's better that we don't rely on user killing the app as a mechanism to get out of a certain workflow.

PR #2232, opened 2020-05-06
Fix position of context menu for android 7 devices

ShivamPokhriyal:
https://issuetracker.google.com/issues/37118658
Android had a bug in v7, If you have a listview with lots of items, it will position the context menu at the bottom of the screen(allowing only couple items in the menu with a scroll bar to get to the other items) when user long presses **the topmost list item**. For items further down the list, the context menu starts to show all the items in it. 

WhatsApp-Video-2020-05-06-at-853


app/res/values-v14/themes.xml
```diff
@@ -4,36 +4,10 @@
         Base application theme for API 14+. This theme completely replaces
         AppBaseTheme from res/values/styles.xml on API 14+ devices.
     -->
-    <style name="AppBaseTheme" parent="android:Theme.Holo.Light">
+    <style name="AppBaseTheme" parent="CommonTheme">
         <!-- API 14 theme customizations can go here. -->
-        <item name="android:colorBackground">@color/cc_neutral_bg</item>
```
shubham1g5, line 9: confirming these have been removed since they are same as v-11 values ?

shubham1g5: We can drop v-11 now actually insted of moving these to v-11 since min supported version is 14.

ShivamPokhriyal: or maybe we can use the v-14 as the default one.

shubham1g5: Not sure if I unserstand the difference correctly though won't v-14 will become the default one by itself once we drop v11 ?

ShivamPokhriyal: We have a different `AppBaseTheme` style in the current default `themes.xml` which would never be used, considering we're only supporting v14+ devices. 
So, we can just replace the default `themes.xml` contents with the contents of themes.xml(v-14) instead of maintaining 2 different themes, right?

shubham1g5: yup, got it. That does seem like the right behaviour to me. Would definitely love to get rid of the extra theme files.

PR #2238, opened 2020-05-19
Use a separate form index to loop over all questions while saving form

ShivamPokhriyal:
The earlier code used to modify the currentFormIndex to loop over all the questions. But if at the same time, some other thread also wants to get hold of the current question in screen, it'll have to use the currentFormIndex which might give wrong value, if this thread has already modified it. And hence may produce NPE in worst cases.

Since, we've to loop over all the questions here, we don't really need to modify the currentFormIndex rather we can simply have another form index to get all the questions. And this will also make sure that the other threads who might be using the currentFormIndex at the same time don't get wrong values.


shubham1g5:

PR #2239, opened 2020-05-19
Permit tls 1.2 on Android 19

ctsims:
For older versions of Android, allow force-enabling TLS 1.2 support

cross-request: https://github.com/dimagi/commcare-core/pull/896

### DIFF
--- a/app/src/org/commcare/CommCareApplication.java
+++ b/app/src/org/commcare/CommCareApplication.java
@@ -98,6 +98,7 @@ import org.commcare.utils.PendingCalcs;
 import org.commcare.utils.SessionActivityRegistration;
 import org.commcare.utils.SessionStateUninitException;
 import org.commcare.utils.SessionUnavailableException;
+import org.conscrypt.Conscrypt;
 import org.javarosa.core.model.User;
 import org.javarosa.core.reference.ReferenceManager;
 import org.javarosa.core.reference.RootTranslator;
@@ -108,6 +109,7 @@ import org.javarosa.core.util.PropertyUtils;
 import org.javarosa.core.util.externalizable.PrototypeFactory;
 
 import java.io.File;
+import java.security.Security;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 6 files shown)


app/src/org/commcare/CommCareApplication.java
```diff
@@ -234,6 +238,12 @@ public void onCreate() {
         LocalePreferences.saveDeviceLocale(Locale.getDefault());
     }
 
+    private void initTls12IfNeeded() {
+        if (Build.VERSION.SDK_INT >= 16 && Build.VERSION.SDK_INT < 22) {
```
shubham1g5, line 242: Should not it be [less than 20](https://developer.android.com/reference/javax/net/ssl/SSLSocket.html#protocols) ?

ctsims: The Stripe team that I got the code from identified that there were some 20/21 API devices which still didn't have the update. I think from the data that our audience of devices which are off KitKat but not on a significantly newer version is pretty small, so I figured I'd take their word for it. 

I should add links to the sources, though, which will clarify where that info came from, thanks for catching.

app/src/org/commcare/CommCareApplication.java
```diff
@@ -240,8 +240,12 @@ public void onCreate() {
         LocalePreferences.saveDeviceLocale(Locale.getDefault());
     }
 
+    public static boolean isRoboUnitTest() {
+        return "robolectric".equals(Build.FINGERPRINT);
+    }
+
     private void initTls12IfNeeded() {
-        if (Build.VERSION.SDK_INT >= 16 && Build.VERSION.SDK_INT < 20) {
+        if (Build.VERSION.SDK_INT >= 16 && Build.VERSION.SDK_INT < 20 && !isRoboUnitTest()) {
```
shubham1g5, line 248: Wondering if we can override this function in `CommCareTestApplication` instead ?

ctsims: Ah, clever idea. I was racing to get this in on the hotfix, but that makes **way** more sense. I'll make this change in the long-term version.

shubham1g5:

PR #2240, opened 2020-05-19
allow tls 1.2 on Android 19

ctsims:


ctsims:

shubham1g5:
@ctsims Does this change affect ICDS as well ? Wondering if we need a LTS hotfix as well.

ctsims:
@shubham1g5 Not yet, we had to update the gateways on the prod cloud for some specific scans. Danny and I are working out what the long term configuration will need to be. thankfully ICDS is ahead of the game, because the mobile endpoints have a distinct DNS entry, on prod it's all the same, which is why we couldn't fork the accepted Cipher set / TLS protocol

PR #2241, opened 2020-05-20
Bumps max update attempts to 10

shubham1g5:
looking at update resets [events](https://console.firebase.google.com/project/commcare-a57e4/analytics/app/android:org.commcare.dalvik/events/~2Freport~2Fcommon_commcare_event%3Ft%3D1589964381694&fpn%3D768065119425&swu%3D1&sgu%3D1&sus%3Dupgraded&params%3D_u..pageSize%253D25%2526_u.dateOption%253Dlast30Days%2526_r..eventId%253Dcommon_commcare_event&cs%3Dapp.m.events.detail&r%3Ddefault&g%3D1), it seems like in last 30 days around 81k users (who updated to 2.48) encountered 3,859,016 update resets due to maxing out the current limit of 5 attempts per update trial.

### DIFF
--- a/app/src/org/commcare/logging/analytics/UpdateStats.java
+++ b/app/src/org/commcare/logging/analytics/UpdateStats.java
@@ -21,7 +21,7 @@ public class UpdateStats implements InstallStatsLogger, Serializable {
     private static final String TOP_LEVEL_STATS_KEY = "top-level-update-exceptions";
     private static final String UPGRADE_STATS_KEY = "upgrade_table_stats";
     private static final long TWO_WEEKS_IN_MS = 1000 * 60 * 60 * 24 * 24;
-    private static final int ATTEMPTS_UNTIL_UPDATE_STALE = 5;
+    private static final int ATTEMPTS_UNTIL_UPDATE_STALE = 10;
 
     private final Hashtable<String, InstallAttempts<String>> resourceInstallStats;
     private long startInstallTime;


ctsims:
Hey @shubham1g5 - I agree we should make a change, but I'm wondering whether 10 is even nearly enough.

Maybe what I'm missing is what qualifies as a "failure," and maybe we need to be more lenient with it to let data trickle down. If nothing else, it might make sense for us to never treat a 429 or 503 as a failure, but generally it might be the case that we should be falling back on either the 2 week time period as the restart trigger, or on a much, much higher number (like 50 or 100), since they're clearly trying hundreds of times per user already.

Side-point: I'd be willing to push a change to this as an LTS hotfix if you think it would help you experiment with how to improve updates for CAS - that's a tremendous amount of update cycling that I think we should be addressing proactively. Great job on setting up the analytics to give us the insight!

shubham1g5:
@ctsims Right now we do count any kind of update failure as an attempt for this calculation. I do like your suggestion of not counting rate limiting as a failure here, think it would be good to extend it to any kind of network errors as these errors don't imply an issue with the update.

ctsims:
Yeah, it's a bit of a trade-off in my mind because the point of failing "at some point" was definitely intended to restart the process regardless of what the reason was (to give us a safe fallback so phones were never stuck in 'waiting to update' mode permanently), and that has provided value historically, but I think it's now being very clearly outweighed by the determent of how infrequently users can hit the endpoints.

If 503's count as failures, I would 100% support pushing a hotfix which prevents that from being the case, although my guess based on seeing the proxy lately is that as many people are receiving other unrelated network failures as 503's. We always have the 2 week timeout to fall back on, I'm wondering whether we should set a switch somewhere for LTS (or using some other proxy indicator that is set for CAS) to just disable "Network issue, or 503, or 429" as a reason to fail an update entirely.

shubham1g5:
I am a bit worried about setting up another custom indicator as that will need an app change and a release while using an existent indicator (other than comparing using hardcoded domain names) would be a bit hacky as that's originally meant to control some other function. Also I think we definitely want to ignore 429s and 503s irrespective of the indicator. Therefore it might just be enough to do these couple changes - 

1. Don't count 503/429 as failures
2. Increase the update attempt threshold to 50 to counter for network failures

Going to initiate the hotfix conversation separately over email

ctsims:
@shubham1g5 Good point. Agreed that having 5 switches for different behaviors isn't getting us a lot of value. 

I think what I've been considering is whether updating to 50 actually gets us anything v. just killing the retry counter entirely. The updates reset either way after 14 days, so if they are only auto-retrying once per day, setting it to 50 is equivalent to removing the failed attempt counter.

My concern is that the reason we have the existing retry counter is that otherwise a 'bad' update that has an error in the config will block the phone for two weeks.

one counter-proposal would be
- Leave the retry counter set to 5
- Only increment the failed attempt counter for errors which are NOT network related in any way

The only question is whether we believe can reliably catch the second condition, and historically we've had challenges associating network errors of different types with the network.

If we can't meet that goal, then ignoring the reset counter entirely seems fine to try (especially for ICDS), just noting that setting it to 50 is essentially the same thing.

shubham1g5:
@ctsims Agree that 50 seems equivalent to ignoring the counter. Based on Analytics data (73k users retrying 2.2 million times ) for last couple weeks, the retry average seems to be ~30 every couple weeks which further validates this argument. 

Ignoring network errors while preserving the counter does seem like a better option to me. On Identifying network errors properly, `UnreliableSourceException` seems to be set up correctly to separate out these errors though unfortunately it gets handled as a [general URE](https://github.com/dimagi/commcare-android/blob/commcare_2.48/app/src/org/commcare/tasks/UpdateTask.java#L140) today and loose it's meaning in the end result. It's an easy change to explicity capture it and tag it as a specific result to make a distinction for retry counter.

shubham1g5:
> Agree that 50 seems equivalent to ignoring the counter. Based on Analytics data (73k users retrying 2.2 million times ) for last couple weeks, the retry average seems to be ~30 every couple weeks which further validates this argument.

Ahh, made a mistake here. 73k users getting reset 2.2 million times => 2.2 * 5 million retries => ~150 retries per user in 2 weeks. so 50 doesn't seem equivalent to disabling the counter. Though still think not considering rate limiting plus network errors is way to go here.

shubham1g5:
Pr'ed a hotfix here - https://github.com/dimagi/commcare-android/pull/2244/files based on this conversation.

PR #2243, opened 2020-05-21
Adds lazy donwnload for multimedia for App Updates

shubham1g5:
Spec: https://docs.google.com/document/d/1WWhgG6cyKm7nN2pUn8lty72F3ZN6D1xWqTZsIlHX4z8/edit#heading=h.xlb7kbytn7rw

- Parses a new 'lazy' atrribute for `resource definition` 
- Skips downloading the lazy media resource during install 
- Initiate a Work manager task `MissingMediaDownloadWorker` on login to initiate lazy media download. This task today takes help of our existing multimedia verification to find out the lazy resources and installs them one by one in background
- [Changes Media Layouts](https://drive.google.com/file/d/1Q5rRCKB3JAxjCXV_ug3_oQrdwIFzDTlr/view?usp=sharing) to support requesting for missing media on user click

cross-request: https://github.com/dimagi/commcare-core/pull/899


shubham1g5:

app/src/org/commcare/engine/resource/AndroidResourceUtils.kt
```diff
@@ -0,0 +1,44 @@
...
+import java.io.File
+import java.io.IOException
+import java.util.*
+
+object AndroidResourceUtils {
+
+    // loops over all lazy resources and checks if one of them has the same local path as {@param problem} URI
+    @JvmStatic
+    fun ifUriBelongsToALazyResource(problem: MissingMediaException, lazyResources: Vector<Resource>): Boolean {
+        for (lazyResource in lazyResources) {
```
ctsims, line 17: Not sure where this list is populated from, but heads up that iterating over resources is a really expensive operation on really large apps like CAS. This might need to get cached into a table which is indexed by the file URI

app/src/org/commcare/engine/resource/AndroidResourceUtils.kt
```diff
@@ -0,0 +1,44 @@
...
+        if (resource.installer is MediaFileAndroidInstaller) {
+            try {
+                val resourceUri = (resource.installer as MediaFileAndroidInstaller).localLocation
+                val resourcePath = ReferenceManager.instance().DeriveReference(resourceUri).localURI
+                val problemPath = ReferenceManager.instance().DeriveReference(uri).localURI
+                if (File(resourcePath).canonicalPath.contentEquals(File(problemPath).canonicalPath)) {
+                    return true
+                }
+            } catch (e: InvalidReferenceException) {
+                e.printStackTrace()
```
ctsims, line 37: I think it's really risky to swallow these. IRE's are almost always the result of a bug, and the IOException is absorbing a pretty large surface area

app/src/org/commcare/mediadownload/MissingMediaDownloadHelper.kt
```diff
@@ -0,0 +1,196 @@
...
+
+    // Verifies all application media and re-downloads any missing lazy resources
+    fun downloadAllMissingMedia() {
+        val platform = CommCareApplication.instance().commCarePlatform
+        val global = platform.globalResourceTable
+
+        global.setInstallCancellationChecker(this)
+
+        val problems = Vector<MissingMediaException>()
+        global.verifyInstallation(problems, platform)
```
ctsims, line 80: Have you timed this operation with CAS, out of curiosity?

app/src/org/commcare/mediadownload/MissingMediaDownloadHelper.kt
```diff
@@ -0,0 +1,196 @@
...
+                        missingMediaDownloadListener.onComplete(result)
+                    }
+                }
+        )
+    }
+
+    private fun downloadMissingMediaResource(uri: String): MissingMediaDownloadResult {
+        val platform = CommCareApplication.instance().commCarePlatform
+        val global = platform.globalResourceTable
+        val lazyResources: Vector<Resource> = global.lazyResources!!
```
ctsims, line 136: Same comment as above that I think this is likely to be too expensive to be practical if this list is as big as CAS's

app/src/org/commcare/mediadownload/MissingMediaDownloadHelper.kt
```diff
@@ -0,0 +1,196 @@
...
+import java.util.concurrent.TimeUnit
+import kotlin.collections.ArrayList
+
+// Contains helper functions to download lazy or missing media resources
+object MissingMediaDownloadHelper : TableStateListener, InstallCancelled {
+
+    var resourceInProgress: Resource? = null
+        private set
+
+    private val jobs = ArrayList<Job>()
```
ctsims, line 30: Shouldn't there be a lock of some sort around each one of these jobs, or is that inherent to the job class?

app/src/org/commcare/mediadownload/MissingMediaDownloadHelper.kt
shubham1g5: I have seen it varying from 14 to 30 seconds on lower mid range to mid range devices but can imagine it taking more closer to a minute on really low end devices.

app/src/org/commcare/mediadownload/MissingMediaDownloadHelper.kt
shubham1g5: this particular call is quite cheap (takes .1 sec on MotoG, around .4 seconds on a low performant device I have) as we are [indexing the lazy property](https://github.com/dimagi/commcare-core/pull/899/files#diff-4ebf66124223ceb5a85393ee19f18ba5R335) in the Resource Table.

app/src/org/commcare/mediadownload/MissingMediaDownloadHelper.kt
shubham1g5: each `job` in `jobs` depends on the synchronized  `recoverResource` function in this class which takes care of protecting against spawning multiple download requests for a resource. The `jobs` itself is only meant to provide a way for cancelling all jobs at once.  So it's fine for it to not be thread safe and have duplicate job entries as a result.

app/src/org/commcare/engine/resource/AndroidResourceUtils.kt
shubham1g5: Agree

shubham1g5:
@ctsims My approach to caching was to provide the index on Resource table for the `lazy` property but I see that that's not going to solve the purpose of indexing by the file URI. Listing out a few ways to do that - 

1. Considering all of these processes happen in background (asynchronously), Provide a in memory cache that gets initialized the first time someone access lazy resources after startup. 

2. Add another DB table to maintain uri to lazy resource index as you have been suggesting. 

3. Add a location index on Resource tables directly. Putting it as an option here since right now the downloading is limited to a lazy resource only though I do think that we should really be downloading any missing resources we identify in background and not just the lazy resources. If we wanna do so , having a in memory table like #1 or a sql table #2  just for the lazy resources will be a limiting factor. 

4. If we don't wanna mess up with ResourceTable, have a in memory version of #3 where we maintain a cache with an index over all missing resources.

shubham1g5:
@ctsims Made the discussed changes in last few commits.

app/src/org/commcare/preferences/HiddenPreferences.java
```diff
@@ -511,4 +513,23 @@ public static void disableRateLimitPopup(boolean disable) {
                 .putBoolean(DISABLE_RATE_LIMIT_POPUP, disable)
                 .apply();
     }
+
+    public static void updateLazyMediaDownloadResult() {
+        String prefKey = String.format(LAZY_MEDIA_DOWNLOAD_COMPLETE, AppUtils.getCurrentAppVersion());
```
ctsims, line 512: I think it's dangerous that this is implicit.

I think we we should either receive this argument as a parameter, or we should explicitly state in method name what the scope it returns over (IE "ForCurrentlySeatedApp")

app/src/org/commcare/mediadownload/MissingMediaDownloadHelper.kt
```diff
@@ -72,29 +72,32 @@ object MissingMediaDownloadHelper : TableStateListener, InstallCancelled {
...
-                            .takeWhile { !wasInstallCancelled() }
-                            .map { global.getResource(it.value) }
-                            .filter { isResourceMissing(it) }
-                            .takeWhile { !wasInstallCancelled() }
-                            .onEach { recoverResource(platform, it) }.toList()
-                }.onFailure {
-                    Logger.log(LogTypes.TYPE_MAINTENANCE, "An error occured while lazy downloading a media resource : " + it.message)
-                    return handleRecoverResourceFailure(it).data
-                }
+        if (!HiddenPreferences.hasLazyMediaDownloaded()) {
```
ctsims, line 73: I'm not sure why we'd ever been running lazy media installers for multiple app versions. Are we lazy installing the upgrade table as well as the installed version? or is this due to some behavior that happens after handing over between updates being applied??

app/src/org/commcare/mediadownload/MissingMediaDownloadWorker.kt
```diff
@@ -3,21 +3,16 @@ package org.commcare.mediadownload
...
-        kotlin.runCatching {
-            MissingMediaDownloadHelper.downloadAllLazyMedia()
-        }.onSuccess {
-            return Result.success()
-        }.onFailure {
-            return Result.failure()
-        }
-        return Result.failure()
+        val result = MissingMediaDownloadHelper.downloadAllLazyMedia()
+        return if (result == AppInstallStatus.Installed) Result.success() else Result.retry()
```
ctsims, line 15: Don't we only want to retry automatically on specific errors?

In the [handling list above](https://github.com/dimagi/commcare-android/pull/2243/commits/a5fb86158cb77cb120d00ae8a1e34026b9987108#diff-15b34a5a58efc4e31c602ee832bd632eR103), 3 of the 4 errors wouldn't be good for us to run a retry on

app/src/org/commcare/mediadownload/MissingMediaDownloadHelper.kt
shubham1g5: A new app version can introduce new lazy resources, so once we have successfully run this task for version x, we don't need to run it again for the same version until user updates to a different version. 

Retrospectively, a better change here will be to just maintain a single app pref across all versions and reset that flag after an app update.

ctsims: Ah, got it. I was reviewing by commit, which was silly, and I missed this was old. Sorry for the noise.

shubham1g5:
@ctsims  thanks. Also want to make sure that you have had a chance to go through the cross-requested core PR.

shubham1g5:
Made one more functional change to only show the audio missing toast when the audio button is visible.

shubham1g5:
@ctsims  I just realized that we don't actually rate limit media requests but only the profile files which is the starting point for an update. Therefore I am worried that these lazy mm requests won't get rate limited at all slowing down other parts of the system. Is it worth adding a flag to these request saying that these are not part of a blocking update and hence it's fine to rate limit these.

ctsims:
@shubham1g5 I think that's a really good idea, and we can avoid setting that flag for Media that's actually on the screen when the user makes the direct request to download a file

PR #2244, opened 2020-05-24
[Hotfix] Update resets

shubham1g5:
* Segregrates update failures to selectively increment update reset counter 
* corrects a bug where `InvalidResourceException` was incorrectly getting logged as `AppInstallStatus.UnknownFailure`

Release Note: Tunes update bheaviour to be more intelligent in resuming update downloads from failures

### DIFF
--- a/app/src/org/commcare/logging/analytics/UpdateStats.java
+++ b/app/src/org/commcare/logging/analytics/UpdateStats.java
@@ -1,6 +1,7 @@
 package org.commcare.logging.analytics;
 
 import org.commcare.CommCareApp;
+import org.commcare.engine.resource.AppInstallStatus;
 import org.commcare.resources.model.InstallStatsLogger;
 
 import java.io.PrintWriter;
@@ -24,6 +25,7 @@ public class UpdateStats implements InstallStatsLogger, Serializable {
     private final Hashtable<String, InstallAttempts<String>> resourceInstallStats;
     private long startInstallTime;
     private int restartCount = 0;
+    private int resetCounter = 0;
 
     private UpdateStats() {
         startInstallTime = new Date().getTime();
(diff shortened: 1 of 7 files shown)


shubham1g5:
Build will fail as I have not opened a core branch correspondign to 2.48

app/src/org/commcare/logging/analytics/UpdateStats.java
```diff
@@ -62,6 +64,7 @@ public void resetStats(CommCareApp app) {
         startInstallTime = new Date().getTime();
         resourceInstallStats.clear();
         restartCount = 0;
+        resetCounter = 0;
```
ctsims, line 67: What's the logic behind splitting both of these out rather than just changing what hits restartCount?

shubham1g5: the restart counter gets used in determining the [retry back off time](https://github.com/dimagi/commcare-android/blob/commcare_2.48/app/src/org/commcare/engine/resource/AndroidResourceManager.java#L250), this code has been removed in `master` since the retry time is managed by the `WorkManager` and hence I will just maintain one counter when I port it in `master`

shubham1g5:
> We are only rolling out as a hotfix for LTS, right? Not prod? (since this is attempting to help with updates on the rate limited requests
Yes, that's right.

> Do you happen to be clear on what happens when the ICDS bandwidth is saturated in terms of the mobile failure presentation? I'd expect it to be a hard fault in the network layer that hopefully presents in ResourceInstallUtils line 131, but wanted to give you a warning ahead of time that if we fail to catch a specific condition that it'll be there.
yeah , all network timeouts does get captured as `UnreliableSourceException` so it should get caught as a network error. 

On a side note, I was thinking of segrregrating these bandwidth errors in anlaytics though based on the slack chat it seems impossible to do so.

PR #2248, opened 2020-05-26
Improvements to Update Error handling

shubham1g5:
Port of https://github.com/dimagi/commcare-android/pull/2244 along with additional error handling -
* Tag parse errors as InvalidResourceException instead of URE
* indentify errors writing incoming resource to a file and throw a `LocalStorageUnavailableException`

cross-request: https://github.com/dimagi/commcare-core/pull/904

### DIFF
--- a/app/src/org/commcare/android/resource/installers/ProfileAndroidInstaller.java
+++ b/app/src/org/commcare/android/resource/installers/ProfileAndroidInstaller.java
@@ -119,7 +119,7 @@ public class ProfileAndroidInstaller extends FileSystemInstaller {
         } catch (XmlPullParserException | InvalidReferenceException | IOException e) {
             e.printStackTrace();
         } catch (InvalidStructureException e) {
-            throw new UnresolvedResourceException(r, "Invalid content in the Profile Definition: " + e.getMessage(), true);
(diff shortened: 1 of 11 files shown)


app/src/org/commcare/android/resource/installers/ProfileAndroidInstaller.java
```diff
@@ -119,7 +119,7 @@ public boolean install(Resource r, ResourceLocation location, Reference ref,
         } catch (XmlPullParserException | InvalidReferenceException | IOException e) {
             e.printStackTrace();
         } catch (InvalidStructureException e) {
-            throw new UnresolvedResourceException(r, "Invalid content in the Profile Definition: " + e.getMessage(), true);
```
ctsims, line 122: Hey @shubham1g5 I believe this is the result of the fact that the profile file (being the first entry point for the app) can sometimes not point to a profile file at all due to a captive portal or similar. I'm not sure what the downstream implications of changing this would be, there, but wanted to surface it in case it jogged any memories

shubham1g5: I see, still makes sense to me to classify it as `InvalidResourceException`. User is going to see the same error message in both cases and there is no difference in downstream handeling of `UnresolvedResourceException` vs `InvalidResourceException`. The change just helps in better classifying the errors for Analytics purpose.

PR #2253, opened 2020-05-30
Updates to AGP 4.0, Gradle 6

shubham1g5:


shubham1g5:
@ShivamPokhriyal Would be good to verify that it doesn't break the setup for you, also heads up best to try it when you are not actively working as you will need to update both your AS and Gradle which can take nearly a GB of internet.

gradle.properties
```diff
@@ -17,8 +17,6 @@ org.gradle.parallel=true
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:configuration_on_demand
 org.gradle.configureondemand=false
 
-android.enableUnitTestBinaryResources=true
```
shubham1g5, line 20: deprecated

ShivamPokhriyal:
Additional Note(can probably be adjusted in this PR)
I get 
```
WARNING: API 'variant.getMergeResources()' is obsolete and has been replaced with 'variant.getMergeResourcesProvider()'.
It will be removed at the end of 2019.
```
 in gradle sync(not sure if you're still getting the same) and it seems like it's an issue with [google-services 4.3.1](https://firebase.google.com/support/release-notes/android#google-services_plugin_v4-3-2) and can be fixed by upgrading to the latest version(4.3.3 though I cannot see any release notes for it😅).

shubham1g5:
Thanks for flagging @ShivamPokhriyal . pushed a bunch of lib updates along with `google-services`.

PR #2257, opened 2020-06-01
Use current form index to check constraints

ShivamPokhriyal:
cross-request: https://github.com/dimagi/commcare-core/pull/908
A fix for this issue: https://dimagi-dev.atlassian.net/browse/QA-1354 caused by [this PR](https://github.com/dimagi/commcare-android/pull/2238). We've changed the SaveToDisk task to use a separate form index to loop over all the questions while saving them. But one of the internal [method call](https://github.com/dimagi/commcare-core/pull/908/files#diff-df805588fd3f80aa4ffb8036e891165aL171) was still using the static formIndex variable in FormModel which was out of bounds causing an NPE.


shubham1g5:
@ShivamPokhriyal can you add more info in description of this PR  around the incorrect functional behaviour that this PR fixes so that I can contextualize it better.

ShivamPokhriyal:
> @ShivamPokhriyal can you add more info in description of this PR around the incorrect functional behaviour that this PR fixes so that I can contextualize it better.

Updated the description!

PR #2260, opened 2020-06-03
Default to current version while getting latestcommcareversion

ShivamPokhriyal:
A last minute fix for https://dimagi-dev.atlassian.net/browse/QA-1359

### DIFF
--- a/app/src/org/commcare/preferences/HiddenPreferences.java
+++ b/app/src/org/commcare/preferences/HiddenPreferences.java
@@ -5,6 +5,7 @@ import android.content.SharedPreferences;
 import org.commcare.CommCareApp;
 import org.commcare.CommCareApplication;
 import org.commcare.activities.GeoPointActivity;
+import org.commcare.android.logging.ReportingUtils;
 import org.commcare.utils.AndroidCommCarePlatform;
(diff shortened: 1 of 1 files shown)


shubham1g5:
@ShivamPokhriyal Wondering if we should also set this latest version info pref from the result obtained from the In app update APIs ?

ShivamPokhriyal:
> @ShivamPokhriyal Wondering if we should also set this latest version info pref from the result obtained from the In app update APIs ?

Unfortunately using the [AppUpdateInfo](https://developer.android.com/reference/com/google/android/play/core/appupdate/AppUpdateInfo) we can only fetch the versionCode. So it's probably not possible to set the versionName using in-app-update apis.

ShivamPokhriyal:
@shubham1g5 On second thought, Do you think it'd be a good idea to set the prefs from [here](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/tasks/ResourceEngineTask.java#L118)

shubham1g5:
@ShivamPokhriyal I see . The required version is not always the latest version, an app can set that to whatever they want from app settings. So no other changes are required based on your comment.

PR #2261, opened 2020-06-03
Set latest CC version from Heartbeat response

shubham1g5:
This was only getting set currently from the Recovery measures response which is not active for most of the projects. Since we are now using this version info for the udpate prompt's In App Update feature, setting it from the heartbeat response as well so that we have this info set before we prompt for update.

### DIFF
--- a/app/src/org/commcare/heartbeat/HeartbeatRequester.java
+++ b/app/src/org/commcare/heartbeat/HeartbeatRequester.java
@@ -156,6 +156,7 @@ public class HeartbeatRequester extends GetAndParseActor {
                     if (latestVersionInfo.has("force")) {
                         forceString = latestVersionInfo.getString("force");
                     }
+                    HiddenPreferences.setLatestCommcareVersion(versionValue);
                     UpdateToPrompt updateToPrompt = new UpdateToPrompt(versionValue, forceString, updateType);
                     updateToPrompt.registerWithSystem();
                 }


app/src/org/commcare/heartbeat/HeartbeatRequester.java
```diff
@@ -156,6 +156,7 @@ private static void parseUpdateToPrompt(JSONObject latestVersionInfo, UpdateToPr
                     if (latestVersionInfo.has("force")) {
                         forceString = latestVersionInfo.getString("force");
                     }
+                    HiddenPreferences.setLatestCommcareVersion(versionValue);
```
ShivamPokhriyal, line 159: @shubham1g5 Just noticed one thing. This particular method `parseUpdateToPrompt` is being called from both `attemptCczUpdateParse` and `attemptApkUpdateParse`. 
But for both of them we're using `setLatestCommcareVersion` when we should've used `setLatestAppVersion`. 
I'm not sure about the actual impact of this bug, but seems like we should fix this in this release. Can you please check?

shubham1g5: ahh good catch. We should wrap this setter in a `updateType` check. The impact would not have been noticeable as no one is using the recovery feature it's used in currently.  Also can we also set this latest version info from the playstore update API for in-app updates ?

ShivamPokhriyal: > can we also set this latest version info from the playstore update API for in-app updates?

Yeah, I think we can get the `versionCode` from the playstore api https://developer.android.com/reference/com/google/android/play/core/appupdate/AppUpdateInfo#availableVersionCode(), but not the version name.

shubham1g5: ok we do need the version name here to be compatible with the surrounding code.

shubham1g5: https://github.com/dimagi/commcare-android/pull/2509

PR #2269, opened 2020-06-11
Commcare 2.49 release

ShivamPokhriyal:
cross-request: https://github.com/dimagi/commcare-core/pull/913


app/AndroidManifest.xml
```diff
@@ -3,7 +3,7 @@
           xmlns:tools="http://schemas.android.com/tools"
           package="org.commcare.dalvik"
           android:versionCode="106"
-          android:versionName="2.49">
+          android:versionName="2.49.0">
```
shubham1g5, line 6: should be 2.50

app/res/values/strings.xml
```diff
@@ -59,7 +59,7 @@
 
     <integer-array name="commcare_version">
         <item>2</item>
-        <item>48</item>
+        <item>49</item>
```
shubham1g5, line 62: 50

app/AndroidManifest.xml
```diff
@@ -3,7 +3,7 @@
           xmlns:tools="http://schemas.android.com/tools"
           package="org.commcare.dalvik"
           android:versionCode="106"
-          android:versionName="2.49">
+          android:versionName="2.50">
```
ShivamPokhriyal, line 6: Okay, I haven't seen this before. I updated `commcare_2.49` branch to have 2 files `AndroidManifest` and `strings.xml` same as master. And it did updated those files as well. 
But now this PR somehow is showing a difference when the two files are identical in these branches.

shubham1g5: @ShivamPokhriyal  you just need to merge master back in to resolve this. I just did it and it's no longer showing up now.

ShivamPokhriyal: Apparently there was some weird cache which wouldn't update even after refreshing the page. Glad that it's working as expected.

ShivamPokhriyal: Wow, it was your commit that did the trick here. But I'm still not wondering why would these 2 files show up in the PR when both the branches had same content in those files.

ShivamPokhriyal: I guess [this answer](https://stackoverflow.com/a/26986320/6671572) might be the reason behind this. But I've no way to confirm it so I guess I'm left wondering why it ever happened.

shubham1g5: possibly because the 2 commits making the change had different hashes ?

ShivamPokhriyal: Maybe yeah, But then the `Files changed` section of PR usually shows the final changes in the files of 2 branches though we can see `Files changed` for a particular commit but that's different.

Aah, maybe that answer is right, `commcare_2.49` had no idea of the new changes to master since it diverted. But then I've seen PRs flagging conflicts(when I'm changing the part of file that somebody else also changed and merged) so it should've raised a merge conflict in the first place right?

ShivamPokhriyal: Now when I go back to [this commit 38d7b79](https://github.com/dimagi/commcare-android/pull/2269/commits/38d7b79b9f1d2318bd1014c36ffaa4cafe2fc9d6) and do a `git rebase master` it shows up a bunch of conflicts. So perhaps that answer is right. 
The only thing that I'm still wondering is how github shows merge conflicts in PRs when it has no idea of the new changes to the target branch.

PR #2275, opened 2020-06-15
[2.49.1 hotfix] Remove install stats exception logging from update stats

shubham1g5:
Fix for [Crash](https://console.firebase.google.com/u/1/project/commcare-a57e4/crashlytics/app/android:org.commcare.lts/issues/89eebf691ef96001ffe22811b8033bf6?time=last-thirty-days)

These `resourceInstallStats` logs  are not getting used anywhere and adds unneccesary failure points in the code today. This code was added 5 years back as part of auto updates feature and if we have not used it since then, hightly likely that we are not going to use it in future as well. 

cross-request: https://github.com/dimagi/commcare-core/pull/915


ctsims:
Hi @shubham1g5 -  removing the objects from the serialization in this way will break deserializing existing objects, won't it? Did you test what happens when this code runs for the first time with objects which have already filled in these tables and it tries to deserializing the new object without knowing about the tables?

I was assuming we'd have to be a bit more surgical - Removing all of the code that ADDS to the tables, but not removing the tables themselves.

shubham1g5:
@ctsims  that's right, it does break the deserialization upon which the `updateStats` [get reset](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/logging/analytics/PrefStats.java#L84). I think resetting these stats once don't have a significant downside in longer term. So I have deliberately remove all of this code, but can add back the code required to not break the serialization if that's a necessity.

PR #2277, opened 2020-06-16
Remove kujaku network change receiver

ShivamPokhriyal:
[Crash Fix](https://console.firebase.google.com/u/1/project/commcare-a57e4/crashlytics/app/android:org.commcare.dalvik/issues/12e997a8b50eb71f33ec03df4925332a)
This receiver `KujakuNetworkChangeReceiver` starts  `MapboxOfflineDownloaderService` which is giving a lot of crashes. So removing it from here.

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -459,5 +459,8 @@
             android:name="firebase_crashlytics_collection_enabled"
             android:value="false" />
 
+        <receiver android:name="io.ona.kujaku.receivers.KujakuNetworkChangeReceiver"
+            tools:node="remove"/>
+
     </application>
 </manifest>
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -459,6 +459,11 @@
             android:name="firebase_crashlytics_collection_enabled"
             android:value="false" />
 
+        <!-- With App Bundles, we're seeing a number of MissingLibraryExceptions
+            which is coming from `MapboxOfflineDownloaderService` in Kujaku library.
+            One way to start this service is by using this network change receiver.
+            So removing this entry from the manifest file.
+         -->
         <receiver android:name="io.ona.kujaku.receivers.KujakuNetworkChangeReceiver"
             tools:node="remove"/>
 


ShivamPokhriyal:
> nice, didn't realize we can remove explicit stuff from manifest like this. Can you add a comment in code so that it's obvious why we are doing this.

Actually, at first I thought, this might be enough for stopping the crashes. But then I realised that the same crash is happening on Android 9 device as well. And I think the Connectivity Change receiver needs to be registered via code to be able to work on android 9 devices. 
Kujaku does register it [here conditionally](https://github.com/onaio/kujaku/blob/master/library/src/main/java/io/ona/kujaku/KujakuLibrary.java#L57) which should've never been triggered as we're not setting [this value](https://github.com/onaio/kujaku/blob/d4d76cb59c99d7806aea59c86c1f5bacfcc349a0/library/src/main/java/io/ona/kujaku/KujakuLibrary.java#L87) from our code. So perhaps this is not enough to stop the crashes. 
But yeah I think we should definitely remove this receiver from the manifest.

shubham1g5:
@ShivamPokhriyal  agree that we should remove this as instances of this issue on Android 9 might just be due to the custom android images on those phones still allowing the manifest recievers to get fired.

ShivamPokhriyal:
Cool! I'm gonna merge it after doing some testing.

ShivamPokhriyal:
Amazing! It's really really easy to reproduce in a pre-Nougat device. I used android 6. You just need to install the `aab` into your device(I used `bundletool`) and then after opening commcare, just toggle internet. And boom it'll [crash](https://console.firebase.google.com/u/0/project/commcare-a57e4/crashlytics/app/android:org.commcare.dalvik/issues/12e997a8b50eb71f33ec03df4925332a?time=last-hour&sessionId=5EE8B7C90388000125ADE79CA421AA44_DNE_0_v2).
Though it might be possible that the above repro is because of the alternate theory behind this crash, i.e. a sideload installation. I tried uploading the `aab` to internal track, but internet literally sucks here, it stopped at 84%. 
(PS: It was also mentioned in [one of the comments](https://issuetracker.google.com/issues/109448556#comment17) that bundletool installation is same as playstore so I'm gonna assume that the crash is indeed reproducible.)

ShivamPokhriyal:
@shubham1g5 This fix worked for the above crash. 
But the app is still crashing when we use kujaku. I ran the fix in my device, android 7, and used your [sample app](https://www.commcarehq.org/a/shubhamgoyaltest/apps/view/9dd730562105c120791a5e8a0f51bd40/) and it crashes as soon as I click button to record/walk boundary. [Here is the stacktrace](https://console.firebase.google.com/u/0/project/commcare-a57e4/crashlytics/app/android:org.commcare.dalvik/issues/12e997a8b50eb71f33ec03df4925332a?time=last-hour&sessionId=5EE8BE85039E00012B3B93A18F747DDF_DNE_0_v2)

So perhaps, we should upload `apk` and not `aab`?  But I'm yet to with the apk, so let's see if that'd work or not.

shubham1g5:
@ShivamPokhriyal  thats great. We should push this one out asap with an .aab , I don't think anyone is going to use Kujaku in coming month so fine to keep it crashing for now. I will work with the Kujaku team meanwhile to see if there is a way to fix it.  Thanks a bunch for looking in this.

shubham1g5:
@ShivamPokhriyal  Flagging that you need to add the PR attributes before merging the PRs.

shubham1g5:
@ShivamPokhriyal Wondering if this should be reverted since the actual issue turned out to be in our build ?

ShivamPokhriyal:
Yeah, thanks for flagging!

PR #2279, opened 2020-06-17
Hides manual quarantine behind developer option

shubham1g5:
Couple changes: 
1. Only show manual quarantine based on a advance setting
2. Logs the Quarantine event to Analytics 

https://dimagi-dev.atlassian.net/browse/ICDS-1568

Product Note: Hides the manual form Quarantine option from the scan record integrity dialog


app/src/org/commcare/activities/FormRecordListActivity.java
```diff
@@ -464,10 +464,13 @@ private void createFormRecordScanResultDialog(Pair<Boolean, String> result, Form
...
         if (FormRecordFilter.Pending.containsStatus(record.getStatus())) {
-            dialog.setNegativeButton(Localization.get("app.workflow.forms.quarantine"), (dialog1, which) -> {
-                manuallyQuarantineRecord(record);
-                dismissAlertDialog();
-            });
+            if (DeveloperPreferences.isManualFormQuarantineAllowed()) {
+                dialog.setNegativeButton(Localization.get("app.workflow.forms.quarantine"), (dialog1, which) -> {
+                    manuallyQuarantineRecord(record);
+                    dismissAlertDialog();
+                    DeveloperPreferences.disableManualFormQuarantine();
```
ctsims, line 471: Is there a reason this is "spring loaded" so to speak and only lets you quarantine a single form? That will get quite painful if there is some strange event that causes issues for a lot of forms in a row.

app/src/org/commcare/preferences/DeveloperPreferences.java
```diff
@@ -56,6 +56,7 @@
     private final static String ENFORCE_SECURE_ENDPOINT = "cc-enforce-secure-endpoint";
     private final static String ALLOW_SPACE_IN_SELECT_CHOICES = "cc-allow-space-in-select-choices";
     private final static String USE_EXPRESSION_CACHING_IN_FORMS = "use-expression-caching-in-forms";
+    private final static String ALLOW_MANUAL_FORM_QUARANTINE = "cc-allow-manual-form-quarantine";
```
ctsims, line 59: I think this needs to be behind the user visible / controllable developer preferences, not the "Special Dimagi" level preferences.

Manual quarantining is a support tool that we need non-Dimagi personnel to be able to trigger to recover a clean state on their devices.

shubham1g5:
@ctsims  Both of your concerns are valid, Though the idea behind this change is to make it as hard as possible for a user to manually quarantine a form and shifts to an approach where the CommCare detects most cases of these errors by itself rather than allowing user to control this behaviour. So I would for example prefer if we don't allow manual quarantine at all in favour of us pushing apk updates to cover a failure edge case as a result of which form should have got quarantined in the first place. I am afraid giving this capability to users have more data integrity sideeffects than us providing dedicated support to users to get these issues resolved at user end.

ctsims:
Hey @shubham1g5 

I totally understand where you are coming from and do generally agree that it's always better to failfast and fix the real problem than to smooth it away. I think it's a judgment and maintenance call with no right or wrong answer here.

One big, big reason that we introduced manual quarantine is because users did find a way to smooth things over even without it, which was to completely blow away their app installation and start over. With the advent of Andoid removing a universal Device ID indicator that action is even more invisible to us, and affects the userbase pretty negatively.

I think this is your call to make, so I'll switch over to approve and let you make the final decision, since you're the senior member of the mobile team, and ultimately the good and bad from the decision will end up on your place. One thing I do want to suggest, though, is that it would be useful to know how this is being used outside of CAS, since this decision affects users across the userbase.

shubham1g5:
@ctsims  Makes sense. I am going to loop in support team on this and see if they have inputs on this. Thanks.

shubham1g5:
@ctsims I had a chat with Kishan on this and based on that we decided to move the setting controlling the "manual quarantine" visibility to the advanced settigs screen to avoid exposing rest of the developers option to user when asking them to perform a manual quarantine. Though we are still keeping the setting per form quarantine.

PR #2280, opened 2020-06-18
Adds support for tiered lists and barcode scan in Case Claim prompts

shubham1g5:
https://trello.com/c/mci8bSOn/17-improve-case-claim-parameter-usage
(leaves out 2 pane UI in this PR from the mobiletasks listed in the card)

Screenshot after UI changes - 

cross-request: https://github.com/dimagi/commcare-core/pull/916


app/src/org/commcare/activities/QueryRequestActivity.java
```diff
@@ -116,56 +132,176 @@ private void buildPromptUI() {
...
+            choices[i] = item.getLabelInnerText();
+            if(item.getValue().equals(answer)){
+                selectedPosition = i;
+            }
+        }
+
+        ArrayAdapter<String> adapter = new ArrayAdapter<>(this, android.R.layout.simple_spinner_item, choices);
+        promptSpinner.setAdapter(adapter);
+        if(selectedPosition != -1) {
+            promptSpinner.setSelection(selectedPosition);
```
ctsims, line 208: Does this handle blank entries effectively? It was unclear from the UI and I can't quite tell from the code how easy it is to "undo" entering a value into the spinner at all (which we definitely want to support)

app/src/org/commcare/activities/QueryRequestActivity.java
```diff
@@ -116,56 +132,176 @@ private void buildPromptUI() {
...
+        for (int i = 0; i < items.size(); i++) {
+            SelectChoice item = items.get(i);
+            choices[i] = item.getLabelInnerText();
+            if(item.getValue().equals(answer)){
+                selectedPosition = i;
+            }
+        }
+
+        ArrayAdapter<String> adapter = new ArrayAdapter<>(this, android.R.layout.simple_spinner_item, choices);
+        promptSpinner.setAdapter(adapter);
```
ctsims, line 206: Does this trigger a "selection changed" event at all, or is it only in like 208 that this gets processed?

app/src/org/commcare/activities/QueryRequestActivity.java
```diff
@@ -116,56 +132,176 @@ private void buildPromptUI() {
...
+        promptSpinner.setVisibility(View.VISIBLE);
+        promptView.findViewById(R.id.prompt_et).setVisibility(View.GONE);
+
+        promptSpinner.setOnItemSelectedListener(new AdapterView.OnItemSelectedListener() {
+            @Override
+            public void onItemSelected(AdapterView<?> parent, View view, int position, long id) {
+                Vector<SelectChoice> choices = queryPrompt.getItemsetBinding().getChoices();
+                SelectChoice selectChoice = choices.get(position);
+                String oldAnswer = userAnswers.get(queryPrompt.getKey());
+                if(oldAnswer == null || !oldAnswer.contentEquals(selectChoice.getValue())) {
```
ctsims, line 175: Am I right in assuming that it's this event loop that propagates out of order changes? 

IE: If this spinner view is the last item in the 213 loop, and its value is changed on 208, this selected method is fired, and the loop is redone in case a previous spinner needs to be changed?

shubham1g5:
@ctsims  took a stab at rewiring the itemset dependency loop so that one selection change only triggers the model updation once in the core library. Can you take a look again. 

One question - Do we expect the select prompts to be dependent only on the select prompts or we wanna handle dependency of select prompts on the free text input fields as well ?

ctsims:
@shubham1g5 Let me know if the system already functions as I outlined re: plaintext variables in the xpath reference set. If so, this is approved.

shubham1g5:
@ctsims It was not complex for us to implement select choices dependency on plain text inputs, so this PR now allows for select options to depend on the plain text inputs as well.

shubham1g5:

shubham1g5:
Made rotation related fixes plus tests.

ctsims (approved):
Looks good, and thanks for the expanding test coverage on it as well!

PR #2282, opened 2020-06-22
Use FusedLocation client for capturing location

ShivamPokhriyal:
Jira: https://dimagi-dev.atlassian.net/browse/SAAS-10937


app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -984,9 +1000,7 @@ protected void deliverError(FormEntryActivity receiver, Exception e) {
     }
 
     private void handleFormLoadCompletion(AndroidFormController fc) {
-        if (GeoUtils.ACTION_CHECK_GPS_ENABLED.equals(locationRecieverErrorAction)) {
```
shubham1g5, line 987: so earlier the gps error will only show up if the receiver was fired before the form load completes ?

app/src/org/commcare/activities/GeoPointActivity.java
```diff
@@ -1,42 +1,66 @@
...
 import android.view.View.OnClickListener;
+import android.widget.Toast;
+
+import com.google.android.gms.common.api.ResolvableApiException;
 
 import org.commcare.activities.components.FormEntryConstants;
 import org.commcare.dalvik.R;
+import org.commcare.interfaces.RuntimePermissionRequester;
 import org.commcare.interfaces.TimerListener;
+import org.commcare.location.CommCareFusedLocationController;
```
shubham1g5, line 32: unused

app/src/org/commcare/location/CommCareProviderLocationController.kt
```diff
@@ -0,0 +1,63 @@
...
+        override fun onStatusChanged(provider: String?, status: Int, extras: Bundle?) {
+            //This callback will never be invoked.
+        }
+
+        override fun onProviderEnabled(provider: String?) {
+            TODO("Not yet implemented")
+        }
+
+        override fun onProviderDisabled(provider: String?) {
+            TODO("Not yet implemented")
```
shubham1g5, line 35: should we not communicate this to `mListener` so that it can take actions in case the provider disabled ?

app/src/org/commcare/location/CommCareFusedLocationController.kt
```diff
@@ -0,0 +1,58 @@
+package org.commcare.location
+
+import android.content.Context
+import android.location.Location
+import com.google.android.gms.location.*
+
+/**
+ * @author $|-|!˅@M
+ */
+class CommCareFusedLocationController(private val mContext: Context,
```
shubham1g5, line 10: Wohoo! Kotlin \m/

app/src/org/commcare/location/CommCareFusedLocationController.kt
```diff
@@ -0,0 +1,58 @@
...
+                                      private val mListener: CommCareLocationListener): CommCareLocationController {
+
+    private val mFusedLocationClient = LocationServices.getFusedLocationProviderClient(mContext)
+    private val mLocationRequest = LocationRequest.create().apply {
+        priority = LocationRequest.PRIORITY_HIGH_ACCURACY
+        interval = 5000
```
shubham1g5, line 16: should use a constant for ``5000`

app/src/org/commcare/activities/GeoPointActivity.java
```diff
@@ -47,9 +71,7 @@ protected void onCreate(Bundle savedInstanceState) {
         setTitle(StringUtils.getStringRobust(this, R.string.application_name) +
                 " > " + StringUtils.getStringRobust(this, R.string.get_location));
 
-        locationManager = (LocationManager)getSystemService(Context.LOCATION_SERVICE);
-
-        providers = GeoUtils.evaluateProviders(locationManager);
+        locationController = CommCareLocationControllerFactory.Companion.getLocationController(this, this);
```
shubham1g5, line 74: you can provide a `@JvmStatic` annotation for `getLocationController` to avoid using `Companion` in references.

app/src/org/commcare/activities/GeoPointActivity.java
```diff
@@ -67,36 +89,48 @@ protected void onCreate(Bundle savedInstanceState) {
...
-
-        // stops the GPS. Note that this will turn off the GPS if the screen goes to sleep.
-        if (ContextCompat.checkSelfPermission(this, Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED ||
-                ContextCompat.checkSelfPermission(this, Manifest.permission.ACCESS_COARSE_LOCATION) == PackageManager.PERMISSION_GRANTED) {
-            locationManager.removeUpdates(this);
-        }
-
-        // We're not using managed dialogs, so we have to dismiss the dialog to prevent it from
-        // leaking memory.
-        if (locationDialog != null && locationDialog.isShowing()) {
```
shubham1g5, line 79: shouldn't we still `dismiss` the dialog ?

app/src/org/commcare/activities/FormEntryActivity.java
ShivamPokhriyal: Yeah.

app/src/org/commcare/activities/GeoPointActivity.java
ShivamPokhriyal: Yay! Thanks! I didn't know about this.

app/src/org/commcare/activities/GeoPointActivity.java
ShivamPokhriyal: Aha! Thanks for catching this.

app/src/org/commcare/location/CommCareProviderLocationController.kt
ShivamPokhriyal: Maybe yeah! I just copied over our previous code for this and it wasn't implemented.

app/src/org/commcare/android/javarosa/PollSensorController.java
```diff
@@ -112,27 +111,43 @@ public void onLocationChanged(Location location) {
...
+            Exception exception = ((CommCareLocationListener.Failure.ApiException) failure).getException();
+            if (exception instanceof ResolvableApiException) {
+                apiException = (ResolvableApiException) exception;
+                sendBroadcast(context);
+            } else {
+                // ignore and return, we can't do anything.
+            }
+        } else {
+            noProviders = true;
+            context.registerReceiver(
```
shubham1g5, line 139: where does this receiver get fired from ?

app/src/org/commcare/android/javarosa/PollSensorController.java
```diff
@@ -112,27 +111,43 @@ public void onLocationChanged(Location location) {
         }
     }
 
-    @Override
-    public void onProviderDisabled(String provider) {
+    private void sendBroadcast(Context context) {
+        Intent noGPSIntent = new Intent(GeoUtils.ACTION_CHECK_GPS_ENABLED);
```
shubham1g5, line 115: it seems like this broadcase is getting used to convey any kind of location errors, I would suggest 2 naming changes based on that - 
`sendBroadcast` -> `broadcastLocationError` 
`ACTION_CHECK_GPS_ENABLED` -> `ACTION_LOCATION_ERROR`

app/src/org/commcare/android/javarosa/PollSensorController.java
ShivamPokhriyal: Yeah, so this else part comes when the providers weren't enabled for this device so it sends a broadcast to the activity to ask users to enable providers `GeoUtils.goToProperLocationSettingsScreen`. 
So we register a receiver to listen to events when the providers are changed by the users from settings.

app/src/org/commcare/location/CommCareFusedLocationController.kt
```diff
@@ -24,6 +24,10 @@ class CommCareFusedLocationController(private val mContext: Context,
         }
     }
 
+    companion object {
```
shubham1g5, line 27: nit: should be defined at the top of the class

app/src/org/commcare/location/CommCareProviderLocationController.kt
shubham1g5: yeah we should check for this since the purpose of this work is to make this feature more robust.

ShivamPokhriyal: I've been thinking about this, and it feels like these 2 methods are not needed!

shubham1g5: @ShivamPokhriyal It would be great to append your statements with some reasoning! It's really hard to understand what you have been thinking otherwise.

ShivamPokhriyal: Yeah sorry about that. So, the onProviderDisabled update will come when one of the 2 providers we're using(GPS and network) gets disabled, which simply means that the location updates will cease for that provider, and as soon as it gets enabled again, the updates will resume.  

And we already have a check [in place ](https://github.com/dimagi/commcare-android/pull/2282/files/820f167f3e358de42212fc5df23c56f9792a24bb#diff-784d69745186464f14c4e514ad19ef93R45) which ensures that if the providers are already disabled, we send the user to the appropriate settings page.

shubham1g5: @ShivamPokhriyal  Thanks for the explanation. What if the user stops the provider after our initial check ? would not we want to show the error to the user in that case ?

ShivamPokhriyal: I updated the code to use the broadcast receiver from `PollSensorController` here only. So now, we'll listen to providers change events in the broadcast receivers and these methods can be ignored fully.

app/src/org/commcare/android/javarosa/PollSensorController.java
```diff
@@ -47,22 +62,8 @@ void startLocationPolling(PollSensorAction action) {
...
-
-                // This thread can't take action on the UI, so instead send
-                // a message that actual activities notice and then display
-                // a dialog asking user to enable location access
-                Intent noGPSIntent = new Intent(GeoUtils.ACTION_CHECK_GPS_ENABLED);
-                context.sendStickyBroadcast(noGPSIntent);
-            }
-            requestLocationUpdates(providers);
+            mLocationController = CommCareLocationControllerFactory.getLocationController(context, this);
+            mLocationController.start();
```
shubham1g5, line 66: nit: would be good to delegate it to `requestLocationUpdates`

app/src/org/commcare/location/CommCareProviderLocationController.kt
```diff
@@ -0,0 +1,95 @@
...
+        }
+        startLocationRequest()
+    }
+
+    private fun startLocationRequest() {
+        if (mLocationRequestStarted) {
+            // We've already started location request so no need to request again!
+            return
+        }
+        mLocationRequestStarted = true
```
shubham1g5, line 82: shoudn't this get reset to `false` in `stop()` ?

ShivamPokhriyal: Yup, Thanks for catching this.

ShivamPokhriyal:

ShivamPokhriyal:
@shubham1g5 need another approval.

PR #2284, opened 2020-06-24
Specify ndk version to fix CI builds

shubham1g5:
Jenkins builds are failing from yesterday with the error - `No version of NDK matched the requested version 21.0.6113669. Versions available locally: 21.3.6528147` because of a bug in AGP 4.0 which doesn't download the required ndk version by default.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -206,7 +206,8 @@ static def getDate() {
 
 android {
     compileSdkVersion 29
-
+    // only required to get past a AGP 4.0 bug and should be removed when we update to AGP 4.1+
+    ndkVersion "21.3.6528147"
     lintOptions {
         abortOnError false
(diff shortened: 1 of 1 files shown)


ShivamPokhriyal:
Actually I might be the reason behind these errors. I installed `ndk;21.3.6528147` on the jenkins machine yesterday.

ShivamPokhriyal:
I can perhaps try removing the `ndk tool` from the jenkins machine and see if that resolves the issue.

ShivamPokhriyal:
> I can perhaps try removing the `ndk tool` from the jenkins machine and see if that resolves the issue.

@shubham1g5 Builds are working after removing the ndk tool. I guess we're good to close this one.

shubham1g5:
@ShivamPokhriyal sounds good.

PR #2285, opened 2020-06-26
Navigate form to failed question when form finishes

ShivamPokhriyal:
Fixes side-effect caused by this PR: https://github.com/dimagi/commcare-android/pull/2238. 
When finish button is clicked, and one of the previous questions had a validation error, the form wouldn't go to that failed question, instead it would stay there giving a feeling that the finish button isn't working at all.  

More details here: https://dimagi-dev.atlassian.net/browse/SAAS-11042


ShivamPokhriyal:
Build failed due to the upstream cc_core branch in prbuilder. [Here](https://jenkins.dimagi.com/job/commcare-android-fake/82/) is the successful build

PR #2288, opened 2020-07-09
Use in-built audio widget as default to record audio

ShivamPokhriyal:
Product Note:- The audio-record questions will now use CommCare's in-built audio widget by default. And the older UX can still be achieved using `legacy` appearance attribute. The new audio widget won't have the the option to choose audio from file system by default, but it can be achieved using `acquire-or-upload` appearance attribute

### DIFF
--- a/app/src/org/commcare/views/widgets/CommCareAudioWidget.java
+++ b/app/src/org/commcare/views/widgets/CommCareAudioWidget.java
@@ -42,6 +42,7 @@ public class CommCareAudioWidget extends AudioWidget
     private ImageButton mPlayButton;
     private TextView recordingNameText;
     private MediaPlayer player;
+    private static final String ACQUIRE_UPLOAD_FIELD = "acquire-or-upload";
 
     public CommCareAudioWidget(Context context, FormEntryPrompt prompt,
                                PendingCalloutInterface pic) {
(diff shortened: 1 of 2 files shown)


shubham1g5:
@ctsims It came up on the Field/Dev that the inbuilt audio capture (prototype)  should not have an option to let user choose file for audio capture. The visisbility of the "choose file" option today is managed by another appearance attribute - `acquire` in the external audio widget, but it's always visible in the prototyped audio widget. I am confused in why you think that prototype should not have the choose file option visible instead of controlling visibility of this setting by the `acquire` appearance attribute like in the external audio capture widget. 

Whatever we decide, it would be good to include those changes in this PR and the corresponding user forum post.

ctsims:
Hi Shubham,

For clarity, I simply thought the new audio widget did not already support selecting files from the file system. 

I agree that extending the widget functionality to check the appearance attribute for `acquire` and hiding the button if it is present is the preferred solution. I think we may want to come up with an 'opposite' appearance for `acquire` as well, like `acquire-or-upload` so that if we need to change the default in the future we can tell people to manually enforce that the button should be there. 

I think there are good odds that when we transition to the new widget we might want to default to `acquire` as the initial behavior, since it's more intuitive for what one would expect, and I believe more teams using the capture widgets disable the upload functionality

shubham1g5:
@ctsims Makes sense. Agree on those points. 

@ShivamPokhriyal  couple action items based on that -
 
1. Default to `acquire`("Choose files" disabled) in the inbuilt Audio widget
2. Provide another appearance attr `acquire-or-upload` for making the "Choose Files" button visible.

PR #2289, opened 2020-07-09
Quarantines form records on form processing failures

shubham1g5:
Earlier Form processing failures was resulting in the form getting saved on phone which used to get quarantined during form submission due to a 422 response to the submission. This PR changes this behaviour to delete such form records as soon as user "finish" the form. 

Product Note: Deletes any erroneous forms on form save by user rather than later quarantining them at the time of form submission.

### DIFF
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -58,6 +58,7 @@ import org.commcare.interfaces.WithUIController;
 import org.commcare.logging.analytics.TimedStatsTracker;
 import org.commcare.logic.AndroidFormController;
 import org.commcare.models.AndroidSessionWrapper;
+import org.commcare.models.FormRecordProcessor;
 import org.commcare.models.ODKStorage;
 import org.commcare.models.database.SqlStorage;
 import org.commcare.tasks.FormLoaderTask;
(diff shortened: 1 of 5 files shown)


ctsims:
Hi @shubham1g5 

Is there a reason we would delete this form rather than quarantining it? (We could quarantine it directly, rather than waiting for HQ to quarantine it)

I think there's little risk in maintaining quarantined forms indefinitely (until they get wiped by the saved form cleanup task), and leaving a quarantined form allows us to 

- Provide visibility to the user about an error which occurred, with the permanent ability to go re-view that error
- Report up to HQ in the heartbeat when lots of quarantining is occurring
- Potentially retry parsing the form later if the structural error was a mistake

The downside of quarantining forms seem mild to me (stacking up the files on disk for longer) compared to the potential upside.

shubham1g5:
@ctsims  yeah, that's a good point and I do see how quarantining a form here is useful.

I am mostly concerned how will that play out in terms of maitaining case transactions over time i.e if due to an error a Form A gets quarantined on save showing an error to the user. Since this form errored, user fills up another instance of this same form A and fills up another Form B as a follow up to the Form A. Now if user was able to un-quarantine and re-process the first errored instance of quarantine form, it will overwrite any case trasactions by Form B right ? It's bit unlikely that this will play out as I have mentioned but possible and will be a real difficult data integrity issue to resolve. I also realize these issues are even possible today with the current quarantine framework, though still checking if this is a possibility that we consciously allow for today.

ctsims:
It's a fair point to say that the recovery path for quarantined forms is almost exclusively re-enty, and that we shouldn't expect "re-try form parse" to be a user facing action. Even removing that bullet from the list I think the other two points still add value over deletion.

Side question: We never actually enabled the functionality to retry quarantined forms, right? That's just a hypothetical workflow?

shubham1g5:
> We never actually enabled the functionality to retry quarantined forms, right? That's just a hypothetical workflow?

@ctsims we do have the ability to r[estore quarantine forms](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/activities/FormRecordListActivity.java#L434) that marks the form as unsent and allow to be submitted to HQ again.

PR #2292, opened 2020-07-17
Use dev pref for file oversize warning setting

ShivamPokhriyal:
Jira https://dimagi-dev.atlassian.net/browse/SAAS-11072

Product Note: Uses a custom property `cc-disable-file-oversize-warning` on HQ, which when set to `yes` will disable file-oversized warnings in mobile.


shubham1g5:
Don't think it should be added as a developer option as this setting is mostly inclined towards the users rather than app developers. We can instead add it in advanced settings or just have it controlled via HQ with no visible control on mobile.

app/src/org/commcare/preferences/HiddenPreferences.java
```diff
@@ -525,4 +526,13 @@ public static boolean isLazyMediaDownloadComplete() {
                 .getBoolean(LAZY_MEDIA_DOWNLOAD_COMPLETE, false);
     }
 
+    public static boolean isFileOversizeWarningDisabled() {
+        if (CommCareApplication.instance().isConsumerApp()) {
```
shubham1g5, line 530: why are we overiding for consumer apps ?

ShivamPokhriyal: Ohh I kept it same as `isIncompleteFormsEnabled`. I thought we don't allow custom settings in consumer apps?

shubham1g5: I think thats only true for incomplete forms setting and not a general rule as all other custom settings doesn't have this check.

ShivamPokhriyal:

PR #2295, opened 2020-07-22
Fixes for Lazy Media Download

shubham1g5:
- Better localized Error messages
- Bug Fix: In case when resource download is already in progress, earlier the download listener was not getting notified on download completion causing the UI to get stuck in a "In Progress" state.


shubham1g5:
@ctsims definitely. `suspend` is used to denote a code block that can suspend a `coroutine` without blocking the current thread.  Since the primary pusrpose of using `suspend` is in correlation with a `coroutine` , a `suspend` function can only be called from a `coroutine` or another `suspend` function.  So `suspend` functions are infact just coroutine-supporting buliding block in Kotlin that have the ability to inherit the scope of a `coroutine` it's called from. 

In this particular case, I needed to use the `withContext` block in `recoverResource` to be able to swtich the code to the main thread. the `withContext` is again something that can only be used inside a `suspend` function or a coroutine. That led me to convert `recoverResource` into a `suspend` function as well as `downloadAllLazyMedia`  since a suspend function can only be called from another `suspend` block. Since `downloadAllLazyMedia` only gets called from the Coroutine worker it was alright for it to be defined as `suspend`.

ctsims:
It sounds like the crux of the change is that since you are still invoking the newly `suspend`able functions in `runBlocking` or with [this explicit coroutine scope](https://github.com/dimagi/commcare-android/blob/2897749b0ec3b92916eb9ed080e3e604d4e090b7/app/src/org/commcare/mediadownload/MissingMediaDownloadHelper.kt#L152) that ultimately the code will execute with the same blocking as before, but can deliver results to the main thread. 

No complaints from my side on that, but does seem like the Kotlin coroutine model will introduce a pretty significant amount of complexity as we use it along side traditional Thread concurrency, so we should be mindful of introducing it piecemeal. I'm not even sure if inline docs would have made it more clear to me, so it might be useful for me (and Shivam if he isn't already familiar) to do some directed learning on Kotlin coroutines if we are going to use them.

shubham1g5:
@ctsims Yeah that's a good idea. I will try to be more explicit on PRs around some of these concepts whenever I do end up using them. Though I do think that Android in general is shifting a lot towards Kotlin and Coroutines and Google itself is backing Coroutines as replacement for traditional async routines like Async tasks. So I would support the idea of team spending some time in learning more about these in general.

PR #2301, opened 2020-07-30
Removes ForceTLS12BuilderConfig 

shubham1g5:
I can see a [lot of logs in sumo](https://service.us2.sumologic.com/ui/#/search/BNkbM3wtKZiawwG3WDc3KxEuJSGYpwYCW7xslxN6) with message - `Network.OkHttpTLSCompat: java.lang.IllegalStateException[Unable to extract the trust manager on AndroidPlatform, sslSocketFactory is class org.commcare.network.Tls12SocketFactory]` which means that [this code](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/network/ForceTLS12BuilderConfig.java#L17) to force tls1.2 is failing. Though these devices are not really facing any sort of network issues because of this or at-least the devices I checked were able to sync and submit forms. Therefore it seems like this code is not really needed and the `Conscrypt` libs are handling the tls1.2 transition pretty well without needing to add additional config for forcing TLS1.2. 
I am able to repro the same behaviour on 4.4 devices on broweserstack as well where the tlsv1.2 fails though the network requests work successfully. It's an [issue with OkHTTP](https://github.com/square/okhttp/issues/2427).


shubham1g5:
> It might be worth testing removing this entirely and seeing if the other layer we added is picking up the slack, since OkHTTP is possibly using the conscrypt trust manager anyway/`

@ctsims tested this and this is in fact what's happening which also explains why we never noticed issues due to this error itself. Given that I think removing the `ForceTLS12BuilderConfig.java` altogether makes most sense to remove ambiguity on this in future.

ctsims:
@shubham1g5 Perfect, that sounds right. Sorry for the resulting confusion from this needing two layers

shubham1g5:
@ctsims I did test on a Android 5 device and things do work correctly. Though you did mention [here](https://github.com/dimagi/commcare-android/pull/2239/files#r427740263) that there might be some 5.x devices that doesn't have TLS1.2 as default. Though we also have not heard of any issues happening currently with 5.x devices with the current code to force tls1.2 failing as that would have caused all the app network requests to fail on 5.x devices with less than tls v1.2as default. With that as evidence I am inclined to not extend conscrypt to 5.x devices unless we hear of any issues in future.

ctsims:
@shubham1g5 agreed, thanks for checking

PR #2303, opened 2020-08-01
sub-detail fixes

shubham1g5:
- Retain `EntitySubnodeDetailFragment` to avoid reloading data on configuration changes
- Fixes [the crash](https://console.firebase.google.com/u/0/project/commcare-a57e4/crashlytics/app/android:org.commcare.dalvik/issues/59ddfc4cbe077a4dcc1af280?time=last-thirty-days&sessionId=5F214311007C0001769F8DCE188D03F7_DNE_0_v2) : This crash happens when the user moves out of the sub detail fragment before the `EntityLoaderTask` completes resulting in the `deliverLoadResult` to operate on a changed session scope. The PR just discards the result of the `EntityLoaderTask` if fragment is not visible.

### DIFF
--- a/app/src/org/commcare/views/TabbedDetailView.java
+++ b/app/src/org/commcare/views/TabbedDetailView.java
@@ -73,11 +73,9 @@ public class TabbedDetailView extends RelativeLayout {
 
         mMenu = root.findViewById(R.id.tabbed_detail_menu);
         mViewPager = root.findViewById(R.id.tabbed_detail_pager);
-        mViewPager.setId(AndroidUtil.generateViewId());
-
         mViewPageTabStrip = root.findViewById(R.id.pager_tab_strip);
 
-        mViewPager.setOnPageChangeListener(new ViewPager.OnPageChangeListener() {
+        mViewPager.addOnPageChangeListener(new ViewPager.OnPageChangeListener() {
 
             @Override
             public void onPageSelected(int position) {
--- a/app/src/org/commcare/fragments/EntitySubnodeDetailFragment.java
+++ b/app/src/org/commcare/fragments/EntitySubnodeDetailFragment.java
@@ -22,6 +22,8 @@ import org.javarosa.core.model.instance.TreeReference;
 
 import java.util.List;
 
+import androidx.annotation.Nullable;
+
 /**
  * Created by jschweers on 8/26/2015.
  * <p/>
@@ -33,6 +35,12 @@ public class EntitySubnodeDetailFragment extends EntityDetailFragment implements
     private ListView listView;
 
     public EntitySubnodeDetailFragment() {
+    }
+
+    @Override
+    public void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setRetainInstance(true);
     }
 
     @Override
@@ -65,6 +73,8 @@ public class EntitySubnodeDetailFragment extends EntityDetailFragment implements
             headerLayout.removeAllViews();
             headerLayout.addView(headerView);
             headerLayout.setVisibility(View.VISIBLE);
+        } else if (adapter != null) {
+            listView.setAdapter((ListAdapter)adapter);
         }
 
         return rootView;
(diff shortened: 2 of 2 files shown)


app/src/org/commcare/views/TabbedDetailView.java
```diff
@@ -73,11 +73,9 @@ public void setRoot(ViewGroup root) {
 
         mMenu = root.findViewById(R.id.tabbed_detail_menu);
         mViewPager = root.findViewById(R.id.tabbed_detail_pager);
-        mViewPager.setId(AndroidUtil.generateViewId());
```
shubham1g5, line 76: this already has an id set in xml and setting dynamic ids result in fragment not getting re-attached to the activity.

app/src/org/commcare/views/TabbedDetailView.java
```diff
@@ -73,11 +73,9 @@ public void setRoot(ViewGroup root) {
 
         mMenu = root.findViewById(R.id.tabbed_detail_menu);
         mViewPager = root.findViewById(R.id.tabbed_detail_pager);
-        mViewPager.setId(AndroidUtil.generateViewId());
-
         mViewPageTabStrip = root.findViewById(R.id.pager_tab_strip);
 
-        mViewPager.setOnPageChangeListener(new ViewPager.OnPageChangeListener() {
+        mViewPager.addOnPageChangeListener(new ViewPager.OnPageChangeListener() {
```
shubham1g5, line 78: no-op change since `setOnPageChangeListener ` is deprecated.

app/src/org/commcare/views/TabbedDetailView.java
ctsims: Good catch. It seems really strange that the original line was set at all, do you have a sense of what was trying to be accomplished with it? The only reason I can think you'd set the ID manually would be if you intended to have two ViewPagers within the same Activity.

I went back to check on the git blames and it does look like that is explicitly why that change was made:
https://github.com/dimagi/commcare-android/commit/d2eab7ce90d0f6b7d1f30d81fddb8ad9159b9560

The only context where I can think this occurs is with *persistent case tiles* which do have their own view pager. The previous commit from the one I linked above is

https://github.com/dimagi/commcare-android/commit/a5b7f23591a72992788d69060db77c510beaeb67

which does discuss persistent case tiles, and it's in the persistent case tile PR
https://github.com/dimagi/commcare-android/pull/81

I think the key thing to test is loading a case list screen with a view pager, along with a persistent case tile that has its own view pager. I think both end up in the same general view hierarchy, so they need distinct ID's just based on how the views work but it's possible that other refactors have made that unecessary.

shubham1g5:
@ctsims  thanks for the context. I tested that this PR does break the persistent case tiles when present along with a viewpager. I am going to investigate if there is a better way to handle these both together.

ctsims:
Is there a regression test somewhere that would have caught this change in QA? If not, we should make sure to add one, given how close we were to missing this. I was very close to not catching the issue and only remembered what was going on because it seemed really strange that we would have been doing the original behavior.

shubham1g5:
@ctsims  took another look at it and it doesn't seem possible today for a single screen to contain both the viewpagers. the reason being - 

1. On HQ, there doesn't seem to be a way currently to configure the case list with both an expanding persistent tile and a detail-confirm screen. Checking the "Embed case details in case tile pull-down from the menu" over [here](https://www.commcarehq.org/a/commcare-tests/apps/view/96fc5696f3faddd3a22ca024cd68596c/module/30b841b83e9f51074f2265f14d000729926b512c/#case-detail-screen-config-tab) results into `detail-confirm` being skipped from the datum definition. 

2. On mobile, we don't allow the persistent case tile to be expanded on the case detail screen. Only once a case has been selected and confirmed, the persistent tile can be expanded. This UX makes a lot of sense as well since there is no need to see the expand the tile to see the detail when it's already present on the screen.

shubham1g5:
@ctsims bumping this up for the repro of the edge case this PR breaks.

ctsims:
Hi Shubham, I offlined with a replication for the state I was describing, as seen below

shubham1g5:
@ctsims  thanks I can repro the issue with this. I will followup with some test on it either with QA or Espresso.

damagatchi:

PR #2304, opened 2020-08-01
Only allow the Entity List Action button to be clicked once

shubham1g5:
Issue: On clicking Entity action twice session gets into a weird state causing form save to fail

Repro: 
1. Go to a Entity List screen with an Actions to enter a new form
2. Click the Action button, while the transition to FormEntry happens click the Action button again. 
3. Saving form results into an error as the second action click [clears the form id](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/models/AndroidSessionWrapper.java#L360) set in session 

This is just one example of an issue that I have was able to repro with multiple clicks on actions, but there can be numerous different weird behaviours it can result into like [this crash](https://console.firebase.google.com/u/0/project/commcare-a57e4/crashlytics/app/android:org.commcare.lts/issues/5bedca22f8b88c2963154b02?time=last-thirty-days&sessionId=5F00BFB30205000114285CEE74F2A7F0_DNE_0_v2). The PR just disables the button once it's been clicked.


app/src/org/commcare/views/EntityActionViewUtils.java
```diff
@@ -64,6 +66,9 @@ private static void setupActionClickListener(FrameLayout actionCardView,
                                                  final Action action,
                                                  final CommCareActivity commCareActivity) {
         CardView cardView = actionCardView.findViewById(R.id.card_body);
-        cardView.setOnClickListener(v -> EntitySelectActivity.triggerDetailAction(action, commCareActivity));
+        cardView.setOnClickListener(v -> {
+            cardView.setEnabled(false);
```
ctsims, line 70: Extremely minor point: I don't think this is a real concern, but if the EntityList view was recycled entirely (IE: The use pressed 'back' and the view came up exactly as it was) it would be the case that this button would remain disabled. 

Again, I don't think this applies to the way we use the views, but I didn't want to discount it entirely in case you knew differently, since you have worked with this more recently than  I have.

shubham1g5: @ctsims Thanks for flagging. You are right that it's a non concern as `onClick` here calls `EntitySelectActivity.triggerDetailAction` which is a terminal operation in the sense that it always `finish` the current activity. So pressing back always results in CommCare session to recreate the `EntityListActivity` again re-enabling the action button.

ctsims: great, thanks for confirming

PR #2305, opened 2020-08-03
Add idling resource to avoid AppNotIdleExceptions

ShivamPokhriyal:
The tests are sometimes failing with `AppNotIdleException`. This PR will add an idling resource to make tests more robust.

### DIFF
--- a/app/instrumentation-tests/src/org/commcare/utils/CommCareIdlingResource.kt
+++ b/app/instrumentation-tests/src/org/commcare/utils/CommCareIdlingResource.kt
@@ -1,0 +1,55 @@
+package org.commcare.utils
+
+import android.os.Handler
+import android.os.Looper
+import androidx.test.espresso.IdlingResource
+import androidx.test.platform.app.InstrumentationRegistry
+import org.commcare.CommCareInstrumentationTestApplication
+import org.commcare.activities.CommCareActivity
+
+/**
+ * An implementation of idling resource useful for monitoring idleness of progress dialogs.
+ *
+ * It has an additional idleness constraint that the progress dialog must not be shown for a set
+ * period of time before we declare the resource idle.
+ *
+ * Useful for cases where we make multiple network calls in succession where each response triggers
+ * another request until loading is complete.
+ * So the app will be like Not-Idle(Progress bar showing) -> Idle(Progress bar closed) -> Not-Idle -> Idle...
+ * And we don't want to report idle when this happens.
+ */
+class CommCareIdlingResource: IdlingResource {
+
+    companion object {
+        const val timeoutMs = 1000L
+    }
+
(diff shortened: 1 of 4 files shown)


app/instrumentation-tests/src/org/commcare/utils/CommCareIdlingResource.kt
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * It has an additional idleness constraint that the progress dialog must not be shown for a set
+ * period of time before we declare the resource idle.
+ *
+ * Useful for cases where we make multiple network calls in succession where each response triggers
+ * another request until loading is complete.
+ * So the app will be like Not-Idle(Progress bar showing) -> Idle(Progress bar closed) -> Not-Idle -> Idle...
+ * And we don't want to report idle when this happens.
+ */
+class CommCareIdlingResource: IdlingResource {
```
shubham1g5, line 21: nit: `CommCareIdlingResource ` -> `ProgressIdlingResource`

app/instrumentation-tests/src/org/commcare/utils/CommCareIdlingResource.kt
```diff
@@ -0,0 +1,55 @@
...
+    var isIdle = false
+    val handler = Handler(Looper.getMainLooper())
+
+    override fun getName(): String {
+        return CommCareIdlingResource::class.java.name
+    }
+
+    override fun isIdleNow(): Boolean {
+        val activity = getCommCareActivity() ?: return true
+        if (!(activity.currentProgressDialog != null && activity.currentProgressDialog.isAdded)) {
```
shubham1g5, line 37: so effectlvely what we are trying to do is we are marking the resource as `Idle` a sec after the progress dialog has been removed. What's the behaviour if we don't add this at all ? The app never goes to `Idle` or takes more time to do so ?

ShivamPokhriyal: The app would take 1 more sec before going to idle.

shubham1g5: got it. thanks!

shubham1g5: might be worth adding that as comment either in the class javadoc or inline comment as that's the crux of why we need this class.

ShivamPokhriyal: Ooops Sorry I missed this comment before merging. 
BTW, I've already added docs here https://github.com/dimagi/commcare-android/pull/2305/files#diff-b30c4312c3ff1a1c9e97d4a741c20500R10-R20

PR #2307, opened 2020-08-04
[2.49.5 hotfix] Fix Update Resumptions

shubham1g5:
- Corrects update resumptions to only reset table with specific install codes
- Corrects auto update retries to retry on network failures 
- Clears updateStats anytime we clear upgrade


shubham1g5:
@ctsims Wanted to push this sooner, therefore making this PR without tests for now. I am going to followup with QA on this so that we get this behaviour tested by QA as well before we push it out on both LTS and prod. I will separately follow up with unit tests around these sometime later as it's going to be a bit of work to explore different ways to mock requests and different failure modes.

PR #2318, opened 2020-08-18
Add jacoco code coverage plugin

ShivamPokhriyal:

PR #2319, opened 2020-08-18
Fixes success check for Update worker

shubham1g5:
- Corrects check for success of the Update worker. Marking the update worker run as `success` vs `failure` should not cause any issues in the field so not pushing a hotfix for this. 
- Adds tests for Update worker results along with a test to check that update is not cleared after network failures.


shubham1g5:
Build failure should be fixed once we get in https://github.com/dimagi/commcare-android/pull/2318

dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/461818/)

PR #2320, opened 2020-08-19
[2.49.7 hotfix] Fix to current location for Maps Location Capture

shubham1g5:
* Corrects the accuracy check to let the initial location fix go through 
* Don't allow 0,0,0,0 location to be recorded

Produce Notes: 
* Move the maps to user's initial location on start up in case no existing location is set
* Don't allow locatin with values "0.0 0.0 0.0 0.0" to be recorded as a valid location

### DIFF
--- a/app/src/org/commcare/activities/GeoPointMapActivity.java
+++ b/app/src/org/commcare/activities/GeoPointMapActivity.java
@@ -123,7 +123,7 @@ public class GeoPointMapActivity extends Activity
     }
 
     private void returnLocation() {
-        if (location != null &&
+        if (location == null ||
                 (location.getLatitude() == 0d && location.getLongitude() == 0d && location.getAccuracy() == 0d)) {
             // we do not have a location fix yet, show an error to the user
             Toast.makeText(
--- a/app/src/org/commcare/activities/GeoPointMapActivity.java
+++ b/app/src/org/commcare/activities/GeoPointMapActivity.java
@@ -14,6 +14,7 @@ import android.view.View;
 import android.view.Window;
 import android.widget.Button;
 import android.widget.TextView;
+import android.widget.Toast;
 
 import com.fasterxml.jackson.databind.type.MapType;
 import com.google.android.gms.maps.CameraUpdate;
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/activities/GeoPointMapActivity.java
```diff
@@ -153,7 +164,8 @@ private void loadProviders() {
     public void onLocationChanged(Location location) {
         if (!inViewMode && !isManualSelectedLocation) {
             if (location != null &&
-                    (this.location == null || (location.getAccuracy() < this.location.getAccuracy()))) {
```
shubham1g5, line 156: this was breaking when there is no initial location set on the map as new location's accuracy is always greater than 0 while base location's accuracy is set to 0 on start up

app/src/org/commcare/activities/GeoPointMapActivity.java
```diff
@@ -120,12 +123,20 @@ private void changeMapLayer() {
...
+                (location.getLatitude() == 0d && location.getLongitude() == 0d && location.getAccuracy() == 0d)) {
+            // we do not have a location fix yet, show an error to the user
+            Toast.makeText(
+                    this,
+                    StringUtils.getStringRobust(CommCareApplication.instance(), R.string.wait_for_location_fix),
+                    Toast.LENGTH_LONG).show();
+
+        } else {
             Intent i = new Intent();
             i.putExtra(FormEntryConstants.LOCATION_RESULT, GeoUtils.locationToString(location));
```
ShivamPokhriyal, line 136: What if location is null?

app/src/org/commcare/activities/GeoPointMapActivity.java
ShivamPokhriyal: Shouldn't the `this.location == null` check work when there's no initial location?

shubham1g5: location is [initialized](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/activities/GeoPointMapActivity.java#L54) to non null value to start with. So I don't think if it will ever be null.

app/src/org/commcare/activities/GeoPointMapActivity.java
shubham1g5: my bad, corrected.

app/src/org/commcare/activities/GeoPointMapActivity.java
ShivamPokhriyal: Got it, so `this.location == null` is never going to be true or do you think there can be a possibility of a null location?

shubham1g5: yeah, I don't think there is such a possibility though have not removed it to minimize changes in the hotfix

ShivamPokhriyal (approved):
Looks good to me. Though the build fails, so I'll use a different job to run the build and will ping you when it succeeds.

shubham1g5:
@ShivamPokhriyal  you can track the build status [here](https://jenkins.dimagi.com/job/commcare-android-2.49/)

ShivamPokhriyal:
> @ShivamPokhriyal you can track the build status [here](https://jenkins.dimagi.com/job/commcare-android-2.49/)

Ohh awesome. Thanks!

PR #2322, opened 2020-08-23
Implements request age and source in Install requests headers

shubham1g5:
Spec: https://docs.google.com/document/d/1apnkFHZ9c38DKmUSarb67w5PJYQ74L0Xwy56Nb13Gpo/edit (Approach 1)

Also contains:

* Implements a `ResourceInstallContext` object in `platform` that keeps tracks of `InstallRequestSource` during resource installs. It also removes the need of having another `recovery` param in ResourceInstallers. 
* Fixes an issue where lazy media would get skipped during App install as well
* Adds tests around lazy media download
* Adds Mockk to assist with mocking objects in unit tests

cross-request: https://github.com/dimagi/commcare-core/pull/929


app/src/org/commcare/network/CommcareRequestGenerator.java
```diff
@@ -226,17 +233,19 @@ private static String getDigest() {
...
     public Response<ResponseBody> simpleGet(String uri) throws IOException {
-        return simpleGet(uri, new HashMap());
+        return simpleGet(uri, new HashMap<>(), new HashMap<>());
     }
 
     @Override
-    public Response<ResponseBody> simpleGet(String uri, Map<String, String> httpParams) throws IOException {
+    public Response<ResponseBody> simpleGet(String uri, Map<String, String> httpParams, Map<String, String> httpHeaders) throws IOException {
+        HashMap<String, String> headers = getHeaders("");
+        headers.putAll(httpHeaders);
```
ctsims, line 242: It's best practice to treat these map objects as immutable to ensure that params don't leak between requests [as is done here](https://github.com/dimagi/commcare-android/pull/2322/files#diff-20bb24bd598ae6ae4f34b3bdb1cfbf2aR207)

app/src/org/commcare/network/RequestStats.kt
```diff
@@ -0,0 +1,73 @@
...
+ */
+object RequestStats {
+
+    @JvmStatic
+    fun register(installRequestSource: InstallRequestSource) {
+        register(CommCareApplication.instance().currentApp, installRequestSource)
+    }
+
+    @JvmStatic
+    fun register(app: CommCareApp, installRequestSource: InstallRequestSource) {
```
ctsims, line 23: This is going to make the SharedPreferences object enormous, right? With potentially thousands of entries?

I don't think that's a best practice for how we use these buckets. It could affect their perforamnce for what we expected to be very fast I/O lookups, and it will be hard for us to do things like migrate them. I think it would be better to store this data in its own db tables like we do for a few other models like the geocoder entity cache

shubham1g5:
@ctsims  thanks that's great advice. I will follow up on this PR with the implementation on `InstallContext`.

app/src/org/commcare/network/RequestStats.kt
shubham1g5: > This is going to make the SharedPreferences object enormous, right? With potentially thousands of entries?

Not really, we are not making an entry here per resource but per install workflow which gets erased on a successful run.  Furthermore, for multiple failed runs of a install workflow, we only ever maintain the time of the first failed attempt to get the time since first failed attempt. So there can only be at max 6 entries each corresponding to each of the install workflow from `InstallRequestSource`

ctsims: We had a big problem with the other update status collector object blowing up like crazy, so I still think that if we are going to keep some number of these records, we should make sure that number is a fixed LIFO queue (Even if it's fixed at some reasonably large size like 100), rather than letting it grow indefinitely. 

I'd also say based on  previous challenges that the mechanism that keeps it fixed should also be upfront (when the new item is added) rather than relying on event based removal, even if event based removal is still present (and keeps the list small in normal operation).

ctsims:
@shubham1g5 Cool, thanks, happy to presume that the Install context stuff happens in a following PR if you'd rather get this in and then refactor, just let me know.

app/src/org/commcare/network/RequestStats.kt
shubham1g5: @ctsims Would not the [check](https://github.com/dimagi/commcare-android/blob/updateRequestHeaders/app/src/org/commcare/network/RequestStats.kt#L25) for the first attempt here automatically gurantees that there is only ever one entry in preference per Install workflow restricting the max number of these prefs to the # of enum entries [here](https://github.com/dimagi/commcare-core/blob/updateRequestHeaders/src/main/java/org/commcare/resources/model/InstallRequestSource.java#L6).

ctsims: I see, I totally misread this code. I thought it was one bucket per Install Attempt per Source (max of 6 buckets per Install attempt), not just one bucket per source for a max of 6 buckets total. 

Thanks for clarifying, you can close this out in that case.

app/src/org/commcare/AppUtils.java
```diff
@@ -177,6 +177,13 @@ public static String getCurrentVersionString() {
         return Localization.get(application.getString(R.string.app_version_string), new String[]{pi.versionName, String.valueOf(pi.versionCode), ccv, buildNumber, buildDate, profileVersion});
     }
 
+    public static String getCommCareVersion() throws PackageManager.NameNotFoundException {
+        CommCareApplication application = CommCareApplication.instance();
+        PackageManager pm = application.getPackageManager();
+        PackageInfo pi = pm.getPackageInfo(application.getPackageName(), 0);
+        return pi.versionName;
```
ShivamPokhriyal, line 184: Curious how is it different from `BuildConfig.VERSION_NAME`

shubham1g5: ahh I should definitely be using `BuildConfig.VERSION_NAME` here. thanks for flagging.

app/src/org/commcare/mediadownload/MissingMediaDownloadHelper.kt
```diff
@@ -171,7 +177,7 @@ object MissingMediaDownloadHelper : TableStateListener {
     }
 
 
-    private suspend fun downloadMissingMediaResource(uri: String): MissingMediaDownloadResult {
+    private fun downloadMissingMediaResource(uri: String): MissingMediaDownloadResult {
```
ShivamPokhriyal, line 180: Also wondering why we removed `suspend` here?

app/src/org/commcare/network/CommcareRequestGenerator.java
```diff
@@ -226,17 +242,19 @@ private static String getDigest() {
...
     @Override
     public Response<ResponseBody> simpleGet(String uri) throws IOException {
-        return simpleGet(uri, new HashMap());
+        return simpleGet(uri, new HashMap<>(), new HashMap<>());
     }
 
     @Override
-    public Response<ResponseBody> simpleGet(String uri, Map<String, String> httpParams) throws IOException {
+    public Response<ResponseBody> simpleGet(String uri, Map<String, String> httpParams, Map<String, String> httpHeaders) throws IOException {
+        HashMap<String, String> headers = new HashMap<>(getHeaders(null));
```
ShivamPokhriyal, line 246: Earlier, we were using `getHeaders("")` and now we're using `getHeaders(null)` which means now we'd not be having `X_COMMCAREHQ_LAST_SYNC_TOKEN` in the headers. Is this expected?

app/src/org/commcare/mediadownload/MissingMediaDownloadHelper.kt
shubham1g5: `suspend` is only required when we launch a corountine from this method or this method calls another `suspend` functions. So it didn't serve any purpose to declare this method as `suspend`

app/src/org/commcare/network/CommcareRequestGenerator.java
shubham1g5: yes, I don't think it serves any purpose to have `X_COMMCAREHQ_LAST_SYNC_TOKEN` as empty in headers.

app/src/org/commcare/network/RequestStats.kt
```diff
@@ -0,0 +1,73 @@
...
+        val firstAttempt = getFirstRequestAttempt(app, installRequestSource);
+        if (firstAttempt == -1L) {
```
ShivamPokhriyal, line 25: nit: we can move -1 to a constant

app/src/org/commcare/update/UpdateHelper.java
```diff
@@ -81,11 +83,11 @@ public static UpdateHelper getNewInstance(boolean autoUpdate, UpdateProgressList
...
 
     // Main UpdateHelper function for staging updates
-    public ResultAndError<AppInstallStatus> update(String profileRef) {
-        setupUpdate(profileRef);
+    public ResultAndError<AppInstallStatus> update(String profileRef, InstallRequestSource installRequestSource) {
+        setupUpdate(profileRef, installRequestSource);
 
         try {
-            return new ResultAndError<>(stageUpdate(profileRef));
+            return new ResultAndError<>(stageUpdate(profileRef, installRequestSource));
```
ShivamPokhriyal, line 90: Should we call `markSuccess` here?

ShivamPokhriyal: Ohh nvm, it's getting set from `UpdateTask`. Perhaps, it'd be good to move the call to `register` outside as well?

app/src/org/commcare/mediadownload/MissingMediaDownloadHelper.kt
```diff
@@ -147,14 +153,14 @@ object MissingMediaDownloadHelper : TableStateListener {
      * Downloads a resource with it's location represented by {@param mediaUri}
      */
     @JvmStatic
-    fun requestMediaDownload(mediaUri: String, missingMediaDownloadListener: MissingMediaDownloadListener) {
+    fun requestMediaDownload(mediaUri: String, defaultDispatcher: CoroutineDispatcher,
```
shubham1g5, line 156: was changed to support running on main thread for unit tests.

app/src/org/commcare/update/UpdateHelper.java
shubham1g5: Good point, calling both `register` and `markSuccess` inside [now](https://github.com/dimagi/commcare-android/pull/2322/commits/6f8c007476e1a49900426a2c4c43dd2ac536fdbc)

shubham1g5:
@ctsims This is ready for another review. Also FYI that I have merged https://github.com/dimagi/commcare-android/pull/2325 as well in this since changes in that PR were too coupled with this.

shubham1g5:
@ctsims bumping this up in your queue for your review as this seems to be only major review left for 2.50

ctsims (approved):

ShivamPokhriyal (approved):

PR #2324, opened 2020-08-23
null adapter check in menu to avoid NPE

shubham1g5:
[Crashlytics](https://console.firebase.google.com/u/0/project/commcare-a57e4/crashlytics/app/android:org.commcare.lts/issues/e3110befb09bd58913f9cff4b716f036?time=last-thirty-days&sessionId=5F04A74D038C00010B02783790F45A2E_DNE_0_v2)

Not sure why adapter is `null` in `onPrepareOptionsMenu`. Looks like `onPrepareOptionsMenu` might be getting called out of order but not sure how. I think it's better to show an empty menu which user can possibly correct by going back in menu in these rare cases rather than crashing the app so just putting a null check to solve this.

### DIFF
--- a/app/src/org/commcare/activities/FormRecordListActivity.java
+++ b/app/src/org/commcare/activities/FormRecordListActivity.java
@@ -557,19 +557,21 @@ public class FormRecordListActivity extends SessionAwareCommCareActivity<FormRec
     public boolean onPrepareOptionsMenu(Menu menu) {
         super.onPrepareOptionsMenu(menu);
 
-        MenuItem quarantine = menu.findItem(MENU_SUBMIT_QUARANTINE_REPORT);
-        if (quarantine != null) {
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/activities/FormRecordListActivity.java
```diff
@@ -557,19 +557,21 @@ public boolean onQueryTextChange(String newText) {
     public boolean onPrepareOptionsMenu(Menu menu) {
         super.onPrepareOptionsMenu(menu);
 
-        MenuItem quarantine = menu.findItem(MENU_SUBMIT_QUARANTINE_REPORT);
-        if (quarantine != null) {
-            quarantine.setVisible(FormRecordFilter.Limbo.equals(adapter.getFilter()));
-        }
+        if (adapter != null) {
```
shubham1g5, line 560: diff is not super clear though the change is just that this null check is now wrapping the rest of the code in `onPrepareOptionsMenu`

PR #2326, opened 2020-08-24
Show downloaded app update info in case of no internet

ShivamPokhriyal:
In case an app update is already downloaded and user enters app update screen but doesn't have internet connectivity, we'll show the information of already downloaded update.

### DIFF
--- a/app/src/org/commcare/activities/UpdateActivity.java
+++ b/app/src/org/commcare/activities/UpdateActivity.java
@@ -196,9 +196,15 @@ public class UpdateActivity extends CommCareActivity<UpdateActivity>
 
         NSDDiscoveryTools.registerForNsdServices(this, this);
 
-        if (!ConnectivityStatus.isNetworkAvailable(this) && offlineUpdateRef == null) {
-            uiController.noConnectivityUiState();
-            return;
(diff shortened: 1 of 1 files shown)


shubham1g5 (commented):
@ShivamPokhriyal  It seems like the `onCreate` calls the `setupUpdateTask` without doing a connectivity check. So whatever UI state we set here should be getting overriden by the result of the updateTask in `handleTaskCompletion`. Wondering if you are not seeing the same happen when testing this ?

ShivamPokhriyal:
> @ShivamPokhriyal It seems like the `onCreate` calls the `setupUpdateTask` without doing a connectivity check. So whatever UI state we set here should be getting overriden by the result of the updateTask in `handleTaskCompletion`. Wondering if you are not seeing the same happen when testing this ?

That call actually ends up doing nothing, cuz the check here https://github.com/dimagi/commcare-android/blob/77022ecc013cbec9c1c3dc33eda554a1963ba4ce/app/src/org/commcare/activities/UpdateActivity.java#L151-L152 would fail when there's no internet.

shubham1g5 (approved):
cool. Looks good in that case.

PR #2328, opened 2020-08-25
Adds set alarm permission

shubham1g5:
Today in case an application wishes to set an alarm using a custom callout, they get an error because of this permission not being requested by the application. Since `SET_ALARM` is of [normal](https://developer.android.com/guide/topics/permissions/overview#normal_permissions) protection level, these permissions are auto granted without user approval.

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -29,6 +29,9 @@
     <uses-permission android:name="android.permission.CAMERA" />
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
 
+    <!-- used to create an alarm from a custom callout -->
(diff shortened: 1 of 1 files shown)


shubham1g5:
@ctsims This was flagged by Ahmad for a requirement with rapid tests where they want to set alarms to check back on the results.

ctsims:
Hi @shubham1g5 I'll offline with Ahmad and make sure he knows about the RDT stuff, I think we'd probably rather direct any rapid test elements over to that, but will confirm whether it's a hard requirement

shubham1g5:
@ctsims Closing this for now assuming that it's not required anymore.

PR #2330, opened 2020-08-27
Fix audio recorder filename

ShivamPokhriyal:
Uses a simple unique fileName(using Date.getTime) for audio files.


app/src/org/commcare/views/widgets/CommCareAudioWidget.java
```diff
@@ -138,12 +138,6 @@ public void onRecordingCompletion(String audioFile) {
         recordingNameText.setText(Localization.get("recording.custom"));
     }
 
-    @Override
-    public String getFileUniqueIdentifier() {
-        String formFileName = FormEntryInstanceState.mFormRecordPath.substring(FormEntryInstanceState.mFormRecordPath.lastIndexOf("/") + 1);
-        return formFileName + "_" + mPrompt.getIndex().toString() + "_" + (new Date().getTime());
```
ShivamPokhriyal, line 144: The fileName returned by this is really weird. It has a `.xml`and `,`(a comma if mPrompt.getIndex returns something like 2,3).

app/src/org/commcare/views/widgets/RecordingFragment.java
```diff
@@ -41,8 +43,6 @@
     public static final String AUDIO_FILE_PATH_ARG_KEY = "audio_file_path";
 
     private String fileName;
-    private static final String FILE_BASE =
-            "/Android/data/org.commcare.dalvik/temp/Custom_Recording";
```
ShivamPokhriyal, line 45: This was causing failures in debug builds as we have a different appID.

shubham1g5:
@ShivamPokhriyal  Is this change just for a temporary file and the file we attach to the form ends up being different than this ? Curious since this file is stored in `temp` and I think we delete all files from `temp` folder on a fresh application startup .

ShivamPokhriyal:
> @ShivamPokhriyal Is this change just for a temporary file and the file we attach to the form ends up being different than this ? Curious since this file is stored in `temp` and I think we delete all files from `temp` folder on a fresh application startup .

Yep, this is just for the temporary file. When the file recording is completed, it triggers the `RecordingCompletionListener` which would end up calling `MediaWidget.setBinaryData` and it would always create a new file https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/views/widgets/MediaWidget.java#L207-L211

shubham1g5 (approved):
looks good.

PR #2333, opened 2020-08-28
Better log tag for AppUpdateManager

ShivamPokhriyal:


app/src/org/commcare/appupdate/CommcareFlexibleAppUpdateManager.java
```diff
@@ -21,7 +21,7 @@
  */
 public class CommcareFlexibleAppUpdateManager implements FlexibleAppUpdateController {
 
-    private static final String TAG = CommcareFlexibleAppUpdateManager.class.getName();
+    private static final String TAG = CommcareFlexibleAppUpdateManager.class.getSimpleName();
```
shubham1g5, line 24: Hey we should use more generalized tag names from `LogTypes` instead of class names as tags

shubham1g5: we are still writing logs with this tag.

shubham1g5 (changes requested):

shubham1g5 (approved):

PR #2336, opened 2020-09-03
Inflate MediaLayout using xml [WIP]

ShivamPokhriyal:
### Breaking Change.

An attempt to fix the UI change caused by #2243 probably [this](https://github.com/dimagi/commcare-android/pull/2243/files#diff-5f60968a18182e1d222e7378017fe2a7R162) (Inline Video View is really shrinked with a lot of white spaces See screenshot)
Screenshot_2020-08-31-15-48-46-984_org commcare dalvik debug

I've moved the layout logic to it's own xml which also avoids the redundant nested view groups.

### TODO: 

* [ ] ~~Allow multiple missing media views to be present when the question contains multiple missing media.~~
     - Seems like we only show inline-video view when both inline-video view and image are present in the question. 
* [x]  Fix showImageAboveText. 
* [x]  Fix video view visiiblilty issue once the video is downloaded(using missing media UI)
* [x] Add a background color to the missing media view.
* [x] Show missing media view for invalid image reference.
* [x] Check ResizingImageView, it was showing some error in the layout editor.
* [ ] Validate this new change doesn't break any existing UI. Possibly through screenshot-tests #2339


app/src/org/commcare/views/media/MediaLayout.java
```diff
@@ -114,69 +104,57 @@ public static MediaLayout buildComprehensiveLayout(Context context,
...
+    }
 
-        if (mediaView != null) {
-            RelativeLayout.LayoutParams mediaViewParams =
-                    new RelativeLayout.LayoutParams(LayoutParams.WRAP_CONTENT, LayoutParams.WRAP_CONTENT);
-            mediaViewParams.addRule(CENTER_IN_PARENT, mediaView.getId());
-            mediaPane.addView(mediaView, mediaViewParams);
+    private void setupStandardAudio(String audioURI, int questionIndex) {
+        if (audioURI != null) {
+            boolean mediaPresent = FileUtil.referenceFileExists(audioURI);
```
ShivamPokhriyal, line 143: Wondering if we need to do something here?

app/src/org/commcare/views/media/MediaLayout.java
```diff
@@ -342,129 +232,66 @@ private View setupImage(String imageURI, String bigImageURI) {
...
+                showMissingMediaView(imageURI,
+                        StringUtils.getStringRobust(getContext(), R.string.image_download_prompt),
+                        true,
+                        () -> {
+                            hideMissingMediaView();
+                            setupImage(imageURI, bigImageURI);
+                        });
             }
         } catch (InvalidReferenceException e) {
             Log.e(TAG, "image invalid reference exception");
```
ShivamPokhriyal, line 249: Same, do we need to show something here as well?

app/src/org/commcare/views/media/MediaLayout.java
shubham1g5: the audio button should be handling the missing media state by itself, providing an option to download the file in case it's missing.

ShivamPokhriyal: awesome. Thanks!

shubham1g5:
@ShivamPokhriyal Wondering if you are seeing a different behaviour on the currently released Playstore version. To me both the released and master video playback behaves exactly same with same amount of whitespace.

app/src/org/commcare/views/media/MediaLayout.java
shubham1g5: we can follow the same pattern as videoView.

ShivamPokhriyal:
> @ShivamPokhriyal Wondering if you are seeing a different behaviour on the currently released Playstore version. To me both the released and master video playback behaves exactly same with same amount of whitespace.

Yeah. the master build shows a lot of space to me, while the released version doesn't. You can try [this app](https://www.commcarehq.org/a/srishti-1/apps/view/2f91bcb96f94afa828ac8b908e888973/releases/).

shubham1g5:
Got it. I can repro using your app. I was using a portrait video and possibly the bug doesn't surface with those. Would be good for you to test your solution with [this app maybe](https://www.commcarehq.org/a/shubhamgoyaltest/apps/view/1956b120ec4e173ed6a5a160b76fb648/form/ca9d66211df4a2140bd2323038dd07d3256fe375/source/#form/woman_name) so that we make sure we are not breaking anything. 

Also the [diff link](https://github.com/dimagi/commcare-android/pull/2243/files#diff-5f60968a18182e1d222e7378017fe2a7R162) in the PR description doesn't point me to anything specific but the PR and it seems like you were trying to point to something specific which broke this. Can you maybe try relinking the code point where it breaks.

ShivamPokhriyal:
> Can you maybe try relinking the code point where it breaks.

Sure, I think this https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/views/media/MediaLayout.java#L160-L162 code adds the unwanted space and shrinks the videoview.

shubham1g5:
@ShivamPokhriyal Sounds good. Let me know when it's ready to review again as we would want to get this in soon given the code freeze on Friday.

ShivamPokhriyal:
@shubham1g5 While carrying out the todo for `showImageAboveText` I realised that this layout hierarchy doesn't consider multiple missing medias. 
Also, I'm really afraid that this change might break a lot of UI(paddings/margins). So I was thinking of setting up the screenshot tests to validate that this change doesn't break anything. So I guess it's gonna take a lot more time.  

I've also edited the description of the PR with a list of TODOs.

shubham1g5:
@ShivamPokhriyal  moving this to 2.51 in that case.

shubham1g5:
@ShivamPokhriyal I just noticed that the `showImageAboveText` is also breaking in master which makes it critical for us to fix this in 2.50. Wondering how do you feel about completing it for 2.50 without the screenshot tests (QA should be able to take care of that). If it's not likely I can attempt to fix these issues separately without xml layout in 2.50

ShivamPokhriyal:
> @ShivamPokhriyal I just noticed that the `showImageAboveText` is also breaking in master which makes it critical for us to fix this in 2.50. Wondering how do you feel about completing it for 2.50 without the screenshot tests (QA should be able to take care of that). If it's not likely I can attempt to fix these issues separately without xml layout in 2.50

TBH I'm not really confident with this change😅. Though the only thing remaining in this PR is the video view visibility issue which is really tricky and I haven't been able to figure it out yet. If you think QA can take care of testing this change then I can definitely try fixing the video view issue and get it merged before QA starts.

app/src/org/commcare/views/media/MediaLayout.java
```diff
@@ -350,13 +351,13 @@ private void setupInlineVideoView(String inlineVideoURI) {
...
-        int[] maxBounds = getMaxCenterViewBounds();
         //These surprisingly get re-jiggered as soon as the video is loaded, so we
         //just want to give it the _max_ bounds, it'll pick the limiter and shrink
         //itself when it's ready.
+        DisplayMetrics metrics = this.getContext().getResources().getDisplayMetrics();
         ViewGroup.LayoutParams params = videoView.getLayoutParams();
-        params.width = maxBounds[0];
-        params.height = maxBounds[1];
+        params.width = metrics.widthPixels;
+        params.height = metrics.heightPixels;
```
ShivamPokhriyal, line 360: the reason why videoView doesn't appear after download completion was this maxBounds[1]. In my case, the textView had a really long text and I had to scroll down to see the videoView. 

The `getMaxCenterViewBounds` method subtracts the text height from the total height
`maxHeight = maxHeight - textViewContainer.getChildAt(0).getHeight();` making this `maxBounds[1]` to be 0. 

Since, videoView will automatically configure itself with the min height possible, we're good to give it max height here.

app/src/org/commcare/views/media/MediaLayout.java
ShivamPokhriyal: FYI, I'll remove this unused variable then.

app/src/org/commcare/views/media/MediaLayout.java
```diff
@@ -473,64 +337,75 @@ private View getInlineVideoView(String inlineVideoURI, RelativeLayout.LayoutPara
...
-
-                videoView.setId(INLINE_VIDEO_PANE_ID);
-                return videoView;
+                videoView.setVisibility(VISIBLE);
             }
         } catch (InvalidReferenceException ire) {
             Log.e(TAG, "invalid video reference exception");
             ire.printStackTrace();
-            return getMissingMediaView(inlineVideoURI, "Invalid reference: " + ire.getReferenceString(), false);
+            showMissingMediaView(inlineVideoURI, "Invalid reference: " + ire.getReferenceString(), false, null);
```
ShivamPokhriyal, line 346: @shubham1g5 Do we need to localize this text?

ShivamPokhriyal: Also the information here isn't really informative imo. Would it be better if we say: `Can't show media due to: `?

app/src/org/commcare/views/media/MediaLayout.java
```diff
@@ -342,129 +271,64 @@ private View setupImage(String imageURI, String bigImageURI) {
...
-
             final File videoFile = new File(videoFilename);
             if (!videoFile.exists()) {
-                return getMissingMediaView(inlineVideoURI,
+                showMissingMediaView(inlineVideoURI,
                         StringUtils.getStringRobust(getContext(), R.string.video_download_prompt),
-                        true);
+                        true,
+                        () -> {
+                            hideMissingMediaView();
```
shubham1g5, line 303: stylistic point that you can ignore: `hideMissingMediaView` can be called from `showMissingMediaView` to DRY the completion logic further.

app/src/org/commcare/views/media/MediaLayout.java
```diff
@@ -541,49 +416,39 @@ private boolean useResizingImageView() {
...
-            dividerParams.addRule(RelativeLayout.BELOW, audioButton.getId());
-        } else if (viewText != null) {
-            // No picture
-            dividerParams.addRule(RelativeLayout.BELOW, viewText.getId());
-        } else {
-            Log.e(TAG, "Tried to add divider to uninitialized ATVWidget");
-            return;
-        }
-        addView(v, dividerParams);
+    private void resetView() {
```
shubham1g5, line 438: this method can be avoided by setting the initial visibility of all views as gone in xml layout itself.

shubham1g5 (changes requested):
Looks quite good to me. Left some very minor points for you to consider. On testing it's regressions, we should check the 2.50 test plan and see if our current tests have testcases for 1. QR view setup in the image 2. Missing Media (these would definitely need to get added) 3. `showImageAboveText` usecases.

app/src/org/commcare/views/media/MediaLayout.java
```diff
@@ -144,12 +146,12 @@ private void showMediaAboveText() {
...
         alignMediaAtTop(imageView);
         alignMediaAtTop(resizingImageView);
+        alignMediaAtTop(missingMediaBackground);
         alignMediaAtTop(downloadIcon);
         alignMediaAtTop(progressBar);
 
-        // Next align the text, audiobutton and videoButton below mediaView.
+        // Next align the text, audiobutton below mediaView.
         alignTextContainerBelowMediaView(audioButton);
-        alignTextContainerBelowMediaView(videoButton);
```
shubham1g5: confused why did we remove `videoButton` from here

app/src/org/commcare/views/media/MediaLayout.java
```diff
@@ -417,6 +426,7 @@ private boolean useResizingImageView() {
     }
 
     private void initView(Context context) {
+        setId(AndroidUtil.generateViewId());
```
shubham1g5, line 429: why is this required ?

shubham1g5 (commented):
left Couple questions, though build seems to be failing on this one.

app/src/org/commcare/views/media/MediaLayout.java
ShivamPokhriyal: If not set, `this.getId()` call above returns -1.

app/src/org/commcare/views/media/MediaLayout.java
ShivamPokhriyal: `videoButton` has a constraint which places it below `audioButton`. So we only need to align `audioButton` top constraint while shifting it below media view.

ShivamPokhriyal:

app/src/org/commcare/views/media/MediaLayout.java
shubham1g5: Ahh Makes sense.

app/res/layout/media_layout.xml
```diff
@@ -1,12 +1,21 @@
...
+        android:layout_height="wrap_content"
+        android:orientation="horizontal"
+        app:layout_constraintGuide_begin="0dp"/>
 
     <org.commcare.views.media.AudioPlaybackButton
         android:id="@+id/audio_button"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
-        android:visibility="gone"
+        tools:visibility="gone"
```
shubham1g5, line 18: tools:visibility should be `visible` while keeping the `android:visibility` as `gone`

ShivamPokhriyal: Ahh Damn, silly me! I thought it behaves the same as `android:visibility`

app/src/org/commcare/views/media/MediaLayout.java
```diff
@@ -40,54 +44,46 @@
...
+        super(context, attrs);
+        initView(context);
+    }
+
+    public MediaLayout(@NonNull Context context, @Nullable AttributeSet attrs, int defStyleAttr) {
+        super(context, attrs, defStyleAttr);
+        initView(context);
     }
 
     public static MediaLayout buildAudioImageLayout(Context context, TextView text, String audioURI, String imageURI) {
```
shubham1g5, line 89: this seems ununsed and can be removed.

app/src/org/commcare/views/media/MediaLayout.java
```diff
@@ -114,69 +110,75 @@ public static MediaLayout buildComprehensiveLayout(Context context,
...
+        // media(video, image qr ) above text.
 
-        LayoutParams mediaPaneParams = refreshMediaView();
+        // Change the questionBottomBarrier to align at the top of parent which would automatically
+        // align the mediaViews(including missing media view) to the top.
+        questionBottomBarrier.setReferencedIds(new int[]{ R.id.top_guideline });
 
-        addAudioVideoButtonsToView(questionTextPane);
+        // Next align the text, audiobutton below mediaView.
+        alignTextContainerBelowMediaView(audioButton);
```
shubham1g5, line 151: Wondering if it's possible to avoid multiple calls of `alignTextContainerBelowMediaView` as well and instead use barrier/guideline mechanism to shift these views ?

shubham1g5 (approved):
Looks great! Kudos 🎆

PR #2340, opened 2020-09-09
Increase no of allowed apps to 4

shubham1g5:
https://dimagi-dev.atlassian.net/browse/SAAS-11265

Product Note: Increase the limit on number of allowed CommCare apps on a single CommCare instance to four


shubham1g5:

PR #2344, opened 2020-09-17
Fix radio button visibility

ShivamPokhriyal:
A fix for: https://dimagi-dev.atlassian.net/browse/QA-1900


shubham1g5:
@ShivamPokhriyal There are few other places where we add media Layout into heirarchy and we should verify if this same change needs to be replicated for them as well.

ShivamPokhriyal:
> @ShivamPokhriyal There are few other places where we add media Layout into heirarchy and we should verify if this same change needs to be replicated for them as well.

It's working fine in other places.

PR #2297, opened 2020-07-27
Don't crash when no form is found in storage while XForm Uninstall

shubham1g5:
Fix for - 
````
Caused by java.util.NoSuchElementException: No record in table form_def for ID 41
       at org.commcare.models.database.SqlStorage.readBytes(SqlStorage.java:458)
       at org.commcare.models.database.SqlStorage.read(SqlStorage.java:449)
       at org.commcare.android.resource.installers.XFormAndroidInstaller.uninstall(XFormAndroidInstaller.java:129)
       at org.commcare.android.resource.installers.XFormAndroidInstaller.uninstall(XFormAndroidInstaller.java:52)
       at org.commcare.resources.model.ResourceTable.uninstallResourcesForStatus(ResourceTable.java:877)
       at org.commcare.resources.model.ResourceTable.clearUpgrade(ResourceTable.java:852)
````
[CL link](https://console.firebase.google.com/u/0/project/commcare-a57e4/crashlytics/app/android:org.commcare.lts/issues/5b51ed486007d59fcd2cb39f?time=last-thirty-days&sessionId=5F1F227202CD000116960D4D7F5E5B43_DNE_0_v2)

I have not been able to reproduce it but I think it's ok behaviour for us to not crash the App when the form id of a form we are uninstalling is not found in the storage as we were going to remove that form id anyway.

PR #2349, opened 2020-09-18
Fixes XForm Uninstall

shubham1g5:
https://dimagi-dev.atlassian.net/browse/QA-1904

It's a regression of https://github.com/dimagi/commcare-android/pull/2297/files where I am calling `super.uninstall` at a wrong place as we should only remove the associlated XForm files when we remove the corresponding form def entry from the DB.


ShivamPokhriyal (approved):
Ohh wow!

ShivamPokhriyal:
@shubham1g5 Can this be merged?

PR #2350, opened 2020-09-18
fix update completion indicator in case of background update

shubham1g5:
Issue: When user opens up the Update check while a background update is in progress, the update UI fails to transition to update "ready to install" state because of incorrect check on updateStats in case of a success

### DIFF
--- a/app/src/org/commcare/activities/UpdateActivity.java
+++ b/app/src/org/commcare/activities/UpdateActivity.java
@@ -188,8 +188,9 @@ public class UpdateActivity extends CommCareActivity<UpdateActivity>
 
         // Update Stats are only maintained in case of an unsuccessful update,
         // so check whether we have a ready to install update instead
-        if (lastUpdateResult == null && ResourceInstallUtils.isUpdateReadyToInstall()) {
-            lastUpdateResult = new ResultAndError<>(AppInstallStatus.UpdateStaged);
+        if (lastUpdateResult == null) {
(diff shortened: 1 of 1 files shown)


ShivamPokhriyal:
Is there is a Jira ticket for this one?

shubham1g5:
no, It's not coming from QA.

app/src/org/commcare/activities/UpdateActivity.java
```diff
@@ -177,17 +177,22 @@ private void connectToUpdateWorker() {
...
-                        }
+                        handleTaskCompletion(getlastStageUpdateResult());
                     }
                 });
     }
 
     private ResultAndError<AppInstallStatus> getlastStageUpdateResult() {
         UpdateStats updateStats = UpdateStats.loadUpdateStats(CommCareApplication.instance().getCurrentApp());
-        return updateStats.getLastStageUpdateResult();
+        ResultAndError<AppInstallStatus> lastUpdateResult = updateStats.getLastStageUpdateResult();
```
ShivamPokhriyal, line 187: Why are we using this one if `ResourceInstallUtils.isUpdateReadyToInstall()` gives the correct result?

shubham1g5: `ResourceInstallUtils.isUpdateReadyToInstall()` is just an indication of the update's success or failure while in case of failures, the exact reason of the failure is contained in `lastUpdateResult`

ShivamPokhriyal: got it. Thanks for explaining that!

ShivamPokhriyal (approved):
Looks good to me.

PR #2351, opened 2020-09-20
Orphan Progress dialog fix

shubham1g5:
https://dimagi-dev.atlassian.net/browse/QA-1893

Because we were using `dialog.show` instead of `dialog.showNow`, it allowed Fragment manager to lazily execute on committed fragment transactions resulting in a state where `dismiss` was called even before the `show` transaction executes resulting in a hanging progress dialog as described in the JIRA linked above. 

This PR changes the `showProgressDialog` to use `dialog.showNow` and then make other supporting changes so that a task only dismisses the progress dialog generated by itself. This is done to avoid incorrect dialog cancellation in case of a task transition. For example in case of sync, if we don't use task specific dialog cancellation the `DataPullTask` new dialog gets cancelled by the concluding `ProcessAndSendTask` dismiss dialog call resulting in the pull task dialog to never gets visible. 

It is not clear to me why this issue is only surfacing now in 2.50. Nothing related seems to have changed in between 2.49 and 2.50 and the whole code path is same line to line.

### DIFF
--- a/app/src/org/commcare/activities/CommCareActivity.java
+++ b/app/src/org/commcare/activities/CommCareActivity.java
@@ -7,11 +7,6 @@ import android.content.DialogInterface;
 import android.graphics.Rect;
 import android.os.Build;
 import android.os.Bundle;
-
-import androidx.fragment.app.DialogFragment;
-import androidx.fragment.app.FragmentActivity;
-import androidx.fragment.app.FragmentManager;
-
 import android.text.Spannable;
 import android.util.DisplayMetrics;
 import android.util.Log;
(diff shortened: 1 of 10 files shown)


app/src/org/commcare/activities/CommCareActivity.java
```diff
@@ -591,7 +593,7 @@ public void showProgressDialog(int taskId) {
         if (taskId >= 0) {
             CustomProgressDialog dialog = generateProgressDialog(taskId);
             if (dialog != null) {
-                dialog.show(getSupportFragmentManager(), KEY_PROGRESS_DIALOG_FRAG);
+                dialog.showNow(getSupportFragmentManager(), KEY_PROGRESS_DIALOG_FRAG);
```
shubham1g5, line 596: crux of the issue

ShivamPokhriyal (approved):

ShivamPokhriyal:
@shubham1g5 test failures seem to be real here
```
01:53:00 org.commcare.android.tests.DemoUserRestoreTest > demoUserRestoreAndUpdateTest FAILED
01:53:00     java.lang.AssertionError at DemoUserRestoreTest.java:120
01:53:26 
01:53:26 org.commcare.android.tests.activities.PostRequestActivityTest > retryClaimTest FAILED
01:53:26     java.lang.RuntimeException at PostRequestActivityTest.java:167
01:53:26         Caused by: java.lang.IllegalStateException at PostRequestActivityTest.java:167
```

ShivamPokhriyal (approved):

PR #2353, opened 2020-09-21
Lazy Media fixes

shubham1g5:
* https://dimagi-dev.atlassian.net/browse/QA-1906 - was happening because of a NPE on `mResourceInProgressListener` which was getting set to null on a background thread and later on accessed on UI thread without null check. 

* https://dimagi-dev.atlassian.net/browse/QA-1915 - Error message correction

* Earlier the `recoverResource` which gets fired by both background worker and the foreground download was invoking the UI thread to notify the listener. This logic is now moved to the `downloadResource` which is invoked by foreground download process only to avoid the background worker suspending because of unavailablity of the main thread.

### DIFF
--- a/app/src/org/commcare/mediadownload/MissingMediaDownloadHelper.kt
+++ b/app/src/org/commcare/mediadownload/MissingMediaDownloadHelper.kt
@@ -1,5 +1,6 @@
 package org.commcare.mediadownload
 
+import android.util.Log
 import androidx.annotation.WorkerThread
 import androidx.work.*
 import kotlinx.coroutines.*
@@ -15,7 +16,6 @@ import org.commcare.resources.model.*
 import org.commcare.network.RequestStats
 import org.commcare.resources.ResourceInstallContext
 import org.commcare.tasks.ResultAndError
-import org.commcare.util.LogTypes
 import org.commcare.utils.AndroidCommCarePlatform
 import org.commcare.utils.FileUtil
 import org.commcare.utils.StringUtils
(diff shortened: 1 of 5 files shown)


app/src/org/commcare/mediadownload/MissingMediaDownloadHelper.kt
```diff
@@ -10,27 +10,29 @@ import org.commcare.engine.resource.AndroidResourceUtils
...
 import java.util.concurrent.TimeUnit
 
 // Contains helper functions to download lazy or missing media resources
 object MissingMediaDownloadHelper : TableStateListener {
 
+    @Volatile
     private var mResourceInProgressListener: MissingMediaDownloadListener? = null
+
+    @Volatile
     private var resourceInProgress: Resource? = null
```
shubham1g5, line 34: any vars which are shared between thread should be marked as `volatile` to ensure that the writes to these are published to all threads immediately

app/src/org/commcare/views/media/MediaLayout.java
```diff
@@ -378,7 +378,7 @@ private void showMissingMediaView(String mediaUri, String errorMessage, boolean
                     progressBar.setVisibility(GONE);
                     downloadIcon.setVisibility(VISIBLE);
                     downloadIcon.setEnabled(true);
-                    missingMediaStatus.setText(StringUtils.getStringRobust(getContext(), R.string.media_download_failed));
+                    missingMediaStatus.setText(result.getMessage());
```
shubham1g5, line 381: https://dimagi-dev.atlassian.net/browse/QA-1915

ShivamPokhriyal (approved):

shubham1g5:

PR #2358, opened 2020-09-24
Add fixtures test

ShivamPokhriyal:
Added FixtureTest from calabash to espresso.


app/instrumentation-tests/src/org/commcare/androidTests/FixturesTest.kt
```diff
@@ -0,0 +1,166 @@
...
+                .check(matches(isDisplayed()))
+
+        onView(withText("Boston"))
+                .perform(click())
+        onView(withId(R.id.nav_btn_finish))
+                .perform(click())
+    }
+
+    @Test
+    fun testFixture_with1Mb_works() {
```
ShivamPokhriyal, line 122: This should fail on master right now.

app/instrumentation-tests/src/org/commcare/androidTests/FixturesTest.kt
```diff
@@ -0,0 +1,166 @@
...
+                .check(matches(isDisplayed()))
+    }
+
+    @Test
+    fun testFixturesWork() {
+        InstrumentationUtility.login("fixtures_works", "123")
+        openFixtureForm_andSelectFirstTwoCheckboxes()
+        onView(withText("Essex"))
+                .perform(click())
+        onView(withId(R.id.nav_btn_next))
```
shubham1g5, line 50: both next and prev navigations can be encapsulated into functions so that if we ever change the view id of buttons, we only need to change the ui tests in one place.

shubham1g5 (approved):
one minor comment for you to take or leave, looks good otherwise.

app/instrumentation-tests/src/org/commcare/androidTests/FixturesTest.kt
ShivamPokhriyal: Noting that test failed successfully.

app/instrumentation-tests/src/org/commcare/androidTests/FixturesTest.kt
ShivamPokhriyal: I think it goes with all the strings/ids that we're using in androidTests and it's always hard to have them at one place. So I feel like not doing it.

ShivamPokhriyal:
@shubham1g5 Another review please.

shubham1g5 (dismissed):

dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/461976/)

ShivamPokhriyal:
@shubham1g5 one more approval please.

shubham1g5 (approved):

PR #2367, opened 2020-10-05
Add clear buttons to all questions with radio buttons

ShivamPokhriyal:
Product Note: Add support for clearing selection of ListWidget using clear button. 

Screenshot_2020-10-05-16-38-21-149_org commcare dalvik debug
Screenshot_2020-10-05-16-34-50-236_org commcare dalvik debug


dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/461993/)

orangejenny:
this looks great!

ShivamPokhriyal:

app/src/org/commcare/views/widgets/LabelWidget.java
```diff
@@ -168,6 +177,14 @@ public LabelWidget(Context context, FormEntryPrompt prompt) {
         addView(questionLayout);
     }
 
+    /**
+     * LabelWidget is used for both ListWidget and ListMultiWidget, and we don't show the clear
+     * button in ListMultiWidget. So this method allows removal of clear button.
+     */
+    public void removeClearButton() {
```
shubham1g5, line 184: Few design suggestions here - 
 1. This should move to `ClearableWidget` so that we localize the logic related to clear button in `ClearableWidget` 
2. An alternative approach to this would be to define a method `showClearButton` in `ClearableWidget` and only add the button if it evals to true. This way we avoid adding the clear button at all for widgets implementing `ClearableWidget` but not wanting to show the button. 
3. A even better approach would be use composition instead of Inheritance here for `ClearableWidget` especially since not all subclasses of `ClearableWidget` ends up showing Clear button in this design.

shubham1g5:
@ShivamPokhriyal  @orangejenny  Wondering if we would want to implement this on web apps sometime in future. I am a liitle worried about us providing  different instructions ("X" vs "clear") to user for clearing the answer on different types of questions and ideally I would like to retain "X" vs "clear" for all kinds of question widgets considering it's more compact and fits well for all widgets. I also realize that we already have rolled out the "clear" and it will be a slight design change for user to shift from "Clear" to "X" though I feel like it should not be too much to adapt to for the user. Thoughts ?

orangejenny:
About X vs "Clear": I agree consistency is important, and we all agree clear doesn't fit well in the combined. But I don't think the X looks natural when it's on the bottom left, as it is in the auto-advanced widget and regular single selects. Xs are used so often in the top right to close things it's weird to see one beneath something.

What if instead of an X for each row, we displayed a single "Clear" button to clear them all? From a UX perspective, it would be consistent, and I think it's fine to only allow clearing all questions, but I'm not sure how feasible it is, code-wise, to have a button that acts across multiple questions.

@dimagi/product might also have thoughts.

I had thought web apps didn't support these, but it does - not the quick advance, but the combined question style. I just opened https://github.com/dimagi/commcare-hq/pull/28627 to add a clear button to that, and can update to do an all-questions clear if we decide to do that on mobile.

app/src/org/commcare/views/widgets/LabelWidget.java
ShivamPokhriyal: Gotcha, Composition sounds better than inheritance.

shubham1g5:
> What if instead of an X for each row, we displayed a single "Clear" button to clear them all? 

But then the user won't have a way to clear choice for a single question in the list which can be quite annoying for the user specially if the list is long since user would have to fill up all the choices again which they didn't want to clear in the first place. I would prefer choosing productivity over consistency if we have to make a choice here.

orangejenny:
@shubham1g5 Then are you saying keep the Xs in the combined widget, but leave the "Clear" text in the auto-advanced and regular select questions? I'm okay with that.

shubham1g5:
@orangejenny I think my preferred choice so far is showing "X" everywhere instead of "clear"  but understand that it might appear wierd to show "X" at bottom right instead of top right in which case I am okay with the current inconsistent compromise of showing "X" and "Clear" selectively for different widgets.

orangejenny:
@shubham1g5 Cool, got it. Yeah, I'm pretty opposed to the X at the bottom, so I think the inconsistent compromise is the place where we agree, unless product jumps in to override us.

app/src/org/commcare/views/widgets/LabelWidget.java
ShivamPokhriyal: I have a question here, this `removeClearButton` method is required because `LabelWidget` is used in combination with `ListWidget` and `ListMultiWidget` and we want the clear button only in 1 combination(with `ListWidget`). 
So whatever suggestion I end up using with, I still need this method here I guess.

shubham1g5:
thanks @orangejenny for your inputs here. All of it makes sense and agree that neither of our options are perfect here. 

@ShivamPokhriyal  I also just realized that we already have a way to remove the response by long clicking the question. So maybe we just want to stick to that instead of introducing anything new here since there are downsides of all the alternative options we have explored so far. One other issue I see with a cross inline with choices is it's going to clutter the UI in case there are more than 2 choices per question.

app/src/org/commcare/views/widgets/LabelWidget.java
shubham1g5: You can inverse the check by introducing another method like `shouldShowClearButton` and have the `ClearableWidget` like - 
````
if(shouldShowClearButton()){
// add clear button
````
The difference is just that we are doing a check while adding the clear button instead of adding and then removing the clear button

Comment (ShivamPokhriyal) [reply app/src/org/commcare/views/widgets/LabelWidget.java:184]: Ok, I'll offline regarding this.

**Comment by dimagimon** [comment]
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462027/)

**Comment by ShivamPokhriyal** [comment]
> One other issue I see with a cross inline with choices is it's going to clutter the UI in case there are more than 2 choices per question.

Fully agree with you on this. It's gonna interfere with the UI formatting for existing applications, and might make the choices label break into multi-lines(when the space isn't enough).

**Comment by orangejenny** [comment]
> It's gonna interfere with the UI formatting for existing applications

This is a really good point. There are likely projects that would consider this a breaking change, and I'm not sure how we communicate potentially breaking changes in mobile releases.

Leaving the combined question UI alone, and just having people long press if needed, also sounds fine to me. Although, given that the long press exists, could it make sense to add a "Clear" button that clears all options, and then people can long-press if they need to clear an individual item? I don't feel strongly about this, just another thought. I definitely agree all of these options are imperfect.

**Comment by shubham1g5** [comment]
> Although, given that the long press exists, could it make sense to add a "Clear" button that clears all options, and then people can long-press if they need to clear an individual item? I don't feel strongly about this, just another thought.

No strong thoughts here though I think it's better to allow users to remove answers one question at a time. Also what particular use case would this serve ? It's hard to imagine for me user often wanting to clear all items just on one particular page and not exit the form entirely. If we have to do it, I would prefer hiding it in the menu options so that users use it only when absolutely needed and there are less chances of users mistakenly clicking on it and loose their data.

**Comment by orangejenny** [comment]
My main thought is that a clear all button would be more discoverable than solely relying on long press. But I don't feel that strongly either...since neither of us are that enthusiastic, let's just drop it.

**Comment by shubham1g5** [comment]
@orangejenny Got it. that definitely makes sense. Thanks for helping in thinking through this.

File: app/src/org/commcare/views/widgets/ClearableWidget.java [hunk 2]
```diff
@@ -0,0 +1,39 @@
...
+
+    public ClearableWidget(Context context, FormEntryPrompt p) {
+        super(context, p);
+    }
+
+    public ClearableWidget(Context context, FormEntryPrompt p, boolean inCompactGroup) {
+        super(context, p, inCompactGroup);
+    }
+
+    protected void setupClearButton(Context context, String text, int visibility) {
```
shubham1g5, line 30: @ShivamPokhriyal The way we are using it now, it makes more sense for this to be simply a static utilitiy method rather than a member of the Widget Classes. Thoughts ?

ShivamPokhriyal: Yeah, I'm not really sure if this subclassing is doing any good here, since the actual addition is happening inside the view only.

shubham1g5 (approved):

dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462085/)

dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462087/)

PR #2370, opened 2020-10-06
Fix for login using PIN when no User exists for a UserKeyRecord

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/browse/SUPPORT-4456 (Issue 1)

Repro: 

1. Install App A, Login as user X and set Pin
2. Install App B, Login as same user X and set Pin
3. Uninstall App A (this deleted the User X DB)
4. Login using pin in App B
5. Login fails 

Issue happens because the `DataPullTask` always expects a `password` and [in cases when there is no user in DB for the UKR](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/tasks/ManageKeyRecordTask.java#L121) we were sending a pin as password to the `DataPullTask`

### DIFF
--- a/app/src/org/commcare/tasks/ManageKeyRecordTask.java
+++ b/app/src/org/commcare/tasks/ManageKeyRecordTask.java
@@ -150,7 +150,7 @@ public abstract class ManageKeyRecordTask<R extends DataPullController> extends 
 
     protected void keysReadyForSync(R receiver) {
         // TODO: we only wanna do this on the _first_ try. Not subsequent ones (IE: On return from startDataPull)
-        receiver.startDataPull(this.dataPullMode);
+        receiver.startDataPull(this.dataPullMode, password);
     }
 
     protected void keysLoginComplete(R receiver) {
--- a/app/src/org/commcare/activities/DataPullController.java
+++ b/app/src/org/commcare/activities/DataPullController.java
@@ -18,7 +18,7 @@ public interface DataPullController {
         CCZ_DEMO
     }
 
-    void startDataPull(DataPullMode mode);
+    void startDataPull(DataPullMode mode, String password);
 
     void dataPullCompleted();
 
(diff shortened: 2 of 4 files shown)


dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462000/)

ctsims (approved):

app/src/org/commcare/tasks/ManageKeyRecordTask.java
```diff
@@ -150,7 +150,7 @@ protected void deliverError(R receiver, Exception e) {
 
     protected void keysReadyForSync(R receiver) {
         // TODO: we only wanna do this on the _first_ try. Not subsequent ones (IE: On return from startDataPull)
-        receiver.startDataPull(this.dataPullMode);
+        receiver.startDataPull(this.dataPullMode, password);
```
ShivamPokhriyal, line 153: This might be a silly question, but does this mean that in subsequent logins using pin, this password will be sent as null?  cuz [here](https://github.com/dimagi/commcare-android/pull/2370/files#diff-4da5bd220b747979dec7869dd4d3bc7fR92-R98) it's being set to null for pin login mode

shubham1g5: Password gets set from [here](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/tasks/ManageKeyRecordTask.java#L460) in pin mode, so it should not be null.

ShivamPokhriyal: gotcha

dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462019/)

PR #2371, opened 2020-10-07
Add RTL support to textview

ShivamPokhriyal:
Jira: https://dimagi-dev.atlassian.net/browse/SAAS-11349

This PR will add RTL support to the textviews that weren't already supporting RTL. 

The new UI after this change:
Screenshot_1602054543
Screenshot_1602054549
Screenshot_1602054558


app/src/org/commcare/views/ViewUtil.java
```diff
@@ -195,4 +196,17 @@ private static void divideIndeterminateSpace(int[] widths, int fullSize,
...
         }
     }
+
+
+    /**
+     * Adds RTL support to the textview. It will make the textView ALIGN_LEFT if the view’s resolved
+     * layoutDirection is LTR, and ALIGN_RIGHT otherwise.
+     */
+    public static void addRTLSupport(TextView textView) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
```
ShivamPokhriyal, line 206: We can also probably have add these to the `CommonTheme` which will make all the textviews RTL supportive. 
But then I guess, we'll have to find out all the buttons/textviews which could be broken because of `view_start` textDirection, and will have to manually change the text direction for them.

shubham1g5 (changes requested):
@ShivamPokhriyal I would prefer any view specific attributes to reside in `xml` as much as possible. If not through a theme, it should be fairly easy to inflate a TextView from xml and set these properties there.

ShivamPokhriyal:
> @ShivamPokhriyal I would prefer any view specific attributes to reside in `xml` as much as possible. If not through a theme, it should be fairly easy to inflate a TextView from xml and set these properties there.

Though it's simple, but considering the textView I'd create in xml can only have these 2 attributes `textDirection` and `textAlignment` and the rest of the attributes like `textSize` or `padding` would still need to be configured from java as [here](https://github.com/dimagi/commcare-android/pull/2371/files#diff-be2c74b44c59c9b17b7e5542ee646a50L194-L198) since they have different values in different places. Also I'll have to create 2 xml since these properties only work starting from version 17. So having a TextView xml doesn't really seem like a nice option to me in this case. But I'd like to hear your thoughts about why we used do otherwise.

shubham1g5:
It just feels like a bad design choice to have the view code in Java. It clutters the business logic and makes it hard to debug issues. The idea of having the Java code to swarm with bunch of RTL related calls doesn't appeal to me. It feels messy and unclean. Even though you will be creating multiple Textview xmls here, you would leave the Java counterpart much more cleaner than before in turn making it relatively easier to focus on business logic for anyone debugging issues in these areas.

ShivamPokhriyal:

dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462004/)

app/res/layout-v17/textview_rtl.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<TextView xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="wrap_content"
+    android:layout_height="wrap_content"
+    android:textAlignment="viewStart"
+    android:textDirection="locale" />
```
shubham1g5, line 6: @ShivamPokhriyal curious what does textDirection does here, from screenshots it seems like only thing we need to adjust is alignment ?

ShivamPokhriyal: I used `textDirection` as I had a feeling that it is used to define the direction of the text. Like it would change `him` to `mih` for rtl language, though it's not doing that even when I change it to rtl.

shubham1g5: yeah agree that we don't need `textDirection` for any static texts. It's mostly used to define cursor direction when entering text.

app/src/org/commcare/activities/components/FormLayoutHelpers.java
```diff
@@ -12,6 +12,7 @@
 import org.commcare.activities.FormEntryActivity;
 import org.commcare.dalvik.R;
 import org.commcare.preferences.FormEntryPreferences;
+import org.commcare.views.ViewUtil;
```
shubham1g5, line 15: unused import

app/src/org/commcare/views/widgets/ListMultiWidget.java
```diff
@@ -20,6 +21,7 @@
 import org.commcare.dalvik.R;
 import org.commcare.utils.MediaUtil;
 import org.commcare.utils.StringUtils;
+import org.commcare.views.ViewUtil;
```
shubham1g5, line 24: unused import

shubham1g5 (changes requested):
Looks good except there are few unused imports that needs to be cleaned up

ShivamPokhriyal:

dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462022/)

ShivamPokhriyal:

ShivamPokhriyal:
@shubham1g5 another review please.

shubham1g5 (changes requested):
There are still a bunch of unused imports.

ShivamPokhriyal:

shubham1g5 (approved):

ShivamPokhriyal:

ShivamPokhriyal:

dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462070/)

app/res/layout/question_widget_text.xml
```diff
@@ -4,5 +4,6 @@
     android:layout_height="match_parent"
     android:gravity="start"
     android:scrollHorizontally="false"
+    android:textAppearance="@style/TextAppearance.AppCompat.Body1"
```
shubham1g5, line 7: Can you confirm this doesn't cause any significant size difference as compared to earlier ?

ShivamPokhriyal: Just checked, I don't see any difference in the size.

shubham1g5: thanks for confirming.

shubham1g5 (approved):

Issue #2341, opened 2020-09-11
Move tests to espresso. 


ShivamPokhriyal:
* [x] DeletedUser
* [x] FormEntry
* [x] Fixtures
* [x] MenuTests
* [x] Login
* [x] FormSettings
* [x] CaseSharing
* [x] Languages
* [x] CaseListSearch
* [x] CaseListSort
* [x] Settings
* [x] FormFiltering
* [x] Tiles
* [x] SessionExpiration
* [x] Dialogs
* [x] InstallList
* [x] DemoUserOnline
* [x] DemoUserOffline
* [x] LookupTableOptionSorting
* [x] ManualQuarantine
* [x] LogSubmission

ShivamPokhriyal:
@shubham1g5 Since we've removed the manual quarantine option, will it be okay to not include `ManualQuarantine` in espresso?

shubham1g5:
@ShivamPokhriyal  we still have manual quarantine , we have just added another step to enable this from advanced actions. So we should still make necessary changes for the new UI and implement it.

ShivamPokhriyal:
Ohh great, I'll add it then.

PR #2382, opened 2020-10-19
Add languages test

ShivamPokhriyal:
#2341


app/instrumentation-tests/src/org/commcare/androidTests/LanguagesTest.kt
```diff
@@ -0,0 +1,135 @@
...
+@LargeTest
+class LanguagesTest : BaseTest() {
+
+    companion object {
+        const val CCZ_NAME = "languages.ccz"
+    }
+
+    @Before
+    fun setup() {
+        if (CommCareApplication.instance().currentApp == null) {
```
shubham1g5, line 30: `BaseTest::installApp` do just the same right ?

app/instrumentation-tests/src/org/commcare/androidTests/LanguagesTest.kt
```diff
@@ -0,0 +1,135 @@
...
+        }
+        InstrumentationUtility.login("user_with_no_data", "123")
+    }
+
+    @Test
+    fun testLanguageChanges() {
+        InstrumentationUtility.openForm(0, 0)
+        onView(withText("Enter a name:"))
+                .check(matches(isDisplayed()))
+        exitForm()
```
shubham1g5, line 45: should move this to `InstrumentationUtility` as it's useful for most tests.

app/instrumentation-tests/src/org/commcare/androidTests/LanguagesTest.kt
```diff
@@ -0,0 +1,135 @@
...
+        // Confirm app language remains hindi
+        onView(withText("Start"))
+                .perform(click())
+        onView(withText("Basic Form Tests"))
+                .perform(click())
+        onView(withText("HIN: Languages"))
+                .check(matches(isDisplayed()))
+    }
+
+    private fun selectLanguageChangeOption() {
```
shubham1g5, line 112: This can be generalized and move to `InstrumentationUtility` as `selectMenu(int textId)`,

app/instrumentation-tests/src/org/commcare/androidTests/LanguagesTest.kt
```diff
@@ -0,0 +1,135 @@
...
+        InstrumentationUtility.rotatePortrait()
+        checkLanguageOptions()
+
+        // Clicking just after rotating doesn't work for some reason, so waiting here for a moment.
+        InstrumentationUtility.sleep(2)
+        onView(withText("Hindi"))
+                .perform(click())
+        onView(withText("Start"))
+                .perform(click())
+        onView(withText("Basic Form Tests"))
```
shubham1g5, line 62: can reuse `InstrumentationUtility.openForm(0, 0)`

shubham1g5 (approved):
Left some optional feedback

app/instrumentation-tests/src/org/commcare/androidTests/LanguagesTest.kt
ShivamPokhriyal: Actually there is a very subtle difference between these 2. 
`BaseTest::installApp` method will install the app only when the app is not already installed. 
This method here will install the app every time even if it's already installed. I had to use it to start with a fresh state so that the changes I make in one method doesn't affect the other method.

app/instrumentation-tests/src/org/commcare/androidTests/LanguagesTest.kt
ShivamPokhriyal: Agree on this, will make the changes.

app/instrumentation-tests/src/org/commcare/androidTests/LanguagesTest.kt
ShivamPokhriyal: Strange, I don't know why I did this. Will change it.

dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462074/)

app/instrumentation-tests/src/org/commcare/androidTests/LanguagesTest.kt
shubham1g5: what makes this test special that we are using a different mechanism here than other tests ?

ShivamPokhriyal: The tests in espresso can run in any order. Meaning `testLanguageChanges` can run before/after   `testAppUpdate_dontInterfereLanguageChanges`. 

Now, `testLanguageChanges` [assumes](https://github.com/dimagi/commcare-android/pull/2382/files#diff-df29b73d980cb2f4a06e71e55835a33ddaa78c83b372c5a1e06f26719a6ce217R45) that the device language is English to begin with. 
And `testAppUpdate_dontInterfereLanguageChanges` changes the device language to Hindi when it finishes. 

So if, `testAppUpdate_dontInterfereLanguageChanges` runs before `testLanguageChanges` then `testLanguageChanges` would fail because `"Enter a name:"` text would've been changed.

shubham1g5: Got it. thanks for the explanation. If you are looking to abstract this better you can choose to pass a `force` flag to the `BaseTest::installApp` method that if true clears the app even if it's already installed. That way we won't need to write this logic again and again for any tests that need this similar setup.

dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462077/)

shubham1g5 (approved):
wonderful!

PR #2385, opened 2020-10-20
[2.48.11] Delete raw images from Saved Forms 

shubham1g5:
* Deletes raw folder for saved forms to conserve space for LTS users
* Report free disk available as a user property to Analytics events

### DIFF
--- a/app/src/org/commcare/tasks/CleanRawMedia.kt
+++ b/app/src/org/commcare/tasks/CleanRawMedia.kt
@@ -6,24 +6,25 @@ import androidx.work.WorkerParameters
 import org.commcare.CommCareApplication
 import org.commcare.android.database.user.models.FormRecord
 import org.commcare.utils.CrashUtil
+import org.commcare.utils.StorageUtils.getUnsentCompleteOrSavedFormIdsForCurrentApp
 import java.io.File
-import java.lang.Exception
 
 class CleanRawMedia(appContext: Context, workerParams: WorkerParameters)
     : Worker(appContext, workerParams) {
 
     override fun doWork(): Result {
-        try {
-            val storage = CommCareApplication.instance().getUserStorage(FormRecord::class.java)
-            for (formRecord in storage) {
-                File(formRecord.filePath).parent?.let {
+        val storage = CommCareApplication.instance().getUserStorage(FormRecord::class.java)
+        val recordsToRemove = getUnsentCompleteOrSavedFormIdsForCurrentApp(storage)
+        for (formId in recordsToRemove) {
+            try {
+                val filePath = storage.getMetaDataFieldForRecord(formId, FormRecord.META_FILE_PATH)
+                File(filePath).parent?.let {
                     val rawDirPath = "$it/raw"
                     File(rawDirPath).deleteRecursively()
                 }
+            } catch (e: Exception) {
+                CrashUtil.reportException(e)
             }
-        } catch (e: Exception) {
-            CrashUtil.reportException(e)
-            return Result.failure()
         }
         return Result.success()
     }
--- a/app/src/org/commcare/tasks/CleanRawMedia.kt
+++ b/app/src/org/commcare/tasks/CleanRawMedia.kt
@@ -1,0 +1,30 @@
+package org.commcare.tasks
+
+import android.content.Context
+import androidx.work.Worker
+import androidx.work.WorkerParameters
+import org.commcare.CommCareApplication
+import org.commcare.android.database.user.models.FormRecord
+import org.commcare.utils.CrashUtil
+import java.io.File
+import java.lang.Exception
+
+class CleanRawMedia(appContext: Context, workerParams: WorkerParameters)
+    : Worker(appContext, workerParams) {
+
+    override fun doWork(): Result {
+        try {
+            val storage = CommCareApplication.instance().getUserStorage(FormRecord::class.java)
+            for (formRecord in storage) {
+                File(formRecord.filePath).parent?.let {
+                    val rawDirPath = "$it/raw"
+                    File(rawDirPath).deleteRecursively()
+                }
+            }
+        } catch (e: Exception) {
+            CrashUtil.reportException(e)
+            return Result.failure()
+        }
+        return Result.success()
+    }
+}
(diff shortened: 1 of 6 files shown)


app/src/org/commcare/tasks/CleanRawMedia.kt
```diff
@@ -0,0 +1,30 @@
...
+import java.io.File
+import java.lang.Exception
+
+class CleanRawMedia(appContext: Context, workerParams: WorkerParameters)
+    : Worker(appContext, workerParams) {
+
+    override fun doWork(): Result {
+        try {
+            val storage = CommCareApplication.instance().getUserStorage(FormRecord::class.java)
+            for (formRecord in storage) {
```
ctsims, line 18: I think we should only be doing this for Forms in a specific state, right? It doesn't seem correct to me for this to delete the raw storage for, say, a form that's currently in progress. I wonder if it's safest to, say, only do this for forms that were saved more than a few days ago, too, but that's probably overly conservative

app/src/org/commcare/tasks/CleanRawMedia.kt
```diff
@@ -0,0 +1,30 @@
...
+                File(formRecord.filePath).parent?.let {
+                    val rawDirPath = "$it/raw"
+                    File(rawDirPath).deleteRecursively()
+                }
+            }
+        } catch (e: Exception) {
```
ctsims, line 24: Do we want to fail the loop for one failed record? Given what this task is doing I'd guess we should report each exception (And return failure for the task), but proceed with the next record. Otherwise one bad record will prevent cleaning up the remaining ones

ctsims (changes requested):

app/src/org/commcare/tasks/CleanRawMedia.kt
shubham1g5: Yeah thats a good point. Though given on some devices users face storage issues even in normal conditions, I am inclined to remove it for all completely saved forms. I don't see value in retaining the original images for any forms which are not in progress or incomplete.

app/src/org/commcare/tasks/CleanRawMedia.kt
shubham1g5: Yeah totally agree. Will add this.

app/src/org/commcare/tasks/CleanRawMedia.kt
```diff
@@ -0,0 +1,31 @@
...
+                    val rawDirPath = "$it/raw"
+                    File(rawDirPath).deleteRecursively()
+                }
+            } catch (e: Exception) {
+                CrashUtil.reportException(e)
+            }
+        }
+        return Result.success()
+    }
+}
```
ctsims, line 31: NIT: Missing endline

app/src/org/commcare/tasks/CleanRawMedia.kt
```diff
@@ -0,0 +1,31 @@
...
+        val recordsToRemove = getUnsentCompleteOrSavedFormIdsForCurrentApp(storage)
+        for (formId in recordsToRemove) {
+            try {
+                val filePath = storage.getMetaDataFieldForRecord(formId, FormRecord.META_FILE_PATH)
+                File(filePath).parent?.let {
+                    val rawDirPath = "$it/raw"
+                    File(rawDirPath).deleteRecursively()
+                }
+            } catch (e: Exception) {
+                CrashUtil.reportException(e)
```
ctsims, line 26: I think we should fail at the bottom of the loop still if one iteration didn't process right? That will let it retry later, although if we are launching at-most-one of these per startup maybe it's not worth worrying about the distinciton.

app/src/org/commcare/tasks/CleanRawMedia.kt
```diff
@@ -0,0 +1,31 @@
...
+    : Worker(appContext, workerParams) {
+
+    override fun doWork(): Result {
+        val storage = CommCareApplication.instance().getUserStorage(FormRecord::class.java)
+        val recordsToRemove = getUnsentCompleteOrSavedFormIdsForCurrentApp(storage)
+        for (formId in recordsToRemove) {
+            try {
+                val filePath = storage.getMetaDataFieldForRecord(formId, FormRecord.META_FILE_PATH)
+                File(filePath).parent?.let {
+                    val rawDirPath = "$it/raw"
```
ctsims, line 22: Probably best for us to encode this in a static with the other part of the code, to make sure if a change is made that someone knows it affects this cleanup as well.

ctsims (approved):
Thanks for the quick review. A few points of optional feedback

app/src/org/commcare/tasks/CleanRawMedia.kt
shubham1g5: We are launching these at-most-once on a user login, though still think that it's not worth the distinction here as it's not crticial to keep retrying for a few failed records.

app/src/org/commcare/tasks/CleanRawMedia.kt
shubham1g5: agree though keeping it unchanged for now to avoid another review cycle and  I will implement the suggestion in the `master` version of this PR>

app/src/org/commcare/tasks/CleanRawMedia.kt
shubham1g5: Ahh I am not sure what are you pointing at here ? Is it nicer to give a line break at the end of file ?

shubham1g5: Found an explanation [here](https://stackoverflow.com/a/729795/3811963), so sorted on this. Configured my editor to do this in future for all files.

ShivamPokhriyal: Wow, I didn't know about this. Will configure my editor for the same.

shubham1g5:
@ctsims  Do you have thoughts on how we would want to adopt this approach more largely across the platform. One option is to check for available space and if it's lower than a certain threshold perform the clear.

shubham1g5:
Great, that sounds like a good and simple solution. FormEntryActivity is [already handling](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/views/widgets/ImageWidget.java#L194) the case to show the alternate image when original is not present.

PR #2388, opened 2020-10-22
Add Tiles test

ShivamPokhriyal:
#2341


ShivamPokhriyal:

dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462081/)

app/instrumentation-tests/src/org/commcare/utils/Array+Extension.kt
```diff
@@ -0,0 +1,15 @@
+package org.commcare.utils
+
+import androidx.test.espresso.matcher.ViewMatchers.withText
+
+fun Array<String>.areDisplayed() {
```
shubham1g5, line 5: Does these make more sense as the Matcher Extension rather than an array extension ? For eg. defining a `isArrayDisplayed(items: array)` as a matcher extension. I think so because the primary purpose of these methods is to match the text in arrays and I would classify these methods to be part of Matcher rather than Array, but would like to know if there is a particular benefit of keeping them as Array extensions ?

app/instrumentation-tests/src/org/commcare/androidTests/TilesTest.kt
```diff
@@ -0,0 +1,183 @@
...
+
+    @Before
+    fun setup() {
+        installApp(APP_NAME, CCZ_NAME)
+        InstrumentationUtility.login("tile.test", "123")
+    }
+
+    @Test
+    fun testNoTile() {
+        InstrumentationUtility.openModule(0)
```
shubham1g5, line 37: Wondering if these methods should be based on the text of the module/form rather than position ? It's more clear to anyone reading code that way.

app/instrumentation-tests/src/org/commcare/androidTests/TilesTest.kt
```diff
@@ -0,0 +1,183 @@
...
+        caseDetails.areDisplayed()
+
+        onView(withText("Sally Ride"))
+                .perform(click())
+        withId(R.id.com_tile_holder_btn_open).doesNotExist()
+        withText("Sex F").doesNotExist()
+
+        onView(withText("Continue"))
+                .perform(click())
+        onView(withText("Placeholder"))
```
shubham1g5, line 48: [Step](https://github.com/dimagi/commcare-ui-tests/blob/master/features/tiles.feature#L25) missing

app/instrumentation-tests/src/org/commcare/androidTests/TilesTest.kt
```diff
@@ -0,0 +1,183 @@
...
+        // Expand tile
+        onView(withId(R.id.com_tile_holder_btn_open))
+                .perform(click())
+        caseDetailsExpanded.areDisplayed()
+
+        //Close expanded list
+        onView(withId(R.id.com_tile_holder_btn_open))
+                .perform(click())
+        caseDetails.areDisplayed()
+        withText("Secret").doesNotExist()
```
shubham1g5, line 112: [Missing](https://github.com/dimagi/commcare-ui-tests/blob/master/features/tiles.feature#L115) ?

app/instrumentation-tests/src/org/commcare/androidTests/TilesTest.kt
```diff
@@ -0,0 +1,183 @@
...
+        caseDetails.areDisplayed()
+        withText("Placeholder").isDisplayed()
+        withId(R.id.com_tile_holder_btn_open).isNotDisplayed()
+
+        // Open form
+        onView(withText("Placeholder"))
+                .perform(click())
+
+        // We can still see the case details.
+        caseDetails.areDisplayed()
```
shubham1g5, line 139: [Missing](https://github.com/dimagi/commcare-ui-tests/blob/master/features/tiles.feature#L143) ?

app/instrumentation-tests/src/org/commcare/androidTests/TilesTest.kt
```diff
@@ -0,0 +1,183 @@
...
+        withText("Placeholder").isDisplayed()
+        withText("Secret").doesNotExist()
+        withId(R.id.com_tile_holder_btn_open).isNotDisplayed()
+
+        // go to form and you can still see the case details.
+        onView(withText("Placeholder"))
+                .perform(click())
+
+        caseDetails.areDisplayed()
```
shubham1g5, line 158: [Step missing](https://github.com/dimagi/commcare-ui-tests/blob/master/features/tiles.feature#L167)

app/instrumentation-tests/src/org/commcare/androidTests/TilesTest.kt
```diff
@@ -0,0 +1,183 @@
...
+        withText("Secret").doesNotExist()
+        withId(R.id.com_tile_holder_btn_open).isNotDisplayed()
+    }
+
+    @Test
+    fun testBreadcrumb() {
```
shubham1g5, line 164: [This](https://github.com/dimagi/commcare-ui-tests/blob/master/features/tiles.feature#L184) and [that](https://github.com/dimagi/commcare-ui-tests/blob/master/features/tiles.feature#L194) are not present

shubham1g5 (changes requested):

app/instrumentation-tests/src/org/commcare/androidTests/TilesTest.kt
ShivamPokhriyal: Yeah makes sense. Will update it.

app/instrumentation-tests/src/org/commcare/androidTests/TilesTest.kt
ShivamPokhriyal: Thanks for pointing it.

app/instrumentation-tests/src/org/commcare/utils/Array+Extension.kt
ShivamPokhriyal: I don't think `isArrayDisplayed(items: array)` can be used as an extension. It'll be more like a utility method where we'll have to call it `InstrumentationUtiility.isArrayDisplayed(items)`. 
I wanted to make use of extension here merely because of the way it will be used. Like we can simply call `items.areDisplayed()` and that's it. And it is kinda more readable this way.

shubham1g5: Makes sense. Thanks for the explanation.

shubham1g5 (approved):

dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462084/)

PR #2396, opened 2020-10-29
Add manual quarantine tests

ShivamPokhriyal:


ShivamPokhriyal:
Looks like it's failing because of the `changeWifi` issue in browserstack. I'll keep this PR as open, till I find a workaround for that issue.

shubham1g5:
@ShivamPokhriyal wondering if we can add `@supress` to it so that it gets skipped on BS and then implement another annotation say "@WifiDependent" and have a mechanism to run all these wifi dependent tests locally before every release.

ShivamPokhriyal:
> @ShivamPokhriyal wondering if we can add `@supress` to it so that it gets skipped on BS and then implement another annotation say "@WifiDependent" and have a mechanism to run all these wifi dependent tests locally before every release.

Sounds really great. Will follow up on it in another PR.

app/instrumentation-tests/src/org/commcare/androidTests/ManualQuarantineTest.kt
```diff
@@ -0,0 +1,126 @@
...
+        InstrumentationUtility.gotoHome()
+        InstrumentationUtility.selectOptionItem(withText("Advanced"))
+        onView(withText("Enable Manual Form Quarantine"))
+                .perform(click())
+        InstrumentationUtility.gotoHome()
+    }
+
+    @Test
+    fun test_A_Quarantine() {
+        InstrumentationUtility.openModule(12)
```
shubham1g5, line 54: nit: if we are using position as input to `openModule`, we should put a comment here specifiying the text of module we want to open for better readability or move the value `12` in a constant whose name indicates what module this position is for.

app/instrumentation-tests/src/org/commcare/androidTests/ManualQuarantineTest.kt
```diff
@@ -0,0 +1,126 @@
...
+        )).perform(longClick())
+        // Unsent forms should not be delete-able
+        withText("Delete Record").doesNotExist()
+        onView(withText("Scan Record Integrity"))
+                .perform(click())
+        onView(withText("QUARANTINE FORM"))
+                .perform(click())
+
+        // After quarantining one form we'll have to re-enable quarantine setting.
+        enableFormQuarantine()
```
shubham1g5, line 77: we should add an additional step that just checks that you can't quarantine without this call.

app/instrumentation-tests/src/org/commcare/androidTests/ManualQuarantineTest.kt
```diff
@@ -0,0 +1,126 @@
...
+        const val APP_NAME = "Basic Tests"
+    }
+
+    @Before
+    fun setup() {
+        installApp(APP_NAME, CCZ_NAME)
+        InstrumentationUtility.login("test", "1234")
+        // Enable quarantine
+        enableFormQuarantine()
+        InstrumentationUtility.changeWifi(false)
```
shubham1g5, line 41: Need to suppress below Android 10

shubham1g5 (changes requested):

ShivamPokhriyal:

dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462111/)

ShivamPokhriyal:
@shubham1g5 Another review please.

shubham1g5 (approved):
Looks great.

PR #2400, opened 2020-11-03
Hacky fix for edittext focus issue

ShivamPokhriyal:
More information here: https://dimagi-dev.atlassian.net/browse/SAAS-11370

Product Note: Fixes a bug in the view hierarchy where in making any change to a question in the questionList will cause the screen to scroll up to the first edittext that has focus.

### DIFF
--- a/app/src/org/commcare/views/ClearFocusEditText.java
+++ b/app/src/org/commcare/views/ClearFocusEditText.java
@@ -1,0 +1,63 @@
+package org.commcare.views;
+
+import android.content.Context;
+import android.os.Build;
+import android.util.AttributeSet;
+import android.view.KeyEvent;
+import android.widget.EditText;
+
+import androidx.annotation.RequiresApi;
+
+/**
+ * @author $|-|!˅@M
+ */
+public class ClearFocusEditText extends EditText {
+
+    /**
+     * A boolean value to keep track of whether we explicitly want this editText to remove focus.
+     */
+    boolean shouldRemoveFocus = false;
+
+    public ClearFocusEditText(Context context) {
+        super(context);
+    }
+
+    public ClearFocusEditText(Context context, AttributeSet attrs) {
+        super(context, attrs);
+    }
+
+    public ClearFocusEditText(Context context, AttributeSet attrs, int defStyleAttr) {
+        super(context, attrs, defStyleAttr);
+    }
+
+    @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
+    public ClearFocusEditText(Context context, AttributeSet attrs, int defStyleAttr, int defStyleRes) {
+        super(context, attrs, defStyleAttr, defStyleRes);
+    }
+
+    public void acceptFocus() {
+        shouldRemoveFocus = false;
+    }
+
+    @Override
+    public boolean onKeyPreIme(int keyCode, KeyEvent event) {
+        if (keyCode == KeyEvent.KEYCODE_BACK) {
+            /*
+            this.clearFocus() works in Android 10, and it should've worked in all the android versions.
+            But in Android 6 it doesn't, clearFocus() internally propagates the clearFocus call up the parent hierarchy.
+            And when it does so, the framework tries to give focus to the first focusable view from the top automatically.
+             */
+            shouldRemoveFocus = true;
+            this.clearFocus();
+        }
+        return super.onKeyPreIme(keyCode, event);
+    }
+
+    @Override
+    public boolean isFocused() {
+        if (shouldRemoveFocus) {
+            return false;
+        }
+        return super.isFocused();
+    }
+}
--- a/app/src/org/commcare/views/widgets/StringWidget.java
+++ b/app/src/org/commcare/views/widgets/StringWidget.java
@@ -19,6 +19,7 @@ import android.view.inputmethod.InputMethodManager;
 import android.widget.EditText;
 
 import org.commcare.dalvik.R;
+import org.commcare.views.ClearFocusEditText;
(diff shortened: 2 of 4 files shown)


dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462106/)

app/src/org/commcare/views/ClearFocusEditText.java
```diff
@@ -0,0 +1,63 @@
...
+        shouldRemoveFocus = false;
+    }
+
+    @Override
+    public boolean onKeyPreIme(int keyCode, KeyEvent event) {
+        if (keyCode == KeyEvent.KEYCODE_BACK) {
+            /*
+            this.clearFocus() works in Android 10, and it should've worked in all the android versions.
+            But in Android 6 it doesn't, clearFocus() internally propagates the clearFocus call up the parent hierarchy.
+            And when it does so, the framework tries to give focus to the first focusable view from the top automatically.
```
shubham1g5, line 48: It's unclear to me how does this PR solve for Android 6 behaviour to give focus to the first focusable view from the top and hence is solving this issue. Can you maybe add more details in the PR description on how this approach solves this issue.

shubham1g5:
@ShivamPokhriyal  I think it's critical for us that we determine the right reproduction path for this issue in order to understand better what exactly is happening here. There is also a similar issue mentioned in the [PR here ](https://github.com/dimagi/commcare-android/pull/2120) and it would be good to crosscheck whether the underlying issue is same in both tickets or not.

ShivamPokhriyal:
> @ShivamPokhriyal I think it's critical for us that we determine the right reproduction path for this issue in order to understand better what exactly is happening here. There is also a similar issue mentioned in the [PR here ](https://github.com/dimagi/commcare-android/pull/2120) and it would be good to crosscheck whether the underlying issue is same in both tickets or not.

Thanks for the link. And I think it's most probably related to the same issue you fixed earlier. Which also explains why I was not able to reproduce it using a sample app.

shubham1g5:
>  And I think it's most probably related to the same issue you fixed earlier

Ahh the PR I linked was closed without merge due to some issues. So that issue is still there.

app/src/org/commcare/views/ClearFocusEditText.java
ShivamPokhriyal: Sorry, I just realised that this comment was not sent.

What I'm achieving here is that `isFocused` method would return `false` when we want it to. So when android system will check in the layout whether the textview is focused or not, this method will be invoked and it will return false. So no scrolling will happen.

ShivamPokhriyal:
@shubham1g5 I used your [sample app](https://www.commcarehq.org/a/shubhamgoyaltest/apps/view/a0329273b7139821552d0aa1c7f0682a/releases/) but I was not able to repro it. I also created a sample [here](https://www.commcarehq.org/a/srishti-1/apps/view/a5f996edd10a5d5081d25702faab89e5/form/17fed93914db4ef798b744fb84611bf7/source/#form/question_list) and I'm not able to repro it using this sample as well. Are you able to repro it in any of these forms? If yes, what's the android version you're using?

shubham1g5:
Hey I am able to repro with the [version 12](https://www.commcarehq.org/a/shubhamgoyaltest/apps/view/a0329273b7139821552d0aa1c7f0682a/releases/) of the app on Android 8. Here are the steps I followed - 
1.  Open the form and click on Yes in MC1
2. Touch the Focusable question so that it gets focus and keyboard opens up 
3. while the keyboard is still open, scroll to bottom and touch any option in MC5 
4. Notice the keyboard close and jump to the Focusable question

ShivamPokhriyal:
> Hey I am able to repro with the [version 12](https://www.commcarehq.org/a/shubhamgoyaltest/apps/view/a0329273b7139821552d0aa1c7f0682a/releases/) of the app on Android 8. Here are the steps I followed -
> 
> 1. Open the form and click on Yes in MC1
> 2. Touch the Focusable question so that it gets focus and keyboard opens up
> 3. while the keyboard is still open, scroll to bottom and touch any option in MC5
> 4. Notice the keyboard close and jump to the Focusable question

Ohh yup, that's kinda different scenario and thanks for adding the details. The one that I'm trying to repro here is when keyboard is closed, but the edittext still has focus, then the screen should jump to the Focusable question.

shubham1g5:
got it. In that case we don't need to worry about the other issue. For this one, can you try to separate out the issue by doing a deep dive into why the issue doesn't appear on your sample app while it does for the app mentioned in the linked ticket. It seems like we are missing a critical detail here around the nature of this issue and I would not be confident in rolling out a fix on it without understanding that properly.

app/src/org/commcare/views/widgets/StringWidget.java
```diff
@@ -43,9 +44,13 @@ public StringWidget(Context context, FormEntryPrompt prompt, boolean secret) {
 
     public StringWidget(Context context, FormEntryPrompt prompt, boolean secret, boolean inCompactGroup) {
         super(context, prompt, inCompactGroup);
-        mAnswer = (EditText)LayoutInflater.from(getContext()).inflate(getAnswerLayout(), this, false);
+        mAnswer = (ClearFocusEditText)LayoutInflater.from(getContext()).inflate(getAnswerLayout(), this, false);
         mAnswer.setTextSize(TypedValue.COMPLEX_UNIT_DIP, mAnswerFontSize);
         mAnswer.setOnClickListener(this);
+        mAnswer.setOnTouchListener((v, event) -> {
```
shubham1g5, line 50: what if the view gained focus due to other events like clicking next on Keyboard.

PR #2401, opened 2020-11-05
Fix form jump issues with edittext

ShivamPokhriyal:
Jira: https://dimagi-dev.atlassian.net/browse/SAAS-11370

### Summary
While working in a question list, when a user enters some text in an edittext and then shifts to a non editText question (like RadioButton or DateWidget), then the last changed editText doesn't clear foucs by itself. 
This results into the screen jumping to this last focused EditText when any change is made to any other question. 

This PR will ask the QuestionWidget to remove focus if it's not the last widget user was interacting with. And only StringWidget will respond to the `removeFocus` call. 

### Product Note: 
Fixes a bug in the view hierarchy wherein making any change to a question in the questionList will cause the screen to scroll up to the first edittext that has focus.

### DIFF
--- a/app/src/org/commcare/views/widgets/QuestionWidget.java
+++ b/app/src/org/commcare/views/widgets/QuestionWidget.java
@@ -135,9 +135,6 @@ public abstract class QuestionWidget extends LinearLayout implements QuestionExt
     protected void acceptFocus() {
     }
 
-    public void removeFocus() {
-    }
-
     private void addHelpPlaceholder() {
         if (!mPrompt.hasHelp()) {
             return;
--- a/app/src/org/commcare/views/widgets/QuestionWidget.java
+++ b/app/src/org/commcare/views/widgets/QuestionWidget.java
@@ -135,6 +135,9 @@ public abstract class QuestionWidget extends LinearLayout implements QuestionExt
     protected void acceptFocus() {
     }
 
+    public void removeFocus() {
+    }
+
     private void addHelpPlaceholder() {
         if (!mPrompt.hasHelp()) {
             return;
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/views/widgets/QuestionWidget.java
```diff
@@ -135,6 +135,9 @@ public QuestionWidget(Context context, FormEntryPrompt p, boolean inCompactGroup
     protected void acceptFocus() {
     }
 
+    public void removeFocus() {
```
shubham1g5, line 138: Since this is only ever meant for `StringWidget` we should not include it here. Instead when you are calling this methos, you can do a type check and cast to call this only on `StringWidget`

shubham1g5 (changes requested):
One design feedback, though we should make sure that we test comprehensively for any side effects of this change. Clayton has some advice on how to do that here - https://github.com/dimagi/commcare-android/pull/2120#issuecomment-503353155

ShivamPokhriyal:
> One design feedback, though we should make sure that we test comprehensively for any side effects of this change. Clayton has some advice on how to do that here - [#2120 (comment)](https://github.com/dimagi/commcare-android/pull/2120#issuecomment-503353155)

Sounds good, I'll do that test myself, and would also flag it to QA before releasing.

shubham1g5 (approved):

dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462113/)

dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462114/)

dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462118/)

ShivamPokhriyal:
Tested with [this app](https://www.commcarehq.org/a/srishti-1/apps/view/a5f996edd10a5d5081d25702faab89e5/form/17fed93914db4ef798b744fb84611bf7/source/) and it's working fine.

Issue #2377, opened 2020-10-13
Allow the use of a custom Barcode Scanning Application


ctsims:
Right now we solely use the embedded zxing library, but there are circumstances where users might want to jump out to a much more powerful version that's more appropriate for their device or use case.

The need for this would be negated by implementing https://github.com/dimagi/commcare-android/issues/2321, but that's potentially a significantly larger scope of work compared to a fallback option for telling the barcode scanner to use the traditional zxing intent, rather than the built-in one. 

This seems to me like an App setting rather than a callout specific setting, so I think the best course of action would be to make it an [application preference](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/preferences/MainConfigurablePreferences.java) ("Use intent callout for scanner") that is loaded in the FormEntryActivity and passed into the [widget factory](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/views/widgets/WidgetFactory.java#L121) .

shubham1g5:
@ShivamPokhriyal this would be good for us to prioritize on SAAS for 2.51 . Feel free to drop any questions here.

PR #2405, opened 2020-11-12
Allow use of custom barcode scanner apps

ShivamPokhriyal:
#2377 
Product Note: Added a setting to allow users to use custom applications for scanning barcodes. 

[Video](https://drive.google.com/drive/folders/1cnVOzjFcycAM0mD1VwfkPr51HkVoGEqf?usp=sharing)

### DIFF
--- a/app/src/org/commcare/views/widgets/WidgetUtils.java
+++ b/app/src/org/commcare/views/widgets/WidgetUtils.java
@@ -1,6 +1,7 @@
 package org.commcare.views.widgets;
 
 import android.content.Context;
+import android.content.Intent;
 import android.text.Spannable;
 import android.util.TypedValue;
 import android.view.LayoutInflater;
@@ -8,9 +9,15 @@ import android.view.ViewGroup;
 import android.widget.Button;
 import android.widget.TableLayout;
 
-import org.commcare.dalvik.R;
+import androidx.appcompat.app.AppCompatActivity;
 
-class WidgetUtils {
+import com.google.zxing.client.android.Intents;
+import com.google.zxing.integration.android.IntentIntegrator;
+
+import org.commcare.dalvik.R;
+import org.commcare.preferences.MainConfigurablePreferences;
+
+public class WidgetUtils {
     private static final TableLayout.LayoutParams params;
 
     static {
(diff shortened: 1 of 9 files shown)


shubham1g5 (changes requested):
@ShivamPokhriyal  your approach looks good to me , though there are some more places in the code where we are using the Zxing library as default and we should make similar changes there as well

ShivamPokhriyal:
> @ShivamPokhriyal your approach looks good to me , though there are some more places in the code where we are using the Zxing library as default and we should make similar changes there as well

Got it, I'll check the other places as well.

app/src/org/commcare/views/widgets/WidgetUtils.java
```diff
@@ -33,4 +40,42 @@ public static Button setupClearButton(Context context, ViewGroup root, String te
...
+     * app exists to handle such intent.
+     *
+     * @param formats names of {@link com.google.zxing.BarcodeFormat}s to scan for. eg. {@link com.google.zxing.BarcodeFormat#QR_CODE}
+     */
+    public static Intent createScanIntent(Context context, String... formats) {
+        if (MainConfigurablePreferences.useIntentCalloutForScanner()) {
+            Intent scanIntent = new Intent(Intents.Scan.ACTION);
+            if (scanIntent.resolveActivity(context.getPackageManager()) != null) {
+                if (formats != null) {
+                    StringBuilder joinedByComma = new StringBuilder();
```
shubham1g5, line 65: can use [String.join](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#join-java.lang.CharSequence-java.lang.CharSequence...-)

shubham1g5 (changes requested):
one minor feedback

app/src/org/commcare/views/widgets/WidgetUtils.java
ShivamPokhriyal: Made a change to use `TextUtils` since [String.join](https://developer.android.com/reference/java/lang/String#join(java.lang.CharSequence,%20java.lang.CharSequence...)) is added on api 26

shubham1g5 (dismissed):

dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462135/)

dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462146/)

shubham1g5 (approved):

PR #2407, opened 2020-11-16
Support for CommCare Identity Hooks

shubham1g5:
* Merges CommCare Support Lib in CommCare Android 
* Adds support for [Commcare Identity Hooks](https://github.com/dimagi/commcare-android/blob/supportLibMerge/commcare-support-library/src/main/java/org/commcare/commcaresupportlibrary/identity/identity_integration.md) to support a generalized Identity Integration

Product Note: Adds support for out of box callout based integration with third party Identity Providers


ShivamPokhriyal:
> Merges CommCare Support Lib in CommCare Android

Why are we merging and not using it as a dependency?

shubham1g5:
> Why are we merging and not using it as a dependency?

It's still used as a dependency. What I meant by merge is instead of having the support lib in a separate project, we have it as a module inside CommCare Android. It makes sense to have them under same project as both these projects are tightly coupled with one another and any compatibility errors due to a change in one of the project are more easily surfaced. For eg. - the project won't build if I change the contract for one of the Identity hooks only in support lib and not in Android. Reverts are also easy since changes are more easily put in a single PR

ShivamPokhriyal:
Ohh got it. Sorry, I didn't see the changes yet. 

> we have it as a module inside CommCare Android.

So does this mean that we no longer need to maintain a separate repo for `commcare-support-library`? And we can publish this module to bintray or jitpack whenever there are changes.

shubham1g5:
> So does this mean that we no longer need to maintain a separate repo for commcare-support-library? And we can publish this module to bintray or jitpack whenever there are changes.

Yeah, that's right. The publishing part is manual though for now (off Jenkins) and reuires one to run the publishing gradle task once changes are finalized.

ShivamPokhriyal:
> Yeah, that's right. The publishing part is manual though for now (off Jenkins) and reuires one to run the publishing gradle task once changes are finalized.

Awesome! Can you also create a doc about how to publish the lib, if you haven't already.

app/src/org/commcare/provider/IdentityCalloutHandler.java
```diff
@@ -0,0 +1,174 @@
...
+    private static final String REF_GUID = "guid";
+    private static final String REF_CONFIDENCE = "confidence";
+    private static final String REF_MATCH_STRENGTH = "match_strength";
+
+    public static final String GENERALIZED_IDENTITY_PROVIDER = "generalized_identity_provider";
+
+
+    @StringDef({GENERALIZED_IDENTITY_PROVIDER, SimprintsCalloutProcessing.SIMPRINTS_IDENTITY_PROVIDER})
+    @Retention(RetentionPolicy.SOURCE)
+    public @interface IdentityProvider {
```
ctsims, line 42: Is there still some way to specifically choose each identity provider (with, say, an ID the same way that SimPrints works?)?

I doubt there will often be more than one in use at a time, but I imagine inside of an app, when people configure "Their identify provider" they may want to choose a specific one.

app/src/org/commcare/provider/IdentityCalloutHandler.java
```diff
@@ -0,0 +1,174 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.util.Collections;
+import java.util.Hashtable;
+import java.util.List;
+import java.util.Vector;
+
+import androidx.annotation.StringDef;
+
+public class IdentityCalloutHandler {
```
ctsims, line 30: Is there something unique to how this app works? The pattern of having the class itself consist ~entirely of static members makes it seem like it's doing something specific that I don't quite understand, but I could also see this just being a set of helper methods.

commcare-support-library/src/main/java/org/commcare/commcaresupportlibrary/CaseUtils.java
```diff
@@ -0,0 +1,252 @@
+package org.commcare.commcaresupportlibrary;
```
ctsims, line 1: I'm presuming these were copied over ~identically from the support library and don't need new review, is that assumption correct?

app/src/org/commcare/activities/EntitySelectActivity.java
```diff
@@ -664,9 +669,16 @@ private void handleSearchStringCallout(Intent intent) {
         }
     }
 
-    private void handleFingerprintMatchCallout(Intent intent) {
-        OrderedHashtable<String, String> guidToMatchConfidenceMap =
-                SimprintsCalloutProcessing.getConfidenceMatchesFromCalloutResponse(intent);
+    private void handleFingerprintMatchCallout(
+            Intent intent,
+            @IdentityCalloutHandler.IdentityProvider String identityProvider) {
```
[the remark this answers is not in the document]
shubham1g5: This is just a more memory-efficient replacement for Enums on Android. [This](https://guides.codepath.com/android/Replacing-Enums-with-Enumerated-Annotations) is a quick gide on this in case you are interested.

app/src/org/commcare/provider/IdentityCalloutHandler.java
shubham1g5: Choice of Identity provider totally depends on the callout action the app defines i.e. "org.simprint.action" vs something else. The Identity provider scpecific handling in the app only applies on the Intent reponse handling which we do by checking the format of the response in intent like [here](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/provider/SimprintsCalloutProcessing.java#L41).

app/src/org/commcare/provider/IdentityCalloutHandler.java
shubham1g5: I was trying to mimic the `SimprintsCalloutProcessing` here. Both of these can potentially implement an Identity Provider interface though we won't be able to define all these methods as `static` in that case. Would that be ok ? Not sure if `static` methods are of a particular advantage here.

commcare-support-library/src/main/java/org/commcare/commcaresupportlibrary/CaseUtils.java
shubham1g5: yes. It's an exact copy so no need to review.

shubham1g5: well one [minor change](https://github.com/dimagi/commcare-android/pull/2407/commits/b57827b47f104aa5e2db3bb5c5cadde36d0cad04) that I had to do is target the same android versions as `commcare-android` in order to build successfully and avoid some resource conflicts.

shubham1g5:

shubham1g5:

shubham1g5:
@ctsims  Made some further changes to support duplicate encountered during registration over [here](https://github.com/dimagi/commcare-android/pull/2407/commits/776e6728f0543a1321e7592e492ecaa62aa04c86). I am taking the best match and making it available in the form, not sure if there is a way by which we can handle multiple duplicates passed back from Identity Provider and make them all available in the form.

dimagimon:
[33% (0.0%) vs master 33%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462144/)

shubham1g5:

dimagimon:
[33% (0.0%) vs master 33%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462151/)

dimagimon:
[33% (0.0%) vs master 33%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462184/)

dimagimon:
[33% (0.0%) vs master 33%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462190/)

dimagimon:
[33% (0.0%) vs master 33%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462198/)

shubham1g5:
@ctsims Bumping this for your review

ctsims (approved):
Looking good to me, thanks for all of the different passes on this!

PR #2408, opened 2020-11-18
Optimize android instrumentation test time

ShivamPokhriyal:
Using auto shard strategy from this doc https://www.browserstack.com/docs/app-automate/espresso/test-sharding
Since we've a limit of 5 parallel tests, so I've configured the `numberOfShards` to 5.

### DIFF
--- a/app/instrumentation-tests/src/org/commcare/androidTests/MenuTests.kt
+++ b/app/instrumentation-tests/src/org/commcare/androidTests/MenuTests.kt
@@ -24,7 +24,6 @@ import org.junit.runner.RunWith
 
 @RunWith(AndroidJUnit4::class)
 @LargeTest
-@BrowserstackTests
 class MenuTests: BaseTest() {
 
     companion object {
@@ -55,6 +54,7 @@ class MenuTests: BaseTest() {
     }
 
     @Test
+    @BrowserstackTests
     fun testHomeScreenOptions() {
         InstrumentationUtility.openOptionsMenu()
         checkStringExists(homeMenuItems)
@@ -123,6 +123,7 @@ class MenuTests: BaseTest() {
     }
 
     @Test
+    @BrowserstackTests
     fun testAdvancedActions() {
         openAdvancedOption(0)
         onView(withText("Do you want to send, receive, or submit forms?"))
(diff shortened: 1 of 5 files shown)


shubham1g5 (dismissed):
awesome

app/instrumentation-tests/src/org/commcare/androidTests/MenuTests.kt
```diff
@@ -24,7 +24,6 @@ import org.junit.runner.RunWith
 
 @RunWith(AndroidJUnit4::class)
 @LargeTest
-@BrowserstackTests
```
shubham1g5, line 27: annotation on class doesn't work ?

shubham1g5 (approved):

app/instrumentation-tests/src/org/commcare/androidTests/MenuTests.kt
ShivamPokhriyal: It works. `BrowserstackTests` means that the tests will run on browserstack. and we don't want to run all the tests from this class on browserstack. So I've annotated the tests which need to run there.

shubham1g5: Ahh I completely misunderstood the intent here. Thanks for clarification.

shubham1g5:

ShivamPokhriyal:
I guess we have some tests `CaseSharingTest` which needs to run in order. And with shard enabled, these tests will run on different devices and would fail. 
So perhaps we can use one shard to run these sorts of tests(maybe by using annotation or specifying the class directly).

shubham1g5:
I think 2 diiferent tests should be independent of each other and if they are dependent they should be merged into a single test.

ShivamPokhriyal:
> I think 2 diiferent tests should be independent of each other and if they are dependent they should be merged into a single test.

Yeah, I wasn't really sure about that and they're 5 methods that need to be run in specific order. So, I'll merge them into a big one then.

dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462147/)

shubham1g5 (approved):

shubham1g5:
Nice this last build was just ~25 mins.

PR #2409, opened 2020-11-19
[2.50.3] Lets Encrypt Root CA Change

shubham1g5:
https://letsencrypt.org/2020/11/06/own-two-feet.html

Solution taken from:
https://stackoverflow.com/questions/64844311/certpathvalidatorexception-connecting-to-a-lets-encrypt-host-on-android-m-or-ea

OkHttp Issue: https://github.com/square/okhttp/issues/6403


ShivamPokhriyal:
Great to see this PR.

app/src/org/commcare/CommCareApplication.java
```diff
@@ -244,6 +246,10 @@ public void onCreate() {
         LocalePreferences.saveDeviceLocale(Locale.getDefault());
     }
 
+    private void attachISRGCert() {
+        CommCareNetworkServiceGenerator.customizeRetrofitSetup(new ISRGCertConfig());
```
ShivamPokhriyal, line 250: One question: Should we be adding this only below `Android 7.1`?

ShivamPokhriyal: Ohh nvm, just saw the check here https://github.com/dimagi/commcare-android/pull/2409/files#diff-accec8e023ea67a6abb17d85e70b38eb6f6a80aa1ead1bb96c638f81b3c2f0eaR31

app/src/org/commcare/network/ISRGCertConfig.java
```diff
@@ -0,0 +1,81 @@
...
+                            "mRGunUHBcnWEvgJBQl9nJEiU0Zsnvgc/ubhPgXRR4Xq37Z0j4r7g1SgEEzwxA57d\n" +
+                            "emyPxgcYxn/eR44/KJ4EBs+lVDR3veyJm+kXQ99b21/+jh5Xos1AnX5iItreGCc=\n" +
+                            "-----END CERTIFICATE-----";
+
+            try {
+                CertificateFactory cf = CertificateFactory.getInstance("X.509");
+                Certificate isgCertificate = cf.generateCertificate(new ByteArrayInputStream(isgCert.getBytes("UTF-8")));
+
+                HandshakeCertificates certificates = new HandshakeCertificates.Builder()
+                        .addTrustedCertificate((X509Certificate)isgCertificate)
```
yschimke, line 70: Do you need to support other platform certificates as well? Or just an LE root?

Can you confirm this already works for existing LE certificates?

shubham1g5: > Do you need to support other platform certificates as well?

Uh yes, It's failing for other non LE certs. so seems like we need to add `addPlatformTrustedCertificates()` to allow for other certificates. Appreciate the catch. 

> Can you confirm this already works for existing LE certificates?

Yeah this somehow works without adding the call to `addPlatformTrustedCertificates()`. Not sure how though!

ctsims:

shubham1g5:
@ctsims the build is going to fail on this as we need to build it against 2.50 core. We will be able to see the build status over [here](https://jenkins.dimagi.com/job/commcare-android-2.50/) once merged.

ShivamPokhriyal (approved):

PR #2411, opened 2020-11-30
Text to speech support

ShivamPokhriyal:
Product Note: Add support for text-to-speech which can be enabled by the app setting `Enable TTS` or by setting `tts` content for the question in the form. 

UI:
Screenshot_2020-12-01-10-37-00-102_org commcare dalvik debug


ShivamPokhriyal:

ShivamPokhriyal:
The last build failure needs a fix in the browserstack script. When the number of tests is less than 5, and we run the tests on 5 shards then it throws an error.

ShivamPokhriyal:

dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462159/)

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -167,6 +171,34 @@
...
+                    FormEntryActivity.this,
+                    Localization.get("tts.data.missing.title"),
+                    Localization.get("tts.data.missing.message", language));
+            dialog.setPositiveButton(Localization.get("dialog.ok"), (dialog1, which) -> {
+                Intent installIntent = new Intent();
+                installIntent.setAction(TextToSpeech.Engine.ACTION_INSTALL_TTS_DATA);
+                startActivity(installIntent);
+            });
+            dialog.setNegativeButton(Localization.get("option.cancel"), (dialog1, which) -> {
+                dismissAlertDialog();
```
shubham1g5, line 197: youc can just call `dialog1.dismiss` instead of re-querying for the dialog.

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -167,6 +171,34 @@
...
+        @Override
+        public void voiceDataMissing(String language) {
+            StandardAlertDialog dialog = new StandardAlertDialog(
+                    FormEntryActivity.this,
+                    Localization.get("tts.data.missing.title"),
+                    Localization.get("tts.data.missing.message", language));
+            dialog.setPositiveButton(Localization.get("dialog.ok"), (dialog1, which) -> {
+                Intent installIntent = new Intent();
+                installIntent.setAction(TextToSpeech.Engine.ACTION_INSTALL_TTS_DATA);
+                startActivity(installIntent);
```
shubham1g5, line 194: should dismiss the dialog once user has taken an action

app/src/org/commcare/tts/TextToSpeechConverter.kt
```diff
@@ -0,0 +1,191 @@
...
+     * Initializes the Text-To-Speech engine
+     */
+    private fun initialize(context: Context, text: String) {
+        mTextToSpeech = TextToSpeech(context, TextToSpeech.OnInitListener { status ->
+            if (status == TextToSpeech.SUCCESS) {
+                // set language and speak
+                mTextToSpeech?.let { tts ->
+                    tts.setOnUtteranceProgressListener(mUtteranceProgressListener)
+                    if (setLocale(tts, LinkedList(listOf(
+                                    Locale(Localization.getCurrentLocale()),
```
shubham1g5, line 123: the locale needs to be reapplied when user changes the language in form and we should also pass the conuntry code from default locale to account for accents.

app/src/org/commcare/tts/TextToSpeechConverter.kt
```diff
@@ -0,0 +1,191 @@
...
+                    if (features == null || features.contains("notInstalled")) {
+                        mTTSCallback?.voiceDataMissing(locale.displayLanguage)
+                        return false
+                    }
+                }
+                // voice data is present so return true that we're ready to use.
+                true
+            }
+            TextToSpeech.LANG_MISSING_DATA -> {
+                // Unfortunately this callback doesn't really work.
```
shubham1g5, line 177: if it does not work at all, is there any sense of having it in code ?

app/src/org/commcare/tts/TextToSpeechConverter.kt
```diff
@@ -0,0 +1,191 @@
...
+        return when (tts.isLanguageAvailable(locale)) {
+            TextToSpeech.LANG_AVAILABLE,
+            TextToSpeech.LANG_COUNTRY_AVAILABLE,
+            TextToSpeech.LANG_COUNTRY_VAR_AVAILABLE -> {
+                // Set language
+                tts.language = locale
+
+                // Check if voice data is present or not.
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+                    if (tts.voice != null) {
```
shubham1g5, line 156: should we returning `false` when `tts.voice==null` ?

shubham1g5 (changes requested):

app/src/org/commcare/tts/TextToSpeechConverter.kt
```diff
@@ -0,0 +1,191 @@
...
+            initialize(context, text)
+            return
+        }
+        // Handle empty text
+        if (TextUtils.isEmpty(text)) {
+            return
+        }
+        mTextToSpeech?.let { tts ->
+            if (isTextLong(text)) {
+                text.chunked(4000).forEach {
```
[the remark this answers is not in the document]
ShivamPokhriyal: Ahh yes, thanks for pointing it out.

app/src/org/commcare/views/media/MediaLayout.java
```diff
@@ -103,9 +105,15 @@ public static MediaLayout buildComprehensiveLayout(Context context,
                                                        TextView text, String audioURI, String imageURI,
                                                        final String videoURI, final String bigImageURI,
                                                        final String qrCodeContent, String inlineVideoURI,
+                                                       final String ttsText,
                                                        int questionIndex) {
         MediaLayout mediaLayout = new MediaLayout(context);
         mediaLayout.setAVT(text, audioURI, imageURI, videoURI, bigImageURI, qrCodeContent, inlineVideoURI, false, questionIndex);
+        // Show TTS view only when no other media is present
+        if (ttsText != null && audioURI == null && imageURI == null && videoURI == null
```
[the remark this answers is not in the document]
ShivamPokhriyal: Sure, will add this.

app/src/org/commcare/tts/TextToSpeechConverter.kt
ShivamPokhriyal: actually no. `tts.voice` can be null for any reason but we know for sure that the language is available just that we won't be sure whether the voice data is available. So if we return false here, that means we're saying we can't use TTS at all, but we can. And if the internet is available, tts will work just fine. And if not, we'll simply say a toast that TTS won't work cuz `utteranceListener` `onError` will get called.

app/src/org/commcare/tts/TextToSpeechConverter.kt
ShivamPokhriyal: > we should also pass the conuntry code from default locale to account for accents.

How do I get the country codes?

app/src/org/commcare/tts/TextToSpeechConverter.kt
ShivamPokhriyal: It didn't work for me in different android versions, but it's quite possible that the next TTS update of google might fix this, so I think the code would be useful then.

app/src/org/commcare/tts/TextToSpeechConverter.kt
shubham1g5: You can get the default country code using `Locale.getDefault().getCountry*`

app/src/org/commcare/tts/TextToSpeechConverter.kt
shubham1g5: I see, thanks for the explanation. Would be good to add an explicit else here and return `true` with a comment so that  we don't make a mistake of returning false from here in future.

app/src/org/commcare/tts/TextToSpeechConverter.kt
shubham1g5: makes sense.

app/src/org/commcare/tts/TextToSpeechConverter.kt
ShivamPokhriyal: Ohh, I thought you were talking about the `Localization.getCurrent` one. 
`Locale.getDefault()` returns a locale with the country code like `en_US`.

shubham1g5: So If you comprise the current locale as `Locale(Localization.getCurrentLocale(), Locale.getDefault().getCountry())` you should be able to add the country info user is in to the language right ?

dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462169/)

app/src/org/commcare/tts/TextToSpeechConverter.kt
ShivamPokhriyal: Ohh, I don't think that makes sense, like if `Localization.getCurrentLocale()` returns Arabic, and `Locale.getDefault().getCountry()` returns US. I'm not sure if it works like that.

shubham1g5: What I am thinking is if a user in India who has set the device language to Hindi, `Locale.getDefault().getCountry()` should return India as the country and setting that context will improve the TTS by far in comparison to having a US accent for Hindi.   In case of (Arabic, US) the TTS should still return `TextToSpeech.LANG_AVAILABLE` so the behaviour won't get changed in comparison to just (Arabic).

ShivamPokhriyal: ohh ok. I'll test how it works.

ShivamPokhriyal:

ShivamPokhriyal:

ShivamPokhriyal:

dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462179/)

app/src/org/commcare/tts/TextToSpeechConverter.kt
shubham1g5: Wondering if there are any concerns with this approach from the testing you did ?

ShivamPokhriyal: It worked fine for me.

app/src/org/commcare/views/media/MediaLayout.java
```diff
@@ -103,9 +105,14 @@ public static MediaLayout buildComprehensiveLayout(Context context,
                                                        TextView text, String audioURI, String imageURI,
                                                        final String videoURI, final String bigImageURI,
                                                        final String qrCodeContent, String inlineVideoURI,
+                                                       final String ttsText,
                                                        int questionIndex) {
         MediaLayout mediaLayout = new MediaLayout(context);
         mediaLayout.setAVT(text, audioURI, imageURI, videoURI, bigImageURI, qrCodeContent, inlineVideoURI, false, questionIndex);
+        // Show TTS view only when audioURI is not present
+        if (ttsText != null && audioURI == null) {
```
shubham1g5, line 113: How does the layout look when you have TTS plus a videoURI ?

shubham1g5 (approved):
Looks good to me. Would you be able to add a wiki page for this feature.

app/src/org/commcare/views/media/MediaLayout.java
ShivamPokhriyal: Screenshot_2020-12-15-10-12-39-794_org commcare dalvik debug

shubham1g5: thanks, this will work.

app/src/org/commcare/tts/TextToSpeechConverter.kt
```diff
@@ -0,0 +1,208 @@
...
+            text.length > TextToSpeech.getMaxSpeechInputLength()
+        } else {
+            text.length > MAX_TEXT_LENGTH
+        }
+    }
+
+    /**
+     * Initializes the Text-To-Speech engine
+     */
+    private fun initialize(context: Context, text: String) {
```
shubham1g5, line 129: Should not this called again after user change a language in CommCare ?

ShivamPokhriyal: We only need to change the language which is done here https://github.com/dimagi/commcare-android/pull/2411/files#diff-c7cb0f5ab8dcc60f97dfaf06c122ffa5e54311f12db10e5bd388d3d4035feffbR825

shubham1g5: Ahh I missed that. thanks

shubham1g5 (dismissed):

app/src/org/commcare/tts/TextToSpeechConverter.kt
```diff
@@ -163,31 +187,6 @@ object TextToSpeechConverter {
             TextToSpeech.LANG_COUNTRY_VAR_AVAILABLE -> {
                 // Set language
                 tts.language = locale
-
-                // Check if voice data is present or not.
-                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
-                    if (tts.voice != null) {
-                        val features = tts.voice.features
-                        if (features == null
```
ShivamPokhriyal: `features.contains(TextToSpeech.Engine.KEY_FEATURE_NOT_INSTALLED)` returns true irregularly even when the data is installed. Because of this false positive we were showing a dialog to install voice data when it was already installed.

shubham1g5 (approved):

ShivamPokhriyal:

dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462196/)

PR #2412, opened 2020-12-02
Launch Session endpoints from Intent

shubham1g5:
[Spec](https://docs.google.com/document/d/111mch19MnKMcXf91h5dmjvomYnket5x696Io3yLEjcE/edit?ts=5f724532#heading=h.5h6283rsmrrl)

cross-request: https://github.com/dimagi/commcare-core/pull/954


ShivamPokhriyal (approved):

app/src/org/commcare/activities/DispatchActivity.java
```diff
@@ -321,14 +330,24 @@ private void handleUnvalidatedApp() {
...
+                        String sessionRequest = this.getIntent().getStringExtra(SESSION_REQUEST);
+                        SessionStateDescriptor ssd = new SessionStateDescriptor();
+                        ssd.fromBundle(sessionRequest);
+                        CommCareApplication.instance().getCurrentSessionWrapper().loadFromStateDescription(ssd);
+                        Intent i = new Intent(this, StandardHomeActivity.class);
+                        i.putExtra(WAS_EXTERNAL, true);
+                        startActivityForResult(i, HOME_SCREEN);
+                    } else if (getIntent().hasExtra(SESSION_ENDPOINT_ID)) {
+                        String sessionEndpointId = this.getIntent().getStringExtra(SESSION_ENDPOINT_ID);
+                        ArrayList<String> args = this.getIntent().getStringArrayListExtra(SESSION_ENDPOINT_ARGUMENTS);
```
ctsims, line 343: Wouldn't these arguments need to be a Map (name to value) rather than a list?

ctsims: ^ If this were true, we might also want to consider having the args provided as a bundle map.

app/src/org/commcare/activities/HomeScreenBaseActivity.java
```diff
@@ -194,10 +199,59 @@ private void loadInstanceState(Bundle savedInstanceState) {
...
     }
 
+    private void processSessionEndpoint() {
+        if (getIntent().hasExtra(SESSION_ENDPOINT_ID)) {
+            Endpoint endpoint = validateIntentForSessionEndpoint(getIntent());
+            if (endpoint != null) {
+                Vector<String> endpointArguments = endpoint.getArguments();
+                ArrayList<String> intentArguments = getIntent().getStringArrayListExtra(SESSION_ENDPOINT_ARGUMENTS);
+                HashMap<String, String> compositeArguments = new HashMap<>(endpointArguments.size());
+                for (int i = 0; i < endpointArguments.size(); i++) {
```
ctsims, line 214: I mentioned this in the `commcare-core` PR, but I think we should be setting the arguments by setting them as XPath variables on the evaluation context, since that will naturally and robustly substitute them in as needed as the stack ops are calculated.

app/src/org/commcare/activities/HomeScreenBaseActivity.java
```diff
@@ -194,10 +199,59 @@ private void loadInstanceState(Bundle savedInstanceState) {
...
+                    compositeArguments.put(endpointArguments.elementAt(i),
+                            intentArguments.get(i));
+                }
+
+                CommCareApplication.instance().getCurrentSessionWrapper()
```
ctsims, line 219: Should we be ensuring the stack is clean here before firing everything?

ctsims (changes requested):
Mostly carryover feedback from the `cc-core` pr.

app/src/org/commcare/activities/HomeScreenBaseActivity.java
```diff
@@ -189,15 +194,75 @@ private void loadInstanceState(Bundle savedInstanceState) {
...
         }
     }
 
+    private void processSessionEndpoint() {
+        if (getIntent().hasExtra(SESSION_ENDPOINT_ID)) {
+            Endpoint endpoint = validateIntentForSessionEndpoint(getIntent());
+            if (endpoint != null) {
+                Vector<String> endpointArguments = endpoint.getArguments();
+                Bundle intentArguments = getIntent().getBundleExtra(SESSION_ENDPOINT_ARGUMENTS);
+                EvaluationContext evaluationContext = CommCareApplication.instance().getCurrentSessionWrapper().getEvaluationContext();
```
shubham1g5, line 213: Should we be creating a copy of `evaluationContext` here ?

ctsims: Requests to the session wrapper for an EC should always return a new, clean evaluation context, so I don't think you explicitly need to fork it here unless it will be used in more than one subcontext (IE: If, for example, you needed to evaluate expressions with different sets of argument variables being set).

shubham1g5: that makes sense. Thanks!

dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462192/)

shubham1g5:
@ctsims  Added a [commit](https://github.com/dimagi/commcare-android/pull/2412/commits/5567d2a1151a224a61cd7d2909a27393e951a9ab) to support both a map and a list of arguments so that external apps have flexibility to operate without strictly knowing the argument names.

shubham1g5:
@ctsims Also wondering if we would need to allow user to manually add more arguments to an endpoint. I am thinking of the use case when an external app would like to add a random session variable which can then be referenced in a form.

app/src/org/commcare/activities/HomeScreenBaseActivity.java
```diff
@@ -189,15 +194,86 @@ private void loadInstanceState(Bundle savedInstanceState) {
...
+                    if (intentArgumentsAsBundle != null && !intentArgumentsAsBundle.containsKey(argumentName)) {
+                        String noArgumentWithNameError = org.commcare.utils.StringUtils.getStringRobust(
+                                this,
+                                R.string.session_endpoint_no_argument_with_name,
+                                new String[]{argumentName, endpoint.getId()});
+                        UserfacingErrorHandling.createErrorDialog(this, noArgumentWithNameError, true);
+                        return;
+                    }
+
+                    // Bundle Arguments take precdence over list arguments
```
ctsims, line 228: Nit: It feels a bit error prone to me for the processing to not fork to being either Dictionary based or List Order based upfront and then in their own block, rather than inline under the same process, but just a style thing.

app/src/org/commcare/activities/HomeScreenBaseActivity.java
```diff
@@ -189,15 +194,86 @@ private void loadInstanceState(Bundle savedInstanceState) {
...
             sessionNavigator.startNextSessionStep();
         }
     }
 
+    private void processSessionEndpoint() {
+        if (getIntent().hasExtra(SESSION_ENDPOINT_ID)) {
+            Endpoint endpoint = validateIntentForSessionEndpoint(getIntent());
+            if (endpoint != null) {
+                Vector<String> endpointArguments = endpoint.getArguments();
+                EvaluationContext evaluationContext = CommCareApplication.instance().getCurrentSessionWrapper().getEvaluationContext();
```
ctsims, line 212: Just realized: We should be doing the session stack reset here, rather than below, since the EC here depends on the current session state, and this could introduce unpredictable behavior for the stack ops depending on the session state when the callout was received.

ctsims: Another consideration on that point: We may need the Endpoint definition to be able to provide its own `<instance>` connectors, since it won't necessarily be connected to any of the other components where connectors are defined. This can be "future" addition, since most of the inbound expressions now will just be Strings or whatnot, but I'd expect us to run into it at some point.

app/src/org/commcare/activities/HomeScreenBaseActivity.java
```diff
@@ -189,15 +194,86 @@ private void loadInstanceState(Bundle savedInstanceState) {
...
+                                new String[]{argumentName, endpoint.getId()});
+                        UserfacingErrorHandling.createErrorDialog(this, noArgumentWithNameError, true);
+                        return;
+                    }
+
+                    // Bundle Arguments take precdence over list arguments
+                    String argValue = intentArgumentsAsBundle == null ? intentArgumentsAsList.get(i)
+                            : intentArgumentsAsBundle.getString(argumentName);
+
+                    evaluationContext.setVariable(argumentName, argValue);
```
ctsims, line 232: Idea: The notion of supporting either a list or a dictionary to populate the list feels like it is in a funny place where it's part of the core app contract but implemented in Android.

I wonder whether it would be helpful to put more of this code into `commcare-core` in a method that receives either a list or dict and populates (or even retrieves then populates, based on the line 212 comment) an EC, just so the behavior is clearly part of the core contract.

app/src/org/commcare/activities/HomeScreenBaseActivity.java
```diff
@@ -189,15 +194,86 @@ private void loadInstanceState(Bundle savedInstanceState) {
...
+                }
+
+                CommCareApplication.instance().getCurrentSessionWrapper().reset();
+                CommCareApplication.instance().getCurrentSessionWrapper()
+                        .executeStackActions(endpoint.getStackOperations(), evaluationContext);
+            }
+        }
+    }
+
+    private Endpoint validateIntentForSessionEndpoint(Intent intent) {
```
ctsims, line 242: do we need the check block on line 219 if this validation code runs over things first?

ctsims (changes requested):
Hi Shubham. A few totally optional thoughts here. One change I didn't realize until re-reading that I think we should definitely make though (resetting the Session upfront, rather than right before executing the stack ops).

shubham1g5:

app/src/org/commcare/activities/HomeScreenBaseActivity.java
shubham1g5: Sounds good. Keeping it as a future addition.

dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462200/)

ctsims (approved):

shubham1g5:
> Also wondering if we would need to allow user to manually add more arguments to an endpoint on HQ. I am thinking of the use case when an external app would like to add a random session variable which can then be referenced in a form.

@ctsims any thoughts on this ?

PR #2416, opened 2020-12-04
Remove header from menu list

ShivamPokhriyal:
Jira: https://dimagi-dev.atlassian.net/browse/SAAS-11501
Product Note: Removes header title from sub-menu screen as it's already present in the action bar title.


ShivamPokhriyal:

dimagimon:
[32% (0.0%) vs master 32%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462178/)

shubham1g5:
@ctsims checking if you you can forsee any side effects of this change.

ctsims:
Honestly not 100% sure. 

The redundant header was part of Biyeun's UI redesign many years ago, not something we added from user feedback or usability testing, so we don't really have a pre/post expectation.

One thing I do think is important to highlight is that I don't believe that the majority of our users are anchored on Android UI patterns, so the idea that the text in the Action Bar describes the content of the menus may be much weaker than the visual alignment between this label and the content of the menu.

Another consideration that I honestly don't remember is whether the "Root" menu label is displayed on the menu bar, I thought there was at least one label that only ever was shown on the inline menu text, but I'm probably wrong about that.

I'm not quite sure how to explore the value of the existing label or how to determine whether it is worth preserving. If we wanted to be conservative, I'd say we should follow the "Disable by default" pattern where we make it optional and have HQ do the new behavior by default for new apps.

shubham1g5:
@ctsims  Thanks for your thoughts. 

> Another consideration that I honestly don't remember is whether the "Root" menu label is displayed on the menu bar

Root menu is just "Start" or in other words don't have a label. So we just show the CC App name in the action bar. 

We do get this issue as [feedback](https://dimagi-dev.atlassian.net/browse/SUPPORT-5219) both externally and internally so I am inclined to push it through on this PR. It seems like a bit of an overkill to control this behaviour by use of a HQ setting.

ctsims:
@shubham1g5 That outcome sounds right to me overall.

shubham1g5 (approved):

Issue #2321, opened 2020-08-20
feature request: allow arbitrarily named string as IntentCallout responses


shawnsarwar:
The IntentCallout activity assumes that called services have knowledge of ODK's incredibly specific expectations. Namely that the result be bundled and named `odk_intent_data`. This makes the callout feature unusable when trying to receive a result from most applications.

Would it be possible to see if `odk_intent_data` is actually returned, and if not try to `getString()` for the available keys that match the requested keys?  Even if you can only handle strings, it would go a long way to making generally IntentCallout useful.

shubham1g5:
Hey Shawn, That's a valid point and we do realize that it's constraining when you want to integrate with a general Android application you don't have control over. I think the purpose that having a specific callout key like `odk_intent_data` serve is to make sure that you are only integrating with Android apps that are meant to integrate with a ODK based Android app. Till now we have rarely encountered use cases where someone wants to receive data into CommCare from a third party Android app thats not meant to feed data into ODK/CommCare. Can you talk about your use case a bit more so that we can evaluate whether it's actually useful to relax this constraint. Thanks.

shawnsarwar:
Sure, we have a requirement to scan some 2d barcodes that aren't handled well by ZXing/ the integrated CC barcode reader as they're very small. We've found another off the shelf Android app that does well with them and has an intent level API which would allow it to be a drop in replacment.

> I think the purpose that having a specific callout key like odk_intent_data serve is to make sure that you are only integrating with Android apps that are meant to integrate with a ODK based Android app.

Shouldn't this be a decision for the form designer? Why would you want to limit the apps that Commcare can integrate with? One can still call whatever app they like, they just can't get any useful data returned. I understand expecting a certain format because being able to arbitrarily deserialize _anything_ from an intent is a lot to ask, but being so opinionated in convention that it breaks the intended behavior of intents doesn't feel like a feature.

shubham1g5:
@shawnsarwar Thanks for the explanation. That makes a lot of sense and I do agree that CommCare should support Android Intents better. We will try to prioritize this work for CommCare 2.51 though we only plan to release it on Playstore by Dec 2020/Jan 2021.

shawnsarwar:
@shubham1g5 Would a PR make this go faster? I might be able to find someone to handle the ticket.

shubham1g5:
[...]

PR #2425, opened 2021-01-12
Allow strings as intent callout response. 

ShivamPokhriyal:
https://github.com/dimagi/commcare-android/issues/2321

Product Note: Allow arbitrary named strings as intent callout response.


app/src/org/commcare/android/javarosa/IntentCallout.java
```diff
@@ -207,28 +213,44 @@ private boolean processOdkResponse(Intent intent, TreeReference intentQuestionRe
...
+                if (hasExtra) {
+                    return true;
                 }
             }
         }
         return (result != null);
     }
 
+    private void setOdkResponseValue(TreeReference intentQuestionRef, File destination, String responseValue, Vector<TreeReference> responseRefs) {
+        if (responseValue == null) {
```
shubham1g5, line 245: I think a better approach would be just to skip the call to `processResponseItem` when `responseValue` is null.

shubham1g5 (changes requested):
One minor suggestion but otherwise we should add the respective tests for this behaviour in `IntentCalloutTests`

app/src/org/commcare/android/javarosa/IntentCallout.java
ShivamPokhriyal: The earlier code uses empty value when `responseValue` is null. If that needs to be changed, then perhaps we can `return` here when the value is null?

shubham1g5: Ahh didn't realize it's like this before, no need for a change in behaviour in that case.

app/src/org/commcare/views/widgets/BarcodeWidget.java
```diff
@@ -31,7 +31,7 @@ public BarcodeWidget(Context context, FormEntryPrompt prompt, PendingCalloutInte
     }
 
     @Override
-    protected void performCallout() {
+    public void performCallout() {
```
shubham1g5, line 34: we should avoid changing method visibility for tests as much as possible. Instead you can just trigger the UI action that results in this being called out.

shubham1g5 (changes requested):

app/src/org/commcare/views/widgets/BarcodeWidget.java
ShivamPokhriyal: makes sense. Will do the changes.

shubham1g5 (approved):
looks great.

ShivamPokhriyal:

ShivamPokhriyal:

ShivamPokhriyal:

dimagimon:
[33% (0.0%) vs master 33%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462236/)

PR #2427, opened 2021-01-12
Move registering/unregistering of location listener to onStart/onStop

ShivamPokhriyal:
## Issue description
https://dimagi-dev.atlassian.net/browse/SAAS-11668
`Distance from current location` case detail display property is not working on mobile. The property won't even show up on the detail screen. 

The way it works is, it calculates the distance between location returned by `here()` function and the location of case. And if any of those locations is empty then it doesn't display the field at all. `AndroidHereFunctionHandler` class handles `here()` xpath function. It has `allowGpsUse` method which is called by `EntitySelectActivity` to request for location updates. But `EntitySelectActivity` doesn't call `allowGpsUse` method until `here()` function is evaluated atleast once. So it registers a listener `HereFunctionHandlerListener` which gets called when the `here()` function is evaluated. 

The issue was happening because `register` and `unregister` calls were happening in `onResume` and `onPause` so by time `here()` function is evaluated, `onPause` was already called and the listener was unregistered so `here()` function would always return empty. 

## Solution

Moved the calls to `register` and `unregister` listener in `onStart` and `onStop` of activity.


ShivamPokhriyal:

shubham1g5 (approved):

dimagimon:
[33% (0.0%) vs master 33%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462239/)

PR #2428, opened 2021-01-21
Scoped storage changes

ShivamPokhriyal:
Due to the restrictions from Android 10, we can no longer read files created using `File(filePath)` when `filePath` lies outside `app-specific directory`. 
This PR makes sure that we're always copying the file to `app-specific-directory` before doing any operation on it.


shubham1g5 (commented):
I think there are a lot more places in the code where we choose an external file and do similar processing, for eg - ccz install. Do we not need to make similar changes over there ? 

Separately If BS have an Android 11 device, we should run our instrumentation tests on it to catch any other regressions of Android 11. If not, we should check with QA if they have an Android 11 device available for the upcoming qa cycle.

ShivamPokhriyal:
> I think there are a lot more places in the code where we choose an external file and do similar processing, for eg - ccz install. Do we not need to make similar changes over there ?

Yeah, I'll see if that needs to be changed as well. 

> Separately If BS have an Android 11 device, we should run our instrumentation tests on it to catch any other regressions of Android 11. If not, we should check with QA if they have an Android 11 device available for the upcoming qa cycle.

They do have it the last time I checked, will use that to run androidTests in a separate job.

app/src/org/commcare/utils/FileUtil.java
```diff
@@ -632,19 +651,23 @@ public static void updateFileLocationFromIntent(Context context, Intent intent,
...
         Uri uriPath = intent.getData();
         if (uriPath != null) {
-            String filePath;
+            String fileName = getFileName(context.getContentResolver(), uriPath);
+            File destinationFile = new File(context.getCacheDir(), fileName);
             try {
-                filePath = UriToFilePath.getPathFromUri(CommCareApplication.instance(), uriPath);
-            } catch (UriToFilePath.NoDataColumnForUriException e) {
-                filePath = uriPath.toString();
+                InputStream inputStream = context.getContentResolver().openInputStream(uriPath);
```
shubham1g5, line 663: we should not need to make an unnecessary file copy here as consumers of result of this function like the `UnzipTask` already makes [that copy for internal processing](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/tasks/UnzipTask.java#L67). We should just return the uri from here and have callers of this method always work on a uri instead of the path (all these paths should already have uri handling since this function is already expected to return uri in case of errors during file path resolution).

shubham1g5 (changes requested):
Can you make sure to cover all the paths by checking usage for `UriToFilePath.getPathFromUri`

dimagimon:
[33% (0.0%) vs master 33%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462233/)

app/res/layout/file_pref_dialog.xml
```diff
@@ -29,6 +29,12 @@
 
             <EditText
                 android:id="@android:id/edit"
+                android:layout_width="10dp"
```
ShivamPokhriyal, line 32: This is the editText whose value is saved in the preference. I've made it invisible and added another edittext which gets shown to the user which we keep using this one for saving preference value.

ShivamPokhriyal:

ShivamPokhriyal:

app/src/org/commcare/utils/FileUtil.java
```diff
@@ -455,6 +458,22 @@ public static String getExtension(String filePath) {
...
+     * Retrieve a file's name using contentUri.
+     * https://developer.android.com/training/secure-file-sharing/retrieve-info.html#RetrieveFileInfo
+     */
+    public static String getFileName(ContentResolver contentResolver, Uri uri) {
+        Cursor cursor = contentResolver.query(uri, null, null, null, null);
+        if (cursor == null) {
+            return "";
+        }
+        int nameIndex = cursor.getColumnIndex(OpenableColumns.DISPLAY_NAME);
+        cursor.moveToFirst();
```
shubham1g5, line 471: This would crash if there is 0 row returned by the cursor.

shubham1g5 (changes requested):
Checking if you have been able to test functionality of any one of the `FilePreference` ? If not, I think the custom restore one should be straightforward to test.

app/src/org/commcare/activities/InstallArchiveActivity.java
```diff
@@ -82,7 +84,7 @@ protected void onCreate(Bundle savedInstanceState) {
         btnInstallArchive.setOnClickListener(v -> {
             String archivePath = editFileLocation.getText().toString();
             HiddenPreferences.setLastKnownCczLocation(archivePath);
-            ZipUtils.UnzipFile(this, archivePath, getTargetFolder());
+            ZipUtils.UnzipFile(this, selectedFileUri.toString(), getTargetFolder());
```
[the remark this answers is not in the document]
ShivamPokhriyal: Yeah, I missed the `HiddenPreference` thing here. It needs to be updated to use `selectedFileUri`.

app/src/org/commcare/activities/components/ImageCaptureProcessing.java
```diff
@@ -150,8 +143,7 @@ private static void processImageGivenFileUri(FormEntryActivity activity, String
         }
 
         // First make a copy of the image to operate on and then pass it to the File function
-        String extension = FileUtil.getExtension(imageUri.getPath());
-        String imageFilename = "tempfile" + "." + extension;
+        String imageFilename = FileUtil.getFileName(activity.getContentResolver(), imageUri);
```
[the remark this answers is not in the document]
ShivamPokhriyal: Not really sure what I had in mind at that time. I'll revert it to use tempfile only.

app/src/org/commcare/utils/FileUtil.java
ShivamPokhriyal: ohh, thanks for pointing.

app/src/org/commcare/activities/components/ImageCaptureProcessing.java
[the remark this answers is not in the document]
ShivamPokhriyal: Ohh, I get it now. `FileUtil.getExtension` won't necessarily return the file extension since the URIs won't have the extension.

dimagimon:
[33% (0.0%) vs master 33%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462245/)

ShivamPokhriyal:
> Checking if you have been able to test functionality of any one of the `FilePreference` ? If not, I think the custom restore one should be straightforward to test.

I tested the restore functionality and it shows sync successful.

dimagimon:
[33% (0.0%) vs master 33%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462246/)

app/src/org/commcare/activities/InstallArchiveActivity.java
```diff
@@ -80,7 +82,7 @@ protected void onCreate(Bundle savedInstanceState) {
         });
 
         btnInstallArchive.setOnClickListener(v -> {
-            String archivePath = editFileLocation.getText().toString();
+            String archivePath = selectedFileUri.toString();
             HiddenPreferences.setLastKnownCczLocation(archivePath);
```
shubham1g5, line 86: We are [operating](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/recovery/measures/ScanCczTask.java#L94) on the last known ccz location as new File(`uri`) , How does that work ?

ShivamPokhriyal: Ohh I missed that.

app/src/org/commcare/recovery/measures/ScanCczTask.java
```diff
@@ -94,7 +97,19 @@ private void locateAppCcz(File[] files, int depth) {
         String lastKnownCczLocation = HiddenPreferences.getLastKnownCczLocation();
         ArrayList<File> filePathsToScan = new ArrayList<>(3);
         if (!StringUtils.isEmpty(lastKnownCczLocation)) {
-            filePathsToScan.add(new File(lastKnownCczLocation));
+            try {
+                if (lastKnownCczLocation.startsWith("content://")) {
```
shubham1g5, line 101: It's a bad idea to copy the ccz given some of them can be really huge in size. I think we should make the caller of this function to work on file uri instead.

app/src/org/commcare/recovery/measures/ScanCczTask.java
```diff
@@ -94,7 +97,19 @@ private void locateAppCcz(File[] files, int depth) {
...
+                    filePathsToScan.add(destination);
+                } else {
+                    filePathsToScan.add(new File(lastKnownCczLocation));
+                }
+            } catch (IOException e) {
+                e.printStackTrace();
+                Logger.log(LogTypes.TYPE_ERROR_STORAGE, "Last known ccz location failed with : " + e.getMessage());
+            }
         }
         filePathsToScan.add(Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOWNLOADS));
```
shubham1g5, line 114: I think these paths will also throw the security errors right ?

ShivamPokhriyal: Yeah, I think so.

dimagimon:
[33% (0.0%) vs master 33%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462248/)

app/src/org/commcare/recovery/measures/ExecuteRecoveryMeasuresPresenter.java
```diff
@@ -155,12 +155,8 @@ private int executeMeasure() {
                     return STATUS_EXECUTED;
                 }
             case MEASURE_TYPE_APP_OFFLINE_REINSTALL_AND_UPDATE:
-                if (AppUtils.notOnLatestAppVersion()) {
-                    initateAutoCczScan();
-                    return STATUS_WAITING;
-                } else {
-                    return STATUS_EXECUTED;
-                }
+                onCCZScanComplete();
```
shubham1g5, line 158: This can be removed as well and replaced by -  

```
mActivity.hideReinstall()
setCczSelectionVisibility(true)
` ``

shubham1g5 (approved):

shubham1g5:
Looks good to me, I would just do some more testing around these changes to see if everything works as it supposed to and do a thorough pass looking into code for any places that might require more such changes. 

Great work with these 💯

dimagimon:
[33% (0.0%) vs master 33%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462252/)

ShivamPokhriyal:
> Looks good to me, I would just do some more testing around these changes to see if everything works as it supposed to and do a thorough pass looking into code for any places that might require more such changes.

Sure, I'll test this on different android versions to see that nothing breaks, But will merge this one now and create a followup if I find something.

PR #2430, opened 2021-01-22
Fix linking to external files

ShivamPokhriyal:
Jira Ticket:: https://dimagi-dev.atlassian.net/browse/SAAS-11757

I've added a custom plugin for handling clicks of markdown text. 

While testing it out on Android 10, I found that `ACTION_VIEW` intent won't show the images of external storage.(I modified the paths for the same `<external-path name="external_files" path="." />`). But it would simply show an error in the gallery app `File format isn't supported or file is corrupted`. 

 I created an app [here](https://www.commcarehq.org/a/srishti-1/apps/view/e75ca00a7e53385e292049d16c0fa1ac/form/936705e7ee9c4f5bbe77708a795f2b0a/source/#form/total_repeat_count) and only the first question shows the image correctly.

### DIFF
--- a/app/res/xml/provider_paths.xml
+++ b/app/res/xml/provider_paths.xml
@@ -2,4 +2,5 @@
 <paths>
     <external-files-path name="all_external_files" path="."/>
     <external-files-path name="external_temp" path="temp/external/"/>
+    <external-path name="external_files" path="." />
 </paths>
--- a/app/src/org/commcare/CommCareApplication.java
+++ b/app/src/org/commcare/CommCareApplication.java
@@ -92,6 +92,7 @@ import org.commcare.utils.CrashUtil;
 import org.commcare.utils.DeviceIdentifier;
 import org.commcare.utils.FileUtil;
 import org.commcare.utils.GlobalConstants;
+import org.commcare.utils.MarkupUtil;
 import org.commcare.utils.MultipleAppsUtil;
 import org.commcare.utils.PendingCalcs;
 import org.commcare.utils.SessionActivityRegistration;
@@ -131,6 +132,7 @@ import androidx.work.NetworkType;
 import androidx.work.OneTimeWorkRequest;
 import androidx.work.PeriodicWorkRequest;
 import androidx.work.WorkManager;
+
 import io.noties.markwon.Markwon;
 import io.noties.markwon.ext.strikethrough.StrikethroughPlugin;
 import io.noties.markwon.ext.tables.TablePlugin;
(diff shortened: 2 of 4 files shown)


app/res/xml/provider_paths.xml
```diff
@@ -2,4 +2,5 @@
 <paths>
     <external-files-path name="all_external_files" path="."/>
     <external-files-path name="external_temp" path="temp/external/"/>
+    <external-path name="external_files" path="." />
```
shubham1g5, line 5: what does this do ?

shubham1g5 (commented):
Can you please make sure to add Jira tickets in the PRs. 

I am confused why doesn't it work for external files ? For eg. How does a Gallery app on Android 11 works if it can't display contents of the `Camera` folder ?

app/res/xml/provider_paths.xml
ShivamPokhriyal: `external-files-path` gives content uri for `storage/emulated/0/data/org.commcare.dalvik/` directory and `external-path` gives content uri for `storage/emulated/0/` directory. So I added this to get share contentUris of external files like the ones in downloads or dcim folder.

ShivamPokhriyal:
> How does a Gallery app on Android 11 works if it can't display contents of the `Camera` folder ?

I think those are the manufacturer's gallery app and I'm assuming they have system level permissions.

shubham1g5:
I am quite confused with this one. If we don't have access to files in say Downloads folder how are we able to copy them in internal storage in https://github.com/dimagi/commcare-android/pull/2428

ShivamPokhriyal:
> I am quite confused with this one. If we don't have access to files in say Downloads folder how are we able to copy them in commcare's scope storage in #2428

We use `content uri` for copying the files to scoped storage. Here we don't have a content URI and we can't use the file URI directly since that isn't accessible.

dimagimon:
[33% (0.0%) vs master 33%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462242/)

shubham1g5 (approved):
I see, thanks for the explanation. I think this behaviour is only mostly used to link a file to Downloads folder, so while this PR looks good to me and is good to be merged, we should update the documentation for this feature to mark it as deprecated and communicate the same to support team

ShivamPokhriyal:
> I see, thanks for the explanation. I think this behaviour is only mostly used to link a file to Downloads folder, so while this PR looks good to me and is good to be merged, we should update the documentation for this feature to mark it as deprecated sighting the reason behind it and communicate the same to support team if they wanna further communicate it to projects.

Yup, will inform the support team about this.

PR #2431, opened 2021-01-25
Discard image when file limit exceeds. 

ShivamPokhriyal:
Jira: https://dimagi-dev.atlassian.net/browse/SAAS-11633

Product Note: Show an error dialog when file of size more than 15 MB is selected by the user.

### DIFF
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -71,6 +71,7 @@ import org.commcare.utils.StringUtils;
 import org.commcare.views.QuestionsView;
 import org.commcare.views.ResizingImageView;
 import org.commcare.views.UserfacingErrorHandling;
+import org.commcare.views.dialogs.CommCareAlertDialog;
 import org.commcare.views.dialogs.CustomProgressDialog;
 import org.commcare.views.dialogs.StandardAlertDialog;
 import org.commcare.views.widgets.BarcodeWidget;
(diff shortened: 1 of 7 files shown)


app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -400,6 +409,8 @@ private void processChooserResponse(Intent intent) {
             Toast.makeText(FormEntryActivity.this,
                     Localization.get("form.attachment.invalid"),
                     Toast.LENGTH_LONG).show();
+        } else if (FileUtil.isFileTooLargeToUpload(getContentResolver(), media)) {
```
shubham1g5, line 412: this seems to be check on original file again instead of the scaled image

shubham1g5 (changes requested):

app/src/org/commcare/activities/FormEntryActivity.java
ShivamPokhriyal: yeah this is for audio and video. We only compress images, so the check is on the original file.

shubham1g5: Makes sense

shubham1g5 (dismissed):

shubham1g5 (approved):
Thanks! This looks much cleaner :)

ShivamPokhriyal:
I'm merging this without waiting for the tests result since they're taking too long. 
I'll be running tests on different devices along with QA.

dimagimon:
[33% (0.0%) vs master 33%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462255/)

PR #2433, opened 2021-01-27
Launcher icon fix

ShivamPokhriyal:
Jira: https://dimagi-dev.atlassian.net/browse/QA-2415

CommCare launcher icon was showing up as the default android launcher icon. Even though we had `tools:replace="android:icon"` it was not working. 

My theory behind the issue is that it's happening on devices that support round icon. And we didn't had a `ic_launcher_round` icon for CommCare. So it was taking up that icon from `commcare-support-library`. I tried setting `tools:replace="android:roundIcon"` as well but it didn't work.

So I removed the launcher icons from `commcare-support-library`.  I'm not sure if this requires a release into bintray. But since we're using `implementation project(':commcare-support-library')` this will work for us but it won't work for someone who's using bintray. 

At first I thought the icon was coming from `zebra-print-android` library, atleast that's what the android studio told me. So I removed the icon from there as well https://github.com/dimagi/zebra-print-android/pull/6 . But it didn't work after that because the icon was also present in `commcare-support-library`.

I do think that we should also rename our launcher icon to something like `commcare_launcher`. Or maybe we can have our own roundIcons and that might also solve this.


shubham1g5:
>  this will work for us but it won't work for someone who's using bintray.

What do you mean here ? What would break for someone using support lib from bintray 

> I do think that we should also rename our launcher icon to something like `commcare_launcher`

that should be totally fine.

ShivamPokhriyal:
> What do you mean here ? What would break for someone using support lib from bintray

Oh, I meant that if someone is also not using roundIcons and has a dependency on `commcare-support-library` then he'd also face the same issue unless we publish these changes to bintray and release a new version? Or I might be wrong.

shubham1g5 (approved):

PR #2434, opened 2021-02-02
Hide save button when recorded sound is shown

ShivamPokhriyal:
This PR will remove the save button from the recording fragment, when it's loaded with a saved recording. 

 
recording


shubham1g5:
@ShivamPokhriyal Is it not possible for user to redo/replace the recording ? If yes, will the save button appear in case user decided to record again

ShivamPokhriyal:
> @ShivamPokhriyal Is it not possible for user to redo/replace the recording ? If yes, will the save button appear in case user decided to record again

There's a record again button for that.

shubham1g5 (approved):

PR #2435, opened 2021-02-02
skip invalid case search prompts

shubham1g5:
@sravfeyn highlighted that as Web Apps is implementing more case search widgets, it would be a good idea for mobile to skip displaying any widgets it doesn't implement so as not to break searches in case of incompatible input types.

### DIFF
--- a/app/src/org/commcare/activities/QueryRequestActivity.java
+++ b/app/src/org/commcare/activities/QueryRequestActivity.java
@@ -156,7 +156,7 @@ public class QueryRequestActivity
         }
     }
 
-    private boolean isPromptValid(QueryPrompt queryPrompt) {
+    private boolean isPromptSupported(QueryPrompt queryPrompt) {
         return queryPrompt.getInput() == null || queryPrompt.getInput().contentEquals(INPUT_TYPE_SELECT1);
     }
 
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/activities/QueryRequestActivity.java
```diff
@@ -142,16 +142,22 @@ private void buildPromptEntry(LinearLayout promptsLayout, String promptId,
...
+
+            promptsLayout.addView(promptView);
+            promptsBoxes.put(promptId, inputView);
         }
-        setUpBarCodeScanButton(promptView, promptId, queryPrompt);
+    }
 
-        promptsLayout.addView(promptView);
-        promptsBoxes.put(promptId, inputView);
+    private boolean isPromptValid(QueryPrompt queryPrompt) {
```
ctsims, line 159: Nit: I would find 'valid' to be a bit confusing if I ran across this code. I might rename this as something like 'isPromptSupported'

ctsims (approved):

app/src/org/commcare/activities/QueryRequestActivity.java
shubham1g5:
```suggestion
    private boolean isPromptSupported(QueryPrompt queryPrompt) {
```

PR #2441, opened 2021-03-01
Search view fix

ShivamPokhriyal:
Jira: https://dimagi-dev.atlassian.net/browse/SUPPORT-8015

This PR fixes a NPE while setting text in searchview after scanning barcode.  
```
Caused by java.lang.NullPointerException: Attempt to invoke virtual method 'void android.widget.EditText.setText(java.lang.CharSequence)' on a null object reference
       at org.commcare.activities.EntitySelectSearchUI.setSearchText(EntitySelectSearchUI.java:124)
       at org.commcare.activities.EntitySelectActivity.processBarcodeFetch(EntitySelectActivity.java:642)
       at org.commcare.activities.EntitySelectActivity.onActivityResultSessionSafe(EntitySelectActivity.java:574)
       at org.commcare.activities.SessionAwareHelper.onActivityResultHelper(SessionAwareHelper.java:49)
       at org.commcare.activities.SessionAwareCommCareActivity.onActivityResult(SessionAwareCommCareActivity.java:39)
       at android.app.Activity.dispatchActivityResult(Activity.java:8316)
       at android.app.ActivityThread.deliverResults(ActivityThread.java:5206)
       at android.app.ActivityThread.performResumeActivity(ActivityThread.java:4618)
       at android.app.ActivityThread.handleResumeActivity(ActivityThread.java:4663)
       at android.app.servertransaction.ResumeActivityItem.execute(ResumeActivityItem.java:52)
       at android.app.servertransaction.TransactionExecutor.executeLifecycleState(TransactionExecutor.java:176)
       at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:97)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2246)
       at android.os.Looper.loop(Looper.java:233)
       at android.app.ActivityThread.main(ActivityThread.java:8010)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:631)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:978)
```

The issue happens when the screen orientation changes while the user is scanning the barcode. 
So when the scan is completed, `onActivityResult` will be called. But because of orientation change, searchView becomes null and need to be set again from  `onOptionsMenuItem`. But the call to `onActivityResult` happens before `onOptionsMenuItem`
so it throws the NPE.

### DIFF
--- a/app/src/org/commcare/activities/EntitySelectSearchUI.java
+++ b/app/src/org/commcare/activities/EntitySelectSearchUI.java
@@ -1,13 +1,11 @@
 package org.commcare.activities;
 
 import android.annotation.TargetApi;
-import android.content.Context;
 import android.os.Build;
 import android.text.Editable;
 import android.text.TextWatcher;
 import android.view.MenuItem;
 import android.view.View;
-import android.view.inputmethod.InputMethodManager;
 import android.widget.EditText;
 import android.widget.ImageButton;
 import android.widget.TextView;
@@ -15,7 +13,6 @@ import android.widget.TextView;
 import org.commcare.activities.components.EntitySelectCalloutSetup;
 import org.commcare.dalvik.R;
 import org.commcare.suite.model.Callout;
-import org.javarosa.core.services.locale.Localization;
 
 import androidx.appcompat.widget.SearchView;
 import androidx.core.view.MenuItemCompat;
@@ -29,7 +26,6 @@ class EntitySelectSearchUI implements TextWatcher {
     private SearchView searchView;
     private MenuItem searchMenuItem;
     private MenuItem barcodeMenuItem;
-    private EditText preHoneycombSearchBox;
     private TextView searchResultStatus;
     private ImageButton clearSearchButton;
     private View searchBanner;
@@ -108,9 +104,8 @@ class EntitySelectSearchUI implements TextWatcher {
     protected CharSequence getSearchText() {
         if (isUsingActionBar()) {
             return searchView.getQuery();
-        } else {
-            return preHoneycombSearchBox.getText();
         }
+        return "";
     }
 
     @SuppressWarnings("NewApi")
@@ -120,8 +115,6 @@ class EntitySelectSearchUI implements TextWatcher {
                 MenuItemCompat.expandActionView(searchMenuItem);
             }
             searchView.setQuery(text, false);
-        } else {
-            preHoneycombSearchBox.setText(text);
         }
     }
 
--- a/app/src/org/commcare/activities/EntitySelectSearchUI.java
+++ b/app/src/org/commcare/activities/EntitySelectSearchUI.java
@@ -32,7 +32,6 @@ class EntitySelectSearchUI implements TextWatcher {
     private final EntitySelectActivity activity;
 
     private String filterString = "";
-    private String searchText;
 
     EntitySelectSearchUI(EntitySelectActivity activity) {
         this.activity = activity;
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/activities/EntitySelectSearchUI.java
```diff
@@ -93,9 +94,7 @@ public void setupActionImage(Callout callout) {
     private void restoreLastQuery() {
         String lastQueryString = activity.getLastQueryString();
         if (lastQueryString != null && lastQueryString.length() > 0) {
-            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.ICE_CREAM_SANDWICH) {
-                MenuItemCompat.expandActionView(searchMenuItem);
-            }
+            searchMenuItem.expandActionView();
```
ShivamPokhriyal, line 97: `MenuItemCompat.expandActionView` is deprecated.

app/src/org/commcare/activities/EntitySelectSearchUI.java
```diff
@@ -106,22 +105,19 @@ private void restoreLastQuery() {
...
-        if (isUsingActionBar()) {
+        if (searchView != null) {
             return searchView.getQuery();
-        } else {
-            return preHoneycombSearchBox.getText();
         }
+        return "";
     }
 
     @SuppressWarnings("NewApi")
```
shubham1g5, line 114: can be removed ?

app/src/org/commcare/activities/EntitySelectSearchUI.java
shubham1g5: would be good to test that this doesn't cause any side-effects on Android 4 devices.

shubham1g5 (commented):
From the stacktrace, It seems like issue happens because of `preHoneycombSearchBox` being null and as I see it this EditText doesn't seem to be set from anywhere. Can you confirm ?

ShivamPokhriyal:
> From the stacktrace, It seems like issue happens because of `preHoneycombSearchBox` being null and as I see it this EditText doesn't seem to be set from anywhere. Can you confirm ?

That's correct. The crash happens because `preHoneycombSearchBox` is null(which is always null). But the codePath only gets triggered when `searchView` is null, which shouldn't be.

app/src/org/commcare/activities/EntitySelectSearchUI.java
```diff
@@ -56,14 +52,13 @@ private void initUIComponents() {
...
-        return (searchItem, searchView, barcodeItem) -> {
-            EntitySelectSearchUI.this.searchMenuItem = searchItem;
-            EntitySelectSearchUI.this.searchView = searchView;
-            EntitySelectSearchUI.this.barcodeMenuItem = barcodeItem;
-
+        return (searchItem, view, barcodeItem) -> {
+            searchMenuItem = searchItem;
+            searchView = view;
+            barcodeMenuItem = barcodeItem;
             restoreLastQuery();
```
shubham1g5, line 59: Looks like this method already restores the search after rotation. Is it not working as intended ?

ShivamPokhriyal: Ohh, I overlooked that. I'm not sure how it's working right now, but I'll check it out.

ShivamPokhriyal: I checked it and seems like `restoreLastQuery` handles rotation only when entity select activity is on top. But when user opens  scanner and then rotate the screen, it won't work.

shubham1g5: And what's the reason behind that. Last query seems to be saved as part of CC session so it should not get cleared right ?

shubham1g5 (approved):

PR #2443, opened 2021-03-08
Fix Memory leaks

ShivamPokhriyal:
There has been an increase in the OOMs with CommCare 2.51. This PR targets to remove some memory leaks so as to avoid those OOMs.


dimagimon:
[33% (0.0%) vs master 33%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462276/)

shubham1g5 (approved):

PR #2448, opened 2021-03-16
Handle receiver register/unregister calls in onResume and onPause only

ShivamPokhriyal:
Jira: https://dimagi-dev.atlassian.net/browse/SAAS-11948

`onActivityResultHelper` was calling `SessionActivityRegistration.handleOrListenForSessionExpiration` which was also registering the broadcast receiver, thereby making the activity to register  twice for the same receiver and causing a memory leak.

This PR will seperate the code to register and unregister the broadcast receiver in different methods so that they can be used only in `onResume` and `onPause` of the activity.

### DIFF
--- a/app/src/org/commcare/utils/SessionActivityRegistration.java
+++ b/app/src/org/commcare/utils/SessionActivityRegistration.java
@@ -41,8 +41,6 @@ public class SessionActivityRegistration {
      * methods of activities that are session sensitive.
      */
     public static boolean handleOrListenForSessionExpiration(AppCompatActivity activity) {
-        activity.registerReceiver(userSessionExpiredReceiver, expirationFilter);
-
         synchronized (registrationLock) {
             if (unredirectedSessionExpiration) {
                 unredirectedSessionExpiration = false;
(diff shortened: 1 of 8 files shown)


app/src/org/commcare/utils/SessionActivityRegistration.java
```diff
@@ -41,8 +41,6 @@ public void onReceive(Context context, Intent intent) {
      * methods of activities that are session sensitive.
      */
     public static boolean handleOrListenForSessionExpiration(AppCompatActivity activity) {
```
shubham1g5, line 43: can we rename it to just `handleSessionExpiration` since it doesn't register the receiver anymore.

shubham1g5 (approved):

ShivamPokhriyal:
@shubham1g5 Should we name `SessionActivityRegistration` as well? Since it isn't an activity, we can perhaps change it to `SessionRegistrationHelper`?

shubham1g5:
I am good with that.

shubham1g5 (approved):

PR #2449, opened 2021-03-17
Add logs to DrawActivity

ShivamPokhriyal:
Jira: https://dimagi-dev.atlassian.net/browse/SAAS-11641

Some users have reported that they couldn't save the signature. More details on the ticket. 
This PR will add logging to the DrawActivity to grab more information about the issue.

### DIFF
--- a/app/src/org/commcare/activities/DrawActivity.java
+++ b/app/src/org/commcare/activities/DrawActivity.java
@@ -203,6 +203,7 @@ public class DrawActivity extends AppCompatActivity {
 
     private void saveAndClose() {
         try {
+            Logger.log(LogTypes.SOFT_ASSERT, "Attempting to save signature");
             saveFile(output);
             setResult(AppCompatActivity.RESULT_OK);
         } catch (FileNotFoundException e) {
--- a/app/src/org/commcare/activities/DrawActivity.java
+++ b/app/src/org/commcare/activities/DrawActivity.java
@@ -23,6 +23,7 @@ import android.widget.LinearLayout;
 import android.widget.RelativeLayout;
 
 import org.commcare.dalvik.R;
+import org.commcare.util.LogTypes;
 import org.commcare.utils.FileUtil;
 import org.commcare.utils.MediaUtil;
 import org.commcare.utils.StringUtils;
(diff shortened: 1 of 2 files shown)


shubham1g5 (approved):

app/src/org/commcare/activities/DrawActivity.java
```diff
@@ -205,6 +207,8 @@ private void saveAndClose() {
             setResult(AppCompatActivity.RESULT_OK);
         } catch (FileNotFoundException e) {
             e.printStackTrace();
+            // We shouldn't have gotten a file not found exception since we're using ImageWidget.getTempFileForImageCapture.
+            Logger.log(LogTypes.TYPE_ERROR_DESIGN, "Couldn't save signature at " + output.toString() + " because of" + e.getMessage());
```
shubham1g5, line 211: can `output` be null ?

shubham1g5:
Another area to survey and add logging might be when `DrawActivity` return result to the Form entry and see if the result is reliably returned to Form Entry or not.

app/src/org/commcare/activities/DrawActivity.java
ShivamPokhriyal: That's a good question. Umm, Although it gets set in `onCreate` [here](https://github.com/dimagi/commcare-android/blob/30f899656c672b22e7784a49eac8287ce2d2b277/app/src/org/commcare/activities/DrawActivity.java#L96) and [here](https://github.com/dimagi/commcare-android/blob/30f899656c672b22e7784a49eac8287ce2d2b277/app/src/org/commcare/activities/DrawActivity.java#L134-L136) but it could get null maybe. 

Though, I think the `FileNotFoundException` exception would only be thrown when the `output` is non-null. Otherwise it'd have thrown an NPE right? So I'm assuming that it's safe to access it here.

ShivamPokhriyal:
> see if the result is reliably returned to Form Entry or not.

I'm afraid I don't understand what you meant here. 
Do you mean it's possible that `DrawActivity` `setResult` and finishes but `FormEntryActivity` doesn't get callback on `onActivityResult`?

shubham1g5:
> Do you mean it's possible that DrawActivity setResult and finishes but FormEntryActivity doesn't get callback on onActivityResult?

yes, I am worried because of the rotation if something funny is happening on the device causing the result to get lost.

app/src/org/commcare/activities/DrawActivity.java
shubham1g5: sounds good

shubham1g5 (approved):

PR #2450, opened 2021-03-18
Backmerge Commcare 2.51.2

ShivamPokhriyal:


shubham1g5 (approved):
we will need a core backmerge too

ShivamPokhriyal:
> we will need a core backmerge too

Ahh, thanks for reminding.

PR #2467, opened 2021-05-07
Use file backed storage for reading fixture while purging case

ShivamPokhriyal:
**Jira**: https://dimagi-dev.atlassian.net/browse/SAAS-12191

**StackTrace**: 
```
Non-fatal Exception: java.lang.NullPointerException: Attempt to get length of null array
       at java.io.ByteArrayInputStream.<init>(ByteArrayInputStream.java:106)
       at org.commcare.models.database.SqlStorage.newObject(SqlStorage.java:249)
       at org.commcare.models.database.SqlStorage.read(SqlStorage.java:450)
       at org.commcare.models.database.SqlStorage.read(SqlStorage.java:43)
       at org.commcare.utils.CommCareUtil.loadFixture(CommCareUtil.java:32)
       at org.commcare.engine.cases.CaseUtils.purgeCases(CaseUtils.java:93)
       at org.commcare.tasks.DataPullTask.doTaskBackgroundHelper(DataPullTask.java:166)
       at org.commcare.tasks.DataPullTask.doTaskBackground(DataPullTask.java:139)
       at org.commcare.tasks.DataPullTask.doTaskBackground(DataPullTask.java:64)
       at org.commcare.tasks.templates.CommCareTask.doInBackground(CommCareTask.java:40)
       at android.os.AsyncTask$3.call(AsyncTask.java:378)
       at java.util.concurrent.FutureTask.run(FutureTask.java:266)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
       at java.lang.Thread.run(Thread.java:919)
```

**Summary**: 
The issue was happening in the sync process. While doing a case purge, we try to read the `user-groups` fixture in `CommCareUtil` class which used `SqlStorage` instead of `HybridFileBackedSqlStorage`. 

At the time of writing this `user-groups` fixture to the memory, the object size was more than 1 MB, so rather than storing the object to `DATA_COL` we instead stored it in a file and saved the file location to `FILE_COL`.

But, at the time of reading the fixture from db, we used `SqlStorage` which only checks the value in `DATA_COL` which was null and caused the NPE.

### DIFF
--- a/app/src/org/commcare/utils/CommCareUtil.java
+++ b/app/src/org/commcare/utils/CommCareUtil.java
@@ -22,8 +22,8 @@ import java.util.Vector;
  */
 public class CommCareUtil {
     public static FormInstance loadFixture(String refId, String userId) {
-        IStorageUtilityIndexed<FormInstance> userFixtureStorage = CommCareApplication.instance().getUserStorage("fixture", FormInstance.class);
-        IStorageUtilityIndexed<FormInstance> appFixtureStorage = CommCareApplication.instance().getAppStorage("fixture", FormInstance.class);
+        IStorageUtilityIndexed<FormInstance> userFixtureStorage = CommCareApplication.instance().getFileBackedUserStorage("fixture", FormInstance.class);
+        IStorageUtilityIndexed<FormInstance> appFixtureStorage = CommCareApplication.instance().getFileBackedAppStorage("fixture", FormInstance.class);
 
         Vector<Integer> userFixtures = userFixtureStorage.getIDsForValue(FormInstance.META_ID, refId);
         ///... Nooooot so clean.


ctsims:
Hi Shivam,

I'm a bit confused, are these two fixtures always stored in `FileBackedUserStorage`, or only when their size fails over to a large size? This seems like a huge change but maybe I'm missing something where we are already failing over to this normally, but this particular call only uses the old object.

ShivamPokhriyal:
> I'm a bit confused, are these two fixtures always stored in `FileBackedUserStorage`, or only when their size fails over to a large size? This seems like a huge change but maybe I'm missing something where we are already failing over to this normally, but this particular call only uses the old object.

I think they both are always stored in file backed storage.
https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/models/database/AndroidSandbox.java#L79-L87

Even the `HybridFileBackedSqlStorage` has a constant for fixture tablename, which I guess should mean that fixtures are always file backed.  
https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/models/database/HybridFileBackedSqlStorage.java#L46

ShivamPokhriyal:
> create a playstore beta pre-release 

Do you mean create a beta pre-release for the next major version `2.52.0` ? And I'm guessing, the beta pre-release doesn't need to go through the QA cycle right?

ctsims:
Correct, and correct. I would let them know specifically that the pre-release version hasn't undergone QA, but is available at their own risk.

Alternatively we can just share the Jenkins build artifact directly, but I think it's probably good for us to be leveraging pre-release  channels a bit more frequently for areas like this where programs might want access to non-critical bugfixes for program rollout.

ShivamPokhriyal:

shubham1g5 (approved):

ShivamPokhriyal:

dimagimon:
[33% (0.0%) vs master 33%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462394/)

Issue #2473, opened 2021-05-12
Not able to build commcare android 


siddesh13:
Hello,
As I import commcare android in android studio I am not able to build project because of the following issue
Unable to resolve dependency for ':app@standaloneRelease/compileClasspath': Could not resolve io.ona.kujaku:library:0.7.2.
Could not GET 'http://dl.bintray.com/ona/kujaku/io/ona/kujaku/library/0.7.2/library-0.7.2.pom'. Received status code 403 from server: Forbidden
Please suggest
Thanks
Siddesh S

ShivamPokhriyal:
Hi @siddesh13, Sorry to hear that the build is failing. 

Is your branch up to date with `master`? It seems like `Could not resolve io.ona.kujaku:library:0.7.2` is trying to use version `0.7.2` of kujaku and we're on version `0.8.7` currently. 

Have you made any changes to `commcare`?

siddesh13:
Having same error  on master 
Could not GET 'http://dl.bintray.com/ona/kujaku/io/ona/kujaku/library/0.8.7/library-0.8.7.pom'. Received status code 403 from server: Forbidden

shubham1g5:
It's probably due to [this](https://github.com/onaio/kujaku/issues/352), waiting for a response there. In the meanwhile you can try commenting this dependency and `BaseMapboxActivity` and it's subclasses to build.

ekigamba:
@shubham1g5  Replied https://github.com/onaio/kujaku/issues/352#issuecomment-839698480

siddesh13:
Could not GET 'https://maven.pkg.github.com/onaio/kujaku/io/ona/kujaku/library/0.9.0/library-0.9.0.pom'. Received status code 401 from server: Unauthorized

ekigamba:
@siddesh13 Kindly make sure you have the github token configured and username with access to read-packages

shubham1g5:
@siddesh13 this should be resolved now. Can you pull again and try. Thanks!

PR #2475, opened 2021-05-12
Shifts Kujaku to download from github packages

shubham1g5:
https://github.com/dimagi/commcare-android/issues/2473
https://github.com/onaio/kujaku/issues/352#issuecomment-839698480

This will require devs to generate a [personal access token](https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token) from github and add [github.properties](https://smartregister.atlassian.net/wiki/spaces/Documentation/pages/2641035265/Publishing+to+GitHub+Package+Registry#2.-Create-a-github.properties-file )

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -19,13 +19,23 @@ apply plugin: 'kotlin-android-extensions'
 apply plugin: 'com.google.firebase.crashlytics'
 apply from: 'jacoco.gradle'
 
+def githubProperties = new Properties()
+githubProperties.load(new FileInputStream(rootProject.file("github.properties")))
+
 repositories {
     // for local aar inclusion
     flatDir {
         dirs 'libs'
     }
     maven { url "https://jitpack.io" }
-    maven { url "http://dl.bintray.com/ona/kujaku" }
+    maven {
+        name = "GitHubPackages"
+        url = uri("https://maven.pkg.github.com/onaio/kujaku")
+        credentials {
+            username = githubProperties['gpr.usr'] ?: System.getenv("GPR_USER")
+            password = githubProperties['gpr.key'] ?: System.getenv("GPR_API_KEY")
+        }
+    }
 }
 
 configurations {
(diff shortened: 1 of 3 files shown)


ShivamPokhriyal (approved):

app/build.gradle
```diff
@@ -19,13 +19,23 @@ apply plugin: 'kotlin-android-extensions'
...
     flatDir {
         dirs 'libs'
     }
     maven { url "https://jitpack.io" }
-    maven { url "http://dl.bintray.com/ona/kujaku" }
+    maven {
+        name = "GitHubPackages"
+        url = uri("https://maven.pkg.github.com/onaio/kujaku")
+        credentials {
+            username = githubProperties['gpr.usr'] ?: System.getenv("GPR_USER")
```
ShivamPokhriyal, line 35: `System.getenv("GPR_USER")` is getting value from jenkins job config right? 
If yes, Should we rather use `gradle.properties` here, since creating new jobs in jenkins would require adding this property.

ctsims:
Oof, this is a really rough change. This adds a new dependency where anyone downloading the codebase needs a github packages PAT? Have we asked the Kujaku project maintainers to change that? I think this calls the dependency into question.

app/build.gradle
shubham1g5: good point, will update.

shubham1g5:
@ctsims that's true unfortunately, though I think Github Packages is a temporary workaround from them until they are able to migrate to [maven central](https://github.com/onaio/kujaku/issues/352). @ekigamba Wondering if you will be open to shift to more open artifact managers like Jitpack for example as I assume migrating to maven central will take some time for you.

ekigamba:
@ctsims @shubham1g5 Github packages is a temporary solution for now. We are open to any other artifact manager that is easy to integrate. We are open to any suggestions on the issue https://github.com/onaio/kujaku/issues/352

shubham1g5:
Closing in favour of https://github.com/dimagi/commcare-android/pull/2481

PR #2502, opened 2021-06-17
Encryption support for NFC Payload

shubham1g5:
## Summary
https://dimagi-dev.atlassian.net/browse/SC-1414

## Product Description
Apps would now be able to read/write ecrypted data on a NFC tag using a `encryption_key` specified on NFC read and write callouts. 

## Safety Assurance

- [x ] If the PR is high risk, "High Risk" label is set

### Automated test coverage

Encryption methods has test coverage as part of the earlier `encrypt-string` function. 

### Safety story

- Manual testing would be needed as part of the mobile release QA
- Did local testing with a basic app [here](https://www.commcarehq.org/a/shubhamgoyaltest/apps/view/972fd0b2174543e8a60b9fc811968e55/form/0c7e5e35eb9c4d45ab5391731400552e/source/#form/nfc_write-to-bracelet_callout)
- Backward Compatibility: Mostly verified through [test](https://github.com/dimagi/commcare-android/blob/encryptNfc/app/unit-tests/src/org/commcare/android/nfc/NfcManagerTest.java#L64). In particular, any payload without a tag could be read with a callout with no encryption_key and entity_id specified on the callout or by setting a `ignore_tag` key to `true` in the callout extras. 

cross-request: https://github.com/dimagi/commcare-core/pull/1003


app/src/org/commcare/android/nfc/NfcActivity.java
```diff
@@ -25,12 +29,17 @@
 
     protected static final String NFC_PAYLOAD_MULT_TYPES_ARG = "types";
     protected static final String NFC_PAYLOAD_SINGLE_TYPE_ARG = "type";
+    protected static final String NFC_ENCRYPTION_KEY_ARG = "encryption_key";
     protected static final String NFC_DOMAIN_ARG = "domain";
+    protected static final String NFC_ENCRYPTION_SCHEME = "encryption_aes_v1";
```
shubham1g5, line 34: @ctsims This is what I am adding to the encrypted message before writing to the tag. I didn't think case_id would always be available so decided to not incorporate it in this. Does this look ok ?

dimagimon:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Error">
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false"><a href="https://github.com/dimagi/commcare-android/blob/5cc525d273baeaac7cd0e87a0af6e0251ba9443c/app/src/org/commcare/android/nfc/NfcActivity.java#L135">app/src/org/commcare/android/nfc/NfcActivity.java#L135</a> - Overriding method should call <code>super.onNewIntent</code></td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false"><a href="https://github.com/dimagi/commcare-android/blob/5cc525d273baeaac7cd0e87a0af6e0251ba9443c/app/src/org/commcare/android/nfc/NfcActivity.java#L83">app/src/org/commcare/android/nfc/NfcActivity.java#L83</a> - Unnecessary; SDK_INT is never &lt; 16</td>
    </tr>
  </tbody>
</table>

</p>

damagatchi:
[33% (0.0%) vs master 33%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462486/)

shubham1g5:
@ctsims bumping this for review.

app/src/org/commcare/android/nfc/NfcReadActivity.java
```diff
@@ -104,6 +107,14 @@ private void readFromNfcTag(Tag tag) {
         }
     }
 
+    private String decryptValue(String message) throws EncryptionUtils.EncryptionException {
+        if (encryptionKey != null && message.startsWith(NFC_ENCRYPTION_SCHEME)) {
```
ctsims, line 111: I imagine this is fine as long as there's no chance that these prefix based tags might conflict with message contents.  

It might be worth thinking through whether there are circumstances where a prefix won't be enough later (for instance if we needed to write another key into the message, like a salt), but I can't quite come up with something that doesn't feel like over-solving the current problem

app/src/org/commcare/android/nfc/NfcActivity.java
ctsims: As small point of personal preference I'd probably think of this scheme tag not as an encryption scheme, but as a general serialization / deserialization scheme tag.

Wasn't the `case_id` thing a pretty hard part of the requirements that they really wanted in place? I think it should be possible to handle both the case where a key is available and one where it isn't. We could have `scheme_encryption_aes_v1` and `scheme_encryption_and_key_aes_v1` and fork behavior on the latter to have a delimiter encoded tag or something.

Alternatively, don't the tags themselves come with keys or buckets that we're writing to? Maybe those can be how we determine where things go

ctsims (approved):
Hey @shubham1g5 .

Code looks fine, but made some suggestions about the scheme.

In the safety story, can you also confirm how you derisked backwards compatibility? The biggest risk of the PR from my perspective is that previously written tags fail due to an unexpected issue.

app/src/org/commcare/android/nfc/NfcActivity.java
shubham1g5: >  but as a general serialization / deserialization scheme tag.

I am not sure I follow what you mean by a serialization tag. Would you be able to point me to one. 

> Wasn't the case_id thing a pretty hard part of the requirements that they really wanted in place?..

You mean add another extra to the NFC callout called "case_id" and use "scheme_encryption_and_($case_id)_aes_v1" as a scheme instead ? I was looking for a generalized key for all projects but agree we can use `case_id` if available. OOC, the benefit this will have is make the tag we are adding more unique and dynamic ?

ctsims: On your first point: Oh, sorry, I just meant the title of the field. IE: This is 'tagging' the encryption scheme, but I think its purpose is more of a general "scheme" of what we are writing, not just how we are encrypting it. IE: `NFC_ENCRYPTION_SCHEME -> NFC_PAYLOAD_SCHEME`

ctsims: On your second point:

I wasn't sure, but I thought the NFC code wrote to buckets, so I was hoping we could just write to a bucket based on the case Id, but that might not be correct.

If we can't write to a named bucket, I wasn't quite thinking what you wrote, I was thinking more that we would write one of three things to the card. I'll refer to the `case_id` as `entity_id` below, since I don't think it's necessarily specific to the case id guid.

```
[No key, No entity_id] - > $RAW_PAYLOAD
[key, no entity_id] -> scheme_aes_v1|$RAW_PAYLOAD
[key, entity_id] -> scheme_entity_aes_v1|$ENTITY_ID|$RAW_PAYLOAD
```

Alternatively if you wanted a universal scheme, I think if we had delimiting, you could just leave the entity id blank unless it was present.

```
[No key, No entity_id] - > $RAW_PAYLOAD
[key, no entity_id] -> scheme_aes_v1||$RAW_PAYLOAD
[key, entity_id] -> scheme_aes_v1|$ENTITY_ID|$RAW_PAYLOAD
```

We'd need a delimiter that we had confidence couldn't be a part of the entity id (Could just throw an exception if it's in there) or a part of the base64 character set.

shubham1g5: @ctsims Added this in https://github.com/dimagi/commcare-android/pull/2502/commits/1a299ea0457ee729c23800cc04108bfac01ed4ca

app/unit-tests/src/org/commcare/android/nfc/NfcManagerTest.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Test
+    public void payloadWithoutTagTest() throws EncryptionUtils.EncryptionException {
+        // Decrypt an old payload without specifying encryptionKey and entityId
+        assert new NfcManager(CommCareTestApplication.instance(), "", "")
+                .decryptValue(PAYLOAD).contentEquals(PAYLOAD);
+
+        // Decrypt an old payload with some encryptionKey and entityId
+        assert new NfcManager(CommCareTestApplication.instance(), ENCRYPTION_KEY, ENTITY_ID)
```
shubham1g5, line 70: @ctsims This test currently fails with a `InvalidPayloadTagException`. I am unsure what should be right behaviour here when you try to read a payload written on old CC versions with a NFC callout that has specified an `encryptionKey` and `entityID`. Is it a right expectation to set that if you specify one of `encryptionKey` and `entityID`, you expect to only read payloads that were written with these exact params in the write callout.

ctsims: As per our conversation, I think it's fine to either fail in the case you are describing or to gracefully update a written entity which doesn't have a version.

I definitely think if someone has a record written with an `entityID` though, that if there's another record with a different entity ID we should have clearly documented expectations about the failure mode.

shubham1g5: Does [this](https://github.com/dimagi/commcare-android/pull/2502/commits/a38384d516e428e688026610c62c7730b2175424) like an ok upgrade pathway ?

> I definitely think if someone has a record written with an entityID though, that if there's another record with a different entity ID we should have clearly documented expectations about the failure mode.

Can you elaborate a little more on this. Today this case will fail as long as the `ignore_tag` parameter introduced in last commit is `false` but user have an option to read the whole payload as it is when `ignore_tag` is set to `true`.

ctsims:
@shubham1g5 Hm, I think the name on the new behavior ("ignore tag") might be a bit confusing, since I don't think it would be desirable behavior to ignore a *mismatched* tag. 

I think I'd either label the tag as "allow_v0_read" or "allow_untagged_read" something that makes it clearer that what it's allowing is reading a payload from the previous, unprotected format, rather than ignoring the tag protection. 

Otherwise, the behavior looks good to me

damagatchi:
[33% (0.0%) vs master 33%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462540/)

ctsims (approved):

shubham1g5:

damagatchi:
[33% (0.0%) vs master 33%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462542/)

PR #2506, opened 2021-06-28
Moves AndroidCaseInstanceTreeElement to core

shubham1g5:
## Summary

- No-op refactor to merge duplicate code between mobile and formplayer(https://github.com/dimagi/formplayer/pull/981)

## Safety Assurance

### Automated test coverage

There is test coverage around case db queries already. 

### Safety story

- Doesn't result in any effective code change, just moves the code to super class

cross-request: https://github.com/dimagi/commcare-core/pull/1008


dimagimon:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

damagatchi:
[33% (0.0%) vs master 33%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462511/)

ctsims:
Hey @shubham1g5 

One thing - I think this set of PR's needs a bit of further de-risking.

It is true that there are unit tests for the area, and those tests do trigger the optimizations currently, but the tests don't technically ensure that the optimizations actually fire since they are quite deep in the code. They will catch an error where an optimization's code breaks the evaluation entirely, or produces the wrong result, but they won't catch an issue which results in an optimization no longer being applied. 

The easiest way to derisk, I think, would be to make sure the tests output a  [Reduced Trace Report, as in this example](https://github.com/dimagi/commcare-core/blob/d9066ea561f4ce85717b25263f9ddad695c98b29/src/main/java/org/commcare/cases/entity/NodeEntityFactory.java#L98) and manually inspect it to make sure the query plan didn't change before and after the tests. We could in theory hard code some aspect of the query plan into an actual testable condition  as well, which we've avoided in the past because it makes the tests dependent on the trace reporter, but might be worth revisiting now.

shubham1g5:
@ctsims Added to tests in https://github.com/dimagi/commcare-core/pull/1006/commits/e13dd2da349e32efde3039ba5bd82e5b59ae287b to test for optimizations. I am simply testing for the the trace to include a stroage lookup text, not sure if that's what you were referring to as well or if there is a better way.

shubham1g5:
checking if the test [here](https://github.com/dimagi/commcare-core/commit/e13dd2da349e32efde3039ba5bd82e5b59ae287b) looks good on this.

ctsims (approved):
Yup, I think that test should cover my concerns, thanks!

snopoke (approved):

PR #2508, opened 2021-07-01
Adds daterange picker for case search

shubham1g5:
## Summary
https://dimagi-dev.atlassian.net/browse/SC-1465

Adds a widget to select a date range in case search

## Feature Flag

https://www.commcarehq.org/hq/flags/edit/search_claim/

## Product Description

Allows user to search a case property in a date range. 

Clicking edit in the query screen opens up the date picker as shown below - 

HQ changes were merged earlier in https://github.com/dimagi/commcare-hq/pull/29035 

## Safety Assurance

### Automated test coverage

Added test around date conversion utility used for the widget which handles major functionality around the datepicker. Would rely on Manual QA for end to end testing on this as datepicker range is proving to be difficult to test with robolectric. 

### Safety story

Tested locally, would go through manual QA as well,
Impact if any would largely be limited to the apps using the case claim with the daterange widget. To use the datepicker widget I had to change the app theme to a [Material Component Bridge theme](https://github.com/material-components/material-components-android/blob/master/docs/getting-started.md#bridge-themes). Though this should not cause any UI changes, there can be minor UI sideeffects elsewhere in app.

### DIFF
--- a/app/res/values/themes.xml
+++ b/app/res/values/themes.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources xmlns:android="http://schemas.android.com/apk/res/android">
 
-    <style name="CommonTheme" parent="Theme.AppCompat.Light.DarkActionBar">
+    <style name="CommonTheme" parent="Theme.MaterialComponents.Light.DarkActionBar.Bridge">
         <item name="colorPrimary">@color/cc_brand_color</item>
         <item name="colorPrimaryDark">@color/cc_brand_text</item>
         <item name="colorAccent">@color/cc_dark_cool_accent_color</item>
@@ -31,6 +31,11 @@
         <item name="android:scrollViewStyle">@style/fading_edge</item>
         <item name="android:windowContentOverlay">@null</item>
         <item name="android:dropDownListViewStyle">@style/dropDownListView</item>
+
+        <!--  Material Components styling that can be removed when we migrate to a material component theme from the bridge theme-->
+        <item name="materialCalendarStyle">@style/Widget.MaterialComponents.MaterialCalendar</item>
+        <item name="materialCalendarFullscreenTheme">@style/ThemeOverlay.MaterialComponents.MaterialCalendar.Fullscreen</item>
+        <item name="materialCalendarTheme">@style/ThemeOverlay.MaterialComponents.MaterialCalendar</item>
     </style>
 
     <style name="AppBaseTheme" parent="CommonTheme" />
--- a/app/src/org/commcare/utils/DateRangeUtils.java
+++ b/app/src/org/commcare/utils/DateRangeUtils.java
@@ -7,6 +7,8 @@ import java.util.Locale;
 
 import javax.annotation.Nullable;
 
+import androidx.core.util.Pair;
+
 /**
  * Utility functions for DateRangePicker widget
  */
@@ -24,14 +26,14 @@ public class DateRangeUtils {
      * @throws ParseException if the given humanReadableDateRange is not in 'yyyy-mm-dd to yyyy-mm-dd' format
      */
     @Nullable
-    public static androidx.core.util.Pair<Long, Long> parseHumanReadableDate(String humanReadableDateRange) throws ParseException {
+    public static Pair<Long, Long> parseHumanReadableDate(String humanReadableDateRange) throws ParseException {
         if (humanReadableDateRange.contains(DATE_RANGE_ANSWER_HUMAN_READABLE_DELIMITER)) {
             String[] humanReadableDateRangeSplit = humanReadableDateRange.split(DATE_RANGE_ANSWER_HUMAN_READABLE_DELIMITER);
             if (humanReadableDateRangeSplit.length == 2) {
                 SimpleDateFormat sdf = new SimpleDateFormat(DATE_FORMAT, Locale.US);
                 Date startDate = sdf.parse(humanReadableDateRangeSplit[0]);
                 Date endDate = sdf.parse(humanReadableDateRangeSplit[1]);
-                return new androidx.core.util.Pair<>(getTimeFromDateOffsettingTz(startDate), getTimeFromDateOffsettingTz(endDate));
+                return new Pair<>(getTimeFromDateOffsettingTz(startDate), getTimeFromDateOffsettingTz(endDate));
             }
         }
         throw new ParseException("Argument " + humanReadableDateRange + " should be formatted as 'yyyy-mm-dd to yyyy-mm-dd'", 0);
--- a/app/src/org/commcare/utils/DateRangeUtils.java
+++ b/app/src/org/commcare/utils/DateRangeUtils.java
@@ -1,0 +1,75 @@
+package org.commcare.utils;
+
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Locale;
+
+import javax.annotation.Nullable;
+
+/**
+ * Utility functions for DateRangePicker widget
+ */
+public class DateRangeUtils {
+
+    // Changing this will require changing this format on ES end as well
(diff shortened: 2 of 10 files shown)


app/res/values/themes.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources xmlns:android="http://schemas.android.com/apk/res/android">
 
-    <style name="CommonTheme" parent="Theme.AppCompat.Light.DarkActionBar">
+    <style name="CommonTheme" parent="Theme.MaterialComponents.Light.DarkActionBar.Bridge">
```
ShivamPokhriyal, line 4: Nice, `Bridge` is the reason why it's working on pre-lollipop right?

app/src/org/commcare/utils/DateRangeUtils.java
```diff
@@ -0,0 +1,75 @@
...
+                SimpleDateFormat sdf = new SimpleDateFormat(DATE_FORMAT, Locale.US);
+                Date startDate = sdf.parse(humanReadableDateRangeSplit[0]);
+                Date endDate = sdf.parse(humanReadableDateRangeSplit[1]);
+                return new androidx.core.util.Pair<>(getTimeFromDateOffsettingTz(startDate), getTimeFromDateOffsettingTz(endDate));
+            }
+        }
+        throw new ParseException("Argument " + humanReadableDateRange + " should be formatted as 'yyyy-mm-dd to yyyy-mm-dd'", 0);
+    }
+
+    private static Long getTimeFromDateOffsettingTz(Date date) {
```
ShivamPokhriyal, line 40: I'm not sure what it returns? Can you see whether a similar utility exists in DataUtils?

ShivamPokhriyal (approved):
Looks good to me.

ctsims:
Hi Shubham

I think we should do some updates on the documentation for search and claim? This PR raised some questions for me about the available inputs and how the structures work and I realized that I don't think it's documented super clearly since we've been rolling it out more on the experimental stream. 

Can you do a pass on adding some details (I'm presuming to the `commcare-core` wiki) about the search and claim process and the options? I might ask for some quick clarifications on process there as well afterwards.

shubham1g5:
@ctsims USH has quite thorough [docs](https://confluence.dimagi.com/display/USH/Case+Search+Configuration) on each of these widgets. Do you mean adding more dev-level docs on it ?

app/res/values/themes.xml
shubham1g5: No not really, Bridge theme is just a workaround for apps not comfortable with changing their theme to a MaterialComponent theme. Bridge theme inherits from the AppCompat Theme and adds any properties defined by the MaterialComponent theme in it. So it has nothing to do with device compatibility. The library support the 14+ devices by default.

app/src/org/commcare/utils/DateRangeUtils.java
shubham1g5: I wasn't able to find it anywhere else. It removes the timezone offset from the date in millis and return the date. This is because when we set the date to the daterange widget we are setting time in milliseconds without any indication of timezone and the date widget itself doesn't offset the timezone from it to arrive at the offsetted date.

ctsims:
> @ctsims USH has quite thorough [docs](https://confluence.dimagi.com/display/USH/Case+Search+Configuration) on each of these widgets. Do you mean adding more dev-level docs on it ?

Yeah, exactly. I think the internal docs are good for usage, but we should also have lower level documentation about the code, XML, and baseline structures outside of the broader use case. Those docs are more useful than the implementation docs for building test plans and understanding the discrete function of the individual components.

shubham1g5:
@ctsims  Added the missing attributes for case search in [this remote-request section](https://github.com/dimagi/commcare-core/wiki/Suite20#remote-request). Let me know if you are looking for something more detailed on this.

ctsims:
Thanks! This is definitely really helpful.

I think the documentation was still too oriented towards the feature definition rather than providing the specification of the underlying behavior. I took a shot at [updating the spec docs](https://github.com/dimagi/commcare-core/wiki/Suite20#remote-request) in a way that is a bit more oriented at explaining the spec structure, can you take a look and see if it looks right to you?

I did test to confirm that `<query>` steps work in arbitrary session templates (like an `<entry>`), one thing I wasn't quite clear on was the notes I mentioned about the `default_search` parameter. 

Namely: 
* Are `<prompt>` defaults included in the autosearch, or just `<data>` pairs?
* Is it correct that navigating "back" into an autosearch query should bring up the prompts?

shubham1g5:
That does look great! 

> Are <prompt> defaults included in the autosearch, or just <data> pairs?

It's just the <data> pairs. We regard <prompt> defaults as visible user inputs as well and treat them more as defaults for user input rather than defaults for search. 

> Is it correct that navigating "back" into an autosearch query should bring up the prompts?

No. Navigating back from here would take you to whatever screen you came from. Though you can add another action from the search results screen (Search Again button) to take you to prompt screen. This is bit of a hack as the search again action points to the same `remote-request` and in turn same `query` with `default_search` set as `true` but the Web Apps recognize that this was an action performed from the Search Results screen and forcefully override the  default_search using the [forceManualAction](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/application/MenuController.java#L58) param.  We basically couldn't come up with a solution that can put this state in suite file instead at that time. 

Did an update to the doc to remove these two specifications.

ctsims:
Thanks for those updates, that's super helpful context. This looks good to me for now.

I do think there's one additional piece missing that we should put on our roadmap to address. If I'm consider either the audience of someone who was trying to write their own engine, or someone who was trying to write their own <query> receiver, I don't think they have enough context to go on. I think the technical contract about the Query process does need one extra piece of documentation, which is what the literal request / response cycle actually looks like. 

As an example, I think the way the date range widget is implemented now is still a little ambiguous. We do mention the string pattern in the docstring, but we don't' really discuss date serialization formats, or whether the ranges are inclusive or exclusive, etc. I think this page is probably not the right level to be documenting those details, though, and we should probably create a separate specification (there might be a starting point somewhere already) in the wiki about the end-to-end process.

ctsims (dismissed):

shubham1g5:
That makes sense to me. I added it on [USH backlog](https://dimagi-dev.atlassian.net/browse/USH-1172) and will probably take a pass on it soon.

shubham1g5:

dimagimon:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

damagatchi:
[33% (0.0%) vs master 33%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462537/)

ctsims (approved):

shubham1g5:
@ctsims I am trying to put in a more detailed wiki on remote request workflow [here](https://github.com/dimagi/commcare-core/wiki/Remote-Request-Query-and-Post). It would be good to get a brief sense on pointers this doc should cover from your perspective.

ctsims:
Hi Shubham,

Looks like a good start, although I might suggest a slightly different approach.

I think it might make the most sense to retain the raw specs as part of the suite XML doc, and focus in this doc on highlighting how they plug together to build the overall user interaction, linking back to the other doc to provide pointers to the full spec.

In that doc instead of spec XML, you could provide some samples for reference, and maybe also provide some more detailed descriptions of the various configuration details, and how they influence the baseline workflow you outline to describe the feature.

shubham1g5:
@ctsims  thanks did an iteration on the [same](https://github.com/dimagi/commcare-core/wiki/Remote-Request-Query-and-Post-Configurations) based on your advice. This page now covers all the different possible workflows. Is there any other information that you think I should add to it ?

ctsims:
This aligns well with what I was thinking about, thanks for the iteration!

PR #2513, opened 2021-07-08
Map fix

ShivamPokhriyal:
## Summary

Fix for https://dimagi-dev.atlassian.net/browse/QA-3258

Basically, 3 issues: 
- currentLocation button click wasn't working properly, because we were relying on the underlying `onLocationChange` event which could take time to trigger. 
- improvements to permissions. We don't need to ask for permissions here since Kujaku handles it internally. Kujaku requests for `ACCESS_FINE_LOCATION` https://github.com/onaio/kujaku/blob/master/library/src/main/java/io/ona/kujaku/helpers/PermissionsHelper.java#L22
- While replacing location, mapview was automatically moving to current location because of `onLocationChange` event. We set `isManualSelection` to `true` to avoid this when we are replacing an already selected location. 

## Safety Assurance

### Safety story
 
Will be QAed

### DIFF
--- a/app/src/org/commcare/gis/MapboxLocationPickerActivity.kt
+++ b/app/src/org/commcare/gis/MapboxLocationPickerActivity.kt
@@ -6,6 +6,7 @@ import android.content.IntentSender.SendIntentException
 import android.location.Location
 import android.os.Bundle
 import android.view.View
+import android.widget.Toast
 import androidx.core.content.ContextCompat
 import androidx.lifecycle.ViewModelProvider
 import com.google.android.gms.common.api.ResolvableApiException
@@ -32,6 +33,7 @@ import org.commcare.location.CommCareLocationListener
 import org.commcare.utils.GeoUtils
 import org.commcare.utils.Permissions
 import org.commcare.views.widgets.GeoPointWidget
+import org.javarosa.core.services.locale.Localization
 
 class MapboxLocationPickerActivity : BaseMapboxActivity(), CommCareLocationListener {
 
@@ -59,6 +61,10 @@ class MapboxLocationPickerActivity : BaseMapboxActivity(), CommCareLocationListe
             // Add marker.
             updateMarker(point)
             viewModel.reverseGeocode(point)
+        } else {
+            Toast.makeText(this,
+                    Localization.get("permission.location.denial.message"),
+                    Toast.LENGTH_LONG).show()
         }
         true
     }
--- a/app/src/org/commcare/gis/MapboxLocationPickerActivity.kt
+++ b/app/src/org/commcare/gis/MapboxLocationPickerActivity.kt
@@ -82,6 +82,7 @@ class MapboxLocationPickerActivity : BaseMapboxActivity(), CommCareLocationListe
                     .zoom(10.0)
                     .tilt(20.0)
                     .build()
+            isManualSelectedLocation = true
         }
         mapView.showCurrentLocationBtn(false)
     }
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/gis/MapboxLocationPickerActivity.kt
```diff
@@ -57,10 +54,12 @@ class MapboxLocationPickerActivity : BaseMapboxActivity(), CommCareLocationListe
     private var symbolManager: SymbolManager? = null
     private var symbol: Symbol? = null
     private val mapClickListener = MapboxMap.OnMapClickListener { point ->
-        isManualSelectedLocation = true
-        // Add marker.
-        updateMarker(point)
-        viewModel.reverseGeocode(point)
+        if (!Permissions.missingAppPermission(this, LOCATION_PERMISSIONS)) {
```
shubham1g5, line 57: Why do we need the check for location permission for manual selection ?

dimagimon:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

app/src/org/commcare/gis/MapboxLocationPickerActivity.kt
ShivamPokhriyal: We don't want to respond to map click events if the location permissions haven't been granted. So this checks prevents us from doing that.

shubham1g5: > We don't want to respond to map click events if the location permissions haven't been granted 

I am not totally sure that's a right statement or if there is a valid reasoning behind that. Location permissions allows  phone to access to the current location of the user and here itself is choosing the location on map. User should be allowed to mark his current location without giving phone/CC access to his location and not doing so wouldn't be a good citizen Android policy as we are forcing user to give CC access to his location even when he doesn't want/need to.

ShivamPokhriyal: >Location permissions allows phone to access to the current location of the user and here itself is choosing the location on map.

Without the location permission, the map load event won't take the user to his current location. It would rather be a whole map and then user will have to zoom in and navigate all the way to their current location. Is that something we should support? 
Wouldn't that cause issues if a user clicks on any weird location and simply confirms it(without verifying that it's a correct location), and then later on project would complain that the captured locations aren't correct.

shubham1g5: That's a good point. Do we know what's the behaviour for normal map activity. Probably best to be consistent with that here.

ShivamPokhriyal: I checked the normal map activity. It finishes if a user denies location permission. 
I think this is gonna be tricky to do with mapbox activity because the `KujakuMapView` is handling the location permissions internally. So if we use our own permissions check outside, then there will be 2 permission popups back to back which will be a weird user experience.

ShivamPokhriyal:
Got the browserstack api failure issue again! I'm gonna file a support ticket to them. 

app/src/org/commcare/gis/MapboxLocationPickerActivity.kt
shubham1g5: sounds good, though we should atleast show a message to the user here if there is no permisson given otherwise user will keep clicking the map without any UI feedback

ShivamPokhriyal: Yup, that sounds good to me. Will display a toast.

shubham1g5 (approved):

damagatchi:
[33% (0.0%) vs master 33%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462530/)

PR #2515, opened 2021-07-09
Handle graph labels more carefully

ShivamPokhriyal:
## Summary

cross-request: https://github.com/dimagi/commcare-core/pull/1015

A fix for https://dimagi-dev.atlassian.net/browse/QA-3223

Updated c3 and d3. Use different character limit for different screen sizes.

## Safety Assurance

### Safety story
 
Will be QAed

Update to Product Notes: 
- We improved the graph UI to rotate the x labels by 75 degrees and at the same time added a limit of 15 characters to the label.


dimagimon:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

damagatchi:
[33% (0.0%) vs master 33%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462533/)

shubham1g5 (changes requested):
What if we set the height to 0 ? The graph looks just fine to me when I do that. 

Also I don't think it's a good idea to restrict labels as lot of projects also use tabs for their workflow where I am guessing larger labels won't be an issue. We should  probably just add a hint text on HQ as an advice to keep labels short without hard pressing projects to use shorter labels. If we need to retain the limit, we should make it variable based on screen size.

ShivamPokhriyal:
> What if we set the height to 0 ? The graph looks just fine to me when I do that.

We're based on an old version so height = 0 doesn't work. I tried upgrading to the latest version(I couldn't find a migration guide), and was able to resolve a few issues but they just kept popping up so I didn't go forward with that. 

> We should probably just add a hint text on HQ as an advice to keep labels short without hard pressing projects to use shorter labels

I don't really think it's always possible for projects to define their own labels. And it's probably not in their control to limit the number of characters. 

> If we need to retain the limit, we should make it variable based on screen size.
This sounds good, I'll see if we can do this.

shubham1g5:
> I don't really think it's always possible for projects to define their own labels. And it's probably not in their control to limit the number of characters.

I see, what's the reason behind that ? I though one can define labels on HQ.

ShivamPokhriyal:
> I see, what's the reason behind that ? I thought one can define labels on HQ.

Yeah, but they won't always use that custom configuration. The most common use-case would be to setup a report using some case property, without providing custom configurations. Now, the case-property in itself can be of any character length.  
In order to impose restrictions, they'd have a create another case property which would have first 15 chars.

ShivamPokhriyal:
cross-request: https://github.com/dimagi/commcare-core/pull/1015

app/assets/graphing/d3.min.js
```diff
@@ -1,5 +1,2 @@
...
...a=[];o=n.pop();)if(t(o),e=o.children)for(r=0,i=e.length;r<i;++r)a.push(e[r])}while(a.length);return this},eachAfter:function(t){for(var n,e,r,i=this,o=[i],a=[];i=o.pop();)if(a.push(i),n=i.children)for(e=0,r=n.length;e<r;++e)o.push(n[e]);for(;i=a.pop();)t(i);return this},eachBefore:function(t){for(var n,e,r=this,i=[r];r=i.pop();)if(t(r),n=r.children)for(e=n.length-1;e>=0;--e)i.push(n[e]);return this},sum:function(t){return this.eachAfter(function(n){for(var e=+t(n.data)||0,r=n.children,i=r&&r.length;--i>=0;)e+=r[i].value;n.value=e})},sort:function(t){return this.eachBefore(function(n){n.children&&n.children.sort(t)})},path:function(t){for(var n=this,e=function(t,n){if(t===n)return t;var e=t.ancestors(),r=n.ancestors(),i=null;for(t=e.pop(),n=r.pop();t===n;)i=t,t=e.pop(),n=r.pop();return i}(n,t),r=[n];n!==e;)n=n.parent,r.push(n);for(var i=r.length;t!==e;)r.splice(i,0,t),t=t.parent;return r},ancestors:function(){for(var t=this,n=[t];t=t.parent;)n.push(t);return n},descendants:function(){var t=[];return this.each(function(n){t.push(n)}),t},leaves:function(){var t=[];return this.eachBefore(function(n){n.children||t.push(n)}),t},links:function(){var t=this,n=[];return t.each(function(e){e
```
shubham1g5, line 2: Assuming that both c3 and d3 files are direct replacements from the source and we have made any changes on top on them right ?

shubham1g5 (approved):
Thanks for getting this update done. Can you comment on if QA has good test coverage on graphs to test these updates ? We should make sure that this PR go through some extensive graph tests bfore getting released.

app/assets/graphing/d3.min.js
ShivamPokhriyal: Yeah, Just copied the file from the zipped artifacts.

ShivamPokhriyal:
> Thanks for getting this update done. Can you comment on if QA has good test coverage on graphs to test these updates ? We should make sure that this PR go through some extensive graph tests before getting released.

We don't have currently. That's also one reason why I wasn't going for updating in the first place. But I think since we're making changes, we should also take this opportunity to update. 
QA doesn't have a good test plan right now, but I can talk to QA team to add more tests around graphs.

shubham1g5:
Definitely agree that update is the right decision. Though if we don't have good test coverage I don't feel comfortable on releasing this without testing. When you talk to QA, I think we have two options - 

1. We merge this and add the test coverage in QA for this for 2.52
2. If QA don't have the space right now, We rollback the graph changes for this release and push this PR to 2.53 instead.

ShivamPokhriyal:

damagatchi:
[33% (0.0%) vs master 33%](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/462538/)

PR #2546, opened 2022-02-24
Adds ability to not exit after form submission on external launch

shubham1g5:
## Summary

Adds Intent argument 'ccodk_exit_after_form_submission' to control exit from external launch. 

## Safety Assurance


dimagimon:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

app/src/org/commcare/activities/DispatchActivity.java
```diff
@@ -333,23 +334,26 @@ private void handleUnvalidatedApp() {
...
-                        i.putExtra(WAS_EXTERNAL, true);
+                        i = new Intent(this, StandardHomeActivity.class);
                         i.putExtra(SESSION_ENDPOINT_ID, sessionEndpointId);
                         i.putExtra(SESSION_ENDPOINT_ARGUMENTS_BUNDLE, args);
                         i.putStringArrayListExtra(SESSION_ENDPOINT_ARGUMENTS_LIST, argsList);
+                    }
+                    if (i != null) {
+                        i.putExtra(WAS_EXTERNAL, true);
+                        i.putExtra(EXIT_AFTER_FORM_SUBMISSION,
+                                getIntent().getBooleanExtra(EXIT_AFTER_FORM_SUBMISSION, true));
```
ctsims, line 356: It's a small thing, but the 'true' default here is essentially set twice - once here and once in the base activity below. 

I think that could create some confusion later if someone though one of these two was the 'authoritative' default and didn't realize it was also getting defaulted upstream. 

Maybe it would be a little cleaner to check whether `EXIT_AFTER_FORM_SUBMISSION` exists in the calling intent and only set it here if so, to prevent the fallback and let the default in the `HomeScreenBaseActivity` be the only one?

ctsims (approved):

shubham1g5:

app/src/org/commcare/activities/DispatchActivity.java
shubham1g5: So like this - ?
```
if(getIntent().hasExtra(EXIT_AFTER_FORM_SUBMISSION)) {
                            i.putExtra(EXIT_AFTER_FORM_SUBMISSION,
                                    getIntent().getBooleanExtra(EXIT_AFTER_FORM_SUBMISSION, true));
                        }
```

even if we check for the extra, `getIntent().getBooleanExtra` requires a fallback. Would that be alright ?

shubham1g5: That does make sense. Implemented [here](https://github.com/dimagi/commcare-android/pull/2551)

PR #2553, opened 2022-03-16
Catch error and show message to user

Charl1996:
## Summary
[Ticket](https://dimagi-dev.atlassian.net/browse/SC-1956)

This PR simply adds the `XPathException` to the `refreshView` in `FormHierarchyActivity.java`. I have followed the same logic as the exception that is already being caught by showing a toast with the error.

Note:
I could not reproduce the _exact_ scenario error as of yet (although still trying to); I have followed [this](https://dimagi-dev.atlassian.net/browse/SC-1956?focusedCommentId=186043) comment, but it is still not reaching the code I included as it shows the error earlier on during the session before this code is reached. I have nonetheless decided to create this PR in the intrim, since I could find the problem area based on the ticket description, although I am wondering if it's necessary since an error message is already shown to the user as in the case of [this](https://dimagi-dev.atlassian.net/browse/SC-1956?focusedCommentId=186043) scenario (unless the particular issue can be reproduced in a different way that would result in the same error in the ticket description).

## Feature Flag
No specific FF

## Safety Assurance

### Automated test coverage
No automated tests added.

### Safety story
See Summary.


dimagimon:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

ctsims:
Hi @Charl1996, happy first PR! 🎈 

I think it would be worth continuing to try and replicate the issue, because now that I've reviewed the code paths looks as though we probably want to make a couple of adjustments to the process here, and making sure the error handling is smooth would be really helpful.

For context, the [previous error handling path ](https://github.com/dimagi/commcare-android/pull/2553/files#diff-9af3b8ae14cabd949c0f2013e3cf6f7ad9552671dc445d7ecf9f705049662487R169) is capturing an overly specific XPath exception `XPathTypeMismatch` rather than a more generic one `XPathException` for legacy reasons. 

When [this code was introduced](https://github.com/dimagi/commcare-android/pull/1078/files), these were the standard exceptions to catch, but since then the other exception handling paths have become generic. [The current state](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/activities/FormEntryActivity.java#L488) of the "peer" exception handlers now use `XPathException` as the catch, so I think we can simply open up the handled exception, rather than adding a separate catch.

Now that I've seen how the other error handling paths have changed since PR #1078 introduced this code, though, I think we should go one step further by replacing the callout to a Toast (which is a transient, less-than-ideal way to communicate errors to users) with the same [UserfacingErrorHandling.logErrorAndShowDialog()](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/activities/FormEntryActivity.java#L489) mechanism that's now used by peer error handling paths. That method also handles the XPath Exception logging on line 169, so we can eliminate that line as well.

It would be good to ensure that those broader changes actually work, though, so getting to the point of replicating the issue will be important. You mentioned that you tried the process in the attached PR, but that it didn't work, can you elaborate on what did happen when you gave that a shot?

damagatchi:

Charl1996:
I have actually made some local changes after the comments on this PR which I need to test still; I'll prioritise it this week.

avazirna:

avazirna:

avazirna:

avazirna:

Charl1996:
I tried hitting the error path in this PR, but instead got [this error](https://dimagi-dev.atlassian.net/browse/SC-1956?focusedCommentId=245049).

Charl1996:
After doing the following steps I managed to hit the error path being addressed in this PR:
- (web) Create lookup table
- (mobile) Fill out incomplete form
- (web) Remove lookup table
- (mobile) Full sync
- (mobile) Navigate to incomplete form

Here’s a screenshot of what the caught XpathException looks like for the user:

error

After this the user is redirected to the screen where he/she chooses a fresh form to fill out. Is this OK? Personally I would expect it to simply redirect "back", i.e. to where the user would choose between which incomplete form to fill out, but open to suggestions.

shubham1g5:
@Charl1996 Great that we finally have a relibale replication on this. Reg. the screenshot, I think we would want to replace the transient toast message with a more permanent error dialog [like here](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/activities/FormEntryActivity.java#L489).

app/src/org/commcare/activities/FormHierarchyActivity.java
```diff
@@ -165,13 +181,18 @@ private void refreshView() {
         String hierarchyPath;
         try {
             hierarchyPath = FormHierarchyBuilder.populateHierarchyList(this, formList);
-        } catch (XPathTypeMismatchException e) {
+        } catch (XPathException e) {
             XPathErrorLogger.INSTANCE.logErrorToCurrentApp(e);
 
-            final String errorMsg = "Encounted xpath error: " + e.getMessage();
-            Toast.makeText(this, errorMsg, Toast.LENGTH_LONG).show();
-            setResult(RESULT_XPATH_ERROR);
```
Charl1996, line 173: I removed this in favour of [this line](https://github.com/dimagi/commcare-android/blob/7a683f0ff101f3fd2a551ff2dcd44a0d6b7a4b45/app/src/org/commcare/views/UserfacingErrorHandling.java#L86), but [this code](https://github.com/dimagi/commcare-android/blob/7a683f0ff101f3fd2a551ff2dcd44a0d6b7a4b45/app/src/org/commcare/activities/FormEntryActivity.java#L355) suggests that the `RESULT_XPATH_ERROR` might still be needed, or I need to change it to `RESULT_CANCELED`?

Charl1996:
I updated the code. See screenshot for results:

image

app/src/org/commcare/views/UserfacingErrorHandling.java
```diff
@@ -64,6 +72,35 @@ public static void createErrorDialog(final CommCareActivity activity, String err
                     activity.dismissAlertDialog();
                 };
         factory.setPositiveButton(buttonDisplayText, buttonListener);
-        activity.showAlertDialog(factory);
+        return factory;
+    }
+
+    public static StandardAlertDialog getErrorDialog(final FormHierarchyActivity activity, String errorMsg,
```
shubham1g5, line 78: @Charl1996 your code looks right to me though looks like we can optimise the design here to avoid code duplication like this. The way to do it would be to refactor this class to take a generalised parameterised object rather than a concrete class like `CommCareActivity` or `FormHierarchyActivity`. Here is an example of how we can do this - 

1. Create a new interface for alert dialog methods used in this class - 

````
public interface AlertDialogController {

    /**
     * Show the alert dialog provided by the given AlertDialogFactory
     */
    void showAlertDialog(CommCareAlertDialog dialog);

    /**
     * Dismiss the current alert dialog
     */
    void dismissAlertDialog();
}

````

These methods are also present in `DialogController` . So we can modify it by removing these methods from `DialogController` and have it extend `AlertDialogController` 

`public interface DialogController extends AlertDialogController {`



2. Next thing we need to is utilise our new interface `AlertDialogController` in `UserfacingErrorHandling`. It seems best to do it by using a parametrised argument.

````
public class UserfacingErrorHandling<T extends AppCompatActivity & AlertDialogController> {
 
 public void logErrorAndShowDialog(final T dialogActivity,
                                             XPathException exception,
                                             final boolean shouldExit)
}
````
we will need to change methods in this class from `static` to `non static` to use parametrised arguments. That means we will need to modify the existing calls to these functions as - 

`new UserfacingErrorHandling().logErrorAndShowDialog(this, e, true);` 

that is initialise the instance of `UserfacingErrorHandling` before calling it.


Apologies that we are getting into a bit of design refactoring as part of this ticket when it was supposed to be minor work but think it's going to be good learning and intro to some of the Java code design.

**Comment by Charl1996** [comment]
@shubham1g5 This was really my first time working with Java's generic types. Just 2 questions:
1. When specifying a bounded generic type like `<T extends A & B>`, does that mean the argument type _must_ be a subclass/interface of both A and B, or only from either? From general programming knowledge the `&` suggests "and", but I may be wrong in this scenario.
2. From what I gather on the web, the diamond operator is used for better type inference by the compiler. I've seen examples where it's used to simplify the following,
`Map<String, Map<Integer>> = Map<Map<String, Map<Integer>>>();`
into 
`Map<String, Map<Integer>> = Map<>();`

I get that, but is it necessary in the [following case](https://github.com/dimagi/commcare-android/blob/4e3cc04e73ba81e218c0a3b35d2e8db1dc05a2e9/app/src/org/commcare/activities/EntityDetailActivity.java#L284)?
`new UserfacingErrorHandling<>().logErrorAndShowDialog(this, xe, true);`

**Comment by shubham1g5** [comment]
@Charl1996 Nice work on this! 

1. Both A and B
2. You should be able to omit the dimond operator in this case and simply use `new UserfacingErrorHandling().logErrorAndShowDialog(this, xe, true)` because Java can infer the cast in this case from the method parameter itself.

File: app/src/org/commcare/activities/FormHierarchyActivity.java [hunk 3]
```diff
@@ -156,6 +162,22 @@ public static String getCurrentPath() {
         return path.substring(0, path.length() - 2);
     }
 
+    @Override
+    public void showAlertDialog(CommCareAlertDialog dialog) {}
```
shubham1g5, line 166: we should call the code to show dialog here

app/src/org/commcare/activities/FormHierarchyActivity.java
```diff
@@ -165,13 +187,18 @@ private void refreshView() {
...
             XPathErrorLogger.INSTANCE.logErrorToCurrentApp(e);
 
-            final String errorMsg = "Encounted xpath error: " + e.getMessage();
-            Toast.makeText(this, errorMsg, Toast.LENGTH_LONG).show();
-            setResult(RESULT_XPATH_ERROR);
-            finish();
+            final String title = StringUtils.getStringRobust(this, org.commcare.dalvik.R.string.error_occured);
+            final String errorMsg = e.getMessage();
+
+            AlertDialogFragment.fromCommCareAlertDialog(
```
shubham1g5, line 196: You should now be able to modify this to just call `UserfacingErrorHandling(). logErrorAndShowDialog` here instead of showing the dialog manually.

shubham1g5 (changes requested):

Charl1996:
@shubham1g5 I see in the CommCareActivity class the `showAlertDialog` method has a check for [areFragmentsPaused](https://github.com/dimagi/commcare-android/blob/fc06cedfd0f75c4038060bf1f84eb165d5ce5ae0/app/src/org/commcare/activities/CommCareActivity.java#L669); under what circumstances would the framents be "paused" and is this something we should check [here](https://github.com/dimagi/commcare-android/blob/4e7d5612c9ed01057c662603d1908226702bf6a1/app/src/org/commcare/activities/FormHierarchyActivity.java#L167)?

shubham1g5:
@Charl1996 Sorry forgot to reply earlier. That handling is for when the error happens when the Application is in background. Fragments paused here just means that the app is in background and we should not show the error dialog to the user because our app is in background. It's required for error handling around asynchronous tasks that runs in background and can error even when user is not on CommCare app. For this case since the error is happening from main UI thread it's fine to skip that handling.

shubham1g5 (dismissed):
Great job getting this to completion @Charl1996. Looks good to merge once the test passes on CI.

shubham1g5:

shubham1g5:

Charl1996:

Charl1996:
Rerunning this. Added a print statement to inspect the response of the curl command

Charl1996:
{u'error': u'Invalid username or password.'}

@shubham1g5 Did the credentials change recently?

Charl1996:

shubham1g5:

shubham1g5:

shubham1g5:

shubham1g5 (approved):

ctsims:
@Charl1996 🎉

PR #2569, opened 2022-07-25
Android dependencies

shubham1g5:
## Summary

[Spec](https://docs.google.com/document/d/1F_nKK8dSvpEKB3QOqq2A8Rb1AcMKZwpg9_TLK_IrhJ4/edit#)

## Product Description

Adds a check to see whether all Android dependencies are installed, if not prompts user to do in a blocking fashion - 

1. Only one unsatisified dependency - 

2. More than one unsatisified Dependencies - 

## Safety Assurance

### Automated test coverage

Added Instrumentation tests to test for the feature 

### Safety story

- Instrumentation tests only, not even planning to get this into manual QA plan as it looks like instrumention tests should suffice on this 

Easier to review the PR as a whole.

cross-request: https://github.com/dimagi/commcare-core/pull/1140


shubham1g5:

shubham1g5:

shubham1g5:

shubham1g5:

dimagimon:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

app/src/org/commcare/utils/PlaystoreUtils.kt
```diff
@@ -0,0 +1,24 @@
...
+import android.content.Intent
+import android.content.pm.PackageManager
+import android.net.Uri
+import org.commcare.CommCareApplication
+
+object PlaystoreUtils {
+
+    fun isApkInstalled(packageName: String): Boolean {
+        return try {
+            val packageManager = CommCareApplication.instance().packageManager
```
ctsims, line 13: What is the failure mode of this message if we don't have **permission** to check for this package?  It seems like that's likely to occur as a source of confusion.

Also, I was thinking that there might be a slightly different way to signal for the installed presence of an app for apps without our "first class" permissions check here. I wonder if we can perform a background callout to a content provider or broadcast mechanism to ask the other app if it is installed. It might be fairly easy to provide that mechanism with the Support Library. Not a "Phase 1" requirement by any means, but worth considering.

ctsims (dismissed):
Left some minor feedback about the text of these prompts.

question about the implementation: If there is more than one required app, is the workflow "clean" in that the user just has to keep pressing the one button that pops up until their app starts? I'm wondering if we should display a list of the required dependencies *only* if there's more than one, since that'd give the user some visual feedback after the first dependency is installed that something happened in between seeing the "Install from Store" dialog the second time.

shubham1g5:
> I'm wondering if we should display a list of the required dependencies only if there's more than one

Agree that we should give some clue to the user on this. I think we should be able to modify the dialog messaging to facilitate this. Although I am not sure if we should display all pending dpendencies or just the dependency we are opening playstore page for. Something like -

Option1: Commcare needs below applications to be i
Option2:  CommCare needs 'Reminders App' to be installed on your device before you can continue.

app/src/org/commcare/utils/PlaystoreUtils.kt
shubham1g5: > What is the failure mode of this message if we don't have permission to check for this package?

It's same as app not being installed on the device i.e. the catched `PackageManager.NameNotFoundException` below. 

>  I wonder if we can perform a background callout to a content provider or broadcast mechanism to ask the other app if it is installed

You mean signalling from External App to CommCare that it's installed ? That sounds feasible for Dimagi maintained apps although it won't work for other generic apps like WhatsApp for example.

ctsims: Huh, unfortunate that Android doesn't differentiate on that for us. It would be much, much better if we were able to flag that an app permission couldn't be checked.

> You mean signaling from External App to CommCare that it's installed ? That sounds feasible for Dimagi maintained apps although it won't work for other generic apps like WhatsApp for example.

Yeah. I'm imagining that part of the required package config model could be, say, a Content Provider endpoint or something similar with a known schema response that could signal that the app is on device.

It wouldn't work with non-CommCare specific apps, but it does give a third party the ability to use the mechanism with equal independence to Dimagi (or us the ability to use the mechanism for a new app we make without or before doing a new mobile release)

shubham1g5: Got it, that definitely makes sense. I have detailed out this approach [here](https://docs.google.com/document/d/1F_nKK8dSvpEKB3QOqq2A8Rb1AcMKZwpg9_TLK_IrhJ4/edit#heading=h.rup8i9jup2h8).

shubham1g5:
Sounds good! Thanks for the detailed thoughts on that.

shubham1g5:

shubham1g5:

shubham1g5:

app/src/org/commcare/activities/CommCareSetupActivity.java
```diff
@@ -290,6 +292,12 @@ protected void onResume() {
             setResult(RESULT_OK);
             finish();
         }
+
+        // if we were in dependency dialog state and resumed CC, take the next step
+        // that we otherwise would have taken on a successful app install
+        if(uiState.equals(UiState.DEPENDENCY_DIALOG)){
+            launchNextActivityOnAppInstall();
```
ctsims, line 299: Out of curiosity does this end up retaining the "multiple" state when the previous user action transitioned the number of '"pending" dependencies from 2 to 1? I don't see a way for it to do so. I'm not sure what the cleanest way to retain that state is, but it seems like it would regress to the other behavior in between.

ctsims (approved):
One question for followup, but other than that the code looks great.

Another thing I wanted to confirm: Where is this getting documented in the profile spec? I wanna make sure we have the spec and details in the open source repo as well.

app/src/org/commcare/activities/CommCareSetupActivity.java
shubham1g5: that's right. Today we only show the dialog once after app install. I did that since I didn't want to block user indefinitely at this stage as there are other possible app manager functions that user might want to access to. We can probably check for the no. of unsatisfied apps here again and if the number decreased show the dialog again ?

shubham1g5:
Merging for now, can followup on any other nice to haves separately.

PR #2588, opened 2022-09-22
Resolve Intent redirection vulnerability

avazirna:
## Summary
This PR is to resolve a vulnerability that could be exploited to access protected CommCare components. The solution is to check if the source of the intent is CommCare before sharing any data. Note, it doesn't seem necessary to notify the user about when the check fails.

JIRA ticket: [SAAS-13974](https://dimagi-dev.atlassian.net/browse/SAAS-13974)

## Safety Assurance

### DIFF
--- a/app/src/org/commcare/activities/CommCareSetupActivity.java
+++ b/app/src/org/commcare/activities/CommCareSetupActivity.java
@@ -163,9 +163,7 @@ public class CommCareSetupActivity extends CommCareActivity<CommCareSetupActivit
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-
         fromManager = getIntent().getBooleanExtra(AppManagerActivity.KEY_LAUNCH_FROM_MANAGER, false);
-
         if (checkForMultipleAppsViolation()) {
             return;
         }
@@ -760,6 +758,9 @@ public class CommCareSetupActivity extends CommCareActivity<CommCareSetupActivit
             Intent i = new Intent(getIntent());
             setResult(RESULT_OK, i);
         }
+        else
+            fail(Localization.get("install.invalid.launch"));
+
         finish();
     }
 
--- a/app/src/org/commcare/activities/CommCareSetupActivity.java
+++ b/app/src/org/commcare/activities/CommCareSetupActivity.java
@@ -756,7 +756,7 @@ public class CommCareSetupActivity extends CommCareActivity<CommCareSetupActivit
                 Intent i = new Intent(getApplicationContext(), DispatchActivity.class);
                 startActivity(i);
             }
-        } else {
+        } else if (getCallingActivity().getPackageName().startsWith(BuildConfig.APPLICATION_ID)){
             Intent i = new Intent(getIntent());
             setResult(RESULT_OK, i);
         }
--- a/app/src/org/commcare/activities/CommCareSetupActivity.java
+++ b/app/src/org/commcare/activities/CommCareSetupActivity.java
@@ -754,7 +754,7 @@ public class CommCareSetupActivity extends CommCareActivity<CommCareSetupActivit
                 Intent i = new Intent(getApplicationContext(), DispatchActivity.class);
                 startActivity(i);
             }
-        } else if (getCallingActivity().getPackageName().startsWith(BuildConfig.APPLICATION_ID)){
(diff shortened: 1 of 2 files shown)


dimagimon:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Message">
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">:link: <a href="https://dimagi-dev.atlassian.net/browse/SAAS-13974">SAAS-13974</a></td>
    </tr>
  </tbody>
</table>

</p>

app/src/org/commcare/activities/CommCareSetupActivity.java
```diff
@@ -756,7 +756,7 @@ private void launchNextActivityOnAppInstall() {
                 Intent i = new Intent(getApplicationContext(), DispatchActivity.class);
                 startActivity(i);
             }
-        } else {
+        } else if (getCallingActivity().getPackageName().startsWith(BuildConfig.APPLICATION_ID)){
             Intent i = new Intent(getIntent());
             setResult(RESULT_OK, i);
         }
```
shubham1g5, line 762: we should add another else statement and fail there with an error message like - 'Invalid CommCare launch, External applications should use action `android.intent.action.VIEW` to launch CommCare'. You can utilise the existing `fail` method in this class to do that.

shubham1g5 (changes requested):

app/src/org/commcare/activities/CommCareSetupActivity.java
avazirna: Noted. I battled a little with that, I was divided between notifying the user and letting it die silently.

shubham1g5 (approved):

app/src/org/commcare/activities/CommCareSetupActivity.java
```diff
@@ -756,10 +754,13 @@ private void launchNextActivityOnAppInstall() {
                 Intent i = new Intent(getApplicationContext(), DispatchActivity.class);
                 startActivity(i);
             }
-        } else {
+        } else if (getCallingActivity().getPackageName().startsWith(BuildConfig.APPLICATION_ID)){
```
shubham1g5, line 757: this is crashing `AppDependenciesTest` with - 

````
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String android.content.ComponentName.getPackageName()' on a null object reference
	at org.commcare.activities.CommCareSetupActivity.launchNextActivityOnAppInstall(SourceFile:757)
	at org.commcare.activities.CommCareSetupActivity.onResume(SourceFile:297)
	at android.app.Instrumentation.callActivityOnResume(Instrumentation.java:1361)
````

We will need to do a null check on `getCallingActivity()`

**Review by shubham1g5 — CHANGES_REQUESTED** [review]

File: app/src/org/commcare/activities/CommCareSetupActivity.java (hunk 2 above)
Comment (shubham1g5) [reply app/src/org/commcare/activities/CommCareSetupActivity.java:757]: Also is there a reason to not exact match the package strings ?

Comment (avazirna) [reply app/src/org/commcare/activities/CommCareSetupActivity.java]: Not a very strong reason, just wanted to be able to respond to calls from other apps in the same domain. Should we make this stricter?

Comment (shubham1g5) [reply app/src/org/commcare/activities/CommCareSetupActivity.java]: Yeah I think we should do a strict match as we should only come here using internal launch from CommCare itself. All other calls would be external and should use VIEW action on intent.

File: app/src/org/commcare/activities/CommCareSetupActivity.java [hunk 3]
```diff
@@ -754,7 +754,7 @@ private void launchNextActivityOnAppInstall() {
                 Intent i = new Intent(getApplicationContext(), DispatchActivity.class);
                 startActivity(i);
             }
-        } else if (getCallingActivity().getPackageName().startsWith(BuildConfig.APPLICATION_ID)){
+        } else if (getCallingActivity().getPackageName()!= null && getCallingActivity().getPackageName().startsWith(BuildConfig.APPLICATION_ID)){
```
shubham1g5, line 757: null check should be on `getCallingActivity` instead.

avazirna: 🤦my bad, I misread the exception message.

shubham1g5 (approved):

shubham1g5:

PR #2602, opened 2022-11-03
Robolectric 4.8.2

shubham1g5:
## Summary

Updates robolectric to 4.8.2 (this also resolves some issues for M1 users in running tests)

## Safety Assurance

commit by commit

### DIFF
--- a/app/unit-tests/src/org/commcare/CommCareTestApplication.java
+++ b/app/unit-tests/src/org/commcare/CommCareTestApplication.java
@@ -17,6 +17,7 @@ import org.commcare.core.network.ModernHttpRequester;
 import org.commcare.dalvik.BuildConfig;
 import org.commcare.heartbeat.HeartbeatRequester;
 import org.commcare.heartbeat.TestHeartbeatRequester;
+import org.commcare.logging.DataChangeLogger;
 import org.commcare.models.AndroidPrototypeFactory;
 import org.commcare.models.database.AndroidPrototypeFactorySetup;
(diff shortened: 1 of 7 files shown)


app/unit-tests/src/org/commcare/CommCareTestApplication.java
```diff
@@ -66,6 +67,9 @@ public class CommCareTestApplication extends CommCareApplication implements Test
 
     @Override
     public void onCreate() {
+        // set if before calling super to initialte the dataChangeLogger correctly
```
avazirna, line 70: @shubham1g5 are these changes motivated/required by the Robolectric update?

shubham1g5: Yeah after the update the data change log test started failing [over here](https://github.com/dimagi/commcare-android/blob/master/app/unit-tests/src/org/commcare/logging/DataChangeLoggerTest.java#L28) as the application onCreate() was no longer getting called before [setup function](https://github.com/dimagi/commcare-android/blob/master/app/unit-tests/src/org/commcare/logging/DataChangeLoggerTest.java#L23) where we were earlier initialising the logger. I didn't go deep into what aspect of robolectric update caused this, but instead tried to fix the [original data change logger init](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/CommCareApplication.java#L208) as part of application onCreate() by setting the storage here.

avazirna (approved):
Sounds good to me.

PR #2608, opened 2022-11-14
ensure singular value for last query string

shubham1g5:
## Summary

1. Since we have changed StackFrameStep.getExtra to thrown [an error](https://github.com/dimagi/commcare-core/blob/formplayer/src/main/java/org/commcare/suite/model/StackFrameStep.java#L167) when multiple extras with same key are present, this ensures that we only save singular value for `LAST_QUERY_STRING` in the stack frame step extras
2.  Adds Proguard config for Guava from [here](https://github.com/google/guava/wiki/UsingProGuardWithGuava)

## Safety Assurance

### Automated test coverage

Changes are motivated from Instrumentation test failures. 

cross-request: https://github.com/dimagi/commcare-core/pull/1163

### DIFF
--- a/app/proguard.cfg
+++ b/app/proguard.cfg
@@ -122,3 +122,101 @@
 -keep class kotlin.collections.ArraysKt { *; }
 -keep class kotlin.collections.CollectionsKt { *; }
 -keep class kotlin.text.StringsKt { *; }
+
+# Guava Config Start
+-dontwarn javax.lang.model.element.Modifier
+
+# Note: We intentionally don't add the flags we'd need to make Enums work.
+# That's because the Proguard configuration required to make it work on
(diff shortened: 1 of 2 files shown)


app/proguard.cfg
```diff
@@ -122,3 +122,107 @@
 -keep class kotlin.collections.ArraysKt { *; }
 -keep class kotlin.collections.CollectionsKt { *; }
 -keep class kotlin.text.StringsKt { *; }
+
+# Guava Config Start
+-dontwarn javax.lang.model.element.Modifier
```
snopoke, line 127: was it intentional to include these proguard changes?

snopoke (approved):

PR #2634, opened 2023-02-06
Adds UI Controller for Query Request Activity

shubham1g5:
## Summary

Refactors Query Request Activity to move UI logic out into a UI controller

## Safety Assurance

### Automated test coverage

[Unit tests](https://github.com/dimagi/commcare-android/blob/master/app/unit-tests/src/org/commcare/android/tests/activities/QueryRequestActivityTest.java) and [Instrumentation tests](https://github.com/dimagi/commcare-android/blob/master/app/instrumentation-tests/src/org/commcare/androidTests/CaseListSearchTest.kt)

### Safety story

Changes are no-op in nature and are well covered by tests

I would not suggest reviewer to look at all code in detail here as it's only moving around of QueryRequestActivity code to another Kotlin class for UI controller. A good reviewing strategy here is to look at tests linked above and flag it if you feel unsure about test coverage.

### DIFF
--- a/app/src/org/commcare/activities/QueryRequestActivity.java
+++ b/app/src/org/commcare/activities/QueryRequestActivity.java
@@ -215,13 +215,16 @@ public class QueryRequestActivity
 
     @Override
     public void initUIController() {
+        initRemoteQuerySessionManager();
+        mRequestUiController = new QueryRequestUiController(this, remoteQuerySessionManager);
+    }
+
+    private void initRemoteQuerySessionManager() {
         try {
             AndroidSessionWrapper sessionWrapper = CommCareApplication.instance().getCurrentSessionWrapper();
-
             try {
-                remoteQuerySessionManager =
-                        RemoteQuerySessionManager.buildQuerySessionManager(sessionWrapper.getSession(),
-                                sessionWrapper.getEvaluationContext(), getSupportedPrompts());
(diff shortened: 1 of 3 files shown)


avazirna (approved):
This looks good to me. As per the review notes:

- Instrumentation tests - I think you meant to link to the [CaseClaimTest](https://github.com/dimagi/commcare-android/blob/master/app/instrumentation-tests/src/org/commcare/androidTests/CaseClaimTest.java) and not the [CaseListSearchTest](https://github.com/dimagi/commcare-android/blob/master/app/instrumentation-tests/src/org/commcare/androidTests/CaseListSearchTest.kt), is that correct?
- Employing `isFinishing` to prevent initialization tasks from running when the activity has been cancelled is a logical move to me, however, I might not have a full picture regarding other classes that derive from `CommCareActivity` so curious to hear from someone with more context.

shubham1g5:
> I think you meant to link to the [CaseClaimTest](https://github.com/dimagi/commcare-android/blob/master/app/instrumentation-tests/src/org/commcare/androidTests/CaseClaimTest.java) and not the [CaseListSearchTest](https://github.com/dimagi/commcare-android/blob/master/app/instrumentation-tests/src/org/commcare/androidTests/CaseListSearchTest.kt), is that correct?

Ahh yes correcting in description. Thanks!

PR #2638, opened 2023-02-13
Add button to SSL errors to take user to date settings

OrangeAndGreen:
## Summary
https://dimagi-dev.atlassian.net/browse/SAAS-10301

## Feature Flag
None

## Product Description
When encountering an SSL error in the app, the detailed error message which indicates that an incorrect date/time may be responsible for the issue now includes a button the user can press to take them directly to the date/time settings.

## Safety Assurance

### Automated test coverage
None

### Safety story
The change involves providing more consistent messaging for SSL errors and the ability to launch a settings page. This functionality is only shown in the rare event of an SSL error, and otherwise has no effect on app functionality.

cross-request: https://github.com/dimagi/commcare-core/pull/1220

### DIFF
--- a/app/src/org/commcare/activities/FormAndDataSyncer.java
+++ b/app/src/org/commcare/activities/FormAndDataSyncer.java
@@ -79,45 +79,10 @@ public class FormAndDataSyncer {
                             ((WithUIController)receiver).getUIController().refreshView();
                         }
 
-                        switch (result) {
-                            case FULL_SUCCESS:
-                                receiver.handleFormSendResult(getLabelForFormsSent(), true);
+                        receiver.handleFormUploadResult(result, getLabelForFormsSent(), userTriggered);
 
-                                if (syncAfterwards) {
-                                    syncDataForLoggedInUser(receiver, true, userTriggered);
-                                }
-                                break;
-                            case AUTH_FAILURE:
-                                receiver.handleFormSendResult(Localization.get("sync.fail.auth.loggedin"), false);
-                                break;
-                            case TRANSPORT_FAILURE:
-                                receiver.handleFormSendResult(Localization.get("sync.fail.bad.network"), false);
-                                break;
-                            case BAD_CERTIFICATE:
-                                CommCareApplication.notificationManager().reportNotificationMessage(
-                                        NotificationMessageFactory.message(NotificationMessageFactory.StockMessages.BadSSLCertificate,
-                                                NotificationActionButtonInfo.ButtonAction.LAUNCH_DATE_SETTINGS));
-                                receiver.handleFormSendResult(Localization.get("sync.fail.badcert"), false);
-                                break;
-                            case PROCESSING_FAILURE:
-                                receiver.handleFormSendResult(Localization.get("sync.fail.server.error"), false);
-                                break;
-                            case RECORD_FAILURE:
-                                receiver.handleFormSendResult(Localization.get("sync.fail.individual"), false);
-                                break;
-                            case ACTIONABLE_FAILURE:
-                                receiver.handleFormSendResult(result.getErrorMessage(), false);
-                                break;
-                            case RATE_LIMITED:
-                                receiver.showRateLimitError(userTriggered);
-                                break;
-                            case INVALID_CASE_GRAPH:
-                                receiver.handleFormSendResult(Localization.get("sync.fail.invalid.case.graph"), false);
-                                break;
-                            case FAILURE:
-                            default:
-                                receiver.handleFormSendResult(Localization.get("sync.fail.unknown"), false);
-                                break;
+                        if(result == FormUploadResult.FULL_SUCCESS && syncAfterwards) {
+                            syncDataForLoggedInUser(receiver, true, userTriggered);
                         }
                     }
 
@@ -127,14 +92,13 @@ public class FormAndDataSyncer {
 
                     @Override
                     protected void deliverError(SyncCapableCommCareActivity receiver, Exception e) {
-                        receiver.handleFormSendResult(Localization.get("sync.fail.unsent"), false);
+                        receiver.handleFormUploadResult(FormUploadResult.UNSENT, getLabelForFormsSent(), userTriggered);
                     }
 
                     @Override
                     protected void handleCancellation(SyncCapableCommCareActivity receiver) {
                         super.handleCancellation(receiver);
-                        receiver.handleFormSendResult(Localization.get("activity.task.cancelled.message")
-                                + " " + getLabelForFormsSent(), false);
+                        receiver.handleFormUploadResult(FormUploadResult.CANCELLED, getLabelForFormsSent(), userTriggered);
                     }
                 };
 
--- a/app/src/org/commcare/activities/FormAndDataSyncer.java
+++ b/app/src/org/commcare/activities/FormAndDataSyncer.java
@@ -23,6 +23,8 @@ import org.commcare.tasks.PullTaskResultReceiver;
 import org.commcare.tasks.ResultAndError;
 import org.commcare.utils.FormUploadResult;
 import org.commcare.utils.StorageUtils;
+import org.commcare.views.notifications.NotificationActionButtonInfo;
+import org.commcare.views.notifications.NotificationMessageFactory;
 import org.javarosa.core.model.User;
 import org.javarosa.core.reference.InvalidReferenceException;
 import org.javarosa.core.reference.ReferenceManager;
@@ -90,6 +92,12 @@ public class FormAndDataSyncer {
                                 break;
                             case TRANSPORT_FAILURE:
                                 receiver.handleFormSendResult(Localization.get("sync.fail.bad.network"), false);
+                                break;
+                            case BAD_CERTIFICATE:
+                                CommCareApplication.notificationManager().reportNotificationMessage(
+                                        NotificationMessageFactory.message(NotificationMessageFactory.StockMessages.BadSSLCertificate,
+                                                NotificationActionButtonInfo.ButtonAction.LAUNCH_DATE_SETTINGS));
+                                receiver.handleFormSendResult(Localization.get("sync.fail.badcert"), false);
                                 break;
                             case PROCESSING_FAILURE:
                                 receiver.handleFormSendResult(Localization.get("sync.fail.server.error"), false);
--- a/app/src/org/commcare/activities/MessageActivity.java
+++ b/app/src/org/commcare/activities/MessageActivity.java
@@ -73,15 +73,15 @@ public class MessageActivity extends CommcareListActivity {
                     action.setText(actionText);
                 }
 
-                if(msg.getButtonInfo() == null) {
-                    actionButton.setVisibility(View.GONE);
-                } else {
+                if(msg.getButtonInfo() != null) {
+                    actionButton.setVisibility(View.VISIBLE);
                     actionButton.setText(msg.getButtonInfo().getButtonText());
 
                     actionButton.setOnClickListener(v -> {
                         switch(msg.getButtonInfo().getButtonAction()) {
                             case LAUNCH_DATE_SETTINGS -> SettingsHelper.launchDateSettings(getContext());
                             //Future actions will be added here once implemented
+                            default -> throw new RuntimeException("Unhandled button action: " + msg.getButtonInfo().getButtonAction());
                         }
                     });
                 }
--- a/app/src/org/commcare/activities/MessageActivity.java
+++ b/app/src/org/commcare/activities/MessageActivity.java
@@ -7,10 +7,12 @@ import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.ArrayAdapter;
+import android.widget.Button;
 import android.widget.TextView;
 
 import org.commcare.CommCareApplication;
 import org.commcare.dalvik.R;
+import org.commcare.views.notifications.NotificationActionButtonInfo;
 import org.commcare.views.notifications.NotificationMessage;
 
 import java.text.DateFormat;
@@ -58,6 +60,8 @@ public class MessageActivity extends CommcareListActivity {
                 TextView body = messageView.findViewById(R.id.layout_note_msg_body);
                 TextView date = messageView.findViewById(R.id.layout_note_msg_date);
                 TextView action = messageView.findViewById(R.id.layout_note_msg_action);
+                Button actionButton = messageView.findViewById(R.id.layout_note_msg_action_button);
+
                 title.setText(msg.getTitle());
                 body.setText(msg.getDetails());
                 date.setText(DateUtils.formatSameDayTime(msg.getDate().getTime(), System.currentTimeMillis(), DateFormat.DEFAULT, DateFormat.DEFAULT));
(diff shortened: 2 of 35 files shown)


app/src/org/commcare/network/HttpCalloutTask.java
```diff
@@ -38,7 +38,7 @@ public enum HttpCalloutOutcomes {
         BadResponse,
         AuthFailed,
         UnknownError,
-        BadCertificate,
+        BadSSLCertificate,
```
shubham1g5, line 41: nit: call it `BadSslCertificate`

app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -684,7 +689,8 @@ public enum PullTaskResult {
         RATE_LIMITED_SERVER_ERROR(AnalyticsParamValue.SYNC_FAIL_RATE_LIMITED_SERVER_ERROR),
         STORAGE_FULL(AnalyticsParamValue.SYNC_FAIL_STORAGE_FULL),
         CAPTIVE_PORTAL(AnalyticsParamValue.SYNC_FAIL_CAPTIVE_PORTAL),
-        AUTH_OVER_HTTP(AnalyticsParamValue.SYNC_FAIL_AUTH_OVER_HTTP);
+        AUTH_OVER_HTTP(AnalyticsParamValue.SYNC_FAIL_AUTH_OVER_HTTP),
+        BAD_CERTIFICATE(AnalyticsParamValue.SYNC_FAIL_BAD_CERTIFICATE);
```
shubham1g5, line 693: nit: rename it to BadSslCertificate as well to maintain consistency with other parts of code.

app/src/org/commcare/utils/FormUploadResult.java
```diff
@@ -69,7 +69,13 @@ public enum FormUploadResult {
...
      * User is behind a captive portal, no need to try re-submissions
      */
-    CAPTIVE_PORTAL(12)
+    CAPTIVE_PORTAL(12),
+
+    /**
+     * Error involving SSL certificate, no need to try re-submission
+     * This is often related to local clock settings
+     */
+    BAD_CERTIFICATE(13)
```
shubham1g5, line 78: nit: call it `BadSSLCertificate`

app/src/org/commcare/views/notifications/NotificationMessage.java
```diff
@@ -24,18 +25,26 @@ public int describeContents() {
 
     @Override
     public void writeToParcel(Parcel dest, int flags) {
-        dest.writeStringArray(new String[]{category, title, details, actions});
+        String buttonText = buttonInfo != null ? buttonInfo.getButtonText() : "";
```
shubham1g5, line 28: I think a more scalable design here is to make `NotificationActionButtonInfo` implements  `Parcelable` and localise the code to write and load `NotificationActionButtonInfo` inside itself and then use `dest.writeParcelable();` and `source.readParcelable()` respectively in this class to write the button info.

app/src/org/commcare/views/notifications/NotificationMessage.java
```diff
@@ -44,15 +53,23 @@ public NotificationMessage[] newArray(int size) {
...
+        this(context, title, details, action, date, null);
+    }
+
+    public NotificationMessage(String context, String title, String details, String action, Date date, NotificationActionButtonInfo buttonInfo) {
         if (context == null || title == null || details == null || date == null) {
-            throw new NullPointerException("None of the arguments for creating a NotificationMessage may be null except for action");
+            throw new NullPointerException("None of the arguments for creating a NotificationMessage may be null except for action and buttonInfo");
+        }
+        if (buttonInfo != null && buttonInfo.getButtonText() == null) {
+            throw new NullPointerException("NotificationMessage with buttonInfo can not have null button text");
```
shubham1g5, line 65: move it to the buttonInfo constructor itself ?

app/src/org/commcare/views/notifications/NotificationMessage.java
```diff
@@ -83,6 +104,20 @@ public boolean equals(Object o) {
...
+        }
+        if(nm.buttonInfo != null) {
+            if(!nm.buttonInfo.getButtonText().equals(this.buttonInfo.getButtonText())) {
+                return false;
+            }
+
+            if(!nm.buttonInfo.getButtonAction().equals(this.buttonInfo.getButtonAction())) {
+                return false;
+            }
+        }
```
shubham1g5, line 119: move this inside a `equals` method in button info class and here simply compare the 2 button info objects.

app/src/org/commcare/views/notifications/NotificationMessageFactory.java
```diff
@@ -211,7 +227,18 @@ public static NotificationMessage message(MessageTag message, String[] parameter
                 //No big deal, key doesn't need to exist
             }
 
-            return new NotificationMessage(customCategory, title, detail, action, new Date());
+            NotificationActionButtonInfo buttonInfo = null;
+            try {
+                String button = parameters[3] == null ? Localization.get(base + ".button") : Localization.get(base + ".button", new String[]{parameters[3]});
+                buttonInfo = new NotificationActionButtonInfo(button, buttonAction);
+            } catch (Exception e) {
+                if(buttonAction != NotificationActionButtonInfo.ButtonAction.NONE) {
```
shubham1g5, line 235: wrap the `String button` evaluation above under this `if` instead of adding in the exception catcher and then remove the `catch` block completely from here ?

app/src/org/commcare/views/notifications/NotificationMessageFactory.java
```diff
@@ -211,7 +227,18 @@ public static NotificationMessage message(MessageTag message, String[] parameter
                 //No big deal, key doesn't need to exist
             }
 
-            return new NotificationMessage(customCategory, title, detail, action, new Date());
+            NotificationActionButtonInfo buttonInfo = null;
+            try {
+                String button = parameters[3] == null ? Localization.get(base + ".button") : Localization.get(base + ".button", new String[]{parameters[3]});
```
shubham1g5, line 232: I don't think we need to support a parameterised localisation for button text. We can simply keep `parameters` as an array of [3] as before and eval button as just `Localization.get(base + ".button")`

shubham1g5 (commented):
@OrangeAndGreen looks solid 🥳 I left a few stylish comments on this. Except that will you be able to add a screenshot of the UI change in Product description for this PR to have a review on UI as well. Thanks!

app/src/org/commcare/activities/FormAndDataSyncer.java
```diff
@@ -91,6 +93,12 @@ protected void deliverResult(SyncCapableCommCareActivity receiver, FormUploadRes
                             case TRANSPORT_FAILURE:
                                 receiver.handleFormSendResult(Localization.get("sync.fail.bad.network"), false);
                                 break;
+                            case BAD_CERTIFICATE:
+                                CommCareApplication.notificationManager().reportNotificationMessage(
```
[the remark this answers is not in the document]
OrangeAndGreen: That makes sense. The issue I was having was that the call back to the receiver already reduced the FormUploadResult to a string so the receiver couldn't determine when to show the notification. So to allow that I created a new handleFormUploadResult method on the receiver that takes the FormUploadResult instead of a string message.

At this point it seemed to make sense to make FormUploadResult a MessageTag and use that to map the enum values to the translatable strings. So I did that and used the mappings from the switch structure to populate the enum mappings.

But then it seemed to make sense to move the entire switch from FormAndDataSyncer to the new method, since most of the cases were simply translating enum to message anyway. That leaves only a little success-based logic in FormAndDataSyncer and otherwise puts the logic all in SyncCapableCommCareActivity.

Finally, I added two more cases to FormUploadRestul for UNSENT and CANCELLED. That allowed me to replace the other two calls in FormAndDataSyncer to handleFormSendResult, such that only handleFormUploadResult is called now.

app/src/org/commcare/activities/MessageActivity.java
```diff
@@ -69,6 +73,19 @@ public View getView(int position, View convertView, @NonNull ViewGroup parent) {
                     action.setText(actionText);
                 }
 
+                if(msg.getButtonInfo() == null) {
+                    actionButton.setVisibility(View.GONE);
```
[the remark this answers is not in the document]
OrangeAndGreen: Agreed

app/src/org/commcare/network/HttpCalloutTask.java
OrangeAndGreen: Sounds good. There are three enums that had the BadSSLCertificate value (HttpCalloutTask, StockMessages, and AppInstallStatus), I changed all three to use BadSslCertificate instead.

app/src/org/commcare/views/notifications/NotificationMessage.java
OrangeAndGreen: Agreed

app/src/org/commcare/views/notifications/NotificationMessage.java
OrangeAndGreen: Agreed

app/src/org/commcare/activities/UpdateActivity.java
```diff
@@ -314,7 +316,16 @@ public void handleTaskCompletion(ResultAndError<AppInstallStatus> result) {
...
             }
         } else {
-            reportFailureToNotifications(result.errorMessage);
+            switch(result.data) {
+                case BadSSLCertificate:
+                    reportFailureToNotifications(result.data, NotificationActionButtonInfo.ButtonAction.LAUNCH_DATE_SETTINGS);
+                    break;
+                default:
+                    //QUESTION: AppInstallStatus maps to MessageTags, shouldn't we be using that? i.e.
+                    //reportFailureToNotifications(result.data, NotificationActionButtonInfo.ButtonAction.NONE);
```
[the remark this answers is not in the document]
OrangeAndGreen: Ok, I rearranged the code a bit so that the checks for which notification to show all happen in the reportFailureToNotification method. I also added a fallback case to create the NotificationMessage simply from the AppInstallStatus (no extra message).

The way the code is now, the UpdateFailedGeneral message with additional errorMessage is still shown if the errorMessage is not empty. The errorMessage usually comes from calling getMessage() on the caught exception.

app/src/org/commcare/activities/SyncCapableCommCareActivity.java
```diff
@@ -237,9 +272,8 @@ public void handleFormSendResult(String message, boolean success) {
     }
 
     public void showRateLimitError(boolean userTriggered) {
-
         if (HiddenPreferences.isRateLimitPopupDisabled() || !userTriggered) {
-            handleFormSendResult(Localization.get("form.send.rate.limit.error.toast"), false);
+            handleFormUploadResult(FormUploadResult.RATE_LIMITED, null, userTriggered);
```
shubham1g5, line 276: good thinking on making this a `FormUploadResult`

app/src/org/commcare/activities/UpdateActivity.java
```diff
@@ -336,34 +328,40 @@ public void handleTaskCompletion(ResultAndError<AppInstallStatus> result) {
...
+                                    NotificationActionButtonInfo.ButtonAction.NONE);
+                } else if (!"".equals(errorMessage)) {
+                    notificationMessage =
+                            NotificationMessageFactory.message(AppInstallStatus.UpdateFailedGeneral,
+                                    new String[]{null, errorMessage, null},
+                                    NotificationActionButtonInfo.ButtonAction.NONE);
+                } else {
+                    notificationMessage =
+                            NotificationMessageFactory.message(status,
+                                    NotificationActionButtonInfo.ButtonAction.NONE);
```
shubham1g5, line 356: Is this a change in behaviour from earlier i.e. we are showing a notification even when errorMessage is empty ?

app/src/org/commcare/activities/SyncCapableCommCareActivity.java
```diff
@@ -109,6 +112,7 @@ public void handlePullTaskResult(ResultAndError<DataPullTask.PullTaskResult> res
                 updateUiAfterDataPullOrSend(Localization.get("sync.fail.bad.data"), FAIL);
                 break;
             case DOWNLOAD_SUCCESS:
+                CommCareApplication.notificationManager().clearNotifications(NotificationMessageFactory.StockMessages.BadSslCertificate.getCategory());
```
shubham1g5, line 115: Thinking around if we add more error handling paths that result into error notifications from sync, it would be helpful if we can delete all notifications related to sync errros here instead of just one corresponding to `BadSslCertificate`.

app/src/org/commcare/views/notifications/NotificationActionButtonInfo.java
```diff
@@ -0,0 +1,73 @@
...
+    @Override
+    public int describeContents() {
+        return 0;
+    }
+
+    @Override
+    public void writeToParcel(Parcel dest, int flags) {
+        String buttonText = getButtonText();
+        String buttonAction = getButtonAction().name();
+        dest.writeStringArray(new String[]{buttonText, buttonAction});
```
shubham1g5, line 26: I think a cleaner way to do this is without arrays - 
````
dest.writeString(text)
dest.writeString(action.name) 
````
and then reading it back from source the same way - 
````
new NotificationActionButtonInfo(source.readString(), source.readString());
````

**Review by shubham1g5 — COMMENTED** [review]
nice improvements overall :)

File: app/src/org/commcare/activities/UpdateActivity.java (hunk 13 above)
Comment (OrangeAndGreen) [reply app/src/org/commcare/activities/UpdateActivity.java:356]: Yes, the additional else case changes the behavior to show the translatable message that maps to each AppInstallStatus when the error message is otherwise empty. Removing this case would restore the prior behavior, i.e. not raising a notification unless there's an error message. I can put it back that way if you prefer.

File: app/src/org/commcare/views/notifications/NotificationActionButtonInfo.java (hunk 15 above)
Comment (OrangeAndGreen) [reply app/src/org/commcare/views/notifications/NotificationActionButtonInfo.java:26]: Agreed

File: app/src/org/commcare/activities/UpdateActivity.java (hunk 13 above)
Comment (shubham1g5) [reply app/src/org/commcare/activities/UpdateActivity.java:356]: Yeah my guess here is we only want to show notifications when there is something more around errors than what's presented on screen to the user. So showing notifications in case of no error messages might get redundant in usage or result in too many error notifications. So probably best to preserve the original behaviour unless we do some testing around the change to see if it causes redunandant/many notifications.

File: app/src/org/commcare/activities/UpdateActivity.java [hunk 16]
```diff
@@ -325,26 +328,36 @@ public void handleTaskCompletion(ResultAndError<AppInstallStatus> result) {
...
+                }
 
-        if (notificationMessage != null) {
-            CommCareApplication.notificationManager()
-                    .reportNotificationMessage(notificationMessage, true);
-            uiController.setNotificationsVisible();
+                break;
         }
+
+        CommCareApplication.notificationManager()
```
shubham1g5, line 358: we should add back `if (notificationMessage != null) {` check here since notification can be null now

OrangeAndGreen: Agreed!

shubham1g5 (approved):
great work on this 💯

PR #2654, opened 2023-03-09
disbaling death on strict mode violations

shubham1g5:
## Summary

Strict mode crashes is making it impossible today to debug application without crashing. I am disabling it for now and we should turn it back on once we are able to solve for existing violations. 

## Safety Assurance

### Automated test coverage

NA

### Safety story
 
Only affect debug apk.


avazirna (approved):
@shubham1g5 what do you think about creating an Issue about this, so we can make sure it gets picked in the future.

shubham1g5:

PR #2658, opened 2023-03-27
Fix starting Foreground services from the background

avazirna:
## Summary
This PR resolves a restriction on Android 12+ with [starting foreground services while the application is in the background](https://developer.android.com/guide/components/foreground-services#background-start-restrictions), which is causing 
[ForegroundServiceStartNotAllowedException](https://developer.android.com/reference/android/app/ForegroundServiceStartNotAllowedException) errors.

## Safety Assurance

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -349,8 +349,9 @@
         <activity android:name="org.commcare.activities.MessageActivity">
         </activity>
 
-        <receiver android:name="org.commcare.views.notifications.NotificationClearReceiver">
-        </receiver>
+        <receiver android:name="org.commcare.views.notifications.NotificationClearReceiver"/>
+
+        <receiver android:name="org.commcare.services.CommCareSessionInitiatorReceiver"/>
 
         <activity
             android:label="@string/title_activity_report_problem"
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -28,7 +28,7 @@
     <uses-permission android:name="android.permission.RECORD_AUDIO" />
     <uses-permission android:name="android.permission.CAMERA" />
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
-
+    <uses-permission android:name="android.permission.SCHEDULE_EXACT_ALARM" />
 
     <uses-feature
         android:name="android.hardware.telephony"
--- a/app/src/org/commcare/services/CommCareSessionService.java
+++ b/app/src/org/commcare/services/CommCareSessionService.java
@@ -69,8 +69,8 @@ public class CommCareSessionService extends Service {
     /**
      * Names of the extra data for the CommCareServiceInitiatorReceiver intent
      */
-    private static final String EXTRA_NOTIFICATION_ID = "commcare.intent.extra.notification-id";
-    private static final String EXTRA_NOTIFICATION_OBJ = "commcare.intent.extra.notification";
+    public static final String EXTRA_NOTIFICATION_ID = "commcare.intent.extra.notification-id";
+    public static final String EXTRA_NOTIFICATION_OBJ = "commcare.intent.extra.notification";
 
     private NotificationManager mNM;
 
--- a/app/src/org/commcare/services/CommCareSessionService.java
+++ b/app/src/org/commcare/services/CommCareSessionService.java
@@ -71,6 +71,7 @@ public class CommCareSessionService extends Service {
      */
     public static final String EXTRA_NOTIFICATION_ID = "commcare.intent.extra.notification-id";
     public static final String EXTRA_NOTIFICATION_OBJ = "commcare.intent.extra.notification";
+    public static final String EXTRA_COMPONENT_TYPE = "commcare.intent.extra.component-type";
 
     private NotificationManager mNM;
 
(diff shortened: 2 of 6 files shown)


app/AndroidManifest.xml
```diff
@@ -28,7 +28,7 @@
     <uses-permission android:name="android.permission.RECORD_AUDIO" />
     <uses-permission android:name="android.permission.CAMERA" />
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
-
+    <uses-permission android:name="android.permission.SCHEDULE_EXACT_ALARM" />
```
shubham1g5, line 31: This permission can be revoked by the user or system at any time and we need to check if the permission is granted before using it otherwise direct user to grant this permission. [reference](https://developer.android.com/training/scheduling/alarms#using-schedule-exact-permission)

app/src/org/commcare/services/CommCareSessionService.java
```diff
@@ -236,8 +246,38 @@ public void showLoggedInNotification(@Nullable User user) {
...
-        //Send the notification.
-        this.startForeground(NOTIFICATION, notificationBuilder.build());
+        //Send the notification
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S && CommCareApplication.isAppInBackground())
+            startForegroundNotificationWithAlarmManager(NOTIFICATION, notificationBuilder.build());
+        else
+            this.startForeground(NOTIFICATION, notificationBuilder.build());
+    }
+
+    private boolean isAppInTheBackground() {
```
shubham1g5, line 256: ununsed.

shubham1g5 (changes requested):
The code itself looks great on it. Though I have a couple points worth mentioning here - 

1. This looks like a fine soultion for the purpose of this hotfix but I think we want to have a more long term strategy here for the next release (2.54). We possibly want to move away from the model of always having a foreground service running with the notification with these new Android restrictions in place. @ctsims I believe this will be a substantial product change for users as they might be accustomed to see the CC logged-in notification. Do you have thoughts on making this change. The trade off here is asking for a new permission to all users (exact alarm) vs retaining the foreground session notification. 

2. Have you been able to replicate this bug and verify that this PR actually fixes it ? Since it looks a bit tricky to repro this bug, it would be nice to have the repro steps mentioned in the description of this PR.

ctsims:
Hey @shubham1g5 

> This looks like a fine soultion for the purpose of this hotfix but I think we want to have a more long term strategy here for the next release (2.54). We possibly want to move away from the model of always having a foreground service running with the notification with these new Android restrictions in place. @ctsims I believe this will be a substantial product change for users as they might be accustomed to see the CC logged-in notification. Do you have thoughts on making this change. The trade off here is asking for a new permission to all users (exact alarm) vs retaining the foreground session notification.

Can you start a spec doc outlining this transition, including the motivations and underlying changes to Android's permissions framework that are making it necessary?

It's definitely a very big transition to move to a model where we don't have a persistent service, because the persistent service is the mechanism we use to maintain a persistent in-memory-only footprint for the decrypted database keys without Android killing the process. Without the service, many things could happen which would require the user to re-enter their password constantly, which is something we'd like to actively avoid since most of our users are provided CommCare as the main or sole function of the phone.

It's plausible there are new mechanisms we can take advantage of for that purpose, though. We already have it on the docket to explore other ways to manage other sensitive tokens outside of process memory, so we should possibly explore these concepts together.

shubham1g5:
>  because the persistent service is the mechanism we use to maintain a persistent in-memory-only footprint for the decrypted database keys without Android killing the process. 

Ahh I didn't realise that's the primary reason for the session service. I will see if I can find alternatives here and list them out in another doc.

app/src/org/commcare/utils/Permissions.java
```diff
@@ -88,38 +107,147 @@ public static boolean shouldShowPermissionRationale(AppCompatActivity activity,
...
+    private static String getSpecialPermissionDialogTitle(String specialPerm) {
+        switch(specialPerm){
+            case Manifest.permission.SCHEDULE_EXACT_ALARM:
+                return "permission.schedule.exact.alarm.title";
+            default:
+                throw new RuntimeException("Invalid special permission " + specialPerm);
+        }
+    }
+
+    // TODO: In addition to scrolling to where CommCare app is, we should highlight the item
```
shubham1g5, line 212: Does the page this intent send us to not CC specific and user need to locate CC from a list of apps ?

app/src/org/commcare/utils/Permissions.java
```diff
@@ -88,38 +107,147 @@ public static boolean shouldShowPermissionRationale(AppCompatActivity activity,
...
+    private static void startAlarmsAndRemindersActivity(AppCompatActivity activity) {
+        requestingSpecialAppPermission = true;
+        Intent intent = new Intent(Settings.ACTION_REQUEST_SCHEDULE_EXACT_ALARM);
+
+        intent.setData(Uri.parse("package:"+activity.getPackageName()));
+        activity.startActivityForResult(intent, SCHEDULE_EXACT_ALARM_PERMISSION_REQUEST);
+    }
+
+    private static boolean shouldRequestSpecialPermission(AppCompatActivity activity, String specialPerm) {
+        if (!requestingSpecialAppPermission) {
```
shubham1g5, line 223: what would cause `requestingSpecialAppPermission` to be `true` here ?

app/src/org/commcare/services/CommCareSessionService.java
```diff
@@ -236,8 +246,34 @@ public void showLoggedInNotification(@Nullable User user) {
             notificationBuilder.setContentText(contentText);
         }
 
-        //Send the notification.
-        this.startForeground(NOTIFICATION, notificationBuilder.build());
+        //Send the notification
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S && CommCareApplication.isAppInBackground())
```
shubham1g5, line 250: Have you been able to repro the crash with the foreground notification here ? I would think that since the foreground service has already started the app will always be in the foreground.

app/src/org/commcare/services/CommCareSessionService.java
```diff
@@ -236,8 +246,34 @@ public void showLoggedInNotification(@Nullable User user) {
...
+        receiverIntent.putExtra(EXTRA_NOTIFICATION_OBJ, notification);
+
+        PendingIntent pendingIntent = PendingIntent.getBroadcast(
+                this,
+                foregroundComponentType.ordinal(),
+                receiverIntent,
+                PendingIntent.FLAG_IMMUTABLE);
+
+        AlarmManager alarmmanager = (AlarmManager) getSystemService(ALARM_SERVICE);
+        alarmmanager.setExactAndAllowWhileIdle(AlarmManager.RTC_WAKEUP, Calendar.getInstance().getTimeInMillis(), pendingIntent);
```
shubham1g5, line 276: should check for `canScheduleExactAlarms()` before scheduling this.

shubham1g5 (changes requested):
The code looks good to me overall except a few comments and I do think this solution will work out for this issue in short term atleast. And I am happy for this to go out as a hotfix but would like QA to test this workflow in detail before we do so. 

Although I do think that the better approach here would be to get rid of foreground service altogether instead of adding another permission that user needs to grant with this special request framing.

PR #2660, opened 2023-04-12
ConnectID Initial Implementation

OrangeAndGreen:
## Summary
Initial development of ConnectID functionality, including registration workflow.

## Feature Flag
CommCare Connect

## Product Description
When enabled, adds UI methods to LoginController for user to sign in to ConnectID.
First-time user can create an account with the Connect server.

## Safety Assurance

### Automated test coverage
No tests.

### Safety story
This code is disabled by a hard-coded boolean at the top of the ConnectIDManager class, and when disabled results in no changes to existing functionality.

cross-request: https://github.com/dimagi/commcare-core/pull/1318


shubham1g5:
@OrangeAndGreen Is it possible to add screenshots for different screens this PR adds in the description in order to get more context around the functionality here. Thanks!

app/build.gradle
```diff
@@ -249,7 +250,7 @@ android {
     }
 
     defaultConfig {
-        minSdkVersion 16
+        minSdkVersion 23
```
shubham1g5, line 253: noting that we would not be able to merge these PRs with minSdk 23.

OrangeAndGreen:
cross-request: [dimagi/commcare-core#1290](https://github.com/dimagi/commcare-core/pull/1290)

app/build.gradle
OrangeAndGreen: This is fixed now. Instead of using the EncryptedFile helper class (required min SDK 23) I encrypt the Connect DB passphrase manually using a Cipher and store the encrypted passphrase in the global DB. Min SDK is back to 16!

OrangeAndGreen:
The various workflows are detailed via screenshots [here](https://docs.google.com/presentation/d/1RGOyqSonkGagU_y0OXdPXCdRpshzevWs7HZUrVJHGWQ/edit?usp=sharing).

app/src/org/commcare/activities/connect/ConnectIDManager.java
```diff
@@ -0,0 +1,565 @@
...
+
+    public static boolean shouldShowSignInMenuOption() {
+        return AppManagerDeveloperPreferences.isConnectIDEnabled() && getInstance().connectStatus == ConnectIDStatus.NotIntroduced;
+    }
+
+    public static boolean shouldShowSignOutMenuOption() {
+        return AppManagerDeveloperPreferences.isConnectIDEnabled() && getInstance().connectStatus == ConnectIDStatus.LoggedIn;
+    }
+
+    public static boolean isConnectIDActivity(int requestCode) {
```
shubham1g5, line 82: you should not need this method but can directly call `ConnectIDManager.handleFinishedActivity(requestCode, resultCode, intent)`, the `handleFinishedActivity` can return a boolean indicating whether it handled the activity response or not for calling activity to know whether the activity response was successfully handled by connect.

app/src/org/commcare/activities/connect/ConnectIDNetworkHelper.java
```diff
@@ -0,0 +1,267 @@
...
+            @Override
+            public void stopTaskTransition(int taskId) {}
+
+            @Override
+            public void hideTaskCancelButton() {}
+        });
+        postTask.executeParallel();
+    }
+
+    public static void post(Context context, String url, AuthInfo authInfo, Multimap<String, String> params, INetworkResultHandler handler) {
```
shubham1g5, line 106: we should be able to use this as the only `post` method here and always pass params as multimap.

app/src/org/commcare/activities/connect/ConnectIDNetworkHelper.java
```diff
@@ -0,0 +1,267 @@
...
+        void processSuccess(int responseCode, InputStream responseData);
+        void processFailure(int responseCode, IOException e);
+    }
+
+    public static void post(Context context, String url, AuthInfo authInfo, HashMap<String, String> params, INetworkResultHandler handler) {
+        Gson gson = new Gson();
+        String json = gson.toJson(params);
+
+        RequestBody requestBody = RequestBody.create(MediaType.parse("application/json"), json);
+        ModernHttpTask postTask =
```
shubham1g5, line 43: (Optional)There is potentially an opportunity here to experiment with Kotlin Coroutines instead of relying on async task framework. Async tasks are deprecated and we want to move away from them and Connect stuff is potentially idle to implement that first without introducing regressions in the main CC code.

app/src/org/commcare/utils/CommCareConnectUtils.java
```diff
@@ -0,0 +1,482 @@
...
+import javax.crypto.spec.SecretKeySpec;
+import javax.net.ssl.SSLException;
+
+import okhttp3.MediaType;
+import okhttp3.MultipartBody;
+import okhttp3.RequestBody;
+import okhttp3.ResponseBody;
+import retrofit2.Response;
+
+public class CommCareConnectUtils {
```
shubham1g5, line 52: should be removed ?

app/src/org/commcare/utils/EncryptionUtils.java
```diff
@@ -1,17 +1,212 @@
...
+            keystore.load(null);
+        }
+
+        KeyStore.Entry existingKey = keystore.getEntry(SECRET_NAME, null);
+        if (existingKey instanceof KeyStore.SecretKeyEntry) {
+            return ((KeyStore.SecretKeyEntry) existingKey).getSecretKey();
+        }
+
+        //Create key
+        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.M) {
```
shubham1g5, line 66: if we remove this check, what exactly fails on devices less than Android M ?

app/src/org/commcare/utils/EncryptionUtils.java
```diff
@@ -1,17 +1,212 @@
...
+            return generator.generateKey();
+        }
+
+        return null;
+    }
+
+    //Generate a random passphrase
+    public static byte[] generatePassphrase() {
+        Random random;
+        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.O) {
```
shubham1g5, line 85: What is api check protecting here ? I see `SecureRandom` is available before Android O.

app/src/org/commcare/utils/EncryptionUtils.java
```diff
@@ -1,17 +1,212 @@
...
+        } else {
+            random = new Random();
+        }
+
+        byte[] result = new byte[PASSPHRASE_LENGTH];
+
+        while (true) {
+            random.nextBytes(result);
+
+            //Make sure there are no zeroes in the passphrase (bad for SQLCipher)
```
shubham1g5, line 100: would be great to add a reference on why it's bad for SqlCipher.

app/src/org/commcare/utils/EncryptionUtils.java
```diff
@@ -1,17 +1,212 @@
...
+    //Gets the SecretKey from the Android KeyStore (creates a new one the first time)
+    private static SecretKey getKey()
+            throws CertificateException, KeyStoreException, IOException, NoSuchAlgorithmException,
+            UnrecoverableEntryException, InvalidAlgorithmParameterException {
+        if(keystore == null) {
+            keystore = KeyStore.getInstance("AndroidKeyStore");
+            keystore.load(null);
+        }
+
+        KeyStore.Entry existingKey = keystore.getEntry(SECRET_NAME, null);
```
shubham1g5, line 60: Does this key ever change ? For eg. if we clear the app data on the phone.

shubham1g5 (changes requested):
I plan to take another pass sometime later on these changes around UI specifically, but putting some thoughts I had meanwhile.

avazirna:

avazirna:

app/assets/locales/android_translatable_strings.txt
```diff
@@ -175,16 +177,115 @@ login.password=Password
 login.pin.password=PIN
 login.primed.prompt=Your password has been saved for you! Just press 'Log In'.
 login.button=Log In
+login.button.connect=Sign in with ConnectID
```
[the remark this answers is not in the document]
OrangeAndGreen: Ah, that makes sense, I'll move them over.

app/src/org/commcare/CommCareApplication.java
```diff
@@ -203,7 +203,7 @@ public class CommCareApplication extends MultiDexApplication {
     public void onCreate() {
         super.onCreate();
 
-        turnOnStrictMode();
+        //turnOnStrictMode();
```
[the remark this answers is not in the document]
OrangeAndGreen: Great!

app/src/org/commcare/activities/BiometricsHelper.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    private static final int StrongBiometric = BIOMETRIC_STRONG;
+    private static final int PinBiometric = DEVICE_CREDENTIAL;
+
+    public static ConfigurationStatus checkFingerprintStatus(BiometricManager biometricManager) {
+        return checkStatus(biometricManager, StrongBiometric);
+    }
+
+    public static boolean isFingerprintAvailable(BiometricManager biometricManager) {
```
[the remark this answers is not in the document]
OrangeAndGreen: Correct, this method returns true only if the fingerprint sensor is available (may or may not be configured yet).

app/src/org/commcare/activities/connect/ConnectIDDatabaseHelper.java
```diff
@@ -0,0 +1,102 @@
...
+            }
+
+            //If we get here, the passphrase hasn't been created yet
+            byte[] passphrase = EncryptionUtils.generatePassphrase();
+            ConnectKeyRecord record = new ConnectKeyRecord("", EncryptionUtils.encryptToBase64String(passphrase));
+            CommCareApplication.instance().getGlobalStorage(ConnectKeyRecord.class).write(record);
+            return passphrase;
+        }
+
+        return null;
```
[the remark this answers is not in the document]
OrangeAndGreen: Ah, yes it does. There should be a TODO on this, not sure what to do when we can't use the keystore to encrypt the Connect DB...

app/src/org/commcare/utils/CommCareConnectUtils.java
OrangeAndGreen: Oops, yes it should!

app/src/org/commcare/activities/connect/ConnectIDKeyboardHelper.java
```diff
@@ -0,0 +1,22 @@
+package org.commcare.activities.connect;
+
+import android.app.Activity;
+import android.content.Context;
+import android.view.View;
+import android.view.inputmethod.InputMethodManager;
+
+public class ConnectIDKeyboardHelper {
```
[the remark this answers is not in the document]
OrangeAndGreen: Sounds good, I'll move it over.

app/src/org/commcare/activities/connect/ConnectIDNetworkHelper.java
OrangeAndGreen: These two post() functions are actually sending data differently. The first version encodes the params input as JSON in the body, while the second uses form encoding. I use the first call for everything going to the ConnectID server EXCEPT for OAuth, which requires the second call. Perhaps the functions should have different names to be more clear, i.e. postWithJson and postWithFormEncoding?

app/src/org/commcare/utils/EncryptionUtils.java
OrangeAndGreen: It's the call to `getInstanceStrong()` that isn't supported in earlier versions. I'll add some code that attempts to create the `SecureRandom` a different way when possible (which includes specifying the algorithm).

app/src/org/commcare/utils/EncryptionUtils.java
OrangeAndGreen: My understanding [from](https://stackoverflow.com/questions/47343659/does-clear-data-on-app-settings-clears-all-android-keystore-stored-for-that-cert) here is that the keystore data for an app is erased when the user clears all app data. But since the DBs are also erased in that case, a new DB will have to be created and it's okay for a different key to be generated and used going forward.

app/src/org/commcare/utils/EncryptionUtils.java
OrangeAndGreen: No problem

app/src/org/commcare/utils/EncryptionUtils.java
OrangeAndGreen: Everything inside the if-block requires at least Android M , starting with the `KeyGenParameterSpec.Builder()` constructor, including the 4 setters called on the `builder`, and the final call to `builder.build()`.

The two (private) callers for this method both require Android M, so depending on how we want older versions of Android to handle this we should make this method consistent with the two callers.

app/src/org/commcare/activities/connect/ConnectIDDatabaseHelper.java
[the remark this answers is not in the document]
OrangeAndGreen: This has been solved now. In earlier versions of Android (before M), a RSA key pair is generated manually. The public key is used to encrypt the DB passphrase and then discarded, while the private key is stored in the KeyStore so the passphrase can be decrypted in the future.

app/src/org/commcare/utils/EncryptionUtils.java
OrangeAndGreen: This code was replaced in the implementation that supports Android before version M.

app/src/org/commcare/activities/BiometricsHelper.java
```diff
@@ -0,0 +1,81 @@
+package org.commcare.activities;
```
shubham1g5, line 1: Maybe better to move it out of `org.commcare.utils`

app/src/org/commcare/utils/EncryptionUtils.java
```diff
@@ -1,17 +1,274 @@
...
+    //Decryption backed by Android KeyStore
+    private static byte[] decrypt(Context context, byte[] bytes)
+            throws NoSuchPaddingException, NoSuchAlgorithmException, InvalidAlgorithmParameterException,
+            InvalidKeyException, IllegalBlockSizeException, BadPaddingException,
+            UnrecoverableEntryException, CertificateException, KeyStoreException, IOException, NoSuchProviderException {
+        int readIndex = 0;
+        int ivLength = bytes[readIndex];
+        readIndex++;
+        if(ivLength < 0) {
+            //TODO: Early chance to catch decryption error
```
shubham1g5, line 227: probably should throw a specific exception from here for callers to catch and act on

app/src/org/commcare/utils/EncryptionUtils.java
```diff
@@ -1,17 +1,274 @@
...
+        }
+        else {
+            transformation = "RSA/ECB/PKCS1Padding";
+        }
+
+        return transformation;
+    }
+
+    //Encryption backed by Android KeyStore
+    private static byte[] encrypt(Context context, byte[] bytes)
```
shubham1g5, line 189: Would be nice to write basic tests for `encrypt` and `decrypt` methods in this class.

app/src/org/commcare/utils/EncryptionUtils.java
```diff
@@ -1,17 +1,274 @@
...
+                    .setBlockModes(KeyProperties.BLOCK_MODE_CBC)
+                    .setEncryptionPaddings(KeyProperties.ENCRYPTION_PADDING_PKCS7)
+                    //.setUserAuthenticationValidityDurationSeconds(AUTH_VALID_DURATION_IN_SECOND)
+                    .build();
+
+            keyGenerator.init(keySpec);
+            return keyGenerator.generateKey();
+        }
+        else {
+            KeyPairGenerator generator = KeyPairGenerator.getInstance("RSA", KEYSTORE_NAME);
```
shubham1g5, line 114: The [SO post](https://stackoverflow.com/a/50389146/3811963) actually suggests to not use this key pair directly for encryption but use it to encrypt another AES key that we generate and actually use for encrypting the data. That way we can always use AES encryption to encrypt data and the use of RSA key pair is just for making the key secure to be able to store it on the device.

shubham1g5 (changes requested):
Left a bunch of nits and a very few major action items. Another thing we should do before merge is format your code in Android Studio using [our Style files](https://github.com/dimagi/commcare-android#code-style-settings).

shubham1g5:

app/src/org/commcare/activities/BiometricsHelper.java
OrangeAndGreen: No problem.

app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -392,13 +409,54 @@ public void dataPullCompleted() {
         ViewUtil.hideVirtualKeyboard(LoginActivity.this);
         CommCareApplication.notificationManager().clearNotifications(NOTIFICATION_MESSAGE_LOGIN);
 
+        String appId = CommCareApplication.instance().getCurrentApp().getUniqueId();
+        String username = uiController.getEnteredUsername();
+        String pass = uiController.getEnteredPasswordOrPin();
+        ConnectIDManager.rememberAppCredentials(appId, username, pass);
+
```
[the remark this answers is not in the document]
OrangeAndGreen: Done.

app/src/org/commcare/activities/connect/ConnectIDDatabaseHelper.java
```diff
@@ -0,0 +1,122 @@
...
+                    }
+                    return connectDatabase;
+                }
+            }
+        });
+    }
+
+    public static ConnectUserRecord getUser(Context context) {
+        ConnectUserRecord user = null;
+        for (ConnectUserRecord r : getConnectStorage(context, ConnectUserRecord.class)) {
```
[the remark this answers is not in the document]
OrangeAndGreen: No, right now we only support one Connect user on the device.

app/src/org/commcare/activities/connect/ConnectIDDatabaseHelper.java
```diff
@@ -0,0 +1,122 @@
...
+
+        return user;
+    }
+
+    public static void storeUser(Context context, ConnectUserRecord user) {
+        getConnectStorage(context, ConnectUserRecord.class).write(user);
+    }
+
+    public static void forgetUser(Context context) {
+        getConnectStorage(context, ConnectUserRecord.class).removeAll();//(getUser().getID());
```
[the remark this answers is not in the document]
OrangeAndGreen: Done.

app/src/org/commcare/activities/connect/ConnectIDDatabaseHelper.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    public static ConnectLinkedAppRecord getAppData(Context context, String appId, String username) {
+        ConnectLinkedAppRecord record = null;
+        for (ConnectLinkedAppRecord r : getConnectStorage(context, ConnectLinkedAppRecord.class)) {
```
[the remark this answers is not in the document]
OrangeAndGreen: Done.

app/src/org/commcare/activities/connect/ConnectIDDatabaseHelper.java
```diff
@@ -0,0 +1,122 @@
...
+    private static byte[] getConnectDBPassphrase(Context context) {
+        try {
+            for (ConnectKeyRecord r : CommCareApplication.instance().getGlobalStorage(ConnectKeyRecord.class)) {
+                return EncryptionUtils.decryptFromBase64String(context, r.getEncryptedPassphrase());
+            }
+
+            //If we get here, the passphrase hasn't been created yet
+            byte[] passphrase = EncryptionUtils.generatePassphrase();
+
+            ConnectKeyRecord record = new ConnectKeyRecord("", EncryptionUtils.encryptToBase64String(context, passphrase));
```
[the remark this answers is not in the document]
OrangeAndGreen: Yes it was, I removed the field.

app/src/org/commcare/utils/EncryptionUtils.java
```diff
@@ -1,17 +1,274 @@
...
+        return keystoreSingleton;
+    }
+
+    //Gets the SecretKey from the Android KeyStore (creates a new one the first time)
+    private static Key getKey(Context context, boolean trueForEncrypt)
+            throws CertificateException, KeyStoreException, IOException, NoSuchAlgorithmException,
+            UnrecoverableEntryException, InvalidAlgorithmParameterException, NoSuchProviderException {
+        KeyStore keystore = getKeystore();
+
+        if(doesKeystoreContainEncryptionKey()) {
```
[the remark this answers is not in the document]
OrangeAndGreen: Done.

app/src/org/commcare/activities/connect/ConnectIDDatabaseHelper.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.Vector;
+
+public class ConnectIDDatabaseHelper {
+    private static final Object connectDbHandleLock = new Object();
+    private static SQLiteDatabase connectDatabase;
+
+    private static byte[] getConnectDBPassphrase(Context context) {
+        try {
+            for (ConnectKeyRecord r : CommCareApplication.instance().getGlobalStorage(ConnectKeyRecord.class)) {
+                return EncryptionUtils.decryptFromBase64String(context, r.getEncryptedPassphrase());
```
shubham1g5, line 30: This crashes for me sometimes as - 

````
W/System.err: java.lang.ArrayIndexOutOfBoundsException: length=0; index=0
W/System.err:     at org.commcare.utils.EncryptionUtils.decrypt(EncryptionUtils.java:223)
W/System.err:     at org.commcare.utils.EncryptionUtils.decryptFromBase64String(EncryptionUtils.java:268)
W/System.err:     at org.commcare.activities.connect.ConnectIDDatabaseHelper.getConnectDBPassphrase(ConnectIDDatabaseHelper.java:30)
W/System.err:     at org.commcare.activities.connect.ConnectIDDatabaseHelper.init(ConnectIDDatabaseHelper.java:48)
W/System.err:     at org.commcare.activities.connect.ConnectIDManager.init(ConnectIDManager.java:60)
W/System.err:     at org.commcare.activities.LoginActivity.onCreate(LoginActivity.java:125)
````

We probbaly need to either ensure ` r.getEncryptedPassphrase()` can't be null or handle code for it being null

File: app/src/org/commcare/activities/connect/ConnectIDPhoneVerificationActivity.java [hunk 26]
```diff
@@ -0,0 +1,281 @@
...
+        String url = getString(R.string.ConnectURL) + command;
+
+        ConnectIDNetworkHelper.post(this, url, authInfo, params, false, new ConnectIDNetworkHelper.INetworkResultHandler() {
+            @Override
+            public void processSuccess(int responseCode, InputStream responseData) {
+                try {
+                    String responseAsString = new String(StreamsUtil.inputStreamToByteArray(responseData));
+                    if(responseAsString.length() > 0) {
+                        JSONObject json = new JSONObject(responseAsString);
+                        String key = "secret";
```
shubham1g5, line 158: any json keys should be defined as constants.

app/src/org/commcare/models/database/connect/DatabaseConnectOpenHelper.java
```diff
@@ -0,0 +1,74 @@
...
+        public void onCreate(SQLiteDatabase database) {
+            database.beginTransaction();
+            try {
+                TableBuilder builder = new TableBuilder(ConnectUserRecord.class);
+                database.execSQL(builder.getTableCreateString());
+
+                builder = new TableBuilder(ConnectLinkedAppRecord.class);
+                database.execSQL(builder.getTableCreateString());
+
+                DbUtil.createNumbersTable(database);
```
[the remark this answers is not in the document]
shubham1g5: Based on Clayton's comment on slack, lets keep it just to have a  consistent `SqlStorage` implementation.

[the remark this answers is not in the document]
ctsims: Agreed. We would need to extract the CoveringIndexIterator from the implementation if we wanted to remove it, which isn't a huge amount of work but seems like added complexity for something that's stable.

app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -392,13 +409,54 @@ public void dataPullCompleted() {
...
+    }
+
+    private void updateConnectButton() {
+        uiController.setConnectButtonText(ConnectIDManager.getConnectButtonText(this));
+        uiController.setConnectButtonVisible(ConnectIDManager.shouldShowConnectButton());
+
+        uiController.updateConnectLoginState();
+    }
+
+    private void checkForSavedCredentials() {
```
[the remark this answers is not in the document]
OrangeAndGreen: Agreed, although I'll use isUnlocked instead since this code only matters if the account is configured AND unlocked.

app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -532,35 +613,65 @@ protected void restoreEnteredTextFromRotation() {
...
+    protected boolean populateAppSpinner(ArrayList<ApplicationRecord> readyApps) {
         ArrayList<String> appNames = new ArrayList<>();
+
         appIdDropdownList.clear();
+
+        boolean includeDefault = ConnectIDManager.isConnectIDIntroduced() && !ConnectIDManager.isSignedIn();
+        if(includeDefault) {
+            appNames.add(Localization.get("login.app.direct"));
+            appIdDropdownList.add("");
+        }
```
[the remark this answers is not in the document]
OrangeAndGreen: The default that this code refers to is an additional option that says "Login to your app directly" and server as a placeholder when we're trying to encourage the user to unlock ConnectID instead of proceeding directly to their traditional credentials. When ConnectID is not introduced or is already unlocked, the selector is set to the currently seated app. Choosing the selector value happens a few lines later, around lines 640-648.

app/src/org/commcare/activities/connect/ConnectIDPhoneVerificationActivity.java
OrangeAndGreen: Done.

app/src/org/commcare/activities/connect/ConnectIDNetworkHelper.java
OrangeAndGreen: Done.

app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -532,35 +613,65 @@ protected void restoreEnteredTextFromRotation() {
         }
     }
 
-    protected void populateAppSpinner(ArrayList<ApplicationRecord> readyApps) {
+    protected boolean populateAppSpinner(ArrayList<ApplicationRecord> readyApps) {
         ArrayList<String> appNames = new ArrayList<>();
+
         appIdDropdownList.clear();
+
+        boolean includeDefault = ConnectIDManager.isConnectIDIntroduced() && !ConnectIDManager.isSignedIn();
```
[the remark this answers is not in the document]
OrangeAndGreen: Done, added a requiresUnlock() function and using that.

app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -532,35 +613,65 @@ protected void restoreEnteredTextFromRotation() {
...
-            SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(this);
-            prefs.edit().putString(KEY_LAST_APP, appId).commit();
-
-            // Launch the activity to seat the new app
-            Intent i = new Intent(this, SeatAppActivity.class);
-            i.putExtra(SeatAppActivity.KEY_APP_TO_SEAT, appId);
-            this.startActivityForResult(i, SEAT_APP_ACTIVITY);
+        if(!ConnectIDManager.isConnectIDIntroduced() || ConnectIDManager.isSignedIn() || position > 0) {
+            // Retrieve the app record corresponding to the app selected
+            selectedAppIndex = position;// - (ConnectIDManager.isConnectIDIntroduced() ? 1 : 0);
```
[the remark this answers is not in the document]
OrangeAndGreen: Done

app/src/org/commcare/activities/connect/ConnectIDManager.java
```diff
@@ -0,0 +1,575 @@
...
+                }
+                else {
+                    Date passwordDate = user.getLastPasswordDate();
+                    boolean forcePassword = passwordDate == null;
+                    if (!forcePassword) {
+                        //See how much time has passed since last password login
+                        long millis = (new Date()).getTime() - passwordDate.getTime();
+                        long days = TimeUnit.DAYS.convert(millis, TimeUnit.MILLISECONDS);
+                        forcePassword = days >= 7;
+                    }
```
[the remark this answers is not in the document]
OrangeAndGreen: Done

app/src/org/commcare/activities/connect/ConnectIDManager.java
```diff
@@ -0,0 +1,575 @@
...
+                nextActivity = ConnectIDPasswordVerificationActivity.class;
+
+                params.put(ConnectIDConstants.PHONE, recoveryPhone);
+                params.put(ConnectIDConstants.SECRET, recoverySecret);
+            }
+            case ConnectIDConstants.CONNECT_RECOVERY_ALT_PHONE_MESSAGE -> {
+                //Show message screen indicating plan to use alt phone
+                nextActivity = ConnectIDMessageActivity.class;
+
+                params.put(ConnectIDConstants.TITLE, parentActivity.getString(R.string.connect_recovery_alt_title));
```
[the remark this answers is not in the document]
OrangeAndGreen: Done

app/src/org/commcare/utils/EncryptionUtils.java
```diff
@@ -1,17 +1,274 @@
...
+
+    private static Key generateKeyInKeystore(Context context, boolean trueForEncrypt) throws NoSuchAlgorithmException, NoSuchProviderException, InvalidAlgorithmParameterException {
+        if(android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.M) {
+            KeyGenerator keyGenerator = KeyGenerator.getInstance(
+                    KeyProperties.KEY_ALGORITHM_AES, KEYSTORE_NAME);
+            KeyGenParameterSpec keySpec = new KeyGenParameterSpec.Builder(SECRET_NAME,
+                    KeyProperties.PURPOSE_ENCRYPT | KeyProperties.PURPOSE_DECRYPT)
+                    .setBlockModes(KeyProperties.BLOCK_MODE_CBC)
+                    .setEncryptionPaddings(KeyProperties.ENCRYPTION_PADDING_PKCS7)
+                    //.setUserAuthenticationValidityDurationSeconds(AUTH_VALID_DURATION_IN_SECOND)
```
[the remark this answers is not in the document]
OrangeAndGreen: We aren't restricting usage of the key based on user authentication in that way, i.e. we never make a call to setUserAuthenticationRequired(), so this setting doesn't apply. I removed the commented line.

app/src/org/commcare/activities/connect/ConnectIDDatabaseHelper.java
OrangeAndGreen: I added a null check for the encryptedPassphrase, although I'm confused as to how it ever could have been null since the ConnectKeyRecord would only ever be created with a valid passphrase, i.e. later in the same function.

app/src/org/commcare/activities/connect/ConnectIDManager.java
```diff
@@ -0,0 +1,575 @@
...
+    }
+
+    private void continueWorkflow() {
+        //Determine activity to launch for next phase
+        Class<?> nextActivity = null;
+        Map<String, String> params = new HashMap<>();
+        ConnectUserRecord user = ConnectIDDatabaseHelper.getUser(parentActivity);
+
+        switch (phase) {
+            case ConnectIDConstants.CONNECT_REGISTER_OR_RECOVER_DECISION -> {
```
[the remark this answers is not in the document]
OrangeAndGreen: I created a ConnectIDTask enum and baked the requestCode and nextActivity into that. The params are loaded dynamically so are still in ConnectIDManager.continueWorkflow()

app/src/org/commcare/utils/EncryptionUtils.java
OrangeAndGreen: Done

app/src/org/commcare/android/database/connect/models/ConnectLinkedAppRecord.java
```diff
@@ -11,11 +12,16 @@ public class ConnectLinkedAppRecord extends Persisted {
      */
     public static final String STORAGE_KEY = "app_info";
 
+    public static final String META_APP_ID = "app_id";
+    public static final String META_USER_ID = "user_id";
+
     @Persisting(1)
-    private String appID;
+    @MetaField(META_APP_ID)
+    private String app_id;
```
shubham1g5, line 22: why change is to `_`  ? We should keep them camelcased as before if possible as per Java conventions.

app/src/org/commcare/android/database/global/models/ConnectKeyRecord.java
```diff
@@ -7,18 +7,14 @@
 @Table(ConnectKeyRecord.STORAGE_KEY)
 public class ConnectKeyRecord extends Persisted {
     public static final String STORAGE_KEY = "connect_key";
-
-    @Persisting(1)
-    String userId;
```
shubham1g5: curious why we removed `userID` ? Is it because only one user can exist on a device ?

app/src/org/commcare/activities/connect/ConnectIDDatabaseHelper.java
shubham1g5: I think `null` check here is more harmful than useful since it will silently eat those errors. I am happy with us removing the `null` check and let the code crash so that we fail fast if we don't see anyway `r.getEncryptedPassphrase()` can be `null`. Also noticed that this whole code block is in a `try catch` with us swallowing the exceptions, wondering if it's better to let app crash in those scenarios as well.

app/src/org/commcare/activities/connect/ConnectIDNetworkHelper.java
```diff
@@ -45,6 +45,8 @@ public PostResult(int responseCode, InputStream responseStream, IOException e) {
         }
     }
 
+    private static boolean isBusy = false;
```
shubham1g5, line 48: we should make `ConnectIDNetworkHelper` a singleton otherwise multiple threads can still do duplicate calls.

shubham1g5 (changes requested):

app/src/org/commcare/android/database/connect/models/ConnectLinkedAppRecord.java
OrangeAndGreen: Done

app/src/org/commcare/android/database/global/models/ConnectKeyRecord.java
OrangeAndGreen: That's correct. If we decide later to support multiple CID users on a device, this may or may not need to change depending on whether we decide to create separate ConnectID databases for each user or use on DB for all.

app/src/org/commcare/activities/connect/ConnectIDDatabaseHelper.java
OrangeAndGreen: The try-catch at this point in the code is to avoid having a bunch of possible exceptions continue bubbling up the call-chain  such that more and more functions have to add a bunch of `throws` to their signatures. I'm now throwing a `RuntimeException` wrapping the original exception in that `catch` block, that way we can still bubble the exception up and let the app crash without having to add a bunch of `throws` elsewhere. Does that work?

app/src/org/commcare/activities/connect/ConnectIDManager.java
```diff
@@ -558,48 +565,126 @@ public static void handleFinishedActivity(int requestCode, int resultCode, Inten
...
+        ConnectUserRecord user = ConnectIDDatabaseHelper.getUser(manager.parentActivity);
+
+        HashMap<String, String> params = new HashMap<>();
+        params.put("client_id", "zqFUtAAMrxmjnC1Ji74KAa6ZpY1mZly0J0PlalIa");
+        params.put("scope", "openid");
+        params.put("grant_type", "password");
+        params.put("username", user.getUserID());
+        params.put("password", user.getPassword());
+
+        String url = manager.parentActivity.getString(R.string.ConnectURL) + "/o/token/";
```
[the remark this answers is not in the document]
OrangeAndGreen: Done

app/src/org/commcare/activities/connect/ConnectIDManager.java
```diff
@@ -558,48 +565,126 @@ public static void handleFinishedActivity(int requestCode, int resultCode, Inten
...
+
+        ConnectIDNetworkHelper.post(manager.parentActivity, url, new AuthInfo.NoAuth(), params, true, new ConnectIDNetworkHelper.INetworkResultHandler() {
+            @Override
+            public void processSuccess(int responseCode, InputStream responseData) {
+                try {
+                    String responseAsString = new String(StreamsUtil.inputStreamToByteArray(responseData));
+                    JSONObject json = new JSONObject(responseAsString);
+                    String key = "access_token";
```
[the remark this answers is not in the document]
OrangeAndGreen: Done

app/src/org/commcare/activities/connect/ConnectIDManager.java
```diff
@@ -558,48 +565,126 @@ public static void handleFinishedActivity(int requestCode, int resultCode, Inten
...
 
-    //TODO: Re-enable this once we're ready for OIDC
-//    public static void getConnectToken() {
-//        ConnectIDManager manager = getInstance();
-//        ConnectUserRecord user = ConnectIDDatabaseHelper.getUser(manager.parentActivity);
-//
-//        HashMap<String, String> params = new HashMap<>();
-//        params.put("client_id", "zqFUtAAMrxmjnC1Ji74KAa6ZpY1mZly0J0PlalIa");
-//        params.put("scope", "openid");
+    public static void getConnectToken(String hqUsername, String hqPassword) {
```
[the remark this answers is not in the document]
OrangeAndGreen: Done

app/src/org/commcare/activities/connect/ConnectIDManager.java
```diff
@@ -558,48 +565,126 @@ public static void handleFinishedActivity(int requestCode, int resultCode, Inten
...
-//        ConnectIDNetworkHelper.post(manager.parentActivity, url, new AuthInfo.NoAuth(), params, true, new ConnectIDNetworkHelper.INetworkResultHandler() {
-//            @Override
-//            public void processSuccess(int responseCode, InputStream responseData) {
+//        params.put("username", hqUsername + "@first-project-2.commcarehq.org");
+        params.put("token", connectToken);
+
+        String url = ServerUrls.getKeyServer().replace("phone/keys/", "settings/users/commcare/link_connectid_user/");
+        //https://www.commcarehq.org/a/dimagi-intro/phone/keys/
+
+        //localhost:8000/oauth/token?grant_type=password&client_id=CLIENT_ID&username=100@commcarehq.commcarehq.org&password=PASS&scope=sync
```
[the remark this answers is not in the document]
OrangeAndGreen: Oops, sorry! These are gone now.

app/src/org/commcare/activities/connect/ConnectIDManager.java
```diff
@@ -558,48 +565,126 @@ public static void handleFinishedActivity(int requestCode, int resultCode, Inten
...
+        //String url = manager.parentActivity.getString(R.string.ConnectURL) + "/o/token/";
+
+        ConnectIDNetworkHelper.post(manager.parentActivity, url, new AuthInfo.NoAuth(), params, true, new ConnectIDNetworkHelper.INetworkResultHandler() {
+            @Override
+            public void processSuccess(int responseCode, InputStream responseData) {
+                try {
+                    String responseAsString = new String(StreamsUtil.inputStreamToByteArray(responseData));
+                    JSONObject json = new JSONObject(responseAsString);
+                    String key = "access_token";
```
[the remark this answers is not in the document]
OrangeAndGreen: Done

OrangeAndGreen:
cross-request: [https://github.com/dimagi/commcare-core/pull/1290](https://github.com/dimagi/commcare-core/pull/1318)

app/src/org/commcare/CommCareApplication.java
```diff
@@ -1153,10 +1154,18 @@ public ModernHttpRequester buildHttpRequester(Context context, String url,
...
                                                   @Nullable HttpResponseProcessor responseProcessor, boolean retry) {
 
-        CommCareNetworkService networkService;
+        CommCareNetworkService networkService = null;
         if (authInfo instanceof AuthInfo.NoAuth) {
             networkService = CommCareNetworkServiceGenerator.createNoAuthCommCareNetworkService();
-        } else {
+        } else if(authInfo instanceof AuthInfo.CurrentAuth) {
+            //Try to get SSO token
+            AuthInfo.TokenAuth tokenAuth = ConnectIDManager.acquireSSOTokenSync();
```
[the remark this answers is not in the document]
OrangeAndGreen: Indeed! I created a corresponding PR for core [here](https://github.com/dimagi/commcare-core/pull/1318)

app/src/org/commcare/activities/connect/ConnectIDNetworkHelper.java
OrangeAndGreen: Done

app/src/org/commcare/activities/connect/ConnectIDNetworkHelper.java
```diff
@@ -0,0 +1,262 @@
...
+
+        RequestBody requestBody = RequestBody.create(MediaType.parse("application/json"), json);
+        ModernHttpTask postTask =
+                new ModernHttpTask(context, url,
+                        ImmutableMultimap.of(),
+                        new HashMap<>(),
+                        requestBody,
+                        HTTPMethod.POST,
+                        authInfo);
+        postTask.connect(new ConnectorWithHttpResponseProcessor<>() {
```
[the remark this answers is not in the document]
OrangeAndGreen: Done

app/src/org/commcare/utils/EncryptionUtils.java
OrangeAndGreen: I could be mistaken (still getting comfortable with cryptography), but I think the current code is as suggested in that SO post. The post says (for API < 23)  we should use an RSA key pair to encrypt a random AES key, with the RSA key stored in the KeyStore and the encrypted AES key stored somewhere easily accessible. Then later we can retrieve the RSA key, decrypt the AES key, and finally access our DB.

In our implementation, SQLCipher handles the AES encryption. We use ConnectIDDatabaseHelper.getConnectDBPassphrase() to handle retrieving (or creating the first time) the AES key. The first time accessing the DB, we create the AES key, encrypt it, encode it to a base64 string, and store it in the CommCare Global DB. Diving into the code that encrypts the AES key: EncryptionUtils.getKey() looks for an existing key in the KeyStore, and otherwise calls generateInKeystore(). This is the function that generates the key in the KeyStore directly (for API >= 23) or manually generates an RSA key pair to store (for earlier APIs).

By the time we attempt to access the encrypted Connect DB, the RSA key has already been generated and stored in the KeyStore, and the AES key has been generated, encrypted, encoded, and stored in the global DB. So now it's just a matter of retrieving the RSA key, decrypting the AES key, and passing it to the Connect DB.

app/src/org/commcare/activities/connect/ConnectIDDatabaseHelper.java
shubham1g5: yes that's perfect.

app/src/org/commcare/activities/connect/ConnectIDManager.java
```diff
@@ -558,48 +565,126 @@ public static void handleFinishedActivity(int requestCode, int resultCode, Inten
...
+        ConnectUserRecord user = ConnectIDDatabaseHelper.getUser(manager.parentActivity);
+
+        HashMap<String, String> params = new HashMap<>();
+        params.put("client_id", "4eHlQad1oasGZF0lPiycZIjyL0SY1zx7ZblA6SCV");
+        params.put("scope", "sync");
+        params.put("grant_type", "password");
+        params.put("username", hqUsername + "@first-project-2.commcarehq.org");
+        params.put("password", connectToken);
+
+        String url = ServerUrls.getKeyServer().replace("a/first-project-2/phone/keys/", "oauth/token/");
```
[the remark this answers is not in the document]
shubham1g5: @calellowitz bumping for your thoughts.

app/src/org/commcare/utils/EncryptionUtils.java
OrangeAndGreen: Done, I added a new EncryptionUtilsTest class with a test for the encrypt and decrypt functions. Note that the test generates its own RSA key pair to use, since the KeyStore isn't available in the test environment.

app/src/org/commcare/utils/EncryptionUtils.java
shubham1g5: Thanks for the detailed explanation. That makes sense to me clears things up.

Although I think there is a fault in this whole mechanism that we are depening on the OS version to decide the Encryption algorithm [here](https://github.com/dimagi/commcare-android/blob/dv/connect_id/app/src/org/commcare/utils/EncryptionUtils.java#L174). This can lock an user out from their data after doing an OS update. We somehow need to know the `transformation` based on the `key` stored in keystore itself to avoid this.

app/res/values/strings.xml
```diff
@@ -450,4 +452,144 @@
...
+
+    <!-- ConnectID strings -->
+
+    <string name="connect_id_enabled" cc:translatable="true">ConnectID Enabled</string>
+    <string name="login_button_connect" cc:translatable="true">Unlock ConnectID</string>
+    <string name="login_welcome_connect_signed_out" cc:translatable="true">Unlock ConnectID first and then enter your app.</string>
+    <string name="login_welcome_connect_signed_in" cc:translatable="true">Welcome %s!</string>
+
+    <string name="choice_or" cc:translatable="true">_____________    OR    _____________</string>
+    <string name="busy_message" cc:translatable="true">Still busy…</string>
```
shubham1g5, line 466: Maybe give a more descriptive message - `We are still busy processing your previous request. Please try again in some time`

shubham1g5 (changes requested):

app/src/org/commcare/activities/connect/ConnectIDPhoneVerificationActivity.java
```diff
@@ -0,0 +1,298 @@
...
+            @Override
+            public void processFailure(int responseCode, IOException e) {
+                String message = "";
+                if (responseCode > 0) {
+                    message = String.format(Locale.getDefault(), "(%d)", responseCode);
+                } else if (e != null) {
+                    message = e.toString();
+                }
+                logRecoveryResult(false);
+                uiController.setErrorMessage(String.format("Error verifying SMS code. %s", message));
```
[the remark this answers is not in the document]
OrangeAndGreen: Done

app/res/values/strings.xml
```diff
@@ -450,4 +452,144 @@
...
+
+
+    <string name="connect_verify_fingerprint" cc:translatable="true">Fingerprint</string>
+    <string name="connect_verify_pin" cc:translatable="true">PIN</string>
+    <string name="connect_verify_button_configured" cc:translatable="true">Continue</string>
+    <string name="connect_verify_button_password" cc:translatable="true">Use Password Only</string>
+
+    <string name="connect_verify_phone_title" cc:translatable="true">Verification Code</string>
+    <string name="connect_verify_phone_title2" cc:translatable="true">Enter verification code</string>
+    <string name="connect_verify_phone_label" cc:translatable="true">We sent you a code on number %s.</string>
```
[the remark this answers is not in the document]
OrangeAndGreen: I changed the message to indicate that we attempted to send an OTP. If the request fails an error message shows up a little further down the page.

app/src/org/commcare/activities/connect/ConnectIDPhoneActivity.java
```diff
@@ -0,0 +1,227 @@
...
+                        boolean isBusy = !ConnectIDNetworkHelper.get(this, url, new AuthInfo.NoAuth(), params, new ConnectIDNetworkHelper.INetworkResultHandler() {
+                            @Override
+                            public void processSuccess(int responseCode, InputStream responseData) {
+                                uiController.setAvailabilityText(getString(R.string.connect_phone_available));
+                                uiController.setOkButtonEnabled(true);
+                            }
+
+                            @Override
+                            public void processFailure(int responseCode, IOException e) {
+                                String text = getString(R.string.connect_phone_unavailable);
```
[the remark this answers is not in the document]
OrangeAndGreen: Done.

app/src/org/commcare/activities/connect/ConnectIDPasswordActivity.java
```diff
@@ -0,0 +1,143 @@
...
+
+    public void handleButtonPress() {
+        String password = uiController.getPasswordText();
+
+        HashMap<String, String> params = new HashMap<>();
+        AuthInfo authInfo;
+        String command;
+        if (username != null && username.length() > 0 && oldPassword != null && oldPassword.length() > 0) {
+            authInfo = new AuthInfo.ProvidedAuth(username, oldPassword, false);
+            command = "/users/change_password";
```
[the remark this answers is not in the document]
OrangeAndGreen: Done.

app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -392,13 +406,63 @@ public void dataPullCompleted() {
...
+        String pass = uiController.getEnteredPasswordOrPin();
+        ConnectIDManager.rememberAppCredentials(appId, username, pass);
+    }
+
+    public void handleConnectButtonPress() {
+        selectedAppIndex = -1;
+        ConnectIDManager.handleConnectButtonPress(success -> {
+            updateConnectButton();
+
+            //NOTE: Enable this to launch new app picker activity
```
shubham1g5, line 430: should remove it if not needed anymore.

app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -392,13 +406,63 @@ public void dataPullCompleted() {
...
+            String selectedAppId = appIdDropdownList.size() > 0 ? appIdDropdownList.get(selectorIndex) : "";
+            String seatedAppId = CommCareApplication.instance().getCurrentApp().getUniqueId();
+            if (!uiController.isAppSelectorVisible() || selectedAppId.equals(seatedAppId)) {
+                AuthInfo.ProvidedAuth credentials = ConnectIDManager.getCredentialsForApp(seatedAppId, uiController.getEnteredUsername());
+                if (credentials != null) {
+                    //uiController.setUsername(credentials.username);
+                    uiController.setPasswordOrPin(credentials.password);
+
+                    //NOTE: This would auto-login, but we decided not to
+                    //initiateLoginAttempt(uiController.isRestoreSessionChecked());
```
shubham1g5, line 458: should remove as no longer needed.

app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -392,13 +406,63 @@ public void dataPullCompleted() {
...
+                } else {
+                    uiController.setPasswordOrPin("");
```
shubham1g5, line 460: checking again on https://github.com/dimagi/commcare-android/pull/2660#discussion_r1256267233

app/src/org/commcare/utils/EncryptionUtils.java
shubham1g5: bumping

app/src/org/commcare/activities/connect/ConnectIDDatabaseHelper.java
```diff
@@ -0,0 +1,135 @@
...
+
+public class ConnectIDDatabaseHelper {
+    private static final Object connectDbHandleLock = new Object();
+    private static SQLiteDatabase connectDatabase;
+
+    private static byte[] getConnectDBPassphrase(Context context) {
+        try {
+            for (ConnectKeyRecord r : CommCareApplication.instance().getGlobalStorage(ConnectKeyRecord.class)) {
+                String encryptedPassphrase = r.getEncryptedPassphrase();
+                if (encryptedPassphrase != null) {
```
shubham1g5, line 30: Bumping https://github.com/dimagi/commcare-android/pull/2660#discussion_r1291530841

app/src/org/commcare/activities/connect/ConnectIDPhoneVerificationActivity.java
```diff
@@ -0,0 +1,316 @@
...
+                params.put("phone", username);
+                params.put("secret_key", password);
+            }
+            default -> {
+                urlId = R.string.ConnectValidatePhoneURL;
+                authInfo = new AuthInfo.ProvidedAuth(username, password, false);
+            }
+        }
+
+        boolean isBusy = !ConnectIDNetworkHelper.post(this, getString(urlId), authInfo, params, false, new ConnectIDNetworkHelper.INetworkResultHandler() {
```
shubham1g5, line 159: Bumping https://github.com/dimagi/commcare-android/pull/2660#discussion_r1294984079

app/src/org/commcare/CommCareApplication.java
```diff
@@ -1153,10 +1155,18 @@ public ModernHttpRequester buildHttpRequester(Context context, String url,
...
                                                   @Nullable HttpResponseProcessor responseProcessor, boolean retry) {
 
-        CommCareNetworkService networkService;
+        CommCareNetworkService networkService = null;
         if (authInfo instanceof AuthInfo.NoAuth) {
             networkService = CommCareNetworkServiceGenerator.createNoAuthCommCareNetworkService();
-        } else {
+        } else if(authInfo instanceof AuthInfo.CurrentAuth) {
+            //Try to get SSO token
+            AuthInfo.TokenAuth tokenAuth = ConnectIDSSOHelper.acquireSSOTokenSync(context);
```
shubham1g5, line 1163: should this only happen if connect ID is enabled ?

shubham1g5 (changes requested):

app/src/org/commcare/CommCareApplication.java
OrangeAndGreen: I added a check at the top of acquireSSOTokenSync() so it returns null if ConnectID is locked.

app/src/org/commcare/activities/connect/ConnectIDDatabaseHelper.java
OrangeAndGreen: Done

app/src/org/commcare/activities/LoginActivity.java
OrangeAndGreen: This code is used to clear the pre-populated password, i.e. if the user chooses to lock or forget ConnectID. I could handle that somewhere more explicitly so it's not part of this method, is that worth doing?

app/src/org/commcare/activities/LoginActivity.java
OrangeAndGreen: Done

app/src/org/commcare/activities/LoginActivity.java
OrangeAndGreen: Done

app/src/org/commcare/utils/EncryptionUtils.java
OrangeAndGreen: Excellent point, the DB would definitely break when a user upgrades from API <31 to >=31.

I changed the code so that the key retrieval and generation code (i.e. getKey()) now also returns the transformation string (via a new KeyAndTransform wrapper class). Now when generating the key, we can also create the correct transformation depending on the API level. When retrieving the key from the Keystore, we can tell which encryption was used during DB initialization depending on whether the retrieved key is a SecretKeyEntry (for AES) or PrivateKeyEntry (for RSA), and create the appropriate transformation string.

A caveat of this solution is that we'll continue using RSA encryption after the user upgrades their OS, but it seems ideally we would want to migrate from RSA to AES to take advantage of the increased security. Is this something we should explore further?

app/src/org/commcare/activities/connect/ConnectIDPhoneVerificationActivity.java
OrangeAndGreen: Done

app/src/org/commcare/utils/EncryptionUtils.java
```diff
@@ -171,25 +183,24 @@ public static byte[] generatePassphrase() {
         return result;
     }
 
-    private static String getTransformationString() {
+    public static String getTransformationString(boolean forceRSA) {
         String transformation;
-        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.M) {
-            transformation = String.format("%s/%s/%s", ALGORITHM, BLOCK_MODE, PADDING);
+        if (forceRSA || android.os.Build.VERSION.SDK_INT < android.os.Build.VERSION_CODES.M) {
```
shubham1g5, line 188: The android version check doesn't seem to be meaningful now to me. We can probably just decide the transformation based on `forceRSA` ?

app/src/org/commcare/activities/connect/ConnectIDNetworkHelper.java
```diff
@@ -192,43 +202,48 @@ private boolean getInternal(Context context, String url, AuthInfo authInfo, Mult
...
             public void processSuccess(int responseCode, InputStream responseData) {
                 isBusy = false;
+                dismissProgressDialog(context);
                 handler.processSuccess(responseCode, responseData);
             }
 
             @Override
             public void processClientError(int responseCode) {
                 isBusy = false;
+                dismissProgressDialog(context);
```
shubham1g5, line 223: can abstract this into another internal method something like - `onFinishProcessing()`

shubham1g5 (changes requested):
Very minor comments, looks quite close to merge. Can we resolve conflicts as well on this.

app/src/org/commcare/utils/EncryptionUtils.java
```diff
@@ -69,18 +69,29 @@ private static KeyStore getKeystore() throws KeyStoreException, CertificateExcep
         return keystoreSingleton;
     }
```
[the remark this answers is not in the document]
OrangeAndGreen: Done

app/src/org/commcare/utils/EncryptionUtils.java
OrangeAndGreen: Yup, you're right, forceRSA will always be true for API<M. But we get a compile warning about using the ALGORITHM, BLOCK_MODE, and PADDING constants if I remove that check, so I left it there to keep the compiler happy. Would it be better to remove it and ignore/suppress the warning?

app/src/org/commcare/activities/connect/ConnectIDNetworkHelper.java
OrangeAndGreen: Done

app/src/org/commcare/utils/EncryptionUtils.java
shubham1g5: > Would it be better to remove it and ignore/suppress the warning?

Yeah I would support that since we are already doing this check outside to determine `forceRSA`

OrangeAndGreen: Done, I removed the extra API check and suppressed the warning.

shubham1g5:
@OrangeAndGreen the changes looks good though the build seems to be failing on Jenkins.

app/src/org/commcare/CommCareApplication.java
```diff
@@ -251,6 +253,11 @@ public void onCreate() {
         GraphUtil.setLabelCharacterLimit(getResources().getInteger(R.integer.graph_label_char_limit));
 
         FirebaseMessagingUtil.verifyToken();
+
+        if(getEncryptionKeyProvider() == null) {
```
shubham1g5, line 257: Think we can remove the null check here. As a best practice, child classed should be expected to call `super.OnCreate()` as the first thing in the override method. They should be able to call `setEncryptionKeyProvider` after `super.OnCreate()` and override the defaults.

app/src/org/commcare/activities/connect/ConnectIDNetworkHelper.java
OrangeAndGreen: Done

app/src/org/commcare/CommCareApplication.java
OrangeAndGreen: Done

shubham1g5 (approved):

PR #2667, opened 2023-05-01
Referral Notifications - FCM Push Notifications

avazirna:
## Summary
This is the first PR related with the implementation of the [Referral Notification feature](https://docs.google.com/document/d/1JRKWF8Pe_LE1cr51UnTPaZPpT2gwgCV_ZAoZqe7jhV8). This part is responsible for the integration of CommCare with Firebase Cloud Message (FCM) as the channel to be used when sending [Push Notifications to CommCare apps](https://docs.google.com/document/d/1m-atLS7gcPRKWWdPd_pW62JLfrW122hAxCkIVurWeyQ). The Push Notifications are supposed to trigger syncs in CommCare, when certain conditions are met, allowing the user access to newly created referrals.

In order to test the service, follow the steps below:

1. Once CommCare is running, Firebase SDK will generate a FCM registration token, if there isn’t one yet, that will be shared with HQ via Heartbeat requests;
    1. For testing purposes, the token is currently outputted with the logs. Look for "New registration token was generated"
2. Once the FCM token is known, it’s possible to start sending Push notifications to the device using any REST Client, such as Postman or even cURL. 
    1. The first step is to obtain a valid access token with [Google OAuth playground](https://developers.google.com/oauthplayground/). 
        1. In Select & Authorize APIs, set the scope to https://www.googleapis.com/auth/firebase.messaging under _Firebase Cloud Messaging API v1_ and press _**Authorize APIs**_. This will require authentication with a valid email address.
        2. Once the authentication is complete and the authorization code is available, click _**Exchange authorization code for tokens**_ to get a refresh and access tokens. 
    2. Now that the access token is generated, submit a POST request to `https://fcm.googleapis.com/v1/projects/<FIREBASE_PROJECT_ID>/messages:send` with the request boby in the following format:
          ```
         {
            "message":{
               "token":<FCM_REGISTRATION_TOKEN>,
                "notification":{
                   "body": <MESSAGE_BODY>,
                   "title": <MESSAGE_TITLE>
                },
                "data":{
                   "action" : "sync",
                   "username" : <MOBILE_USERNAME>,
                   "domain" : <FULLY_QUALIFIED_DOMAIN>, 
                   "created_at": <SOME_DATE>
                }
             }
          }
          ```
     3. Monitor _logcat_ and look for "Message received ..." logs;

## Safety Assurance

cross-request: https://github.com/dimagi/commcare-core/pull/1275

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -110,7 +110,7 @@
         <!--TODO: We might want to find a better option to reference the notification channel Id-->
         <meta-data
             android:name="com.google.firebase.messaging.default_notification_channel_id"
-            android:value="notification-channel-server-communications" />
+            android:value="@string/fcm_default_notification_channel" />
 
         <activity
             android:label="@string/application_name"
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -107,6 +107,11 @@
         android:usesCleartextTraffic="true"
         android:theme="@style/AppBaseTheme">
 
+        <!--TODO: We might want to find a better option to reference the notification channel Id-->
+        <meta-data
+            android:name="com.google.firebase.messaging.default_notification_channel_id"
+            android:value="notification-channel-server-communications" />
+
         <activity
             android:label="@string/application_name"
             android:exported="true"
@@ -547,6 +552,14 @@
             android:exported="true"
             android:name="com.dimagi.android.zebraprinttool.PrintReceiverActivity"
             tools:node="merge"/>
+
+        <service
+            android:name="org.commcare.services.CommCareFirebaseMessagingService"
+            android:exported="true">
+            <intent-filter>
+                <action android:name="com.google.firebase.MESSAGING_EVENT" />
+            </intent-filter>
+        </service>
 
     </application>
 
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -107,7 +107,6 @@
         android:usesCleartextTraffic="true"
         android:theme="@style/AppBaseTheme">
 
-        <!--TODO: We might want to find a better option to reference the notification channel Id-->
         <meta-data
             android:name="com.google.firebase.messaging.default_notification_channel_id"
             android:value="@string/fcm_default_notification_channel" />
@@ -555,7 +554,7 @@
 
         <service
             android:name="org.commcare.services.CommCareFirebaseMessagingService"
-            android:exported="true">
+            android:exported="false">
             <intent-filter>
                 <action android:name="com.google.firebase.MESSAGING_EVENT" />
             </intent-filter>
--- a/app/src/org/commcare/heartbeat/HeartbeatRequester.java
+++ b/app/src/org/commcare/heartbeat/HeartbeatRequester.java
@@ -68,12 +68,13 @@ public class HeartbeatRequester extends GetAndParseActor {
         params.put(CC_VERSION, ReportingUtils.getCommCareVersionString());
         params.put(QUARANTINED_FORMS_PARAM, String.valueOf(StorageUtils.getNumQuarantinedForms()));
         params.put(UNSENT_FORMS_PARAM, String.valueOf(StorageUtils.getNumUnsentForms()));
-        params.put(LAST_SYNC_TIME_PARAM, getISO8601FormattedLastSyncTime());
+        params.put(LAST_SYNC_TIME_PARAM, convertTimeInMsToISO8601(SyncDetailCalculations.getLastSyncTime()));
         params.put(CURRENT_DRIFT, String.valueOf(DriftHelper.getCurrentDrift()));
         params.put(MAX_DRIFT_SINCE_LAST_HEARTBEAT, String.valueOf(DriftHelper.getMaxDriftSinceLastHeartbeat()));
         //TODO: Encode the FCM registration token
         params.put(FCM_TOKEN, FirebaseMessagingUtil.getFCMToken());
-        params.put(FCM_TOKEN_TIME, String.valueOf(FirebaseMessagingUtil.getFCMTokenTime()));
+        // TODO: Convert the Date to ISO 8601 format
+        params.put(FCM_TOKEN_TIME, convertTimeInMsToISO8601(FirebaseMessagingUtil.getFCMTokenTime()));
         return params;
     }
 
@@ -82,14 +83,14 @@ public class HeartbeatRequester extends GetAndParseActor {
         return new AuthInfo.CurrentAuth();
     }
 
-    private static String getISO8601FormattedLastSyncTime() {
-        long lastSyncTime = SyncDetailCalculations.getLastSyncTime();
-        if (lastSyncTime == 0) {
+    // TODO: Move this method to DateUtils
+    private static String convertTimeInMsToISO8601(long ms) {
+        if (ms == 0) {
             return "";
         } else {
             DateFormat df = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm'Z'");
             df.setTimeZone(TimeZone.getTimeZone("UTC"));
-            return df.format(lastSyncTime);
+            return df.format(ms);
         }
     }
 
--- a/app/src/org/commcare/heartbeat/HeartbeatRequester.java
+++ b/app/src/org/commcare/heartbeat/HeartbeatRequester.java
@@ -19,6 +19,7 @@ import org.commcare.utils.FirebaseMessagingUtil;
 import org.commcare.utils.SessionUnavailableException;
 import org.commcare.utils.StorageUtils;
 import org.commcare.utils.SyncDetailCalculations;
+import org.javarosa.core.model.utils.DateUtils;
 import org.javarosa.core.services.Logger;
 import org.json.JSONException;
 import org.json.JSONObject;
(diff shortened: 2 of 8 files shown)


shubham1g5 (changes requested):

app/AndroidManifest.xml
```diff
@@ -107,6 +107,11 @@
         android:usesCleartextTraffic="true"
         android:theme="@style/AppBaseTheme">
 
+        <!--TODO: We might want to find a better option to reference the notification channel Id-->
+        <meta-data
+            android:name="com.google.firebase.messaging.default_notification_channel_id"
+            android:value="notification-channel-server-communications" />
```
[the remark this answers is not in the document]
avazirna: Yes, I will move those to `strings.xml`

app/AndroidManifest.xml
```diff
@@ -548,6 +553,14 @@
             android:name="com.dimagi.android.zebraprinttool.PrintReceiverActivity"
             tools:node="merge"/>
 
+        <service
+            android:name="org.commcare.services.CommCareFirebaseMessagingService"
+            android:exported="true">
```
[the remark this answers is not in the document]
avazirna: Nop, just some residue from testing I was conducting. I will switch that off.

app/src/org/commcare/preferences/HiddenPreferences.java
```diff
@@ -551,4 +552,14 @@ public static void markRawMediaCleanUpComplete() {
         CommCareApplication.instance().getCurrentApp().getAppPreferences()
                 .edit().putBoolean(RAW_MEDIA_CLEANUP_COMPLETE, true).apply();
     }
+
+    public static boolean isPostLoginSyncNeeded() {
+        return CommCareApplication.instance().getCurrentApp().getAppPreferences()
+                .getBoolean(POST_LOGIN_SYNC_NEEDED, false);
+    }
+    public static void setPostLoginSyncNeeded(boolean syncNeeded) {
```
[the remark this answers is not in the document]
avazirna: @shubham1g5  This is a good point. I wonder if we should include some app info in the process to ensure that we trigger the sync for the right app.

avazirna:
@damagatchi retest this please?

app/src/org/commcare/preferences/HiddenPreferences.java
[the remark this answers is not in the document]
shubham1g5: Since cases are domain level and can be associated with any of the apps in the domain, I don't think it's possible to direct the sync notification to a specific app unless we build custom rules in HQ messaging engine to associate a particular rule with a specific app. But until we do that we should assume notifications to be non app-specific.

avazirna:
@damagatchi test this please

app/AndroidManifest.xml
```diff
@@ -107,6 +107,11 @@
         android:usesCleartextTraffic="true"
         android:theme="@style/AppBaseTheme">
 
+        <!--TODO: We might want to find a better option to reference the notification channel Id-->
```
shubham1g5, line 110: Is this todo still valid ?

app/AndroidManifest.xml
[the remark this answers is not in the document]
shubham1g5: Bumping in  latest review for visibility

app/src/org/commcare/services/CommCareFirebaseMessagingService.java
```diff
@@ -0,0 +1,148 @@
...
+    *    app is in the background, which means that the data object won't be processed from here
+    * 2) Verify if the message contains a data object and trigger the necessary steps according
+    *    to the action it carries
+    *
+    */
+    @Override
+    public void onMessageReceived(RemoteMessage remoteMessage) {
+        Logger.log(LogTypes.TYPE_FCM, "Message received: " + remoteMessage.getMessageId());
+        Map<String, String> payloadData = remoteMessage.getData();
+        RemoteMessage.Notification payloadNotification = remoteMessage.getNotification();
```
shubham1g5, line 49: Can a notification contains both `data` and `notification` object ?

app/src/org/commcare/utils/FirebaseMessagingUtil.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    public static void updateFCMToken(String newToken) {
+        SharedPreferences sharedPreferences = PreferenceManager
+                .getDefaultSharedPreferences(CommCareApplication.instance());
+        sharedPreferences.edit().putString(FCM_TOKEN, newToken).apply();
+        sharedPreferences.edit().putLong(FCM_TOKEN_TIME,System.currentTimeMillis()).apply();
+    }
+
+    public static String removeServerUrlFromUserDomain(String userDomain) {
```
shubham1g5, line 39: looks unused

app/src/org/commcare/services/CommCareFirebaseMessagingService.java
```diff
@@ -0,0 +1,148 @@
...
+
+        // Check if the message contains a data object, there is no further action if not
+        if (payloadData.size() == 0){
+            return;
+        }
+
+        FCMMessageData fcmMessageData = new FCMMessageData(payloadData);
+
+        switch(fcmMessageData.action){
+            case SYNC -> {} // trigger sync for fcmMessageData
```
shubham1g5, line 63: should call `setPendingSyncRequestFromServer` here ?

shubham1g5 (changes requested):

avazirna:

shubham1g5 (approved):
Looks safe to merge, we can continue any followups from here in other PRs.

app/src/org/commcare/utils/FirebaseMessagingUtil.java
```diff
@@ -46,4 +51,19 @@ public static String removeServerUrlFromUserDomain(String userDomain) {
         }
         return userDomain;
     }
+
+    public static void verifyToken() {
+        if(!BuildConfig.DEBUG) {
```
shubham1g5, line 56: Why not init token in debug ?

app/src/org/commcare/heartbeat/HeartbeatRequester.java
```diff
@@ -75,14 +83,14 @@ public AuthInfo getAuth() {
         return new AuthInfo.CurrentAuth();
     }
 
-    private static String getISO8601FormattedLastSyncTime() {
-        long lastSyncTime = SyncDetailCalculations.getLastSyncTime();
-        if (lastSyncTime == 0) {
+    // TODO: Move this method to DateUtils
```
shubham1g5, line 86: pending todo

app/src/org/commcare/services/CommCareFirebaseMessagingService.java
shubham1g5: bumping

app/src/org/commcare/utils/FirebaseMessagingUtil.java
avazirna: Nothing in particular, just trying to avoid noise but I'm happy to revert that.

shubham1g5: yeah we should revert to be able to have receive push notifications with debug apk as well.

app/src/org/commcare/services/CommCareFirebaseMessagingService.java
avazirna: > I don't completely understand. How do we handle the sync notifcation when the app is in background ? Do we just discard the notification ? Or it's routed thorough a different pathway ?

Yeah, this is a bit tricky. So, when the app is in the background and the message contains a `notification` object, Firebase SDK takes care of delivering the notification to the _Notification Drawer_. In case the message also contains a `data` object, Firebase SDK adds it to the extras of the intent of the launcher activity of the app. This is the default behaviour, so when the user taps the notification it opens the app launcher with the `data` object as an extra. 

> Also when onMessageReceived gets called (app is in foreground), It seems like we can only have one of notification or data payload. 

We can have both, but in this case because we want to trigger the sync when the app is in the background too, we are having messages with only the `data` object

avazirna:

avazirna:

avazirna:

shubham1g5:

app/src/org/commcare/services/CommCareFirebaseMessagingService.java
shubham1g5: > In case the message also contains a data object, Firebase SDK adds it to the extras of the intent of the launcher activity of the app. This is the default behaviour, so when the user taps the notification it opens the app launcher with the data object as an extra.

That means we should be handling these intents on app startup right ? Or they by default gets delivered to this messaging service when user clicks on notification and pull the app in foreground ?

shubham1g5:

app/src/org/commcare/services/CommCareFirebaseMessagingService.java
avazirna: @shubham1g5 This is supposed to be handled on app startup but it's not yet implemented because our current implementation on HQ doesn't trigger FCM messages with both `notification` and `data` objects (see image below). So, the intention is to incorporate that on a later stage depending on the feedback we get as well. 
image

shubham1g5: Should not we still need to handle only `data` messages that gets send when app is in background ? if not, How does the app handle sync data messages that gets delivered to the phone when the CC is in background ? For ex, my current understanding is -  

1. We trigger a notification with Action 'Background Sync' and type data messages
2. This notification gets delivered to phone when CC is in background. Therefore Firebase adds the notification to the notification drawer ? (Not sure if it's true for only data messages)
3. On clicking notification, Firebase launches CC with the intent containing data action. But since we are not handling intents with data payload, background sync doesn't get triggered. 

Does it sound right to you ?

avazirna: When the `message` doesn't contain a `notification` object and the app is in the background, the message is delivered to the FCM Service, so the `onMessageDelivered` is called. More about this can be found [here](https://firebase.google.com/docs/cloud-messaging/android/receive#handling_messages).

shubham1g5: got it, thanks for confirming.

shubham1g5 (approved):

PR #2668, opened 2023-05-01
Add hint text to QueryRequestUIController

Charl1996:
## Summary
[Ticket](https://dimagi-dev.atlassian.net/browse/SC-2367)

This PR is to align the webapp and Android app Case Claim UX by adding a hint button on Android which brings up a dialog when clicked, showing the hint/help text configured on HQ, similar to what is seen on the webapps.

When no help/hint text is configured, the hint icon is not shown.

## Feature Flag
None specific.

## Product Description
The image below shows what the user would have seen before this PR (as well as in the case that no hint text is configured),
 
without_hint_icon

When hint text is configured on HQ, the user will see the following button,

hint_icon

Now when the user clicks on the button, a dialog appears with the configured hint text, as seen below,

hint_popup

## Safety Assurance

### Automated test coverage
Change not covered in automated tests (don't think it's necessary?).

### Safety story
Did local testing with and without hint text configured.

### DIFF
--- a/app/unit-tests/src/org/commcare/android/tests/activities/QueryRequestActivityTest.java
+++ b/app/unit-tests/src/org/commcare/android/tests/activities/QueryRequestActivityTest.java
@@ -219,10 +219,10 @@ public class QueryRequestActivityTest {
         assertEquals("francisco", patientName.getText().toString());
 
         ImageButton imageButtonWithoutHint = promptsLayout.getChildAt(0).findViewById(R.id.prompt_hint_button);
-        assertEquals(8, imageButtonWithoutHint.getVisibility());
+        assertEquals(View.GONE, imageButtonWithoutHint.getVisibility());
 
         ImageButton imageButtonWithHint = promptsLayout.getChildAt(4).findViewById(R.id.prompt_hint_button);
-        assertEquals(0, imageButtonWithHint.getVisibility());
+        assertEquals(View.VISIBLE, imageButtonWithHint.getVisibility());
 
         Spinner stateSpinner = promptsLayout.getChildAt(2).findViewById(R.id.prompt_spinner);
         assertEquals(2, stateSpinner.getSelectedItemPosition());


app/src/org/commcare/activities/QueryRequestUiController.kt
```diff
@@ -407,7 +411,43 @@ class QueryRequestUiController(
...
     }
 
+    private fun setHintPlaceholder(promptView: View, queryPrompt: QueryPrompt) {
+        val hintText = getHintText(queryPrompt)
+        if (hintText != "") {
+            val promptHelpButton = promptView.findViewById<View>(R.id.prompt_hint_button)
+            promptHelpButton.visibility = View.VISIBLE
+
+            promptHelpButton.setOnClickListener {
+                AlertDialogWrapper.alertDialog(queryRequestActivity, "Hint", hintText)
```
shubham1g5, line 421: Curios what you think about removing the wrapper and directly calling the `StandardAlertDialog.getBasicAlertDialog()` method from here.

app/res/layout/query_prompt_layout.xml
```diff
@@ -18,6 +18,15 @@
         android:gravity="center|left"
         android:layout_height="wrap_content" />
 
+    <ImageButton
+        android:id="@+id/prompt_hint_button"
+        android:layout_width="0dp"
```
shubham1g5, line 23: curious what happens if we set the width here to `wrap_content` and remove `layout_weight` from below ? I want to avoid using `layout_weight` for help icon since it may result in unexpected behaviours for different screen sizes. We are setting `layout_weight` on other views to `1` since we want them to take all the available space but this is not true for help view.

app/res/layout/query_prompt_layout.xml
```diff
@@ -20,7 +20,7 @@
 
     <ImageButton
         android:id="@+id/prompt_hint_button"
-        android:layout_width="0dp"
+        android:layout_width="wrap_content"
         android:layout_height="wrap_content"
         android:layout_weight="0.5"
```
shubham1g5, line 25: we should be able to remove `layout_weight` if we are able to set the `layout_width` to `wrap_content`

shubham1g5:
Thanks @Charl1996 . This looks good to me. Although I think we need to do 2 other things to finalise this PR-
1. Merge `master` in the current branch to get the build passing on this PR
2. Add a test possibly as part of [this roboelectric test](https://github.com/dimagi/commcare-android/blob/master/app/unit-tests/src/org/commcare/android/tests/activities/QueryRequestActivityTest.java#L198) to test that `hint` is visible for one of the views when it's configured in the app. You will need to manually modify the test ccz to add the `hint` xml for one of the case search input field.

Charl1996:
> Thanks @Charl1996 . This looks good to me. Although I think we need to do 2 other things to finalise this PR-
> 
> 1. Merge `master` in the current branch to get the build passing on this PR
> 2. Add a test possibly as part of [this roboelectric test](https://github.com/dimagi/commcare-android/blob/master/app/unit-tests/src/org/commcare/android/tests/activities/QueryRequestActivityTest.java#L198) to test that `hint` is visible for one of the views when it's configured in the app. You will need to manually modify the test ccz to add the `hint` xml for one of the case search input field.

Thanks! Will do

app/src/org/commcare/activities/QueryRequestUiController.kt
```diff
@@ -407,6 +411,32 @@ class QueryRequestUiController(
...
+                ))
+            }
+        }
+    }
+
+    private fun getHintText(queryPrompt: QueryPrompt): String {
+        val displayData = queryPrompt.display.evaluate()
+        val hintText = displayData.hintText ?: return "";
+
+        return Localizer.processArguments(hintText, arrayOf("")).trim { it <= ' ' }
```
shubham1g5, line 437: I don't think we need to call `Localizer. processArguments` here as calling `queryPrompt.display.evaluate()` should evaluate any localised strings itself.

shubham1g5 (commented):
Looks great, just have one final comment that I didn't notice initially.

PR #2672, opened 2023-05-12
Testing - don't catch improper exceptions during view build

ctsims:
## Safety Assurance


ctsims:

PR #2678, opened 2023-06-28
Improve error handling for restore parse errors

shubham1g5:
## Summary
[Jira](https://dimagi-dev.atlassian.net/jira/servicedesk/projects/SUPPORT/queues/issue/SUPPORT-17011)

In case of bad restore data, we were not showing the actual exception as there is no locale key present corresponding to the particular error's action. 

I think it makes sense to pass the action as it is for some errors like this when there is no locale key present. 

Improved error response - 

Screenshot_1687971421

## Safety Assurance

### DIFF
--- a/app/src/org/commcare/views/notifications/NotificationMessageFactory.java
+++ b/app/src/org/commcare/views/notifications/NotificationMessageFactory.java
@@ -225,6 +225,8 @@ public class NotificationMessageFactory {
                 action = parameters[2] == null ? Localization.get(base + ".action") : Localization.get(base + ".action", new String[]{parameters[2]});
             } catch (Exception e) {
                 //No big deal, key doesn't need to exist
+                // pass in parameter value as it is
+                action = parameters[2];
             }
 
             NotificationActionButtonInfo buttonInfo = null;


shubham1g5:

ctsims:
@shubham1g5  - Hm., it should be possible to have this behavior occur by having the specific error types include an action message that includes only the variable locale key (IE: `${0}` or whatnot) right? 

I think its' often a fairly counter-productive anti-pattern for error messages to dump the stacktraces or technical errors outright, since they are rarely actionable for users and including the big dump of technical context can prevent people from seeing and communicating the clear part of the message. 

If a user sees 

```
= Couldn't restore your data = 
The app couldn't understand the data it got back from the server
```

They are likely to be able to communicate that clearly to someone, whereas if they see

```
= Couldn't restore your data = 
The app couldn't understand the data it got back from the server
Java$Exception$$$DivideByZeroError at RRRANDomCompil.CLASS$33:43322 
caught by
ElementParseHelper.getNewMessageBus:44:3222 ($FFa, %A_1)
```

They are like to assume the error is just technical jargon and not understand which parts need to be communicated and which don't. The example you cited is a pretty good error in this regard as nearly 50% of it is actionable, but that's still not a great density. 

I'm a little inclined to counter-propose having people reveal errors by updating the localization templates for their project if they want them to come through rather than making it the default. From my perspective the end-user is the wrong person to receive this error context anyway, and the ideal outcome would be for it to accessible through HQ to a supervisor, or to the support team so that someone who can action the error is seeing it. If we set an expectation that the way to get this context is to navigate to this screen and take a screenshot, we're going to make ourselves accountable for ensuring that this screen can contain the context in full, and isn't getting truncated, which might set up to put bad-work-after-good.

One potentially interesting option would be to enable submitting a specific error directly to HQ from this screen kind of how we enabled with the [old Report Button Flag](https://confluence.dimagi.com/display/GS/CommCare+Developer+Options+--+Internal#CommCareDeveloperOptionsInternal-HomeReportButtonenabled) which could include any extra context in this bucket when there's no action key that specifies how to translate it to the user, although I acknowledge that might be hard/impossible if the issues are impacting the user's ability to log in or use.

shubham1g5:
That makes sense in general. We can potentially change the implementation here to only show the exception details for the restore parse errors. 

The reason I thought It would be fine to put the exception details on this screen is because this screen is separated from the main user flow by a 'See Error Details' button. But I do agree that end user is not really the right audience for any of these errors. 

> One potentially interesting option would be to enable submitting a specific error directly to HQ from this screen kind of how we enabled with the [old Report Button Flag](https://confluence.dimagi.com/display/GS/CommCare+Developer+Options+--+Internal#CommCareDeveloperOptionsInternal-HomeReportButtonenabled) which could include any extra context in this bucket when there's no action key that specifies how to translate it to the user

I think we can make this button available from the error screen itself. I don't think the report issue mechanism should rely on user login strictly. Although I am afraid that this might result into a bunch of errors on Support end, that Support won't really know how to meaningfully followup on. 

Another potential mechanism here is probably Firebase Crashlytics/ Sumo,  but since user is not logged in it's hard to tag the error meaningfully to enable easy discovery of these errors/logs. Although we should probably be able to tag these with the fleeting `username` user tried to login with before hitting this error.  But Firebase and Sumo are also limited to few tech and support team members and doesn't help the delivery to troubleshoot these.

ctsims:
>That makes sense in general. We can potentially change the implementation here to only show the exception details for the restore parse errors.

Yeah, and my preference even if we did that would be to change the localization key default value for these errors and not to change the code. 

The screen is behind a notification, as you mention, but this is also the only way for users to get any information about these kinds of errors since they happen in the background, so I think they're still fundamentally user facing. 

> Although I am afraid that this might result into a bunch of errors on Support end, that Support won't really know how to meaningfully followup on.

Yeah, for more clarity, my preference would be that we submit the logs to HQ through a similar mechanism but that the reports wouldn't actually go to support directly but would show up in HQ somewhere for project space admins or another end user to then action (potentially by sending up the chain to support). That would just provide a clearer way for an issue to go from a device to a project admin to support in a way that has the right info for the right parties at each step.

shubham1g5:
Thanks, that makes sense although I am closing this out without taking any follow up actions (I don't think it will be a priority for any of the division to take that work)

PR #2681, opened 2023-07-18
Enable Certificate Transparency with OkHTTP Update

avazirna:
## Summary
This PR addresses a vulnerability flagged by the ABDM auditors, according to which an attacker is able to use open source tools to intercepts requests and perform MITM attacks. The initial recommendation was to adopt Certificate Pinning but after some investigation we opted for [Certificate Transparency (CT)](https://certificate.transparency.dev/), considering the restrictions and maintenance burden related with pinning certificates. Certificate Transparency verifies the validity of the host's SSL certificate by checking the existence of Signed Certificate Timestamps (SCTs). SCTs are basically a confirmation that the certificate was submitted to a CT log by the Certificate Authorities (CA) when a new certificate is issued. 

## Feature Flag
This feature is behind a CommCare custom property called `cc-enable-certificate-transparency`

## Product Description
When the verification fails, meaning there are no SCTs that accompany the Certificate, the connection is rejected and an exception such as the one below is raised.  
```
 www.commcarehq.org -> Failure: This certificate does not have any Signed Certificate Timestamps in it.
23:38:44.175  W  logger> warning-network: Detecting captive portal failed with exceptionCLEARTEXT communication to www.commcarehq.org not permitted by network security policy
23:38:44.186  W  javax.net.ssl.SSLPeerUnverifiedException: Certificate transparency failed
23:38:44.186  W  	at com.appmattus.certificatetransparency.internal.verifier.CertificateTransparencyInterceptor.intercept(CertificateTransparencyInterceptor.kt:78)
23:38:44.186  W  	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
23:38:44.186  W  	at org.commcare.core.network.CommCareNetworkServiceGenerator.lambda$static$0(CommCareNetworkServiceGenerator.java:48)
23:38:44.186  W  	at org.commcare.core.network.CommCareNetworkServiceGenerator$$ExternalSyntheticLambda0.intercept(Unknown Source:0)
```
Considering that this feature is supposed to be enabled at the CommCare App level, a new Advanced menu option called `Certificate Transparency` has been added to allow the users to disable the feature without having to update the app. 

## Safety Assurance

Cross-request: https://github.com/dimagi/commcare-core/pull/1311


avazirna:

app/src/org/commcare/network/CTInterceptorConfig.kt
```diff
@@ -0,0 +1,48 @@
...
+            return client.addNetworkInterceptor(getInterceptor)
+        }
+
+        // In case there are CT Interceptors already attached
+        removeCTInterceptors(client)
+        return client
+    }
+
+    private val getInterceptor = certificateTransparencyInterceptor {
+        logger = object : CTLogger {
```
ctsims, line 31: What's the frequency of this request pattern? The example lists a failed log, but it looks like at a minimum we would also expect getting a log record (or multiple) for each channel negotiation, which I don't think makes Logger the right channel unless we're filtering for failure more specifically

Based on the noisiness of logging I've been trying to push us back from using the CommCare logs for any events which are "unexceptional" and which would occur with higher frequency than user sessions. IE: If something normal / expected happens once per login it's fine to log, but if it happens on "machine frequency" (IE: a background process is retrying over and over) we should push to LogCat rather than CommCare's logs. It's also good to log blocking *failures* (Since eventually a user recovers from them), we just don't want to be constantly filling our sumo allocations with that data indefinitely.

app/src/org/commcare/network/CTInterceptorConfig.kt
```diff
@@ -0,0 +1,48 @@
...
+        if(HiddenPreferences.isCertificateTransparencyEnabled()) {
+            return client.addNetworkInterceptor(getInterceptor)
+        }
+
+        // In case there are CT Interceptors already attached
+        removeCTInterceptors(client)
+        return client
+    }
+
+    private val getInterceptor = certificateTransparencyInterceptor {
```
shubham1g5, line 30: maybe lazy init this to make sure we are not initialising the interceptor when the certificate transparency is disabled.

app/src/org/commcare/network/CTInterceptorConfig.kt
```diff
@@ -0,0 +1,48 @@
...
+class CTInterceptorConfig:HttpBuilderConfig  {
+
+
+    override fun performCustomConfig(client: OkHttpClient.Builder): OkHttpClient.Builder {
+        if(HiddenPreferences.isCertificateTransparencyEnabled()) {
+            return client.addNetworkInterceptor(getInterceptor)
+        }
+
+        // In case there are CT Interceptors already attached
+        removeCTInterceptors(client)
```
ctsims, line 26: I don't think this is possible, since the builder starts fresh each time in `customizeRetrofitSetup()`, so I'd prefer to avoid the redundancy, and especially the class name reflection check. Did you test specifically to see if things were getting added multiple times without this?

app/src/org/commcare/CommCareApplication.java
```diff
@@ -1183,4 +1182,8 @@ public AndroidPackageUtils getAndroidPackageUtils() {
     public boolean isNsdServicesEnabled() {
         return true;
     }
+
+    public void initCertificateTransparency() {
+        CommCareNetworkServiceGenerator.customizeRetrofitSetup(new CTInterceptorConfig());
```
ctsims, line 1187: Hm, are we sure this chains in this way? my read from [the current code](https://github.com/dimagi/commcare-core/blob/1e35c9faaa18743380a30abc7efded5533b941b2/src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java#L116) is that this may be [overwriting this config](https://github.com/dimagi/commcare-android/blob/8a1bf52325ee7198e122eb1e43a21900e33ed29a/app/src/org/commcare/CommCareApplication.java#L255C20-L255C34) rather than chaining it, and we may need a cleaner way to compose these two HttpBuilderConfig classes. It should be easy too, since they both act on the builder, we should just be able to make a composite that runs multiple.

ctsims (changes requested):
If I'm interpreting the current behavior correctly there are some changes we should make on these, but also satisfied with clarity in how it was determined that the behavior is different from what I'm interpreting.

shubham1g5 (changes requested):
+1 to Clayton's comments. 

> a new Developers option called Certificate Transparency has been added to allow the users to disable the feature without having to update the app.

I would not expect users to interact with Developers Option at all and if we expect users to be able to turn this off - should we have it somewhere in Advanced Settings sections instead ? But I am also curious why do we want the user to have the ability to turn this off if an app specfically enforces it for it's users.

app/src/org/commcare/network/CTInterceptorConfig.kt
avazirna: Yes, that's what I thought too but everything in the `CommCareNetworkServiceGenerator` is static and so we get the same `OkHttpClient.Builder` object with all the interceptors added previously. So this is just to remove the CT interceptors before the `build()` method is called. Unless I'm missing something.

app/src/org/commcare/network/CTInterceptorConfig.kt
avazirna: Completely, indeed this would create a lot of noise as it's supposed to log the result of every negotiation. I will change it to only log failures (only once), which will allow us to know when someone gets blocked. However, I don't think in that situation we would receive anything from the device.

ctsims: Cool, thanks for confirming. It's totally possible we could get the logged data from a request later if, for example, someone is trying to trick your WIFI with a MITM temporarily but you switch to mobile internet because it's not working.

app/src/org/commcare/network/CTInterceptorConfig.kt
ctsims: Ah, got it. 

That does seem tricky to deal with. I am a bit concerned about the structure of this approach being a straight string comparison, though. 

Could we move [this comparison](https://github.com/dimagi/commcare-android/pull/2681/files/0ac900e58e0b5d60b931e996deabf632b5d2cbde#diff-3ef7953a7a8b4016c33692a7f7330318e5e47b976fc5c2d05ad84efad0bd812bR39) to an `instanceof` check instead? That would prevent future drift issues that require a tight coupling of that string and the name

avazirna: We can't because the `CertificateTransparencyInterceptor` class is internal, but we can switch to an object comparison.

avazirna:

avazirna:

ctsims:
> This is a like a fail-safe mechanism, to be used if, for some reason, CT stops validating the certificate. I imagine that this can happen when the certificate expires, or it's renewed with a different CA and it takes some time to propagate, or even, due to a failure on CT infra side, but I'm just guessing

There are a few scenarios to consider, but the two most relevant are

1. Customer is using their own cert upfront on a server where CT doesn't work 
2. Customer is already on a server that will fail CT and can add the setting ahead of time after they catch the need with the announced pre-release
3. Google's Certificate Transparency service check disappears in the future

The first is the main one that requires a local option, since otherwise they wouldn't be able to download the app which would configure their device to not use it. I would mildly prefer to hide it behind developer options, simply because the need to use it is hypothetical and I don't want to introduce it unless there's a real demonstrated need or use case. If we do determine that for some reason CT works less well than we expect I could see moving it to a normal user setting instead.

avazirna:

avazirna:
> Details



> 

I will it back to `Developer Options`

avazirna:

avazirna:

app/src/org/commcare/CommCareApplication.java
avazirna: @ctsims I had the opportunity to retest this and confirm my understanding. So, this is not supposed to impact any of the existing configuration of the OkHttpClient object such as the [SSLSocketFactory](https://github.com/dimagi/commcare-android/blob/8a1bf52325ee7198e122eb1e43a21900e33ed29a/app/src/org/commcare/network/ISRGCertConfig.java#L74C17-L74C30), it just adds a new network interceptor and it sort of operates at a higher level. For instance, we can set a TrustManager for the Network Interceptor and it shouldn't override the TrustManager of the OkHttpClient. But I do like the idea of having a single entry point for these config classes.

app/build.gradle
```diff
@@ -308,6 +313,9 @@ android {
     }
 
     compileOptions {
+        // Flag to enable support for API desugaring
+        coreLibraryDesugaringEnabled true
```
shubham1g5, line 317: should be another PR as well ideally as it's not related to CT at all

app/src/org/commcare/CommCareApplication.java
```diff
@@ -220,7 +220,7 @@ public void onCreate() {
         // improperly, so the second https request in a short time period will flop)
         System.setProperty("http.keepAlive", "false");
 
-        attachISRGCert();
+        loadOkHttpBuilderCustomConfig();
```
shubham1g5, line 223: nit: rename to `customiseOkHttp()`

app/src/org/commcare/network/CTInterceptorConfig.kt
```diff
@@ -0,0 +1,62 @@
...
+            }
+        }
+        return interceptor!!
+    }
+
+    private fun removeCTInterceptors(client: OkHttpClient.Builder) {
+        client.networkInterceptors().removeAll { it === interceptor }
+    }
+
+    companion object {
```
shubham1g5, line 57: nit: companion object should be defined at the top of class

app/src/org/commcare/network/CTInterceptorConfig.kt
```diff
@@ -0,0 +1,62 @@
...
+    private fun getCTInterceptor(): Interceptor {
+        if (interceptor == null) {
+            interceptor = certificateTransparencyInterceptor {
+                logger = object : CTLogger {
+                    override fun log(host: String, result: VerificationResult) {
+                        if (result is VerificationResult.Failure && !previousRequestFailed) {
+                            Logger.log(LogTypes.TYPE_NETWORK, "$host -> $result")
+                            previousRequestFailed = true
+                        } else if (result is VerificationResult.Success && previousRequestFailed) {
+                            previousRequestFailed = false;
```
shubham1g5, line 43: nit: unnecessary semi colon

app/src/org/commcare/network/CTInterceptorConfig.kt
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    private fun removeCTInterceptors(client: OkHttpClient.Builder) {
+        client.networkInterceptors().removeAll { it === interceptor }
+    }
+
+    companion object {
+        private var interceptor: Interceptor? = null
+        private var previousRequestFailed = false
+        private var interceptorEnabled = false
```
shubham1g5, line 60: nit: rename to `interceptorAttached`

app/src/org/commcare/network/CTInterceptorConfig.kt
```diff
@@ -0,0 +1,62 @@
...
+            }
+        }
+        else {
+            // In case there are CT Interceptors already attached
+            if (interceptorEnabled) {
+                removeCTInterceptors(client)
+                interceptorEnabled = false;
+            }
+        }
+    }
```
shubham1g5, line 32: Can be simplified to - 
````
 fun toggleCertificateTransparency(client: OkHttpClient.Builder){
        if (HiddenPreferences.isCertificateTransparencyEnabled()){                
                client.addNetworkInterceptor(getCTInterceptor())            
        } else if (interceptor!=null) {        
                removeCTInterceptors(client)
                interceptor = null
        }
    }
````

File: app/src/org/commcare/network/CTInterceptorConfig.kt [hunk 11]
```diff
@@ -0,0 +1,62 @@
...
+import org.commcare.util.LogTypes
+import org.javarosa.core.services.Logger
+
+/**
+ * Adds and removes Certificate Transparency (CT) network interceptor
+ *
+ */
+class CTInterceptorConfig {
+
+    public fun toggleCertificateTransparency(client: OkHttpClient.Builder){
```
shubham1g5, line 18: no need for `public`

shubham1g5 (changes requested):

app/build.gradle
shubham1g5: I am actually going to split this out in another PR as we need it in `master` right away.

avazirna:

avazirna:

avazirna:

avazirna:

app/src/org/commcare/network/CTInterceptorConfig.kt
```diff
@@ -0,0 +1,59 @@
...
+        }
+    }
+
+    private fun getCTInterceptor(): Interceptor {
+        if (interceptor == null) {
+            interceptor = certificateTransparencyInterceptor {
+                logger = object : CTLogger {
+                    override fun log(host: String, result: VerificationResult) {
+                        if (result is VerificationResult.Failure && !previousRequestFailed) {
+                            Logger.log(LogTypes.TYPE_NETWORK, "$host -> $result")
```
shubham1g5, line 43: can we add an additional message to the log - ` Logger.log(LogTypes.TYPE_NETWORK, "Certification verification failed: $host -> $result")`

app/src/org/commcare/network/CTInterceptorConfig.kt
```diff
@@ -0,0 +1,59 @@
...
+                            previousRequestFailed = true
+                        } else if (result is VerificationResult.Success && previousRequestFailed) {
+                            previousRequestFailed = false
```
shubham1g5, line 46: This assignment can be moved out of if else block - `previousRequestFailed =  result is VerificationResult.Success`

app/src/org/commcare/network/OkHttpBuilderCustomConfig.java
```diff
@@ -0,0 +1,35 @@
...
+ * configs. As it stands, it serves to attach the ISRG Root X1 Certificate and enable/disable
+ * Certificate Transparency
+ */
+public class OkHttpBuilderCustomConfig implements HttpBuilderConfig {
+    private CTInterceptorConfig ctInterceptorConfig;
+    private ISRGCertConfig isrgCertConfig;
+
+    public OkHttpBuilderCustomConfig(){
+         ctInterceptorConfig = new CTInterceptorConfig();
+         isrgCertConfig = new ISRGCertConfig();
```
shubham1g5, line 20: Lets initialise them in ` performCustomConfig` itself. That way we can make sure they are only initialised if needed. For ex. we don't need to init  `ISRGCertConfig` if ` if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.N_MR1) {` is not true.

shubham1g5 (dismissed):
Some minor code feedback but looks good to go as it is.

app/src/org/commcare/network/OkHttpBuilderCustomConfig.java
avazirna: good point

shubham1g5: can you say more to what you are thinking ?

app/src/org/commcare/network/CTInterceptorConfig.kt
avazirna: I think the idea is to compare it against `VerificationResult.Failure`

app/src/org/commcare/network/OkHttpBuilderCustomConfig.java
avazirna: Considering also that all the `CommCareNetworkServiceGenerator` content is static we don't actually need instances of `CTInterceptorConfig` and `ISRGCertConfig` to apply these custom configurations, right?

avazirna: @shubham1g5 in concrete terms, [this](https://github.com/dimagi/commcare-android/pull/2681/commits/6a720db57319cf69d38f48e857bf5ac9693198a5) is what I'm thinking.

app/src/org/commcare/CommCareApplication.java
```diff
@@ -540,6 +535,10 @@ public void initializeAppResources(CommCareApp app) {
             FirebaseAnalyticsUtil.reportCorruptAppState();
         }
         app.setAppResourceState(resourceState);
+
+        // This is part of the CommCare app initialization because it needs to be applied during
+        // app initialization, update and when switching the seated app
+        customiseOkHttp();
```
avazirna, line 541: @shubham1g5 I thought I had included in the initial PR. Basically, because the Certificate Transparency is a CommCare app level configuration, we need to be able to apply it in case it changes between app updates and when the user switches the seated app.

avazirna:

shubham1g5 (changes requested):
Couple questions - 

1. How have you tested this change ? Can we specify it in the Safety story for PR 

2. This setting needs to be QA'ed as part of regular releases and therefore we should have a QA note with QA note label set.

app/src/org/commcare/CommCareApplication.java
avazirna: @shubham1g5 true, when CommCare is installed there is not CC app to initialize and the configuration won't applied. I think consolidating configurations of different cycles/types is probably a stretch, but we can continue applying the configuration during app startup, just to cover all the scenarios.

avazirna: [0e53d88](https://github.com/dimagi/commcare-android/pull/2681/commits/0e53d88234bfa72704c8f6fb9e075eedb1896730)

shubham1g5 (approved):

avazirna:

avazirna:

avazirna:

shubham1g5 (approved):

PR #2682, opened 2023-07-18
Enable secure flag

avazirna:
## Summary
This PR addresses an issue raised by the ABDM auditors in which sensitive data could be accessible to attackers in certain parts of the Android OS, such as in the Recent Apps. This PR enables the FLAG_SECURE flag, which enforces security features of the window, this includes blanking the window when the application goes to the background and preventing screenshots.

## Feature Flag
This feature is behind a CommCare custom property called `cc-enable-secure-flag`

## Product Description
When this flag is enabled, here's what the user sees when they see the app in the Recent apps. 
Sync attempt during Form Entry

## Safety Assurance

### DIFF
--- a/app/src/org/commcare/preferences/HiddenPreferences.java
+++ b/app/src/org/commcare/preferences/HiddenPreferences.java
@@ -554,7 +554,6 @@ public class HiddenPreferences {
     }
 
     public static boolean isFlagSecureEnabled() {
-        SharedPreferences properties = CommCareApplication.instance().getCurrentApp().getAppPreferences();
-        return PrefValues.YES.equals(properties.getString(ENABLE_SECURE_FLAG, PrefValues.NO));
+        return DeveloperPreferences.doesPropertyMatch(ENABLE_SECURE_FLAG, PrefValues.NO, PrefValues.YES);
     }
 }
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/preferences/HiddenPreferences.java
```diff
@@ -96,6 +96,7 @@ public class HiddenPreferences {
 
 
     private static final long NO_OF_HOURS_TO_WAIT_TO_RESUME_BACKGROUND_WORK = 36;
+    private static final String ENABLE_SECURE_FLAG = "cc-enable-secure-flag";
```
shubham1g5, line 99: Can we name it `cc-enable-secure-app-window` instead to imply it's usage better.

ctsims: +1 that if we aren't going to wrap the functionality behind a more generic preference of bundled features (or cross platform features) we should articulate its purpose as specifically as possible (IE: "cc-enable-android-window-secure-flag") to prevent people from misinterpreting it as more powerful or more important than it is in practice.

shubham1g5 (approved):
Looks good given build passing. 

 Also can we add the flag to [the wiki](https://confluence.dimagi.com/display/GS/CommCare+Developer+Options+--+Internal) as well. 
 
 Thanks.

PR #2684, opened 2023-07-25
Set audio recorder sample rate

avazirna:
## Summary
Some users have reported having trouble recording audio, CommCare simple crashes with the following exception being thrown:
```
Fatal Exception: java.lang.IllegalStateException:
       at android.media.MediaRecorder.start(MediaRecorder.java)
       at org.commcare.views.widgets.RecordingFragment.startRecording(RecordingFragment.java:168)
       at org.commcare.views.widgets.RecordingFragment.lambda$prepareButtons$3(RecordingFragment.java:135)
       at org.commcare.views.widgets.RecordingFragment.$r8$lambda$_f4gpgcFmBpYAjMTw-zPR52rFnE(RecordingFragment.java)
       at org.commcare.views.widgets.RecordingFragment$$InternalSyntheticLambda$0$cc1961f5746e23f7491a3d6e1f5a06d308627bd6d301c3274e25a58d59953a04$1.onClick(RecordingFragment.java:2)
       at android.view.View.performClick(View.java:7522)
       at android.view.View.performClickInternal(View.java:7499)
       at android.view.View.access$3700(View.java:835)
       at android.view.View$PerformClick.run(View.java:29039)
       at android.os.Handler.handleCallback(Handler.java:938)
       at android.os.Looper.loopOnce(Looper.java:201)
       at android.os.Looper.loop(Looper.java:288)
       at android.app.ActivityThread.main(ActivityThread.java:7952)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1022)
```
Initial findings led us to consider that maybe the devices stopped supporting the HE-AAC codec, but after some more investigation it was possible to identify another cause for `IllegalStateException` exceptions as not setting the Audio Sample Rate for the recorder. This PR changes that.
 

## Safety Assurance

cross-request: https://github.com/dimagi/commcare-core/pull/1154

### DIFF
--- a/app/src/org/commcare/views/widgets/RecordingFragment.java
+++ b/app/src/org/commcare/views/widgets/RecordingFragment.java
@@ -52,6 +52,9 @@ public class RecordingFragment extends DialogFragment {
 
     private static final String MIMETYPE_AUDIO_AAC = "audio/mp4a-latm";
 
+    private static final int HEAAC_SAMPLE_RATE = 44100;
+    private static final int AMRNB_SAMPLE_RATE = 8000;
+
     private String fileName;
     private static final String FILE_EXT = ".mp3";
 
@@ -192,10 +195,13 @@ public class RecordingFragment extends DialogFragment {
             recorder = new MediaRecorder();
         }
 
+        boolean isHeAacSupported = isHeAacEncoderSupported();
+
         recorder.setAudioSource(MediaRecorder.AudioSource.MIC);
+        recorder.setAudioSamplingRate(isHeAacSupported ? HEAAC_SAMPLE_RATE : AMRNB_SAMPLE_RATE);
         recorder.setOutputFormat(MediaRecorder.OutputFormat.MPEG_4);
         recorder.setOutputFile(fileName);
-        if (isHeAacEncoderSupported()) {
+        if (isHeAacSupported) {
             recorder.setAudioEncoder(MediaRecorder.AudioEncoder.HE_AAC);
         } else {
             recorder.setAudioEncoder(MediaRecorder.AudioEncoder.AMR_NB);


shubham1g5:
@avazirna Based on PR description, It's not clear to me how we established the correlation between `IllegalStateException` and the audio sample rate. Can you speak more about it. Thanks!

avazirna:
> @avazirna Based on PR description, It's not clear to me how we established the correlation between `IllegalStateException` and the audio sample rate. Can you speak more about it. Thanks!

@shubham1g5 The documentation states that `setAudioSamplingRate` should be called before `prepare()` and [this](https://medium.com/@anuandriesz/android-audio-recording-guide-mediarecorder-72fb814165c7) page also suggests that the configuration of the MediaRecorder should be set in a specific order and that failing to do so can result in `IllegalStateExceptions` being thrown. But I have to stress that while this fixed the issue on my Android emulator, I still need confirmation from the field.

shubham1g5:
> The documentation states that setAudioSamplingRate should be called before prepare() 

Can you link the doc you are referring to. I looked at the sample code [here](https://developer.android.com/guide/topics/media/platform/mediarecorder#sample-code) but didn't find a mention of it. 

> But I have to stress that while this fixed the issue on my Android emulator, I still need confirmation from the field.

Got it, Yeah I would wait to see if it fixes the issue for on field user before merging this.

app/src/org/commcare/views/widgets/RecordingFragment.java
```diff
@@ -193,6 +195,7 @@ private void setupRecorder() {
         }
 
         recorder.setAudioSource(MediaRecorder.AudioSource.MIC);
+        recorder.setAudioSamplingRate(RECORDING_SAMPLE_RATE);
```
shubham1g5, line 198: Is the audio sampling rate same for `HE_AAC` and `AMR_NB` ?

avazirna:
> > The documentation states that setAudioSamplingRate should be called before prepare()
> 
> Can you link the doc you are referring to. I looked at the sample code [here](https://developer.android.com/guide/topics/media/platform/mediarecorder#sample-code) but didn't find a mention of it.

Yes, I missed the link, [here](https://developer.android.com/reference/android/media/MediaRecorder#setAudioSamplingRate(int)) is where it says that the method should be called before `prepare()`
> 
> > But I have to stress that while this fixed the issue on my Android emulator, I still need confirmation from the field.
> 
> Got it, Yeah I would wait to see if it fixes the issue for on field user before merging this.

app/src/org/commcare/views/widgets/RecordingFragment.java
avazirna: Nop, [here](https://developer.android.com/guide/topics/media/platform/supported-formats#audio-formats) it says that `AMR_NB` should be sampled at 8kHz. In that scenario the audio will be downsampled to the appropriate rate before compression.

shubham1g5: Is it correct then to set the sampling rate to `44100` for both HE_AAC and AMR_NB ?

shubham1g5:
> Yes, I missed the link, [here](https://developer.android.com/reference/android/media/MediaRecorder#setAudioSamplingRate(int)) is where it says that the method should be called before prepare()

I think it just means that if we call the `setAudioSamplingRate` , we must call it before `prepare()` . But it should not mean we need to necessarily call `setAudioSamplingRate` before `prepare()`

avazirna:
> > Yes, I missed the link, [here](https://developer.android.com/reference/android/media/MediaRecorder#setAudioSamplingRate(int)) is where it says that the method should be called before prepare()
> 
> I think it just means that if we call the `setAudioSamplingRate` , we must call it before `prepare()` . But it should not mean we need to necessarily call `setAudioSamplingRate` before `prepare()`

True, but it's also an indication that it's called when prepping for the recording. And in the aforementioned documentation it says that the instantiation of MediaRecorder objects should follow:
image

But let's err on the side of caution, we can set Sampling Rate according to the supported codecs.

app/src/org/commcare/views/widgets/RecordingFragment.java
avazirna: @shubham1g5 I have seen in some places, but as suggested let's set the value according to the codec.

avazirna:

avazirna:

avazirna:

ctsims (approved):
Thanks for the iteration, looks good to me!

PR #2686, opened 2023-08-02
Java 8+ API desugaring support

shubham1g5:
## Summary

Enable desugaring to support Java 8 APIs like stream() for older devices (Android API 21 - 23) 

Reference: https://developer.android.com/studio/write/java8-support#library-desugaring

## Safety Assurance

### Safety story

Build level change to add support for Java 8+ APIs

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -133,6 +133,7 @@ dependencies {
 
     implementation 'com.google.android.material:material:1.3.0'
     implementation 'com.google.guava:guava:31.1-jre'
+    coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:2.0.3'
 }
 
 ext {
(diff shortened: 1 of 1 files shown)


shubham1g5:
Causing tests to fail with -
````
java.lang.NoSuchMethodError: No static method toMap(Ljava/util/function/Function;Ljava/util/function/Function;)Lj$/util/stream/Collector; in class Lj$/util/stream/Collectors; or its super classes (declaration of 'j$.util.stream.Collectors' appears in /data/app/org.commcare.dalvik.test-rb9q8Vi83H-CwA9Zib9YlA==/base.apk:classes2.dex)
````

I inspected the release apk and `java.util.stream.Collectors` seems to be missing. I am not sure why. Going to leave it as it is for now and change the code to not use `Stream` Apis.

PR #2690, opened 2023-08-08
Safe starts session with error handling while handling activity result

shubham1g5:
## Summary

[Jira](https://dimagi-dev.atlassian.net/browse/QA-5457)

When a fixture instance used in case search prompts is not present, mobile crashes with no error handling. 

## Feature Flag

CASE_SEARCH

## Safety Assurance

### Automated test coverage

NA

### Safety story
 
Only error handling


avazirna (approved):
Looks safer :-)

PR #2691, opened 2023-08-10
Trims location attributes before formatting as a string

shubham1g5:
## Summary

[JIRA](https://dimagi-dev.atlassian.net/browse/SAAS-14313)
[Related HQ PR](https://github.com/dimagi/commcare-hq/pull/33338)

@mjriley  noticed that some of the location strings has leading spaces into the location string like - ` lat    lng alt acc`. This PR trims the location attributes to make sure we keep the formatting consistent on mobile side. 

## Safety Assurance

### Automated test coverage

None, but covered in manual QA tests

### Safety story

Very minor change, will go through mobile release QA

### DIFF
--- a/app/src/org/commcare/utils/GeoUtils.java
+++ b/app/src/org/commcare/utils/GeoUtils.java
@@ -47,7 +47,11 @@ public class GeoUtils {
      * @return String in format "<latitude> <longitude> <altitude> <accuracy>"
      */
     public static String locationToString(Location location) {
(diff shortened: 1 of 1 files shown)


ctsims (changes requested):
I don't think this bug is occurring as outlined in the source ticket, and I don't think that the `Location` outputs here are capable of returning incorrectly formatted strings. I've updated the original ticket with the findings, happy to revisit if I missed something.

shubham1g5:
Ahh thanks for verifying, I missed that we now allow editing on forms. Closing out.

PR #2701, opened 2023-08-29
Tech5 Integration - Enrollment updates

avazirna:
## Summary
This PR is part of an ongoing initiative to integrate CommCare with Tech5 SDK. The goal of the current changes is to be able to capture biometric templates and send back to CommCare in a Base64 encoded String as part of a app callout question response. 

## Safety Assurance


commcare-support-library/src/main/java/org/commcare/commcaresupportlibrary/BiometricUtils.java
```diff
@@ -0,0 +1,68 @@
...
+import com.google.gson.Gson;
+import com.google.gson.JsonSyntaxException;
+import com.google.gson.reflect.TypeToken;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class BiometricUtils {
+
+    public enum BiometricIdentifier {
```
shubham1g5, line 14: potentially worth to put into a separate file of it's own.

commcare-support-library/src/main/java/org/commcare/commcaresupportlibrary/BiometricUtils.java
```diff
@@ -0,0 +1,68 @@
...
+import android.util.Base64;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonSyntaxException;
+import com.google.gson.reflect.TypeToken;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class BiometricUtils {
```
shubham1g5, line 12: code formatting is off in the class.

commcare-support-library/src/main/java/org/commcare/commcaresupportlibrary/BiometricUtils.java
```diff
@@ -0,0 +1,68 @@
...
+     * @param templates Map containing biometric templates
+     * @return Base64 encoded string
+     */
+    public static String convertMapTemplatesToBase64String(Map<BiometricUtils.BiometricIdentifier,
+            byte[]> templates){
+        // In order to reduce the size of the templates, we are converting each byte array into a
+        // Base64 encoded String
+        Map<Integer, String> templatesBase64Encoded = new HashMap<>(templates.size());
+        for (Map.Entry<BiometricUtils.BiometricIdentifier, byte[]> template : templates.entrySet()) {
+            templatesBase64Encoded.put(template.getKey().ordinal(),
```
shubham1g5, line 32: any reason we are double encoding templates, here and then later again at the end of this method ?

shubham1g5 (changes requested):
PR looks good to me though I think lack completion - 
1. We should update the [API docs](https://github.com/dimagi/commcare-android/blob/master/commcare-support-library/src/main/java/org/commcare/commcaresupportlibrary/identity/identity_integration.md)
2. We need to write tests for new API in [here](https://github.com/dimagi/commcare-android/blob/master/app/unit-tests/src/org/commcare/android/tests/formentry/IdentityCalloutTests.kt)

commcare-support-library/src/main/java/org/commcare/commcaresupportlibrary/BiometricUtils.java
avazirna: This is because `JSON` is used to represent the Map object, and considering that the values are byte arrays, the first encoding avoids representing them in JSON, which would increment the size due to the comas used to split the elements.

avazirna:

avazirna:

avazirna:

commcare-support-library/src/main/java/org/commcare/commcaresupportlibrary/BiometricIdentifier.java
```diff
@@ -0,0 +1,7 @@
+package org.commcare.commcaresupportlibrary;
```
shubham1g5, line 1: can we make sure that all biometric classes are in `org.commcare.commcaresupportlibrary.identity` package.

commcare-support-library/src/main/java/org/commcare/commcaresupportlibrary/BiometricIdentifier.java
```diff
@@ -0,0 +1,7 @@
...
+
+public enum BiometricIdentifier {
```
shubham1g5, line 3: I don't think we wanna allow ` UNKNOWN` or  `INVALID` here.

commcare-support-library/src/main/java/org/commcare/commcaresupportlibrary/BiometricUtils.java
shubham1g5: I am struggling to understand how will this gets used in the form. With current design, Seems like we are going to store into form the whole `template` string instead of storing the individual templates in individula field like Simprint. Any reason we are deviating away from Simprints pattern ?

commcare-support-library/src/main/java/org/commcare/commcaresupportlibrary/BiometricUtils.java
```diff
@@ -0,0 +1,64 @@
...
+        return Base64.encodeToString(templatesInJson.getBytes(), Base64.DEFAULT);
+    }
+
+    /**
+     * This method converts the Base64 encoded biometric templates to its original form
+     *
+     * @param base64EncodedTemplates String containing Base64 encoded biometric templates
+     * @return Map containing biometric templates
+     */
+    public static Map<BiometricIdentifier, byte[]> convertBase64StringTemplatesToMap(
```
shubham1g5, line 41: where will this be used ?

app/unit-tests/src/org/commcare/android/tests/formentry/IdentityCalloutTests.kt
```diff
@@ -96,6 +98,24 @@ class IdentityCalloutTests {
         assertEquals(guidToConfidenceMap.elementAt(2), "★★★")
     }
 
+    @Test
+    fun testRegistrationWithTemplates() {
+        val formEntryActivity = ActivityLaunchUtils.launchFormEntry("m0-f1")
+
+        var templates : HashMap<BiometricIdentifier, ByteArray> =
```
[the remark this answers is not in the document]
avazirna: yeah, in the principle of separation of concerns, but we can move it to a global var

commcare-support-library/src/main/java/org/commcare/commcaresupportlibrary/BiometricUtils.java
avazirna: In the biometric app, once the biometric templates are retrieved from the CC database, this will them be used to revert them into the original form.

commcare-support-library/src/main/java/org/commcare/commcaresupportlibrary/identity/identity_integration.md
```diff
@@ -20,6 +20,11 @@ Often you will need this generated guid to be passed back to CommCare so that it
 IdentityResponseBuilder.registrationResponse(guid)
     .finalizeResponse(activity)
 ````
+Alternatively, in case the biometric templates are to be stored in CommCare, use the following instead -
```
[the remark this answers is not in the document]
avazirna: Absolutely, I missed that one

commcare-support-library/src/main/java/org/commcare/commcaresupportlibrary/BiometricUtils.java
avazirna: @shubham1g5 the principle here is to use CC just to store the biometric templates and in the most simple way, so in a case property. any biometric operations will be preceded by the retrieval and restoration of the templates to their original format.

shubham1g5: I meant why not store individiual templates in separate fields in form like [here](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/provider/SimprintsCalloutProcessing.java#L95) ?

app/unit-tests/src/org/commcare/android/tests/formentry/IdentityCalloutTests.kt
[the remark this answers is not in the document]
shubham1g5: you can just pass the templates as a method param instead of making it a global var

commcare-support-library/src/main/java/org/commcare/commcaresupportlibrary/BiometricUtils.java
avazirna: It's more to avoid having to handle different form questions and case properties, when there is no apparent advantage to doing that. From my understanding, we are not doing anything with the biometric templates in CC. But I think that for the sake of standartization we could adopt the same pattern, but I wonder if you see any other gains from this.

app/unit-tests/src/org/commcare/android/tests/formentry/IdentityCalloutTests.kt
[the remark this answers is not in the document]
avazirna: That's a better option

commcare-support-library/src/main/java/org/commcare/commcaresupportlibrary/BiometricUtils.java
shubham1g5: If we store it together and an app is capturing multiple templates, how will they separate this data out in future without writing a script ? I think storing them seprately makes any future data management on those templates much simpler for project teams.

commcare-support-library/src/main/java/org/commcare/commcaresupportlibrary/identity/identity_integration.md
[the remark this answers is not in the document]
avazirna: c49a74e

commcare-support-library/src/main/java/org/commcare/commcaresupportlibrary/BiometricIdentifier.java
avazirna: 73d6a98c861ffa373fb7195728375b0ab06bd018

commcare-support-library/src/main/java/org/commcare/commcaresupportlibrary/BiometricUtils.java
avazirna: c19496bced827842183f3daeca51dc425d0ec157

app/unit-tests/src/org/commcare/android/tests/formentry/IdentityCalloutTests.kt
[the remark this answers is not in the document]
avazirna: 9d5df752257e2318ede4b31116c36e8d83ed800c

avazirna:

avazirna:

avazirna:

app/unit-tests/src/org/commcare/android/tests/formentry/IdentityCalloutTests.kt
```diff
@@ -96,6 +99,25 @@ class IdentityCalloutTests {
         assertEquals(guidToConfidenceMap.elementAt(2), "★★★")
     }
 
+    @Test
+    fun testRegistrationWithTemplates() {
```
shubham1g5, line 103: Can we also verify that converting back base 64 encoded templates to original templates using the support lib method indeed matches with original templates as part of this test.

commcare-support-library/src/main/java/org/commcare/commcaresupportlibrary/identity/identity_integration.md
```diff
@@ -20,6 +20,12 @@ Often you will need this generated guid to be passed back to CommCare so that it
 IdentityResponseBuilder.registrationResponse(guid)
     .finalizeResponse(activity)
 ````
+Alternatively, in case the biometric templates are to be stored in CommCare, use the following instead -
+````
+IdentityResponseBuilder.registrationResponse(guid, templates)
+    .finalizeResponse(activity)
+````
+* `templates` is a [`Map`](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html) containing all the biometric templates and whose _keys_ are [`BiometricIdentifier`](BiometricIdentifier.java) elements and _values_ are the actual biometric templates in the form of a byte array
```
shubham1g5, line 28: We should add a sample code below this to give an example of how to construct templates using Biometric indentifiers.

shubham1g5 (changes requested):

app/src/org/commcare/provider/IdentityCalloutHandler.java
```diff
@@ -212,4 +236,11 @@ public static boolean isRegistrationResponse(Intent intent) {
     public static boolean isVerificationResponse(Intent intent) {
         return intent.hasExtra(IdentityResponseBuilder.VERIFICATION);
     }
+
+    public static void main(String[] args) {
```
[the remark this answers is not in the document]
avazirna: Shoot, just residual code from some testing I was conducting, I could swear that I removed everything by somehow some was left behind. My bad.

app/src/org/commcare/provider/IdentityCalloutHandler.java
```diff
@@ -109,8 +110,31 @@ private static boolean processRegistrationReponse(FormDef formDef,
...
+            numOfTemplates++;
+        }
+
+        String result;
+        if (registrationResult.getTemplates().isEmpty()) {
+            result = guid;
+        } else {
+            result = Localization.get("intent.callout.biometrics.capture.result",
+                    new String[]{String.valueOf(numOfTemplatesStored), String.valueOf(numOfTemplates)});
+        }
```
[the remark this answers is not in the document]
avazirna: My reasoning here is to provide some feedback to the user in terms of the number of biometrics that was successfully captured (and stored). It can be that we scan 4 fingers but the callout handler is only able to store 2 due to a misconfiguration in the app. But not a hard requirement, I'm happy to drop this and maybe wait for feedback.

app/unit-tests/resources/commcare-apps/identity_callouts/modules-0/forms-0.xml
```diff
@@ -1,58 +1,91 @@
 <h:html xmlns:h="http://www.w3.org/1999/xhtml" xmlns:orx="http://openrosa.org/jr/xforms" xmlns="http://www.w3.org/2002/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:jr="http://openrosa.org/javarosa" xmlns:vellum="http://commcarehq.org/xforms/vellum">
-	<h:head>
```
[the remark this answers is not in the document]
avazirna: it is confusing, do we have a practice around this? I just made the modification in HQ, downloaded the app and replaced the files.

commcare-support-library/src/main/java/org/commcare/commcaresupportlibrary/identity/BiometricUtils.java
```diff
@@ -0,0 +1,60 @@
...
+     * @return Base64 encoded string
+     */
+    public static String convertTemplateToBase64String(BiometricIdentifier biometricIdentifier,
+                                                       byte[] template) {
+        // In order to reduce the size of the outpput, the byte array is converted to Base64 before
+        // the instantiation of the pair object
+        Pair<Integer, String> templatePairBase64Encoded = new Pair<>(biometricIdentifier.ordinal(),
+                Base64.encodeToString(template, Base64.DEFAULT));
+        String templatesInJson = new Gson().toJson(templatePairBase64Encoded);
+        return Base64.encodeToString(templatesInJson.getBytes(), Base64.DEFAULT);
```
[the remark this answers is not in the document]
avazirna: Here's the idea is to bundle the `BiometricIdentifier` with the `template`. It is redundant as we use the `callout response key` to identify the type of biometric but I think it could be useful to have a second layer of identification. For instance, the app builder can make a poor choice of case property name and have difficulties later to know if the template is of a `finger` or a `face`. We could also use the position `0` of the array for this. Any thoughts?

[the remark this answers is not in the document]
shubham1g5: > For instance, the app builder can make a poor choice of case property name and have difficulties later to know if the template is of a finger or a face

I don't think its a valid concern. App builders can name the properties of other fields poorly too (First Name to Last Name for ex) and we should not take into account such mistakes to make decisions at platform level. I would rather rgo for simplicity here and just store the template directly in the case property.

app/unit-tests/resources/commcare-apps/identity_callouts/modules-0/forms-0.xml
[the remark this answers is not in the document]
shubham1g5: We should only change the parts that changed. I generally edit the files by hand manually to add the new/modified parts.

app/src/org/commcare/provider/IdentityCalloutHandler.java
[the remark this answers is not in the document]
shubham1g5: Got it. That makes sense to me although I am worried about supporting 2 different return formats to the callout question.  I am thinking about case exports where some fields might have guid and some fields this string which might be confusing to interpret for the data systems incorporated by a project.

That said I do like the idea of telling to user whether everything was saved correctly or not. How about we just say `All data stored successfully` or `Failed to store some data`  to indicate failure vs success in both templates and non templates use cases here.

app/unit-tests/resources/commcare-apps/identity_callouts/modules-0/forms-0.xml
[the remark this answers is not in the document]
avazirna: I undid this and followed your advice - e81b21b8afcc47968eb929efc08e236ee995582f

app/unit-tests/src/org/commcare/android/util/TestUtils.java
```diff
@@ -362,12 +363,19 @@ private static EvaluationContext buildContextWithInstances(UserSandbox sandbox,
...
         errorMsg = ExprEvalUtils.expectedEval(expr, instance, null, expectedValue, null);
         assertTrue(errorMsg, "".equals(errorMsg));
     }
+
+    public static Object getFormValue(String expr) throws XPathSyntaxException {
+        FormDef formDef = FormEntryActivity.mFormController.getFormEntryController().getModel().getForm();
+        FormInstance instance = formDef.getMainInstance();
+        EvaluationContext ec = new EvaluationContext(instance);
+        return ExprEvalUtils.xpathEval(ec, expr);
+    }
```
avazirna, line 380: @shubham1g5 I have created this method to retrieve a value from a form question, but I wonder if there isn't a method to do this already.

app/src/org/commcare/provider/IdentityCalloutHandler.java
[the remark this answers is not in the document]
avazirna: 756586fa289c69c28791b4d7b083813d3d5a5daf

commcare-support-library/build.gradle
```diff
@@ -57,6 +57,8 @@ dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
     implementation 'com.android.support:appcompat-v7:26.1.0'
     implementation 'com.android.support.constraint:constraint-layout:1.0.2'
+    implementation 'com.google.code.gson:gson:2.9.0'
```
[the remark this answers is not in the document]
avazirna: good catch, it's not needed.

shubham1g5 (approved):

shubham1g5 (approved):

PR #2702, opened 2023-09-01
Fix barcode screen glitch

avazirna:
## Summary
This PR fixes an issue with the Case List screen in which a barcode search view is shown momentarily when the case list is being rendered (see image below).
image

## Safety Assurance

### DIFF
--- a/app/res/layout/entity_select_layout.xml
+++ b/app/res/layout/entity_select_layout.xml
@@ -111,7 +111,8 @@
         android:id="@+id/searchfooter"
         android:orientation="horizontal"
         android:background="@drawable/border_top_black"
-        android:paddingTop="1dp">
+        android:paddingTop="1dp"
+        android:visibility="gone">
 
         <TextView
             android:id="@+id/screen_entity_select_search_label"


ctsims:
@avazirna I think the code looks good here, but a few points on the PR:

1- This PR is really two meaningful changes (or one change for two big reasons). One is eliminating legacy code that no longer has an execution path, the other is fixing a bug that happens to get fixed when that code is removed.

I'd make sure that both of those intents are communicated here. If someone git blames back to this PR, it would be super confusing for them to see that the reason that 80 lines of code and a bunch of views were removed was to fix a bug where a view temporarily flickered and disappeared. 

2 - You have the safety story checked but don't provide an explanation on why it's safe. I think that context does exist from our chats (IE: It's clear looking at Shubham's PR that there used to be a meaningful conditional where a bunch of code wasn't activated, and now there isn't), so it's good to include that background info here so a reviewer can see that this code couldn't possibly be used. Also good to confirm that you've tested the new view layouts on the two screens correctly and confirmed that they behave right still after eliminating the view (and needing to re-anchor on a new view)

avazirna:
> @avazirna I think the code looks good here, but a few points on the PR:
> 
> 1- This PR is really two meaningful changes (or one change for two big reasons). One is eliminating legacy code that no longer has an execution path, the other is fixing a bug that happens to get fixed when that code is removed.
> 
> I'd make sure that both of those intents are communicated here. If someone git blames back to this PR, it would be super confusing for them to see that the reason that 80 lines of code and a bunch of views were removed was to fix a bug where a view temporarily flickered and disappeared.
> 
> 2 - You have the safety story checked but don't provide an explanation on why it's safe. I think that context does exist from our chats (IE: It's clear looking at Shubham's PR that there used to be a meaningful conditional where a bunch of code wasn't activated, and now there isn't), so it's good to include that background info here so a reviewer can see that this code couldn't possibly be used. Also good to confirm that you've tested the new view layouts on the two screens correctly and confirmed that they behave right still after eliminating the view (and needing to re-anchor on a new view)

@ctsims thanks for the notes, I will update the PR to make sure this removal is well documented. As per the testing, it seems to be working as expected in newer versions of Android, but I still need to test on older versions.

avazirna:

avazirna:

PR #2705, opened 2023-09-18
Adds wait in Async Restore Instrumentation Tests

shubham1g5:
Adds wait after making HQ API calls to make sure HQ registers those changes well before we do login. 

Only a test change and therefore skipping the PR template.


shubham1g5:
Closing, I think the real issue is that HQ has become better at generating restore in time and is not resturning 202 response code anymore due to that. I am going to try exploading cases for the users instead.

PR #2709, opened 2023-09-21
Android 13 Tests Updates

avazirna:
## Summary
This PR brings a few changes needed to our Instrumentation Tests to successfully run on Android 13.

## Safety Assurance


shubham1g5 (approved):
Looks good, can we try adding a Android 13 device on Browserstack to see how well our tests run on Browserstack for an Android 13 device. It would be great to add more variety across Android OS versions in BS test runs.

avazirna:
> Looks good, can we try adding a Android 13 device on Browserstack to see how well our tests run on Browserstack for an Android 13 device. It would be great to add more variety across Android OS versions in BS test runs.

Yeah, I was looking at the supported devices yesterday and they have many options.

PR #2710, opened 2023-09-22
Intent for launching a specific App In CommCare 

shubham1g5:
## Summary
Adds a way to launch a particular CC app using an intent as - 

````
val intent = Intent("org.commcare.dalvik.action.CommCareSession")
intent.putExtra(DispatchActivity.SESSION_ENDPOINT_APP_ID, appId)
startActivity(intent)
````

commit by commit

## Safety Assurance

### Automated test coverage

coverage for new functionality is in PR, installs and logins have ample tests. 

### Safety story

Tests should cover for any regressions on the dipatch and login pages.

### DIFF
--- a/app/src/org/commcare/activities/DispatchActivity.java
+++ b/app/src/org/commcare/activities/DispatchActivity.java
@@ -207,7 +207,7 @@ public class DispatchActivity extends AppCompatActivity {
         if (sesssionEndpointAppID != null) {
             CommCareApp currentApp = CommCareApplication.instance().getCurrentApp();
             if (currentApp != null) {
-                return !currentApp.getUniqueId().contentEquals(sesssionEndpointAppID);
+                return !currentApp.getUniqueId().equals(sesssionEndpointAppID);
             }
         }
         return false;
--- a/app/src/org/commcare/activities/DispatchActivity.java
+++ b/app/src/org/commcare/activities/DispatchActivity.java
@@ -1,5 +1,7 @@
 package org.commcare.activities;
 
+import static org.commcare.commcaresupportlibrary.CommCareLauncher.SESSION_ENDPOINT_APP_ID;
+
 import android.content.Intent;
 import android.os.Bundle;
 import android.util.Log;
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: app/src/org/commcare/activities/DispatchActivity.java [hunk 1]
```diff
@@ -195,6 +202,17 @@ private void dispatch() {
         }
     }
 
+    private boolean needAnotherAppLogin() {
+        String sesssionEndpointAppID = getSessionEndpointAppId();
+        if (sesssionEndpointAppID != null) {
+            CommCareApp currentApp = CommCareApplication.instance().getCurrentApp();
+            if (currentApp != null) {
+                return !currentApp.getUniqueId().contentEquals(sesssionEndpointAppID);
```
OrangeAndGreen, line 210: Curious about the use of contentEquals() here instead of just equals(). From what I understand contentEquals() just calls equals() under the hood when the passed input is a String, so functionally there should be no difference. So this may just be a consistency nit.

But in LoginActivity line 561 I see we're only calling equals(). When I search for "getUniqueId." I see 4 times that we use equals() and one that we use contentEquals(). Okay to ignore this if you don't think it's worth changing.

app/src/org/commcare/activities/LoginActivityUIController.java
```diff
@@ -230,6 +228,23 @@ public void refreshView() {
         }
     }
 
+    @Nullable
+    private ApplicationRecord getPresetAppRecord(ArrayList<ApplicationRecord> readyApps) {
+        String presetAppId = activity.getPresetAppID();
+        if (presetAppId != null) {
+            for (ApplicationRecord readyApp : readyApps) {
+                if (readyApp.getUniqueId().contentEquals(presetAppId)) {
```
OrangeAndGreen, line 236: contentEquals() again, see above comment.

OrangeAndGreen (approved):
One or two minor comments but otherwise this looks great!

OrangeAndGreen (approved):

PR #2711, opened 2023-09-26
Improved biometric enrollment for Android versions before API 30

OrangeAndGreen:
## Summary
Added checks for the API level when the user chooses to enroll in a biometric (fingerprint or PIN). When the API is less than 30, there's one additional check to see if it is >=28 and if so the fingerprint enrollment can still be launched. Otherwise, a message is shown to the user indicating that they will need to configure the biometric manually in Android settings, and a button is shown that will take them directly to the Security page in Android settings.

## Product Description
For users with API 28 or 29, a new fingerprint enrollment will now appear. Otherwise, any enrollment with API <30 now exhibits the behavior described above.

New enrollment when user chooses to enroll fingerprint using API 28 or 29:
image

New message shown when app can't auto-direct user all the way to Android enrollment:
image

### DIFF
--- a/app/src/org/commcare/activities/connect/ConnectIdManager.java
+++ b/app/src/org/commcare/activities/connect/ConnectIdManager.java
@@ -365,10 +365,9 @@ public class ConnectIdManager {
                     boolean configured = intent.getBooleanExtra(ConnectIdConstants.CONFIGURED, false);
                     manager.passwordOnlyWorkflow = intent.getBooleanExtra(ConnectIdConstants.PASSWORD, false);
                     boolean failedEnrollment = intent.getBooleanExtra(ConnectIdConstants.ENROLL_FAIL, false);
-                    if(failedEnrollment) {
+                    if (failedEnrollment) {
                         nextRequestCode = ConnectIdTask.CONNECT_BIOMETRIC_ENROLL_FAIL;
-                    }
-                    else {
+                    } else {
                         nextRequestCode = configured && !manager.passwordOnlyWorkflow ?
                                 ConnectIdTask.CONNECT_REGISTRATION_UNLOCK_BIOMETRIC :
                                 ConnectIdTask.CONNECT_REGISTRATION_VERIFY_PRIMARY_PHONE;
@@ -544,7 +543,7 @@ public class ConnectIdManager {
             }
             case CONNECT_BIOMETRIC_ENROLL_FAIL -> {
                 nextRequestCode = ConnectIdTask.CONNECT_REGISTRATION_CONFIGURE_BIOMETRICS;
-                if(success) {
+                if (success) {
                     //Go to settings
                     launchSecuritySettings = true;
                 }
@@ -562,7 +561,7 @@ public class ConnectIdManager {
 
         manager.continueWorkflow();
 
-        if(launchSecuritySettings) {
+        if (launchSecuritySettings) {
             //Launch after continuing workflow so previous activity is still there when user returns
             SettingsHelper.launchSecuritySettings(manager.parentActivity);
         }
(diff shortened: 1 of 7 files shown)


app/src/org/commcare/activities/connect/ConnectIdManager.java
```diff
@@ -529,6 +542,13 @@ public static void handleFinishedActivity(int requestCode, int resultCode, Inten
                     }
                 }
             }
+            case CONNECT_BIOMETRIC_ENROLL_FAIL -> {
+                nextRequestCode = ConnectIdTask.CONNECT_REGISTRATION_CONFIGURE_BIOMETRICS;
+                if(success) {
+                    //Go to settings
+                    launchSecuritySettings = true;
```
shubham1g5, line 549: why not fire the settings screen directly from here ?

app/src/org/commcare/utils/BiometricsHelper.java
```diff
@@ -151,11 +151,24 @@ private static int canAuthenticate(Context context, BiometricManager biometricMa
...
+            //Best case, handles both fingerprint and PIN
+            enrollIntent = new Intent(Settings.ACTION_BIOMETRIC_ENROLL);
+            enrollIntent.putExtra(Settings.EXTRA_BIOMETRIC_AUTHENTICATORS_ALLOWED,
+                    authenticator);
+        }
+        else if(authenticator == StrongBiometric && Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+            //An alternative for fingerprint enroll that might be available
+            enrollIntent = new Intent(Settings.ACTION_FINGERPRINT_ENROLL);
+        } else {
+            //No way to enroll, have to fail
```
shubham1g5, line 167: curious why did you decide to have a separate status enroll failed instead of directing user to settings screen from here ?

shubham1g5 (changes requested):
Code looks good to me. Although requesting 2 minor changes for completion - 
1. Can we format the new code properly
2. Can we add a screenshot in PR to be able to have a better idea how the error message and go to settings button present itself on UI.

app/src/org/commcare/utils/BiometricsHelper.java
OrangeAndGreen: Since in this case we're going to dump them on the Security settings screen and expect them to navigate manually from there (i.e. open Screen Lock setting, the configure desired biometrics), I'm navigating back up to where we can display the formal message screen so we can tell them what's about to happen and let them click the button to go on from there. Launching the ConnectIdMessageActivity from here seemed like unnecessarily breaking the pattern.

app/src/org/commcare/activities/connect/ConnectIdManager.java
OrangeAndGreen: The call to continueWorkflow() launches the next Connect activity (in this case it re-launches ConnectIdVerificationActivity after showing the message activity). I still want that activity to launch again, but then I want the settings activity to launch on top of it therefore I need it to be launched second.

OrangeAndGreen:
Code formatting done in [489c14a](https://github.com/dimagi/commcare-android/pull/2711/commits/489c14a2953782e4d08f2ab599c45e68e22d1400)

app/src/org/commcare/activities/connect/ConnectIdManager.java
shubham1g5: The navigation here seems a bit complex to me. I would think a more simplified workflow here would be for Verification activity to show the error and settings screen button directly instead of us going back and forth with verification and message activity.

shubham1g5 (approved):
Looks Ok to merge.

PR #2712, opened 2023-09-26
Remove AdMob

avazirna:
## Summary
With Consumer Apps being now obsolete, there is no use for AdMob in CommCare and considering that Google Play is forcing us to update the firebase-ads dependency, which would in turn require updating the Consumer Apps scripts, a decision was made to completely remove the AdMobManager class (and all its references) and the dependency.

## Safety Assurance


avazirna:

shubham1g5 (approved):

PR #2719, opened 2023-10-04
Connect App verification

shubham1g5:
## Summary

Verifies Connect App after installing it and move to next sceeen only if verification is successful

## Safety Assurance

### Automated test coverage

None

### Safety story

Changes only touch connect screens, CC impact is limited to verification activity which adds a variable `fromConnect` to handle launch from connect but looks like a minimal safe change.

### DIFF
--- a/app/src/org/commcare/activities/connect/ConnectActivity.java
+++ b/app/src/org/commcare/activities/connect/ConnectActivity.java
@@ -25,7 +25,7 @@ public class ConnectActivity extends CommCareActivity<ResourceEngineListener> {
                 if (result.getResultCode() == Activity.RESULT_OK) {
                     ConnectDownloadingFragment connectDownloadFragment = getConnectDownloadFragment();
                     if (connectDownloadFragment != null) {
-                        connectDownloadFragment.startLearning();
+                        connectDownloadFragment.onSuccessfulVerification();
                     }
                 }
             });
--- a/app/src/org/commcare/activities/connect/ConnectActivity.java
+++ b/app/src/org/commcare/activities/connect/ConnectActivity.java
@@ -1,11 +1,16 @@
 package org.commcare.activities.connect;
 
+import android.app.Activity;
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/activities/connect/ConnectActivity.java
```diff
@@ -14,6 +19,17 @@
 
 public class ConnectActivity extends CommCareActivity<ResourceEngineListener> {
 
+    ActivityResultLauncher<Intent> verificationLauncher = registerForActivityResult(
+            new ActivityResultContracts.StartActivityForResult(),
+            result -> {
+                if (result.getResultCode() == Activity.RESULT_OK) {
+                    ConnectDownloadingFragment connectDownloadFragment = getConnectDownloadFragment();
+                    if (connectDownloadFragment != null) {
+                        connectDownloadFragment.startLearning();
```
OrangeAndGreen, line 28: This appears to be hard-coded for the learning app, but we'll need the functionality for both learn and delivery apps. There's an "isLearning" boolean now being passed into ConnectDownloadingFragment, maybe just a matter of accessing it here to either call startLearning or startDelivery?

OrangeAndGreen (changes requested):
Change requested to support both learn and deliver apps, see comment below.

app/src/org/commcare/activities/connect/ConnectActivity.java
shubham1g5: `startLearning` does do the learn or deliver correctly but I have [renamed the method ](https://github.com/dimagi/commcare-android/pull/2719/commits/5f405dbde9b3e3832b9e6cfdedab138c1f6e1b70) to avoid the confusion

OrangeAndGreen: Ah, that makes sense but thanks for making it more clear.

OrangeAndGreen (approved):

PR #2724, opened 2023-10-11
Not showing apps managed by Connect in LoginActivity selector

OrangeAndGreen:
This is to address [CCPC-141](https://dimagi-dev.atlassian.net/browse/CCPC-141). Hiding apps that are managed by Connect in the selector on the login page, so there's no way for the user to access those apps from there.

"Apps managed by Connect" are identified as having a stored ConnectLinkedAppRecord with a username equal to the ConnectID username (a long random alphanumeric string).

The code allows the presetAppId to remain in the list (i.e. when launched from Connect), since it needs to be included in order for the traditional login functionality to work.

I also renamed the presetAppID and getter to Id since the linter doesn't like acronyms to be in all caps.

### DIFF
--- a/app/src/org/commcare/activities/LoginActivityUiController.java
+++ b/app/src/org/commcare/activities/LoginActivityUiController.java
@@ -20,6 +20,7 @@ import android.widget.TextView;
 import androidx.preference.PreferenceManager;
 
 import java.util.ArrayList;
+import java.util.Locale;
 import java.util.Vector;
 
 import org.commcare.CommCareApplication;
@@ -28,6 +29,7 @@ import org.commcare.activities.connect.ConnectDatabaseHelper;
 import org.commcare.activities.connect.ConnectManager;
 import org.commcare.android.database.app.models.UserKeyRecord;
 import org.commcare.android.database.global.models.ApplicationRecord;
+import org.commcare.core.network.AuthInfo;
 import org.commcare.dalvik.R;
 import org.commcare.google.services.analytics.FirebaseAnalyticsUtil;
 import org.commcare.interfaces.CommCareActivityUIController;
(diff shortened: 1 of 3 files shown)


shubham1g5:
Why don't we want user to access Connect Apps directly from login ?

OrangeAndGreen:
I think there are several reasons this change was requested:
1) To create separation between Connect-managed apps and user-managed apps.

2) The user experience is odd since the auto-populated username is the ConnectID username (a long alphanumeric string that they won't recognize).

3) The selector gets cluttered up by a bunch of apps, when really the only reason for the user to use those inputs is for their user-managed apps. So we might as well show them a smaller list.

4) If they don't have any user-managed apps, they won't see the login inputs at all

app/src/org/commcare/activities/LoginActivityUiController.java
```diff
@@ -221,9 +223,28 @@ public void refreshView() {
...
         // Decide whether or not to show the app selection spinner based upon # of usable apps
         ArrayList<ApplicationRecord> readyApps = MultipleAppsUtil.getUsableAppRecords();
+        if(ConnectManager.isConnectIdIntroduced()) {
+            //We need to remove any apps that are managed by Connect
+            String username = ConnectManager.getUser(activity).getUserId().toLowerCase(Locale.getDefault());
+            for(int i= readyApps.size()-1; i>=0; i--) {
+                String appId = readyApps.get(i).getUniqueId();
+                //Preset app needs to be in the list if set
+                if(!appId.equals(activity.getPresetAppId())) {
+                    AuthInfo.ProvidedAuth auth = ConnectManager.getCredentialsForApp(appId, username);
```
shubham1g5, line 233: abstract as ConnectManager.isConnectEnabledForApp()

app/src/org/commcare/activities/LoginActivityUiController.java
```diff
@@ -221,9 +223,28 @@ public void refreshView() {
 
         // Decide whether or not to show the app selection spinner based upon # of usable apps
         ArrayList<ApplicationRecord> readyApps = MultipleAppsUtil.getUsableAppRecords();
+        if(ConnectManager.isConnectIdIntroduced()) {
```
shubham1g5, line 226: should abstract into another method - filterConnectApps

app/src/org/commcare/activities/connect/ConnectManager.java
```diff
@@ -588,12 +588,10 @@ public static void forgetAppCredentials(String appId, String userId) {
     }
 
     public static AuthInfo.ProvidedAuth getCredentialsForApp(String appId, String userId) {
```
shubham1g5, line 590: add `@Nullable` to signify possible null return

shubham1g5 (changes requested):

app/src/org/commcare/activities/connect/ConnectManager.java
OrangeAndGreen: [49aef9b](https://github.com/dimagi/commcare-android/pull/2724/commits/49aef9b1630835182867aa080277d058dbd245c5)

app/src/org/commcare/activities/LoginActivityUiController.java
OrangeAndGreen: [49aef9b](https://github.com/dimagi/commcare-android/pull/2724/commits/49aef9b1630835182867aa080277d058dbd245c5)

app/src/org/commcare/activities/LoginActivityUiController.java
OrangeAndGreen: I left this out for now after abstracting the filterConnectManagedApps function. This would be a one-liner to check that the retrieved auth is not null, but I would still want to keep the other code out (i.e. retrieving Connect username) for simplicity.

OrangeAndGreen:
I need to merge this PR in order to get a new build for QA before tomorrow.

PR #2725, opened 2023-10-12
Going directly to Connect jobs list after user unlocks CID

OrangeAndGreen:
Once the user chooses to unlock ConnectID on the login page, we'll go straight to the jobs list instead of back to the login page. This is a user request.


OrangeAndGreen:
I need to merge this PR in order to get a new build for QA before tomorrow.

PR #2730, opened 2023-11-01
Added cccStaging build flavor

OrangeAndGreen:
New cccStaging build flavor in order to support Connect production and staging versions.


app/build.gradle
```diff
@@ -333,6 +336,10 @@ android {
             manifest.srcFile 'CommcareAndroidManifest.xml'
         }
 
+        cccStaging {
+            manifest.srcFile 'CommcareAndroidManifest.xml'
```
shubham1g5, line 340: is this required ?

shubham1g5 (approved):

app/build.gradle
OrangeAndGreen: Hmm... good question! I assumed so since it's specified instead of the default for both commcare and lts, but looking into CommcareAndroidManifest.xml it seems like maybe it's just for adding some extra shortcuts?

OrangeAndGreen: I can try removing it and see if things work, although I'm not sure what error to look for other than just making sure the app installs and launches successfully.

shubham1g5:

shubham1g5:
Merging to unblock other PRs

PR #2731, opened 2023-11-20
Adds a new notification channel for push notifications

shubham1g5:
## Summary

Simple change to direct firebase notifications to a new notification channel with default priority (instead of non obstructive min priority for server notifications) 

## Safety Assurance

### Safety story

Small change, almost no-op. Should be tested by QA before release.

### DIFF
--- a/app/res/values/strings.xml
+++ b/app/res/values/strings.xml
@@ -393,6 +393,8 @@
     <string name="notification_channel_user_session_description" cc:translatable="true">Notification containing information around current user session like when the current session is scheduled to expire</string>
     <string name="notification_channel_server_communication_title" cc:translatable="true">Server Communications</string>
     <string name="notification_channel_server_communication_description" cc:translatable="true">Notifications for status of any background server communication happening at the moment like CommCare App updates, log submission etc.</string>
+    <string name="notification_channel_push_notfications_title" cc:translatable="true">Priority notifications from Server</string>
+    <string name="notification_channel_push_notfications_description" cc:translatable="true">Notifications sent from server directly to notify user of priority tasks</string>
 
     <!--Recovery Measures-->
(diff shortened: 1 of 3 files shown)


avazirna (approved):

ctsims:
@shubham1g5 - If this is for CommCare Connect, I think we should be getting ahead of the curve here on creating a specific category for these notifications that we hold the Connect team accountable for using reliably.

Especially if these are going to be related to Connect Opportunities, I think it's pretty key that users receiving notifications about new ops are able to reasonably identify and opt-out of those notifications for us to be in compliance with the rules. They can still be high priority or whatnot (if opted in) but it needs to not be possible to mix these notifications with those from another channel, and 'default' sounds like the wrong framing for Connect market notifications, which are pretty narrowly focused.

shubham1g5:
@ctsims Unfortunately firebase doesn't have a way for us to meaningfully categorise the notification based on it's content. As such we can only create one bucket for all firebase push notifications connect or non-connect. 

We will have to have a internal mechanism for connect users instead if we want them to have a way of opting out of these notifications based on which server don't send notifications to a set of users.

calellowitz (approved):

calellowitz:
> I think it's pretty key that users receiving notifications about new ops are able to reasonably identify and opt-out of those notifications for us to be in compliance with the rules.

New opportunity notifications are likely to be a small subset of the notifications connect sends (unless the user never joins an opp, otherwise they will also receive ones related to active opps). I am not sure if that changes the calculus, but I agree with shubham that providing a way in app to opt out of new opportunity notifications, including the sms, will likely be the best way to do that.

shubham1g5:
merging this for now (as there is no other way to handle this as notification channels) but happy to continue the conversation around custom opt-outs.

PR #2732, opened 2023-11-29
Moves Async entity Implementation to Core

shubham1g5:
## Summary

https://dimagi-dev.atlassian.net/browse/USH-3790

Moves Async Entity [to core](https://github.com/dimagi/commcare-core/pull/1386) in order to be able to implement Cache and Indexing feature on Web Apps. 

## Feature Flag

Cache and Index

## Safety Assurance

### Automated test coverage

No substantial coverage 

### Safety story

- Behind feature flag
- Refactor is straightforward without making any sort of functional changes 
- Will go through QA

QA Note: Test Cache and Index functionality. 

review commit by commit

cross-request: https://github.com/dimagi/commcare-core/pull/1387


shubham1g5:
bumping this for a review.

app/src/org/commcare/models/AsyncEntity.java
```diff
@@ -128,24 +124,14 @@ public String getNormalizedField(int i) {
...
-        try {
-            db = CommCareApplication.instance().getUserDbHandle();
-        } catch (SessionUnavailableException e) {
-            return null;
-        }
-
-        //get the db lock
-        db.beginTransaction();
-        try {
+        if (EntityStorageCache.lockCache()) {
```
snopoke: What about making this implement `AutoCloseable` and then use `try-with-resource`?

Or perhaps `lockCache` returns an `AutoClosable` object and raises and exception if the transaction can't be started.

app/src/org/commcare/models/database/user/models/CommCareEntityStorageCache.java
```diff
@@ -163,4 +195,90 @@ public static int getEntityCacheWipedPref(String uuid) {
...
+        if (SqlStorage.STORAGE_OUTPUT_DEBUG) {
+            Log.d(TAG, "Sequential Cache Load: " + (System.currentTimeMillis() - now) + "ms");
+        }
+    }
+
+    public String getCacheKey(String detailId, String mFieldId) {
+        return detailId + "_" + mFieldId;
+    }
+
+    private static String buildValidKeys(Vector<Integer> sortKeys, DetailField[] fields) {
```
avazirna, line 239: `buildValidKeys`, `buildKeyNameWhereClause` and `populateEntitySet` are private static methods called by non-static methods, do they need to be `static`?

avazirna (commented):
@shubham1g5 looks good, I just left a minor comment.

app/src/org/commcare/models/database/user/models/CommCareEntityStorageCache.java
```diff
@@ -72,18 +73,12 @@ public static void createIndexes(SQLiteDatabase db) {
...
-            db = CommCareApplication.instance().getUserDbHandle();
-        } catch (SessionUnavailableException e) {
-            return false;
-        }
-
+        SQLiteDatabase db = CommCareApplication.instance().getUserDbHandle();
         //get the db lock
         db.beginTransaction();
-        return true;
+        return db;
```
snopoke, line 81: Doing this will result in the DB connection being closed as well (instead of just ending the transaction). Is that what we want?

app/src/org/commcare/models/database/user/models/CommCareEntityStorageCache.java
shubham1g5: I prefer to make methods `private static` when they can be to signify that none of these methods changes the member variables of the class.

app/src/org/commcare/models/database/user/models/CommCareEntityStorageCache.java
shubham1g5: Ahh right! I don't think we want to `close` the db. Should we just return a dummy closeable in that case ?

snopoke: Yes, we can create a new class that will just end the transaction in the close method.

snopoke (approved):

shubham1g5:

PR #2735, opened 2023-12-07
Update android_translatable_strings.txt

orangejenny:
Replacement for https://github.com/dimagi/commcare-translations/pull/80


shubham1g5 (approved):

orangejenny:
I'm going to merge without fixing the lint errors, though I do appreciate the reminder that it's "Wi-Fi" not "Wi-fi".

@shubham1g5 This will add this string to the default set of translations when users bulk download CCUI translations, right?

PR #2741, opened 2024-02-08
Set app callout response for biometric identification

avazirna:
## Summary
This is a small improvement to the biometric identification workflow using the app callout widget. Currently there is no indication of a successful identification (see image below) which makes it hard for the user to realize the outcome unless they move to the next question, which is when there is usually a text with the result. This PR sets the response of the identity callout to be the score of the biometric match.
Screenshot_20240207-181240_CommCare

## Product Description
The new behaviour sets the response as:
Screenshot_20240207-181902_CommCare

## Safety Assurance


zandre-eng (approved):
Nice!

shubham1g5 (approved):

shubham1g5:
Should this be merged or closed ?

PR #2744, opened 2024-03-01
Allow Unsynced Time Limit to be set in hours

avazirna:
## Summary
Currently, the minimum frequency for CommCare to warn users that a sync is needed is daily, this PR allows setting time limits shorter than that. The property `Unsynced Time Limit` is expressed in days, so this change makes it possible to pass down a decimal number resulting in frequencies lower than 24 hours, when the value is lower than `1`.

Ticket: https://dimagi.atlassian.net/browse/SC-3223

## Safety Assurance

### Automated test coverage
Unit tests don't cover this logic. I was tempted to add some, but this code lives in a private method so I would need to broaden the scope of the test.

### DIFF
--- a/app/src/org/commcare/utils/SyncDetailCalculations.java
+++ b/app/src/org/commcare/utils/SyncDetailCalculations.java
@@ -153,12 +153,11 @@ public class SyncDetailCalculations {
     private static boolean unsentFormTimeLimitExceeded(long lastSyncTime) {
         SharedPreferences prefs = CommCareApplication.instance().getCurrentApp().getAppPreferences();
         double unsentFormTimeLimitInDays = Double.parseDouble(prefs.getString(UNSENT_FORM_TIME_KEY, "5"));
-        int unsentFormTimeLimitInMinutes = (int)(unsentFormTimeLimitInDays * 24 * 60);
+        long unsentFormTimeLimitInMsecs = (int)(unsentFormTimeLimitInDays * 24 * 60 * 60 * 1000);
 
         long now = new Date().getTime();
-        int secsSinceLastSync = (int)((now - lastSyncTime) / 1000);
-        int minutesSinceLastSync = (int)(secsSinceLastSync / 60);
+        long msecsSinceLastSync = (now - lastSyncTime);
 
-        return minutesSinceLastSync > unsentFormTimeLimitInMinutes;
+        return msecsSinceLastSync > unsentFormTimeLimitInMsecs;
     }
 }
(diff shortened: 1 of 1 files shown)


avazirna:

OrangeAndGreen (dismissed):
The math looks good. Nice that the inverted days_ago variable goes away.

app/src/org/commcare/utils/SyncDetailCalculations.java
```diff
@@ -152,12 +152,13 @@ private static boolean unsentFormNumberLimitExceeded(int numUnsentForms) {
 
     private static boolean unsentFormTimeLimitExceeded(long lastSyncTime) {
         SharedPreferences prefs = CommCareApplication.instance().getCurrentApp().getAppPreferences();
-        int unsentFormTimeLimit = Integer.parseInt(prefs.getString(UNSENT_FORM_TIME_KEY, "5"));
+        double unsentFormTimeLimitInDays = Double.parseDouble(prefs.getString(UNSENT_FORM_TIME_KEY, "5"));
+        int unsentFormTimeLimitInMinutes = (int)(unsentFormTimeLimitInDays * 24 * 60);
```
mkangia, line 156: How about converting this to seconds instead of minutes for even better precision?

mkangia: or miliseconds?
that way you avoid any conversion in the comparison later.

avazirna: I battled with the same question myself and ended up settling with minutes.

app/src/org/commcare/utils/SyncDetailCalculations.java
```diff
@@ -153,12 +153,11 @@ private static boolean unsentFormNumberLimitExceeded(int numUnsentForms) {
...
-        int unsentFormTimeLimitInMinutes = (int)(unsentFormTimeLimitInDays * 24 * 60);
+        long unsentFormTimeLimitInMsecs = (int)(unsentFormTimeLimitInDays * 24 * 60 * 60 * 1000);
 
         long now = new Date().getTime();
-        int secsSinceLastSync = (int)((now - lastSyncTime) / 1000);
-        int minutesSinceLastSync = (int)(secsSinceLastSync / 60);
+        long msecsSinceLastSync = (now - lastSyncTime);
 
-        return minutesSinceLastSync > unsentFormTimeLimitInMinutes;
+        return msecsSinceLastSync > unsentFormTimeLimitInMsecs;
```
mkangia, line 183: Nice!

mkangia (commented):
thanks @avazirna , I will leave the review to mobile team.

OrangeAndGreen (approved):

shubham1g5 (approved):

PR #2747, opened 2024-03-18
Add fullscreen option to in-line videos

avazirna:
## Summary
This PR adds the option to view inline videos in fullscreen mode. The initial approach was to use a third-party video app, but in order to offer a seamless user experience, an internal activity preferred. 

Ticket: https://dimagi.atlassian.net/browse/SAAS-15133 

## Product Description
With this change, there will be an Fullscreen mode button in the media controller panel, which when pressed it transitions to fullscreen mode. And once in fullscreen mode, the same button can be used to transition back:

[](https://github.com/dimagi/commcare-android/assets/19228119/9dac8543-2409-46f1-99e7-10dfbc2927ca)

## Safety Assurance


OrangeAndGreen (dismissed):
One little nit but otherwise all looks good.

app/src/org/commcare/views/media/CommCareMediaController.java
```diff
@@ -50,4 +73,49 @@ public void hide() {
...
+                // if in fullscreen mode, we exit
+                if (fullscreenMode) {
+                    Intent i = new Intent();
+                    i.putExtra(CommCareMediaController.INLINE_VIDEO_TIME_POSITION, videoView.getCurrentPosition());
+                    ((AppCompatActivity)getContext()).setResult(Activity.RESULT_OK, i);
+                    ((AppCompatActivity)getContext()).finish();
+                } else {
+                    Intent intent = new Intent(getContext(), FullscreenVideoViewActivity.class);
+                    intent.setData(FileUtil.getUriForExternalFile(getContext(),
+                            new File(videoView.getVideoPath())));
```
shubham1g5, line 108: we should explicitly validate videoView.getVideoPath() is not null and crash with an exception otherwise.

shubham1g5 (changes requested):
Looks good overall, the PR should have a QA note and release note. We should also include a video/gif of the transition back and forth to get to know how the transition UX feels like.

app/src/org/commcare/views/media/CommCareMediaController.java
```diff
@@ -1,33 +1,56 @@
...
  */
 public class CommCareMediaController extends MediaController {
 
+    public static final String INLINE_VIDEO_TIME_POSITION = "inline-video-time-position";
+
     // A mock to superclass' isShowing property.
     // {@link https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/widget/MediaController.java#96}
     private boolean _isShowing = false;
+    private ImageButton fullscreenBtn;
+    private boolean fullscreenMode;
```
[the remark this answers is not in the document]
avazirna: In the constructor, the fullscreen mode is set during the initialization of the media controller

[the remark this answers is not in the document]
shubham1g5: Maybe I am missing something, but in the diff I am only seeing it set to `false` . Can you point me to the line where you are setting it to `true`

app/res/values/styles.xml
```diff
@@ -293,4 +293,16 @@
     <color name="login_edit_text_color">@color/cc_core_text</color>
     <color name="login_edit_text_color_error">@color/cc_attention_negative_color</color>
     <dimen name="map_button_min_width">120dp</dimen>
+
+    <style name="MediaButton">
+        <item name="background">@null</item>
+        <item name="android:layout_width">71dip</item>
+        <item name="android:layout_height">52dip</item>
```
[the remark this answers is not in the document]
shubham1g5: sounds good in that case.

app/src/org/commcare/views/media/CommCareMediaController.java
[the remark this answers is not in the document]
avazirna: Yes, [here](https://github.com/dimagi/commcare-android/blob/123590ee3a6f6c3d3a29fbf89c13c92f9a9544de/app/src/org/commcare/activities/FullscreenVideoViewActivity.kt#L33), when setting the media controller for the fullscreen mode activity.

app/src/org/commcare/activities/FullscreenVideoViewActivity.kt
```diff
@@ -0,0 +1,86 @@
...
+    private lateinit var viewBinding: ActivityFullscreenVideoViewBinding
+    private var lastPosition = -1
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        viewBinding = ActivityFullscreenVideoViewBinding.inflate(layoutInflater)
+        setContentView(viewBinding.root)
+
+        // Get video URI from intent, finish if no URI is available
+        // TODO: we should inform the user when we finish because there is no data
```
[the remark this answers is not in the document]
avazirna: instead of crashing, I opted for [canceling the activity and informing the user.](https://github.com/dimagi/commcare-android/pull/2747/commits/123590ee3a6f6c3d3a29fbf89c13c92f9a9544de)

app/src/org/commcare/views/media/CommCareMediaController.java
avazirna: The [validation of the of the video path](https://github.com/dimagi/commcare-android/blob/123590ee3a6f6c3d3a29fbf89c13c92f9a9544de/app/src/org/commcare/views/media/MediaLayout.java#L306) happens before the instantiation of the CommCareMediaController, so this branch would only be executed if the file exists.

avazirna:

avazirna:

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -332,6 +332,8 @@ public void onActivityResultSessionSafe(int requestCode, int resultCode, Intent
             } else if (requestCode == FormEntryConstants.INTENT_CALLOUT) {
                 processIntentResponse(intent, true);
                 Toast.makeText(this, Localization.get("intent.callout.cancelled"), Toast.LENGTH_SHORT).show();
+            } else if (requestCode == FormEntryConstants.VIEW_VIDEO_FULLSCREEN) {
+                Toast.makeText(this, Localization.get("media.invalid.video.path"), Toast.LENGTH_SHORT).show();
```
shubham1g5, line 336: This doesn't seem like the ideal validation place to me, we can potentially have other code path ways that can cancel the full screen activity. I think we should not even start the `FullscreenVideoViewActivity.kt` if the video uri is null or invalid and crash hard if we encounter the video path to be null inside the `FullscreenVideoViewActivity`

shubham1g5 (changes requested):

app/src/org/commcare/activities/FormEntryActivity.java
avazirna: Done. https://github.com/dimagi/commcare-android/pull/2747/commits/858864592f1a1190807e43ae7c128dcdea68c3d8

shubham1g5: we should remove the validation here now right ?

avazirna: Yes, it has been removed already, the code there is outdated. Can you double check?

PR #2752, opened 2024-04-26
Improve audio recording configuration

avazirna:
## Summary
This PR was motivated by an issue with the Audio Recording feature in which the resulting recording had no sound. The issue was replicated by audio recording with CommCare and another voice recorder app at the same time, it was noticed that navigating away from CommCare would cause the recording to go silence. A few options were entertained as potential solutions, but ultimately taking advantage of some recording configuration elements introduced in the latest versions of Android, proved to be sufficient to ensure that CommCare audio recording feature is handled with higher priority by the system. 
While the testing was able to, it's important to note that the actual conditions in which the issue occurred could not be established as access to the list of apps in the affected devices were not possible. 

## Product Description
There are a couple change visible to the user:
- CommCare now checks if there is an ongoing recording before initiating a new one, and informs the user with the necessary message when that's the case:

- During the recording, the user is advised not to leave CommCare:

## Safety Assurance

### Safety story
These changes were successfully tested on Android 9, 10, 11 and 12.

### DIFF
--- a/app/assets/locales/android_translatable_strings.txt
+++ b/app/assets/locales/android_translatable_strings.txt
@@ -444,6 +444,7 @@ custom.restore.file.not.set=Custom restore file path is not set in preferences
 custom.restore.error=Error loading custom sync
 
 start.recording=Start Recording
+start.recording.failed=Unable to start recording while another application is recording!
 stop.recording=Stop Recording
 recording.header=Record a sound
 before.recording=Tap to start recording
--- a/app/src/org/commcare/views/widgets/RecordingFragment.java
+++ b/app/src/org/commcare/views/widgets/RecordingFragment.java
@@ -23,9 +23,11 @@ import android.widget.ImageButton;
 import android.widget.LinearLayout;
 import android.widget.ProgressBar;
 import android.widget.TextView;
+import android.widget.Toast;
 
 import org.commcare.CommCareApplication;
 import org.commcare.dalvik.R;
+import org.commcare.utils.MediaUtil;
 import org.javarosa.core.services.locale.Localization;
 
 import java.io.File;
@@ -165,6 +167,13 @@ public class RecordingFragment extends DialogFragment {
     }
 
     private void startRecording() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+            if (MediaUtil.isRecordingActive(getContext())) {
+                Toast.makeText(getContext(), Localization.get("start.recording.failed"), Toast.LENGTH_SHORT).show();
+                return;
+            }
+        }
+
         disableScreenRotation((AppCompatActivity)getContext());
         setCancelable(false);
         setupRecorder();
--- a/app/src/org/commcare/views/widgets/RecordingFragment.java
+++ b/app/src/org/commcare/views/widgets/RecordingFragment.java
@@ -205,8 +205,11 @@ public class RecordingFragment extends DialogFragment {
         }
 
         boolean isHeAacSupported = isHeAacEncoderSupported();
-
-        recorder.setAudioSource(MediaRecorder.AudioSource.MIC);
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+            recorder.setAudioSource(MediaRecorder.AudioSource.VOICE_COMMUNICATION);
+       } else {
+            recorder.setAudioSource(MediaRecorder.AudioSource.MIC);
+       }
         recorder.setAudioSamplingRate(isHeAacSupported ? HEAAC_SAMPLE_RATE : AMRNB_SAMPLE_RATE);
         recorder.setOutputFormat(MediaRecorder.OutputFormat.MPEG_4);
         recorder.setOutputFile(fileName);
--- a/app/src/org/commcare/views/widgets/RecordingFragment.java
+++ b/app/src/org/commcare/views/widgets/RecordingFragment.java
@@ -210,6 +210,10 @@ public class RecordingFragment extends DialogFragment {
        } else {
             recorder.setAudioSource(MediaRecorder.AudioSource.MIC);
        }
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+            recorder.setPrivacySensitive(true);
+        }
         recorder.setAudioSamplingRate(isHeAacSupported ? HEAAC_SAMPLE_RATE : AMRNB_SAMPLE_RATE);
         recorder.setOutputFormat(MediaRecorder.OutputFormat.MPEG_4);
         recorder.setOutputFile(fileName);
--- a/app/src/org/commcare/utils/MediaUtil.java
+++ b/app/src/org/commcare/utils/MediaUtil.java
@@ -4,6 +4,8 @@ import android.content.Context;
 import android.content.res.TypedArray;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
+import android.media.AudioManager;
+import android.os.Build;
 import android.util.DisplayMetrics;
 import android.util.Log;
 import android.util.Pair;
@@ -11,8 +13,6 @@ import android.view.WindowManager;
 
 import org.commcare.CommCareApplication;
 import org.commcare.engine.references.JavaFileReference;
-import org.commcare.google.services.analytics.AnalyticsParamValue;
-import org.commcare.google.services.analytics.FirebaseAnalyticsUtil;
 import org.commcare.preferences.HiddenPreferences;
 import org.commcare.util.LogTypes;
 import org.javarosa.core.reference.InvalidReferenceException;
@@ -27,6 +27,7 @@ import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 
 import androidx.annotation.Nullable;
+import androidx.annotation.RequiresApi;
 
 /**
  * @author ctsims
@@ -504,6 +505,12 @@ public class MediaUtil {
         } catch (OutOfMemoryError e) {
             return performSafeScaleDown(imageFilepath, scaleDownFactor + 1, depth + 1);
         }
+    }
+
+    @RequiresApi(api = Build.VERSION_CODES.N)
+    public static boolean isRecordingActive(Context context){
+        return ((AudioManager) context.getSystemService(Context.AUDIO_SERVICE))
+                .getActiveRecordingConfigurations().size() > 0;
     }
 
 }


app/src/org/commcare/views/widgets/RecordingFragment.java
```diff
@@ -196,8 +205,15 @@ private void setupRecorder() {
         }
 
         boolean isHeAacSupported = isHeAacEncoderSupported();
-
-        recorder.setAudioSource(MediaRecorder.AudioSource.MIC);
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+            recorder.setAudioSource(MediaRecorder.AudioSource.VOICE_COMMUNICATION);
+       } else {
+            recorder.setAudioSource(MediaRecorder.AudioSource.MIC);
+       }
```
shubham1g5, line 212: We should test that the recording with source set to `VOICE_COMMUNICATION` are playable in browser (instead of download) on HQ.

shubham1g5 (commented):
Couple general comments - 

1. Can we add a Jira link and repro steps in the PR description to help anyone coming across this PR in future. 

2. We also discussed setting up the audio config callback and pause recording on our side if we get `isClientSilenced() as true to handle situations in which user navigates away from CC recording and start a new recording. Curious if there is a reason so that we decided to not implement it ?

avazirna:
> Curious if there is a reason so that we decided to not implement it ?
I'm indecisive about it. I did implement it but during tests, the changes to the configuration seemed sufficient to ensure that CC continues recording even though other apps are trying to record (the other apps get not sound). However, I don't have certainty about whether the third-party apps that I used have the same configuration, so maybe better to include. What do you think?

shubham1g5:
I would think that we should include it as it seems to be the most fullproof solution to have empty recordings (callback works irrespective of Android version ?)

avazirna:
> callback works irrespective of Android version ?

From Android 7.1. But the tricky part is in assessing when the recording has gone silent, `isClientSilenced ()` was introduced in Android 11. For versions prior to 11, I was considering using [getMaxAmplitude()](https://developer.android.com/reference/android/media/MediaRecorder#getMaxAmplitude()) but I wasn't able to thoroughly test its reliability. Any thoughts?

avazirna:
>  callback works irrespective of Android version ?
> 
From Android 7.1. But the tricky part is in assessing when the recording has gone silent, `isClientSilenced ()` was introduced in Android 10. For versions prior to 10, I was considering using [getMaxAmplitude()](https://developer.android.com/reference/android/media/MediaRecorder#getMaxAmplitude()) but I wasn't able to thoroughly test its reliability. Any thoughts?

app/src/org/commcare/views/widgets/RecordingFragment.java
```diff
@@ -320,6 +329,9 @@ private void saveRecording() {
         if (listener != null) {
             listener.onRecordingCompletion(fileName);
         }
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+            unregisterAudioRecordingConfigurationChangeCallback();
```
shubham1g5, line 336: Would this callback wil never get unregistered if user don't call save() ?

app/src/org/commcare/views/widgets/RecordingFragment.java
```diff
@@ -428,6 +428,19 @@ private void registerAudioRecordingConfigurationChangeCallback() {
...
+                    return;
+                }
+
+                if (hasRecordingGoneSilent(configs)) {
+                    if (!inPausedState) {
+                        pauseRecording();
+                    }
+                } else {
+                    if (inPausedState) {
+                        resumeRecording();
```
shubham1g5, line 453: why should we ever resume a silent recording ?

app/assets/locales/android_translatable_strings.txt
```diff
@@ -459,6 +459,8 @@ recording.clear=Clear
 recording.prompt.with.file.chooser=Record or choose sound below
 recording.prompt.without.file.chooser=Record sound below
 recording.custom=Recorded Sound
+recording.paused.due.another.app.recording.title=CommCare Audio Recording
+recording.paused.due.another.app.recording.message=Recording paused as another app with higher privileges started recording. Click here to resume the recording!
```
shubham1g5, line 463: > with higher privileges

technical details should be omitted for users

app/src/org/commcare/views/widgets/RecordingFragment.java
```diff
@@ -433,10 +438,20 @@ public void onRecordingConfigChanged(List<AudioRecordingConfiguration> configs)
                 if (hasRecordingGoneSilent(configs)) {
                     if (!inPausedState) {
                         pauseRecording();
+                        NotificationUtil.showNotification(
```
shubham1g5, line 441: We should also update the status on the recording dialog to show this message.

app/src/org/commcare/views/widgets/RecordingFragment.java
```diff
@@ -433,10 +438,20 @@ public void onRecordingConfigChanged(List<AudioRecordingConfiguration> configs)
                 if (hasRecordingGoneSilent(configs)) {
                     if (!inPausedState) {
                         pauseRecording();
+                        NotificationUtil.showNotification(
+                                getContext(),
+                                CommCareNoficationManager.NOTIFICATION_CHANNEL_USER_SESSION_ID,
+                                RECORDING_NOTIFICATION_ID,
+                                Localization.get("recording.paused.due.another.app.recording.title"),
+                                Localization.get("recording.paused.due.another.app.recording.message"),
+                                new Intent(getContext(), DispatchActivity.class)
```
shubham1g5, line 447: Does on clicking notification, user resumes the recording screen or get taken out of form ?

app/src/org/commcare/utils/NotificationUtil.java
```diff
@@ -0,0 +1,48 @@
...
+import org.commcare.activities.DispatchActivity;
+import org.commcare.dalvik.R;
+
+/**
+ * Set of methods to post notifications to the user.
+ *
+ * @author avazirna
+ */
+public class NotificationUtil {
+    public static void showNotification(Context context, String notificationChannel, int notificationId,
```
shubham1g5, line 22: can use `NotificationManager.reportNotificationMessage()` ?

shubham1g5 (changes requested):
> For versions prior to 10, I was considering using [getMaxAmplitude()](https://developer.android.com/reference/android/media/MediaRecorder#getMaxAmplitude()) but I wasn't able to thoroughly test its reliability. Any thoughts?

I think the only way to check will be to test it out. Have you been facing difficulties testing ?

app/src/org/commcare/views/widgets/RecordingFragment.java
avazirna: If we navigate back to CommCare, bringing it to the top and depending on the configuration of the other voice recording app, this might trigger a change in the recording configuration and CommCare will start capturing sound again. In this situation, we want to resume the recording without any action from the user.

app/src/org/commcare/views/widgets/RecordingFragment.java
avazirna: I think once the recording starts, we [can't dismiss the dialog](https://github.com/dimagi/commcare-android/blob/2df74444f4fce60f3c0545022fab8f837c73b90b/app/src/org/commcare/views/widgets/RecordingFragment.java#L189C9-L189C25) nor [go back](https://github.com/dimagi/commcare-android/blob/2df74444f4fce60f3c0545022fab8f837c73b90b/app/src/org/commcare/views/widgets/RecordingFragment.java#L169) without saving.

app/src/org/commcare/utils/NotificationUtil.java
avazirna: As discussed, because the goal is to go back to the recording the notification action can't be to go to the `MessageActivity`. This utility class is also to be used to consolidate other places where we post notifications.

app/src/org/commcare/views/widgets/RecordingFragment.java
shubham1g5: got it, can we also check if the callback works after screen rotation i.e rotate the screen after you start recording and then change audio config and see if the callback still gets triggered?

app/src/org/commcare/views/widgets/RecordingFragment.java
shubham1g5: I tend to disagree with this, If we resume recording without any user intervention it would seem to user that recording was never paused. Insted when user comes back they should see a message that recording was paused due to user navigating away and give user option to resume recording.

app/src/org/commcare/utils/NotificationUtil.java
shubham1g5: I see, is it possible to reuse this method in `NotificationManager.reportNotificationMessage()` that case. Just wondering if there is a way to avoid code duplication.

avazirna: Yes, the plan is to refactor everywhere where we have this code to this utility class in a separate PR.

app/src/org/commcare/views/widgets/RecordingFragment.java
avazirna: I understand. The rationale here was that because the user didn't directly action to pause the recording, CC should resume it in the same terms. But given that we will provide guidance to the user, it makes sense to let the user resume the recording.

avazirna:
> I think the only way to check will be to test it out. Have you been facing difficulties testing ?

The testing was successful. The tricky part is in having certainty that 0 (zero) amplitude always mean that the recording has gone silent and not that maybe that there is no sound to record, i.e. people stopped talking and there is no background noise.

app/src/org/commcare/views/widgets/RecordingFragment.java
avazirna: screen rotation is disabled in [onCreateView()](https://github.com/dimagi/commcare-android/blob/2df74444f4fce60f3c0545022fab8f837c73b90b/app/src/org/commcare/views/widgets/RecordingFragment.java#L81) and [startRecording()](https://github.com/dimagi/commcare-android/blob/2df74444f4fce60f3c0545022fab8f837c73b90b/app/src/org/commcare/views/widgets/RecordingFragment.java#L168)

app/src/org/commcare/views/widgets/RecordingFragment.java
shubham1g5: Any updates here ?

shubham1g5:
>  people stopped talking and there is no background noise.

Are you getting a 0 output from the method in this case ?

app/src/org/commcare/views/widgets/RecordingFragment.java
avazirna: Yes, I tested it and it plays normally.

avazirna:
> > people stopped talking and there is no background noise.
> 
> Are you getting a 0 output from the method in this case ?

Not necessarily, which gives me 80% confidence that this option should work fine, so happy to push this out and have QA confirm that that's the case. Agree?

shubham1g5:
> Not necessarily, which gives me 80% confidence that this option should work fine, so happy to push this out and have QA confirm that that's the case. Agree?

Sounds good to me.

app/src/org/commcare/views/widgets/RecordingFragment.java
```diff
@@ -469,16 +469,35 @@ private void unregisterAudioRecordingConfigurationChangeCallback() {
...
+            // this loop is to get at least 3 zero amplitude readings to confirm that the microphone has gone
+            // silent
+            for (int i = 0; i < 3; i++) {
+                if (recorder == null || recorder.getMaxAmplitude() != 0) {
+                    return false;
+                }
+
+                try {
+                    Thread.sleep(500);
+                } catch (InterruptedException e) {
```
shubham1g5, line 506: This doesn't feel right to me. If we don't think `recorder.getMaxAmplitude()`, lets not use it at all instead of hacking something ambiguous here. I would suggest that we only do a one time check on `recorder.getMaxAmplitude()` and treat it as reliable unless we find otherwise

shubham1g5 (changes requested):

shubham1g5 (approved):
Looks good :)

avazirna:

avazirna:

avazirna:

PR #2774, opened 2024-06-28
Jp/french language support

j13panchal:
Added French translation of CCC mobile UI
Added Add CCC job ID to analytics message

## Summary
https://dimagi.atlassian.net/browse/CCCT-338
https://dimagi.atlassian.net/browse/CCCT-340

## Safety Assurance


app/src/org/commcare/CommCareApplication.java
```diff
@@ -275,9 +279,42 @@ protected void turnOnStrictMode() {
...
 
     @Override
-    public void onConfigurationChanged(Configuration newConfig) {
+    public void onConfigurationChanged(@NonNull Configuration newConfig) {
         super.onConfigurationChanged(newConfig);
         LocalePreferences.saveDeviceLocale(newConfig.locale);
+        updateLocale(this);
+    }
+
+    private void updateLocale(Context context) {
```
OrangeAndGreen, line 288: Is this method needed? Seems like the OS properly updates views when the user switches languages, even if I comment this out.

app/res/values-fr/strings.xml
```diff
@@ -161,7 +161,7 @@ License.
     <string name="choose_sound">Choisissez le Son</string>
     <string name="choose_video">Choisissez la Vidéo</string>
     <string name="choose_image">Choisissez l\'Image</string>
-    <string name="selected">"Sélectioné: "</string>
+    <string name="selected" tools:ignore="PrivateResource">"Sélectioné: "</string>
```
OrangeAndGreen, line 164: Curious about why the change here?

OrangeAndGreen (commented):
I added a couple questions. Also, I started going through the various screens and generating screenshots to identify issues and/or strings that still need to be translated. I linked the screenshots doc to the Jira ticket. So far I've gone through most of the Connect screens but we still need to look at the ConnectId screens as well.

app/res/values-fr/strings.xml
j13panchal: No other changes were made here. I just resolved the warning that was shown.

app/src/org/commcare/CommCareApplication.java
j13panchal: Yes, you're right. The OS automatically updates the view after changing the language from the system settings without needing to set the locale in the code.

j13panchal: Code updated

app/src/org/commcare/CommCareApplication.java
```diff
@@ -275,9 +277,8 @@ protected void turnOnStrictMode() {
     }
 
     @Override
-    public void onConfigurationChanged(Configuration newConfig) {
+    public void onConfigurationChanged(@NonNull Configuration newConfig) {
         super.onConfigurationChanged(newConfig);
-        LocalePreferences.saveDeviceLocale(newConfig.locale);
```
OrangeAndGreen, line 280: I think this line may still be necessary for other parts of the code

app/src/org/commcare/activities/CommCareSetupActivity.java
```diff
@@ -888,6 +888,7 @@ private boolean isSingleAppBuild() {
     public void onRequestPermissionsResult(int requestCode,
                                            @NonNull String[] permissions,
                                            @NonNull int[] grantResults) {
+        super.onRequestPermissionsResult(requestCode, permissions, grantResults);
```
OrangeAndGreen, line 891: Thinking this should be okay but just want to make sure it wasn't left out on purpose since this is traditional CC code. @shubham1g5?

app/src/org/commcare/activities/LoginActivityUiController.java
```diff
@@ -550,6 +550,7 @@ protected void setConnectIdLoginState(boolean useConnectId) {
             text = Localization.get("login.button");
         }
         loginButton.setText(text);
+        connectLoginButton.setText(activity.getString(R.string.connect_button_logged_in));
```
OrangeAndGreen, line 553: This is set statically in the XML, does it need to be updated here?

j13panchal: I included these lines because when a user changes the language from the system settings while in the app, the strings don't translate correctly when they return to the app. That's why I added this code.

app/src/org/commcare/activities/CommCareSetupActivity.java
j13panchal: I don't think this will cause any issues in the future. The code was showing an error, which is why I resolved it. However, if you still want to remove this line, I can do that.

app/src/org/commcare/CommCareApplication.java
j13panchal: Yes, that's correct. This line is already present in the onCreate method, which is why I removed it from here.

app/src/org/commcare/activities/LoginActivityUiController.java
OrangeAndGreen: Ah, yeah I see it happening too, not sure why. Add a comment above explaining why we're doing it just so we don't forget.

app/src/org/commcare/CommCareApplication.java
j13panchal: Yes, LocalePreferences should update correctly after changing the language while the app is running.

I have already tested this scenario.

app/src/org/commcare/activities/LoginActivityUiController.java
j13panchal: Okay

OrangeAndGreen (approved):

PR #2778, opened 2024-07-04
Override RxJava Global Error Handler

avazirna:
## Summary
RxJava 2.x brought changes to the way Throwable errors are handled, causing them to not be swallowed. This means that when any exception that is not being handled occurs, RxJava now bubbles it up through the current thread's uncaught exception handler, which causes CommCare to crash. This PR overrides RxJava Global error handler to offer a more controlled behaviour.  

## Safety Assurance

### Safety story
This is only to swallow RxJava RequestAbandonedException exceptions, other exceptions should continue being handled the same way.

### DIFF
--- a/app/src/org/commcare/CommCareApplication.java
+++ b/app/src/org/commcare/CommCareApplication.java
@@ -137,6 +137,7 @@ import javax.crypto.SecretKey;
 import io.noties.markwon.Markwon;
 import io.noties.markwon.ext.strikethrough.StrikethroughPlugin;
 import io.noties.markwon.ext.tables.TablePlugin;
+import io.reactivex.exceptions.UndeliverableException;
 import io.reactivex.plugins.RxJavaPlugins;
 import okhttp3.MultipartBody;
 import okhttp3.RequestBody;
@@ -1211,7 +1212,11 @@ public class CommCareApplication extends MultiDexApplication {
 
     private void setRxJavaGlobalHandler() {
         RxJavaPlugins.setErrorHandler(throwable -> {
-            // Not sure the type of exceptions we should handle here
+            // Swallow UndeliverableException
+            if (throwable instanceof UndeliverableException) {
+                return;
+            }
+            Thread.currentThread().getUncaughtExceptionHandler().uncaughtException(Thread.currentThread(), throwable);
         });
     }
 }
--- a/app/src/org/commcare/CommCareApplication.java
+++ b/app/src/org/commcare/CommCareApplication.java
@@ -137,6 +137,7 @@ import javax.crypto.SecretKey;
 import io.noties.markwon.Markwon;
 import io.noties.markwon.ext.strikethrough.StrikethroughPlugin;
 import io.noties.markwon.ext.tables.TablePlugin;
+import io.reactivex.plugins.RxJavaPlugins;
 import okhttp3.MultipartBody;
 import okhttp3.RequestBody;
 
@@ -251,8 +252,9 @@ public class CommCareApplication extends MultiDexApplication {
         FirebaseMessagingUtil.verifyToken();
 
         customiseOkHttp();
-    }
 
+        setRxJavaGlobalHandler();
+    }
 
     protected void loadSqliteLibs() {
         SQLiteDatabase.loadLibs(this);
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/CommCareApplication.java
```diff
@@ -1206,4 +1209,14 @@ public boolean isNsdServicesEnabled() {
...
         CommCareNetworkServiceGenerator.customizeRetrofitSetup(new OkHttpBuilderCustomConfig());
     }
+
+    private void setRxJavaGlobalHandler() {
+        RxJavaPlugins.setErrorHandler(throwable -> {
+            // Swallow UndeliverableException
+            if (throwable instanceof UndeliverableException) {
+                return;
+            }
+            Thread.currentThread().getUncaughtExceptionHandler().uncaughtException(Thread.currentThread(), throwable);
```
shubham1g5, line 1219: Does this deliver the exception to `CommCareExceptionHandler` ?

app/src/org/commcare/CommCareApplication.java
```diff
@@ -1206,4 +1209,14 @@ public boolean isNsdServicesEnabled() {
     public void customiseOkHttp() {
         CommCareNetworkServiceGenerator.customizeRetrofitSetup(new OkHttpBuilderCustomConfig());
     }
+
+    private void setRxJavaGlobalHandler() {
+        RxJavaPlugins.setErrorHandler(throwable -> {
+            // Swallow UndeliverableException
+            if (throwable instanceof UndeliverableException) {
+                return;
```
shubham1g5, line 1217: Instead of completely swallowing the error we should deliver the wrapped exception to the uncaught ex handler similar to [here](https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#error-handling)

shubham1g5 (changes requested):
can we add the sample firebase crash stacktrace to the PR description for which we are making the change as well.

app/src/org/commcare/CommCareApplication.java
avazirna: Yes, it does.

app/src/org/commcare/CommCareApplication.java
avazirna: [There](https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#error-handling), they don't deliver the exception to the handler, they log it, right?

shubham1g5: They deliver it to handler - 

````
  if (e instanceof UndeliverableException) {
        e = e.getCause();
  }
  // further checks on new e retrieved from above
````

avazirna: But we should swallow when the wrapped exception is `RequestAbandonedException`

shubham1g5: whats the reasoning behind swallowing `RequestAbandonedException` ?

avazirna: We throw it when the Thread is in the interrupted state and I didn't see any exception handling around it, so it seems that it was being swallowed in the previous versions of RxJava and that was intentional. Right?

shubham1g5: I don't think it was intentional, unless there is a reason to swallow it we should do the same handling for it as well.

avazirna: Let me see if I can find any context on that, I also don't see any messages associated with that exception.

ctsims: Confirming that `RequestAbandonedException` is intended to kill its thread (or allow it to terminate its semantic operation and return to a thread pool) as quickly as possible without bubbling back up to the main thread, so swallowing in this fashion is inline with expectation.

shubham1g5: makes sense, thanks for clarification. Just to clarify It doesn't seem like we were swallowing this earlier and it would have been crashing the app which I agree doesn't sound like right behaviour.

avazirna: I think RxJava was swallowing as part of its behaviour something that changed in later versions as described [here](https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#error-handling). What puzzles me is why this only started manifesting when we bumped to version `2.2.x`, from a quick search in Crashlytics we can see that crashes only started when we bumped to that version.

shubham1g5 (approved):

PR #2781, opened 2024-07-08
CCCT-385- daily limit bug fix

pm-dimagi:
https://dimagi.atlassian.net/browse/CCCT-385
Daily limit bug fix

## Safety Assurance


app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
```diff
@@ -371,10 +371,12 @@ public String getMoneyString(int value) {
         return String.format(Locale.getDefault(), "%d%s", value, currency);
     }
 
+
     public int numberOfDeliveriesToday() {
         int dailyVisitCount = 0;
         Date today = new Date();
         for (ConnectJobDeliveryRecord record : deliveries) {
+            record.getDate();
```
OrangeAndGreen, line 379: This is extra, should be removed

app/src/org/commcare/android/database/connect/models/ConnectJobDeliveryRecord.java
```diff
@@ -74,7 +74,7 @@ public static ConnectJobDeliveryRecord fromJson(JSONObject json, int jobId) thro
         delivery.lastUpdate = new Date();
 
         delivery.deliveryId = json.has(META_ID) ? json.getInt(META_ID) : -1;
-        delivery.date = json.has(META_DATE) ? ConnectNetworkHelper.parseDate(json.getString(META_DATE)) : new Date();
+        delivery.date = json.has(META_DATE) ? ConnectNetworkHelper.convertUTCToLocal(json.getString(META_DATE)) : new Date();
```
OrangeAndGreen, line 77: I think the value that gets stored in the database should stay in UTC. The problem is that they may travel to another time zone later and we'll still want to know the correct dates. If we know the DB value is always UTC, we can always adjust it properly later.

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
```diff
@@ -403,10 +405,20 @@ public Hashtable<String, Integer> getDeliveryCountsPerPaymentUnit(boolean todayO
         for(int i = 0; i < deliveries.size(); i++) {
             ConnectJobDeliveryRecord delivery = deliveries.get(i);
             int oldCount = 0;
-            if(paymentCounts.containsKey(delivery.getSlug())) {
-                oldCount = paymentCounts.get(delivery.getSlug());
+            boolean isSameDay=false;
```
OrangeAndGreen, line 408: This variable can move into the todayOnly case

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
```diff
@@ -403,10 +405,20 @@ public Hashtable<String, Integer> getDeliveryCountsPerPaymentUnit(boolean todayO
...
+            if(todayOnly){
+                Date today = new Date();
+                isSameDay=sameDay(today,delivery.getDate());
+                if(paymentCounts.containsKey(delivery.getSlug()) && isSameDay) {
```
OrangeAndGreen, line 412: Would be better not to duplicate this piece of code. Instead how about defining a boolean before the if-structure like:
boolean shouldInclude = true;

Then in the todayOnly case, you can update the value like:
shouldInclude = sameDay;

The else case goes away. Then after exiting the if-structure, you can do the following:
if(shouldInclude && paymentCounts.containsKey(delivery.getSlug()) ) {
   oldCount = paymentCounts.get(delivery.getSlug());
}

app/src/org/commcare/connect/network/ConnectNetworkHelper.java
```diff
@@ -81,7 +82,35 @@ private static ConnectNetworkHelper getInstance() {
 
     private static final DateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd", Locale.getDefault());
     public static Date parseDate(String dateStr) throws ParseException {
-        return dateFormat.parse(dateStr);
+        Date issueDate=dateFormat.parse(dateStr);
+        return issueDate;
+    }
+
+
+    public static Date convertUTCToLocal(String utcDateString) {
```
OrangeAndGreen, line 90: Proposed change: instead of this method, create a method that will create a local date given a UTC date. Then have ConnectJobRecord.sameDay() convert the UTC dates to local before it checks to see if they're the same day.

OrangeAndGreen (changes requested):

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
```diff
@@ -403,10 +404,10 @@ public Hashtable<String, Integer> getDeliveryCountsPerPaymentUnit(boolean todayO
         for(int i = 0; i < deliveries.size(); i++) {
             ConnectJobDeliveryRecord delivery = deliveries.get(i);
             int oldCount = 0;
-            if(paymentCounts.containsKey(delivery.getSlug())) {
-                oldCount = paymentCounts.get(delivery.getSlug());
-            }
-
+            boolean isSameDay=sameDay(new Date(),delivery.getDate());
+                if(paymentCounts.containsKey(delivery.getSlug()) && (isSameDay && todayOnly)) {
```
OrangeAndGreen, line 408: I think this:
(isSameDay && todayOnly)
needs to change to:
(isSameDay || !todayOnly)

app/src/org/commcare/connect/network/ConnectNetworkHelper.java
```diff
@@ -81,7 +82,64 @@ private static ConnectNetworkHelper getInstance() {
...
```
OrangeAndGreen, line 90: No callers for this function anymore, it should be deleted.

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
```diff
@@ -416,14 +417,12 @@ public Hashtable<String, Integer> getDeliveryCountsPerPaymentUnit(boolean todayO
         Hashtable<String, Integer> paymentCounts = new Hashtable<>();
         for(int i = 0; i < deliveries.size(); i++) {
             ConnectJobDeliveryRecord delivery = deliveries.get(i);
-            if(!todayOnly || sameDay(new Date(), delivery.getDate())) {
-                int oldCount = 0;
-                if (paymentCounts.containsKey(delivery.getSlug())) {
+            int oldCount = 0;
+            boolean isSameDay=sameDay(new Date(),delivery.getDate());
+                if(paymentCounts.containsKey(delivery.getSlug()) && (isSameDay && todayOnly)) {
```
OrangeAndGreen, line 422: Pretty sure the logic here is still incorrect. For example, if todayOnly=false (i.e. we want the all-time counts), the code in the if-statement never gets called.

I wasn't sure whether we'd get this ticket finished before the deadline for QA and wanted to make sure at least this bug in the code was fixed, so in the meantime I added what I believe is a correct fix which you can see [here](https://github.com/dimagi/commcare-android/blob/97c31019f963f2821e69359d16a225449fc193c7/app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java#L437C13-L444C14).

The idea is to only consider records when either:
- We want all-time (i.e. todayOnly=false)
OR
- The record is from today

If either check above passes, then the record is sorted in to paymentCounts.

app/src/org/commcare/connect/network/ConnectNetworkHelper.java
```diff
@@ -81,7 +82,37 @@ private static ConnectNetworkHelper getInstance() {
 
     private static final DateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd", Locale.getDefault());
     public static Date parseDate(String dateStr) throws ParseException {
-        return dateFormat.parse(dateStr);
+        Date issueDate=dateFormat.parse(dateStr);
+        return issueDate;
+    }
+
+    public static Date convertUTCToDate(String utcDateString) {
+        SimpleDateFormat utcFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'", Locale.getDefault());
```
OrangeAndGreen, line 90: Nit: Let's make the SimpleDateFormats private static final variables (like line 83). They can easily be reused, and I think they're unfortunately expensive to REPEATEDLY create and destroy.

app/src/org/commcare/connect/network/ConnectNetworkHelper.java
```diff
@@ -81,7 +82,37 @@ private static ConnectNetworkHelper getInstance() {
...
+            date = utcFormat.parse(utcDateString);
+        } catch (ParseException e) {
+            e.printStackTrace();
+        }
+
+        return date;
+    }
+
+    public static Date convertDateToLocal(Date utcDate) {
+        SimpleDateFormat localFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'", Locale.getDefault());
```
OrangeAndGreen, line 104: Same comment as above, reuse this

OrangeAndGreen (changes requested):

OrangeAndGreen (approved):

PR #2786, opened 2024-07-12
Set icon for CommCare update navigation drawer item

avazirna:
## Summary
When in `Root module menu as home screen` mode, after 3 unsuccessful prompts to update CommCare, CommCare displays an item in the navigation drawer for the user to trigger the update. This item doesn't have an image resource associated, causing CC to crash when this item needs to be rendered. 

Note for reviewers, please provide some feedback around the icon as well, I'm not 100% happy with it.
<table>
<tr><td>Initial</td><td>Final</td></tr>
<tr><td>

</td>
<td></td></tr></table>

## Feature Flag
Custom property `cc-use-root-menu-as-home-screen`

## Product Description
Adds an icon for the `CommCare update`  item in the navigation drawer.

## Safety Assurance

### Safety story
No concerns, but we might need to update our QA plan to make sure this is noted during testing.


avazirna:

avazirna:
Feedback from @OrangeAndGreen: Maybe the CC logo inside the circular arrow, i.e. inverted from your current design?

app/src/org/commcare/activities/HomeNavDrawerController.java
```diff
@@ -221,6 +221,8 @@ private static int getItemIcon(String id) {
                 return R.drawable.ic_logout_nav_drawer;
             case TRAINING_DRAWER_ITEM_ID:
                 return R.drawable.ic_training_nav_drawer;
+            case UPDATE_CC_DRAWER_ITEM_ID:
+                return R.drawable.ic_cc_update_nav_drawer;
```
shubham1g5, line 225: we can probably use https://github.com/dimagi/commcare-android/blob/master/app/res/drawable/update_download_icon.xml instead ?

avazirna: It seems a valid option, let me see how is looks

avazirna: This is how it looks. I don't think it properly conveys the action, but with the text next it it should be fine.

shubham1g5: Actually can we use [this instead](https://github.com/dimagi/commcare-android/blob/master/app/original_vector_graphics/update_download.svg) to get rid of the background circle for consistency with other icons

avazirna: I might need to change the colour as well, let me check

avazirna: Much better

shubham1g5: yes, looks great!

avazirna: approved?

shubham1g5 (approved):

app/res/drawable/ic_cc_update.xml
```diff
@@ -0,0 +1,5 @@
+<vector android:height="25dp" android:viewportHeight="232.5"
```
shubham1g5, line 1: any reason we can't use the existing `.svg` file here?

avazirna: I converted it to vector drawable because their rendering tends to be faster and compatibility with older android versions is higher. Is there any particular advantage with using `svg` files?

shubham1g5: no was just curious why we need to create another file.

PR #2789, opened 2024-07-17
PM_CCCt_366- pr upgrade

pm-dimagi:
https://dimagi.atlassian.net/browse/CCCT-366

## Safety Assurance

### DIFF
--- a/app/src/org/commcare/activities/connect/ConnectIdPhoneActivity.java
+++ b/app/src/org/commcare/activities/connect/ConnectIdPhoneActivity.java
@@ -35,6 +35,8 @@ public class ConnectIdPhoneActivity extends CommCareActivity<ConnectIdPhoneActiv
     private String existingPhone;
     private ConnectIdPhoneActivityUiController uiController;
 
+    protected boolean skipPhoneNumberCheck = false;
+
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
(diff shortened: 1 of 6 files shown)


OrangeAndGreen (changes requested):
Getting closer... I'm seeing the phone number picker now in both register and recover scenarios.

There's an issue in the register scenario where the API call is being called too many times (3 instead of 1).

Another issue: in the recovery scenario, when I select a number from the picker, the country code changes from +1 to just 1. This causes a 500 error a bit later when I try to request an OTP from the server using the incorrect number (missing the +).

Also some code cleanup to do.

app/src/org/commcare/activities/connect/ConnectIdPhoneActivity.java
```diff
@@ -90,33 +86,19 @@ void displayNumber(String fullNumber) {
         if (fullNumber != null && fullNumber.startsWith("+" + codeText)) {
             fullNumber = fullNumber.substring(codeText.length() + 1);
         }
-
+        skipPhoneNumberCheck = false;
         uiController.setPhoneNumber(fullNumber);
+        skipPhoneNumberCheck = true;
```
OrangeAndGreen, line 91: Should the line below (setCountryCode) be moved up above line 89? Thinking we want to make sure that both the number and country code have been updated to the latest selection before letting the API call trigger

app/src/org/commcare/activities/connect/ConnectIdPhoneActivity.java
```diff
@@ -170,6 +152,7 @@ public void handleButtonPress() {
             IApiCallback callback = new IApiCallback() {
                 @Override
                 public void processSuccess(int responseCode, InputStream responseData) {
+                    skipPhoneNumberCheck=false;
```
OrangeAndGreen, line 155: Need to add this for the three error conditions as well

app/src/org/commcare/activities/connect/ConnectIdPhoneActivity.java
pm-dimagi: checked this on debug

OrangeAndGreen (approved):

PR #2790, opened 2024-07-18
CCCT-77 || Push Notification Redirection

j13panchal:
## Summary
https://dimagi.atlassian.net/browse/CCCT-77

## Safety Assurance


app/src/org/commcare/activities/DispatchActivity.java
```diff
@@ -17,10 +17,12 @@
 import org.commcare.preferences.DeveloperPreferences;
 import org.commcare.recovery.measures.ExecuteRecoveryMeasuresActivity;
 import org.commcare.recovery.measures.RecoveryMeasuresHelper;
+import org.commcare.util.LogTypes;
```
OrangeAndGreen, line 20: Oops, a couple unnecessary imports here

app/src/org/commcare/services/CommCareFirebaseMessagingService.java
```diff
@@ -28,81 +29,95 @@
...
     public void onMessageReceived(RemoteMessage remoteMessage) {
-        Logger.log(LogTypes.TYPE_FCM, "Message received: " + remoteMessage.getMessageId());
+        Logger.log(LogTypes.TYPE_FCM, "CommCareFirebaseMessagingService Message received: " + remoteMessage.getData());
         Map<String, String> payloadData = remoteMessage.getData();
-        RemoteMessage.Notification payloadNotification = remoteMessage.getNotification();
-
-        if (payloadNotification != null) {
-            showNotification(payloadNotification);
-        }
+//      RemoteMessage.Notification payloadNotification = remoteMessage.getNotification();
```
OrangeAndGreen, line 56: Delete commented code

app/src/org/commcare/services/CommCareFirebaseMessagingService.java
```diff
@@ -28,81 +29,95 @@
...
+            intent.addCategory(Intent.CATEGORY_LAUNCHER);
+        }
+        intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP | Intent.FLAG_ACTIVITY_SINGLE_TOP | Intent.FLAG_ACTIVITY_NEW_TASK);
 
         PendingIntent contentIntent;
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M)
-            contentIntent = PendingIntent.getActivity(this, 0, i, PendingIntent.FLAG_IMMUTABLE);
-        else
-            contentIntent = PendingIntent.getActivity(this, 0, i, 0);
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
```
OrangeAndGreen, line 107: Nit: Code in the if-else can just specify an (int) flags variable, then in one line below:
PendingIntent contentIntent = PendingIntent.getActivity(this, 0, intent,  flags);

app/src/org/commcare/utils/FirebaseMessagingUtil.java
```diff
@@ -11,6 +11,7 @@
 import org.commcare.CommCareApplication;
 import org.commcare.dalvik.BuildConfig;
 import org.commcare.services.FCMMessageData;
+import org.commcare.util.LogTypes;
```
OrangeAndGreen, line 14: Unused, delete

app/src/org/commcare/activities/connect/ConnectActivity.java
```diff
@@ -8,11 +8,27 @@
...
 import org.commcare.android.database.connect.models.ConnectJobRecord;
 import org.commcare.connect.ConnectManager;
+import org.commcare.connect.network.ApiConnect;
+import org.commcare.connect.network.ConnectNetworkHelper;
+import org.commcare.connect.network.IApiCallback;
 import org.commcare.dalvik.R;
 import org.commcare.fragments.connect.ConnectDownloadingFragment;
 import org.commcare.google.services.analytics.FirebaseAnalyticsUtil;
 import org.commcare.tasks.ResourceEngineListener;
+import org.commcare.util.LogTypes;
```
OrangeAndGreen, line 18: Unused, delete

app/src/org/commcare/activities/connect/ConnectActivity.java
```diff
@@ -69,22 +89,82 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
...
+     * Returns the fragment ID based on the redirection action.
+     * <p>
+     * This method determines which fragment should be displayed based on the value of the redirectionAction.
+     * It maps specific actions to their corresponding fragment IDs.
+     *
+     * @return The ID of the fragment to be displayed.
+     */
+    private int getFragmentId() {
+        int fragmentId;
+        if (redirectionAction.equals("ccc_opportunity_summary_page")) {
```
OrangeAndGreen, line 114: Let's make these strings private static final strings defined for the class (so we only have to type them out once).

app/src/org/commcare/services/CommCareFirebaseMessagingService.java
```diff
@@ -28,81 +29,95 @@
...
+        String notificationTitle = payloadData.get("title");
+        String notificationText = payloadData.get("body");
         NotificationManager mNM = (NotificationManager) getSystemService(NOTIFICATION_SERVICE);
 
-        Intent i = new Intent(this, DispatchActivity.class);
-        i.setAction(Intent.ACTION_MAIN);
-        i.addCategory(Intent.CATEGORY_LAUNCHER);
+        Intent intent;
+
+        if (payloadData.containsKey("action")) {
```
OrangeAndGreen, line 95: Will this code behave correctly for the existing SYNC action that was sent in existing CommCare, i.e. not related to Connect? Need to make sure we aren't passing the SYNC action on to Connect instead of using the default path

app/src/org/commcare/activities/connect/ConnectActivity.java
```diff
@@ -147,4 +232,47 @@ public void startAppValidation() {
...
+    public void getJobDetails() {
+        ApiConnect.getConnectOpportunities(ConnectActivity.this, new IApiCallback() {
+            @Override
+            public void processSuccess(int responseCode, InputStream responseData) {
+                try {
+                    String responseAsString = new String(StreamsUtil.inputStreamToByteArray(responseData));
+                    if (responseAsString.length() > 0) {
+                        //Parse the JSON
+                        JSONArray json = new JSONArray(responseAsString);
+                        List<ConnectJobRecord> jobs = new ArrayList<>(json.length());
```
OrangeAndGreen, line 245: We should save the new jobs list to the DB since we just got the most updated info. See ConnectDatabaseHelper.storeJobs, you can set pruneMissing=true in this case.

app/src/org/commcare/activities/connect/ConnectActivity.java
```diff
@@ -69,22 +89,82 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
                     .setPopUpTo(navController.getGraph().getStartDestinationId(), true)
                     .build();
             navController.navigate(fragmentId, bundle, options);
+        } else {
```
OrangeAndGreen, line 92: Nit: this can just be:
} else if (redirectionAction != null) {

app/src/org/commcare/activities/connect/ConnectActivity.java
```diff
@@ -147,4 +232,47 @@ public void startAppValidation() {
...
+                        setFragmentRedirection();
+                    }
+                } catch (IOException | JSONException | ParseException e) {
+                    Logger.exception("Parsing return from Opportunities request", e);
+                }
+            }
+
+            @Override
+            public void processFailure(int responseCode, IOException e) {
+                Logger.log("ERROR", String.format(Locale.getDefault(), "Opportunities call failed: %d", responseCode));
```
OrangeAndGreen, line 264: What does the user see when this happens? Thinking they should be taken to the Connect jobs list and somehow shown a message saying there was a problem talking to the server

app/src/org/commcare/activities/connect/ConnectActivity.java
```diff
@@ -69,22 +89,82 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
...
+     */
+    private void setFragmentRedirection() {
+        int fragmentId = getFragmentId();
+
+        boolean buttons = getIntent().getBooleanExtra("buttons", true);
+        Bundle bundle = new Bundle();
+        bundle.putBoolean("showLaunch", buttons);
+
+        // Set the tab position in the bundle based on the redirection action
+        if (redirectionAction.equals("ccc_delivery_progress")) {
```
OrangeAndGreen, line 143: Reference new constant strings here (see previous comment)

OrangeAndGreen (commented):
I didn't get a chance to setup my debug code so I could test this out today, but reviewed the code and left some comments

app/src/org/commcare/activities/DispatchActivity.java
j13panchal: Removed

app/src/org/commcare/activities/connect/ConnectActivity.java
j13panchal: Removed

app/src/org/commcare/activities/connect/ConnectActivity.java
j13panchal: Done

app/src/org/commcare/activities/connect/ConnectActivity.java
j13panchal: Done

app/src/org/commcare/activities/connect/ConnectActivity.java
j13panchal: Done

app/src/org/commcare/services/CommCareFirebaseMessagingService.java
j13panchal: Done

app/src/org/commcare/utils/FirebaseMessagingUtil.java
j13panchal: Done

app/src/org/commcare/services/CommCareFirebaseMessagingService.java
j13panchal: Removed else if condition. looks like that execute same code for both else if and else

app/src/org/commcare/activities/connect/ConnectActivity.java
j13panchal: Done

app/src/org/commcare/activities/connect/ConnectActivity.java
j13panchal: I set a flag to check if an API call is successful. If the flag indicates success, redirection to targeted fragment occurs; otherwise, it does not. 

Additionally, I display a toast message to indicate errors.

Let me know if you would like any changes.

app/src/org/commcare/services/CommCareFirebaseMessagingService.java
j13panchal: It should work because I didn't change the existing sync logic. I just added a condition: if the payload contains the "action" keyword, it redirects to **ConnectActivity**. Otherwise, it redirects to **DispatchActivity**, as previously set.

app/src/org/commcare/services/CommCareFirebaseMessagingService.java
OrangeAndGreen: Hmm, that doesn't look right. The else-if passed FLAG_IMMUTABLE while the else passed 0, do those result in the same behavior?

But my comment was more about not repeating the code that calls getActivity. Since the only thing that changes in the various if-else cases is the flags, let's just choose the flags in the if-structure, then call PendingIntent.getActivity afterward

j13panchal: is this looks correct ?

`int flags = Build.VERSION.SDK_INT >= Build.VERSION_CODES.S
                ? PendingIntent.FLAG_UPDATE_CURRENT | PendingIntent.FLAG_IMMUTABLE | PendingIntent.FLAG_ONE_SHOT
                : PendingIntent.FLAG_UPDATE_CURRENT | PendingIntent.FLAG_ONE_SHOT;

        PendingIntent contentIntent = PendingIntent.getActivity(this, 0, intent, flags);`

app/src/org/commcare/services/CommCareFirebaseMessagingService.java
OrangeAndGreen: I'm still worried that the SYNC command also uses the 'action' field. Have a look at the constructor for FCMMessageData, which is called just after this function (see lines 62-64). It retrieves this.action from the 'action' field in the payloadData. So wouldn't we then incorrectly route this notification to the ConnectActivity?

app/src/org/commcare/activities/connect/ConnectActivity.java
OrangeAndGreen: Oops! Close, but the current code stores the "jobs" list while it's still blank! The call to storeJobs() needs to happen after the for loop that parsed the JSON into the list

j13panchal: Okay I will set storeJobs() after for the loop

app/src/org/commcare/activities/connect/ConnectActivity.java
OrangeAndGreen: I'll need to test this out to see what the user experience is like. Okay to leave for now.

OrangeAndGreen (commented):
Updated the existing comments with a few additional questions and tweaks.

app/src/org/commcare/activities/connect/ConnectActivity.java
j13panchal: Okay

OrangeAndGreen (approved):

PR #2794, opened 2024-07-23
CCCT-415 Stage 1

pm-dimagi:
CCCT-415

PhoneIdActivity to Fragment

## Safety Assurance


app/src/org/commcare/activities/connectId/fragments/PhoneFragment.java
OrangeAndGreen, line 1: Let's create these new fragments next to src/org/commcare/fragments/connect in a new folder called connectid.
Also, name the fragments with the ConnectId prefix, i.e. ConnectIdPhoneFragment.

app/src/org/commcare/activities/connectId/fragments/PhoneFragment.java
```diff
@@ -0,0 +1,315 @@
...
+        messageTextView = view.findViewById(R.id.connect_primary_phone_message);
+        countryCodeInput = view.findViewById(R.id.connect_primary_phone_country_input);
+        phoneInput = view.findViewById(R.id.connect_primary_phone_input);
+        availabilityTextView = view.findViewById(R.id.connect_primary_phone_availability);
+        button = view.findViewById(R.id.connect_primary_phone_button);
+        countryCodeInput.addTextChangedListener(watcher);
+        phoneInput.addTextChangedListener(watcher);
+        button.setOnClickListener(v -> handleButtonPress());
+        requireActivity().setTitle(getString(R.string.connect_phone_page_title));
+//        method = getIntent().getStringExtra(ConnectConstants.METHOD);
```
OrangeAndGreen, line 98: We'll still need these two variables (method and existingPhone) that get passed in by the caller. Perhaps the best way to handle this for now is to pass them in to the newInstance() function?

In the eventual Navigation implementation, this data gets passed in via SafeArgs. See ConnectDownloadingFragment.onCreate for an example.

app/src/org/commcare/activities/connectId/fragments/PhoneFragment.java
```diff
@@ -0,0 +1,315 @@
...
+    @Override
+    public void onActivityResult(int requestCode, int resultCode, Intent data) {
+        super.onActivityResult(requestCode, resultCode, data);
+        String phone = PhoneNumberHelper.handlePhoneNumberPickerResult(requestCode, resultCode, data, getActivity());
+        skipPhoneNumberCheck = false;
+        displayNumber(phone);
+    }
+
+    public void finish(boolean success, String phone) {
+//        Intent intent = new Intent(getIntent());
```
OrangeAndGreen, line 143: Similar to data being passed in, we need a way to indicate that this page is done and pass any data on to the next step. Eventually this will be done via Navigation/SafeArgs. For now, maybe the caller should pass in a callback function so it can handle the next step?

This may require more thought...

OrangeAndGreen (commented):
This is a good start, see comments for a few spots to address.

Per [this guide](https://developer.android.com/guide/navigation/migrate), I think a good next step is to create the single-fragment navigation graph that uses this fragment (call it nav_graph_connectid_phone). Then change ConnectIdPhoneActivity to use the new nav graph and fragment.

Completing the above will help solve a couple current issues involving how data gets passed in and out of the fragment.

app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -469,13 +470,15 @@ private void setResultAndFinish(boolean goToJobInfo) {
...
-        ConnectManager.unlockConnect(this, success -> {
-            if(success) {
-                ConnectManager.goToConnectJobsList();
-                setResult(RESULT_OK);
-                finish();
-            }
-        });
+        Intent intent=new Intent(this, ConnectIdActivity.class);
+        startActivity(intent);
+//        ConnectManager.unlockConnect(this, success -> {
```
OrangeAndGreen, line 475: Delete the old code

app/src/org/commcare/activities/connect/ConnectActivity.java
```diff
@@ -75,8 +75,8 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
 
         destinationListener = FirebaseAnalyticsUtil.getDestinationChangeListener();
 
-        NavHostFragment host = (NavHostFragment) getSupportFragmentManager().findFragmentById(R.id.nav_host_fragment_connect);
-        navController = host.getNavController();
+        NavHostFragment host1 = (NavHostFragment) getSupportFragmentManager().findFragmentById(R.id.nav_host_fragment_connect);
```
OrangeAndGreen, line 78: No need to rename this variable

app/src/org/commcare/activities/connect/ConnectIdActivity.java
```diff
@@ -0,0 +1,24 @@
...
+public class ConnectIdActivity extends AppCompatActivity {
+    NavController controller;
+
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_connect_id);
+        NavHostFragment host2 = (NavHostFragment) getSupportFragmentManager().findFragmentById(R.id.nav_host_fragment_connectid);
+        assert host2 != null;
```
OrangeAndGreen, line 20: Remove the assertion, we'll crash on the next line anyway

app/src/org/commcare/connect/ConnectConstants.java
```diff
@@ -8,8 +8,15 @@
 public class ConnectConstants {
     public static final int ConnectIdTaskIdOffset = 1000;
 
-    public static int CREDENTIAL_PICKER_REQUEST = 2000;
+    public final  static int CREDENTIAL_PICKER_REQUEST = 2000;
     public static final String METHOD = "METHOD";
+    public static  String recoverPhone ;
```
OrangeAndGreen, line 13: We only want constants (i.e. static final) in this class. These fields should move to ConnectIdActivity ideally or ConnectManager if not.

app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -469,13 +470,15 @@ private void setResultAndFinish(boolean goToJobInfo) {
...
     public void handleConnectButtonPress() {
         selectedAppIndex = -1;
-        ConnectManager.unlockConnect(this, success -> {
-            if(success) {
-                ConnectManager.goToConnectJobsList();
-                setResult(RESULT_OK);
-                finish();
-            }
-        });
+        Intent intent=new Intent(this, ConnectIdActivity.class);
```
OrangeAndGreen, line 473: I don't think this is the right place to launch this activity. When the user presses the Connect button (which calls this function), they already have their ConnectID account configured so they just need to unlock and then proceed to the jobs page (as accomplished by the existing code).

The ConnectIdActivity should be launched when the option is selected from the three-dot menu. Note the same menu option exists in the CommCareSetupActivity.

app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -469,13 +470,15 @@ private void setResultAndFinish(boolean goToJobInfo) {
...
+        startActivity(intent);
```
OrangeAndGreen, line 474: Also, let's move the launcher into a function in ConnectManager like launchConnectId(). That way the LoginActivity and CommCareSetupActivity don't have to copy the code, and we'll have a single entry point for the activity.

app/src/org/commcare/connect/ConnectIdWorkflows.java
```diff
@@ -573,7 +573,7 @@ private static ConnectTask completeUnlock() {
         return ConnectTask.CONNECT_NO_ACTIVITY;
     }
 
-    private static void completeSignIn() {
+    public static void completeSignIn() {
```
OrangeAndGreen, line 576: This class is going away, so calling the callback will have to happen from somewhere else.

app/src/org/commcare/connect/ConnectManager.java
```diff
@@ -109,7 +114,7 @@ public interface ConnectActivityCompleteListener {
     }
 
     private static ConnectManager manager = null;
-    private ConnectIdStatus connectStatus = ConnectIdStatus.NotIntroduced;
+    public static ConnectIdStatus connectStatus = ConnectIdStatus.NotIntroduced;
```
OrangeAndGreen, line 117: This should stay as a member of the class. If we need it accessible from elsewhere, let's create:
public static ConnectIdStatus getStatus() {
   return getInstance().status;
}

app/src/org/commcare/fragments/connectId/ConnectIdMessageFragment.java
OrangeAndGreen, line 1: Something to consider and we can discuss more: Would it make the code and navigation graph cleaner to create four separate fragments (one for each message) instead of reusing a single one? Then we could setup the text for each fragment directly in the XML, only include a second button when appropriate, and know where to go next when the user presses the button. And the nav graph would definitely look less messy!

OrangeAndGreen (commented):
Great steps in the right direction! I left some initial comments. A big thing right now is the entry points and return when the activity finishes.

I think the key here is to define a couple entry points in ConnectManager:
-handleRegisterMenuSelection (launches ConnectIdActivity)
-handleConnectButtonPress (launches ConnectActivity)

These correspond to the two UI elements related to ConnectID and Connect in the two relevant UI pages (LoginActivity and CommCareSetupActivity). Both pages contain the menu option as well as Connect button, so the two functions above will each have two callers.

PR #2797, opened 2024-07-31
Remove test credentials from release build

avazirna:
## Summary
Our release `BuildConfig` class contains test credentials that are used by instrumentation tests. This PR makes these credentials only available to the Instrumentation tests `BuildConfig` class. 

After this change, here are the fields available in the release `BuildConfig` class
```
# static fields
.field public static final ANALYTICS_TRACKING_ID_DEV:Ljava/lang/String; = ""
.field public static final ANALYTICS_TRACKING_ID_LIVE:Ljava/lang/String; = ""
.field public static final APPLICATION_ID:Ljava/lang/String; = "org.commcare.dalvik"
.field public static final BUILD_DATE:Ljava/lang/String; = "2024-08-13"
.field public static final BUILD_NUMBER:Ljava/lang/String; = "..."
.field public static final BUILD_TYPE:Ljava/lang/String; = "release"
.field public static final CCC_HOST:Ljava/lang/String; = "connect.dimagi.com"
.field public static final CC_AUTHORITY:Ljava/lang/String; = "org.commcare.dalvik"
.field public static final CONSUMER_APP_PASSWORD:Ljava/lang/String; = ""
.field public static final CONSUMER_APP_USERNAME:Ljava/lang/String; = ""
.field public static final DEBUG:Z = false
.field public static final FIREBASE_DATABASE_URL:Ljava/lang/String; = "..."
.field public static final FLAVOR:Ljava/lang/String; = "commcare"
.field public static final IS_CONSUMER_APP:Z = false
.field public static final IS_SINGLE_APP_BUILD:Z = false
.field public static final MAPBOX_SDK_API_KEY:Ljava/lang/String; = "..."
.field public static final ODK_AUTHORITY:Ljava/lang/String; = "org.commcare.android.provider.odk"
.field public static final TRUSTED_SOURCE_PUBLIC_KEY:Ljava/lang/String; = "xxx"
.field public static final USE_CRASHLYTICS:Z = true
.field public static final VERSION_CODE:I = 0x72102
.field public static final VERSION_NAME:Ljava/lang/String; = "2.54"
```

## Safety Assurance

### Safety story
This is only relevant to the build process.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -556,7 +556,7 @@ tasks.whenTaskAdded { task ->
             && "true" == isConsumerApp && "true" == runDownloadScripts) {
         task.dependsOn downloadRestoreFile
     }
-    if ((task.name != null) && (task.name.startsWith('assemble') && task.name.endsWith('AndroidTest'))) {
+    if (task.name == 'assembleCommcareReleaseAndroidTest') {
         android.defaultConfig.buildConfigField "String", "HQ_API_USERNAME", "\"${project.ext.HQ_API_USERNAME}\""
         android.defaultConfig.buildConfigField "String", "HQ_API_PASSWORD", "\"${project.ext.HQ_API_PASSWORD}\""
     }
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -276,8 +276,6 @@ android {
         buildConfigField 'String', 'ANALYTICS_TRACKING_ID_DEV', "\"${project.ext.ANALYTICS_TRACKING_ID_DEV}\""
         buildConfigField 'String', 'MAPBOX_SDK_API_KEY', "\"${project.ext.MAPBOX_SDK_API_KEY}\""
 
-        buildConfigField "String", "HQ_API_USERNAME", "\"${project.ext.HQ_API_USERNAME}\""
-        buildConfigField "String", "HQ_API_PASSWORD", "\"${project.ext.HQ_API_PASSWORD}\""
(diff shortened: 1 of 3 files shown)


avazirna:

avazirna:

app/build.gradle
```diff
@@ -558,6 +556,10 @@ tasks.whenTaskAdded { task ->
             && "true" == isConsumerApp && "true" == runDownloadScripts) {
         task.dependsOn downloadRestoreFile
     }
+    if ((task.name != null) && (task.name.startsWith('assemble') && task.name.endsWith('AndroidTest'))) {
```
shubham1g5, line 559: can't we use exact task name here ?

avazirna: because of the different `flavours` and build types we have a few combinations:
image

shubham1g5: but think we only want to do this for `assebleCommcareReleaseAndroidTest` as that's the apk that gets passed to Browserstack ultimately.

avazirna: True and also, it seems that LTS is not relevant anymore. But what about CCConnect?

shubham1g5: LTS is still relevant though we have not lately pushed an update there (blocked on Android OS update). Though other flavors don't matter as we do all instrumentation testing on the main production flavour.

shubham1g5 (approved):

PR #2799, opened 2024-08-01
Fixes Crash on App Resume after Session Expiration

shubham1g5:
## Summary

- Handles SessionExpiration in the home button view setup, this escapes the default session expiration and causes the app to crash when resuming after a session expiration event.  Other buttons already have this safety implemented and sync was the only one remaining. 

- Adds some helpful logs around session expiration

## Safety Assurance


shubham1g5:

avazirna:
@shubham1g5 this looks right to me. I wonder if we should consider adding this to `2.54`

shubham1g5:
>  I wonder if we should consider adding this to 2.54

Yeah that sounds good to me.

shubham1g5:
Closing in favour of https://github.com/dimagi/commcare-android/pull/2814

PR #2802, opened 2024-08-06
Auto-Save Form on Pause

shubham1g5:
## Summary

[Context Doc](https://docs.google.com/document/d/1j5ume0J0UwwraijxGw-Pq-8ZYFC7L07cKb7CXL30poo/edit#heading=h.ferq7q7l8jai)

## Feature Flag
None at the moment, will be behind a custom property when merged to `master`

## Product Description

Auto-Saves a CC form when CC goes in background

## Safety Assurance

### Automated test coverage

None

### Safety story
 
We call an existing method we already use to save form session at the time of user session expiry

cross-request: https://github.com/dimagi/commcare-core/pull/1426


app/src/org/commcare/logic/AndroidFormController.java
```diff
@@ -88,4 +86,14 @@ public FormAnalyticsHelper getFormAnalyticsHelper() {
     public FormDef getFormDef() {
         return mFormEntryController.getModel().getForm();
     }
+
+    // TODO: we should cache this
+    public String getSerializedFormIndex() {
+        try{
+            byte[] serializedFormIndex = SerializationUtil.serialize(getFormIndex());
+            return Base64.encodeToString(serializedFormIndex, Base64.DEFAULT);
```
shubham1g5, line 94: can the value for `Base64.DEFAULT` change with OS updates ? If so, I would like to pass a concrete encoding here like `Charset.forName("UTF-8")`

app/src/org/commcare/logic/AndroidFormController.java
```diff
@@ -88,4 +86,14 @@ public FormAnalyticsHelper getFormAnalyticsHelper() {
...
+        } catch (Exception e){
+            return null;
```
shubham1g5, line 96: Think it's better to fail fast here instead of swallowing the errors here unless there is a strong reason to do otherwise. At the very least we should be logging the exceptions encountered here.

shubham1g5: Actually think again it makes sense to catch any errors here as we would not want to loose on saving form because of an error here.

app/src/org/commcare/logic/AndroidFormController.java
avazirna: That's a good point, I'm not sure, let's lean to the safe side (but I will take note and look into that later)

PR #2806, opened 2024-08-07
Load last Form Index when auto resuming session

avazirna:
## Summary
Load last Form Index when auto resuming session.

## Safety Assurance


app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -1112,12 +1116,17 @@ private void handleFormLoadCompletion(AndroidFormController fc) {
...
+        mFormController.returnToStoredIndex();
+
         uiController.refreshView();
         FormNavigationUI.updateNavigationCues(this, mFormController, uiController.questionsView);
         if (isRestartAfterSessionExpiration) {
             Toast.makeText(this,
                     Localization.get("form.entry.restart.after.expiration"), Toast.LENGTH_LONG).show();
         }
+
+        HiddenPreferences.clearInterruptedFormIndex();
```
ctsims, line 1129: So I'm concerned a bit here about the atomicity and idempotence of how this field is being managed.

It can't be perfect, because of the difference in how the data is stored, but I'm concerned that there are possibilities where this field is set and the other model is removed, and future apps have issues not with the Form Index being null, but with it being misaligned with the form being loaded. I'm also a little concerned that whether the other incomplete form is wiped on recovery could change and because this code isn't close to the recovery/load event it won't be clear that we should change this. 

There are a few ways to potentially improve this, but I think the most straightforward is to stick to the same structure of this PR, but to add one more specific piece, which is adding the form ID (ideally the GUID, but that may not be available easily, so a database ID or some other convenient secondary signal) itself to the item we put into shared preferences, confirming that the form ID matches the recovered form before applying it, and performing this wipe if there's a mismatch to prevent potential problems. 

I'm also not sure if there's an easy or clear way to ensure that the stored FormIndex is valid for the current form, in case something changed in the background, or the form structure is odd somehow (say, there's a section that disappears based on the date), but I'm a little uneasy without the ability to either apply the form index as a wrapped call, or somehow check it before applying the nav.

ctsims (changes requested):
This approach looks 90% of the way there to me, with one or two minor points to make sure it's safely compartmentalized.

app/src/org/commcare/logic/AndroidFormController.java
```diff
@@ -96,4 +99,17 @@ public String getSerializedFormIndex() {
...
         }
     }
+
+    private FormIndex deserializeFormIndex(String serializedFormIndex) {
+        if (serializedFormIndex != null) {
+            try{
+                byte[] decodedFormIndex = Base64.decode(serializedFormIndex, Base64.DEFAULT);
+                return SerializationUtil.deserialize(decodedFormIndex, FormIndex.class);
+            } catch(Exception e) {
+                Logger.log(LogTypes.TYPE_FORM_ENTRY,
```
shubham1g5, line 109: should log with `logger.exception` to get the full stacktrace here.

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -1024,6 +1025,9 @@ private void loadForm() {
                             instanceState);
                     formId = instanceAndStatus.first;
                     instanceIsReadOnly = instanceAndStatus.second;
+
+                    // only retrieve a potentially stored form index when loading an existing form record
+                    serializedFormIndex = HiddenPreferences.getInterruptedFormIndex();
```
shubham1g5, line 1030: nit: probably cleaner to directly access it in `AndroidFormController` instead of passing it through the loader task to controller ?

ctsims: I think there are somewhat competing design constraints. Ideally, I think it's best for Android OS lifecycle type code to be isolated into Activities, or other clearly-coupled-to-the-OS locations, so it does make sense to me that the root entry point for actually loading the form index should be in the activity. 

In thinking about where to put these call (and what format to put them in) my two main priorities would be:

* I think it's good for the retrieval code, the storage code, and the serialization/deserialization to all be as compartmentalized as possible so there's little ambiguity about whether a future change has completely covered them
* I think we should try and minimize state uncertainty. As written one big potential point of confusion is that there's lots of uncertainty about this return object. Does it exist at all? Is it serialized correctly? Is it valid for the current form? It worries me that uncertainty is passed through later to the loader task, since it increases the number of touchpoints where a reader can't be certain about what might happen with it.

app/src/org/commcare/preferences/HiddenPreferences.java
```diff
@@ -637,4 +637,17 @@ public static void setInterruptedFormIndex(String serializedFormIndex) {
                 .putString(INTERRUPTED_FORM_INDEX + currentUserId,  serializedFormIndex)
                 .apply();
     }
+
+    public static String getInterruptedFormIndex() {
+        String currentUserId = CommCareApplication.instance().getCurrentUserId();
```
ctsims, line 642: Nit, but if available this should probably also be keyed to the App ID, since a user's ID is shared between two apps in the same project space.

app/src/org/commcare/activities/FormEntryActivity.java
avazirna: > there are possibilities where this field is set and the other model is removed, and future apps have issues not with the Form Index being null, but with it being misaligned with the form being loaded.
I circled a bit around how much decoupled these should be, envisioning a scenario where they could exist independently. But yeah, there is no point in having a form index that is not directly associated with a session state, so that was me overthinking. One potential candidate ID that could allow some verification during form loading is the SSD record ID.

ctsims: Yeah, SSD record ID is totally fine if that replicates in an identical way before/after, even just the database ID of the incomplete forms row.  I think whatever's convenient and readily available in both the save and load context is great. Just looking for a gut check on the system knowing the FormIndex is actually intended for the recovery context.

app/src/org/commcare/preferences/HiddenPreferences.java
avazirna: About this, is it still necessary to have the App ID if we are validating the FormIndex based on the SSD Id? Is there any other reason to still have it there?

app/src/org/commcare/preferences/HiddenPreferences.java
```diff
@@ -631,10 +631,23 @@ public static boolean isBackgroundSyncEnabled() {
         return DeveloperPreferences.doesPropertyMatch(ENABLE_BACKGROUND_SYNC, PrefValues.NO, PrefValues.YES);
     }
 
-    public static void setInterruptedFormIndex(String serializedFormIndex) {
+    public static void setInterruptedFormIndex(int ssdId, String serializedFormIndex) {
         String currentUserId = CommCareApplication.instance().getCurrentUserId();
         CommCareApplication.instance().getCurrentApp().getAppPreferences().edit()
-                .putString(INTERRUPTED_FORM_INDEX + currentUserId,  serializedFormIndex)
+                .putString(INTERRUPTED_FORM_INDEX + currentUserId, ssdId + " " + serializedFormIndex)
```
shubham1g5, line 637: think it's better to have `ssdId` as part of key rather than value. It will simplify code while retrieving the form index as well.

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -1083,6 +1088,31 @@ protected void deliverError(FormEntryActivity receiver, Exception e) {
...
+                                spaceIndex + 1, interruptedFormIndex.length());
+                if (ssdId == sessionDescriptorId) {
+                    return serializedFormIndex;
+                }
+            } catch (Exception e) {
+                // something went wrong
+            }
+        }
+        // data format is invalid, so better to clear the data
+        HiddenPreferences.clearInterruptedFormIndex();
```
shubham1g5, line 1112: we will also reach here when `ssdId == sessionDescriptorId` eval to false which is not a case of invalid data format right ?

app/src/org/commcare/logic/AndroidFormController.java
```diff
@@ -96,4 +99,17 @@ public String getSerializedFormIndex() {
             return null;
         }
     }
+
+    private FormIndex deserializeFormIndex(String serializedFormIndex) {
```
shubham1g5, line 103: can we perform the deserialization at the same time when we retrive the form index in `FormEntryActivity`. Today the logic to retrive the saved value of the index seems to be split into multiple place which is not ideal.

shubham1g5 (changes requested):

app/src/org/commcare/activities/FormEntryActivity.java
avazirna: Yeah, this is supposed to be short-lived, so if the `ssdIDs` don't match we can just clear it out, but happy to keep it there

app/src/org/commcare/preferences/HiddenPreferences.java
shubham1g5: That's a good point.

app/src/org/commcare/activities/FormEntryActivity.java
shubham1g5: agree that it makes sense to clear it.

app/src/org/commcare/logic/AndroidFormController.java
avazirna: Very true, I wanted to keep the `serialization` and `deserialization` logic together, but this is method doesn't rely on the state of the class

app/src/org/commcare/utils/StringUtils.java
```diff
@@ -71,4 +66,19 @@ public static String convertPairToJsonString(Pair<? extends Serializable, ? exte
...
+        Gson gson = new Gson();
+        if (stringInJsonFormat == null) {
+            return null;
+        }
+
+        try{
+            return gson.fromJson(stringInJsonFormat, Pair.class);
+        } catch(JsonIOException e){
+            // default to null
+            return null;
```
shubham1g5, line 81: we should log all catched exceptions here, similar comment for pair to JSON conversion method.

shubham1g5:
Merging to unblock QA, but happy to have more discussions here.

app/src/org/commcare/preferences/HiddenPreferences.java
ctsims: I think this should still be keyed on the app ID. The SSD Id is likely to have collisions on its own, since it's just a DB row id, right?

shubham1g5: I would think we should use form instance id , app id doesn't make a difference here as the preferences we are writing to are app specific.

avazirna: >  app id doesn't make a difference here as the preferences we are writing to are app specific.

Good point, taking that into considering plus that we only load the interrupted form index when recovering an expired session, which happens right after the next login, our current check seems to be sufficient. I have tested it with a few scenarios, such as logging back in an app from the same domain and with the same user, with form linking, and worked fine. Let me know if there are other scenarios we should be concerned about.  

As per the form instance Id, I think this it the `form record id` (right, @shubham1g5 ?), we can always add it to strengthen the validation, this is very low loe as the `SessionStateDescriptor` also stores the form record Id.

shubham1g5: >  we can always add it to strengthen the validation, this is very low loe as the SessionStateDescriptor also stores the form record Id.

@avazirna think the idea here is to make this collision resistant. Even though `SessionStateDescriptor` stores form record Id, here we are only storing the database id of the DB row for the corresponding  `SessionStateDescriptor`. This will be an Integer number and is not very collision resistive in case something goes wrong that lets say clear the `SessionStateDescriptor` table and cause it to reuse the integer ids. In contrast form instance ids are unique uuids which are much less likely to collide in context of a single app.

PR #2810, opened 2024-08-09
Minor cleanup after adding InterruptedFormState model

avazirna:
## Summary
This is to remove some unused methods and outdated comments. In addition to that, a small change to when the stored form index is cleared, this is now at the top of the [handledFormLoadCompletion(...)](https://github.com/dimagi/commcare-android/blob/755913bfaf583c2960c36bf5d5de1d457c2d7c71/app/src/org/commcare/activities/FormEntryActivity.java#L1124) method for visibility.

## Safety Assurance

### DIFF
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -1107,6 +1107,8 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
     }
 
     private void handleFormLoadCompletion(AndroidFormController fc) {
+        HiddenPreferences.clearInterruptedFormState();
+
         if (PollSensorAction.XPATH_ERROR_ACTION.equals(locationRecieverErrorAction)) {
             handleXpathErrorBroadcast();
         }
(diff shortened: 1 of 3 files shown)


app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -1109,19 +1106,9 @@ private FormIndex retrieveAndValidateFormIndex(int sessionDescriptorId) {
...
-                return SerializationUtil.deserialize(decodedFormIndex, FormIndex.class);
-            } catch(Exception e) {
-                Logger.exception("Deserialization of last form index failed ", e);
-            }
-        }
-        return null;
-    }
-
     private void handleFormLoadCompletion(AndroidFormController fc) {
+        HiddenPreferences.clearInterruptedFormState();
```
shubham1g5, line 1110: is there any bheavior change caused by the change in position for this call ?

shubham1g5 (approved):

app/src/org/commcare/activities/FormEntryActivity.java
avazirna: Not necessarily, more for visibility. Initially though, I thought that this call could be prevented by [this return](https://github.com/dimagi/commcare-android/blob/f87f6e4863c6f868cfdf4adda7b25cc34285cfc7/app/src/org/commcare/activities/FormEntryActivity.java#L1146), which is triggered when saved forms are loaded, but because we restore the session immediately after the login, it seems unlikely.

avazirna:

PR #2814, opened 2024-08-14
Fixes Crash on App Resume after Session Expiration + Logs

shubham1g5:
## Summary

- Handles SessionExpiration in the home button view setup, this escapes the default session expiration and causes the app to crash when resuming after a session expiration event.  Other buttons already have this safety implemented and sync was the only one remaining. 

- Adds some helpful logs around session expiration

## Safety Assurance


avazirna (approved):
@shubham1g5 this looks good to me. But it seems that we might need to send this to QA, what do you think?

shubham1g5:
I don't see a need to QA this as it just adds a try-catch block. Are there specific concerns you are thinking of ?

avazirna:
> I don't see a need to QA this as it just adds a try-catch block. Are there specific concerns you are thinking of ?

No concern in particular, I'm happy to follow your lead on this.

PR #2820, opened 2024-08-19
Copy connect v7 global database upgrades

avazirna:
## Summary
This PR is to copy Global DB changes that were done as part of the CCC work. The goal with bringing these changes is to ensure a healthy update from that version to production. Changes included here are:
- Added is_local column to ConnectKeyRecord table in global DB

## Safety Assurance

### Safety story
These changes have already been released in the beta stream, the risk is in the update itself and there are plans to test that update before releasing this to production users.

### DIFF
--- a/app/src/org/commcare/models/database/global/GlobalDatabaseUpgrader.java
+++ b/app/src/org/commcare/models/database/global/GlobalDatabaseUpgrader.java
@@ -132,7 +132,7 @@ class GlobalDatabaseUpgrader {
     }
 
     private boolean upgradeFiveSix(SQLiteDatabase db) {
-        return addTableForNewModel(db, ConnectKeyRecord.STORAGE_KEY, new ConnectKeyRecord());
+        return addTableForNewModel(db, ConnectKeyRecord.STORAGE_KEY, new ConnectKeyRecordV6());
     }
 
     private boolean upgradeSixSeven(SQLiteDatabase db) {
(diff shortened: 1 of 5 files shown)


app/src/org/commcare/models/database/global/GlobalDatabaseUpgrader.java
```diff
@@ -124,7 +132,44 @@ private boolean upgradeFourFive(SQLiteDatabase db) {
...
+                ConnectKeyRecord newRecord = ConnectKeyRecord.fromV6(oldRecord);
+                //set this new record to have same ID as the old one
+                newRecord.setID(oldRecord.getID());
+                newStorage.write(newRecord);
+            }
+
+            db.setTransactionSuccessful();
+            return true;
+        } catch(Exception e) {
+            return false;
```
shubham1g5, line 169: think we should not be catching exceptions here.

shubham1g5 (changes requested):

app/src/org/commcare/models/database/global/GlobalDatabaseUpgrader.java
avazirna: True. @OrangeAndGreen any particular reason for not letting this bubble up? We might not be able to pick this up if not.

shubham1g5: Think it was a mistake and Dave is planning to remove it from Connect branch as well

OrangeAndGreen: Yeah, Shubham is correct, we shouldn't catch the exception and should let it bubble up instead.

shubham1g5 (approved):

PR #2822, opened 2024-08-22
Badge size to correctly display Overflow of characters

shubham1g5:
## Summary

https://dimagi.atlassian.net/browse/SC-3315

## Feature Flag

Custom Icon Badges for modules and forms 

## Product Description

Correctly display an overflow of text for Badge View. 

## Safety Assurance

### Automated test coverage

Feature has coverage around normal use case in MenuBadgeTests without testing for overflow of characters. I can add it but feels like an overdo for a small thing. Happy to reconsider. 

### Safety story

tested locally.


avazirna (approved):
Looks good. I wasn't sure about why we are not letting the NFE to bubble up, but then remembered that we can have non-numerical content in badges.

PR #2828, opened 2024-08-26
Print memory info for critical level

avazirna:
## Summary
This PR adjusts the `onTrimMemory` level options that trigger an output of the state of the memory. There is a level that haven been consistently seen before app kills and with this change we expect to get a better sense of how low the available memory can get before the low memory killer is triggered.

## Safety Assurance


OrangeAndGreen (approved):

avazirna:

avazirna:

PR #2830, opened 2024-08-27
CCCT-470 || Country-Code Phone Edittext UI Component

j13panchal:
## Summary
https://dimagi.atlassian.net/browse/CCCT-470

## Safety Assurance


app/AndroidManifest.xml
OrangeAndGreen, line 1: Per feedback from the other mobile devs, we should avoid doing these large reorganizations as part of other PRs. Let's back this one up so it only has any required changes. We can also do the organizational refactor over on the master branch and create a PR for it there, then pull the change up to our branch once it's accepted in master.

OrangeAndGreen (changes requested):
See the comment about change to the manifest, but otherwise everything looks great!

app/AndroidManifest.xml
j13panchal: I revert all reorganizations code from Manifest file

OrangeAndGreen (approved):

PR #2835, opened 2024-09-05
Do not save read only forms

shubham1g5:
## Summary
Makes 2 changes in context of completed/read only forms -

1. Do not allow re-saving completed forms 
2. Reword Form's "FINISH" button to say "EXIT" for completed forms

## Safety Assurance

### Safety story

Tested locally and verified that we are able to save non-completed forms as usual.


avazirna (approved):
Looks fine to me.

PR #2836, opened 2024-09-05
Potential fix for incorrect session recovery

shubham1g5:
## Summary

This is a potential fix for some of the issues we are seeing where user gets restored to a saved session state even after saving the form as complete. Today [we clear this state after the Form Saved ](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/activities/FormEntryActivity.java#L1237) async task has been finished successfully but that may cause issues in scenarios when the activity get deattached from the async task itself.

I have not been able to reproduce these issues but the change is harmless and would love to see if that fixes the issue for users.  If not, also added some helpful logs to get more info on if we are managing the interrupted state correctly. 

## Safety Assurance

### Safety story
 
Small code change with all code pathways already in usage

### DIFF
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -839,6 +839,11 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
             return;
         }
 
+        if (complete) {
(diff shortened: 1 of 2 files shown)


avazirna (approved):
Looks good to me!

avazirna:

PR #2837, opened 2024-09-05
Prevent read only forms form being auto saved

avazirna:
## Summary
This PR prevents read only forms from being auto saved as incomplete, which would allow them to be loaded and submitted again.

## Safety Assurance

### Safety story
This only limits the scenarios where the auto-save feature is triggered.


shubham1g5:
@avazirna I already PR'ed this - https://github.com/dimagi/commcare-android/pull/2835 here, not sure if you are following slack messages.

PR #2842, opened 2024-09-10
Fixes a bug causing Updates to get stale

shubham1g5:
## Summary

https://dimagi.atlassian.net/browse/SC-3923

Bug Fix for App updates. 

Background updates were failing due to SessionNotAvailable exception due to us unneccesarily checking for user session being active in the Update Worker. This PR makes 2 important changes to eliminate this - 

1. Remove the check for user session from UpdateWrapper, we already perform user updates outside user sessions. 
2. If the error is caused by the wrapper code in UpdateWorker, we can simply result a task failure instead of handling it as an update code error and resetting the update downloads. 

## Safety Assurance

### Automated test coverage

Covered in [UpdateWorkerTests](https://github.com/dimagi/commcare-android/blob/master/app/unit-tests/src/org/commcare/update/UpdateWorkerTest.kt)

### Safety story

Mostly relying on tests coverage here plus the change not affecting any functional update code.

### DIFF
--- a/app/src/org/commcare/heartbeat/HeartbeatWorker.kt
+++ b/app/src/org/commcare/heartbeat/HeartbeatWorker.kt
@@ -29,9 +29,10 @@ class HeartbeatWorker(context: Context, workerParams: WorkerParameters):
                     Result.retry()
                 }
                 else -> {
-                    Logger.log(LogTypes.TYPE_ERROR_SERVER_COMMS,
-                            "Encountered unexpected exception during heartbeat communications: "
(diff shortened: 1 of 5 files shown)


avazirna:
I'm excited to see this fixed.

shubham1g5:
@avazirna Think we should roll this out in 2.53.1 as well with GD stuff.

shubham1g5:
Bumping this for review

shubham1g5:
PR'ed directly for hotfixing https://github.com/dimagi/commcare-android/pull/2850

PR #2846, opened 2024-09-12
Add FormInstanceId to InterruptedFormState 

avazirna:
## Summary
This PR add FormInstance Id to InterruptedFormState, reducing that way potential collisions around the SessionStateDescriptor Id.

## Safety Assurance

### Safety story
This is a simple change, just complementing an exiting model with a database Id, so risks associated are almost inexistent. The only risk is when updating from the previous version which didn't have the `formInstanceid`, but this was mitigated by defaulting `formInstanceId` to -1 when non-existent.


app/src/org/commcare/models/database/InterruptedFormState.java
```diff
@@ -33,12 +35,14 @@ public void readExternal(DataInputStream in, PrototypeFactory pf)
             throws IOException, DeserializationException {
         sessionStateDescriptorId = ExtUtil.readInt(in);
         formIndex = (FormIndex)ExtUtil.read(in, FormIndex.class, pf);
+        formRecordId = ExtUtil.readInt(in);
```
shubham1g5, line 38: this will fail when trying to recover a form which has been saved on older version of app as there will be nothing to read. Safety story on this PR should detail out the backward compatibility with this change and look into ways of having the old forms recover successfully after this is rolled out.

avazirna:

PR #2848, opened 2024-09-16
Fix an issue when unregistering Pending Sync Alerts receiver

avazirna:
## Summary
This PR fixes an issue when unregistering the Pending Sync alerts receiver, it seems that for some reason the registration of the receiver failed. Here's the stacktrace:
```
Fatal Exception: java.lang.RuntimeException: Unable to pause activity {org.commcare.dalvik/org.commcare.activities.FormEntryActivity}: java.lang.IllegalArgumentException: Receiver not registered: org.commcare.services.PendingSyncAlertBroadcastReceiver@3592b8b
       at android.app.ActivityThread.performPauseActivityIfNeeded(ActivityThread.java:5266)
       at android.app.ActivityThread.performPauseActivity(ActivityThread.java:5217)
       at android.app.ActivityThread.handlePauseActivity(ActivityThread.java:5168)
       at android.app.servertransaction.PauseActivityItem.execute(PauseActivityItem.java:46)
       at android.app.servertransaction.TransactionExecutor.executeLifecycleState(TransactionExecutor.java:176)
       at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:97)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2336)
       at android.os.Looper.loop(Looper.java:246)
       at android.app.ActivityThread.main(ActivityThread.java:8653)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:602)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1130)
```
Link to the in Crashlytics: [31900d304ee55c06bdbebdc58d102951](https://console.firebase.google.com/project/commcare-a57e4/crashlytics/app/android:org.commcare.dalvik/issues/31900d304ee55c06bdbebdc58d102951)

## Safety Assurance

- [X I have confidence that this PR will not introduce a regression for the reasons below

### DIFF
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -907,11 +907,7 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
         if (broadcastReceiver != null) {
             try {
                 unregisterReceiver(broadcastReceiver);
-            } catch (IllegalArgumentException e) {
-                // Thrown when given receiver isn't registered.
-                Logger.log(LogTypes.TYPE_ERROR_ASSERTION,
-                        "Tried to unregister a BroadcastReceiver that wasn't registered: " + e.getMessage());
-            }
+            } catch (IllegalArgumentException e) {}
         }
     }
 
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -900,7 +900,7 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
         }
         TextToSpeechConverter.INSTANCE.stop();
 
-        unregisterReceiver(pendingSyncAlertBroadcastReceiver);
+        attemptToUnregisterBroadcastReceiver(pendingSyncAlertBroadcastReceiver);
     }
 
     private void attemptToUnregisterBroadcastReceiver(BroadcastReceiver broadcastReceiver){
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -891,16 +891,7 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
             saveAnswersForCurrentScreen(false);
         }
 
-        if (mLocationServiceIssueReceiver != null) {
-            try {
-                unregisterReceiver(mLocationServiceIssueReceiver);
-            } catch (IllegalArgumentException e) {
-                // Thrown when given receiver isn't registered.
-                // This shouldn't ever happen, but seems to come up in production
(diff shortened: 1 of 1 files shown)


avazirna:

avazirna:

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -909,7 +900,19 @@ protected void onPause() {
...
+        attemptToUnregisterBroadcastReceiver(pendingSyncAlertBroadcastReceiver);
+    }
+
+    private void attemptToUnregisterBroadcastReceiver(BroadcastReceiver broadcastReceiver){
+        if (broadcastReceiver != null) {
+            try {
+                unregisterReceiver(broadcastReceiver);
+            } catch (IllegalArgumentException e) {
+                // Thrown when given receiver isn't registered.
+                Logger.log(LogTypes.TYPE_ERROR_ASSERTION,
```
shubham1g5, line 912: we should log this using `logger.exception()`

shubham1g5 (changes requested):
1. Curious if there is more info on this issue in a ticket /Firebase crash that can be linked in PR description ? 
2. Do we know the reason beahind broadcast not registering successfully ? 
3. Also why is this the only place we are safeguarding unregister calls and not other calls in the code ?

avazirna:
> 2\. Do we know the reason beahind broadcast not registering successfully ?

It's hard to say. I think maybe an error during registration that didn't bubble up, the system didn't register due to resources constraints or there are conditions causing `unregisterReceiver` to be called multiple times

avazirna:
> 3\. Also why is this the only place we are safeguarding unregister calls and not other calls in the code ?

I believe we are actually doing that in many places in the code, [here](https://github.com/dimagi/commcare-android/blob/68d250457c6753d7532a02f333fe57289c511d1f/app/src/org/commcare/utils/SessionRegistrationHelper.java#L62) and [here](https://github.com/dimagi/commcare-android/blob/68d250457c6753d7532a02f333fe57289c511d1f/app/src/org/commcare/location/CommCareProviderLocationController.kt#L58)

app/src/org/commcare/activities/FormEntryActivity.java
ctsims: I'd normally agree with this, but reviewing the [baseline docs](https://developer.android.com/reference/android/content/Context#unregisterReceiver(android.content.BroadcastReceiver)), it's clear that there are only two scenarios where this gets thrown and both involve a semantically clear outcome (the receiver was not registered before, or was already cleared). 

As such, I don't think necessary to log as an exception directly, which I think of as more of a clearing house for things we believe are (or are might be) actual errors. To me this is more like a NumberFormatException catch, which reflects a known condition using the catch callpath.

ctsims (approved):

shubham1g5 (approved):

PR #2851, opened 2024-09-17
Adjust user messages around the Session Pause feature

avazirna:
## Summary
This PR adjusts the messages around the Graceful Session Pause to be more inline with the expected user experience. 

## Feature Flag
This is behind a custom property called `cc-auto-form-save-on-pause`

## Product Description
This change entails:
- Navigating away from CommCare no longer triggers a message to the user;
- When restoring a session, the message is now to:
`CommCare was closed and the form you were in the middle of has been saved and resumed.`
## Safety Assurance

### Safety story
This needs to go through QA to make sure it doesn't impact with the Session Expiration workflow.


app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -252,17 +252,17 @@ public void onCreateSessionSafe(Bundle savedInstanceState) {
     public void formSaveCallback(Runnable listener) {
         // note that we have started saving the form
         customFormSaveCallback = listener;
-        interruptAndSaveForm(exit);
+        interruptAndSaveForm(false);
```
shubham1g5, line 255: I am confused on semantics on the new parameter `autoSaving`. Given the `formSaveCallback` is called to auto-save the form as well, should not it be set to true ?

app/src/org/commcare/models/AndroidSessionWrapper.java
```diff
@@ -185,14 +185,19 @@ private static boolean ssdHasValidFormRecordId(int ssdId,
...
                     CommCareApplication.instance().getUserStorage(SessionStateDescriptor.class);
             SessionStateDescriptor current = sessionStorage.read(sessionStateRecordId);
             InterruptedFormState interruptedFormState = new InterruptedFormState(current.getID(), formIndex);
             HiddenPreferences.setInterruptedSSD(current.getID());
-            HiddenPreferences.setInterruptedFormState(interruptedFormState);
+            if (autoSaving) {
+                HiddenPreferences.setInterruptedFormState(interruptedFormState);
+            } else {
+                HiddenPreferences.clearInterruptedFormState();
+            }
```
shubham1g5, line 199: IMO think we should not introduce any additional non-required control flows in a hotfix branch as I am afraid we will need to do a QA on this otherwise but will let you take a call on this depending on how well you have tested these changes.

shubham1g5 (changes requested):

app/src/org/commcare/logic/AndroidFormController.java
```diff
@@ -20,13 +20,17 @@ public class AndroidFormController extends FormController implements PendingCall
...
+    private boolean restoredAfterSessionPause = false;
 
     private FormAnalyticsHelper formAnalyticsHelper;
 
     public AndroidFormController(FormEntryController fec, boolean readOnly, FormIndex formIndex) {
         super(fec, readOnly);
         formAnalyticsHelper = new FormAnalyticsHelper();
         formIndexToReturnTo = formIndex;
+        if (formIndex != null) {
+            restoredAfterSessionPause = true;
```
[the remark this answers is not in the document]
avazirna: I was trying to balance minimal changes to achieve that goal but I agree that this is not ideal, let me propose another approach

app/src/org/commcare/activities/FormEntryActivity.java
avazirna: Yeah, `autoSaving` as in the `autoSessionPauseSaving`, let me come up with another term.

> Given the formSaveCallback is called to auto-save the form as well, should not it be set to true ?

Considering that the `exit` flag was always set to true for the `formSaveCallback`, I thought about replacing it with another that can convey the same data point. I will review this was well.

app/src/org/commcare/logic/AndroidFormController.java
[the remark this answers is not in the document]
avazirna: I ended up passing the InterruptFormState to the form controller [1836d2f](https://github.com/dimagi/commcare-android/pull/2858/commits/1836d2fb100af6f0cf327b9499a3a72acedef8a2). I agree with include the FormIndex when the session expires but I believe that will require more changes, which might be a bit too far from the scope of this PR. But happy to discuss this a bit more.

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -1153,7 +1153,9 @@ private void handleFormLoadCompletion(AndroidFormController fc) {
         FormNavigationUI.updateNavigationCues(this, mFormController, uiController.questionsView);
         if (isRestartAfterSessionExpiration) {
             Toast.makeText(this,
-                    Localization.get("form.entry.restart.after.expiration"), Toast.LENGTH_LONG).show();
+                    Localization.get("form.entry.restart.after."
+                                    + (fc.getRestoredAfterSessionPause()?"session.pause":"expiration")),
```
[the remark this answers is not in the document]
avazirna: [d643cca](https://github.com/dimagi/commcare-android/pull/2858/commits/d643cca9d44a8fe17e29100fc3b394ea091c9e75)

app/src/org/commcare/activities/FormEntryActivity.java
avazirna: I changed the flags, added a `sessionExpiredOrSustended` ([3023424](https://github.com/dimagi/commcare-android/pull/2858/commits/30234248cc47e11985f484f4e4d296f2211abc6b)) and `userTriggered` ([312ade0](https://github.com/dimagi/commcare-android/pull/2858/commits/312ade00bdcc22e140d5a8087a15ac7a435958c0))

PR #2852, opened 2024-09-17
Change log server URL locally

avazirna:
## Summary
This PR adds a new option to the `Developer Options` menu that allows users the set a new URL for where CommCare logs should be submitted.

Ticket: https://dimagi.atlassian.net/browse/SAAS-15980

## Product Description
Users are able to set the URL by typing it in the text box and remove it by leaving the text box blank:
<video width="200" src="https://github.com/user-attachments/assets/f167c52b-ed63-487f-a743-48d115cd9d0d"/>

## Safety Assurance

### Safety story
Local tests were successful but this should also be tested by QA.


app/res/xml/preferences_developer.xml
```diff
@@ -173,4 +173,9 @@
         android:entryValues="@array/pref_enabled_vals"
         android:key="cc-enable-certificate-transparency"
         android:title="Certificate Transparency"/>
+    <EditTextPreference
+        android:defaultValue="@string/logs_server_url_protocol"
```
shubham1g5, line 177: Wondering if user ever sees this `defaultValue` as an app should always have the `log_receiver_url` populated ?

app/src/org/commcare/preferences/DeveloperPreferences.java
```diff
@@ -120,6 +121,19 @@ protected int getPreferencesResource() {
...
         super.onCreatePreferences(savedInstanceState, rootKey);
         setHasOptionsMenu(true);
+        EditTextPreference editTextPreference = (EditTextPreference)findPreference(PREFS_LOG_POST_URL_KEY);
+        if (editTextPreference != null) {
+            editTextPreference.setOnPreferenceChangeListener((preference, newValue) -> {
+                if (newValue == null || newValue.toString().trim().isEmpty()
+                        || newValue.equals(getString(R.string.logs_server_url_protocol))) {
+                    SharedPreferences appPreferences = CommCareApplication.instance().getCurrentApp()
+                            .getAppPreferences();
+                    appPreferences.edit().remove(PREFS_LOG_POST_URL_KEY).apply();
```
shubham1g5, line 131: think we should be defaulting to the url from app profile xml instead of completely removing it.

shubham1g5 (changes requested):
Few points/questions - 

- Can we add a link to the ticket or github issue which is prompting the change
-  Are we clear and ok with set of steps one need to take to see this setting in dev options ? (Dev options visibility defer in debug and release builds)
- Have we confirmed with the client that this solution works with them ?
- Is it urgent enough for the client to include in a hotfix ?

app/res/xml/preferences_developer.xml
avazirna: they see it when `log_receiver_url` is not set, by [default the logs URL is the Form Submission log](https://github.com/dimagi/commcare-android/blob/831747a00d29328ea0aaa886245e86588fc4af84/app/src/org/commcare/tasks/LogSubmissionTask.java#L112)

app/src/org/commcare/preferences/DeveloperPreferences.java
avazirna: I think this is related to the notion of what is the default URL for the logs submission, right now it seems to be the form submission URL

avazirna:
> Few points/questions -
> 
> * Can we add a link to the ticket or github issue which is prompting the change

Done
> * Are we clear and ok with set of steps one need to take to see this setting in dev options ? (Dev options visibility defer in debug and release builds)

Good point, I may have to whitelist this option, let me check.
> * Have we confirmed with the client that this solution works with them ?

I got sign-off from [this](https://dimagi.atlassian.net/browse/SAAS-15980?focusedCommentId=347077) comment about this option. And I think we are still waiting for the client's response about it [here](https://dimagi.atlassian.net/browse/SUPPORT-21952?focusedCommentId=347123).
> * Is it urgent enough for the client to include in a hotfix ?

Given the context around this client, I believe this is the bare minimum to satisfy their request and considering that they are in the middle of investigating the issues, it does seem relevant enough to take this opportunity to roll it out now.

app/src/org/commcare/preferences/DeveloperPreferences.java
shubham1g5: yeah we should be defaulting it to form submission url instead of removing it completely.

ctsims:
@avazirna I think it's worth confirming briefly that this option can cleanly be set to a different URL, used, then have the custom text deleted, and make sure that the submission reverts to the original receiver. With settings like this where the option is a String value or an empty value, there's always a risk that a blank string setting is treated differently than a null or empty value.

avazirna:
> @avazirna I think it's worth confirming briefly that this option can cleanly be set to a different URL, used, then have the custom text deleted, and make sure that the submission reverts to the original receiver. With settings like this where the option is a String value or an empty value, there's always a risk that a blank string setting is treated differently than a null or empty value.

Yes @ctsims, I tested that scenario a couple of times and also added a [listener to remove the preference](https://github.com/dimagi/commcare-android/pull/2852/commits/a2be44ba7aa194f4692c0364914774d5a70b88ed) when `blank`

app/src/org/commcare/preferences/DeveloperPreferences.java
avazirna: @shubham1g5 that would be changing the way this preference is currently used and create redundancy. Right now, it seems that we only set the preference when the logs URL is different from the form submissions URL. Defaulting to the submissions URL and to keep consistency, would require setting it during app installation as well, right?

ctsims: @shubham1g5 that was my first impulse as well, but I think it's actually too dangerous to set this redundantly for two reasons.

1 - It introduces a new parity requirement. If we changed the submission URL for form submissions on HQ's end, we'd now need to **also** change the log URL, but we can't really do that with HQ settings because it'll blow away the config for anyone who's added the custom setting.
2 - From a usability perspective, it's dangerous because the users don't have a super meaningful way currently to know what **was** in the box for this current UI before. If the Log URL is Empty when you load it, it's clear that setting it back to blank should be the "inverse" that sets things back to neutral. If you load up the field then and its got the HQ default form URL already in it, I think you'd naturally then assume that you need to keep track of that URL so you can set it back if you want to undo the action.

avazirna:

app/src/org/commcare/preferences/DeveloperPreferences.java
shubham1g5: > Defaulting to the submissions URL and to keep consistency, would require setting it during app installation as well, right?

Makes sense, I didn't realise we don't do it already. 

> we can't really do that with HQ settings because it'll blow away the config for anyone who's added the custom setting.

I am not sure I understand it. My thought was to do this as a one time action at app install time -

1. While Installing app, we populate the `logs_receiver_url` in app preferences by copying it from form submission url.

2. When user opens the logs url preference to edit, they would see what this url is pointing to (which would be same as form submission url)

3. User edits the url to some other valid url. We preserve it and show it to user in this preference

4. User edits the url and change is to a non-valid url (empty or something else) -> We show user an error that the change they made is invalid and preserve whatever the old value of the url was. 


No strong thoughts here though overall, I think we want to preserve the existing behaviour in the PR, probably better to rename the field from `Log Submission Server` to `Custom Log Submission Server` to make it more clear that existing log server is not empty when they see it empty.

avazirna:

app/src/org/commcare/preferences/DeveloperPreferences.java
avazirna: @shubham1g5 I think it's a matter of interpretation, my understanding from going through the code is that we don't have a `Log Submission Server` per se, everything is to be sent to the `Form Submission Server`. However, we can have the option to set a `Log Submission Server`. Is this a valid interpretation? I'm happy to follow the `Custom Log Submission Server` approach

app/res/values/strings.xml
```diff
@@ -456,4 +456,5 @@
     <string name="fcm_default_notification_channel">notification-channel-push-notifications</string>
     <string name="app_with_id_not_found">Required CommCare App is not installed on device</string>
     <string name="audio_recording_notification">Audio Recording Notification</string>
+    <string name="logs_server_url_protocol">https://</string>
```
ctsims, line 459: I think this is going to be really confusing if this is the default value. 

From my perspective and the discussion below, we need to frame this as an "override" setting, which redirects logs to a separate path. Having any value in here (like a stub URL) feels very confusing, as it implies that the value was in use before.

I think the "baseline" value of this setting should be an empty string (``) so it's clear that its clear when it's unused, but more critically so that it's clearer on how to reset back to empty.

app/res/xml/preferences_developer.xml
```diff
@@ -173,4 +173,9 @@
...
+        android:enabled="true"
+        android:key="log_receiver_url"
+        android:title="Log Submission Server"/>
```
ctsims, line 180: As per the discussion  below, I think we should rename the text on this to be something like "Override: Custom Log Submission Server" so that it's clear that this is a secondary setting, not the baseline.

app/src/org/commcare/preferences/DeveloperPreferences.java
ctsims: Agreed with "Custom" and updated suggestions on how to be consistent with my take on this above. I think it'd introduce a lot of ambiguity to create a new situation where we **expect** the log submission URL to be set, rather than treating this as an override. 

@shubham1g5 I think the big drawback in your user story is that currently users have an expectation that this URL is synchronized, and the 'custom' version is a rare edge case, so I'd be concerned about situations where we'd need to keep the two in sync across other events or changes by default.

app/res/values/strings.xml
avazirna: I was trying to find a way to inform the user that the protocol needs to be `https` as it will fail otherwise. But happy to explore another option

avazirna:

app/res/xml/preferences_developer.xml
avazirna: Done 9e623dfe3579c385de1f1efe58a49dc5cbb95cb7

avazirna:

app/res/xml/preferences_developer.xml
```diff
@@ -173,4 +173,10 @@
         android:entryValues="@array/pref_enabled_vals"
         android:key="cc-enable-certificate-transparency"
         android:title="Certificate Transparency"/>
+    <EditTextPreference
+        android:dialogTitle="Log Submission Server"
```
shubham1g5, line 177: nit: We should also rename it to have the same `Override` as title.

shubham1g5 (approved):

app/res/xml/preferences_developer.xml
avazirna: And I forgot the word "Custom" too. Corrected [0069632](https://github.com/dimagi/commcare-android/pull/2852/commits/0069632157b68b90c94cf59b3d6d5acf4d9e34d6)

shubham1g5: Ahh I think if we use `Custom` , we should remove the `Override` i.e. use one of them instead of both.

avazirna: Agree

PR #2858, opened 2024-09-18
Adjust user messages around the Session Pause feature

avazirna:
## Summary
Copy of https://github.com/dimagi/commcare-android/pull/2851

cross-request: cross-request: https://github.com/dimagi/commcare-core/pull/1437

### DIFF
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -1159,6 +1159,7 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
         uiController.refreshView();
         FormNavigationUI.updateNavigationCues(this, mFormController, uiController.questionsView);
         if (isRestartAfterSessionExpiration) {
+            // InterruptedFormState null check is important to ensure backward compatibility
             String localeKey =
                     (fc.getInterruptedFormState() == null
                             || fc.getInterruptedFormState().getInterruptedDueToSessionExpiration())
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -1162,7 +1162,7 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
             // InterruptedFormState null check is important to ensure backward compatibility
             String localeKey =
                     (fc.getInterruptedFormState() == null
-                            || fc.getInterruptedFormState().getInterruptedDueToSessionExpiration())
+                            || fc.getInterruptedFormState().isInterruptedDueToSessionExpiration())
                     ? "form.entry.restart.after.expiration" : "form.entry.restart.after.session.pause";
             Toast.makeText(this, Localization.get(localeKey), Toast.LENGTH_LONG).show();
         }
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -249,7 +249,7 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
     }
 
     @Override
-    public void formSaveCallback(boolean exit, Runnable listener) {
+    public void formSaveCallback(Runnable listener) {
         // note that we have started saving the form
         customFormSaveCallback = listener;
         interruptAndSaveForm(exit);
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -259,7 +259,7 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
         if (mFormController != null) {
             // Set flag that will allow us to restore this form when we log back in
             CommCareApplication.instance().getCurrentSessionWrapper().setCurrentStateAsInterrupted(
-                    mFormController.getFormIndex());
+                    mFormController.getFormIndex(), sessionExpiredOrSustended);
 
             // Start saving form; will trigger expireUserSession() on completion
             saveIncompleteFormToDisk(sessionExpiredOrSustended);
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -255,14 +255,14 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
         interruptAndSaveForm(true, false);
     }
 
-    private void interruptAndSaveForm(boolean sessionExpiredOrSustended, boolean userTriggered) {
+    private void interruptAndSaveForm(boolean sessionExpired, boolean userTriggered) {
         if (mFormController != null) {
             // Set flag that will allow us to restore this form when we log back in
             CommCareApplication.instance().getCurrentSessionWrapper().setCurrentStateAsInterrupted(
-                    mFormController.getFormIndex(), sessionExpiredOrSustended);
+                    mFormController.getFormIndex(), sessionExpired);
 
             // Start saving form; will trigger expireUserSession() on completion
-            saveIncompleteFormToDisk(sessionExpiredOrSustended, userTriggered);
+            saveIncompleteFormToDisk(sessionExpired, userTriggered);
         }
     }
 
@@ -1154,8 +1154,8 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
         FormNavigationUI.updateNavigationCues(this, mFormController, uiController.questionsView);
         if (isRestartAfterSessionExpiration) {
             String localeKey =
-                    (fc.getRestoredFormSession() == null
-                            || fc.getRestoredFormSession().getInterruptedDueToSessionExpiration())
+                    (fc.getInterruptedFormState() == null
+                            || fc.getInterruptedFormState().getInterruptedDueToSessionExpiration())
                     ? "form.entry.restart.after.expiration" : "form.entry.restart.after.session.pause";
             Toast.makeText(this, Localization.get(localeKey), Toast.LENGTH_LONG).show();
         }
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -1288,7 +1288,9 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
                     hasSaved = true;
                     break;
                 case SAVED_INCOMPLETE:
-                    toastMessage = Localization.get("form.entry.incomplete.save.success");
+                    if (userTriggered) {
+                        toastMessage = Localization.get("form.entry.incomplete.save.success");
+                    }
                     hasSaved = true;
                     break;
                 case SAVED_AND_EXIT:
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -1153,8 +1153,11 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
         uiController.refreshView();
         FormNavigationUI.updateNavigationCues(this, mFormController, uiController.questionsView);
         if (isRestartAfterSessionExpiration) {
-            Toast.makeText(this,
-                    Localization.get("form.entry.restart.after.expiration"), Toast.LENGTH_LONG).show();
+            String localeKey =
+                    (fc.getRestoredFormSession() == null
+                            || fc.getRestoredFormSession().getInterruptedDueToSessionExpiration())
+                    ? "form.entry.restart.after.expiration" : "form.entry.restart.after.session.pause";
+            Toast.makeText(this, Localization.get(localeKey), Toast.LENGTH_LONG).show();
         }
     }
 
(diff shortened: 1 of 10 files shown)


avazirna:

shubham1g5:
@avazirna It doesn't look like the comments on the original PR are addressed in this PR, let me know if my understanding is wrong.

avazirna:
@shubham1g5 I have entertained a few design alternatives over this, walking the fine line between structural changes and the bare minimum to achieve the goals of this PR. I'm more inclined to leave the structural work for 2.55 where we can have a proper QA and properly assess all the risks. I also don't think that deep changes should be rolled out in a `hotfix`, which I believe you agree with me. So in terms of future work, the main points are:
* `InterruptedFormState` should evolve into `InterruptedSessionState` - we are now redundantly saving the `SessionStateDescriptorId`, [individually](https://github.com/dimagi/commcare-android/blob/2b994dbd9b7a91ca473199b04397e6f27cc0fbcf/app/src/org/commcare/models/AndroidSessionWrapper.java#L195) and in the [InterruptedFormState](https://github.com/dimagi/commcare-android/blob/2b994dbd9b7a91ca473199b04397e6f27cc0fbcf/app/src/org/commcare/models/AndroidSessionWrapper.java#L196) - this means using this model for both Session expiration (and suspension) and Session Pause;
    * This requires a few change in the way we restore expired sessions
    * There may be challenges to ensure backward compatibility, considering that users would be migrating from version with both
* Stop storing the SSDId in SharedPreferences
* Review the workflow around the [is-restart-after-session-expiration](https://github.com/dimagi/commcare-android/blob/5a0578a7ee045c61cd26c6511d4a6696be6f9941/app/src/org/commcare/activities/FormEntryActivity.java#L126) - currently, it implies catering only for Session expiration but the reality is that with the changes we introduced, it's also serving the Session Pause feature

app/src/org/commcare/logic/AndroidFormController.java
```diff
@@ -20,13 +21,17 @@ public class AndroidFormController extends FormController implements PendingCall
     private boolean wasPendingCalloutCancelled;
     private FormIndex formIndexToReturnTo = null;
     private boolean formCompleteAndSaved = false;
+    private InterruptedFormState restoredFormSession;
 
     private FormAnalyticsHelper formAnalyticsHelper;
 
-    public AndroidFormController(FormEntryController fec, boolean readOnly, FormIndex formIndex) {
+    public AndroidFormController(FormEntryController fec, boolean readOnly, InterruptedFormState savedFormSession) {
```
shubham1g5, line 28: nit: inconsistent variable naming. would keep them all as `interruptedFormState`

app/src/org/commcare/logic/AndroidFormController.java
```diff
@@ -20,13 +21,17 @@ public class AndroidFormController extends FormController implements PendingCall
...
         super(fec, readOnly);
         formAnalyticsHelper = new FormAnalyticsHelper();
-        formIndexToReturnTo = formIndex;
+        this.restoredFormSession = savedFormSession;
+        if (savedFormSession !=null ){
+            formIndexToReturnTo = savedFormSession.getFormIndex();
+        }
```
shubham1g5, line 34: think we can remove global reference to `formIndexToReturnTo` and just derive it from `restoredFormSession` when needed

app/src/org/commcare/models/database/InterruptedFormState.java
```diff
@@ -17,10 +18,12 @@ public class InterruptedFormState implements Externalizable {
 
     int sessionStateDescriptorId;
     FormIndex formIndex;
+    private boolean interruptedDueToSessionExpiration = false;
```
shubham1g5, line 21: possibly have it as enum - `InterruptionCause` with values `SESSION_EXPIRATION` and `PROCESS_KILL`

shubham1g5 (approved):
Overall approach looks good to me, left some minor comments around code clarity and readability as optional but nice to have feedback.

shubham1g5:
@avazirna Can you create a ticket describing the pending list of items and make sure it gets prioritised for 2.55.

app/src/org/commcare/models/database/InterruptedFormState.java
avazirna: @shubham1g5 I entertained this idea too, the issue is that there is no implicit call for `PROCESS_KILL`, we kind of do preventive saves. so it would be inconsistent with the triggering process. Should we still call is `PROCESS_KILL` or `PREVENTIVE_SAVE`, I'm ok with either.

app/src/org/commcare/logic/AndroidFormController.java
avazirna: `formIndexToReturnTo` [is nullified after its first use](https://github.com/dimagi/commcare-android/blob/2b994dbd9b7a91ca473199b04397e6f27cc0fbcf/app/src/org/commcare/logic/AndroidFormController.java#L76C14-L76C33), which makes sense because we are supposed to return to that FormIndex only when loading the form after a session expiration or app kill. Am I missing something here?

app/src/org/commcare/models/database/InterruptedFormState.java
shubham1g5: Makes sense, I think fine to leave it as it is.

app/src/org/commcare/logic/AndroidFormController.java
shubham1g5, line 28: can we mark `savedFormSession` and `restoredFormSession` as `@Nullable`

app/src/org/commcare/logic/AndroidFormController.java
shubham1g5: makes sense. no need for a change here.

app/src/org/commcare/logic/AndroidFormController.java
```diff
@@ -20,13 +21,17 @@ public class AndroidFormController extends FormController implements PendingCall
...
 
     private FormAnalyticsHelper formAnalyticsHelper;
 
-    public AndroidFormController(FormEntryController fec, boolean readOnly, FormIndex formIndex) {
+    public AndroidFormController(FormEntryController fec, boolean readOnly, InterruptedFormState savedFormSession) {
         super(fec, readOnly);
         formAnalyticsHelper = new FormAnalyticsHelper();
-        formIndexToReturnTo = formIndex;
+        this.restoredFormSession = savedFormSession;
+        if (savedFormSession !=null ){
```
shubham1g5, line 32: nit:formatting

app/src/org/commcare/logic/AndroidFormController.java
avazirna: [7218aba](https://github.com/dimagi/commcare-android/pull/2858/commits/7218aba874042f32b616a363889118bfc7bd2516)

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -766,18 +766,19 @@ protected void fireCompoundIntentDispatch() {
...
 
     private void saveCompletedFormToDisk(String updatedSaveName) {
-        saveDataToDisk(FormEntryConstants.EXIT, true, updatedSaveName, false);
+        saveDataToDisk(FormEntryConstants.EXIT, true, updatedSaveName, false, true);
     }
 
-    private void saveIncompleteFormToDisk(boolean exit) {
-        saveDataToDisk(exit, false, null, true);
+    private void saveIncompleteFormToDisk(boolean sessionExpiredOrSuspended, boolean userTriggered) {
+        boolean shouldExit = sessionExpiredOrSuspended ? FormEntryConstants.EXIT : FormEntryConstants.DO_NOT_EXIT;
```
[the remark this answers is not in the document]
avazirna: done https://github.com/dimagi/commcare-android/pull/2858/commits/3ade9b262f34d6db51cfa01eb3f6a5102e6d5e25

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -249,20 +249,20 @@ public void onCreateSessionSafe(Bundle savedInstanceState) {
     }
 
     @Override
-    public void formSaveCallback(boolean exit, Runnable listener) {
+    public void formSaveCallback(Runnable listener) {
         // note that we have started saving the form
         customFormSaveCallback = listener;
-        interruptAndSaveForm(exit);
+        interruptAndSaveForm(true, false);
```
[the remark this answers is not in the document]
avazirna: done https://github.com/dimagi/commcare-android/pull/2858/commits/3ade9b262f34d6db51cfa01eb3f6a5102e6d5e25

app/src/org/commcare/models/database/InterruptedFormState.java
```diff
@@ -59,6 +63,10 @@ public FormIndex getFormIndex() {
         return formIndex;
     }
 
+    public boolean getInterruptedDueToSessionExpiration(){
```
shubham1g5, line 66: nit: rename to `isInterrupted..`

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -793,7 +796,7 @@ protected void onExternalAttachmentUpdated() {
         }
         String formStatus = formRecord.getStatus();
         if (FormRecord.STATUS_INCOMPLETE.equals(formStatus)) {
-            saveDataToDisk(false, false, null, true);
+            saveDataToDisk(false, false, null, true, true);
```
shubham1g5, line 799: not sure if this save is`userTriggered` as it happens in response to updating an attachment.

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -1063,7 +1066,8 @@ private void loadForm() {
             }
 
             mFormLoaderTask = new FormLoaderTask<FormEntryActivity>(symetricKey, instanceIsReadOnly,
-                    formEntryRestoreSession.isRecording(), FormEntryInstanceState.mFormRecordPath, this, lastFormIndex) {
+                    formEntryRestoreSession.isRecording(), FormEntryInstanceState.mFormRecordPath, this,
```
shubham1g5, line 1069: this looks weird to have a comma just before method call close. No compilation errors here ?

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -1157,8 +1161,11 @@ private void handleFormLoadCompletion(AndroidFormController fc) {
         uiController.refreshView();
         FormNavigationUI.updateNavigationCues(this, mFormController, uiController.questionsView);
         if (isRestartAfterSessionExpiration) {
-            Toast.makeText(this,
-                    Localization.get("form.entry.restart.after.expiration"), Toast.LENGTH_LONG).show();
+            String localeKey =
+                    (fc.getInterruptedFormState() == null
```
shubham1g5, line 1165: can `fc.getInterruptedFormState() == null` be ever true here given `isRestartAfterSessionExpiration` is already true ?

app/src/org/commcare/models/AndroidSessionWrapper.java
```diff
@@ -185,12 +185,14 @@ private static boolean ssdHasValidFormRecordId(int ssdId,
...
 
-    public void setCurrentStateAsInterrupted(FormIndex formIndex) {
+    public void setCurrentStateAsInterrupted(FormIndex formIndex, boolean sessionExpired) {
         if (sessionStateRecordId != -1) {
             SqlStorage<SessionStateDescriptor> sessionStorage =
                     CommCareApplication.instance().getUserStorage(SessionStateDescriptor.class);
             SessionStateDescriptor current = sessionStorage.read(sessionStateRecordId);
-            InterruptedFormState interruptedFormState = new InterruptedFormState(current.getID(), formIndex, current.getFormRecordId());
+          
+            InterruptedFormState interruptedFormState = new InterruptedFormState(current.getID(), formIndex, current.getFormRecordId(), sessionExpired);
```
shubham1g5, line 194: nit: not sure on formatting here.

shubham1g5 (changes requested):

app/src/org/commcare/activities/FormEntryActivity.java
avazirna: true, good catch

app/src/org/commcare/activities/FormEntryActivity.java
avazirna: Noticed that too, bad conflict resolution. Redid it locally, pushing it in a few.

app/src/org/commcare/activities/FormEntryActivity.java
avazirna: For backward compatibility, we might load a previously saved session using an older version of CC, where the InterruptedFormState was not saved.

shubham1g5: makes sense, can we make a comment  there that the null check is to support backward compatibility and should not be removed.

app/src/org/commcare/models/database/InterruptedFormState.java
avazirna: [dec7568](https://github.com/dimagi/commcare-android/pull/2858/commits/dec75686a49af9085f4b2889769c0a1c812ae85a)

app/src/org/commcare/activities/FormEntryActivity.java
avazirna: [019cd2d](https://github.com/dimagi/commcare-android/pull/2858/commits/019cd2d0521de8f6a49e3c3c1092fe59b0a49e38)

app/src/org/commcare/activities/FormEntryActivity.java
avazirna: [a27ab98](https://github.com/dimagi/commcare-android/pull/2858/commits/a27ab98c1f70f7afd060af5cf31ddd25f8e3def4)

shubham1g5 (approved):

avazirna:

PR #2872, opened 2024-10-14
PM-bug fixes in ui

pm-dimagi:
Suggested ui changes in new ui/ux

## Safety Assurance


app/src/org/commcare/fragments/connect/ConnectLearningProgressFragment.java
```diff
@@ -70,6 +77,16 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
         return view;
     }
 
+    public void expand(View view) { //the function that given to onclick event
```
OrangeAndGreen, line 80: The "View More" link is supposed to navigate to the Job Details page, rather than show/hide the description text. We'll be replacing this tile with a reusable component soon (currently in another PR), so for now can just remove this click handler and leave the card static

OrangeAndGreen (changes requested):
A little cleanup to do to remove the click handler (see comment), but otherwise this looks good

app/src/org/commcare/fragments/connect/ConnectLearningProgressFragment.java
OrangeAndGreen: Ok to ignore since this is about to be refactored.

OrangeAndGreen (approved):

PR #2873, opened 2024-10-17
Migrate to ViewPager2

devanshhooda:
## Summary
Poc to fix: [SAAS-16035](https://dimagi.atlassian.net/browse/SAAS-16035)

## Product Description
Removed the usage of Duolingo's rtl viewpager as the repository is archived and no longer holds support.
It was observed that the issue was reproducible only in release mode while working fine in debug mode hence we suspected that it might be because outdated viewpager dependency.

In terms of UX, the behaviour almost the same as it was with RtlViewPager, the most notable difference is that when swiping through the tabs the active page doesn't change, only when a tab is selected:
### Case Detail
<table>
<tr>
<td>
<video width="200" height="200" src="https://github.com/user-attachments/assets/4608f63e-d4a3-4c89-a83c-584278751aec"/>
</td>
</tr>
</table>

### Case Details in landscape mode on large screens
<table>
<tr>
<td>
<video width="200" height="200" src="https://github.com/user-attachments/assets/6eceec06-2027-48a3-a769-30fc9e06b256"/>
</td>
</tr>
</table>

### Inline Tile
<table>
<tr>
<td>
<video width="200" height="200" src="https://github.com/user-attachments/assets/f3202133-18da-4a57-8a38-8eb39e7702c7"/>
</td>
</tr>
</table>

## Safety Assurance

### Note for QA
This PR brings changes to the tab system used by case details and inline tiles, see videos above. It's important during QA to see if the tabs are working as expected and that right-to-left (RTL) is still supported.


app/res/values/dimens.xml
```diff
@@ -16,6 +16,8 @@
     <dimen name="title_padding">10dp</dimen>
     <dimen name="title_paddingLeft">20dp</dimen>
     <dimen name="tab_gradient_radius">300dp</dimen>
+    <dimen name="detail_view_tab_indicator_height">3dp</dimen>
+    <dimen name="detail_view_tab_padding_horizontal">24dp</dimen>
```
shubham1g5, line 20: Is there a reason for changing these instead of applying the defaults from material design styles ?

app/src/org/commcare/views/TabbedDetailView.java
```diff
@@ -61,79 +61,43 @@
...
-        mMenu.getChildAt(position).setBackgroundDrawable(getResources().getDrawable(R.drawable.title_case_tab_vertical));
     }
 
     public int getCurrentTab() {
         return mViewPager.getCurrentItem();
     }
 
     public int getTabCount() {
-        return mViewPager.getAdapter().getCount();
+        return Objects.requireNonNull(mViewPager.getAdapter()).getItemCount();
```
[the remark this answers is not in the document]
shubham1g5: @avazirna are we sure `mViewPager.getAdapter()` can't be null if a caller calls `getTabCount` out of order on this view ?

app/src/org/commcare/views/TabbedDetailView.java
```diff
@@ -61,79 +61,43 @@ private void loadViewConfig(Context context, AttributeSet attrs) {
...
-        for (int i = 0; i < mMenu.getChildCount(); i++) {
-            mMenu.getChildAt(i).setBackgroundDrawable(getResources().getDrawable(R.drawable.title_neutral_tab_vertical));
+        if (detail.isCompound()) {
+            new TabLayoutMediator(mTabLayout, mViewPager,
+                    (tab, position) -> tab.setText(detail.getDetails()[position].getTitle().getText().evaluate()))
+                    .attach();
+        } else {
+            mTabLayout.setVisibility(GONE);
         }
-        mMenu.getChildAt(position).setBackgroundDrawable(getResources().getDrawable(R.drawable.title_case_tab_vertical));
```
shubham1g5, line 129: why was this removed ?

shubham1g5 (changes requested):
looks quite good, left some clarifying questions. 

Also can we add videos for other places we are using a tabbed view in the app. 

Thanks!

app/res/values/dimens.xml
avazirna: the default tab indicator height seems a bit too thin:
image
And the horizontal padding was already set and I didn't test without, let me do that.

app/src/org/commcare/views/TabbedDetailView.java
avazirna: I believe the default style is enough, but will make sure it looks fine everywhere

avazirna: @shubham1g5 Here's the old style to which those drawables where used:
645ebc7e-74a2-11e4-8de9-cf45e08495a6

shubham1g5: and how does it look now with the changes in PR ?

avazirna: @shubham1g5 my understanding is that this style was used by the old tab system, which was based on the [tabbed_detail_menu](https://github.com/dimagi/commcare-android/blob/423846919137930ca0610e268c2ba8a0f60177be/app/res/layout/tabbed_detail_view.xml#L7C26-L7C44) linear layout. That system was discontinued when `PagerTabStrip` was added to CC, [this](https://github.com/dimagi/commcare-android/pull/401/files#diff-2e88d3fe8ebeb3aa7b671fbeeae867da2e9af67242f2d2e227a38a465b835679) is the PR. So this PR doesn't change the style around the tabs, the current and previous styling is what I shared in [this](https://github.com/dimagi/commcare-android/pull/2873#discussion_r1890009087) comment.

avazirna:
> are we sure mViewPager.getAdapter() can't be null if a caller calls getTabCount out of order on this view ?

@shubham1g5 This logic was not changed as such but I agree that we should null check this

app/res/values/dimens.xml
avazirna: In my opinion, it looks better with `24dp` lateral padding (2nd image)
image
What do you think?

shubham1g5: I agree, thanks for verifying this.

app/src/org/commcare/views/TabbedDetailView.java
shubham1g5: I think my question here is more around how does this change not effect Case Detail tabs for example given `TabbedDetailView` is used there and this line of code seems generic to all `TabbedDetailView` layouts.

avazirna: hannn, in the old tab system, we used to add `HorizontalMediaView` views to [tabbed_detail_menu](https://github.com/dimagi/commcare-android/blob/423846919137930ca0610e268c2ba8a0f60177be/app/res/layout/tabbed_detail_view.xml#L7C26-L7C44) linear layout and apply this style to them, this was removed [here](https://github.com/dimagi/commcare-android/pull/401/commits/0ce8f12bf0bd7b097d4ca803843c29bfe041b9c9#diff-2e88d3fe8ebeb3aa7b671fbeeae867da2e9af67242f2d2e227a38a465b835679L108). And because we don't add views to that linear layout, every time we call [mMenu.getChildCount()](https://github.com/dimagi/commcare-android/blob/423846919137930ca0610e268c2ba8a0f60177be/app/src/org/commcare/views/TabbedDetailView.java#L122C9-L122C49) we get 0 and the execution is interrupted. This to say that we haven't been calling this code for quite some time, it's dead code now.

shubham1g5: got it , thanks for confirming.

PR #2884, opened 2024-10-23
Extend user session during audio recording start

avazirna:
## Summary
This PR implements a mechanism for CommCare to extend the user session when the current widget data can't be recovered in case of session expiration. This extension is basically to prevent the user session from expiring which automatically logs the user out. This implementation entails:
1. Adds a way to extend the user session 
2. Triggers the user session extension logic during audio recording initiation

Context around this change, a project that makes use of the CommCareAudioWidget to capture long audio recordings reported that users were being unexpectedly logged out in the middle of the recordings and losing data, this was later associated to the session expiring in the middle of an active recording. More information on [this](https://dimagi.atlassian.net/browse/SAAS-15417) ticket.

## Safety Assurance

### Safety story
This only adds more time to the current session. Local tests were successfully.


avazirna:

shubham1g5 (approved):

PR #2899, opened 2024-11-20
PM-title bar color changes

pm-dimagi:
Change of the title bar color in the commcare and the launch icon changes

## Feature Flag
https://dimagi.atlassian.net/browse/CCCT-534

## Product Description
https://www.figma.com/design/UWjp1wRNAO0o9q2llwDKMc/CommCare-Cosmetic-Changes-Post-Nov-'22Rebrand?node-id=1322-3886&t=TgPFUaWgJx2XMlmX-1
Screenshot_20241120_032101_CommCare Debug
WhatsApp Image 2024-11-22 at 08 10 47
WhatsApp Image 2024-11-22 at 08 50 40

## PR Checklist


app/res/values/themes.xml
```diff
@@ -73,7 +73,9 @@
 
     <style name="ActionBar.TitleTextStyle">
         <item name="android:textColor">@color/white</item>
-        <item name="android:paddingLeft">8dp</item>
+        <item name="android:paddingLeft">16dp</item>
+        <item name="android:textSize">18sp</item>
+        <item name="android:textStyle">bold</item>
```
shubham1g5, line 78: Are these values coming from figma as well ? I would think we should keep the style same as `TitleTextStyle` as thats standard Android sizes and font styles.

app/res/values/themes.xml
```diff
@@ -74,8 +74,8 @@
     <style name="ActionBar.TitleTextStyle">
         <item name="android:textColor">@color/white</item>
         <item name="android:paddingLeft">16dp</item>
-        <item name="android:textSize">18sp</item>
-        <item name="android:textStyle">bold</item>
+        <item name="android:textSize">22sp</item>
+        <item name="android:fontWeight" tools:targetApi="o">400</item>
```
shubham1g5, line 78: Do we not need this on pre O devices ?

app/res/values/themes.xml
```diff
@@ -60,10 +60,10 @@
     </style>
 
     <style name="ActionBar" parent="Widget.AppCompat.Light.ActionBar">
-        <item name="background">@color/cc_brand_color</item>
+        <item name="background">@color/title_bar_color</item>
         <item name="titleTextStyle">@style/ActionBar.TitleTextStyle</item>
         <item name="subtitleTextStyle">@style/ActionBarSubtitleTextStyle</item>
-        <item name="logo">@drawable/actionbar_modern_spacing</item>
+        <item name="logo">@drawable/logo_spacing</item>
```
shubham1g5, line 66: we should delete ther resources we are no longer using.

app/res/drawable/logo.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
shubham1g5, line 1: nit: Rename to `commcare_logo.xml`

app/res/drawable/logo_spacing.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layer-list
+    xmlns:android="http://schemas.android.com/apk/res/android">
+    <item
+        android:drawable="@drawable/logo"
+        android:right="16dp"/>
```
shubham1g5, line 6: is this intentional to not have equal left and right values like the earlier `actionbar_modern_spacing` ?

app/res/drawable/logo_spacing.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
```
shubham1g5, line 1: nit: rename to `commcare_actionbar_logo_spacing`

shubham1g5 (changes requested):

app/res/values/themes.xml
pm-dimagi: we need it on all but weight 400 is available above this only otherwise we have only bold option

app/res/drawable/logo_spacing.xml
pm-dimagi: yess because as per figma its 16dp in right and by default it takes 16 in left

app/res/values/themes.xml
pm-dimagi: Yes we need it but font weight 400 doses not support below that , we have to use bold or have to get ttf file

app/res/drawable/logo_spacing.xml
pm-dimagi: left is predefined 16dp so no need to add that
Right is earlier 8dp but as per figma it should be 16dp

app/res/values/themes.xml
shubham1g5: can we use bold instead then to be consistent across devices ?

shubham1g5 (changes requested):
Incomplete PR checklist.

shubham1g5 (approved):
Approved, though we should not merge until the base PR is merged into `master`

PR #2911, opened 2024-12-05
Removes some logs which are killing our sumo usage

shubham1g5:
## Summary

Cleaning some not so useful logs which makes up for almost half of the log entries in sumologic in hopes to reduce our sumo usage by a good factor. 

Screenshot 2024-12-05 at 10 40 35 PM

## PR Checklist

### Automated test coverage

NA

### Safety story
 
Logs only

### DIFF
--- a/app/src/org/commcare/appupdate/CommcareFlexibleAppUpdateManager.java
+++ b/app/src/org/commcare/appupdate/CommcareFlexibleAppUpdateManager.java
@@ -161,8 +161,6 @@ public class CommcareFlexibleAppUpdateManager implements FlexibleAppUpdateContro
         if (mAppUpdateState == newState) {
             return;
         }
-        Logger.log(LogTypes.TYPE_CC_UPDATE, "Publishing status update to : " + newState.name() + ", from : "
-                + (mAppUpdateState != null ? mAppUpdateState.name() : "null"));
         mAppUpdateState = newState;
         mCallback.run();
     }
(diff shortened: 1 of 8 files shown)


app/src/org/commcare/appupdate/CommcareFlexibleAppUpdateManager.java
```diff
@@ -161,8 +161,6 @@ private void publishStatus() {
         if (mAppUpdateState == newState) {
             return;
         }
-        Logger.log(LogTypes.TYPE_CC_UPDATE, "Publishing status update to : " + newState.name() + ", from : "
-                + (mAppUpdateState != null ? mAppUpdateState.name() : "null"));
```
avazirna, line 165: @shubham1g5 Are we 100% sure that this is unnecessary? We can consider logging when the state changes.

shubham1g5: This already logs only when state changes (look at the if cond above this). I don't think this gives us any useful debugging info as such as most functionality there is managed by Play services API. Also there is nothing like `100% certainity` with what you might and might not need with logs and think we should avoid this framing in deciding what to log. Instead a better framing for retroactively deleting logs is have you found a particular line of useful while solving any debugging issues in fast ? Happy to re-consider if so.

avazirna: > Instead a better framing for retroactively deleting logs is have you found a particular line of useful while solving any debugging issues in past ?

avazirna (approved):

PR #2928, opened 2024-12-19
Cache and Lazy-Load Modifications

shubham1g5:
## Summary

https://dimagi.atlassian.net/browse/SC-4055

[Spec](https://docs.google.com/document/d/1LM2PBXqDrUOKVfk_p2niu_-D695goeAGLevZ0jrrMVc/edit?tab=t.0#heading=h.mv6dkvomhjog)

Makes 2 improvements for a Cache backed entity list - 

1. Adds a progress indicator to show the loading  progress in different phases of the entity list loading

2. Uses a Work manager task to schedule entity list cache calculation ahead of time and handle synchronocity with user opening the entity list when a background task is already in Progress. All of this is managed with the help of a [new Android specific entity factory class](https://github.com/dimagi/commcare-android/blob/primeCacheAheadOfTime/app/src/org/commcare/entity/AndroidAsyncNodeEntityFactory.kt). 

3. Changes to Cache Layer to be able to selectively cache certain fields while leaving out others while at the same time extending caching capabilities to non-sort values for a detail field. 

## Feature Flag

CACHE_AND_INDEX

## PR Checklist

### Automated test coverage

None

### Safety story

- tested locally different scenarios handled in [AndroidAsyncNodeEntityFactory](https://github.com/dimagi/commcare-android/blob/primeCacheAheadOfTime/app/src/org/commcare/entity/AndroidAsyncNodeEntityFactory.kt) 
- Major changes are behind `detail.shouldCache()` check which depends on a new suite.xml config to turn on the entity caching and therefore should not impact any existing projects. 

QA Note: Test new feature - https://dimagi.atlassian.net/wiki/spaces/GS/pages/2939519055/Feature+Usage+Guide+Case+List+Caching+and+Lazy+Loading+in+CommCare

cross-request: https://github.com/dimagi/commcare-core/pull/1460


shubham1g5:
@codereabbitai full review

app/src/org/commcare/entity/AndroidAsyncNodeEntityFactory.kt
```diff
@@ -0,0 +1,102 @@
...
+                        detail,
+                        entities,
+                        progressListener
+                    )
+                }
+            }
+        } else {
+            super.prepareEntitiesInternal(entities)
+        }
+    }
```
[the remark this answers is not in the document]
shubham1g5: This code is only ever invoked from a background thread. Is there a way I can be explicit about that ?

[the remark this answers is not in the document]
shubham1g5: avoiding to do this for now as the code is shared with CLI which will actually calls this on main thread.

shubham1g5:

app/src/org/commcare/activities/EntitySelectActivity.java
```diff
@@ -852,16 +856,7 @@ public void deliverLoadResult(List<Entity<TreeReference>> entities,
                                   List<TreeReference> references,
                                   NodeEntityFactory factory, int focusTargetIndex) {
         loader = null;
-
-        AdapterView visibleView;
-        if (shortSelect.shouldBeLaidOutInGrid()) {
```
ctsims, line 857: Is this removing the gridview modality of entity selects entirely? I think it is mostly deprecated, but probably important to call out specifically if so to confirm that it's safe to do so.

app/src/org/commcare/activities/EntitySelectActivity.java
```diff
@@ -852,16 +856,7 @@ public void deliverLoadResult(List<Entity<TreeReference>> entities,
...
-        AdapterView visibleView;
-        if (shortSelect.shouldBeLaidOutInGrid()) {
-            visibleView = ((GridView)this.findViewById(R.id.screen_entity_select_grid));
-        } else {
-            ListView listView = this.findViewById(R.id.screen_entity_select_list);
-            EntitySelectViewSetup.setupDivider(this, listView, shortSelect.usesEntityTileView());
-            visibleView = listView;
-        }
-
+        setProgressText(StringUtils.getStringRobust(this, R.string.entity_list_finalizing));
```
ctsims, line 859: This message doesn't seem like the right one to me. Isn't this just the start of the loading process?

app/src/org/commcare/activities/EntitySelectActivity.java
```diff
@@ -995,6 +994,25 @@ public void deliverLoadError(Exception e) {
...
 
+    @Override
+    public void deliverProgress(Integer[] values) {
+        EntityLoadingProgressListener.EntityLoadingProgressPhase phase =
+                EntityLoadingProgressListener.EntityLoadingProgressPhase.fromInt(values[0]);
+        // throttle to not update text too frequently
+        if (values[1] % 100 == 0) {
+            switch (phase) {
+                case PHASE_PROCESSING -> setProgressText(
+                        StringUtils.getStringRobust(this, R.string.entity_list_processing,
```
ctsims, line 1005: Raw count text inside the view might be a tricky way to view this progress, btw, compared to another form of loading indicator. 

In the big views we've used with cache and index historically, sometimes the length of the view is tens-of-thousands of cases, which is going to make a fairly large and pretty messy looking view. Even just using a % might work better.

app/src/org/commcare/entity/AndroidAsyncNodeEntityFactory.kt
```diff
@@ -0,0 +1,116 @@
...
+) : AsyncNodeEntityFactory(d, ec, entityStorageCache) {
+
+    companion object {
+        const val TEN_MINUTES = 10 * 60 * 1000L
+    }
+
+    override fun prepareEntitiesInternal(
+        entities: MutableList<Entity<TreeReference>>
+    ) {
+        if (detail.shouldOptimize() && detail.isCacheEnabled) {
```
ctsims, line 37: There was another place where a similar check was done, which implied there were 3 potential states for this system based on combinations of different configs. I'm a little concerned it'll be hard to the various places these modes are checked in sync over time, it might be helpful to unify the checks in one place so that if changes are made later we can't miss one.

app/src/org/commcare/tasks/PrimeEntityCacheHelper.kt
```diff
@@ -0,0 +1,215 @@
...
+
+    /**
+     * Primes cache for all entity screens in the app
+     * @throws IllegalStateException if a cache prime is already in progress or user session is not active
+     */
+    @Synchronized
+    fun primeEntityCache() {
+        checkPreConditions()
+        try {
+            primeEntityCacheForApp(CommCareApplication.instance().commCarePlatform)
```
ctsims, line 84: I'd be worried managing the parts of the lifecycle here that depend on the current state of the seated application and platform implicitly. I think it's going to be really hard to ensure aligned execution with the app state resources (for example, an app being unseated and a different app being reseated in between calls without a bad read happening in between that kills the worker). 

There are a couple of ways to control for that, but I think the safest might be to make access to the helper/worker come from a different object's lifecycle, like nesting it inside the currently seated application/platform singleton and then associating that helper with the public call, rather than making this helper itself the singleton object. That would make sure that if there was misalignment, it would stay limited to the orphaned helper.

ctsims (changes requested):
Mostly minor feedback, but I think the issues raised with the Singleton pattern for the caching helper need to be addressed. I think it can be implemented without creating a new independent static lifecycle, and that we should be trying to avoid needing new static lifecycles in the future if possible, especially for classes that could be executing independent working code outside of the managed app lifecycle.

app/src/org/commcare/activities/EntitySelectActivity.java
shubham1g5: No, the intention here was to remove duplicattion with [setupUI()](https://github.com/dimagi/commcare-android/blob/dbbcfd48c9a5e7ddc447b494b5d8598e4abf1db8/app/src/org/commcare/activities/EntitySelectActivity.java#L262.) which gets called in Activity's `onCreate`

app/src/org/commcare/tasks/PrimeEntityCacheHelper.kt
shubham1g5: > to make access to the helper/worker come from a different object's lifecycle, like nesting it inside the currently seated application/platform singleton and then associating that helper with the public call, rather than making this helper itself the singleton object.

Would not that mean that there can be multiple instances of the `PrimeEntityCacheHelper`  per application singleton ? I think we also want to make sure that we don't duplicate cache priming work in simultaneous threads.

app/src/org/commcare/activities/EntitySelectActivity.java
ctsims: aha, got it. I didn't see that code duplicated elsewhere so I assumed it was removed for being old/broken. Thanks for clarifying

app/src/org/commcare/tasks/PrimeEntityCacheHelper.kt
ctsims: If we make it a member of the seated singleton platform/app there will only ever be one helper per seated application, but it means that it's much easier to tie lifecycle events to the application itself, since we could make the app platform object final and immutable. 

Specifically, it would let us be sure that in the process of unseating one app and seating another that neither of those objects could end up with a state mismatch. If the currently seated storage changed underneath the one shutting down, it would fail and could close itself out naturally, and the new one spinning up wouldn't have to deal with things like race conditions in the process of closing out old listeners while other ones are being added.

app/src/org/commcare/activities/EntitySelectActivity.java
shubham1g5: https://github.com/dimagi/commcare-android/pull/2928/commits/15cf20ad67bc7290a2ef6502acf2d70049244def

app/src/org/commcare/entity/AndroidAsyncNodeEntityFactory.kt
shubham1g5: https://github.com/dimagi/commcare-android/pull/2928/commits/71981614565917191d1fe077982d9633cbc31411

app/src/org/commcare/tasks/PrimeEntityCacheHelper.kt
shubham1g5: https://github.com/dimagi/commcare-android/pull/2928/commits/aa65ba4f9db4e58cb1d4838cd3c0dc650b89105f, not completely sure if this is what you meant though.

app/src/org/commcare/activities/EntitySelectActivity.java
shubham1g5: changes the message to `soting cases` in https://github.com/dimagi/commcare-android/pull/2928/commits/b24d5800ccee08fcaf272ecf7f55dd12e20586d8 as I think that will be the most time consuming process after setting the adapter. We do things differently based on different configs mode which makes it hard to deduce the right message for different phases but below is a gist to make it more clear - 

1. Normal Entity List (without caching and lazy loading)

- We calculate the entity references and calculate all Entity Fields simultaneously
- Set the List to Adapter which in turn also sorts the list using already calculated fields


2. With Cache Enabled but no lazy loading

- We calculate the async entity references from nodeset
- Subsequently, We calculate all entity fields which can happen from cache or a raw xpath calculation 
- Set the List to Adapter which in turn also sorts the list using already calculated fields


3. With Cache and Lazy Loading Enabled

- We calculate the async entity references from nodeset
- We calculate All NON-OPTIMIZABLE entity fields
- Set the List to Adapter which in turn also sorts the list (We may need to calculate any lazy fields here if sorting is specified on the lazy field)
- We calculate Optimzed fields lazily and cache them as user scroll through the list


In all of these scenarios, the only major thing between setting the list to adapter and user seeing cases on the list is "Sorting" which is why I have now changed this message to signify that.

app/src/org/commcare/entity/AndroidAsyncNodeEntityFactory.kt
```diff
@@ -31,10 +31,16 @@ class AndroidAsyncNodeEntityFactory(
         const val TEN_MINUTES = 10 * 60 * 1000L
     }
 
+    init {
+        if (!detail.shouldOptimize()) {
+            throw RuntimeException(AndroidAsyncNodeEntityFactory::class.simpleName + " can only be used for optimizable case lists");
```
ctsims, line 39: Great change, this is  really solid refactor.

app/src/org/commcare/CommCareApp.java
```diff
@@ -411,6 +414,17 @@ public SQLiteDatabase getHandle() {
         };
     }
 
+    public PrimeEntityCacheHelper getPrimeEntityCacheHelper() {
```
ctsims, line 417: Thanks for the refactor, I think this coupling will make a lot of sense and keep things straight

app/src/org/commcare/tasks/PrimeEntityCacheHelper.kt
```diff
@@ -65,12 +56,11 @@ class PrimeEntityCacheHelper private constructor() : Cancellable {
...
+    }
 
-        @JvmStatic
-        fun cancelWork() {
-            instance?.cancel()
-            WorkManager.getInstance(CommCareApplication.instance()).cancelUniqueWork(PRIME_ENTITY_CACHE_REQUEST)
-        }
+    fun cancelWork() {
+        cancel()
+        WorkManager.getInstance(CommCareApplication.instance()).cancelUniqueWork(PRIME_ENTITY_CACHE_REQUEST)
```
ctsims, line 76: No need to address if I'm overthinking, but if the spin-down process for this getting executed takes some time, will it already be out of the queue reliably by the next worker? If not, should this be keyed against something like the app id / work type as well? It's probably a non-issue since an old worker will error out anyway when its database goes away.

app/res/values/strings.xml
```diff
@@ -460,5 +460,5 @@
     <string name="entity_list_processing" cc:translatable="true">Processing cases: %s</string>
     <string name="entity_list_loading_cache" cc:translatable="true">Loading Cache…</string>
     <string name="entity_list_calculating" cc:translatable="true">Calculating cases %s</string>
-    <string name="entity_list_finalizing" cc:translatable="true">Finalizing list…</string>
+    <string name="entity_list_sorting" cc:translatable="true">Sorting cases…</string>
```
ctsims, line 463: Very nit-picky feedback - I don't know that 'sorting' here adds a lot of detail for users and might be a little tricky to translate, but "Finalizing" is also extremely technical, so I also wouldn't recommend switching back. I wonder if 'Finishing' might be a good, less jargon-y way to communicate.

ctsims (approved):
Re-reviewed commit by commit and I think my core feedback was all addressed. Only remaining input is that it'd be really great to get how this works written down somewhere, and make sure that the underlying states are communicated effectively for debugging. There are now a lot of factors involved in how things get processed.

app/res/values/strings.xml
shubham1g5: https://github.com/dimagi/commcare-android/pull/2928/commits/7530f944acd849ac4731affdb3bb128768fabf2a

app/src/org/commcare/tasks/PrimeEntityCacheHelper.kt
shubham1g5: I would think so but probably better to not rely on assumptions and also it's more consistent with other app specific requests - https://github.com/dimagi/commcare-android/pull/2928/commits/99329d3292b941081cdead072fe559e9c193076e and https://github.com/dimagi/commcare-android/pull/2928/commits/89a054f50ac7a00baa2ed69d02d6a9592f1ef0e0

shubham1g5:
> Only remaining input is that it'd be really great to get how this works written down somewhere, and make sure that the underlying states are communicated effectively for debugging.

That resonates with me. I will prioritize a wiki here and get back on it.

shubham1g5:
@coderabbitai generate docstrings

shubham1g5:

ctsims:
> Also curious if you have opinion on this level of details in Java docs - https://github.com/dimagi/commcare-android/pull/2965 ?

Hm.. I feel like that level of docstring is somewhat anti-useful. Since it's pulling the docs from the method signatures I think it's just replicating the information already there without adding value, while introducing risk that someone only changes one source and not the other

ctsims (approved):

shubham1g5:

avazirna (approved):

PR #2930, opened 2024-12-20
Update component exported attribute

avazirna:
## Summary
[Android 14 brought new restrictions around implicit intents](https://developer.android.com/about/versions/14/behavior-changes-14#safer-intents) and CC app callouts are based on implicit intents, this PR reviews the `exported` attribute of activities targeted by predefined app callout options.
Ticket: https://dimagi.atlassian.net/browse/QA-7356

## PR Checklist

- [X] If I think the PR is high risk, "High Risk" label is set
- [X] Does the PR introduce any major changes worth communicating ? If yes, "Release Note" label is set and a "Release Note" is specified in PR description.

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -93,22 +93,9 @@
     </permission>
 
     <permission
-        android:name="${applicationId}.permission.READ_NFC_DATA"
-        android:label="@string/permission_read_nfc_data_label"
-        android:permissionGroup="commcare.permission-group.INTERNAL_ACTION"
-        android:protectionLevel="signature">
-    </permission>
-
-    <permission
-        android:name="${applicationId}.permission.WRITE_NFC_DATA"
-        android:label="@string/permission_write_nfc_data_label"
-        android:permissionGroup="commcare.permission-group.INTERNAL_ACTION"
-        android:protectionLevel="signature">
-    </permission>
-
-    <permission
-        android:name="${applicationId}.permission.DRAW_BOUNDARY_AREA"
-        android:label="@string/permission_draw_boundary_area_label"
+        android:name="${applicationId}.permission.PROTECT_FALSELY_EXPOSED_COMPONENTS"
+        android:label="@string/permission_protect_falsely_exposed_components_label"
+        android:description="@string/permission_protect_falsely_exposed_components_description"
         android:permissionGroup="commcare.permission-group.INTERNAL_ACTION"
         android:protectionLevel="signature">
     </permission>
@@ -119,13 +106,6 @@
         android:description="@string/permission_update_to_latest_build_description"
         android:permissionGroup="commcare.permission-group.EXTERNAL_ACTION"
         android:protectionLevel="dangerous">
-    </permission>
-
-    <permission
-        android:name="${applicationId}.permission.PRINT"
-        android:label="@string/permission_print_label"
-        android:permissionGroup="commcare.permission-group.INTERNAL_ACTION"
-        android:protectionLevel="signature">
     </permission>
 
     <queries>
@@ -387,7 +367,7 @@
 
         <activity
             android:name="org.commcare.gis.DrawingBoundaryActivity"
-            android:permission="org.commcare.dalvik.permission.DRAW_BOUNDARY_AREA"
+            android:permission="org.commcare.dalvik.permission.PROTECT_FALSELY_EXPOSED_COMPONENTS"
             android:exported="true">
             <intent-filter>
                 <action android:name="org.commcare.dalvik.action.DrawBoundary"/>
@@ -443,7 +423,7 @@
         <activity
             android:exported="true"
             android:name="org.commcare.android.nfc.NfcWriteActivity"
-            android:permission="org.commcare.dalvik.permission.WRITE_NFC_DATA">
+            android:permission="org.commcare.dalvik.permission.PROTECT_FALSELY_EXPOSED_COMPONENTS">
             <intent-filter>
                 <action android:name="org.commcare.nfc.WRITE"/>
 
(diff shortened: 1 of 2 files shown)


shubham1g5:
@avazirna can you link the Android docs reg. restriction you are talking about in PR description.

avazirna:
> @avazirna can you link the Android docs reg. restriction you are talking about in PR description.

done @shubham1g5, we might need to consider having an option to allow app callout questions to use explicit intents.

shubham1g5:
@avazirna I think we need to implement a permission at the very least for these exported components as we don't really want to allow other apps to have access to these.

avazirna:
@coderabbitai review

app/AndroidManifest.xml
```diff
@@ -87,6 +92,42 @@
         android:protectionLevel="dangerous">
     </permission>
 
+    <permission
+        android:name="${applicationId}.permission.READ_NFC_DATA"
```
shubham1g5, line 96: @avazirna Here the permission `READ_NFC_DATA` doesn't make sense since the purpose of this permission is not to give an external app permission to read nfc data (we don't have that functionality for external apps) but merely to protect the app's exposed components that should not ideally be exposed in the first place. As such we should have one single permission like `PROTECT_FALSELY_EXPOSED_COMPONENTS` and apply it to all such components.

shubham1g5 (changes requested):

app/AndroidManifest.xml
avazirna: Makes sense having an umbrella permission for these, so this will be applied to `READ_NFC_DATA`, `WRITE_NFC_DATA`, `DRAW_BOUNDARY_AREA` and `PRINT`

avazirna:
Applied the same permission to `DataChangeLogsActivity`

app/AndroidManifest.xml
```diff
@@ -438,7 +462,8 @@
 
         <receiver
             android:name="org.commcare.provider.RefreshToLatestBuildReceiver"
-            android:exported="false">
+            android:permission="org.commcare.dalvik.permission.UPDATE_TO_LATEST_BUILD"
```
shubham1g5, line 465: `RefreshToLatestBuildReceiver` is a internal dev only tool and introducing a permission here will break the underlying scripts. Given it was marked as `exported=false` earlier, is there a good reason change that to `true` now ?

app/AndroidManifest.xml
```diff
@@ -527,8 +552,9 @@
         <activity android:name="org.commcare.activities.PromptCczUpdateActivity"/>
         <activity
             android:label="@string/title_data_change_logs_activity"
-            android:exported="false"
-            android:name="org.commcare.activities.DataChangeLogsActivity">
+            android:exported="true"
```
shubham1g5, line 555: why do we need to export this activity ? (I don't recall using it in callouts)

app/AndroidManifest.xml
avazirna: I might not have enough context here, but my understanding is that this receiver is for external use, its documentation also states that it should respond to requests via `adb`, in that case it needs to be exported.

> Trigger from command line with:
>  *   adb shell am broadcast -a org.commcare.dalvik.api.action.RefreshToLatestBuildAction

Considering that this is for dev use only, what's the usage? To make sure the feature is working as expected?

shubham1g5: It's a adb command to update the app to latest without interacting with UI and was meant to increase dev efficiency - https://github.com/dimagi/commcare-android/blob/master/scripts/ccc#L63

app/AndroidManifest.xml
avazirna: It's not, but it's supposed to respond to `ACTION_VIEW` actions, so I think it needs to be exported. I didn't test this though, let me do that.

avazirna: Btw from the [Spec](https://docs.google.com/document/d/1LJIaQMoex6SM90gAxJbLRsMw8FC_0FtJLbMbmxGE5dE/edit?tab=t.0), I don't see the reason for this activity to respond o `ACTION_VIEW` actions, can you provide some context here?

shubham1g5: I think that's again because of dev tooling - https://github.com/dimagi/commcare-android/blob/master/scripts/ccc#L217 which I am fine with killing.

app/AndroidManifest.xml
avazirna: Deleting this receiver then, thanks

app/AndroidManifest.xml
avazirna: @shubham1g5 I've reverted the `exported` status of these components, I will kill them or move to a different manifest file in a different PR.

app/res/values/strings.xml
```diff
@@ -153,6 +153,11 @@
     <string name="permission_commcare_logout_label">Allow Auto CommCare Logout</string>
     <string name="permission_commcare_logout_description">Allows this app to logout a user from CommmCare App</string>
 
+    <string name="permission_protect_falsely_exposed_components_label">Protect falsely exposed components</string>
+    <string name="permission_protect_falsely_exposed_components_description">Due to implicit intent constrains, certain intent-aware components are now exposed, this permission is to prevent abusive use by external apps</string>
+
+    <string name="permission_group_internal_action_label">Permission group for CommCare Internal Actions</string>
```
shubham1g5, line 159: can we try to simply these strings and make them user-friendly (remove any tech jargain like intents/callouts) as these may be presented to the users.

shubham1g5 (approved):

app/res/values/strings.xml
avazirna: Done [abe2a90](https://github.com/dimagi/commcare-android/pull/2930/commits/abe2a902dfc8701c9d6e2a61c74fc1203cae328f)

PR #2944, opened 2025-01-21
Expires entity cache for related cases

shubham1g5:
## Summary

[Spec](https://docs.google.com/document/d/1VE6yjo-yOfhnwpHMRZWavgJmbmbRhsoVvWS3Kq3yh2g/edit?tab=t.0#heading=h.ivv3e8sh6t45)

Expires cache for all related cases in order to support caching xpath calculation results for related cases like `instance('casedb')/casedb/case[@case_type='mother'][@case_id=current()/index/parent]/case_name`. Without expiring related cases from cache, these kind of expressions may result in obsolete data to the user. 

## Feature Flag

CACHE_AND_INDEX

## PR Checklist

### Automated test coverage

PR adds a test to demonstrate desired cache expiration. 

### Safety story

To remove the cache for related cases, we are doing a [lookup in the case index table](https://github.com/dimagi/commcare-android/blob/dddccd8acd7c6d98a8dfe75b85367b0d18f59f5e/app/src/org/commcare/models/database/user/models/AndroidCaseIndexTable.java#L212), every time a case gets changed which may have a performance impact. We currently only [index the case index table](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/models/database/user/models/AndroidCaseIndexTable.java#L63) on case_id and index_name columns but not on target case and I am wondering whether we should add an index on target case to make these lookups much faster ?

cross-request: https://github.com/dimagi/commcare-core/pull/1456

### DIFF
--- a/app/src/org/commcare/xml/AndroidBulkCaseXmlParser.java
+++ b/app/src/org/commcare/xml/AndroidBulkCaseXmlParser.java
@@ -1,5 +1,7 @@
 package org.commcare.xml;
 
+import static org.commcare.xml.AndroidCaseXmlParser.clearEntityCacheHelper;
+
 import net.sqlcipher.database.SQLiteDatabase;
 
 import org.commcare.CommCareApplication;
@@ -15,6 +17,7 @@ import org.xmlpull.v1.XmlPullParserException;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
@@ -74,6 +77,7 @@ public class AndroidBulkCaseXmlParser extends BulkProcessingCaseXmlParser {
         SQLiteDatabase db;
         db = getDbHandle();
         ArrayList<Integer> recordIdsToWipe = new ArrayList<>();
+        ArrayList<Case> casesToWipe = new ArrayList<>();
 
         db.beginTransaction();
         try {
@@ -81,10 +85,9 @@ public class AndroidBulkCaseXmlParser extends BulkProcessingCaseXmlParser {
                 Case c = writeLog.get(cid);
                 storage.write(c);
                 recordIdsToWipe.add(c.getID());
+                casesToWipe.add(c);
             }
-            if (mEntityCache != null) {
-                mEntityCache.invalidateCaches(recordIdsToWipe);
-            }
+            clearEntityCache(casesToWipe);
             mCaseIndexTable.clearCaseIndices(recordIdsToWipe);
 
             for (String cid : writeLog.keySet()) {
@@ -96,5 +99,15 @@ public class AndroidBulkCaseXmlParser extends BulkProcessingCaseXmlParser {
             db.endTransaction();
         }
 
+    }
+
+    private void clearEntityCache(ArrayList<Case> casesToWipe) {
+        if (mEntityCache != null) {
+            Set<Integer> recordIdsToWipe = new HashSet<>();
+            for (int i = 0; i < casesToWipe.size(); i++) {
+                clearEntityCacheHelper(casesToWipe.get(i), recordIdsToWipe, mCaseIndexTable, storage);
+            }
+            mEntityCache.invalidateCaches(recordIdsToWipe);
+        }
     }
 }
--- a/app/src/org/commcare/xml/AndroidCaseXmlParser.java
+++ b/app/src/org/commcare/xml/AndroidCaseXmlParser.java
@@ -9,6 +9,7 @@ import net.sqlcipher.database.SQLiteDatabase;
 import org.commcare.CommCareApplication;
 import org.commcare.android.database.user.models.ACase;
 import org.commcare.cases.model.Case;
+import org.commcare.cases.model.CaseIndex;
 import org.commcare.engine.references.JavaHttpReference;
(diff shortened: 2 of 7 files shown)


app/src/org/commcare/xml/AndroidCaseXmlParser.java
```diff
@@ -107,6 +109,46 @@ public void commit(Case parsed) throws IOException {
...
+        if (aCase == null || mEntityCache == null) {
+            return;
+        }
+        Set<Integer> recordsToWipe = new HashSet<>();
+        clearEntityCacheHelper(aCase, recordsToWipe, mCaseIndexTable, storage());
+        // Invalidate the current case in cache
+        mEntityCache.invalidateCaches(recordsToWipe);
+    }
+
+    public static void clearEntityCacheHelper(Case aCase, Set<Integer> visited, AndroidCaseIndexTable caseIndexTable,
```
ctsims, line 122: Instead of being recursive, I would update this methodology to use the [existing case index map](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/models/database/user/models/AndroidCaseIndexTable.java#L95) functionality, similar to how we generate the DAG for [the case purge algorithm](https://github.com/dimagi/commcare-android/blob/2a031f9e206a92ee013b58a1b0880ebca88202cb/app/src/org/commcare/engine/cases/CaseUtils.java#L165), and then just run the purge in a loop with the list of cases. 

That would let us keep logic (and especially optimizations) for doing these kinds of case tree walks in one place, and keep the same guarantees about the shapes of the graphs.

app/src/org/commcare/xml/AndroidBulkCaseXmlParser.java
```diff
@@ -97,4 +100,14 @@ protected void performBulkWrite(LinkedHashMap<String, Case> writeLog) throws IOE
         }
 
     }
+
+    private void clearEntityCache(ArrayList<Case> casesToWipe) {
+        if (mEntityCache != null) {
+            Set<Integer> recordIdsToWipe = new HashSet<>();
+            for (int i = 0; i < casesToWipe.size(); i++) {
+                clearEntityCacheHelper(casesToWipe.get(i), recordIdsToWipe, mCaseIndexTable, storage);
```
ctsims, line 108: The performance characteristics of these methods needs a lot more consideration, I think.

It gets really problematic for CommCare if saving a case takes a long time, since there are lots of contexts where we might update a bunch of cases in a row), and this code introduces a lot of potential reads (and writes) that could end up making what currently takes milliseconds into something that snowballs. There are definitely contexts where it's probably faster to wipe the entire cache than it is to update it record by record.

There's some chance I'm overthinking it, but having worked a lot with case subgraphs in the purge code, it is really expensive to do even one walk over the full case graph. If we were going to make a change like this I'd want to see how it affects performance of high case loads in both sparse (1 parent 1 child) type cases and in dense cases (for instance, 1 person case with 100 visit cases) 

As an alternative, I wonder if instead of updating the invalidation synchronously, we instead added the "cases needing invalidation" to a queue that can process them async, and block on the queue clearing before showing a case list, which would be no delay at all in the fast cases.

This has two other nice advantages: 
1 - with the queue of invalidations you could process them in chunks. If, for instance, if there were 5 updates pending, they could share an [index walk](https://github.com/dimagi/commcare-android/blob/2a031f9e206a92ee013b58a1b0880ebca88202cb/app/src/org/commcare/engine/cases/CaseUtils.java#L165) to accumulate related cases, and then submit all of the cache updates in one transaction. 
2 - Seeing the spool size would let us be smarter about large backlogs. IE: there was a bulk update and 5,000 cases were changed of the 5,500 on the device) were changed, instead of invalidating them one by one we could just wipe the cache

ctsims (commented):
Added a couple of initial thoughts, but I think my feedback here will need to be based in the overall direction of approach for the feature, IE: What [level of confidence](https://docs.google.com/document/d/1VE6yjo-yOfhnwpHMRZWavgJmbmbRhsoVvWS3Kq3yh2g/edit?tab=t.0#bookmark=id.o4vsg7l94cqx)  are you targeting from the spec?

app/src/org/commcare/xml/AndroidBulkCaseXmlParser.java
shubham1g5: > "cases needing invalidation" to a queue that can process them async, 

@ctsims The alternative approach you mentioned here makes a lot of sense to me. One question - How do you feel about adding a new `dirty` field to the existing Cache table as a manner of implementing this queue ? We can mark the cache as dirty for the specific cases synchronously and then asynchronously re-validate these records along with their case relation graph ?

ctsims: Ah, I think that something like that could be a really good approach! It reminds me a lot of something Emord was proposing a while ago about switching pillows from an event bus model to a stale flag processor. The fact that we're never going to have significant parallelization of the processing lends itself well (I think) to flagging this way. 

Only tricky thing is that I **think** in this case it's a "Case -> Case subgraph" extraction process that I think is the time consuming thing that should get onto the queue (IE: Determining which cases are adjacent to the one that needs to get updated). I don't think we could just put the flag onto the Cache tables, since it won't be clear at that point which of the rows to actually tag.

shubham1g5:
Closing for https://github.com/dimagi/commcare-android/pull/2955

PR #2953, opened 2025-02-12
Fix Case detail tab title

avazirna:
## Product Description
This PR fixes an issue in the Case details which causes a mismatch between tab items and their respective content when a display condition is set. The fix is basically to restore a method that was removed during the migration to ViewPager2 that retrieves the title of the page from the visible tabs.

Ticket: https://dimagi.atlassian.net/browse/SAAS-16528

### Safety story
This was tested locally and with different configurations, all seems fine. 

## Labels and Review

### DIFF
--- a/app/src/org/commcare/adapters/EntityDetailPagerAdapter.java
+++ b/app/src/org/commcare/adapters/EntityDetailPagerAdapter.java
@@ -59,6 +59,11 @@ public class EntityDetailPagerAdapter extends FragmentStateAdapter {
         return fragment;
     }
 
+    public CharSequence getPageTitle(int position) {
+        Detail detailShowing = detail.isCompound() ? displayableChildDetails[position] : detail;
+        return detailShowing.getTitle().getText().evaluate();
+    }
+
     @Override
     public int getItemCount() {
         return detail.isCompound() ? displayableChildDetails.length : 1;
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/adapters/EntityDetailPagerAdapter.java
```diff
@@ -59,6 +59,11 @@ public Fragment createFragment(int position) {
         return fragment;
     }
 
+    public CharSequence getPageTitle(int position) {
```
shubham1g5, line 62: is there a reason this method is here and not in `TabbedDetailView` itself ?

shubham1g5 (approved):

app/src/org/commcare/adapters/EntityDetailPagerAdapter.java
avazirna: It might be a stretch but the reason is that the adapter is responsible for providing the content of the tabs and can also provide their titles. The structural aspect is that the [displayableChildDetails](https://github.com/dimagi/commcare-android/blob/d3f29f1cc1bce0fdaf14d8c5d4ce00f64c464c6d/app/src/org/commcare/adapters/EntityDetailPagerAdapter.java#L30) which stores the visible detail children and we would need to refactor this logic which I'm not sure if it's worth the effort. Any thoughts?

PR #2955, opened 2025-02-18
Expires entity cache for related cases

shubham1g5:
## Product Description
[Spec](https://docs.google.com/document/d/1VE6yjo-yOfhnwpHMRZWavgJmbmbRhsoVvWS3Kq3yh2g/edit?tab=t.0#heading=h.ivv3e8sh6t45)

https://dimagi.atlassian.net/browse/SC-4279

Expires cache for all related cases in order to support caching xpath calculation results for related cases like instance('casedb')/casedb/case[@case_type='mother'][@case_id=current()/index/parent]/case_name. Without expiring related cases from cache, these kind of expressions may result in obsolete data to the user.

## Technical Summary

Alternate approach to https://github.com/dimagi/commcare-android/pull/2944

Adds a new field `is_shallow` in the entity cache to signify any cases that have changed from the last time the cache was primed. These shallow records are then further processed by a background worker after form submission and restores to delete any cache records related to these cases or their related cases. 

## Feature Flag

CACHE_AND_LAZY_LOAD

### Safety story

should only affect projects implementing caching for case list. 

### Automated test coverage
Adds a test to verify the cache expiration behavior

### QA Plan
This whole feature will be QA'ed as part of mobile release cycle.

## Labels and Review

- [x] Does the PR introduce any major changes worth communicating ? If yes, the "Release Note" label is set and a "Release Note" is specified in PR description.

cross-request: https://github.com/dimagi/commcare-core/pull/1460

### DIFF
--- a/app/src/org/commcare/activities/HomeScreenBaseActivity.java
+++ b/app/src/org/commcare/activities/HomeScreenBaseActivity.java
@@ -24,10 +24,6 @@ import android.widget.Toast;
 import androidx.appcompat.app.AlertDialog;
 import androidx.appcompat.app.AppCompatActivity;
 import androidx.preference.PreferenceManager;
-import androidx.work.ExistingWorkPolicy;
-import androidx.work.OneTimeWorkRequest;
-import androidx.work.OutOfQuotaPolicy;
-import androidx.work.WorkManager;
 
 import com.google.android.play.core.install.model.InstallErrorCode;
 
@@ -74,9 +70,7 @@ import org.commcare.suite.model.SessionDatum;
 import org.commcare.suite.model.StackFrameStep;
 import org.commcare.suite.model.Text;
 import org.commcare.sync.FirebaseMessagingDataSyncer;
-import org.commcare.sync.FormSubmissionWorker;
 import org.commcare.tasks.DataPullTask;
-import org.commcare.tasks.EntityCacheInvalidationWorker;
 import org.commcare.tasks.FormLoaderTask;
 import org.commcare.tasks.FormRecordCleanupTask;
 import org.commcare.tasks.ResultAndError;
@@ -923,7 +917,7 @@ public abstract class HomeScreenBaseActivity<T> extends SyncCapableCommCareActiv
             // The form is either ready for processing, or not, depending on how it was saved
             if (complete) {
                 startUnsentFormsTask(false, false);
-                CommCareApplication.instance().propogateEntityCacheInvalidation();
+                CommCareApplication.instance().scheduleEntityCacheInvalidation();
                 refreshUI();
 
                 if (exitFromExternalLaunch()) {
--- a/app/src/org/commcare/activities/HomeScreenBaseActivity.java
+++ b/app/src/org/commcare/activities/HomeScreenBaseActivity.java
@@ -24,6 +24,10 @@ import android.widget.Toast;
 import androidx.appcompat.app.AlertDialog;
 import androidx.appcompat.app.AppCompatActivity;
 import androidx.preference.PreferenceManager;
+import androidx.work.ExistingWorkPolicy;
+import androidx.work.OneTimeWorkRequest;
+import androidx.work.OutOfQuotaPolicy;
+import androidx.work.WorkManager;
 
 import com.google.android.play.core.install.model.InstallErrorCode;
 
@@ -70,7 +74,9 @@ import org.commcare.suite.model.SessionDatum;
 import org.commcare.suite.model.StackFrameStep;
 import org.commcare.suite.model.Text;
 import org.commcare.sync.FirebaseMessagingDataSyncer;
+import org.commcare.sync.FormSubmissionWorker;
 import org.commcare.tasks.DataPullTask;
+import org.commcare.tasks.EntityCacheInvalidationWorker;
 import org.commcare.tasks.FormLoaderTask;
 import org.commcare.tasks.FormRecordCleanupTask;
 import org.commcare.tasks.ResultAndError;
@@ -917,6 +923,7 @@ public abstract class HomeScreenBaseActivity<T> extends SyncCapableCommCareActiv
             // The form is either ready for processing, or not, depending on how it was saved
             if (complete) {
                 startUnsentFormsTask(false, false);
+                CommCareApplication.instance().propogateEntityCacheInvalidation();
                 refreshUI();
 
                 if (exitFromExternalLaunch()) {
--- a/app/src/org/commcare/activities/HomeScreenBaseActivity.java
+++ b/app/src/org/commcare/activities/HomeScreenBaseActivity.java
@@ -917,7 +917,6 @@ public abstract class HomeScreenBaseActivity<T> extends SyncCapableCommCareActiv
             // The form is either ready for processing, or not, depending on how it was saved
             if (complete) {
                 startUnsentFormsTask(false, false);
-                CommCareApplication.instance().scheduleEntityCacheInvalidation();
                 refreshUI();
 
                 if (exitFromExternalLaunch()) {
--- a/app/src/org/commcare/engine/cases/CaseUtils.java
+++ b/app/src/org/commcare/engine/cases/CaseUtils.java
@@ -79,38 +79,12 @@ public class CaseUtils {
     public static void purgeCases() throws InvalidCaseGraphException {
         long start = System.currentTimeMillis();
         //We need to determine if we're using ownership for purging. For right now, only in sync mode
-        Vector<String> owners = new Vector<>();
-        Vector<String> users = new Vector<>();
-        for (IStorageIterator<User> userIterator = CommCareApplication.instance()
-                .getUserStorage(User.STORAGE_KEY, User.class).iterate(); userIterator.hasMore(); ) {
-            String id = userIterator.nextRecord().getUniqueId();
-            owners.addElement(id);
-            users.addElement(id);
-        }
-
-        //Now add all of the relevant groups
-        //TODO: Wow. This is.... kind of megasketch
-        for (String userId : users) {
-            DataInstance instance = CommCareUtil.loadFixture("user-groups", userId);
-            if (instance == null) {
-                continue;
-            }
-            EvaluationContext ec = new EvaluationContext(instance);
-            for (TreeReference ref : ec.expandReference(XPathReference.getPathExpr("/groups/group/@id").getReference())) {
-                AbstractTreeElement<AbstractTreeElement> idelement = ec.resolveReference(ref);
-                if (idelement.getValue() != null) {
-                    owners.addElement(idelement.getValue().uncast().getString());
-                }
-            }
-        }
-
-        SQLiteDatabase db;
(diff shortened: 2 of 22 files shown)


shubham1g5:

app/src/org/commcare/models/database/user/models/CommCareEntityStorageCache.java
```diff
@@ -123,19 +133,35 @@ public String retrieveCacheValue(String entityKey, String cacheKey) {
...
-    public void invalidateCache(String recordId) {
-        int removed = db.delete(TABLE_NAME, COL_CACHE_NAME + " = ? AND " + COL_ENTITY_KEY + " = ?", new String[]{mCacheName, recordId});
-        if (SqlStorage.STORAGE_OUTPUT_DEBUG) {
-            Log.d(TAG, "Invalidated " + removed + " cached values for entity " + recordId);
+    public void invalidateRecord(String recordId) {
+        if (isEmpty()) {
+            return;
+        }
+        markRecordsAsShallow(ImmutableList.of(Integer.parseInt(recordId)));
+    }
```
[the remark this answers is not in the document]
shubham1g5: we never anticipate recordId to be non numeric and would like to crash fast and hard if it's not defined so.

app/src/org/commcare/engine/cases/CaseUtils.java
```diff
@@ -151,9 +126,77 @@ public static void purgeCases() throws InvalidCaseGraphException {
...
+            String relatedRecordId = getRecordIdFromCaseId(storage, relatedCaseId);
+            relatedRecordIds.add(relatedRecordId);
+        }
+        return relatedRecordIds;
+    }
+
+    private static String getRecordIdFromCaseId(SqlStorage<ACase> storage, String caseId) {
+        Vector<Integer> result = storage.getIDsForValue(INDEX_CASE_ID, caseId);
+        return String.valueOf(result.get(0));
+    }
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/2955/commits/ed0d420ddb69a40345e6cb047457b7faeae42a0b

app/src/org/commcare/activities/HomeScreenBaseActivity.java
```diff
@@ -917,6 +917,7 @@ private boolean processReturnFromFormEntry(int resultCode, Intent intent) {
             // The form is either ready for processing, or not, depending on how it was saved
             if (complete) {
                 startUnsentFormsTask(false, false);
+                CommCareApplication.instance().scheduleEntityCacheInvalidation();
```
ctsims, line 920: Two quick things:

1- If the form is 'complete' and not 'unsent' that means it wasn't processed, right? A little concerned about the possibility that the UnsentForms task might not have finished processing the form before the actual underlying work here is performed. 

2 - Seems a little odd to queue this if there is no caching going on with the current app. I still don't think we're ready to view caching as the default, so I'm interested  (where it's not a ton of work) in still minimizing the new surface area non-cache-and-index apps are exposed to from the changes.

app/src/org/commcare/engine/cases/CaseUtils.java
```diff
@@ -79,38 +80,12 @@ public static String computeCaseDbHash(SqlStorage<?> storage) {
...
-        SQLiteDatabase db;
-        db = CommCareApplication.instance().getUserDbHandle();
-
-        db.beginTransaction();
+        Vector<String> owners = getAllOwners();
+        SQLiteDatabase db = CommCareApplication.instance().getUserDbHandle();
         int removedCaseCount;
         int removedLedgers;
         try {
+            db.beginTransaction();
```
ctsims, line 88: I believe our pattern is for these to stay [outside of the transaction](https://github.com/dimagi/commcare-android/blob/d3f29f1cc1bce0fdaf14d8c5d4ce00f64c464c6d/app/src/org/commcare/xml/AndroidBulkCaseXmlParser.java#L61) when being used as `begin/success/end`, otherwise if this method itself throws an error and the transaction wasn't started, the `endTransaction` call could unintentionally end a separate transaction instead of this one.

app/src/org/commcare/engine/cases/CaseUtils.java
```diff
@@ -79,38 +80,12 @@ public static String computeCaseDbHash(SqlStorage<?> storage) {
...
-                    owners.addElement(idelement.getValue().uncast().getString());
-                }
-            }
-        }
-
-        SQLiteDatabase db;
-        db = CommCareApplication.instance().getUserDbHandle();
-
-        db.beginTransaction();
+        Vector<String> owners = getAllOwners();
```
ctsims, line 83: Thanks for adding all of these little refactors.

app/src/org/commcare/engine/cases/CaseUtils.java
```diff
@@ -151,9 +126,80 @@ public static void purgeCases() throws InvalidCaseGraphException {
...
+        if (recordIds.isEmpty()) {
+            return recordIds;
+        }
+
+        SqlStorage<ACase> storage = CommCareApplication.instance().getUserStorage(ACase.STORAGE_KEY, ACase.class);
+        DAG<String, int[], String> fullCaseGraph = getFullCaseGraph(storage, new AndroidCaseIndexTable(),
+                new Vector<>());
+        Set<String> caseIds = getCaseIdsFromRecordsIds(storage, recordIds);
+        Set<String> relatedCaseIds = fullCaseGraph.getRelatedRecords(caseIds);
+        Set<String> relatedRecordIds = new HashSet<>();
```
ctsims, line 174: Fully optional, but I bet our current version of java might have a way to apply this MAP operation with an internal collections lambda or something that might be a little more terse.

app/src/org/commcare/engine/cases/CaseUtils.java
```diff
@@ -151,9 +126,80 @@ public static void purgeCases() throws InvalidCaseGraphException {
...
+        for (String relatedCaseId : relatedCaseIds) {
+            String relatedRecordId = getRecordIdFromCaseId(storage, relatedCaseId);
```
ctsims, line 176: When this set of ID's is large (which it might be fairly often) it will be very expensive to be doing these reads one-by-one.

There are a few options for how to do this faster. **One** of the methods above that does a trawl of the full sets of case data must have been able to retain this mapping, so just keeping it around might be enough, but otherwise doing a [bulk read](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/services/storage/IStorageUtilityIndexed.java#L280) type operator is a much more efficient option.

app/src/org/commcare/engine/cases/CaseUtils.java
```diff
@@ -151,9 +126,80 @@ public static void purgeCases() throws InvalidCaseGraphException {
...
+        }
+        return relatedRecordIds;
+    }
+
+    private static String getRecordIdFromCaseId(SqlStorage<ACase> storage, String caseId) {
+        Vector<Integer> result = storage.getIDsForValue(INDEX_CASE_ID, caseId);
+        if (result == null || result.isEmpty()) {
+            throw new IllegalStateException("No record found for case ID: " + caseId);
+        }
+        return String.valueOf(result.get(0));
```
ctsims, line 187: I'm a little confused a to why these ints are getting string casts in/out a bunch. I seems like this is could cause a fair amount of unnecessary memory churn (strings aren't exactly efficient), but would also communicate more clearly if these signatures stayed in integer as much as possible to prevent confusion between CaseID and RecordID.

app/src/org/commcare/models/database/user/models/CommCareEntityStorageCache.java
```diff
@@ -231,6 +274,15 @@ public void primeCache(Hashtable<String, AsyncEntity> entitySet, String[][] cach
         }
     }
 
+    /**
+     * Gets all shallow records and it's related graph and delete those records from the cache
+     */
+    public void processShallowRecords() {
```
ctsims, line 280: Since this is getting schedule asynchronously, I think this method itself might need to be thread protected or something should prevent simultaneous management of the cache while it's happening.

ctsims (changes requested):
Added some points of feedback but most of them are performance related. 

I'm not sure if we know what the spanning set of people using Cache and Index is, but I'm generally pretty worried that this set of changes might have a significant impact on where performance is felt for a current C&I user. 

Not all of my feedback here would be mandatory to process upfront if there was a clean way to disable the caching cleaning strategy if it's having significant impacts on runtime, since I do think that creating new mandatory full walks of the DB can be risky.

app/src/org/commcare/activities/HomeScreenBaseActivity.java
shubham1g5: 1. https://github.com/dimagi/commcare-android/pull/2955/commits/867c25462ac0264a267a21b97cd007ee864a3347
2. https://github.com/dimagi/commcare-android/pull/2955/commits/be429057100fee52a003f241385a4ed77d8e2d2c

app/src/org/commcare/engine/cases/CaseUtils.java
shubham1g5: https://github.com/dimagi/commcare-android/pull/2955/commits/78f2e5ddd95df21034763976fe824899f6395232

app/src/org/commcare/models/database/user/models/CommCareEntityStorageCache.java
shubham1g5: https://github.com/dimagi/commcare-android/pull/2955/commits/4d1e04bf1109939598212b2450870e0982911be5

app/src/org/commcare/engine/cases/CaseUtils.java
shubham1g5: https://github.com/dimagi/commcare-android/pull/2955/commits/fdcabfc9a69461a2a53297d2cd1de09ed8cf7d22

app/src/org/commcare/engine/cases/CaseUtils.java
shubham1g5: https://github.com/dimagi/commcare-android/pull/2955/commits/23836b694221cbc92fa5c894e4eec429c5fc6d1c

app/src/org/commcare/engine/cases/CaseUtils.java
shubham1g5: outdated by https://github.com/dimagi/commcare-android/pull/2955/commits/23836b694221cbc92fa5c894e4eec429c5fc6d1c

shubham1g5:
> I'm not sure if we know what the spanning set of people using Cache and Index is, but I'm generally pretty worried that this set of changes might have a significant impact on where performance is felt for a current C&I user.

@ctsims If required, we can do a check for new cache and index config on the app before scheduling cache invalidation, but given the expensive invalidation is part of background processing, I was not too worried about it.  (the synchronous cache invalidation in case background processing didn't happen in time only happens for new cache and index config)

ctsims (approved):
I think still some things to work through on the core idea but nothing blocking from me for now

PR #2957, opened 2025-02-21
Delete SSO token storage when token auth fails

OrangeAndGreen:
## Technical Summary
https://dimagi.atlassian.net/browse/CCCT-666

When an API call fails due to a 401 error while using TokenAuth, we should delete the token from local storage.
This is most likely to occur due to a clock error (drift or time zone mismatch) on the device.
For whatever reason the token failed, it should be discarded so a new token is retrieved the next time an API call is made.

This code addresses token failures for both HQ and ConnectID tokens in several places:
- [Checking response for HQ calls](https://github.com/dimagi/commcare-android/blob/5113bd1950004148398a740c04add1016ceb242c/app/src/org/commcare/network/CommcareRequestGenerator.java#L215)
- [ResponseProcessor for Connect calls using ConnectID token](https://github.com/dimagi/commcare-android/blob/5113bd1950004148398a740c04add1016ceb242c/app/src/org/commcare/connect/network/ConnectNetworkHelper.java#L356)
- [Failure getting new HQ token using ConnectID token](https://github.com/dimagi/commcare-android/blob/5113bd1950004148398a740c04add1016ceb242c/app/src/org/commcare/connect/network/ApiConnectId.java#L126)

## Feature Flag
ConnectID

### Safety story
Tested by performing a successful HQ API call with an app managed by ConnectID, to get valid ConnectID and HQ tokens.
Then waited for the tokens to expire
Manually set date on the device to the day before so it thought the tokens were still valid
Tried to make a Connect API, observed the 401 call and discarded token, observed retry on the call succeeding (including new token retrieval).
Repeated the test from the beginning (new tokens, then expired), then attempted a managed HQ API call and observed the same results as previous test.

The main idea here is that tokens are pretty cheap, so if a token isn't good when we thought it would be we should just go get a new one and expect things to work at that point.

In normal usage where tokens are managed correctly (no clock errors resulting in using expired tokens), there will be no effect from this new change.

### Automated test coverage

### QA Plan
QA should attempt similar steps for testing to the ones described above, i.e. manually altering the clock to trick the device into attempting an API call with one or both bad tokens.

## Labels and Review

- [ x ] Do we need to enhance the manual QA test coverage ? If yes, the "QA Note" label is set correctly

### DIFF
--- a/app/src/org/commcare/android/database/connect/models/ConnectJobAssessmentRecord.java
+++ b/app/src/org/commcare/android/database/connect/models/ConnectJobAssessmentRecord.java
@@ -1,7 +1,6 @@
 package org.commcare.android.database.connect.models;
 
 import org.commcare.android.storage.framework.Persisted;
-import org.commcare.connect.network.ConnectNetworkHelper;
 import org.commcare.models.framework.Persisting;
 import org.commcare.modern.database.Table;
 import org.commcare.modern.models.MetaField;
--- a/app/src/org/commcare/android/database/connect/models/ConnectJobDeliveryRecord.java
+++ b/app/src/org/commcare/android/database/connect/models/ConnectJobDeliveryRecord.java
@@ -1,7 +1,6 @@
 package org.commcare.android.database.connect.models;
 
 import org.commcare.android.storage.framework.Persisted;
-import org.commcare.connect.network.ConnectNetworkHelper;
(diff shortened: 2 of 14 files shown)


OrangeAndGreen:
Made these changes to the wrong branch. Corrected with  different branch

PR #2958, opened 2025-02-21
Delete SSO token storage when token auth fails

OrangeAndGreen:
## Technical Summary
https://dimagi.atlassian.net/browse/CCCT-666

When an API call fails due to a 401 error while using TokenAuth, we should delete the token from local storage.
This is most likely to occur due to a clock error (drift or time zone mismatch) on the device.
For whatever reason the token failed, it should be discarded so a new token is retrieved the next time an API call is made.

This code addresses token failures for both HQ and ConnectID tokens in several places:
- [Checking response for HQ calls](https://github.com/dimagi/commcare-android/blob/5113bd1950004148398a740c04add1016ceb242c/app/src/org/commcare/network/CommcareRequestGenerator.java#L215)
- [ResponseProcessor for Connect calls using ConnectID token](https://github.com/dimagi/commcare-android/blob/5113bd1950004148398a740c04add1016ceb242c/app/src/org/commcare/connect/network/ConnectNetworkHelper.java#L356)
- [Failure getting new HQ token using ConnectID token](https://github.com/dimagi/commcare-android/blob/5113bd1950004148398a740c04add1016ceb242c/app/src/org/commcare/connect/network/ApiConnectId.java#L126)

## Feature Flag
ConnectID

### Safety story
Tested by performing a successful HQ API call with an app managed by ConnectID, to get valid ConnectID and HQ tokens.
Then waited for the tokens to expire
Manually set date on the device to the day before so it thought the tokens were still valid
Tried to make a Connect API, observed the 401 call and discarded token, observed retry on the call succeeding (including new token retrieval).
Repeated the test from the beginning (new tokens, then expired), then attempted a managed HQ API call and observed the same results as previous test.

The main idea here is that tokens are pretty cheap, so if a token isn't good when we thought it would be we should just go get a new one and expect things to work at that point.

In normal usage where tokens are managed correctly (no clock errors resulting in using expired tokens), there will be no effect from this new change.

### Automated test coverage

### QA Plan
QA should attempt similar steps for testing to the ones described above, i.e. manually altering the clock to trick the device into attempting an API call with one or both bad tokens.

## Labels and Review

### DIFF
--- a/app/src/org/commcare/connect/network/ApiConnectId.java
+++ b/app/src/org/commcare/connect/network/ApiConnectId.java
@@ -77,7 +77,7 @@ public class ApiConnectId {
         ConnectNetworkHelper.PostResult postResult = ConnectNetworkHelper.postSync(context, url,
                 API_VERSION_NONE, new AuthInfo.NoAuth(), params, true, false);
         Logger.log(LogTypes.TYPE_MAINTENANCE, "OAuth Token Post Result " + postResult.responseCode);
-        if (postResult.responseCode >= 200 && postResult.responseCode < 300>) {
+        if (postResult.responseCode >= 200 && postResult.responseCode < 300) {
             try {
                 String responseAsString = new String(StreamsUtil.inputStreamToByteArray(
                         postResult.responseStream));
--- a/app/src/org/commcare/connect/network/ApiConnectId.java
+++ b/app/src/org/commcare/connect/network/ApiConnectId.java
@@ -77,7 +77,7 @@ public class ApiConnectId {
         ConnectNetworkHelper.PostResult postResult = ConnectNetworkHelper.postSync(context, url,
                 API_VERSION_NONE, new AuthInfo.NoAuth(), params, true, false);
         Logger.log(LogTypes.TYPE_MAINTENANCE, "OAuth Token Post Result " + postResult.responseCode);
-        if (postResult.responseCode == 200 || postResult.responseCode == 201) {
+        if (postResult.responseCode >= 200 && postResult.responseCode < 300>) {
             try {
                 String responseAsString = new String(StreamsUtil.inputStreamToByteArray(
                         postResult.responseStream));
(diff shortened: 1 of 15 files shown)


app/src/org/commcare/connect/network/ApiConnectId.java
```diff
@@ -77,7 +77,7 @@ public static AuthInfo.TokenAuth retrieveHqTokenApi(Context context, String hqUs
         ConnectNetworkHelper.PostResult postResult = ConnectNetworkHelper.postSync(context, url,
                 API_VERSION_NONE, new AuthInfo.NoAuth(), params, true, false);
         Logger.log(LogTypes.TYPE_MAINTENANCE, "OAuth Token Post Result " + postResult.responseCode);
-        if (postResult.responseCode == 200) {
+        if (postResult.responseCode == 200 || postResult.responseCode == 201) {
```
shubham1g5, line 80: probably better to check for `responseCode >= 200 && responseCode < 300` as they all mean `success`

shubham1g5 (commented):
The code here looks good to me but I wonder if we should trigger the process to get a new token when we get a auth error ? Correct me if I am wrong but it seems like the request would just fail without us retrying to do the request with a new token ?

app/src/org/commcare/connect/network/ApiConnectId.java
OrangeAndGreen: [97568a1](https://github.com/dimagi/commcare-android/pull/2958/commits/97568a157b1fa29c09f5172b3c54a9c5efafad68)

app/src/org/commcare/connect/network/ApiConnectId.java
```diff
@@ -77,7 +77,7 @@ public static AuthInfo.TokenAuth retrieveHqTokenApi(Context context, String hqUs
         ConnectNetworkHelper.PostResult postResult = ConnectNetworkHelper.postSync(context, url,
                 API_VERSION_NONE, new AuthInfo.NoAuth(), params, true, false);
         Logger.log(LogTypes.TYPE_MAINTENANCE, "OAuth Token Post Result " + postResult.responseCode);
-        if (postResult.responseCode == 200) {
+        if (postResult.responseCode >= 200 && postResult.responseCode < 300>) {
```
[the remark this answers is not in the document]
OrangeAndGreen: Oops, [f16a26e](https://github.com/dimagi/commcare-android/pull/2958/commits/f16a26e9c42732ef8e33d748076f386a68d59d38)

OrangeAndGreen:
Per discussion with Clayton, this seems like something that should happen rarely enough that we can just fail the call and try again later with new tokens rather than trying to support the complex logic of nesting one API call in several more.

Although Jignesh mentioned that for Retrofit calls (so to Connect, not HQ) we can configure an interceptor to handle the 401 and automatically attempt to get a new token, so I think that would be another good addition to schedule soon.

shubham1g5 (approved):

PR #2963, opened 2025-02-25
Remove media permissions

avazirna:
## Product Description
Google issued a new policy that states that [READ_MEDIA_IMAGES](https://developer.android.com/reference/android/Manifest.permission#READ_MEDIA_IMAGES) and [READ_MEDIA_VIDEO](https://developer.android.com/reference/android/Manifest.permission#READ_MEDIA_VIDEO) permissions can only be used by apps whose core functionality requires broad access to images and videos stored in the device. These permissions were introduced in Android 13 and as such, this policy mostly affects that version and later. For apps that don't require this level of access, the recommended approach is to use a system picker, like the [Android photo picker](https://developer.android.com/training/data-storage/shared/photopicker), which is available in Android 13 onward. Additionally, Google is committed to making the photo picker the default behaviour and is working to make it available to older versions too:
- Android 11 and 12 - through Google System updates
- Android 4.4 to 10 - via a backported version distributed through Google Play Services.

This PR implements the necessary changes to comply with this policy considering that CommCare doesn't need broad access to media files. 
**Note to reviewers:** this work also involved assessing whether to adopt the _ActivityResult API_, since _startActivityForResult_ is now deprecated. However, these two approaches can't be mixed, meaning that migrating would require the definition of ActivitResult contracts for every option in the `FormEntryActivity.onAtivityResult` method.

### Safety story
This changes the current approach when selecting media files to upload in CommCare, it sets preference to system photo and video pickers that can handle the [ACTION_PICK_IMAGE](https://developer.android.com/reference/android/provider/MediaStore#ACTION_PICK_IMAGES) action, falling back to `ACTION_GET_CONTENT` when an appropriate component is not available, so is not supposed to break anything. Moreover, these change were tested on Android 5, 6, 8, 9, 10, 11, 12 and 13

### QA Plan
A QA label has been  added to ensure that this gets tested on both  the minimum (Android 5) and maximum (Android 14) supported versions.

## Labels and Review

cross-request: https://github.com/dimagi/commcare-core/pull/1462

### DIFF
--- a/app/src/org/commcare/views/widgets/ImageWidget.java
+++ b/app/src/org/commcare/views/widgets/ImageWidget.java
@@ -136,11 +136,10 @@ public class ImageWidget extends QuestionWidget {
                 ImageCaptureProcessing.setCustomImagePath(null);
             } else {
                 mErrorTextView.setVisibility(View.GONE);
-                Intent i = new Intent(Intent.ACTION_GET_CONTENT);
-                i.setType("image/*");
 
                 try {
-                    ((AppCompatActivity)getContext()).startActivityForResult(i,
+                    ((AppCompatActivity)getContext()).startActivityForResult(
+                            WidgetUtils.createPickMediaIntent (getContext(), "image/*"),
                             FormEntryConstants.IMAGE_CHOOSER);
                     pendingCalloutInterface.setPendingCalloutFormIndex(mPrompt.getIndex());
                 } catch (ActivityNotFoundException e) {
--- a/app/src/org/commcare/views/widgets/ImageWidget.java
+++ b/app/src/org/commcare/views/widgets/ImageWidget.java
@@ -138,9 +138,9 @@ public class ImageWidget extends QuestionWidget {
                 mErrorTextView.setVisibility(View.GONE);
 
                 try {
-                    ((AppCompatActivity)getContext()).startActivityForResult(
-                            WidgetUtils.createPickMediaIntent (getContext(), "image/*"),
-                            FormEntryConstants.IMAGE_CHOOSER);
+                    ((AppCompatActivity)getContext())
+                            .startActivityForResult(WidgetUtils.createPickMediaIntent (getContext(), "image/*"),
+                                    FormEntryConstants.IMAGE_CHOOSER);
                     pendingCalloutInterface.setPendingCalloutFormIndex(mPrompt.getIndex());
                 } catch (ActivityNotFoundException e) {
                     Toast.makeText(getContext(),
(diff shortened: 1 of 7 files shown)


avazirna:

app/src/org/commcare/views/widgets/ImageWidget.java
```diff
@@ -136,11 +136,10 @@ public ImageWidget(final Context context, FormEntryPrompt prompt, PendingCallout
                 ImageCaptureProcessing.setCustomImagePath(null);
             } else {
                 mErrorTextView.setVisibility(View.GONE);
-                Intent i = new Intent(Intent.ACTION_GET_CONTENT);
-                i.setType("image/*");
 
                 try {
-                    ((AppCompatActivity)getContext()).startActivityForResult(i,
+                    ((AppCompatActivity)getContext()).startActivityForResult(
+                            WidgetUtils.createPickMediaIntent (getContext(), "image/*"),
```
shubham1g5, line 142: code formatting seems off

app/src/org/commcare/views/widgets/WidgetUtils.java
```diff
@@ -75,4 +78,22 @@ public static Intent createScanIntent(Context context, String... formats) {
...
+            intent.setAction(MediaStore.ACTION_PICK_IMAGES);
+        } else {
+            intent.setAction(Intent.ACTION_GET_CONTENT);
+        }
+        return intent.setType(mimeType);
+    }
+
+    @SuppressLint("InlinedApi")
+    public static boolean isPhotoPickerSupported(Context context) {
+        Intent intent = new Intent(MediaStore.ACTION_PICK_IMAGES);
```
shubham1g5, line 95: How is this passing on versions older than api 33 as this field is [not present there](https://developer.android.com/reference/android/provider/MediaStore#ACTION_PICK_IMAGES) ?

avazirna: This puzzled me a bit, but there is a Google project called [Mainline](https://source.android.com/docs/core/ota/modular-system) that allows new functionality and libraries to be delivered outside of the normal Android release cycle, making it possible to provide newer functionality to older versions of Android, here's the list of [modules](https://source.android.com/docs/core/ota/modular-system#available-modules). As per the `ACTION_PICK_IMAGES` itself, considering that it is a constant, during build the reference is replaced by its value, so it should be safe.

shubham1g5 (dismissed):

avazirna:

shubham1g5 (dismissed):

avazirna:

avazirna:

avazirna:

avazirna:

avazirna:

shubham1g5 (approved):

PR #2968, opened 2025-03-03
Disable menu items while downloading the applications

Jignesh-dimagi:
## Product Description
Menu items of connect page are disabled while downloading the application

## Technical Summary
https://dimagi.atlassian.net/browse/CCCT-721

## Feature Flag
For better user experience and fixing the bugs by pressing back/menu items while downloading the app

### Safety story
This is small change and test on staging variant

### QA Plan
General QA

### DIFF
--- a/app/src/org/commcare/activities/connect/ConnectActivity.java
+++ b/app/src/org/commcare/activities/connect/ConnectActivity.java
@@ -41,7 +41,7 @@ import androidx.navigation.NavOptions;
 import androidx.navigation.fragment.NavHostFragment;
 
 public class ConnectActivity extends CommCareActivity<ResourceEngineListener> {
-    private boolean backButtonEnabled = true;
+    private boolean backButtonAndActionBarEnabled = true;
     private boolean waitDialogEnabled = true;
     String redirectionAction = "";
     String opportunityId = "";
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/activities/connect/ConnectActivity.java
```diff
@@ -160,6 +160,13 @@ public boolean onCreateOptionsMenu(Menu menu) {
         return super.onCreateOptionsMenu(menu);
     }
 
+    @Override
+    public boolean onPrepareOptionsMenu(Menu menu) {
+        menu.findItem(R.id.action_sync).setEnabled(backButtonAndActionBarEnabled);
```
OrangeAndGreen, line 165: Let's hide these instead of disabling them, that way the user won't be confused seeing them.

Jignesh-dimagi: @OrangeAndGreen I would prefer not to jerk the UI but to show disabled state of icon. Thought?

shubham1g5: >  jerk the UI
@Jignesh-dimagi do you mean there is some sort of actual ui jerk here caused by this ?

It's geerally preferrable to not show the icon at all in order to keep the UI clean (we have seen that our users gets fustrated by disabled icons thinking something is wrong with their phone instead)

Jignesh-dimagi: @shubham1g5 Yes in some sense, icons hide/show will catch an eye of users that something happened. But as you and @OrangeAndGreen are of the view to hide, I will implement it.

OrangeAndGreen: Another option Clayton floated here is not to show the titlebar at all, making the download page is more of a full-screen view. Thoughts @shubham1g5 @Jignesh-dimagi?

shubham1g5: A full screen view here does seem like a good option to me here.

Jignesh-dimagi: Agree with @shubham1g5. It should not be full screen.

OrangeAndGreen: That sounds good to me

OrangeAndGreen (approved):

PR #2971, opened 2025-03-03
Automatic encryption key retrieval for messaging channels

OrangeAndGreen:
https://dimagi.atlassian.net/browse/CCCT-789

## Product Description
Not user visible

## Technical Summary
Attempting to retrieve encryption key for consented channel when a message is received for that channel and the key hasn't been retrieved yet. To support this, I added code to synchronously retrieve the encryption key, and extracted the common code for handling the received key to a helper method.

## Feature Flag
ConnectID Messaging

### Safety story
Low risk, leverages existing code to attempt to retrieve the encryption key on the fly. If it fails, the fallback behavior is to ignore the message.

### Automated test coverage

### QA Plan
Create an auto-consented channel for a Connect opportunity that a mobile worker is part of.
Before ever accessing the channel in the messaging channel list, send a message to the device on that channel.
The message should be received, decrypted, and notified to the user so they can view it.

## Labels and Review


app/src/org/commcare/connect/network/ApiConnectId.java
```diff
@@ -520,6 +537,21 @@ public static void retrieveChannelEncryptionKey(Context context, String channelI
...
+                    StreamsUtil.inputStreamToByteArray(stream));
+
+            if(responseAsString.length() > 0) {
+                JSONObject json = new JSONObject(responseAsString);
+                channel.setKey(json.getString("key"));
+                ConnectDatabaseHelper.storeMessagingChannel(context, channel);
+            }
+        } catch(Exception e) {
+            Logger.exception("Messaging channel encryption key", e);
+        }
```
shubham1g5, line 552: We must move away from practice of catching generalised exceptions in Connect unless there is strong reason to do so.

shubham1g5 (changes requested):

app/src/org/commcare/connect/network/ApiConnectId.java
OrangeAndGreen: [d5e66df](https://github.com/dimagi/commcare-android/pull/2971/commits/d5e66dfe080f094162ab0e619d09ff2e611f624e)

app/src/org/commcare/connect/network/ApiConnectId.java
```diff
@@ -520,6 +536,21 @@ public static void retrieveChannelEncryptionKey(Context context, String channelI
...
+        try {
+            String responseAsString = new String(
+                    StreamsUtil.inputStreamToByteArray(stream));
+
+            if(responseAsString.length() > 0) {
+                JSONObject json = new JSONObject(responseAsString);
+                channel.setKey(json.getString("key"));
+                ConnectDatabaseHelper.storeMessagingChannel(context, channel);
+            }
+        } catch(IOException | JSONException e) {
```
shubham1g5, line 549: Given a `JSONException` here signifies a coding error here, I think this needs to hard crash the application.

shubham1g5 (changes requested):

app/src/org/commcare/connect/network/ApiConnectId.java
OrangeAndGreen: Ah, okay, that makes sense to me too. The compiler won't let me leave those exceptions uncaught so I converted to a RuntimeException ([b2b465f](https://github.com/dimagi/commcare-android/pull/2971/commits/b2b465fae46ca54ef13bf3c6ba50b7b616efc953)), is that the best way to handle this?

shubham1g5: yes, that's what we want but only for `JSONException` and not `IOException` as `IOException` can happen for external reasons (network/disk I/O etc) out of our control.

OrangeAndGreen: Looking into inputStreamFromByteArray a bit more, I see Connect code using it quite a bit so whatever we decide to do will need to be applied in other spots as well. But that's probably work for a new ticket...

When I look at existing usage in the legacy code I see a few ways this has been handled:
* Print stack trace ([here](https://github.com/dimagi/commcare-android/blob/934c13f938648455ce611d93266907199197a3f5/app/src/org/commcare/logging/DataChangeLogger.kt#L170), [here](https://github.com/dimagi/commcare-android/blob/934c13f938648455ce611d93266907199197a3f5/app/src/org/commcare/android/nsd/MicroNode.java#L52), [here](https://github.com/dimagi/commcare-core/blob/876d21563aa4334db6e0f757ca4c087c4e4576dc/src/translate/java/org/javarosa/engine/FunctionExtensions.java#L165))
* RuntimeException ([here](https://github.com/dimagi/commcare-android/blob/934c13f938648455ce611d93266907199197a3f5/app/src/org/commcare/models/database/HybridFileBackedSqlStorage.java#L233))
* UnresolvedResourceException ([here](https://github.com/dimagi/commcare-core/blob/876d21563aa4334db6e0f757ca4c087c4e4576dc/src/main/java/org/commcare/resources/model/installers/OfflineUserRestoreInstaller.java#L58))
* Log ([here](https://github.com/dimagi/commcare-android/blob/934c13f938648455ce611d93266907199197a3f5/app/src/org/commcare/network/GetAndParseActor.java#L77))

OrangeAndGreen: Here's the change for silently logging the IOException if that's acceptable. [0caff08](https://github.com/dimagi/commcare-android/pull/2971/commits/0caff08a62a98ed112d3e0dc40e1b0134b497a73)

shubham1g5: thanks, yeah that's acceptable for `IOException` as they don't necessarily signify a code error on our end but think it's better for them to bubble up to UI level method calls and show a related error on UI corresponding to the error message.

shubham1g5 (approved):

PR #2973, opened 2025-03-05
Auto login app after its downloaded 

Jignesh-dimagi:
## Technical Summary
https://dimagi.atlassian.net/browse/CCCT-800
https://dimagi.atlassian.net/browse/CCCT-730

Previously, there was a bug. After downloading the app from connect screen, app was showing empty list of opportunities. Even though code was pointing out that it should redirect to auto login but it was not happening due to bug. This is rectified in this PR.

## Feature Flag
Not a feature, its a bug

### Safety story
I have tested in staging variant

### QA Plan
Now, after downloading the app, user should be automatically logged in.

### DIFF
--- a/app/src/org/commcare/activities/connect/ConnectActivity.java
+++ b/app/src/org/commcare/activities/connect/ConnectActivity.java
@@ -6,11 +6,9 @@ import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
 import android.graphics.PorterDuff;
-import android.graphics.drawable.ColorDrawable;
 import android.os.Bundle;
 import android.view.Menu;
 import android.view.MenuItem;
-import android.view.Window;
 
 import com.google.common.base.Strings;
 
@@ -19,7 +17,6 @@ import org.commcare.activities.CommCareVerificationActivity;
 import org.commcare.android.database.connect.models.ConnectJobRecord;
 import org.commcare.connect.ConnectDatabaseHelper;
 import org.commcare.connect.ConnectManager;
-import org.commcare.connect.MessageManager;
 import org.commcare.dalvik.R;
 import org.commcare.fragments.connect.ConnectDownloadingFragment;
 import org.commcare.google.services.analytics.FirebaseAnalyticsUtil;
@@ -139,6 +136,12 @@ public class ConnectActivity extends CommCareActivity<ResourceEngineListener> {
     public void setTitle(CharSequence title) {
         super.setTitle(title);
         getSupportActionBar().setTitle(title);
+    }
+
+    @Override
+    protected void onActivityResult(int requestCode, int resultCode, @androidx.annotation.Nullable Intent data) {
+        super.onActivityResult(requestCode, resultCode, data);
+        ConnectManager.handleFinishedActivity(this, requestCode, resultCode, data);
     }
 
     @Override
--- a/app/src/org/commcare/activities/connect/ConnectActivity.java
+++ b/app/src/org/commcare/activities/connect/ConnectActivity.java
@@ -142,12 +142,6 @@ public class ConnectActivity extends CommCareActivity<ResourceEngineListener> {
     }
 
     @Override
-    protected void onActivityResult(int requestCode, int resultCode, @androidx.annotation.Nullable Intent data) {
-        super.onActivityResult(requestCode, resultCode, data);
-        ConnectManager.handleFinishedActivity(this, requestCode, resultCode, data);
-    }
-
-    @Override
     public boolean onCreateOptionsMenu(Menu menu) {
         getMenuInflater().inflate(R.menu.menu_connect, menu);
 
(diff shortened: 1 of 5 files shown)


app/src/org/commcare/activities/connect/ConnectActivity.java
```diff
@@ -141,12 +141,6 @@ public void setTitle(CharSequence title) {
         getSupportActionBar().setTitle(title);
     }
 
-    @Override
-    protected void onActivityResult(int requestCode, int resultCode, @androidx.annotation.Nullable Intent data) {
-        super.onActivityResult(requestCode, resultCode, data);
-        ConnectManager.handleFinishedActivity(this, requestCode, resultCode, data);
```
OrangeAndGreen, line 147: I think this call needs to remain in order to support workflows that start with unlocking ConnectID (it handles routing the result of the biometric unlock activity back to the calling code in ConnectManager).

If it's causing an issue in the case of finishing an app installation and going to auto-login, maybe we could have ConnectManager.handleFinishedActivity return a boolean indicating whether it handled the activity result?

OrangeAndGreen (changes requested):

commcare-support-library/src/main/java/org/commcare/commcaresupportlibrary/CommCareLauncher.java
```diff
@@ -13,6 +13,7 @@ public class CommCareLauncher {
     public static void launchCommCareForAppId(Context context, String appId) {
         Intent intent = new Intent(CC_LAUNCH_ACTION);
         intent.putExtra(SESSION_ENDPOINT_APP_ID, appId);
+        intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP  | Intent.FLAG_ACTIVITY_NEW_TASK);
```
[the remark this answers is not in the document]
shubham1g5: thanks for checking, confirming this seems apt to me here.

app/src/org/commcare/activities/connect/ConnectActivity.java
Jignesh-dimagi: @OrangeAndGreen I am not sure if I fully understand this unlocking ConnectID flow because so far when I browse the app and seen the code, I have not came across such thing. Are you referring to `ConnectUnlockFragment`?

OrangeAndGreen: Understandable, this is a less-travelled path into the app. The ConnectManager.handleFinishedActivity function in question is directly related to ConnectManager.unlockConnect. If an activity calls unlockConnect, it also needs to handle the routing via handleFinishedActivity. This code is a bit outdated and is probably going to be overhauled sometime soon anyway (to be more generic and not distinguish between biometric and PIN/pattern), but for now (if I remember correctly) the routing function is necessary to support the code flow in the PIN/pattern case.

Under normal usage, the user enters Connect functionality (and is prompted to unlockConnect) from either the LoginActivity or SetupActivity. In those cases the unlock prompt is shown over the existing screen (i.e. a fingerprint bottom sheet appears over the login page).

But when the user enters Connect functionality by clicking on a push notification (like when they get invited to a new opp), we still want them to unlockConnect first. So in that case the ConnectActivity is launched as the Intent for the notification, and we pass in a redirection action so it can navigate to a particular spot (i.e. intro page for the new opp). When that happens, ConnectActivity first loads the ConnectUnlockFragment to show a blank screen that the unlock dialog can appear over. The once the user succeeds unlock, the navigation takes them to the specified redirect.

Jignesh-dimagi: @OrangeAndGreen ConnectUnlockFragment  is not connected through any calls. Is this functionality still used?

OrangeAndGreen: Yeah, but it happens via the [navigation graph](https://github.com/dimagi/commcare-android/blob/60a2199d967141ea84f554572de106e4ca825fd8/app/res/navigation/nav_graph_connect.xml#L10). [Here's the code](https://github.com/dimagi/commcare-android/blob/60a2199d967141ea84f554572de106e4ca825fd8/app/src/org/commcare/activities/connect/ConnectActivity.java#L112) that navigates to that fragment.

OrangeAndGreen: I dug in a bit more and saw that the problematic code was erroneously launching the jobs list after unlock. That line is meant to handle launching the jobs list after the user registers/recovers ConnectID from the SetupActivity, so they are automatically taken to that page once configured. I made a change to only make that call when the requestCode indicates that the jobs list is where we want to navigate to. That fixes the bug while still preserving the functionality from SetupActivity after registering/recovering. [1b6aa37](https://github.com/dimagi/commcare-android/pull/2973/commits/1b6aa3786453f3e8e7cadbc8de45dcbb5ab4fe05)

OrangeAndGreen (approved):

PR #2976, opened 2025-03-05
Preview of message in messages channel changed

Jignesh-dimagi:
## Product Description
For Outgoing message: read/unread icon + message body
For incoming message: message body

When last message was outgoing:
image

When last message was incoming:
image

## Technical Summary
https://dimagi.atlassian.net/browse/CCCT-711

## Feature Flag
This is kind of feature where display of message preview changed

### Safety story
Limited effect on safety, UI change

### QA Plan
Need general review for message preview

### DIFF
--- a/app/src/org/commcare/utils/StringUtils.java
+++ b/app/src/org/commcare/utils/StringUtils.java
@@ -1,9 +1,7 @@
 package org.commcare.utils;
 
 import android.content.Context;
-import android.content.res.Resources;
 import android.text.Spannable;
-import android.util.DisplayMetrics;
 
 import org.javarosa.core.services.locale.Localization;
 import org.javarosa.core.util.NoLocalizedTextException;
--- a/app/src/org/commcare/utils/StringUtils.java
+++ b/app/src/org/commcare/utils/StringUtils.java
@@ -1,7 +1,9 @@
 package org.commcare.utils;
 
 import android.content.Context;
+import android.content.res.Resources;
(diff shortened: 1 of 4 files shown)


OrangeAndGreen (approved):

app/src/org/commcare/utils/StringUtils.java
```diff
@@ -49,4 +51,10 @@ public static Spannable getStringSpannableRobust(Context c, int resId, String ar
         }
         return MarkupUtil.styleSpannable(c, ret);
     }
+
+    public static float convertDpToPixel(float dp){
```
shubham1g5, line 55: Maybe better to create a separate Util class , something like - 

````
public class DimensionUtils {
   // I am a little worried about performance here if we end up calling this too many times in a loop or something and therefore might be best retained as a static variable
    private static final float density = Resources.getSystem().getDisplayMetrics(). densityDpi; 

    public static int convertDpToPixel(float dp) { }
}
````

shubham1g5 (changes requested):

app/src/org/commcare/utils/StringUtils.java
Jignesh-dimagi: Yeah I was thinking of creating new but I can see lots of Util files so thought to keep in StringUtils but anyways created new DimensionUtils file

OrangeAndGreen (approved):

PR #2978, opened 2025-03-07
Set CC version in form metadata in SemVer format

avazirna:
## Product Description
This is a small change to ensure that the CommCare versions that appear in the form submission `appVersion` metadata follow the SemVer format. They are currently being computed differently but from the same source, resulting in values such as:
```
<n1:appVersion xmlns:n1="http://commcarehq.org/xforms">CommCare Android, version "2.56"(1). App v16. CommCare Version 2.56.0. Build 1, built on: 2025-03-05</n1:appVersion>
```
These differences are causing discrepancies in the Enterprise Dashboard, more info [here](https://dimagi.atlassian.net/browse/SAAS-16383).

## Technical Summary
It seems that in the past, one used to be sourced from the `PackageManager` and the other from the `BuildConfig.VERSION_NAME`, the latter is the current source for both. The change was done [here](https://github.com/dimagi/commcare-android/commit/9ce804477231844805bf5bd545aabf223ad3cb5b).

### Safety story
Successfully tested locally. However, it was flagged during review that there might be apps with form control logic based on the `appVersion` metadata field. To assess the potential impact of the change, steps were taken to identify such apps and document any required updates, but no apps were found. So the decision is to proceed and monitor potential issues that may arise from this once released, the AEs team will be responsible for that.

### Release Note
Minor change to the format of the `appVersion` metadata in forms - this change can cause unexpected behaviors in apps that use this field to control form logic.

## Labels and Review

### DIFF
--- a/app/assets/locales/android_translatable_strings.txt
+++ b/app/assets/locales/android_translatable_strings.txt
@@ -204,8 +204,8 @@ post.server.error=Server-side error (code ${0}) received from network request.
 post.gone.error=Case is not present on server.
 post.conflict.error=You have already claimed this case.
 
-version.id.long=CommCare Android, version "${0}"(${1}). App v${5}. CommCare Version ${2}. Build ${3}, built on: ${4}
-version.id.short=CCDroid:"${0}"(${1}). v${5} CC${2}b[${3}] on ${4}
+version.id.long=CommCare Android, version "${0}"(${1}). App v${4}. CommCare Version ${0}. Build ${2}, built on: ${3}
+version.id.short=CCDroid:"${0}"(${1}). v${4} CC${0}b[${2}] on ${3}
 commcare.version=CommCare Android, version "${0}"(${1}), built on: ${2}
 
 auth.request.not.using.https=${0} is an invalid URL. CommCare can't connect with authentication to an insecure (HTTP) server.
--- a/app/src/org/commcare/AppUtils.java
+++ b/app/src/org/commcare/AppUtils.java
@@ -173,7 +173,7 @@ public class AppUtils {
         String buildNumber = BuildConfig.BUILD_NUMBER;
 
         return Localization.get(application.getString(R.string.app_version_string), new String[]{
-                ccv, String.valueOf(BuildConfig.VERSION_CODE), ccv, buildNumber, buildDate, profileVersion});
+                ccv, String.valueOf(BuildConfig.VERSION_CODE), buildNumber, buildDate, profileVersion});
     }
 
     public static String getCurrentAppId() {
--- a/app/src/org/commcare/AppUtils.java
+++ b/app/src/org/commcare/AppUtils.java
@@ -173,9 +173,7 @@ public class AppUtils {
         String buildNumber = BuildConfig.BUILD_NUMBER;
 
         return Localization.get(application.getString(R.string.app_version_string), new String[]{
-                BuildConfig.VERSION_NAME,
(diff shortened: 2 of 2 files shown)


app/src/org/commcare/AppUtils.java
```diff
@@ -173,9 +173,7 @@ public static String getCurrentVersionString() {
         String buildNumber = BuildConfig.BUILD_NUMBER;
 
         return Localization.get(application.getString(R.string.app_version_string), new String[]{
-                BuildConfig.VERSION_NAME,
-                String.valueOf(BuildConfig.VERSION_CODE),
-                ccv, buildNumber, buildDate, profileVersion});
+                ccv, String.valueOf(BuildConfig.VERSION_CODE), ccv, buildNumber, buildDate, profileVersion});
```
shubham1g5, line 176: can we not show `ccv` twice ?

shubham1g5 (changes requested):

app/src/org/commcare/AppUtils.java
avazirna: Yes, [7f0f857](https://github.com/dimagi/commcare-android/pull/2978/commits/7f0f857cc7f56a4710438ed6d76ceacd21f9b342)

avazirna: I wonder if we still need this notation, having it twice when the source is the same seems redundant

app/assets/locales/android_translatable_strings.txt
```diff
@@ -204,8 +204,8 @@ post.server.error=Server-side error (code ${0}) received from network request.
 post.gone.error=Case is not present on server.
 post.conflict.error=You have already claimed this case.
 
-version.id.long=CommCare Android, version "${0}"(${1}). App v${5}. CommCare Version ${2}. Build ${3}, built on: ${4}
-version.id.short=CCDroid:"${0}"(${1}). v${5} CC${2}b[${3}] on ${4}
+version.id.long=CommCare Android, version "${0}"(${1}). App v${4}. CommCare Version ${0}. Build ${2}, built on: ${3}
+version.id.short=CCDroid:"${0}"(${1}). v${4} CC${0}b[${2}] on ${3}
```
shubham1g5, line 208: I am just realizing that this format also has reprucssions for `instance('commcaresession')/session/context/appversion` which has the potential to break apps if they were relying on a specific hardcoded format to extract versioning. Therefore I don't think we can change the format here.

shubham1g5 (changes requested):

app/assets/locales/android_translatable_strings.txt
shubham1g5: I don't know if there is one but we can't be sure that there is no app that does that.

avazirna: @shubham1g5 I looked at some form submissions from CC versions with the patch version greater than 0, and I noticed that in those cases, the versioning aligns with SemVer. If this had cause apps to break, we would have seen it by know. Here's an example:
`<n2:appVersion xmlns:n2="http://commcarehq.org/xforms">CommCare Android, version &quot;2.53.1&quot;(1). App v76. CommCare Version 2.53.1. Build 1, built on: 2023-04-11</n2:appVersion>`

avazirna: @shubham1g5 I added a note about this to the `Safety Story` section above.

shubham1g5 (approved):
Can we add a release note here to make sure this gets flagged in release comms as well.

PR #2982, opened 2025-03-12
MInor Tweaks for Cache and Lazy Load

shubham1g5:
## Technical Summary
 A few minor tweaks and pending feedback for cache and lazy load feature. Review commit by commit. 

## Labels and Review

cross-request: https://github.com/dimagi/commcare-core/pull/1463

### DIFF
--- a/app/src/org/commcare/tasks/EntityCacheInvalidationWorker.kt
+++ b/app/src/org/commcare/tasks/EntityCacheInvalidationWorker.kt
@@ -13,7 +13,9 @@ class EntityCacheInvalidationWorker(context: Context, workerParams: WorkerParame
     override suspend fun doWork(): Result {
         try {
             val entityStorageCache: CommCareEntityStorageCache = CommCareEntityStorageCache("case")
-            entityStorageCache.processShallowRecords()
+            if (!entityStorageCache.isEmpty) {
+                entityStorageCache.processShallowRecords()
+            }
         } catch (e: Exception) {
             Logger.exception("Error encountered while invalidating entity cache", e)
             return Result.failure()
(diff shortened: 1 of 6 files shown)


shubham1g5:
https://github.com/coderabbitai full review

shubham1g5:
@coderabbitai full review

shubham1g5:

shubham1g5:
@coderabbitai full review

avazirna (approved):
Looks good to me!

app/src/org/commcare/tasks/EntityCacheInvalidationWorker.kt
```diff
@@ -13,10 +14,17 @@ class EntityCacheInvalidationWorker(context: Context, workerParams: WorkerParame
...
+            if (!entityStorageCache.isEmpty) {
+                entityStorageCache.processShallowRecords()
+            }
         } catch (e: Exception) {
             Logger.exception("Error encountered while invalidating entity cache", e)
             return Result.failure()
+        } finally {
+            // we want to schedule the Prime worker irrespective of result of this task
+            if (CommCareApplication.isSessionActive()) {
+                PrimeEntityCacheHelper.schedulePrimeEntityCacheWorker()
```
ctsims, line 26: Honestly, I don't like that these two methods refer to different static singletons with different lifecycles. I'm worried this code has the risk of further spreading out access to static / singleton lifecycles while also being fairly likely to be executed in contexts where those lifecycles are changing (right after login, or when a user has completed a form and is logging out for the day). 

My guess here is that it's better to make "the other half" of any calls like this that depend on multiple singleton contexts get pushed into one of them, so that at least there's no ambiguity at runtime about which memory space or sandbox the second call is going to be running against.

shubham1g5: Not completely sure if I got you right but did you mean [like this](https://github.com/dimagi/commcare-android/pull/2982/commits/091cd01f1ba904cbf3f5115120061193330666e7) ?

PR #2984, opened 2025-03-12
Additional logging for failed HQ token auth in ConnectID-managed apps.

OrangeAndGreen:
https://dimagi.atlassian.net/browse/CCCT-668

## Technical Summary
Some additional logging to help capture the failed token auths related to ConnectID-managed apps.
We're already logging the case where we're unable to get an HQ token just before making an HQ call.
But currently we still make the call anyway, which is guaranteed to fail.
Even once that's fixed, we still may get surprise failures due to tokens we thought were good actually not being good.
So this PR adds logging after HQ calls fail (key record or data pull requests) due to bad authorization when they should be managed by ConnectID via tokens.

## Feature Flag
ConnectID

### Safety story
Simple addition that silently logs a condition, and checks happen in a try-catch in case something goes wrong.

### Automated test coverage
No automated tests for ConnectID yet

### QA Plan
No extra QA plan necessary, this is for dev tracking.

## Labels and Review

### DIFF
--- a/app/src/org/commcare/activities/LoginActivity.java
+++ b/app/src/org/commcare/activities/LoginActivity.java
@@ -878,8 +878,7 @@ public class LoginActivity extends CommCareActivity<LoginActivity>
                 raiseLoginMessage(StockMessages.Empty_Url, true);
                 break;
             case AUTH_FAILED:
-                String seatedAppId = CommCareApplication.instance().getCurrentApp().getUniqueId();
-                if(ConnectManager.checkForFailedConnectIdAuth(seatedAppId, uiController.getEnteredUsername())) {
+                if(ConnectManager.checkForFailedConnectIdAuth(uiController.getEnteredUsername())) {
                     Logger.exception("Token auth error for connect managed app",
                             new Throwable("Token Auth failed during login for a ConnectID managed app"));
                 }
--- a/app/src/org/commcare/activities/LoginActivity.java
+++ b/app/src/org/commcare/activities/LoginActivity.java
@@ -878,6 +878,11 @@ public class LoginActivity extends CommCareActivity<LoginActivity>
                 raiseLoginMessage(StockMessages.Empty_Url, true);
                 break;
             case AUTH_FAILED:
+                String seatedAppId = CommCareApplication.instance().getCurrentApp().getUniqueId();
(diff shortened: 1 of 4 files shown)


app/src/org/commcare/activities/SyncCapableCommCareActivity.java
```diff
@@ -105,7 +106,12 @@ public void handlePullTaskResult(ResultAndError<DataPullTask.PullTaskResult> res
             case AUTH_FAILED:
                 String seatedAppId = CommCareApplication.instance().getCurrentApp().getUniqueId();
                 String username = CommCareApplication.instance().getRecordForCurrentUser().getUsername();
-                ConnectManager.forgetAppCredentials(seatedAppId, username);
```
shubham1g5, line 108: this is intentional ?

app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -878,6 +878,11 @@ public void handlePullTaskResult(ResultAndError<DataPullTask.PullTaskResult> res
                 raiseLoginMessage(StockMessages.Empty_Url, true);
                 break;
             case AUTH_FAILED:
+                String seatedAppId = CommCareApplication.instance().getCurrentApp().getUniqueId();
+                if(ConnectManager.checkForFailedConnectIdAuth(seatedAppId, uiController.getEnteredUsername())) {
```
shubham1g5, line 882: we can move the `seatedAppId` eval in `checkForFailedConnectIdAuth` itself

app/src/org/commcare/activities/SyncCapableCommCareActivity.java
OrangeAndGreen: Yes, getting rid of that on purpose. It was added with ConnectID-managed traditional apps in mind. But now that we know the token auth could fail for other reasons, we don't want to take the unnecessary step of forcing the app credentials to be setup again. The improved messaging is a better reaction to this error.

app/src/org/commcare/activities/LoginActivity.java
OrangeAndGreen: No problem. [3078ade](https://github.com/dimagi/commcare-android/pull/2984/commits/3078ade57d85e2de7e7886dbbc5d5d3e751d909e)

shubham1g5 (approved):

PR #2988, opened 2025-03-13
Android 15 Edge to Edge support

avazirna:
## Product Description
This PR makes the necessary changes to ensure compatibility with Android 15 Edge-to-edge mode. This consisted mainly of assessing the window insets and then adjust the padding of the system content view. The impact to the UI is:
<table>
<tr>
<th>Android 14 and lower</th>
<th>Android 15+</th>
</tr>
<tr>
<td>

</td>
<td>
image-20250902-121244
</td>
</table>

### Safety story
These changes will have to go through regression testing.

## Labels and Review


app/src/org/commcare/utils/AndroidUtil.java
```diff
@@ -69,4 +76,34 @@ public static HashMap<String, String> bundleAsMap(Bundle bundle) {
...
+
+    /**
+     * Attaches a WindowInsetsListener to the root view of the activity for devices running Android 15+. This
+     * listener applies padding to the view based on the system window insets.
+     *
+     * @param activity   The activity to which the listener is attached.
+     * @param rootViewId The ID of the root view in the activity's layout.
+     */
+    @RequiresApi(api = Build.VERSION_CODES.VANILLA_ICE_CREAM)
+    public static void attachWindowInsetsListener(AppCompatActivity activity, int rootViewId) {
```
shubham1g5, line 90: couple questions:

1. Why not set the padding directly in xml layout instead ?
2. Is there no backward compatibility for edge to edge for pre Android 15 devices ?

avazirna: > Is there no backward compatibility for edge to edge for pre Android 15 devices ?

We can enable it for older versions too, but I believe it's better to focus on Android 15 and make sure that everything is fine.

shubham1g5: > Like fixed padding?
Nvm, I didn't realise the values are getting eval'ed at runtime. 

> We can enable it for older versions too, but I believe it's better to focus on Android 15 and make sure that everything is fine.

I think it's critical for us to be consistent across devices and we would like to make sure that if we support edge to edge, we support it on all devices and not just Android 15.

shubham1g5: Sorry,  Rethinking on backward compatibility, I do see value in only supporting it to Android 15 only as we do expect OS versions to impact UX in general.

app/src/org/commcare/activities/CommCareActivity.java
```diff
@@ -189,6 +196,18 @@ private void persistManagedUiState() {
...
 
+    @Override
+    public void onPostCreate(@Nullable Bundle savedInstanceState) {
+        super.onPostCreate(savedInstanceState);
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.VANILLA_ICE_CREAM) {
+            AndroidUtil.attachWindowInsetsListener(this, getRootViewId());
+        }
+    }
+
+    public int getRootViewId() {
```
shubham1g5, line 207: 1. this should be an abstract method as activities must implement this. 
2. On Connect, a single activity host multiple fragments with their own views. Is there no generalised way to get a root view of the current screen instead of assigning IDs to the xml layouts ?

shubham1g5 (changes requested):

app/src/org/commcare/utils/AndroidUtil.java
avazirna: yes

app/src/org/commcare/activities/NoCommCareActivity.java
```diff
@@ -0,0 +1,23 @@
...
+import android.os.Build;
+import android.os.Bundle;
+
+import org.commcare.utils.AndroidUtil;
+
+import androidx.annotation.Nullable;
+import androidx.appcompat.app.AppCompatActivity;
+import androidx.core.view.WindowInsetsControllerCompat;
+
+public class NoCommCareActivity extends AppCompatActivity {
```
shubham1g5, line 12: Think we should have `CommCareActivity` also extend from this so that we only put this logic in one place, also might make sense to name it `CommonBaseActivity` if we do that.

shubham1g5 (approved):
Left a suggestion but nothing blocking.

PR #2989, opened 2025-03-13
Delivery Progress UI fix

OrangeAndGreen:
https://dimagi.atlassian.net/browse/CCCT-838

## Product Description
Delivery progress tabs no longer squished vertically when changing between tabs.

## Technical Summary
This PR fixes a bug in which part of the UI would sometimes be sized too short (matching a different UI), such that elements were squished. Also while working in this code, I cleaned things up slightly so as not to hold static references to the two fragments (which equated to holding static references to Context objects).

Showing the error:
image

## Feature Flag
Connect

### Safety story
Tested the code out to verify things size correctly now and nothing got broken in my cleanup.
Otherwise this is mostly just about updating the height of a UI element, pretty simple change.

### Automated test coverage
No automated tests for Connect yet.

### QA Plan
See steps to reproduce in the Jira ticket. The UI should no longer get squished when following the steps.

## Labels and Review

### DIFF
--- a/app/src/org/commcare/fragments/connect/ConnectDeliveryProgressFragment.java
+++ b/app/src/org/commcare/fragments/connect/ConnectDeliveryProgressFragment.java
@@ -3,6 +3,7 @@ package org.commcare.fragments.connect;
 import android.content.Context;
 import android.os.Bundle;
 import android.view.LayoutInflater;
+import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.ImageView;
@@ -66,7 +67,7 @@ public class ConnectDeliveryProgressFragment extends Fragment {
     }
 
     @Override
-    public View onCreateView(LayoutInflater inflater, ViewGroup container,
+    public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
                              Bundle savedInstanceState) {
         ConnectJobRecord job = ConnectManager.getActiveJob();
         getActivity().setTitle(R.string.connect_progress_delivery);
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/fragments/connect/ConnectDeliveryProgressFragment.java
```diff
@@ -134,6 +135,12 @@ public void onPageSelected(int position) {
                     TabLayout.Tab tab = tabLayout.getTabAt(position);
                     tabLayout.selectTab(tab);
 
+                    View view = viewStateAdapter.createFragment(position).getView();
+                    if(view != null) {
+                        pager.getLayoutParams().height = view.getMeasuredHeight();
+                        pager.requestLayout();
+                    }
```
shubham1g5, line 142: can you talk more about the issue here ? As though why does the view get cut off and can't it be corrected from view xml instead by adding margin to the xml view ?

OrangeAndGreen: I'm having a hard time figuring out what the underlying issue is here, it seems to me that the FragmentStateAdapter should be handling the size of the fragments. After spending some time and not making much progress I decided to let Github Copilot try to help me! This was the solution it suggested, and when I tried it out it indeed worked.

But I'm open to looking for a better fix, I'd love to understand what's getting confused such that the fragment ends up too short despite having clear layout params. Another odd thing to note is that changing between the tabs seems to work fine at first. The issue only occurs when the user is on the second tab and presses a button (which causes some UI to change), then navigates back to the first tab.

I don't think I'll be able to get to that before the cohort starts on Monday though, so is the current change acceptable for now to fix the user-facing bug? I can create another ticket to revisit this next week and if I'm still struggling then I'll recruit Jignesh for some fresh eyes.

shubham1g5 (approved):

app/src/org/commcare/fragments/connect/ConnectDeliveryProgressFragment.java
shubham1g5: It would also be good to add a code comment there to denote why we are doing it as it's not entirely obvious by looking at the code.

OrangeAndGreen:
Thanks Shubham, I created [a ticket](https://dimagi.atlassian.net/browse/CCCT-863) to revisit this next week. Merging the existing fix for now.

PR #2990, opened 2025-03-14
Job status fix for learning

OrangeAndGreen:
No ticket, fixing a quick bug for Jon.

## Product Description
Fixes a special case where the user has transitioned to learning but hasn't completed a learn module yet. This ensures that the status stays in "learning" and doesn't revert to "available".

## Technical Summary
The underlying issue is that the server doesn't send a status along with each job when we retrieve the jobs. So when parsing and storing the job data, we use a [couple checks](https://github.com/dimagi/commcare-android/blob/489b81a863d341b500b1170d6ce690a0368a9a4f/app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java#L270) to determine the current status. Unfortunately, the check to determine whether we're in learning state is whether the user has >0 learn completion (i.e. they've completed at least one learn module).

When the user starts learning, the mobile code changes the status for the job locally to learning, but when the job gets downloaded again we calculate the status fresh from the payload, resulting in "available" status again. This is handled by taking whichever status is later in the workflow when comparing newly received jobs to what's already in storage. So if local knows we're in learning, we'll override available to learning before storing the job in the DB.

## Feature Flag
Connect

### Safety story
Small change to enforce expected behavior

### Automated test coverage
None

### QA Plan
Not required

## Labels and Review

### DIFF
--- a/app/src/org/commcare/connect/ConnectDatabaseHelper.java
+++ b/app/src/org/commcare/connect/ConnectDatabaseHelper.java
@@ -341,6 +341,12 @@ public class ConnectDatabaseHelper {
             for (ConnectJobRecord incoming : jobs) {
                 if (existing.getJobId() == incoming.getJobId()) {
                     incoming.setID(existing.getID());
+
+                    //Could be the case that app has transitioned to learn and server doesn't know yet
+                    if(existing.getStatus() > incoming.getStatus()) {
+                        incoming.setStatus(existing.getStatus());
+                    }
+
                     stillExists = true;
                     break;
                 }


app/src/org/commcare/connect/ConnectDatabaseHelper.java
```diff
@@ -341,6 +341,12 @@ public static int storeJobs(Context context, List<ConnectJobRecord> jobs, boolea
             for (ConnectJobRecord incoming : jobs) {
                 if (existing.getJobId() == incoming.getJobId()) {
                     incoming.setID(existing.getID());
+
+                    //Could be the case that app has transitioned to learn and server doesn't know yet
+                    if(existing.getStatus() > incoming.getStatus()) {
+                        incoming.setStatus(existing.getStatus());
+                    }
```
shubham1g5, line 348: Can status ever change on server side ? I wonder if it can cause server changes to get overwritten in some cases ?

OrangeAndGreen: Not really since the server doesn't actually send a status for each job, just data that mobile uses to determine the status. Status changes do occur this way though, but only in the "forward" direction (i.e. increasing integer values), i.e. Available >> Learning >> Delivering.

That's why the check here only overrides the value when existing status is greater than incoming status.

shubham1g5: can we always do `incoming.setStatus(existing.getStatus());` in that case to simplify the reasoning here to respect mobile statuses over server always.

shubham1g5 (approved):
Approving with an optional suggestion

app/src/org/commcare/connect/ConnectDatabaseHelper.java
OrangeAndGreen: Hmm, yeah, that could probably work. I was thinking there could be cases where a server-side change is the impetus for moving to a new state, but currently that's not ever the case.

But for example, if we were to add a special state for "assessment", i.e. the user completed all learn modules but hasn't passed the assessment yet. In a case like that, the user would submit an assessment form, and then sometime later mobile would learn (via the jobs API call) that the user had passed the assessment and progressed to delivery state.

So it seemed to me that allowing the server to move us forward in state could logically make sense, if not now then in the future. But nothing should ever move the user backward in the workflow, at least not as we've designed it so far.

Your thoughts?

shubham1g5: Sounds good to me to keep it as it is based on your reasoning here.

PR #2993, opened 2025-03-19
Additional logging to troubleshoot the repeated login issue

shubham1g5:
## Product Description

Logging to troubleshoot https://dimagi.atlassian.net/browse/QA-7555

This is a temporary addition and will need to be reverted before release.


avazirna (approved):
Probably stating the obvious here, but can we make sure to have a ticket to analyse these logs and proceed with the removal once they are no longer needed.

shubham1g5:
> Probably stating the obvious here, but can we make sure to have a ticket to analyse these logs and proceed with the removal once they are no longer needed.

I will track this in https://dimagi.atlassian.net/browse/QA-7555 itself

PR #2998, opened 2025-03-24
Fixes for Entity Cache and Lazy loading work

shubham1g5:
## Technical Summary
 This PR makes following changes for cache and index work - 
 - Nullability corrections https://github.com/dimagi/commcare-android/pull/2998/commits/cdd44e260f78f3bbec749843effcd9ff7a5079ed
- Instead of showing a progress indictor for 3 phases that each goes from 0 to 100 , shows a weighted progress by giving all possible phases equal ewightage. https://github.com/dimagi/commcare-android/pull/2998/commits/980fcad76c288866dbee3756b74ceeedd26e9f81
- Fixes a bug where a normal case list loading gets blocked on background cache work priming work due to [user db being locked ](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/models/database/user/models/CommCareEntityStorageCache.java#L91)by the background work. This is fixed by explicitly calling cancel on the background work before loading the list https://github.com/dimagi/commcare-android/pull/2998/commits/d9bce47ebce1027fede81e57694834b88d137665
- Fixes an issue where a case list cache prime completion event gets registered from the previously emitted value of State Flow. This caused an entity list to show up ahead of time without really completing calculaitons for all cases. https://github.com/dimagi/commcare-android/pull/2998/commits/9061a3fe5b81d2fbc2d5a6962a9af02799673f21

## Feature Flag

CACHE AND LAZY LOADING

### Safety story

All the impacted areas should be behind the feature flag. 

## Labels and Review

### DIFF
--- a/app/src/org/commcare/tasks/EntityLoaderHelper.kt
+++ b/app/src/org/commcare/tasks/EntityLoaderHelper.kt
@@ -2,6 +2,7 @@ package org.commcare.tasks
 
 import android.util.Pair
 import io.reactivex.functions.Cancellable
+import org.commcare.CommCareApplication
 import org.commcare.activities.EntitySelectActivity
 import org.commcare.cases.entity.AsyncNodeEntityFactory
 import org.commcare.cases.entity.Entity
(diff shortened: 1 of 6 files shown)


app/src/org/commcare/tasks/EntityLoaderHelper.kt
```diff
@@ -54,12 +55,23 @@ class EntityLoaderHelper(
...
     ): Pair<List<Entity<TreeReference>>, List<TreeReference>>? {
-        val references = factory.expandReferenceList(nodeset)
-        val entities = loadEntitiesWithReferences(references, progressListener)
-        entities?.let {
-            factory.prepareEntities(entities)
-            factory.printAndClearTraces("build")
-            return Pair<List<Entity<TreeReference>>, List<TreeReference>>(entities, references)
+        if (factory !is AsyncNodeEntityFactory) {
+            // if we are into synchronous mode, cancel background cache work for now to not lock the user db
+            CommCareApplication.instance().currentApp.primeEntityCacheHelper.cancelWork()
```
avazirna, line 60: Two questions:
1.  Given that `cancelUniqueWork()` is async, are we certain that while the cancellation is not complete, there won't be any db locking issues?
2.  It seems that calling `cancelLoading()` in sync mode leads to its default implementation which throws an `RuntimeException`. Is that expected?

avazirna (commented):
@shubham1g5 left a couple of questions.

app/src/org/commcare/tasks/EntityLoaderHelper.kt
shubham1g5: 1. Db might block for a little bit (probably a fraction of second) while the cancellation completes but should be able to get the lock as soon as the canncelled thread interrupts and release the lock. 
2. Yeah that's expected, as we only call this method for it's Async implementations.

shubham1g5:

shubham1g5:

shubham1g5:

avazirna (approved):

PR #3002, opened 2025-03-27
Update Firebase dependencies

avazirna:
## Product Description
This PR is to update Firebase Analytics and Crashlytics to 20.1.2 and 18.3.7, respectively. The main motivation is to make ANRs available in Crashlytics, this will give better insight into the affected users and domains. Crashlytics started supporting ANR collection in 18.4.2. The Firebase Analytics bump is more opportunistic, the current version is from 2020 and from the release notes, this bump seems safe to make.

### Safety story
Caused a crash and was able to confirm its report in `Crashlytics` as well as the Analytics events.

## Labels and Review


shubham1g5 (approved):
We should verify that we get crashes and analytics events with this update.

PR #3005, opened 2025-03-31
CCCT-890

pm-dimagi:
## Product Description
-bug fixes
expired opportunity bug fix
non connect user workflow 
and continue button in recovery flow
https://dimagi.atlassian.net/browse/CCCT-890
https://dimagi.atlassian.net/browse/CCCT-900

### Safety story
Describe:
- tested the changes in dev testing 

In particular consider how existing data may be impacted by this change.

## Labels and Review

### DIFF
--- a/app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
+++ b/app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
@@ -204,7 +204,7 @@ public class ConnectJobRecord extends Persisted implements Serializable {
 
         job.claimed = json.has(META_CLAIM) && !json.isNull(META_CLAIM);
 
-        job.isActive = !json.has(META_IS_ACTIVE) || json.getBoolean(META_IS_ACTIVE);
+        job.isActive = json.optBoolean(META_IS_ACTIVE, true);
 
         job.isUserSuspended = json.optBoolean(META_USER_SUSPENDED, false);
 
--- a/app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
+++ b/app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
@@ -204,7 +204,7 @@ public class ConnectJobRecord extends Persisted implements Serializable {
 
         job.claimed = json.has(META_CLAIM) && !json.isNull(META_CLAIM);
 
-        job.isActive = !json.optBoolean(META_IS_ACTIVE, true);
+        job.isActive = !json.has(META_IS_ACTIVE) || json.getBoolean(META_IS_ACTIVE);
 
         job.isUserSuspended = json.optBoolean(META_USER_SUSPENDED, false);
 
--- a/app/src/org/commcare/connect/database/ConnectJobUtils.java
+++ b/app/src/org/commcare/connect/database/ConnectJobUtils.java
@@ -412,9 +412,13 @@ public class ConnectJobUtils {
     }
 
     public static ConnectAppRecord getAppRecord(Context context, String appId) {
-        Vector<ConnectAppRecord> records = ConnectDatabaseHelper.getConnectStorage(context, ConnectAppRecord.class).getRecordsForValues(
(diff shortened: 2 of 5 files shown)


app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
```diff
@@ -204,7 +204,7 @@ public static ConnectJobRecord fromJson(JSONObject json) throws JSONException, P
 
         job.claimed = json.has(META_CLAIM) && !json.isNull(META_CLAIM);
 
-        job.isActive = !json.optBoolean(META_IS_ACTIVE, true);
+        job.isActive = !json.has(META_IS_ACTIVE) || json.getBoolean(META_IS_ACTIVE);
```
shubham1g5, line 207: can simply be `json.optBoolean(META_IS_ACTIVE, true)` i.e. just remove the `!`

app/src/org/commcare/connect/database/ConnectJobUtils.java
```diff
@@ -412,9 +412,13 @@ public static void storeLearningRecords(Context context, List<ConnectJobLearning
     }
 
     public static ConnectAppRecord getAppRecord(Context context, String appId) {
-        Vector<ConnectAppRecord> records = ConnectDatabaseHelper.getConnectStorage(context, ConnectAppRecord.class).getRecordsForValues(
-                new String[]{ConnectAppRecord.META_APP_ID},
-                new Object[]{appId});
-        return records.isEmpty() ? null : records.firstElement();
+        if (ConnectManager.isConnectIdConfigured()) {
```
shubham1g5, line 415: what if connect id is configured but not for the currently selected app on login screen ?

app/src/org/commcare/fragments/connectId/ConnectIDSignupFragment.java
```diff
@@ -265,21 +265,14 @@ public void checkPhoneNumber() {
                 String finalPhone = phone;
                 switch (callingClass) {
                     case ConnectConstants.CONNECT_REGISTRATION_PRIMARY_PHONE,
-                         ConnectConstants.CONNECT_REGISTRATION_CHANGE_PRIMARY_PHONE,
-                         ConnectConstants.CONNECT_RECOVERY_PRIMARY_PHONE -> {
-                        if (existingPrimary != null && existingPrimary.equals(phone)) {
```
shubham1g5, line 270: what are we trying to solve here ? And why is primary check not required any more ?

shubham1g5 (changes requested):
Can we add Jira links to issues this PR is trying to solve for in PR description.

app/src/org/commcare/fragments/connectId/ConnectIDSignupFragment.java
pm-dimagi: Because it was a issue we are getting existing no when we are changing the method from signup to recover we pass the no entered there  so this condition willl never allow user to check phone no

shubham1g5:
@pm-dimagi How about we PR this to `master` instead and then merge `master` inside `connect_qa`. I am not sure how else we will get these changes in `master` otherwise. Thoughts ?

pm-dimagi:

app/src/org/commcare/connect/database/ConnectJobUtils.java
OrangeAndGreen: The new check is simply whether ConnectID is configured on the device, i.e. the user has completed registration/recovery. This is essentially just a check that the Connect database exists and the user isn't currently mid-way through registration.

The query that follows is specific to the currently selected app (appId is passed in as a param), and will return an empty list (converted to null) if the currently selected app is not linked to ConnectID.

shubham1g5: right, I see that now. thanks!

shubham1g5 (approved):

PR #3017, opened 2025-04-09
Correctly cancels the entity loading task

shubham1g5:
## Product Description

Correctly cancels the entity loading task on finishing activity as we are not going to reattach to the task again

## Technical Summary

Solves an issue where the entity loader task keeps running even after the entity select activity has been finished due to user getting out of that screen. 

This was causing multiple entity loading tasks to get intitiated and run simultaneously when you go back and come in again on the entity list before the entity list has been loaded completely. 

The solution here is 2 parts - 

1. Change the [mayInterruptIfRunning](https://developer.android.com/reference/android/os/AsyncTask#cancel(boolean)) to `true` to make sure that the `onCancelled` gets called on time for AsyncTask. 

2. Correctly cancels the work by intercepting the `stopLoading` flag which is set to `true` on cancellation. 

### Safety story

It's a bit confusing to me that we were setting the [interrruption flag to `false` on cancellation](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/activities/EntitySelectActivity.java#L547) which is hard to know if it was an oversight or is done due to a particular reason I am unaware of. Although think if we are trying to cancel it's better to interrupt otherwise there is no guarantee on when it well get cancelled. 

### Automated test coverage
None

### QA Plan
Not planning

## Labels and Review


avazirna:
I believe [this](https://github.com/dimagi/commcare-android/blob/fcf9a749ee93d4f8995261336e4f8c2a0672d8f4/app/src/org/commcare/tasks/EntityLoaderHelper.kt#L101) block is no longer needed, given when `stopLoading` now, we return the `entities` list

shubham1g5:
you are absolutely right, thanks for catching - https://github.com/dimagi/commcare-android/pull/3017/commits/da553e552874fb2f8b63ed5f567131579b18abc0

PR #3018, opened 2025-04-10
Cache Prime Fix: Reuse entity factory for cache prime

shubham1g5:
## Technical Summary

https://dimagi.atlassian.net/browse/SC-4402

This is necessary to be able to populate cache correctly in primeCache as otherwise we don't have values like [mTemplateIsCachable and mCacheHost correctly initialized](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/commcare/cases/entity/AsyncNodeEntityFactory.java#L94) which results into the  `primeCache` method to exit early and not populate cache into memory resulting into a performance hit. 

## Feature Flag
Cache and Lazy Load 

### Safety story

The change allows us to reuse the factory class which should be harmless for normal use cases that doesn't involve entity caching. The pass on integration test should confirm this further. 

### QA Plan

None
## Labels and Review

### DIFF
--- a/app/src/org/commcare/entity/AndroidAsyncNodeEntityFactory.kt
+++ b/app/src/org/commcare/entity/AndroidAsyncNodeEntityFactory.kt
@@ -57,7 +57,8 @@ class AndroidAsyncNodeEntityFactory(
                         getCurrentCommandId(),
                         detail,
                         entityDatum,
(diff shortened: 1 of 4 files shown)


avazirna (approved):
Makes sense to me.

PR #3026, opened 2025-04-11
Show fingerprint failure error string

OrangeAndGreen:
Quick change to include the error string returned by the OS when the user fails fingerprint authentication, both in the reported non-fatal exception and the toast message shown to the user.

### DIFF
--- a/app/src/org/commcare/fragments/connectId/ConnectIdBiometricConfigFragment.java
+++ b/app/src/org/commcare/fragments/connectId/ConnectIdBiometricConfigFragment.java
@@ -100,10 +100,12 @@ public class ConnectIdBiometricConfigFragment extends Fragment {
                     }
                 }
 
-                Logger.exception("Exhausted biometrics", new Exception(
-                        "Fingerprint error and PIN isn't configured/allowed: " + allowPassword));
+                String whyNoPin = allowPassword ? "configured" : "allowed";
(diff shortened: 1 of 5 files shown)


app/src/org/commcare/fragments/connectId/ConnectIdBiometricConfigFragment.java
```diff
@@ -100,10 +100,12 @@ public void onAuthenticationError(int errorCode,
...
 
-                Logger.exception("Exhausted biometrics", new Exception(
-                        "Fingerprint error and PIN isn't configured/allowed: " + allowPassword));
+                String whyNoPin = allowPassword ? "configured" : "allowed";
+                Logger.exception("Exhausted biometrics", new Exception(String.format(Locale.getDefault(),
+                        "Fingerprint error and PIN isn't %s: %s (%d)", whyNoPin, errString, errorCode)));
 
-                Toast.makeText(context, R.string.connect_verify_configuration_failed, Toast.LENGTH_SHORT).show();
+                String message = context.getString(R.string.connect_verify_configuration_failed, errString);
+                Toast.makeText(context, message, Toast.LENGTH_SHORT).show();
```
shubham1g5, line 108: we should use `LENGTH_LONG` to give user enough time to read as it might be a long string with the `errString`

shubham1g5 (approved):

PR #3034, opened 2025-04-16
Phase 2 Connect Merge Fixes

shubham1g5:
## Technical Summary

Some final set of fixes/refactor before we merge the phase 2 connect code into `master`.  A bunch of commits are just to bring missing code from latest `connect_qa` branch to here. 

Best reviewed commit by commit. 

### Safety story

I have tried to make some calls here to what makes best sense to me and might introduce a regression with a special case handling on Login Screen that I am unaware of. I am mostly relying on review to spot any such edge case handling. I would like reviewer to pay special attention to login inputs visibility and connect app state management. 

### QA Plan
Will be QA'ed with phase 2 QA 

## Labels and Review

### DIFF
--- a/app/src/org/commcare/activities/HomeButtons.java
+++ b/app/src/org/commcare/activities/HomeButtons.java
@@ -28,7 +28,7 @@ import java.util.Vector;
 public class HomeButtons {
 
     private final static String[] buttonNames =
-            new String[]{"start", "training", "saved", "incomplete", "sync", "report", "logout"};
+            new String[]{"start", "training", "saved", "incomplete","connect", "sync", "report", "logout"};
 
     /**
      * Note: The order in which home cards are returned by this method should be consistent with
--- a/app/src/org/commcare/activities/LoginActivity.java
+++ b/app/src/org/commcare/activities/LoginActivity.java
@@ -418,7 +418,7 @@ public class LoginActivity extends CommCareActivity<LoginActivity>
 
             return true;
         } catch (Exception e) {
-            e.printStackTrace();
+            Logger.exception("Error while local login", e);
             return false;
         }
     }
@@ -897,7 +897,7 @@ public class LoginActivity extends CommCareActivity<LoginActivity>
         return selectedNewApp;
     }
 
-    public void setConnectAppState() {
+    protected void setConnectAppState() {
         String seatedAppId = CommCareApplication.instance().getCurrentApp().getUniqueId();
         ConnectIDManager.ConnectAppMangement appState = connectIDManager.evalAppState(this,
                 seatedAppId, uiController.getEnteredUsername());
--- a/app/src/org/commcare/activities/LoginActivity.java
+++ b/app/src/org/commcare/activities/LoginActivity.java
@@ -1,5 +1,7 @@
 package org.commcare.activities;
 
+import static org.commcare.activities.DispatchActivity.REDIRECT_TO_CONNECT_OPPORTUNITY_INFO;
+
 import android.annotation.TargetApi;
 import android.content.Context;
 import android.content.Intent;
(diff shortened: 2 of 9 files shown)


app/src/org/commcare/activities/StandardHomeActivity.java
```diff
@@ -58,7 +58,6 @@ public void onCreateSessionSafe(Bundle savedInstanceState) {
     @Override
     protected void onResume() {
         super.onResume();
-        updateSecondaryPhoneConfirmationTile();
```
pm-dimagi, line 61: The main reason to call updateSecondaryPhoneConfirmationTile() here was to remove the tile as soon as user update the secondaryphone number when he returns back from the connectidactivity

shubham1g5: That's good to know, I added it back in [the commit](https://github.com/dimagi/commcare-android/pull/3034/commits/3a23cc28cc4179c5abc45ec9bdbb1b4006a8e848) along with moving some changes to their right place.

pm-dimagi (approved):

app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -899,29 +912,41 @@ protected boolean seatAppIfNeeded(String appId) {
...
+            setConnectAppState(Unmanaged);
         }
+    }
+
+    protected boolean loginManagedByConnectId() {
+        return connectAppState == ConnectId || connectAppState == Connect;
+    }
 
-        uiController.setConnectIdLoginState(appState);
+    public void setConnectAppState(ConnectIDManager.ConnectAppMangement connectAppState) {
```
OrangeAndGreen, line 945: Optional: Seems like this code could be moved into one place at the end of evaluateConnectAppState? It's called at the end of both cases in that function (and the function never returns early).
The only other usage is when the user chooses to deconfigure ConnectID, in which case it would be just as simple to call evaluateConnectAppState after deleting the user and let the function set the state to Unmanaged.

OrangeAndGreen: Ah, missed the additional call from the UI controller, may not be worth changing anything here.

app/src/org/commcare/activities/StandardHomeActivityUIController.java
```diff
@@ -106,9 +111,31 @@ protected void updateSyncButtonMessage(String message) {
         adapter.notifyItemChanged(adapter.getSyncButtonPosition());
     }
 
-    public void updateConnectTile(boolean show) {
-        ConnectIDManager.getInstance().updateSecondaryPhoneConfirmationTile(activity, connectTile, show, v -> {
-            activity.performSecondaryPhoneVerification();
-        });
+    public void updateSecondaryPhoneConfirmationTile() {
```
OrangeAndGreen, line 114: This tile is also shown on the Connect home page (to be merged later), which is why the common code was extracted to ConnectIdManager.

app/src/org/commcare/connect/ConnectIDManager.java
```diff
@@ -701,6 +655,16 @@ public void setFailureAttempt(int failureAttempt) {
         failedPinAttempts = failureAttempt;
     }
 
+    public boolean shouldShowJobStatus(Context context, String appId) {
```
OrangeAndGreen, line 658: Leave out until the Connect merge?

OrangeAndGreen (commented):
Left a few comments but nothing that necessarily prevents going forward

app/src/org/commcare/activities/HomeButtons.java
```diff
@@ -28,7 +28,7 @@
 public class HomeButtons {
 
     private final static String[] buttonNames =
-            new String[]{"start", "training", "saved", "incomplete", "sync", "report", "logout"};
+            new String[]{"start", "training", "saved", "incomplete","connect", "sync", "report", "logout"};
```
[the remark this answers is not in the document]
shubham1g5: It's a bit hard to strictly separte connect and connectID code and this was added in response to a crash happening due to other code added for this button but missing out on this one. A such this button will be hidden until app is connect so think it's harmless as such.

app/src/org/commcare/activities/StandardHomeActivityUIController.java
shubham1g5: I see, I think we will refactor it with the later merge in that case tobe abstracted out as a widget instead. I think it's important for `ConnectIdManager` to not have any UI code and it be dealing with string Connect ID business logic.

app/src/org/commcare/connect/ConnectIDManager.java
shubham1g5: think it's fine to have it unless you think it'll cause issues with the current merge.

app/src/org/commcare/activities/StandardHomeActivityUIController.java
OrangeAndGreen: That sounds good to me

OrangeAndGreen (approved):

PR #3040, opened 2025-04-18
Show flags for deliveries

OrangeAndGreen:
https://dimagi.atlassian.net/browse/CCCT-607

## Product Description
Showing flags associated with deliveries (such as too short, missing GPS, etc.)

## Technical Summary
New table to store flags associated with a Connect job delivery.
Database upgrade code.
API retrieval and data storage.
Showing the flags as a comma-separated list in the delivery list (in delivery progress page).

## Feature Flag
Connect

### Safety story
Tested locally against prod server.

### Automated test coverage
None for Connect yet.

### QA Plan
In a Connect opportunity, complete one or more deliveries without satisfying the required conditions for the opportunity (i.e. minimum time, GPS requirement, etc.). After submitting the form, go to delivery progress for the opportunity and view the delivery list. The corresponding flags for the delivery should be shown in a comma-separated list.


app/src/org/commcare/android/database/connect/models/ConnectJobDeliveryFlagRecord.java
```diff
@@ -0,0 +1,68 @@
...
+                flag.deliveryId = deliveryId;
+                flag.code = key;
+                flag.description = json.getString(key);
+                flags.add(flag);
+            } catch (JSONException e) {
+                Logger.exception("Error parsing delivery flag", e);
+            }
+        }
+        return flags;
+    }
```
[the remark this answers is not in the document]
OrangeAndGreen: This should never happen

app/src/org/commcare/android/database/connect/models/ConnectJobDeliveryRecord.java
```diff
@@ -91,6 +95,8 @@ public static ConnectJobDeliveryRecord fromJson(JSONObject json, int jobId) thro
             delivery.entityName = json.getString(META_ENTITY_NAME);
             delivery.reason = json.getString(META_REASON);
 
+            delivery.flags = ConnectJobDeliveryFlagRecord.fromJson(json.getJSONObject(META_FLAGS), deliveryId);
```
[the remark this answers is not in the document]
OrangeAndGreen: Should never be null

shubham1g5 (commented):
We need to PR the model and DB changes from here directly to `master`

OrangeAndGreen:
Added these new changes and some other missing ones to the code in master, [PR here](https://github.com/dimagi/commcare-android/pull/3043)

app/src/org/commcare/connect/database/ConnectJobUtils.java
```diff
@@ -213,6 +215,40 @@ public static void storeDeliveries(Context context, List<ConnectJobDeliveryRecor
...
+        SqlStorage<ConnectJobDeliveryFlagRecord> storage = ConnectDatabaseHelper.getConnectStorage(context, ConnectJobDeliveryFlagRecord.class);
+
+        List<ConnectJobDeliveryFlagRecord> existingFlags = getDeliveryFlags(context, deliveryId, storage);
+
+        //Delete flags that are no longer present
+        Vector<Integer> recordIdsToDelete = new Vector<>();
+        for (ConnectJobDeliveryFlagRecord existing : existingFlags) {
+            boolean stillExists = false;
+            for (ConnectJobDeliveryFlagRecord incoming : flags) {
+                if (existing.getDeliveryId() == incoming.getDeliveryId() && Objects.equals(existing.getCode(), incoming.getCode())) {
```
Jignesh-dimagi, line 233: We can override `equals` method in `ConnectJobDeliveryFlagRecord ` and perform the check for deliveryId and code to check if both are objects are equal.

app/src/org/commcare/connect/database/ConnectJobUtils.java
```diff
@@ -213,6 +215,40 @@ public static void storeDeliveries(Context context, List<ConnectJobDeliveryRecor
...
+                    incoming.setID(existing.getID());
```
Jignesh-dimagi, line 234: May be I have less context here so asking, why we need to use ID of existing to new?

app/src/org/commcare/connect/database/ConnectJobUtils.java
```diff
@@ -213,6 +215,40 @@ public static void storeDeliveries(Context context, List<ConnectJobDeliveryRecor
...
+                //Remember the ID so we can delete them all at once after the loop
+                recordIdsToDelete.add(existing.getID());
+            }
+        }
+
+        storage.removeAll(recordIdsToDelete);
+
+        //Now insert/update flags
+        for (ConnectJobDeliveryFlagRecord incomingRecord : flags) {
+            storage.write(incomingRecord);
```
Jignesh-dimagi, line 251: In above code, we are removing flag records which doesn't exists anymore (by comparing with incoming flags) but we are again storing the incoming records which may be already there in existing?

app/src/org/commcare/connect/database/ConnectJobUtils.java
OrangeAndGreen: Good suggestion, but we changed how we're doing this storage so it's not necessary anymore (simply deleting all flags for a delivery and inserting them again).

app/src/org/commcare/android/database/connect/models/ConnectJobDeliveryRecord.java
```diff
@@ -138,7 +138,7 @@ public Date getLastUpdate() {
     }
 
     public String getReason() {
-        return reason;
+        return reason != null && reason.equals("null") ? null : reason;
```
shubham1g5, line 141: given it's a new change we are making, I would actually like us to block on merging this until server can make this change to remove `null` strings. Thoughts ?

app/src/org/commcare/fragments/connect/ConnectDeliveryListFragment.java
```diff
@@ -187,7 +188,21 @@ public void bind(Context context, ConnectJobDeliveryRecord delivery) {
                 nameText.setText(delivery.getEntityName());
                 dateText.setText(ConnectManager.paymentDateFormat(delivery.getDate()));
                 statusText.setText(delivery.getStatus());
-                reasonText.setText(delivery.getReason());
+
+                String descriptionText = delivery.getReason();
+                if(descriptionText == null) {
+                    if(delivery.getFlags() != null) {
```
shubham1g5, line 194: Seems like the logic at the moment here is if delivery reason is not provided use delivery flags. I would have thought it to be other way around i.e. if delivery flags are not provided,use delivery reason just because delivery flags seems to be more explicit. 

I think any behaviour is fine here but just wanted to make sure we have given some thought in the order of execution here.

OrangeAndGreen: I think we want the reason to be the preferred choice when available since it's entered manually by a human when rejecting the delivery, whereas the flags are auto-generated by the system at the time of submission.

shubham1g5: I see, that resonates with me in that case.

app/src/org/commcare/android/database/connect/models/ConnectJobDeliveryRecord.java
OrangeAndGreen: Looking a little closer at this, checking for "null" was a recent change (April 18) to address the behavior that users would see that when no reason was provided yet. It seems that's a reasonable fallback until the server team fixes the issue on their side, at which point the mobile code will work correctly. Either way, we know we eventually don't want this check, so I went ahead and removed it. [65b6a93](https://github.com/dimagi/commcare-android/pull/3040/commits/65b6a93eeb4900895a77982c37864630087c2cb7)

shubham1g5: Good call.

PR #3041, opened 2025-04-18
Fixed two issues with rekeying DB after receiving new passphrase.

OrangeAndGreen:
https://dimagi.atlassian.net/browse/QA-7668

## Product Description
No visible change

## Technical Summary
Bringing some changes over from the hotfix branch.
The changes were prompted by QA testing.

1. Not logging analytics event about rekeying the DB until the passphrase has been stored as local.
2. Using query instead of execSQL to execute the rekey

Not sure why we need to use query instead, but it works and execSQL now throws an exception.

## Feature Flag
ConnectID

### Safety story
Tested locally and via QA.

### Automated test coverage
None

### QA Plan
Existing QA plan discovered this issue.


app/src/org/commcare/models/database/connect/DatabaseConnectOpenHelper.java
```diff
@@ -78,7 +78,7 @@ public static void rekeyDB(SQLiteDatabase db, String newPassphrase) throws Base6
             byte[] newBytes = Base64.decode(newPassphrase);
             String newKeyEncoded = UserSandboxUtils.getSqlCipherEncodedKey(newBytes);
 
-            db.execSQL("PRAGMA rekey = '" + newKeyEncoded + "';");
+            db.query("PRAGMA rekey = '" + newKeyEncoded + "';");
```
shubham1g5, line 81: think `execSQL` is the correct call to make for `PRAGMA` statements. Can you describe the exception thrown by making that call ?

OrangeAndGreen: Yeah, this has me confused too, and it definitely worked the other way when I coded it the first time (I remember encountering this). But the error we're getting is "Queries cannot be performed using execSQL(), use query() instead". Crash logs from Nitin's testing [here](https://console.firebase.google.com/u/0/project/commcare-a57e4/crashlytics/app/android:org.commcare.dalvik/issues/b9b31217959ae9cb493502b0d9eaf841?time=last-seven-days&versions=2.57%20(472004);2.57%20(472002);2.57%20(472001)&sessionEventKey=68023D0502BD000126FA1967A5241E92_2073065453945335715).

shubham1g5: Are we able to repro this on every device ? I wonder if using `rawExecSQL` might yield better result.

PR #3045, opened 2025-04-22
Added safe navigation

Jignesh-dimagi:
## Product Description
This is fix for bug

## Technical Summary
https://dimagi.atlassian.net/browse/CCCT-961

## Feature Flag
Bug fix

### Safety story
Test on staging and prod app for said changes

### QA Plan
This PR needs testing the connect ID sign up and recover flow

## Labels and Review


app/src/org/commcare/utils/CommCareNavigationController.kt
```diff
@@ -0,0 +1,25 @@
...
+
+
+fun NavController.navigateSafely(direction: NavDirections?) {
+    val currentDestination = this.currentDestination
+    if (direction!=null && currentDestination != null) {
+        val navAction = currentDestination.getAction(direction.actionId)
+        if (navAction != null) {
+            val destinationId: Int = navAction.destinationId.orEmpty()
+            val currentNode: NavGraph? = if (currentDestination is NavGraph) currentDestination else currentDestination.parent
+            if (destinationId != 0 && currentNode != null && currentNode.findNode(destinationId) != null) {
```
OrangeAndGreen, line 16: Looking at all the conditions that need to be met on the way to finally calling navigate(), I'm wondering in which cases we might want to log a non-fatal exception to Firebase vs. silently do nothing. Is the else-case in each of the three ifs an error situation that we should track?

Jignesh-dimagi: @OrangeAndGreen We already have crashes in app but still not able to identify the cause as its not giving any useful information to trace the issue. Also, I have feeling that lets not add at other places before seeing the reactions from user so kept at sign up / recovery mode only.

OrangeAndGreen (approved):

PR #3046, opened 2025-04-22
Add support for Document Upload question

minhaminha:
## Product Description

Adds support for a new Document Upload question for mobile. 

Main [HQ PR](https://github.com/dimagi/commcare-hq/pull/36034)

## Technical Summary

[Jira Ticket](https://dimagi.atlassian.net/browse/USH-5366)

Modeled roughly after the VideoWidget, just without the capture and playback buttons and with the intention to upload application/* and text/* type files. 

### Safety story

Tested locally and on staging. The blast radius will be limited by projects that use this new document upload feature. 

### QA Plan

QA originally found this to be not working on mobile so they'll be doing regression testing with the APK built by Jenkins. ([QA ticket here](https://dimagi.atlassian.net/browse/QA-7651))

## Labels and Review


minhaminha:
Would this be considered a "major change" worth communicating?

app/src/org/commcare/views/widgets/DocumentWidget.java
```diff
@@ -0,0 +1,72 @@
...
+    public void unsetListeners() {
+        superUnsetListeners();
+        mChooseButton.setOnLongClickListener(null);
+    }
+
+    @Override
+    public void cancelLongPress() {
+        superCancelLongPress();
+        mChooseButton.cancelLongPress();
+    }
```
shubham1g5, line 71: @minhaminha think we should just do `super. unsetListeners` directly here and have the super class (MediaWidget) implement null checks on individual buttons to supprt these calls.

app/src/org/commcare/views/widgets/MediaWidget.java
```diff
@@ -147,7 +147,9 @@ private SecretKeySpec getSecretKey() {
     }
 
     protected void togglePlayButton(boolean enabled) {
-        mPlayButton.setEnabled(enabled);
+        if (mPlayButton != null) {
```
shubham1g5, line 150: curious if you are running into an error here ? My assumption is that this method only gets called for widgets with play button.

shubham1g5:
> Would this be considered a "major change" worth communicating?

Is this feature flagged on HQ side or publicly available ? Also on HQ, this widget should require users to target a minimum of CommCare 2.57 as that's when this PR will likely go out.

app/src/org/commcare/views/widgets/DocumentWidget.java
minhaminha: Is this not simpler? Otherwise, I'd have to do null checks on at least the `mCapture` and `mPlay` buttons for 3 different functions, which I think would clutter `MediaWidget`, when all it really needs to do is call the grandfather method in QuestionWidget. Would renaming the function to make this clearer help?

app/src/org/commcare/views/widgets/DocumentWidget.java
```diff
@@ -0,0 +1,72 @@
...
+        mChooseButton = new MaterialButton(getContext());
+        WidgetUtils.setupButton(mChooseButton,
+                StringUtils.getStringSpannableRobust(getContext(), R.string.choose_document),
+                !mPrompt.isReadOnly());
+
+        mChooseButton.setOnClickListener(v -> {
+            try {
+                ((AppCompatActivity)getContext())
+                        .startActivityForResult(WidgetUtils.createPickMediaIntent(getContext(), "application/*,text/*"),
+                                FormEntryConstants.AUDIO_VIDEO_FETCH);
```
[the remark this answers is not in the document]
minhaminha: I just renamed it to `AUDIO_VIDEO_DOCUMENT_FETCH`, since it doesn't make sense to have a new case in the switch statement that does the same thing

app/src/org/commcare/views/widgets/DocumentWidget.java
```diff
@@ -0,0 +1,72 @@
...
+                        StringUtils.getStringSpannableRobust(getContext(),
+                                R.string.activity_not_found, "choose document"),
+                        Toast.LENGTH_SHORT).show();
+            }
+        });
+
+        String acq = mPrompt.getAppearanceHint();
+        if (QuestionWidget.ACQUIREFIELD.equalsIgnoreCase(acq)) {
+            mChooseButton.setVisibility(View.GONE);
+        }
```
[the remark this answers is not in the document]
minhaminha: removed! yep that was copied over from MediaWidget, I just forgot to get rid of it.

app/src/org/commcare/views/widgets/DocumentWidget.java
shubham1g5: Hmm I think it's pretty bad practie in Java to bypass super method chain like this (quite unusual I would say) and if we are having to do that, I don't think `MediaWidget` in it's current form is the right choice for super class for `documentWidget`. `MediaWidget` is designed to support media playback with the ability to play, capture and add media from file provider. It seems like we only need one of these functionalities (add file from file provider here) , So think we need to either modify `MediaWidget` to be more flexible by allowing for these use cases by expecting buttons as optional OR have a different `FileChooserWidget` that abstracts the functionality of "adding a file" to a common place and can be further utilized as a super class by both `MediaWidget` and `DocumentWidget`.

minhaminha: That's fair! I went with the first option after realizing that it would actually be less lines of code overall. I was also a bit confused about which `setOnLongClickListener` QuestionWidget would call (I thought it would call MediaWidget) but it does in fact, use the DocumentWidget's, which is good to know.

shubham1g5 (approved):
thanks, looks great! Although bumping on my previous note - 

> Is this feature flagged on HQ side or publicly available ? Also on HQ, this widget should require users to target a minimum of CommCare 2.57 as that's when this PR will likely go out.

minhaminha:
Just noting that I'll address the above comment in the main HQ PR (linked in the description), but this is safe to merge now since it can't affect any change without the HQ PR.

Issue #2742, opened 2024-02-15
Improve image capture resizing


ctsims:
### Bug Description
The current naive strategy for downsampling captured images doesn't perform well when resizing images with light patterns, resulting in a stippled appearance.

### Current Behavior
When images with light patterns are resized using the current downsampling strategy, the resized images appear stippled and amplify underlying shadow overlays like the image attached.

![image](https://github.com/dimagi/commcare-android/assets/155066/a7266856-f9a5-4d4a-860e-db4048647783)

### Expected Behavior
Resized images should maintain the quality of the original, without amplifying light patterns or shadow overlays.

### Reproduction Steps
_**Note:**_
- _We currently don’t have a consistent method to reproduce this issue. To qualify for the bounty, please see the Acceptance Criteria._
- _These steps assume you have basic knowledge of navigating and using the CommCare platform. If you are unfamiliar with how to perform basic setup steps, a [getting started guide is available here](https://github.com/dimagi/commcare-hq/blob/master/local_dev_guide.rst)._

1. Create a CCHQ app that [captures images](https://dimagi.atlassian.net/wiki/spaces/commcarepublic/pages/2143958321/Media+Capture+Questions#Image-Capture).
2. Install the app on an Android device
3. When you get to a screen with an image capture question, capture the photo by clicking a photo possibly with Light Patterns
4. After form submission, navigate to [Submit History report on HQ and download the attachment](https://dimagi.atlassian.net/wiki/spaces/commcarepublic/pages/2143950715/Submit+History#Attachments)
    -Observe that the resized image of the above-given image exhibits a stippled appearance or amplified shadow overlays similar to the one attached above.

### Possible Solution
Experiment with different platform methods for resizing images that do not amplify underlying shadow overlays. Potential solutions may include:

1. Platform-Specific Methods: Utilize built-in platform-specific resizing methods that are known to preserve image quality better in these scenarios.
2. Advanced Downsampling Algorithms: Explore and implement more sophisticated downsampling algorithms that are specifically designed to handle images with light patterns.

### Acceptance Criteria
A successful pull request for this issue will be expected to contain the following:

- A sample image which was captured in real usage of CommCare replicating the compression issues identified in the ticket, and the conditions under which the issue was replicated
   - Keep in mind that it may not be possible to replicate this issue on all devices
- A description of the correction applied in code which prevents the issue 
- A sample of the same image after the fix is applied

### Additional Information/Context
No response

### Environment details (OS name and version, etc.)
No response

[...]

PR #3069, opened 2025-05-04
Improve image downscaling using progressive steps to reduce artifacts. 

rahulrajesh21:
/claim #2742
## Product Description

Improves image downscaling quality in CommCare by implementing a progressive step-wise scaling approach. This reduces visual artifacts like amplifying light patterns when resizing large images.

## Technical Summary

Ticket: #2742 
	•	The original implementation of direct downscaling introduced visual artifacts in images with fine patterns. This change introduces progressive halving to preserve image quality.
	•	DownScaling is done  in multiple halving steps followed by a final scale, instead of a single step.

## Labels and Review

### DIFF
--- a/app/src/org/commcare/utils/FileUtil.java
+++ b/app/src/org/commcare/utils/FileUtil.java
@@ -7,7 +7,9 @@ import android.content.Context;
 import android.content.Intent;
 import android.database.Cursor;
 import android.graphics.Bitmap;
+
 import androidx.exifinterface.media.ExifInterface;
+
 import android.media.MediaMetadataRetriever;
 import android.net.Uri;
 import android.os.Build;
@@ -153,7 +155,7 @@ public class FileUtil {
     /**
      * Copies a source file from a FileProvider Content provider into a file directory local
      * to this application.
-     *
+     * <p>
      * The app needs to already have permissions granted for the external file content.
      *
      * @param contentUri A valid uri to a contentprovider backed external file
@@ -318,7 +320,7 @@ public class FileUtil {
 
     /**
      * http://stackoverflow.com/questions/11281010/how-can-i-get-external-sd-card-path-for-android-4-0
-     *
+     * <p>
      * Used in SD Card functionality to get the location of the SD card for reads and writes
      * Returns a list of available mounts; for our purposes, we just use the first
      */
@@ -620,7 +622,7 @@ public class FileUtil {
 
     private static void copyExifData(ExifInterface sourceExif, ExifInterface destExif, Bitmap scaledBitmap) {
         if (sourceExif == null || destExif == null) {
-          return;
+            return;
         }
         for (String tag : EXIF_TAGS) {
             String value = sourceExif.getAttribute(tag);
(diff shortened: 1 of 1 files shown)


damagatchi:

shubham1g5:
@rahulrajesh21 Thanks for submitting this, can you [fill this up](https://docs.google.com/forms/d/e/1FAIpQLSeIrqqWnawDDPySu6am8I0TafelxMMVBXq89KnD9cPsom0WUQ/viewform) for us to take next steps here.

rahulrajesh21:
@shubham1g5 I have filled out the form.

shubham1g5:
@damagatchi ok to test

shubham1g5 (commented):
@rahulrajesh21 The code here looks good to me, though I am a little worried about performance implications of step down sizing, can you provide a rough timing of how long it takes to perform `stepDownscale` for the sample image you provided in comparison to direct one step scaling.

rahulrajesh21:
@shubham1g5  for sample image (4080 x 3060 pixels ) `direct scaling` is taking around 1 ms  and `stepDownscale` is taking 16 ms. I tested it with a 4k image  (3840 x 2160 pixels ) `direct scaling` took 1 ms and `stepDownscale` took 10 ms

app/src/org/commcare/utils/FileUtil.java
```diff
@@ -698,6 +700,52 @@ public static void writeBitmapToDiskAndCleanupHandles(Bitmap bitmap, ImageType t
...
+            if (currentBitmap != originalBitmap && !currentBitmap.isRecycled()) {
+                currentBitmap.recycle();
+            }
+
+            currentBitmap = tempBitmap;
+        }
+
+        // Final step to exactly match target dimensions
+        if (width != targetWidth || height != targetHeight) {
+            Bitmap finalBitmap = Bitmap.createScaledBitmap(currentBitmap, targetWidth, targetHeight, false);
```
shubham1g5, line 736: I would be curious on your reasoning of putting bilinear filtering as `false` here and `true` above while wondering if it should be `false` even above or `true` even here ?

shubham1g5 (commented):
@rahulrajesh21 Thanks for putting out those timings, I think that's something we should be able to live with. Left one more question for you where it would be nice to get your thoughts on.

app/src/org/commcare/utils/FileUtil.java
shubham1g5: Can you elaborate more on what do you mean by preventing visual artifacts and why does it not matter in the final step when it matters in the intermediary steps ?

rahulrajesh21: Visual artifacts like moiré patterns and stippling occur primarily when downsampling at large ratios. Bilinear filtering averages neighboring pixels, which helps smooth out these artifacts.
When reducing dimensions by large factors, the sampling grid can align improperly with high-frequency details in the original image. This causes interference patterns - similar to strange patterns in digital photos of striped fabrics. In intermediary steps, bilinear filtering helps by averaging pixels, softening these problematic interference patterns.

In the final step, these artifacts matter less because:

1. The scaling ratio is much smaller.
2. The major frequency mismatches that cause interference have already been addressed in previous steps.
3. When sampling distance is smaller, the likelihood of problematic frequency interactions is reduced.

In the final step, bilinear filtering could also be kept `true`, but may reduce image quality and impact performance. So it is kept `false` to preserve sharpness and detail in the final image while maintaining better performance.

shubham1g5: that's insightful, thanks for the detailed explanation here.

shubham1g5 (approved):

PR #3075, opened 2025-05-06
Brings back missing analytics events And Token Exceptions Handling

shubham1g5:
## Product Description
- Make some lint changes and logs cleanup
- Adds missing analytics code that was PR'ed before in https://github.com/dimagi/commcare-android/pull/2687/files

## Labels and Review

### DIFF
--- a/app/assets/locales/android_translatable_strings.txt
+++ b/app/assets/locales/android_translatable_strings.txt
@@ -135,6 +135,8 @@ sync.fail.recovery.failure=There was an error trying to recover from a bad local
 sync.fail.unsent=Having issues communicating with the server to send forms. Will try again later.
 sync.fail.individual=1 or more forms failed to send due to individual problems. See quarantined forms for details.
 sync.fail.invalid.case.graph=1 or more forms introduced an invalid case relationship.
+sync.fail.token.unavailable=Couldn't contact server. Please make sure an internet connection is available or try again later.
+sync.fail.token.denied=Please re-login into your ConnectID account, then try again.
 
 form.send.rate.limit.error.toast=Form submission rate limit reached. Your forms are being saved locally and will be submitted to the server later.
 form.send.rate.limit.error.title=You've reached your form submission rate limit
@@ -561,7 +563,7 @@ notification.token.unavailable.action=Check your device's time and date to make 
 
 notification.token.denied.title=Couldn't connect to the server.
 notification.token.denied.detail=CommCare can no longer communicate with the ConnectID server.
-notification.token.denied.action=Please recover your ConnectID account, then try again.
+notification.token.denied.action=Please re-login into your ConnectID account, then try again.
 
 notification.restore.storagefull.title=Couldn't save your application's resources
 notification.restore.storagefull.detail=CommCare was not able to save one or more of your application's resources on the device. This could be because the entire application was too large, or because an individual fixture exceeded 1MB.
--- a/app/src/org/commcare/activities/SyncCapableCommCareActivity.java
+++ b/app/src/org/commcare/activities/SyncCapableCommCareActivity.java
@@ -106,13 +106,6 @@ public abstract class SyncCapableCommCareActivity<T> extends SessionAwareCommCar
                 updateUiAfterDataPullOrSend(Localization.get("sync.fail.empty.url"), FAIL);
                 break;
             case AUTH_FAILED:
-                String username = CommCareApplication.instance().getRecordForCurrentUser().getUsername();
-
-                if(ConnectIDManager.getInstance().isSeatedAppLinkedToConnectId(username)) {
-                    Logger.exception("Token auth error for connect managed app",
-                            new Throwable("Token Auth failed during sync for a ConnectID managed app"));
-                }
-
                 updateUiAfterDataPullOrSend(Localization.get("sync.fail.auth.loggedin"), FAIL);
                 break;
             case BAD_DATA:
--- a/app/src/org/commcare/activities/SyncCapableCommCareActivity.java
+++ b/app/src/org/commcare/activities/SyncCapableCommCareActivity.java
@@ -14,8 +14,6 @@ import androidx.annotation.AnimRes;
 import androidx.annotation.LayoutRes;
 
 import org.commcare.CommCareApplication;
-import org.commcare.connect.ConnectIDManager;
-import org.commcare.connect.ConnectManager;
 import org.commcare.dalvik.R;
 import org.commcare.google.services.analytics.AnalyticsParamValue;
 import org.commcare.google.services.analytics.FirebaseAnalyticsUtil;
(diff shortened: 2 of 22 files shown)


app/src/org/commcare/activities/SyncCapableCommCareActivity.java
```diff
@@ -106,13 +106,6 @@ public void handlePullTaskResult(ResultAndError<DataPullTask.PullTaskResult> res
                 updateUiAfterDataPullOrSend(Localization.get("sync.fail.empty.url"), FAIL);
                 break;
             case AUTH_FAILED:
-                String username = CommCareApplication.instance().getRecordForCurrentUser().getUsername();
-
-                if(ConnectIDManager.getInstance().isSeatedAppLinkedToConnectId(username)) {
```
OrangeAndGreen, line 111: We should separate this failed token case from other failed auth cases in terms of messaging. That could be done via token unavailable and rejected cases in this structure, or via this if-case since the corresponding exceptions extend IOException and otherwise fall into this case.

app/src/org/commcare/activities/connect/ConnectActivity.java
```diff
@@ -74,7 +74,7 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
 
         updateBackButton();
 
-        destinationListener = FirebaseAnalyticsUtil.getDestinationChangeListener();
+        destinationListener = FirebaseAnalyticsUtil.getNavControllerPageChangeLoggingListener();
```
OrangeAndGreen, line 77: Can we put this in ConnectIDActivity also?

app/src/org/commcare/google/services/analytics/CCAnalyticsEvent.java
```diff
@@ -32,7 +32,7 @@ public class CCAnalyticsEvent {
     static final String IN_APP_UPDATE_EVENT = "in_app_update_event";
     static final String LOGIN_CLICK = "login_click";
     static final String CCC_SIGN_IN = "ccc_sign_in";
-    static final String CCC_SIGN_OUT = "ccc_sign_out";
+    static final String CCC_FORGET = "ccc_forget";
```
OrangeAndGreen, line 35: Wondering how necessary this is? It will make looking at analytics more difficult since we'll need to remember there were two historical names for the same event

app/src/org/commcare/connect/ConnectIDManager.java
```diff
@@ -630,18 +630,18 @@ public boolean isLoggedInWithConnectApp(Context context, String appId) {
         return isloggedIn() && isConnectApp(context, appId);
     }
 
-    public static AuthInfo.TokenAuth getHqTokenIfLinked(String username) throws TokenRequestDeniedException, TokenUnavailableException {
-        if (!manager.isloggedIn()) {
+    public AuthInfo.TokenAuth getHqTokenIfLinked(String username) throws TokenRequestDeniedException, TokenUnavailableException {
+        if (isloggedIn()) {
```
OrangeAndGreen, line 634: Lost the '!'

app/src/org/commcare/network/CommcareRequestGenerator.java
```diff
@@ -176,29 +178,25 @@ public static HashMap<String, String> getHeaders(String lastToken) {
...
+            AuthInfo.TokenAuth tokenAuth = ConnectIDManager.getInstance().getHqTokenIfLinked(username);
             if (tokenAuth != null) {
-                Logger.log(LogTypes.TYPE_MAINTENANCE, "Applying token auth");
                 return tokenAuth;
             } else {
                 if (ConnectIDManager.getInstance().isSeatedAppLinkedToConnectId(username)) {
-                    Logger.exception("Token auth error for connect managed app",
-                            new Throwable("No token Auth available for a connect managed app"));
+                    Logger.log(LogTypes.TYPE_MAINTENANCE,
+                            "Could not get token Auth for a connect managed app");
```
OrangeAndGreen, line 187: Looking again, this will definitely be an error condition that we expect to never happen once the May beta goes out. I mentioned earlier that we're still seeing these events currently, but I forgot then that the token exceptions haven't gone out yet.

In the old code, getting a null here would happen if a token API call failed (i.e. due to bad network), so this was normal usage. But in the new code that same scenario will cause a TokenUnavailableException to bubble up from getHqTokenIfLinked through this method, so we should never get here anymore.

I think we should leave this as a non-fatal in the May beta release, or even consider hard-crashing instead.

OrangeAndGreen (changes requested):

app/src/org/commcare/google/services/analytics/CCAnalyticsEvent.java
shubham1g5: the old event was not even getting logged in the recent apks so think it's fine to rename this.

OrangeAndGreen: Ah, well in that case definitely safe to rename.

app/src/org/commcare/connect/ConnectIDManager.java
shubham1g5: https://github.com/dimagi/commcare-android/pull/3075/commits/4dca13fd89b48edbfabc6cabe1c740527bd2b3d7

app/src/org/commcare/activities/connect/ConnectActivity.java
shubham1g5: https://github.com/dimagi/commcare-android/pull/3075/commits/7cfa93017170d84c5b633c14e85da3d79d8f13c0

app/src/org/commcare/network/CommcareRequestGenerator.java
shubham1g5: https://github.com/dimagi/commcare-android/pull/3075/commits/2461bf35831999be8c9f25b025e9660bc463a902

app/src/org/commcare/activities/SyncCapableCommCareActivity.java
shubham1g5: https://github.com/dimagi/commcare-android/pull/3075/commits/c5c10ab4a935d941b54aec483e3cbde9b8989037 and https://github.com/dimagi/commcare-android/pull/3075/commits/787f46c5468fb0f05a929d01ea03aab5c92d3092

app/assets/locales/android_translatable_strings.txt
```diff
@@ -561,7 +563,7 @@ notification.token.unavailable.action=Check your device's time and date to make
 
 notification.token.denied.title=Couldn't connect to the server.
 notification.token.denied.detail=CommCare can no longer communicate with the ConnectID server.
-notification.token.denied.action=Please recover your ConnectID account, then try again.
+notification.token.denied.action=Please re-login into your ConnectID account, then try again.
```
shubham1g5, line 566: @OrangeAndGreen Given we are moving away from recover workflow, I changed this to re-login but think there is a better word out there that I am missing here.

OrangeAndGreen: I agree in the change but not sure of a better term. This is short-lived anyway I think until we can provide a better user experience for situations like this

OrangeAndGreen (approved):

PR #3097, opened 2025-05-09
Play Integrity Helper

OrangeAndGreen:
Added helper utility to generate a random nonce and asynchronously retrieve the Google Integrity token.

## Technical Summary
New GooglePlayIntegrityUtil class
Function to generate the nonce
Function to retrieve the integrity token

This code is not used in this PR but will be used in upcoming work.

Example Usage:
final String nonce = NonceGenerator.INSTANCE.generateNonce();
GooglePlayIntegrityUtil.INSTANCE.getIntegrityTokenAsync(this, nonce, token -> {
    //Send the nonce and token to the server
    return null;
});

## Feature Flag
ConnectID

### Safety story
Tested code with a signed build and verified that token retrieval works

### Automated test coverage
None

### QA Plan
Nothing for QA to test yet, that will come once this code is put to use.


OrangeAndGreen:

app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -57,6 +57,7 @@
 import org.commcare.tasks.ResultAndError;
 import org.commcare.utils.ConsumerAppsUtil;
 import org.commcare.utils.CrashUtil;
+import org.commcare.utils.FirebaseIntegrityUtil;
```
shubham1g5, line 60: remove ?

app/src/org/commcare/utils/FirebaseIntegrityUtil.kt
```diff
@@ -0,0 +1,56 @@
...
+
+        if (timeoutSeconds <= 0) {
+            throw IllegalArgumentException("Timeout must be greater than 0")
+        }
+
+        Thread {
+            try {
+                val integrityManager: IntegrityManager = IntegrityManagerFactory.create(context)
+                val request = IntegrityTokenRequest.builder().setNonce(nonce).build()
+                val task = integrityManager.requestIntegrityToken(request)
```
Jignesh-dimagi, line 39: @OrangeAndGreen Can we implement something like

```
integrityManager.requestIntegrityToken(request)
    .addOnSuccessListener { token ->
        
    }
    .addOnFailureListener { exception ->
        
    }
```

app/src/org/commcare/utils/FirebaseIntegrityUtil.kt
```diff
@@ -0,0 +1,56 @@
...
+
+                val response = Tasks.await(task, timeoutSeconds, java.util.concurrent.TimeUnit.SECONDS)
+                callback(response.token())
+            } catch (e: Exception) {
+                e.printStackTrace()
```
Jignesh-dimagi, line 44: @OrangeAndGreen Please remove `e.printStackTrace()` as it can leak in production build too

app/src/org/commcare/utils/FirebaseIntegrityUtil.kt
```diff
@@ -0,0 +1,56 @@
...
+    ) {
+        if (nonce.isBlank()) {
+            throw IllegalArgumentException("Nonce cannot be empty")
+        }
+
+        if (timeoutSeconds <= 0) {
+            throw IllegalArgumentException("Timeout must be greater than 0")
+        }
+
+        Thread {
```
shubham1g5, line 35: we should not need to spin off a new thread given the method `requestIntegrityToken` itself is async and provides a callback based future for us to listen to.

app/src/org/commcare/utils/FirebaseIntegrityUtil.kt
```diff
@@ -0,0 +1,56 @@
...
+                callback(response.token())
+            } catch (e: Exception) {
+                e.printStackTrace()
+                Logger.exception("Error retrieving Firebase Integrity token", e)
+                callback(null)
+            }
+        }.start()
+    }
+
+    fun generateNonce(): String {
```
shubham1g5, line 51: nit: would be good to move this to `CryptUtil`.

shubham1g5 (changes requested):
Great to see some Kotlin out there :)

Jignesh-dimagi:
@OrangeAndGreen @shubham1g5 

We can try to implement some part of clean architecture here, not fully but partially due to current app structure. Decision of adding a integrity token in API call should be made by `data` module and not in fragment/activity. While finally calling the API through retrofit/ sync task, decision should be made at that time to add or not integrity token. Also `data` (retrofit / async task class) module is responsible for fetching the integrity token.  This will make our code more robust to keep adding/removing the integrity token in an API call with only single change in `data` module. We can discuss more in details while implementing it.

app/src/org/commcare/utils/FirebaseIntegrityUtil.kt
Jignesh-dimagi: @shubham1g5 agree. We should use as mentioned [here](https://github.com/dimagi/commcare-android/pull/3097#discussion_r2087015384). Also, Google/Firebase must have optimized timeout time and/or other parameter so let it only decide and call `addOnFailureListener ` . This will reduce our boiling code.

app/src/org/commcare/utils/FirebaseIntegrityUtil.kt
OrangeAndGreen: The Base64 class that we're using here is the Android class, and the version we have in commcare-core doesn't include the same method (with NO_WRAP flag). Thoughts on how best to address this? Is our implementation of Base64 in core meant to replicate the functionality provided by the Android class?

app/src/org/commcare/utils/FirebaseIntegrityUtil.kt
```diff
@@ -0,0 +1,56 @@
...
+        callback: (String?) -> Void
+    ) {
+        if (nonce.isBlank()) {
+            throw IllegalArgumentException("Nonce cannot be empty")
+        }
+
+        if (timeoutSeconds <= 0) {
+            throw IllegalArgumentException("Timeout must be greater than 0")
+        }
+
```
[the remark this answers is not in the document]
OrangeAndGreen: [f092a13](https://github.com/dimagi/commcare-android/pull/3097/commits/f092a13d1bced499257a55f4f7229e3e3eb8d2ad)

app/src/org/commcare/utils/FirebaseIntegrityUtil.kt
```diff
@@ -0,0 +1,56 @@
...
+import android.content.Context
+import android.util.Base64
+import com.google.android.gms.tasks.Tasks
+import com.google.android.play.core.integrity.IntegrityManager
+import com.google.android.play.core.integrity.IntegrityManagerFactory
+import com.google.android.play.core.integrity.IntegrityTokenRequest
+import org.javarosa.core.services.Logger
+import java.security.SecureRandom
+
+object FirebaseIntegrityUtil {
```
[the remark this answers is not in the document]
OrangeAndGreen: Ah, excellent point. There was some confusion early on between Google Play Integrity and Firebase App Check, I updated the name to clear up that confusion. [6562692](https://github.com/dimagi/commcare-android/pull/3097/commits/6562692f21a10673aa6f3a55f7a5beaf6ce180e9)

app/src/org/commcare/activities/LoginActivity.java
OrangeAndGreen: [6562692](https://github.com/dimagi/commcare-android/pull/3097/commits/6562692f21a10673aa6f3a55f7a5beaf6ce180e9)

app/src/org/commcare/utils/FirebaseIntegrityUtil.kt
OrangeAndGreen: Thanks for the pointer! [b95c852](https://github.com/dimagi/commcare-android/pull/3097/commits/b95c85233c8c27b221d68416cd39841b0c2f63af)

app/src/org/commcare/utils/FirebaseIntegrityUtil.kt
OrangeAndGreen: Removed, thanks

app/src/org/commcare/utils/FirebaseIntegrityUtil.kt
shubham1g5: If we need the `NO_WRAP` flag, we can just create a new class in Android repo itself, say `NonceGenerator`. 

> Is our implementation of Base64 in core meant to replicate the functionality provided by the Android class?

That seems like a stripped down version of some opensoure base64 library and can only handle simple base64 encode and decode.

OrangeAndGreen: Sounds good. [cd3938c](https://github.com/dimagi/commcare-android/pull/3097/commits/cd3938c7087cb72daa47b6d7a50656264f916625)

shubham1g5 (approved):

PR #3112, opened 2025-05-16
Add photo and isDemo fields to Connect User Record

shubham1g5:
## Technical Summary

Add photo and isDemo fields to Connect User Record. These are to be used to store base 64 encoded photo String and whether a user is a demo user. 

## Labels and Review

### DIFF
--- a/app/src/org/commcare/android/database/connect/models/ConnectUserRecordV13.java
+++ b/app/src/org/commcare/android/database/connect/models/ConnectUserRecordV13.java
@@ -1,0 +1,118 @@
+package org.commcare.android.database.connect.models;
+
+import org.commcare.android.storage.framework.Persisted;
+import org.commcare.connect.ConnectConstants;
+import org.commcare.models.framework.Persisting;
+import org.commcare.modern.database.Table;
+import org.commcare.modern.models.MetaField;
+
+import java.util.Date;
+
+@Table(ConnectUserRecordV13.STORAGE_KEY)
+public class ConnectUserRecordV13 extends Persisted {
+
+    public static final String STORAGE_KEY = "user_info";
+    public static final String META_PIN = "pin";
+    public static final String META_SECONDARY_PHONE_VERIFIED = "secondary_phone_verified";
+    public static final String META_VERIFY_SECONDARY_PHONE_DATE = "verify_secondary_phone_by_date";
+
+    @Persisting(1)
+    private String userId;
+
+    @Persisting(2)
+    private String password;
+
+    @Persisting(3)
+    private String name;
+
+    @Persisting(4)
+    private String primaryPhone;
+
+    @Persisting(5)
+    private String alternatePhone;
+
+    @Persisting(6)
+    private int registrationPhase;
+
+    @Persisting(7)
+    private Date lastPasswordDate;
+
+    @Persisting(value = 8, nullable = true)
+    private String connectToken;
+
+    @Persisting(value = 9, nullable = true)
+    private Date connectTokenExpiration;
+    @Persisting(value = 10, nullable = true)
+    @MetaField(META_PIN)
+    private String pin;
+    @Persisting(11)
+    @MetaField(META_SECONDARY_PHONE_VERIFIED)
+    private boolean secondaryPhoneVerified;
+
+    @Persisting(12)
+    @MetaField(META_VERIFY_SECONDARY_PHONE_DATE)
+    private Date verifySecondaryPhoneByDate;
+
+    public ConnectUserRecordV13() {
+        registrationPhase = ConnectConstants.CONNECT_NO_ACTIVITY;
+        lastPasswordDate = new Date();
+        connectTokenExpiration = new Date();
+        secondaryPhoneVerified = true;
+        verifySecondaryPhoneByDate = new Date();
+    }
+
+    public static ConnectUserRecordV13 fromV5(ConnectUserRecordV5 oldRecord) {
+        ConnectUserRecordV13 newRecord = new ConnectUserRecordV13();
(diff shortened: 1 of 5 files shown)


app/src/org/commcare/models/database/connect/ConnectDatabaseUpgrader.java
```diff
@@ -522,6 +528,31 @@ private void upgradeTwelveThirteen(SQLiteDatabase db) {
         addTableForNewModel(db, ConnectJobDeliveryFlagRecord.STORAGE_KEY, new ConnectJobDeliveryFlagRecord());
     }
 
+    private void upgradeThirteenFourteen(SQLiteDatabase db) {
+        db.beginTransaction();
+        try {
+            SqlStorage<ConnectUserRecordV13> oldStorage = new SqlStorage<>(
```
Jignesh-dimagi, line 534: It is required to add the column in db for `photo` and `isDemo` ?

app/src/org/commcare/models/database/connect/ConnectDatabaseUpgrader.java
```diff
@@ -329,12 +335,12 @@ private void upgradeFiveSix(SQLiteDatabase db) {
...
             SqlStorage<Persistable> newStorage = new SqlStorage<>(
                     ConnectUserRecord.STORAGE_KEY,
-                    ConnectUserRecordV5.class,
+                    ConnectUserRecordV13.class,
                     new ConcreteAndroidDbHelper(c, db));
 
             for (Persistable r : oldStorage) {
                 ConnectUserRecordV5 oldRecord = (ConnectUserRecordV5)r;
-                ConnectUserRecord newRecord = ConnectUserRecord.fromV5(oldRecord);
+                ConnectUserRecordV13 newRecord = ConnectUserRecordV13.fromV5(oldRecord);
```
Jignesh-dimagi, line 343: Just a question: Why only `upgradeFiveSix` is effected and not others like `upgradeSevenEight`?

shubham1g5:

app/src/org/commcare/models/database/connect/ConnectDatabaseUpgrader.java
shubham1g5: we are not adding an explicit column but just changing the serialization scheme (all the clolums are stored as a byte[] in a single blob column in our db unless we need to query on them)

app/src/org/commcare/android/database/connect/models/ConnectUserRecordV13.java
```diff
@@ -0,0 +1,118 @@
...
+
+    public String getName() {
+        return name;
+    }
+
+    public String getPrimaryPhone() {
+        return primaryPhone;
+    }
+
+    public String getAlternatePhone() {
```
pm-dimagi, line 95: we can remove these fields now as they will not be in use

app/src/org/commcare/models/database/connect/ConnectDatabaseUpgrader.java
shubham1g5: this change is just because we renamed the old `ConnectUserRecord` to `ConnectUserRecordV13` as the record to use upto v13, since the migration happens on v5 of the db to V6,  we need to use the class that was applicable for V6 of the DB here (a lot of this has to do with how we serialize data and store it, would suggest taking a look at `SqlStorage` and in particular `TableBuilder` code to understand how we are creating tables in DB.

app/src/org/commcare/android/database/connect/models/ConnectUserRecordV13.java
```diff
@@ -0,0 +1,118 @@
...
+        newRecord.userId = oldRecord.getUserId();
+        newRecord.password = oldRecord.getPassword();
+        newRecord.name = oldRecord.getName();
+        newRecord.primaryPhone = oldRecord.getPrimaryPhone();
+        newRecord.alternatePhone = oldRecord.getAlternatePhone();
+        newRecord.registrationPhase = oldRecord.getRegistrationPhase();
+        newRecord.lastPasswordDate = oldRecord.getLastPasswordDate();
+        newRecord.connectToken = oldRecord.getConnectToken();
+        newRecord.connectTokenExpiration = oldRecord.getConnectTokenExpiration();
+        newRecord.secondaryPhoneVerified = true;
```
pm-dimagi, line 75: there should be the field of photo and isDemo  and we can add one more field of name

app/src/org/commcare/android/database/connect/models/ConnectUserRecordV13.java
shubham1g5: There are multiple usages of this method in this branch and we should do a separate DB cleanup to remove unused methods in a different PR.

app/src/org/commcare/android/database/connect/models/ConnectUserRecordV13.java
pm-dimagi: https://github.com/dimagi/commcare-android/blob/bc59f7b34de9ef9ef677ebe0cde802bc8775e67b/app/src/org/commcare/android/database/connect/models/ConnectJobRecordV7.java#L242
i am comparing with this example

pm-dimagi: yesss got it right .. now understood

app/src/org/commcare/android/database/connect/models/ConnectUserRecordV13.java
pm-dimagi: okay willl have to create the ticket for the same

PR #3113, opened 2025-05-16
- Added firebase otp send and receive helper class

jaypanchal-13:
Added firebase otp send and receive helper classes
- FirebaseAuthService class -> to handle phone number authentication using firebase
- OtpVerificationCallback class -> for callbacks for otp verification operations.
- OtpManager class -> It is a Manager class that wraps authentication service operations for OTP functionality to provide simplified interface for requesting and submitting otp
- AuthService class -> Implementations of this interface are responsible for sending OTP codes to users

Jira ticket -> https://dimagi.atlassian.net/browse/CCCT-1157


pm-dimagi (changes requested):

app/src/org/commcare/utils/OTPManager.java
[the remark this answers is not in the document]
jaypanchal-13: It is a Manager class that wraps authentication service operations for OTP functionality to provide simplified interface for requesting and submitting otp.

Jignesh-dimagi:
@jaypanchal-13 It would be nice if you can show the usage of this code, You can mention here in comment only. Sorry I am bit confused like which / how calls should be made to send Otp and then to verify Otp.

OrangeAndGreen (changes requested):
Almost there, a couple more things to do:
-Add Example Usage to the PR description to show how the code will be called from other places.
-Add javadoc comments to interfaces and public methods

jaypanchal-13:
> Add Example Usage to the PR description to show how the code will be called from other places



> @jaypanchal-13 It would be nice if you can show the usage of this code, You can mention here in comment only. Sorry I am bit confused like which / how calls should be made to send Otp and then to verify Otp.

> @Jignesh-dimagi Here is the usage of this code

- UI triggers OtpManager.requestOtp(phoneNumber)
- OtpManager delegates to FirebaseAuthService to send OTP
- Once OTP is sent, FirebaseAuthService notifies via OtpVerificationCallback
- When the user enters OTP, OtpManager.submitOtp(code) is called
- FirebaseAuthService verifies it and notifies the result through the callback

jaypanchal-13:
> Almost there, a couple more things to do: -Add Example Usage to the PR description to show how the code will be called from other places. -Add javadoc comments to interfaces and public methods

> @OrangeAndGreen  Can you please check my latest commit I have added javadoc comments in it
https://github.com/dimagi/commcare-android/pull/3113/commits/edeeb4d11d02b4251fbcf7be5383119e92fe13e1

app/src/org/commcare/utils/AuthService.java
```diff
@@ -0,0 +1,21 @@
+package org.commcare.utils;
+
+/**
+ * AuthService is an interface that defines the contract for OTP-based authentication operations.Implementations of this interface are responsible for sending OTP codes to users and verifying the OTPs they enter
+ */
+public interface AuthService {
```
shubham1g5, line 6: can we rename to `OtpAuthService` instead.

app/src/org/commcare/utils/FirebaseAuthService.java
```diff
@@ -0,0 +1,123 @@
...
+     * Verifies the OTP entered by the user.
+     *
+     * @param code The OTP code received by the user
+     */
+    @Override
+    public void verifyOtp(String code) {
+        if (verificationId == null || verificationId.isEmpty()) {
+            callback.onFailure("Missing verification ID");
+            return;
+        }
```
shubham1g5, line 94: this signifies an error in our application logic as we are probably calling this method out of order (before `onCodeSent` ). As such we should fail first and crash the application instead with an `IllegalStateException` from here.

app/src/org/commcare/utils/OtpManager.java
```diff
@@ -0,0 +1,47 @@
+package org.commcare.utils;
+
+/**+ * Manager class that wraps authentication service operations for OTP functionality.+ * Provides a simplified interface for requesting and submitting OTPs. */
```
shubham1g5, line 3: Java doc formatting seems off for all classes in the PR.

````
/**
 *
 */
````

File: app/src/org/commcare/utils/OtpManager.java [hunk 4]
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * Constructs an OtpManager with the specified AuthService implementation.
+     *
+     * @param authService An implementation of AuthService used to send and verify OTPs
+     * @throws IllegalArgumentException if authService is null
+     */
+    public OtpManager(AuthService authService) {
+        if (authService == null) {
+            throw new IllegalArgumentException("AuthService cannot be null");
+        }
```
shubham1g5, line 18: nit: can replace this with `Objects.requireNonNull`

app/src/org/commcare/utils/FirebaseAuthService.java
```diff
@@ -0,0 +1,123 @@
...
+        this.verificationId = verificationId;
+    }
+
+    /**
+     * Sends an OTP to the provided phone number using Firebase's phone auth.
+     *
+     * @param phoneNumber The phone number to which the OTP will be sent
+     */
+    @Override
+    public void sendOtp(String phoneNumber) {
```
shubham1g5, line 53: probably better to call this `requestOtp` to be more correct on what's happening here.

shubham1g5 (changes requested):

Jignesh-dimagi:
Thanks @jaypanchal-13 for clarification [here](https://github.com/dimagi/commcare-android/pull/3113#issuecomment-2892984554)

If UI is interacting with `OtpManager` only, how is `activity` set in `FirebaseAuthService`? Basically where/who is instantiating `FirebaseAuthService`. Sorry if I have missed anything in context here and asking same.

app/src/org/commcare/utils/OtpManager.java
jaypanchal-13: Updated

app/src/org/commcare/utils/OtpManager.java
jaypanchal-13: Updated

app/src/org/commcare/utils/AuthService.java
jaypanchal-13: yes renamed

app/src/org/commcare/utils/FirebaseAuthService.java
jaypanchal-13: Yes you are correct at the time of implementation we can apply a check if code is null or not then we can avoid this crash. Still implementation is pending

jaypanchal-13:
> Thanks @jaypanchal-13 for clarification [here](https://github.com/dimagi/commcare-android/pull/3113#issuecomment-2892984554)
> 
> If UI is interacting with `OtpManager` only, how is `activity` set in `FirebaseAuthService`? Basically where/who is instantiating `FirebaseAuthService`. Sorry if I have missed anything in context here and asking same.

> @Jignesh-dimagi  yes you are right OtpManager is directly communicating with ui. Instance of FirebaseAuthService is passed as a param in OtpManager class at the same time. This object (FirebaseAuthService) is probably responsible for handling Firebase OTP sending and verification logic.

app/src/org/commcare/utils/FirebaseAuthService.java
jaypanchal-13: Updated

app/src/org/commcare/utils/FirebaseAuthService.java
```diff
@@ -0,0 +1,123 @@
...
+     * Constructor for verifying OTP with a known verification ID.
+     *
+     * @param activity Android activity context
+     * @param callback Callback to handle OTP events
+     * @param verificationId ID received when OTP was sent
+     */
+    public FirebaseAuthService(Activity activity, OtpVerificationCallback callback, String verificationId) {
+        this(activity, callback);
+        this.verificationId = verificationId;
+    }
```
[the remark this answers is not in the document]
shubham1g5: but the verification ID is getting assigned as part of the `onCodeSent` call itself, as such why should someone need` to re-init this class again ?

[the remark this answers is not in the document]
jaypanchal-13: > but the verification ID is getting assigned as part of the `onCodeSent` call itself, as such why should someone need` to re-init this class again ?

> @shubham1g5 yes you are right. As of now implementation is not added so I created new constructor of it and passed verification id in that. Will remove it at the time of implementation.

Jignesh-dimagi:
@jaypanchal-13 I can see that you are trying to achieve the simple solution for otp send and verify. Its really good you have introduced the OtpManager in between UI and FirebaseAuthService. As per the current implementation, fragment is instantiating FirebaseAuthService and providing to OtpManager.  But it has problem here, OtpManager still depends upon the fragment's (UI module) FirebaseAuthService (data module) instantiation and OtpManager cannot function independently. Also, whenever FirebaseAuthService is instantiated, activity reference is stored in it which is again problem from life cycle point of view. For better solution, `data`  should be independent of UI.

If we can implement simple solution:

Fragement -> instantiate OtpManager(activity, callback) -> FirebaseAuthService() and create and store PhoneAuthOptions.Builder using activity and callback. OtpManager has FirebaseAuthService reference

Send Otp:  OtpManager.sendOtp(phoneNumber) -> firebaseAuthService.sendOtp(phoneNumber) -> add phone number in PhoneAuthOptions.Builder -> build PhoneAuthOptions -> PhoneAuthProvider.verifyPhoneNumber(PhoneAuthOptions )

Verify Otp: OtpManager.verifyOtp(code)

Your thoughts? We can even call to discuss the same. 

@shubham1g5 @OrangeAndGreen if you have other thoughts, please share.

jaypanchal-13:
> Great! @jaypanchal-13 Any concerns with the approach discussed above ?

> Ok @shubham1g5 @Jignesh-dimagi  will update it

shubham1g5 (dismissed):

shubham1g5 (approved):

shubham1g5:
@pm-dimagi can you have another look at this, it's getting blocked because you have requested changes here.

PR #3129, opened 2025-05-27
Shorter Device ID in Firebase Analytics

OrangeAndGreen:
## Technical Summary
Cropping device ID to last 36 characters when adding to Firebase Analytics as a user property. 36 is the max length for user property values, but fortunately our Device IDs are 45 characters and always start with "commcare_". So this code just crops off that opening part.


shubham1g5 (approved):

avazirna:
I'm a little concerned that this would result in having the same `device_id` represented differently in Analytics and HQ. Shouldn't we consider changing the [device ID generation algorithm](https://github.com/dimagi/commcare-android/blob/c01dce1510c5d4f6fe6aa2657c65bb94786e8677/app/src/org/commcare/utils/DeviceIdentifier.java#L33).

avazirna:
> @avazirna Analytics is mostly used for internal usage and should be fine to have a different device id there which we can match to. I would be worried about changing the device id in case external folks are relying on that format in any manner.

Makes sense

PR #3140, opened 2025-05-29
Adding app meta data as per https://dimagi.atlassian.net/browse/CCCT-1296

Jignesh-dimagi:
## Technical Summary
https://dimagi.atlassian.net/browse/CCCT-1296

## Labels and Review

### DIFF
--- a/app/src/org/commcare/utils/AppMetaDataUtil.kt
+++ b/app/src/org/commcare/utils/AppMetaDataUtil.kt
@@ -1,0 +1,8 @@
+package org.commcare.utils
+
+import android.content.Context
+
+object AppMetaDataUtil {
+
+    fun provideAppMetaData(context: Context) = hashMapOf("application_id" to context.packageName)
+}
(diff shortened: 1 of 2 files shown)


OrangeAndGreen (dismissed):

pm-dimagi (dismissed):

app/src/org/commcare/utils/AppMetaDataUtil.kt
```diff
@@ -0,0 +1,8 @@
+package org.commcare.utils
+
+import android.content.Context
+
+object AppMetaDataUtil {
+
+    fun provideAppMetaData(context: Context) = hashMapOf("application_id" to context.packageName)
```
shubham1g5, line 7: I am struggling to understand why have we decided to return it in a hashMap instead of directly using `context.packageName)` ? Are you thinking of having a single method for all app meta data like this ?

Jignesh-dimagi: @shubham1g5 We can keep adding the other app meta data over here in this hashmap and can be used where ever its required.

shubham1g5: Is not it better to have individual methods though, I doubt a single function will need all app meta data together.

Issue #2740, opened 2024-02-07
Divide By Zero Error in Edge Case with Automatic Image resizing


ctsims:
### Bug Description
There is a bug in automated image resizing when an image's existing/native size is exactly the width of the screen. 

### Current Behavior
There is a "divide by zero" error which can be triggered in a form when the app uses the multimedia images in a question label depending on the specific alignment of the width of the image and the size of the device. 

This bug appears to occur when an image is exactly the width of the potential frame size, which is the screen width minus padding.

This was replicated with a specific image and replicated with the following virtual device profile:

<img width="402" alt="image" src="https://github.com/dimagi/commcare-android/assets/155066/5051e937-ead8-4ad2-af9a-95365b1e4df3">

Adding this 456x456 image as the multimedia image for a question label with the above device screen parameters should result in a crash. I can't quite remember if the bug was when this image was 456x456 or 480x480 (exact screen size or size minus padding maybe), but resizing to one of those should work.
![download](https://github.com/dimagi/commcare-android/assets/155066/5a8569e9-10a6-45cd-b922-df9c62d9ed1a)

### Expected Behavior
Opening a form when the app uses the multimedia image as a question label shouldn’t cause a crash, and the form should display the multimedia image as part of the question label. 

### Reproduction Steps
_**Note:** These steps assume you have basic knowledge of navigating and using the CommCare platform. If you are unfamiliar with how to perform basic setup steps, a [getting started guide is available here](https://github.com/dimagi/commcare-hq/blob/master/local_dev_guide.rst)._

1. Create an app which contains the provided image as a question label
   - Use the provided sample image inside of the form for the main label of a question. It doesn’t matter what the question type is
2. Create a virtual device for replication. Using Android Studio’s device manager, replicate the above virtual device profile. The screen resolution is presumed to be the important factor
3. Install the app onto the virtual device and navigate to the form. 
   - The application should crash with the Divide By Zero bug 
   - If you see the question with the multimedia label applied, the replication has failed. You can try changing the image width to one of the other options described above, or check to confirm that your virtual device display has both the correct screen size and pixel density. 

### Possible Solution
This bug is almost certainly the result of the math for the behaviour which resizes images for the screen. 

### Additional Information/Context
Further Context: https://dimagi.atlassian.net/browse/SUPPORT-16630

### Environment details (OS name and version, etc.)
No response

### Contribution Prerequisites
[...]

PR #3143, opened 2025-06-01
Solving #2740 layout size is probably 0 when setting bitmap size

AgCaliva:
## Product Description
This solves #2740 
## Technical Summary

The problem probably was that the parent that you was using to get width for the generated bitmap was 0, because it was called before getting a size. Common problem when programming for android, is gets solved most of the time by defining a ViewTreeObserver for the View that returns size 0.
This kind of bug can be really hard to find, most of the time you get noticed that it exists by automated tests, firebase logs or very cluttered app/low phone resources available when executing.
Thats why you was not able to reproduce.
This could have been more easy to solve probably if we had the firebase logs.
Also i dont see the forum related topic, seems like it dont exists anymore.

## Labels and Review

### DIFF
--- a/app/src/org/commcare/views/EntityView.java
+++ b/app/src/org/commcare/views/EntityView.java
@@ -12,6 +12,7 @@ import android.text.TextUtils;
 import android.text.style.BackgroundColorSpan;
 import android.util.Log;
 import android.view.View;
+import android.view.ViewTreeObserver;
 import android.widget.ImageView;
 import android.widget.LinearLayout;
 import android.widget.TextView;
@@ -353,15 +354,24 @@ public class EntityView extends LinearLayout {
             return;
         }
         if (onMeasureCalled) {
-            int columnWidthInPixels = layout.getLayoutParams().width;
-            Bitmap b = MediaUtil.inflateDisplayImage(getContext(), source, columnWidthInPixels, columnWidthInPixels, true);
-            if (b == null) {
-                // Means the input stream could not be used to derive the bitmap, so showing
-                // error-indicating image
-                iv.setImageDrawable(getResources().getDrawable(R.drawable.ic_menu_archive));
-            } else {
-                iv.setImageBitmap(b);
-            }
+            //this is needed since layout width can be 0 if this function is called before parent called has size
+            layout.getViewTreeObserver().addOnGlobalLayoutListener(new ViewTreeObserver.OnGlobalLayoutListener() {
+                @Override
+                public void onGlobalLayout() {
+                    layout.getViewTreeObserver().removeOnGlobalLayoutListener(this);
+                    //width is ready
+                    int columnWidthInPixels = layout.getLayoutParams().width;
+                    Bitmap b = MediaUtil.inflateDisplayImage(getContext(), source, columnWidthInPixels, columnWidthInPixels, true);
+                    if (b == null) {
+                        // Means the input stream could not be used to derive the bitmap, so showing
+                        // error-indicating image
+                        iv.setImageDrawable(getResources().getDrawable(R.drawable.ic_menu_archive));
+                    } else {
+                        iv.setImageBitmap(b);
+                    }
+                }
+            });
+
         } else {
             // Since case list images are scaled down based on the width of the column they
             // go into, we cannot set up an image layout until onMeasure() has been called


damagatchi:

shubham1g5:
@AgCaliva Can you confirm if you have been able to repro this consistently - https://github.com/dimagi/commcare-android/issues/2740 . Think the bug in issue description is not one of temporal nature but app should consistently crash when using the right image and device size on the right reproduction.

AgCaliva:
@shubham1g5 Hello, you cant reproduce, nobody ever could as far as i understand. You probably knew about this error by reading firebase logs but it was not shared with us. The issue was not related to an imagesize, so you cant reproduce it in another way than with automatic tests, which is mentioned that was used here to confirm its existence. You just need to apply the patch and tell if the error stops displaying in firebase, there is no easy way to "reproduce". But yes im 100% sure why it happens and that it is solved, as confirmed by coderabbit, After the global-layout callback, you can safely use layout.width, which was returning 0, later was tried to divide it by 2, which gives the expected error.
This bug has been here so long because nobody ever could reproduce it following the steps provided, because it has nothing to do with the image used or the image size.

AgCaliva:
@shubham1g5 You didnt read anything i said at first?

> The problem probably was that the parent that you was using to get width for the generated bitmap was not sized yet, so was returning 0 sometimes. Common problem when programming for android, it gets solved most of the time by defining a ViewTreeObserver for the View that returns size 0.
> This kind of bug can be really hard to find, most of the time you get noticed that it exists by automated tests, firebase logs or very cluttered app/low phone resources available when executing.
> Thats why you was not able to reproduce.

I was not able to reproduce, but the bug is there and is solved, somebody that worked in this codebase can confirm it indeed, because its not the first place that adding a ViewTreeObserver method was used in the codebase, but was faulting here.

AgCaliva:
@shubham1g5 i mean  confirm 100% i used the image and tried to reproduce by setting a lot of sizes for my emulator, but none worked because it was never related to image size or image used. It was needed to call addViewTreeObserver() in the parent layout to safely use its width. Not using sizes safely results in this kind of errors like "DivideByZero"

AgCaliva:
Are you saying that this fils is not related with the original issue? If thats the case which file is related, maybe its the exactly same problem but in other file.

AgCaliva:
@kiran-tripathy im not sure who i should tag but i would like to ask for other person to review this if possible.

kiran-tripathy:
@AgCaliva Thank you for taking the time to submit this fix and engage with the issue.

After reviewing the details and attempting to reproduce the bug on our end, we haven’t been able to reliably replicate the issue. As replication is a requirement for issuing bounties in this program, we're unfortunately not able to award a bounty at this time.

We absolutely recognize the complexity here, and given the difficulty in consistently triggering the issue, we’re going to close the issue (https://github.com/dimagi/commcare-android/issues/2740) out for now.

That said, if a replication is identified in the future, whether through logs or further testing, we’ll revisit this fix. If it turns out that your proposed change does resolve the issue upon confirmed replication, we’d be happy to both credit you for the contribution and award the bounty retroactively.

Appreciate your contribution!

AgCaliva:
@kiran-tripathy someone who is not a junior dev can confirm i solved the bounty. You cant just not award it, since you knew it cant be reproduced easily since the start! You lied in the bug description. Please review this with someone with more experience. And you will have to pay me.

shubham1g5:
>  mean i confirm 100% that i used the image and tried to reproduce by setting a lot of sizes for my emulator


@AgCaliva

You still have not detailed here the exact steps taken to repro this that led you to the outcome that this is not reproducible, for example what is the CommCare HQ app you added this image in , what navigation you perform on the app to be able to reach to the form screen, how did you identify this is the related area of code which is related to this bug etc. 

Furthermore, as part of the bounty we expect you to figure out the right solution without any support from our side and it's completely unreasonable to ask for a successful bounty if the PR doesn't even touch the right code components and neither provide a detailed set of steps taken to repro the issue. I am afraid the understanding you have here of the issue is far from the actual issue which is giving you the impression that this is the right solution for the concerned issue. 

We would still be open to award you bounty if 

1. You provide detailed steps you have taken in the app to deduce that the issue is not reproducible.
2. Based on that identify right code components that the issue is present in
3. Make a potential fix in the right code components (even though it's not easily reproducible)

AgCaliva:
To me the easiest way to know if i solved the bounty is:
1. Mergue this PR since it cant break anything, and indeed solves a very feasible bug even if its not the same that triggered the bounty, and makes the code more robust. You have to add `view.getViewTreeObserver().addOnGlobalLayoutListener(new ViewTreeObserver.OnGlobalLayoutListener() { ...`

when you use a view size in a function that can be called before onLayout call.
Here you have a better explanation about:
https://cheesecakelabs.com/blog/understanding-android-views-dimensions-set/

2. Send new version to clients or internal testing. It will be better if more users have the new version since the error can happen very randomly.

3. Check firebase logs to see if it is solved.

The other way would be to use automatic testing to do hundreds of tests, but that requires a not cheap enviroment for testing.

So yes, we have been losing our time. Just merge it. Thank you.

AgCaliva:
@shubham1g5 please read.
You didnt even google about my solution by the way.

AgCaliva:
@kiran-tripathy its not serious to try to solve something without firebase logs and they can be used to track from where it comes.
Also please make the bounty bigger 150$ is too low. This kind of bug is hard to solve, is the result of only hiring juniors

AgCaliva:
I have things to pay. Can you merge it faster?

PR #3164, opened 2025-06-06
Backup code error when not matching criteria

Jignesh-dimagi:
## Technical Summary
https://dimagi.atlassian.net/browse/QA-7824

## Labels and Review

### DIFF
--- a/app/src/org/commcare/fragments/personalId/PersonalIdBackupCodeFragment.java
+++ b/app/src/org/commcare/fragments/personalId/PersonalIdBackupCodeFragment.java
@@ -143,7 +143,7 @@ public class PersonalIdBackupCodeFragment extends Fragment {
             }
         }
 
-        binding.connectBackupCodeErrorMessage.setVisibility(!backupCode1.isEmpty() && !isValid ? View.VISIBLE : View.GONE);
+        binding.connectBackupCodeErrorMessage.setVisibility(TextUtils.isEmpty(errorText) ? View.GONE : View.VISIBLE);
         binding.connectBackupCodeErrorMessage.setText(errorText);
         enableContinueButton(isValid);
     }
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/fragments/personalId/PersonalIdBackupCodeFragment.java
```diff
@@ -143,6 +143,7 @@ private void validateBackupCodeInputs() {
             }
         }
 
+        binding.connectBackupCodeErrorMessage.setVisibility(!backupCode1.isEmpty() && !isValid ? View.VISIBLE : View.GONE);
```
shubham1g5, line 146: we can just do `errorText` not empty check here ?

Jignesh-dimagi: We don't want to show any error message before user starts typing, so check for empty.

Also, you are correct to have separate method when its confirmed that there is error/no-error from multiple places. In this case, unnecessarily checks have to be made, with method overhead.

shubham1g5: > We don't want to show any error message before user starts typing, so check for empty.

why would errorText be non empty in that case ? What I am suggesting here is to tie the visibility to `errorText` itself and then set or clear `errorText` if we need to clear the error.

pm-dimagi: validateBackupCodeInputs() can also do the same thing i guess

Jignesh-dimagi: Agree, my bad, corrected

shubham1g5 (approved):

PR #3169, opened 2025-06-09
Personal ID: Biometric secret Key Invalidation

shubham1g5:
## Product Description

https://dimagi.atlassian.net/browse/CCCT-1148

## Technical Summary

[Some context here](https://docs.google.com/document/d/1ckTVDIBD8iqg70nts374fLkK70TMjT_7vUaxr9nGJVw/edit?tab=t.0#heading=h.q0vrlgncdgh3)

## Labels and Review

- [] Do we need to enhance the manual QA test coverage ? If yes, the "QA Note" label is set correctly

### DIFF
--- a/app/src/org/commcare/android/security/KeyStoreHandler.kt
+++ b/app/src/org/commcare/android/security/KeyStoreHandler.kt
@@ -1,7 +1,9 @@
 package org.commcare.android.security
 
 import org.commcare.utils.EncryptionKeyAndTransform
+import javax.crypto.SecretKey
 
 interface KeyStoreHandler {
     fun getKeyOrGenerate(): EncryptionKeyAndTransform
+    fun isKeyValid(): Boolean
 }
--- a/app/src/org/commcare/connect/PersonalIdManager.java
+++ b/app/src/org/commcare/connect/PersonalIdManager.java
@@ -43,6 +43,8 @@ import org.commcare.google.services.analytics.FirebaseAnalyticsUtil;
 import org.commcare.util.LogTypes;
 import org.commcare.utils.BiometricsHelper;
 import org.commcare.utils.CrashUtil;
+import org.commcare.utils.EncryptionKeyAndTransform;
+import org.commcare.utils.EncryptionKeyProvider;
 import org.commcare.views.dialogs.StandardAlertDialog;
 import org.javarosa.core.io.StreamsUtil;
 import org.javarosa.core.services.Logger;
@@ -62,6 +64,7 @@ import java.util.concurrent.TimeUnit;
  * @author dviggiano
  */
 public class PersonalIdManager {
+    public static final String BIOMETRIC_INVALIDATION_KEY = "biometric-invalidation-key";
     private static final long DAYS_TO_SECOND_OFFER = 30;
 
     /**
@@ -185,7 +188,12 @@ public class PersonalIdManager {
 
             @Override
             public void onAuthenticationSucceeded(@NonNull BiometricPrompt.AuthenticationResult result) {
-                callback.connectActivityComplete(true);
+                if(hasBiometricInvalidated()) {
+                    callback.connectActivityComplete(true);
+                }else {
+                    Toast.makeText(activity, "Biometric Invalidated", Toast.LENGTH_SHORT).show();
+                    callback.connectActivityComplete(false);
+                }
             }
 
             @Override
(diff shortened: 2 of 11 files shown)


app/src/org/commcare/connect/PersonalIdManager.java
```diff
@@ -177,6 +184,14 @@ public boolean isloggedIn() {
     }
 
     public void unlockConnect(CommCareActivity<?> activity, ConnectActivityCompleteListener callback) {
+        if (!isKeyValid()) {
+            GlobalErrorUtil.addError(
+                    new GlobalErrorRecord(new Date(), PERSONALID_BIOMETRIC_INVALIDATED_ERROR.ordinal()));
```
shubham1g5, line 189: @OrangeAndGreen I am not sure what's the behaviour we want here, in other places of global errors we are crashing the CommCare but changing biometric should not really cause a crash. 

I am also not sure if we want this to be a global error like this.  Furthermore, do we want to call `forgetUser` for this error ? I think not, because then there is no point of having a different key. But if we don't then the behaviour is weird as we are asking user to recover without really signing out of the device.

OrangeAndGreen: I had a conversation with Clayton about this last week and it sounds like there are two approaches we could take:
1. If we want to be safe, we can just silently log the error for now so we can see how prevalent this is before deciding on how to react to it. If we do that though, the token will need to automatically be re-established once invalidated so we're able to see if the same user is triggering the event more than once (otherwise the token would only be configured during the device configuration workflow)
2. The alternative would be some kind of deconfiguration so the user has to recover their account to regain trust. In the long run I think this is one place where we might use "soft recovery" (i.e. a simpler recovery process like just OTP or backup code). But for this release, the way we're handling those is with crashDb and an error message when the app starts up next time.

My preference is for the first option.

shubham1g5: #1 doesn't seem to add any security value and I would avoid doing more immediate work here at the end of the QA cycle for something that doesn't add any definite value to the release

#2 Given changing biometric is a practical alid scenario, I doin't really want to crash here. 


I am more inclined to call `forget User` when this happens and direct user to Login page thereafter in that case. Any concerns with that ?

OrangeAndGreen: But if we're going with the second option, I think we should use crashDb and let it add the global error, rather than simply adding the error directly here (we also want the user to be deconfigured and kicked out of whatever activity they're currently doing)

app/src/org/commcare/fragments/personalId/PersonalIdBiometricConfigFragment.java
```diff
@@ -255,6 +257,15 @@ private void navigateForward(boolean enrollmentFailed) {
         }
     }
 
+    /**
+     * Generates a biometric linked key in Android Key Store if not already there
+     */
+    private void storeBiometricInvalidationKey() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+            new EncryptionKeyProvider(requireContext(), true, BIOMETRIC_INVALIDATION_KEY).getKeyForEncryption();
+        }
```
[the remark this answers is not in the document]
OrangeAndGreen: Oops, my mistake about M being the current min! So this should never happen, but should we log an error or crash if it does?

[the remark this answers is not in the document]
shubham1g5: I removed the check instead - https://github.com/dimagi/commcare-android/pull/3169/commits/b55c04fd3b0bbbf05de52ad0a48c06d9e2b7d662

app/src/org/commcare/android/security/KeyStoreHandler.kt
```diff
@@ -1,7 +1,9 @@
 package org.commcare.android.security
 
 import org.commcare.utils.EncryptionKeyAndTransform
+import javax.crypto.SecretKey
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3169/commits/2cc4b6c8c244ebd36e954cd3ad42e188f1b06566

app/src/org/commcare/connect/PersonalIdManager.java
shubham1g5: https://github.com/dimagi/commcare-android/pull/3169/commits/2cc4b6c8c244ebd36e954cd3ad42e188f1b06566

PR #3185, opened 2025-06-11
- Solved crash on changing language

jaypanchal-13:
Solved crash on changing language
https://dimagi.atlassian.net/browse/QA-7844
## Crash Description
Crash was because it was navigating to `PersonalIdPhoneFragment `screen everytime when we are changing the language and coming back and after changing language user comes back app tries to find `PersonalIdPhoneFragment `in stack and we have minimized app from `PersonalIdBiometricConfigFragment `So it in stack app was not able to find `PersonalIdPhoneFragment  `as user minimized from `PersonalIdBiometricConfigFragment  `So I did not changed any other logic just added a sure check that if current fragment is `PersonalIdPhoneFragment  `then only it should take that navigation and checked with multiple scenarios it is not crashing now

**Crash StackTrace**
                                                                                                    Caused by: java.lang.IllegalArgumentException: Navigation action/destination org.commcare.dalvik.debug:id/action_personalid_phone_fragment_self cannot be found from the current destination Destination(org.commcare.dalvik.debug:id/personalid_otp_page) label=fragment_personalid_phone_verify class=org.commcare.fragments.personalId.PersonalIdPhoneVerificationFragment

## Labels and Review

### DIFF
--- a/app/src/org/commcare/activities/connect/PersonalIdActivity.java
+++ b/app/src/org/commcare/activities/connect/PersonalIdActivity.java
@@ -2,16 +2,14 @@ package org.commcare.activities.connect;
 
 import android.content.Intent;
 import android.os.Bundle;
-import android.util.Log;
 
 import org.commcare.activities.NavigationHostCommCareActivity;
-import org.commcare.fragments.personalId.PersonalIdBiometricConfigFragment;
 import org.commcare.connect.ConnectConstants;
 import org.commcare.connect.PersonalIdManager;
 import org.commcare.dalvik.R;
+import org.commcare.fragments.personalId.PersonalIdBiometricConfigFragment;
 import org.commcare.fragments.personalId.PersonalIdPhoneFragmentDirections;
 import org.commcare.views.dialogs.CustomProgressDialog;
-import org.javarosa.core.services.Logger;
 
 import java.util.Objects;
 


app/src/org/commcare/activities/connect/PersonalIdActivity.java
```diff
@@ -80,16 +82,38 @@ private void beginRegistration() {
...
+                    break;
+                }
         }
         if (navDirections == null) {
-            navDirections = PersonalIdPhoneFragmentDirections
-                    .actionPersonalidPhoneFragmentToPersonalidBiometricConfig();
+            if (Objects.equals(getCurrentFragmentName(), "PersonalIdPhoneFragment")){
+                navDirections = PersonalIdPhoneFragmentDirections
+                        .actionPersonalidPhoneFragmentToPersonalidBiometricConfig();
+                navController.navigate(navDirections);
```
Jignesh-dimagi, line 95: @jaypanchal-13 Both the conditions i..e line 85 to 89 and here, are checking for `PersonalIdPhoneFragment`, and one is navigating to phone fragment and other to biometric?

app/src/org/commcare/activities/connect/PersonalIdActivity.java
```diff
@@ -80,16 +82,38 @@ private void beginRegistration() {
...
         switch (PersonalIdManager.getInstance().getStatus()) {
             case NotIntroduced, Registering:
-                navDirections = PersonalIdPhoneFragmentDirections.actionPersonalidPhoneFragmentSelf();
-                break;
+                if (Objects.equals(getCurrentFragmentName(), "PersonalIdPhoneFragment")){
+                    navDirections = PersonalIdPhoneFragmentDirections.actionPersonalidPhoneFragmentSelf();
+                    navController.navigate(navDirections);
+                    break;
+                }
         }
```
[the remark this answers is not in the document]
Jignesh-dimagi: +1

Jignesh-dimagi:
@jaypanchal-13 I think we don't need [this](https://github.com/dimagi/commcare-android/blob/443f9c3e591f6f06dcc523c2367612e954604ace/app/src/org/commcare/activities/connect/PersonalIdActivity.java#L78) method. May be its related to bug originating with this function only.

app/src/org/commcare/activities/connect/PersonalIdActivity.java
jaypanchal-13: Yes @Jignesh-dimagi

jaypanchal-13:
> @jaypanchal-13 I think we don't need [this](https://github.com/dimagi/commcare-android/blob/443f9c3e591f6f06dcc523c2367612e954604ace/app/src/org/commcare/activities/connect/PersonalIdActivity.java#L78) method. May be its related to bug originating with this function only.

> @Jignesh-dimagi  Yes we can do that but as I do not know whole functionality. Did not intend to change it. And kept current logic unchanged

app/src/org/commcare/activities/connect/PersonalIdActivity.java
pm-dimagi: what is the use of checking the current fragment, is there any overlapping fragment also

PR #3192, opened 2025-06-11
Handling Next and Done keys in Backup Code Fragment

OrangeAndGreen:
https://dimagi.atlassian.net/browse/QA-7850

Added key handlers for the Next and Done keys when showing the numeric keypad in the backup code fragment.

When pressed:
-If in recovery: submits the code
-If new registration, first input: proceeds to second input
-If new registration, second input: proceeds to next page

### DIFF
--- a/app/src/org/commcare/fragments/personalId/PersonalIdBackupCodeFragment.java
+++ b/app/src/org/commcare/fragments/personalId/PersonalIdBackupCodeFragment.java
@@ -103,24 +103,22 @@ public class PersonalIdBackupCodeFragment extends Fragment {
 
     private void setupDoneKeys() {
         binding.connectBackupCodeInput.setOnEditorActionListener((v, actionId, event) -> {
-            if (actionId == EditorInfo.IME_ACTION_NEXT || actionId == EditorInfo.IME_ACTION_DONE) {
-                if(isRecovery) {
-                    handleBackupCodeSubmission();
-                } else {
-                    binding.connectBackupCodeRepeatInput.requestFocus();
-                }
+            if ((actionId == EditorInfo.IME_ACTION_NEXT || actionId == EditorInfo.IME_ACTION_DONE)
+                    && isRecovery && binding.connectBackupCodeButton.isEnabled()) {
+                handleBackupCodeSubmission();
                 return true;
             }
+
             return false;
         });
 
         binding.connectBackupCodeRepeatInput.setOnEditorActionListener((v, actionId, event) -> {
-            if (actionId == EditorInfo.IME_ACTION_NEXT || actionId == EditorInfo.IME_ACTION_DONE) {
-                if(binding.connectBackupCodeButton.isEnabled()) {
-                    handleBackupCodeSubmission();
-                }
+            if (actionId == EditorInfo.IME_ACTION_NEXT || actionId == EditorInfo.IME_ACTION_DONE
+                    && binding.connectBackupCodeButton.isEnabled()) {
+                handleBackupCodeSubmission();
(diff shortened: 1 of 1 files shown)


pm-dimagi (approved):

app/src/org/commcare/fragments/personalId/PersonalIdBackupCodeFragment.java
```diff
@@ -99,6 +101,30 @@ private void setupInputFilters() {
         binding.connectBackupCodeRepeatInput.setFilters(filters);
     }
 
+    private void setupDoneKeys() {
+        binding.connectBackupCodeInput.setOnEditorActionListener((v, actionId, event) -> {
+            if (actionId == EditorInfo.IME_ACTION_NEXT || actionId == EditorInfo.IME_ACTION_DONE) {
+                if(isRecovery) {
+                    handleBackupCodeSubmission();
+                } else {
+                    binding.connectBackupCodeRepeatInput.requestFocus();
```
shubham1g5, line 110: think this is not required.

app/src/org/commcare/fragments/personalId/PersonalIdBackupCodeFragment.java
```diff
@@ -99,6 +101,30 @@ private void setupInputFilters() {
         binding.connectBackupCodeRepeatInput.setFilters(filters);
     }
 
+    private void setupDoneKeys() {
+        binding.connectBackupCodeInput.setOnEditorActionListener((v, actionId, event) -> {
+            if (actionId == EditorInfo.IME_ACTION_NEXT || actionId == EditorInfo.IME_ACTION_DONE) {
+                if(isRecovery) {
+                    handleBackupCodeSubmission();
```
shubham1g5, line 108: should we also check ` if(binding.connectBackupCodeButton.isEnabled()) {`

Jignesh-dimagi: It should also close the keyboard in both cases?

OrangeAndGreen: Good catch: [8ec0edd](https://github.com/dimagi/commcare-android/pull/3192/commits/8ec0edd80131ce923b797d0ef18e34fe938cb29e)

I decided to leave the keypad open since a wait dialog is displayed while the call is in progress (so the user can't press keys), and then the system automatically hides the keyboard when showing the failure message or proceeding to the next screen.

app/src/org/commcare/fragments/personalId/PersonalIdBackupCodeFragment.java
OrangeAndGreen: Changed it to rely on the default handler for that case: [8ec0edd](https://github.com/dimagi/commcare-android/pull/3192/commits/8ec0edd80131ce923b797d0ef18e34fe938cb29e)

OrangeAndGreen: The issue I had before was that the handler was returning true on any Next|Done event, even if it wasn't handled. Now returning false in that case, so the default handler moves the focus to the next input

shubham1g5 (approved):

app/src/org/commcare/fragments/personalId/PersonalIdBackupCodeFragment.java
Jignesh-dimagi: @OrangeAndGreen Ok, sounds good.

PR #3198, opened 2025-06-12
Require active job in some places

OrangeAndGreen:
## Product Description
No user facing change

## Technical Summary
Added requireActiveJob function that returns active job or crashes if not present.
Using requireActiveJob instead of getActiveJob wherever the job is required.

## Feature Flag
Connect

### Safety story
The goal is to crash in a consistent way if functionality is encountered which requires an active job but one isn't set.

### Automated test coverage
None

### QA Plan
None, this is to help us with debugging when something goes wrong

### DIFF
--- a/app/src/org/commcare/activities/StandardHomeActivityUIController.java
+++ b/app/src/org/commcare/activities/StandardHomeActivityUIController.java
@@ -73,14 +73,11 @@ public class StandardHomeActivityUIController implements CommCareActivityUIContr
     }
 
     private void updateJobTileDetails() {
-        String appId = CommCareApplication.instance().getCurrentApp().getUniqueId();
-        ConnectAppRecord record = ConnectJobUtils.getAppRecord(activity, appId);
-        boolean show = record != null;
+        ConnectJobRecord job = ConnectManager.getActiveJob();
+        boolean show = job != null;
 
         viewJobCard.setVisibility(show ? View.VISIBLE : View.GONE);
         if (show) {
-            ConnectJobRecord job = ConnectManager.getActiveJob();
-            Objects.requireNonNull(job);
             TextView tvJobTitle = viewJobCard.findViewById(R.id.tv_job_title);
             TextView tvViewMore = viewJobCard.findViewById(R.id.tv_view_more);
             TextView tvJobDescription = viewJobCard.findViewById(R.id.tv_job_description);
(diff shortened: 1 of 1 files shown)


Jignesh-dimagi:
@OrangeAndGreen This looks really good idea to throw exception whenever no active job is set. This can save lot of our debugging efforts as app will crash with defined error and can be immediately found out in Firebase.
 
I was just thinking that in future, there can be many such checks/getter/setter required for connect job, so having a idea to create some base fragment like 'BaseConnectFragment` and rest all fragments `ConnectDeliveryDetailsFragment`, `ConnectDeliveryListFragment`, `ConnectDeliveryProgressDeliveryFragment`...etc can extend it. For this particular case, `OnCreateView` method of  'BaseConnectFragment` can call & check for `ConnectManager.requireActiveJob();` and child fragment can use that `ConnectJobRecord` object throughout. Thoughts?

app/src/org/commcare/connect/ConnectManager.java
```diff
@@ -180,6 +180,13 @@ public static ConnectJobRecord getActiveJob() {
         return activeJob;
     }
 
+    public static  ConnectJobRecord requireActiveJob() {
+        if (activeJob == null) {
+            throw new IllegalStateException("No active job set in ConnectManager");
```
Jignesh-dimagi, line 185: @OrangeAndGreen I am not sure but you are best person to tell if we can add more details here. I am sure we would not like to put any personal info here, but just in case if anything available.

shubham1g5 (dismissed):

shubham1g5 (commented):
How about calling `Objects.requireNotNull(job)` in `onCreate` of these fragments instead of calling `requireJob` in every call ? It simplifies overhead of knowing if a fragment expects a job to start with in comparison of monitoring individual calls and deciding where it can be null and where not.

shubham1g5:
The specific null check (typically done in constructors or init) are generally used to set up developer expectations on the class level around what params are required for proper functioning of the class so that anyone making changes to these classes are aware of those expectations.  I agree in respect to the actual app crash, it doesn't really matter how we do it.  But I do think the provided value is more aligned with code clarity and in-code documentation.

OrangeAndGreen:
Okay, catching and handling early sounds good to me as well. So at this point we've discussed three options:
1. Implement requireActiveJob method to pair with getActiveJob (similar to Android get/require Activity/Context)
2. Add Object.requireNonNull in each fragment that requires an active job
3. Create a ConnectJobFragment base class that retrieves the job and null checks in onCreate, then makes it available to child classes

I'm leaning toward #3, particularly as I could imagine additional uses for the base class in the future.

Jignesh-dimagi (approved):

PR #3207, opened 2025-06-19
Fix input lag on saved forms with multimedia in list questions

avazirna:
## Product Description
This small PR addresses an issue with saved forms that contain list questions with Image widgets. The logic to resave the form to disk is triggered whenever any question within the same list changes, leading to a poor UX due to slowness - especially when typing in a text question. This logic was implemented to update the form on disk when the image file changes, either replaced or removed.

JIRA ticket: https://dimagi.atlassian.net/browse/SAAS-17653

## Technical Summary
This change ensures that the form is resaved only when the Image widget is changed.

### Safety story
Successfully tested locally.

## Labels and Review

### DIFF
--- a/app/src/org/commcare/activities/FormEntryActivityUIController.java
+++ b/app/src/org/commcare/activities/FormEntryActivityUIController.java
@@ -756,12 +756,10 @@ public class FormEntryActivityUIController implements CommCareActivityUIControll
                 continue;
             }
 
-            if (oldWidgets.get(i) instanceof ImageWidget) {
+            if ((i == indexOfLastChangedWidget) && oldWidgets.get(i) instanceof ImageWidget) {
                 // If there was change(in particular image-remove) in an image widget,
                 // only then update the form to disk.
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/activities/FormEntryActivityUIController.java
```diff
@@ -759,7 +759,9 @@ protected void updateFormRelevancies() {
             if (oldWidgets.get(i) instanceof ImageWidget) {
                 // If there was change(in particular image-remove) in an image widget,
                 // only then update the form to disk.
-                activity.onExternalAttachmentUpdated();
+                if (i == indexOfLastChangedWidget) {
```
shubham1g5, line 762: we should put this in the outer `if` itself as the first condition so that we only do the `instanceof ImageWidget` check when required.

shubham1g5 (dismissed):
Nice catch! Can we add a link to the relevant support ticket in the comment.

shubham1g5 (approved):

PR #3209, opened 2025-06-20
Restrict Web users from logging into CC mobile

avazirna:
## Product Description
This PR enforces a restriction in CC mobile to block login attempts from Web users. With this change, Web users will now see a message as follows:

## Technical Summary
Web users are identified by their email address, so this change checks for the presence of an '@' symbol in the username. Access is allowed if the '@' symbol is not found, or it it's found and the username ends with the project space domain name.

### Safety story
Successfully tested locally.

## Labels and Review

### DIFF
--- a/app/src/org/commcare/activities/LoginActivity.java
+++ b/app/src/org/commcare/activities/LoginActivity.java
@@ -243,6 +243,11 @@ public class LoginActivity extends CommCareActivity<LoginActivity>
     }
 
     private void doLogin(LoginMode loginMode, boolean restoreSession, String passwordOrPin) {
+        if (!isUsernameValid(getUniformUsername())) {
+            raiseLoginMessage(StockMessages.Auth_BadCredentials, false);
+            return;
+        }
+
         if ("".equals(passwordOrPin) && loginMode != LoginMode.PRIMED) {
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -268,6 +273,17 @@ private void doLogin(LoginMode loginMode, boolean restoreSession, String passwor
...
 
+    private boolean isUsernameValid(String username) {
+        if (username !=null && !username.isEmpty()) {
+            if (!username.contains("@") || (username.endsWith("@" + HiddenPreferences.getUserDomain()))) {
+                return true;
+            } else {
+                return false;
+            }
+        }
+        return true;
```
shubham1g5, line 284: this should return `false` given username can't be empty.

shubham1g5 (changes requested):

app/src/org/commcare/activities/LoginActivity.java
avazirna: good point. My initial approach was around lack of permission, but then changed to `username` validity. Also changed the order of the checks, the `username` should be first

shubham1g5 (approved):

PR #3216, opened 2025-06-24
PersonalID Lockout

OrangeAndGreen:
## Product Description
After entering the wrong backup code too many times when trying to recover an account, the user will be locked out by the server and shown a message in mobile informing them.

## Technical Summary
Handling failure case when server returns an error_code instead of attempts_left.
Mobile should no longer proceed to the "create backup code" step in this scenario, and should instead exit the workflow.

## Feature Flag
PersonalID


app/src/org/commcare/fragments/personalId/PersonalIdBackupCodeFragment.java
```diff
@@ -202,10 +202,9 @@ private void confirmBackupCode() {
             protected void onSuccess(PersonalIdSessionData sessionData) {
                 if (sessionData.getDbKey() != null) {
                     handleSuccessfulRecovery();
-                } else if (sessionData.getAttemptsLeft() != null && sessionData.getAttemptsLeft() == 0) {
-                    navigateWithMessage(getString(R.string.personalid_recovery_failed_title),
-                            getString(R.string.personalid_recovery_failed_message),
-                            ConnectConstants.PERSONALID_RECOVERY_ACCOUNT_ORPHANED);
+                } else if (sessionData.getSessionFailureCode() != null &&
+                                sessionData.getSessionFailureCode().equalsIgnoreCase("LOCKED_ACCOUNT")) {
+                    handleAccountLockout();
```
Jignesh-dimagi, line 207: If server keep sending both `attempts_left` and `error_code` and if app checks first for `error_code` before `attempts_left`, it should support both versions of users?

shubham1g5: we don't really want to support orphaned account any more which is why we are removing `attempts_left ==0` check.

shubham1g5 (approved):

PR #3238, opened 2025-07-10
QA7909- fix clicking on job status 

pm-dimagi:
## Product Description
https://dimagi.atlassian.net/browse/QA-7909
Issue:- Nav host JobListFragment is calling the api and retuning the result after it got detached and the LearningProgress is up thats why it was crashing 

## Labels and Review

### DIFF
--- a/app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java
+++ b/app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java
@@ -19,6 +19,8 @@ import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewGroup;
 
+import com.google.type.DateTime;
+
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.core.view.MenuProvider;
@@ -69,6 +71,8 @@ public class ConnectJobsListsFragment extends Fragment {
     ArrayList<ConnectLoginJobListModel> jobList;
     ArrayList<ConnectLoginJobListModel> corruptJobs = new ArrayList<>();
 
+    private ConnectUserRecord userRecord;
+
     public ConnectJobsListsFragment() {
         // Required empty public constructor
     }
@@ -79,7 +83,7 @@ public class ConnectJobsListsFragment extends Fragment {
         requireActivity().setTitle(R.string.connect_title);
 
         binding = FragmentConnectJobsListBinding.inflate(inflater, container, false);
-
+        userRecord= ConnectUserDatabaseUtil.getUser(requireContext());
         launcher = (appId, isLearning) -> {
             ConnectAppUtils.INSTANCE.launchApp(getActivity(), isLearning, appId);
         };
@@ -108,8 +112,7 @@ public class ConnectJobsListsFragment extends Fragment {
 
     public void refreshData() {
         corruptJobs.clear();
-        ConnectUserRecord user = ConnectUserDatabaseUtil.getUser(getContext());
-        ApiConnect.getConnectOpportunities(getContext(), user, new IApiCallback() {
+        ApiConnect.getConnectOpportunities(getContext(), userRecord, new IApiCallback() {
             @Override
             public void processSuccess(int responseCode, InputStream responseData) {
                 int totalJobs = 0;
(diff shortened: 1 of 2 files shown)


Jignesh-dimagi:
@pm-dimagi Just a question: When clicking `View Status` button why `ConnectJobsListsFragment` is getting effected? `ConnectJobsListsFragment` should not get instantiate in this case.

app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java
```diff
@@ -19,6 +19,8 @@
 import android.view.View;
 import android.view.ViewGroup;
 
+import com.google.type.DateTime;
+
```
Jignesh-dimagi, line 23: @pm-dimagi Just a question: When clicking View Status button why ConnectJobsListsFragment is getting effected? ConnectJobsListsFragment should not get instantiate in this case.

pm-dimagi: @Jignesh-dimagi So when user clicks on view job status it redirect to dispatch activity then it calls ConnectActivity which calls the host fragment and then it replaces by learning or delivery progress fragment

app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java
```diff
@@ -79,7 +83,7 @@ public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
         requireActivity().setTitle(R.string.connect_title);
 
         binding = FragmentConnectJobsListBinding.inflate(inflater, container, false);
-
+        userRecord= ConnectUserDatabaseUtil.getUser(requireContext());
```
pm-dimagi, line 86: This is to cache the user record so that we dont have to call it ConnectUserDatabaseUtil.getUser(requireContext()) again

app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java
```diff
@@ -234,62 +237,69 @@ private void launchAppForJob(ConnectJobRecord job, boolean isLearning) {
...
-            boolean isDeliverAppInstalled = ConnectAppUtils.INSTANCE.isAppInstalled(job.getDeliveryAppInfo().getAppId());
-
-            switch (jobStatus) {
-                case STATUS_AVAILABLE_NEW, STATUS_AVAILABLE:
-                    if (!finished) {
-                        availableNewJobs.add(createJobModel(job, JOB_NEW_OPPORTUNITY, NEW_APP,
-                                true, true, false, false));
-                    }
-                    break;
+        if(getActivity()!=null) {
```
pm-dimagi, line 240: to check if this fragment is attached to activity because when there is callback from the api call this fragment will not be attached

shubham1g5 (changes requested):
@pm-dimagi From the qa issue, it seems like the issue is easily reproducible and happens everytime ? Is that right? If so, I don't think activity non null check is the right fix here as that will just fail the UI update silently in the callback. 

Also do you know what change we made that's causing this issue now ?

app/src/org/commcare/fragments/connect/ConnectLearningProgressFragment.java
```diff
@@ -52,6 +52,7 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
         setupToolbar();
         setupRefreshButton();
         populateJobCard(job);
+        updateLearningUI();
```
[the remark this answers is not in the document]
pm-dimagi: If we will not call this function it will not show any ui it will show all the textfilelds blank as it shows the cache data in the ui and if there is any change in the data the api is called after that and it will update the UI

app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java
Jignesh-dimagi: @pm-dimagi I think the main issue here is that `ConnectJobsListsFragment` is getting attached instead of status fragment directly. I think we need to remove the `startDestination` default to `connect_jobs_list_fragment`. Instead load it programmatically if no other conditions are true to show status fragment or unlock fragment. 
Here, the issue is that Job list fragment is getting attached initially and trying to show the details when its not attached to activity as code has load status fragment by that time.

Jignesh-dimagi: @pm-dimagi Yeah correct, but we can atleast set start-destination in `onCreate` of activity again

PR #3245, opened 2025-07-14
Connect smallish code improvements

shubham1g5:
## Product Description

Small changes to connect related code I made during review

Review commit by commit 

## Labels and Review

### DIFF
--- a/app/src/org/commcare/connect/ConnectAppUtils.kt
+++ b/app/src/org/commcare/connect/ConnectAppUtils.kt
@@ -79,7 +79,7 @@ object ConnectAppUtils {
 
     fun checkAutoLoginAndOverridePassword(
         context: Context?, appId: String?, username: String?,
-        passwordOrPin: String?, appLaunchedFromConnect: Boolean, uiInAutoLogin: Boolean
+        passwordOrPin: String?, appLaunchedFromConnect: Boolean, personalIdManagedLogin: Boolean
     ): String? {
         var updatedPasswordOrPin = passwordOrPin
         if (PersonalIdManager.getInstance().isloggedIn()) {
@@ -90,11 +90,11 @@ object ConnectAppUtils {
                     appId, username
                 )
                 if (record == null) {
-                    record = prepareConnectManagedApp(context, appId, username)
+                    record = storeNewConnectLinkedAppRecord(context, appId, username)
                 }
 
                 updatedPasswordOrPin = record.password
-            } else if (uiInAutoLogin) {
+            } else if (personalIdManagedLogin) {
                 val seatedAppId = CommCareApplication.instance().currentApp.uniqueId
                 val record = ConnectAppDatabaseUtil.getConnectLinkedAppRecord(
                     context, seatedAppId,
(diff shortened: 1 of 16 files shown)


shubham1g5:
Bumping this for review

app/src/org/commcare/connect/ConnectAppUtils.kt
```diff
@@ -16,30 +16,19 @@ import java.security.SecureRandom
...
 
-    private var downloading = false
-    private var downloadListener: ResourceEngineListener? = null
-
-    fun downloadAppOrResumeUpdates(installUrl: String?, listener: ResourceEngineListener?) {
-        downloadListener = listener
-        if (!downloading) {
-            downloading = true
+    fun downloadApp(installUrl: String?, listener: ResourceEngineListener?) {
+        val downloadListener: ResourceEngineListener? = listener
```
Jignesh-dimagi, line 29: no need to redefine the variable again, directly `listener` can be used

app/src/org/commcare/connect/ConnectAppUtils.kt
```diff
@@ -16,30 +16,19 @@ import java.security.SecureRandom
...
+        if (!isAppDownloading) {
+            isAppDownloading = true
```
Jignesh-dimagi, line 31: `isAppDownloading` should be made true in `startBlockingForTask`?

app/src/org/commcare/connect/ConnectAppUtils.kt
```diff
@@ -78,24 +68,25 @@ object ConnectAppUtils {
     }
 
     fun checkAutoLoginAndOverridePassword(
-        context: Context?, appId: String?, username: String?,
-        passwordOrPin: String?, appLaunchedFromConnect: Boolean, uiInAutoLogin: Boolean
+        context: Context?, username: String?,
+        passwordOrPin: String?, appLaunchedFromConnect: Boolean, personalIdManagedLogin: Boolean
     ): String? {
+        val seatedAppId = CommCareApplication.instance().currentApp.uniqueId
```
Jignesh-dimagi, line 74: @shubham1g5 This is set after login is made? I was not able to find when is this `currentApp` is set during login.

app/src/org/commcare/connect/ConnectAppUtils.kt
shubham1g5: we are setting it even before in this context before installing the task, but I will still add here to be consistent with `stopBlockingForTask`

app/src/org/commcare/connect/ConnectAppUtils.kt
```diff
@@ -16,30 +16,18 @@ import java.security.SecureRandom
 
 object ConnectAppUtils {
     private const val APP_DOWNLOAD_TASK_ID: Int = 4
-    private var primedAppIdForAutoLogin: String? = null
+    private const val IS_LAUNCH_FROM_CONNECT = "is_launch_from_connect"
 
-    fun wasAppLaunchedFromConnect(appId: String?): Boolean {
-        val primed = primedAppIdForAutoLogin
```
OrangeAndGreen, line 22: This changes how this functionality works, in that previously the call would consume the primed appId when called (setting the stored value back to null). I'm worried that changed behavior may have unintended consequences.

I'm having trouble remembering the details of why that consuming logic was necessary, but I also see that connectLaunchPerformed is no longer being used in LoginActivity. I think both of these were related to preventing the activity from attempting future auto-logins after trying once when launched from Connect.

One way or another I think this code should definitely be improved and maybe the change here is all we need now, but we should test thoroughly to make sure this change doesn't break anything.

OrangeAndGreen (approved):
One concern but otherwise approved.

PR #3255, opened 2025-07-18
Corrected the app crash after downloading the learning/delivery app

Jignesh-dimagi:
## Technical Summary
https://dimagi.atlassian.net/browse/QA-7932

### QA Plan
App shouldn't crash after downloading learn/delivery app

## Labels and Review


shubham1g5:
@Jignesh-dimagi it would be good to add more details in `Technical Description` section of the PR template on the root cause of the issue as well.

shubham1g5:
Thanks! That makes sense to me, but more generally we should be providing this info in Technical Description of PR descriptions itself to help anyone looking at the PR to get enough context on why the particular change was made.

PR #3260, opened 2025-07-21
Pm qa 7928 fix

pm-dimagi:
## Product Description
https://dimagi.atlassian.net/browse/QA-7928
Fixed this, it was updating the tile 2nd time before the api response 

## Labels and Review


pm-dimagi:
> @pm-dimagi How does this solve this issue? Previously, without these changes, also update payment confirmation should be successful only.

Yes Jignesh payment is successful only ... but the issue is updatePaymentConfirmationTile() was calling before the api result so it was showing the dialog again because payment is not marked confirmed

Jignesh-dimagi (approved):

PR #3266, opened 2025-07-23
Better "unavailable" error handling in biometric config page

OrangeAndGreen:
https://dimagi.atlassian.net/browse/CI-178

## Product Description
When there is an issue with the user's biometric or PIN hardware being unavailable to the app, the user will see the error message when they click the button corresponding to each option, rather than when the page loads.

## Technical Summary
Changed biometric config page to show "unavailable" error only when relevant security mechanism is chosen (not at page load).
Split PIN/biometric unavailable messages apart for better localization. Cleaned up a couple code warnings in PersonalIdBiometricConfigFragment. Changed unavailable message not to quit activity when finished (so user can try the other option).

## Feature Flag
PersonalID

### Safety story
Tested locally, although I don't have a device that fails the hardware check so could only test greenlight path.

### Automated test coverage
None

### QA Plan
Would be great to test recovery as a Connect user with a device that gives the "fingerprint unavailable" error to make sure the user can successfully proceed with the PIN option, but that may not be possible since this error is not easily reproducible.

### DIFF
--- a/app/src/org/commcare/fragments/personalId/PersonalIdBiometricConfigFragment.java
+++ b/app/src/org/commcare/fragments/personalId/PersonalIdBiometricConfigFragment.java
@@ -10,7 +10,6 @@ import android.widget.Toast;
 import androidx.annotation.NonNull;
 import androidx.biometric.BiometricManager;
 import androidx.biometric.BiometricPrompt;
-import androidx.fragment.app.Fragment;
 import androidx.lifecycle.ViewModelProvider;
 import androidx.navigation.NavDirections;
 import androidx.navigation.Navigation;
@@ -27,6 +26,8 @@ import org.commcare.util.LogTypes;
 import org.commcare.utils.BiometricsHelper;
 import org.commcare.utils.EncryptionKeyProvider;
 import org.javarosa.core.services.Logger;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 import java.util.Locale;
 
@@ -42,7 +43,7 @@ import static org.commcare.utils.ViewUtils.showSnackBarWithOk;
 /**
  * Fragment that handles biometric or PIN verification for Connect ID authentication.
  */
-public class PersonalIdBiometricConfigFragment extends Fragment {
+public class PersonalIdBiometricConfigFragment extends BasePersonalIdFragment {
     private BiometricManager biometricManager;
     private BiometricPrompt.AuthenticationCallback biometricCallback;
     private ScreenPersonalidVerifyBinding binding;
@@ -282,7 +283,7 @@ public class PersonalIdBiometricConfigFragment extends Fragment {
     private void navigateForward(boolean enrollmentFailed) {
         if (enrollmentFailed) {
             FirebaseAnalyticsUtil.reportPersonalIdConfigurationFailure(AnalyticsParamValue.BIOMETRIC_ENROLLMENT_FAILED);
-            Navigation.findNavController(binding.connectVerifyFingerprintButton).navigate(navigateToBiometricEnrollmentFailed());
+            navigateToBiometricEnrollmentFailed();
         } else {
             BiometricsHelper.ConfigurationStatus fingerprint = BiometricsHelper.checkFingerprintStatus(
                     getActivity(), biometricManager);
@@ -316,25 +317,25 @@ public class PersonalIdBiometricConfigFragment extends Fragment {
         }
     }
 
-    private NavDirections navigateToBiometricEnrollmentFailed() {
-        return PersonalIdBiometricConfigFragmentDirections.actionPersonalidBiometricConfigToPersonalidMessage(
+    private void navigateToBiometricEnrollmentFailed() {
+        navigateToMessageDisplay(
                 getString(R.string.connect_biometric_enroll_fail_title),
                 getString(R.string.connect_biometric_enroll_fail_message),
+                true,
                 ConnectConstants.PERSONALID_BIOMETRIC_ENROLL_FAIL,
-                getString(R.string.connect_biometric_enroll_fail_button),
-                null);
+                R.string.connect_biometric_enroll_fail_button);
     }
 
     private void navigateToMessageDisplayForSecurityConfigurationFailure(String errorMessage, boolean allowRetry) {
         int specifier = allowRetry ? ConnectConstants.PERSONALID_DEVICE_CONFIGURATION_ISSUE_WARNING
                 : ConnectConstants.PERSONALID_DEVICE_CONFIGURATION_FAILED;
 
-        NavDirections navDirections =
-                PersonalIdBiometricConfigFragmentDirections.actionPersonalidBiometricConfigToPersonalidMessage(
-                        getString(R.string.personalid_configuration_process_failed_title),
-                        errorMessage, specifier, getString(R.string.ok), null)
-                        .setIsCancellable(false);
-        Navigation.findNavController(requireView()).navigate(navDirections);
+        navigateToMessageDisplay(
+                getString(R.string.personalid_configuration_process_failed_title),
+                errorMessage,
+                false,
+                specifier,
+                R.string.ok);
     }
 
     private NavDirections navigateToOtpScreen() {
@@ -343,5 +344,13 @@ public class PersonalIdBiometricConfigFragment extends Fragment {
 
     private NavDirections navigateToNameScreen() {
         return PersonalIdBiometricConfigFragmentDirections.actionPersonalidBiometricConfigToPersonalidName();
+    }
+
+    @Override
+    protected void navigateToMessageDisplay(@NotNull String title, @Nullable String message, boolean isCancellable,
+            int phase, int buttonText) {
+        NavDirections navDirections = PersonalIdBiometricConfigFragmentDirections.actionPersonalidBiometricConfigToPersonalidMessage(
+                title, message, phase, getString(buttonText), null).setIsCancellable(isCancellable);
+        Navigation.findNavController(binding.getRoot()).navigate(navDirections);
     }
 }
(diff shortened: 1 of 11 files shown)


app/res/values/strings.xml
```diff
@@ -604,7 +604,8 @@
     <string name="personalid_configuration_process_failed_temporary_unavailable">Temporary error in establishing your device eligibility. Please try again after a day.</string>
     <string name="personalid_configuration_process_failed_network_error">Network error in establishing your device eligibility. Please make sure you have a good network connection and try again.</string>
     <string name="personalid_configuration_process_failed_unexpected_error">Unable to establish device eligibility due to an unexpected error. Please contact customer support if the problem persists</string>
-    <string name="personalid_configuration_process_failed_security_subtitle">Your device isn\'t eligible to sign up for PersonalID at this time due to non availability of %s security feature. Please try again on a different device.</string>
+    <string name="personalid_configuration_process_pin_unavailable_message">The PIN security mechanism is unavailable, please fix this issue or try again with a different device.</string>
```
shubham1g5, line 607: > please fix this issue 

Is there something user can do to fix this ? If not, probably just be more plain here and say `your device is not supported to sign up due to unavailibility of pin security configuration` ?

shubham1g5 (changes requested):
Curious why do we want to even show pin or fingerprint options if they are not available on the device ? Instead we should just show the available options and if nothing is available show user a blocking error directly

app/res/values/strings.xml
OrangeAndGreen: There are 5 possible errors that could be returned from BiometricManager, perhaps we want to address each more directly? One situation suggests the user should "try again later", and another corresponds to a security update being required

app/src/org/commcare/fragments/personalId/PersonalIdBiometricConfigFragment.java
```diff
@@ -196,28 +182,65 @@ private void updatePinSection(String buttonText) {
...
+                    navigateForward(true);
+                }
+                break;
         }
     }
 
+    private void showBiometricNotAvailableError() {
+        String message = BiometricsHelper.getBiometricHardwareUnavailableError(requireActivity());
+        FirebaseAnalyticsUtil.reportPersonalIdConfigurationFailure(AnalyticsParamValue.MIN_BIOMETRIC_HARDWARE_ABSENT);
+        Logger.log(LogTypes.TYPE_MAINTENANCE, message);
```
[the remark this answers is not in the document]
OrangeAndGreen: [123ad5f](https://github.com/dimagi/commcare-android/pull/3266/commits/123ad5fa9eb124dd7d0c826007a7294ff2b23f1a)

app/src/org/commcare/fragments/personalId/PersonalIdBiometricConfigFragment.java
```diff
@@ -196,28 +182,65 @@ private void updatePinSection(String buttonText) {
...
+        navigateToMessageDisplayForSecurityConfigurationFailure(message);
```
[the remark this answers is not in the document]
OrangeAndGreen: I think the answer to this depends on how much assumption we want to make about the state of the error. If we want the user to be able to try again easily (see other comments) this seems the way to go, but if we're confident that the only good path forward is to use PIN then we could add the extra navigation after showing the message. I'd probably change the button text on that message to say "Use PIN instead" so it's clear to the user what's going to happen next.

OrangeAndGreen:
I spent some more time looking into the different error states that could come back from BiometricManager and we combine into one "Unavailable" state. It seems like it's worth handling more of those states individually in our code so we can provide better messaging to the user. For example, we could better distinguish between the hardware actually not being present on the device (no fingerprint scanner) vs. the hardware being unavailable due to needing a software update.

But even if/when we improve the messaging, I still worry about hiding options (mostly the fingerprint) when there's an error accessing it. For example, suppose I'm a user that would prefer to use fingerprint rather than PIN if I'm given the option to use either.
* Scenario A: If I go to the page and see both options, then click fingerprint and see an error indicating that my fingerprint sensor isn't available, I may decide to go troubleshoot the fingerprint scanner or I may decide to go the PIN route instead.
* Scenario B: If I go to the page and only see the PIN option, I may never know that fingerprint was supposed to be an option. I'm not given the chance to fix the error, instead I'm just defaulted into lower security.

app/src/org/commcare/fragments/personalId/PersonalIdMessageFragment.java
```diff
@@ -136,7 +136,6 @@ private void finish() {
                 }
 
                 break;
-            case ConnectConstants.PERSONALID_DEVICE_CONFIGURATION_FAILED:
```
Jignesh-dimagi, line 139: Now, this will not close the activity whenever there is error in start_configuration flow or integrity error, as it is using this variable [here](https://github.com/dimagi/commcare-android/blob/519688017e84722d7c8110540f28a455d68d96a4/app/src/org/commcare/fragments/personalId/PersonalIdPhoneFragment.java#L446)

app/src/org/commcare/activities/connect/PersonalIdActivity.java
```diff
@@ -27,7 +27,7 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
         if (requestCode == ConnectConstants.PERSONALID_UNLOCK_PIN
                 || requestCode == ConnectConstants.CONFIGURE_BIOMETRIC_REQUEST_CODE) {
```
Jignesh-dimagi, line 28: ConnectConstants.PERSONALID_UNLOCK_PIN is not used any more, not sure if we need to remove that code

app/src/org/commcare/utils/BiometricsHelper.java
```diff
@@ -217,41 +217,27 @@ private static boolean configureBiometric(Activity activity, int authenticator)
...
-            case BIOMETRIC_TYPE ->
-                    getMinBioMetricHardwareErrorForSecurityIfAny(activity, fingerprintStatus);
-            default -> {
-                crashWithInvalidSecurityTypeException(activity, requiredLock);
-                yield null;
-            }
-        };
+        if (!requiredLock.equals(PIN) && !requiredLock.equals(BIOMETRIC_TYPE)) {
+            crashWithInvalidSecurityTypeException(requiredLock);
+        }
```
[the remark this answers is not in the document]
OrangeAndGreen: Done as part of this commit: [529bb7c](https://github.com/dimagi/commcare-android/pull/3266/commits/529bb7c4f946d1d252b0af926a047643d56eeb7c)

app/src/org/commcare/fragments/personalId/PersonalIdMessageFragment.java
OrangeAndGreen: Yeah, that's intentional since the user may be able to take additional actions and we don't want them to be immediately kicked out of the device configuration workflow. For example, after seeing an error when clicking the fingerprint button, they may choose to use the PIN option instead (if available).

app/res/values/strings.xml
OrangeAndGreen: I broke out some additional error states so we can provide better messaging to the user. In the case where the device doesn't have biometric hardware, we'll either hide that option (for Connect users that can use PIN instead), or go directly to an error message (for non-Connect users where biometrics are required). [529bb7c](https://github.com/dimagi/commcare-android/pull/3266/commits/529bb7c4f946d1d252b0af926a047643d56eeb7c)

app/res/values/strings.xml
```diff
@@ -604,7 +604,11 @@
...
+    <string name="personalid_configuration_process_pin_unavailable_message">The PIN security mechanism is unavailable, please try again later.</string>
+    <string name="personalid_configuration_process_biometric_unavailable_message">The biometric security mechanism is unavailable, please try again later.</string>
+    <string name="personalid_configuration_process_pin_needs_update_message">The PIN security mechanism requires a software update, please apply any available system updates or try again with a different device.</string>
+    <string name="personalid_configuration_process_biometric_needs_update_message">The biometric security mechanism requires a software update, please apply any available system updates or try again with a different device.</string>
+    <string name="personalid_configuration_process_biometric_no_hardware_message">PersonalID requires that your device have a biometric sensor. Please try again with a different device.</string>
```
shubham1g5, line 611: should we explicitly say "fingerprint scanner" instead of "biometric sensor" here ?

app/src/org/commcare/utils/BiometricsHelper.java
```diff
@@ -167,6 +169,14 @@ public static ConfigurationStatus checkStatus(Context context, BiometricManager
             case BiometricManager.BIOMETRIC_ERROR_NONE_ENROLLED -> {
                 return ConfigurationStatus.NotConfigured;
             }
+            case BiometricManager.BIOMETRIC_ERROR_NO_HARDWARE -> {
+                return ConfigurationStatus.NoHardware;
+            }
+            case BiometricManager.BIOMETRIC_STATUS_UNKNOWN,
+                 BiometricManager.BIOMETRIC_ERROR_SECURITY_UPDATE_REQUIRED,
+                 BiometricManager.BIOMETRIC_ERROR_UNSUPPORTED-> {
+                return ConfigurationStatus.NeedsUpdate;
```
shubham1g5, line 178: think NeedsUpdate should only map to `BIOMETRIC_ERROR_SECURITY_UPDATE_REQUIRED` ? and rest of these should still be under `NotAvailable` ?

OrangeAndGreen: From the code documentation on the other values I gathered they're also most likely fixed by updates. Specifically:
* UNKNOWN: 
This status code may be returned on older Android versions due to partial incompatibility
with a newer API. Applications that wish to enable biometric authentication on affected
devices may still call {@code BiometricPrompt#authenticate()} after receiving this status
code but should be prepared to handle possible errors.

* UNSUPPORTED:
The user can't authenticate because the specified options are incompatible with the current
Android version.

app/src/org/commcare/fragments/personalId/PersonalIdMessageFragment.java
shubham1g5: @OrangeAndGreen this is now breaking other handling that uses this code and think you will need to use a specific code for use case you are speaking about instead of removing a general purpose configuration failed handler here.

shubham1g5 (approved):

PR #3289, opened 2025-08-06
Crash analytics bugs

jaypanchal-13:
## Product Description
Ticket -> https://dimagi.atlassian.net/browse/CCCT-1532
https://dimagi.atlassian.net/browse/CCCT-1534
https://dimagi.atlassian.net/browse/CCCT-1533

## Technical Summary
**Bug 1: Crash on notification-based launch (null job)
Cause:**
When ConnectActivity is launched via a notification with GO_TO_JOB_STATUS=true but without a valid OPPORTUNITY_ID, the job remains null, causing a NullPointerException during
job.getStatus().

**Fix:**
Added a null check before accessing job. If it's null, we log the error using Crashlytics and safely default to the job list screen.

**Bug 2: Crash when activity is unexpectedly null in tokenUnavailable()
Cause:**
The activity reference in the tokenUnavailable() callback can be unexpectedly null, especially if the user has backgrounded the app or the activity was destroyed.
Attempting to use it without checking led to a NullPointerException.

**Fix:**
Added a null check for activity.
If null, we log the anomaly using FirebaseCrashlytics for future analysis.

**Bug 3: Crash in StandardHomeActivityUIController due to null ConnectJobRecord and calling updateConnectProgress without before calling setupUI
Cause:**
activity.getActiveJob() could return null if no job was assigned, but code in setupJobTile() and updateConnectProgress() assumed it was non-null, leading to a NullPointerException.

**Fix:**
Added null checks before accessing job fields. The job card is now hidden when no active job is present. Also added Crashlytics logs to track unexpected UI states.

## Labels and Review

### DIFF
--- a/app/src/org/commcare/activities/StandardHomeActivityUIController.java
+++ b/app/src/org/commcare/activities/StandardHomeActivityUIController.java
@@ -12,6 +12,8 @@ import androidx.recyclerview.widget.LinearLayoutManager;
 import androidx.recyclerview.widget.RecyclerView;
 import androidx.recyclerview.widget.StaggeredGridLayoutManager;
 
+import com.google.firebase.crashlytics.FirebaseCrashlytics;
+
 import org.commcare.CommCareApp;
 import org.commcare.CommCareApplication;
 import org.commcare.adapters.ConnectProgressJobSummaryAdapter;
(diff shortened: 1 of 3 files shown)


shubham1g5:
@jaypanchal-13 In general, I don't agree with having a null check as a way to hide the crash. We follow fail first approach in our code base that is we prefer to crash the app instead of doing a null check to hide it. Instead we need to manage the state/lifecycle of code correctly to init the vars at right time. 

Given that, think all these three issues will have separate solution and I find it quite hard to reason for all three in the same thread, will you be able to create separate PRs for each of these issue instead. When you create a new PR can you also include the repro steps one can take to reproduce the issue if that's available in addition to your reasoning on crashes.

PR #3292, opened 2025-08-11
Fixes navigtation drawer to only be visible for required Screens

shubham1g5:
## Product Description

Fixes a bug in nav drawer implementation which was causing this to be visible for all screens. 

## Labels and Review


app/src/org/commcare/activities/LoginActivity.java
```diff
@@ -180,6 +175,13 @@ && new RootBeer(this).isRooted()) {
         }
     }
 
+    private void initPersonaIdManager() {
+        if (personalIdManager == null) {
```
OrangeAndGreen, line 179: Nit: Seems like personalIdManager will always be null here since this method is called in onCreate.

OrangeAndGreen (approved):

PR #3307, opened 2025-08-27
Jignesh/feat/ccct 1365

Jignesh-dimagi:
## Technical Summary
https://dimagi.atlassian.net/browse/CCCT-1365

### QA Plan
QA can test the application by sending different types of push notification and clicking on these push notifications

## Labels and Review

### DIFF
--- a/app/src/org/commcare/activities/DispatchActivity.java
+++ b/app/src/org/commcare/activities/DispatchActivity.java
@@ -22,6 +22,7 @@ import org.commcare.recovery.measures.ExecuteRecoveryMeasuresActivity;
 import org.commcare.recovery.measures.RecoveryMeasuresHelper;
 import org.commcare.utils.AndroidShortcuts;
 import org.commcare.utils.CommCareLifecycleUtils;
+import org.commcare.utils.FirebaseMessagingUtil;
 import org.commcare.utils.MultipleAppsUtil;
 import org.commcare.utils.SessionUnavailableException;
 import org.javarosa.core.services.locale.Localization;
@@ -100,6 +101,14 @@ public class DispatchActivity extends AppCompatActivity {
         }
     }
 
+
+    private void checkIfAnyPNIntentPresent(){
+        Intent pnIntent = FirebaseMessagingUtil.getIntentForPNIfAny(this,getIntent());
+        if(pnIntent!=null){
+            startActivity(pnIntent);
+        }
+    }
+
     /**
      * A workaround required by Android Bug #2373 -- An app launched from the Google Play store
      * has different intent flags than one launched from the App launcher, which ruins the back
@@ -127,6 +136,7 @@ public class DispatchActivity extends AppCompatActivity {
             finish();
         } else {
             dispatch();
+            checkIfAnyPNIntentPresent();
         }
     }
 
--- a/app/src/org/commcare/fragments/connect/ConnectUnlockFragment.java
+++ b/app/src/org/commcare/fragments/connect/ConnectUnlockFragment.java
@@ -4,6 +4,8 @@ import static org.commcare.connect.ConnectConstants.REDIRECT_ACTION;
 import static org.commcare.connect.ConnectConstants.SHOW_LAUNCH_BUTTON;
 
 import android.os.Bundle;
+import android.os.Handler;
+import android.os.Looper;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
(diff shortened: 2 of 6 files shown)


app/src/org/commcare/fragments/connect/ConnectUnlockFragment.java
```diff
@@ -55,17 +57,39 @@ public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
...
-                retrieveOpportunities();
-            } else {
-                requireActivity().finish();
-            }
-        });
-
         return binding.getRoot();
     }
 
+    private final Runnable unlockRunnable = new Runnable() {
```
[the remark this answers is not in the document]
shubham1g5: Note for future: this was required to fix a fragment transaction crash - 

````
 Fatal Exception: java.lang.IllegalStateException: FragmentManager is already executing transactions
   at androidx.fragment.app.FragmentManager.ensureExecReady(FragmentManager.java:1717)
   at androidx.fragment.app.FragmentManager.execPendingActions(FragmentManager.java:1776)
   at androidx.fragment.app.FragmentManager.executePendingTransactions(FragmentManager.java:641)
   at androidx.biometric.BiometricPrompt.findOrAddBiometricFragment(BiometricPrompt.java:1082)
   at androidx.biometric.BiometricPrompt.authenticateInternal(BiometricPrompt.java:992)
   at androidx.biometric.BiometricPrompt.authenticate(BiometricPrompt.java:972)
   at org.commcare.utils.BiometricsHelper.authenticatePinOrBiometric(BiometricsHelper.java:149)
   at org.commcare.utils.BiometricsHelper.authenticatePin(BiometricsHelper.java:127)
   at org.commcare.connect.PersonalIdManager.unlockConnect(PersonalIdManager.java:192)
   at org.commcare.fragments.connect.ConnectUnlockFragment.onCreateView(ConnectUnlockFragment.java:71)
   at androidx.fragment.app.Fragment.performCreateView(Fragment.java:3114)
   ````

**Comment by shubham1g5** [comment]
@Jignesh-dimagi thanks, although still not clear on how this happens - 

> Also, if user clicks on push notification (which is received when user was logged in) and if user is logout, app will not receive any intent which will make sure that it will not start any activity related to PersonalId or connect.

File: app/src/org/commcare/activities/DispatchActivity.java [hunk 2]
```diff
@@ -221,6 +230,7 @@ private void dispatch() {
                 launchLoginScreen();
             }
         }
+        checkIfAnyPNIntentPresent();
```
shubham1g5, line 233: @Jignesh-dimagi Suppose as part of dispatch we end up calling `launchLoginScreen` above (i.e user was not logged into a CC app when they clicked on notificaion), it seems like we will first start the `LoginActivity`  and then on top of it, will start the screen related to the notification intent. This is a problematic behaviour as we are calling startActivity back to back for 2 different activities and think will screw up the navigation stack. 

My sense is that we should instead be calling `checkIfAnyPNIntentPresent` first thing in the dispatch (for personal ID notifications) and ignore the rest of dispatch handling if we redirect user to a screen as part of notification handling. Also think we need to be separating out the Personal ID notification handling here from the other CC notification handling. The idea being, if user is logged into Personal ID and user click on a personal ID notification, we don't need to depend on any other state being managed in `dispatch` and can safely direct user to the related personal ID screen. 

Although, if user has clicked on a notification which is a CommCare notification, we do need to depend on other handling here (like making sure user is logged in first into CC app, before directing user to a screen that should only appear after a user login into the CC App. 


Also open to other ideas if you have thoughts on how to manage this state more seamlessly.

shubham1g5 (changes requested):

shubham1g5:
> It starts from [here](https://github.com/dimagi/commcare-android/blob/a29e137b38405fdf053fa183e145b3bdfb161f90/app/src/org/commcare/utils/FirebaseMessagingUtil.java#L155) and follows the code ahead

This only is true for CCC notifications though, if we call the `handleNotifications` with a general intent with any random action, that will eventually lands to `handleGeneralApplicationPushNotification` and will end up showing a notification even if the intent was not meant to show a notification.

app/src/org/commcare/utils/FirebaseMessagingUtil.java
```diff
@@ -324,7 +325,7 @@ private static Intent handleGeneralApplicationPushNotification(Context context,
         Intent intent = new Intent(context, DispatchActivity.class);
         intent.setAction(Intent.ACTION_MAIN);
         intent.addCategory(Intent.CATEGORY_LAUNCHER);
```
shubham1g5, line 327: we also don't really need to build related intents when showNotification is false.

shubham1g5 (changes requested):

app/src/org/commcare/utils/FirebaseMessagingUtil.java
Jignesh-dimagi: @shubham1g5 Done

shubham1g5 (approved):

Jignesh-dimagi:
@shubham1g5 Your action required again as merged master. Thanks.

Jignesh-dimagi:
@damagatchi retest please

Jignesh-dimagi:

PR #3309, opened 2025-08-27
Fix bug in integrity reporting

OrangeAndGreen:
https://dimagi.atlassian.net/browse/CCCT-1658

## Product Description
No user visible changes.

## Technical Summary
Fixes a crash that was coming out of the integrity reporting code that gets triggered by a flag in the HQ heartbeat.
The code was trying to observer a LiveData object while waiting to retrieve a token.
But calling observeForever from a background thread caused an IllegalStateException.
I wrapped the crashing code to marshal the observer back to the main thread, and confirmed the crash no longer happens.

## Feature Flag
HQ Heartbeat, Integrity Reporting

### Safety story
Forced a test locally first to observe the crash, then to confirm it didn't happen after the fix.

### Automated test coverage
None

### QA Plan
None


app/src/org/commcare/android/integrity/IntegrityTokenApiRequestHelper.kt
```diff
@@ -157,8 +159,12 @@ class IntegrityTokenApiRequestHelper(
                     }
                 }
             }
-            viewModel.providerState.observeForever(observer)
-            cont.invokeOnCancellation { viewModel.providerState.removeObserver(observer) }
+            kotlinx.coroutines.runBlocking {
```
shubham1g5, line 162: no need to use `runBlocking` as we are inside a `suspend` method

shubham1g5 (changes requested):
@OrangeAndGreen I am prettry concerned about this code as it seems that we are trying to force main thread execution from a work manager task and I am not really sure of the repurcussions of doing that. IMO, we should not need rely on `observeForever` call if that require us to switch the execution to main thread. 

There is another Kotlin component called Flow using which you might be able to simplify things as an alternative here - `viewModel.providerState.asFlow()`

PR #3310, opened 2025-08-28
FIx NPE during CommCareSetupActivity menu preparation

avazirna:
## Product Description
Crashlytics issue: https://console.firebase.google.com/project/commcare-a57e4/crashlytics/app/android:org.commcare.dalvik/issues/7f24c229a2dfd2918fbdd7fc8981f50d

Ticket: https://dimagi.atlassian.net/browse/CI-21

## Labels and Review


app/src/org/commcare/activities/CommCareSetupActivity.java
```diff
@@ -522,10 +522,12 @@ public boolean onPrepareOptionsMenu(Menu menu) {
         }
         
         MenuItem refreshItem = menu.findItem(MENU_REFRESH_OPPORTUNITIES);
-        boolean showRefreshMenu =  !fromExternal &&
-                PersonalIdManager.getInstance().isloggedIn() &&
-                !ConnectUserDatabaseUtil.hasConnectAccess(this);
-        refreshItem.setVisible(showRefreshMenu);
+        if (refreshItem != null) {
```
shubham1g5, line 525: Any idea what's causing this to be `null` ?

shubham1g5 (approved):
Looks harmless as such, but would be curious around what's causing this to be `null` and if you were able to repro it ?

app/src/org/commcare/activities/CommCareSetupActivity.java
avazirna: I tried to replicate a few times but I wasn't able to. What I noticed is that we call `invalidateOptionsMenu()` in a few places and that when we start the activity, the method `onPrepareOptionsMenu()` is called twice. The latter makes me think that maybe when the activity is recreated (due to a screen rotation), a delay is introduced, I tried this and couldn't replicate still.

PR #3311, opened 2025-08-28
fix issue with connect hearbeat due to incorrect sequencing

shubham1g5:
## Product Description
https://dimagi.atlassian.net/browse/CCCT-1657

## Technical Summary

If `isloggedIn()` is false, we were immediately calling `continuation.resume(Result.failure())` without exiting the code which in turn call heartbeatRequest, and will eventually invoke onSuccess or onFailure resulting into another call to `continuation.resume` and hence the exception `java.lang.IllegalStateException: Already resumed`

## Labels and Review

### DIFF
--- a/app/src/org/commcare/connect/workers/ConnectHeartbeatWorker.kt
+++ b/app/src/org/commcare/connect/workers/ConnectHeartbeatWorker.kt
@@ -12,18 +12,14 @@ class ConnectHeartbeatWorker(val context: Context, workerParams: WorkerParameter
         CoroutineWorker(context, workerParams) {
 
     override suspend fun doWork(): Result {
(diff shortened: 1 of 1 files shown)


Jignesh-dimagi (approved):

OrangeAndGreen (approved):
Approving as the fix for the logged out scenario should definitely be applied. But it doesn't seem to me that this alone explains the large number of exceptions we've seen coming from this code this week.
The Connect heartbeat doesn't start scheduling until the user has either just configured PersonalID on the device or unlocked (via PIN/biometric) for the first time since starting the app. In either case the device should only be logged out at this point if the user chose to "forget PersonalID" on the device, and then the heartbeat executes again.

Meanwhile, we've recorded nearly 700 of these crashes in under 7 days

PR #3322, opened 2025-09-04
Show update prompt during user session

avazirna:
## Product Description
It was noticed that when the `Prompt Updates` option is set to `Force`, users still need to wait until the next login for the update to be triggered. This workflow is only triggered during the login process and because CommCare session duration is 24 hours by default, forced updates can take up to 1 day to be applied, which is not consistent with how this feature is supposed to work. This PR makes the workflow to be triggered when users navigates back to the `Home Screen` and are not being directed to any other screen, see video below.

<video src="https://github.com/user-attachments/assets/4cc1e89b-8f8e-41c9-bfc3-a8e8d604acab" width="200px"> </video>

Ticket: https://dimagi.atlassian.net/browse/SAAS-15946

### Safety story
This was successfully tested locally. 

## Labels and Review


avazirna:

shubham1g5 (approved):
Looks good, but would be nice to rename the `onlyForcedUpdate` or `showForcedOnly` to use the same variable name - `skipOptionalPrompt`

PR #3329, opened 2025-09-09
Not immediately failing when retrieving an integrity token fails.

OrangeAndGreen:
## Product Description
App no longer immediately blocks the user if retrieving an integrity token fails at the start of the device configuration workflow.
Instead, the server can decide whether the user can still proceed (i.e. for Connect vs. non-Connect users).

## Technical Summary
Changing behavior so workflow proceeds after integrity token error and sends empty strings for the hash and token

## Feature Flag
PersonalID

### Safety story
Tested locally to verify:
1. Normal functionality when integrity token retrieval succeeds
2. Connect user can proceed with empty token values (manually overrode the code to test)
3. Non-Connect user can NOT proceed with empty token values (manually overrode the code to test)

### Automated test coverage
None

### QA Plan
Repeat the tests performed by the dev (see above)

### DIFF
--- a/app/src/org/commcare/fragments/personalId/PersonalIdPhoneFragment.java
+++ b/app/src/org/commcare/fragments/personalId/PersonalIdPhoneFragment.java
@@ -2,7 +2,6 @@ package org.commcare.fragments.personalId;
 
 import android.Manifest;
 import android.app.Activity;
-import android.app.Dialog;
 import android.content.DialogInterface;
 import android.location.Location;
 import android.os.Bundle;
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/fragments/personalId/PersonalIdPhoneFragment.java
```diff
@@ -268,8 +267,9 @@ public void onTokenReceived(@NotNull String requestHash,
 
                     @Override
                     public void onTokenFailure(@NotNull Exception exception) {
-                        onConfigurationFailure(AnalyticsParamValue.START_CONFIGURATION_INTEGRITY_DEVICE_FAILURE,
-                                integrityTokenApiRequestHelper.getErrorForException(requireActivity(), exception));
+                        FirebaseAnalyticsUtil.reportPersonalIdConfigurationFailure(
+                                AnalyticsParamValue.START_CONFIGURATION_INTEGRITY_DEVICE_FAILURE);
```
shubham1g5, line 271: thoughts on removing this analytic now completely given we are not really failing the configuration flow anymore here and we are getting more wider reporting from the heartbeat anyway ? (Afraid that this will confuse anyone looking at those analytics)

PR #3332, opened 2025-09-10
Add perf monitoring trace for app sync duration

avazirna:
## Product Description
This PR implements a performance custom trace to measure app synchronization times. This is part of an effort to collect performance data related key events in CommCare.

Ticket: https://dimagi.atlassian.net/browse/SAAS-18093

### Safety story
This is supposed to be a lightweight, null-safe operation that has no impact on the user experience.

## Labels and Review

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -182,7 +182,8 @@
         <activity android:name="org.commcare.activities.MenuActivity" />
         <activity
             android:name="org.commcare.activities.connect.ConnectActivity"
-            android:exported="false"/>
+            android:exported="false"
+            android:screenOrientation="portrait"/>
         <service android:name="com.google.android.gms.metadata.ModuleDependencies"
             android:enabled="false"
             android:exported="false"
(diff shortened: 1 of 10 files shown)


app/src/org/commcare/tasks/DataPullTask.java
```diff
@@ -377,6 +382,10 @@ private ResultAndError<PullTaskResult> handleSuccessResponseCode(
             updateUserSyncToken(syncToken);
 
             onSuccessfulSync();
+            if (trace != null) {
+                trace.putAttribute(CCPerfMonitoring.ATTR_SYNC_SUCESS, AnalyticsParamValue.SYNC_SUCCESS);
```
shubham1g5, line 386: we don't want to log failures ?

shubham1g5 (changes requested):

app/src/org/commcare/tasks/DataPullTask.java
avazirna: @shubham1g5 the [description](https://docs.google.com/spreadsheets/d/1l5OiT5HCscpUW3h0xs9uPzxt9nHrfeaaArr0CuxxUVc/edit?gid=640482631#gid=640482631&range=C5) says `Duration of a full sync process with the server`, so my understanding is that this is to track complete sync durations. And I don't think there is any value in tracking the amount of time a failed sync takes.

shubham1g5 (approved):

OrangeAndGreen (approved):

PR #3335, opened 2025-09-10
Commcare 2.59.2 back-merge

OrangeAndGreen:
Back-merging the change from the 2.59.2 hotfix


app/AndroidManifest.xml
```diff
@@ -2,7 +2,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
           xmlns:tools="http://schemas.android.com/tools"
           android:versionCode="106"
-          android:versionName="2.60">
+          android:versionName="2.59.2">
```
shubham1g5, line 5: should be 2.60 in `master`

shubham1g5 (changes requested):

app/AndroidManifest.xml
OrangeAndGreen: Interesting, is this something that's missing from the hotfix scripts? I didn't see any error indicating that this step was attempted and failed. Either way, fixed here: [0835f55](https://github.com/dimagi/commcare-android/pull/3335/commits/0835f55abf2b114b35f720948108fa81635d4600)

avazirna: If I'm not mistaken, this happens when you merge `master` into the release branch before backmerging

PR #3348, opened 2025-09-18
Improved Connect DB error handling

OrangeAndGreen:
## Product Description
No visible user changes

## Technical Summary
Removed ambiguous logging about not finding DB passphrase (often false-alarms)
Removed redundant non-fatal exceptions just before DB crashes (which cause fatal exceptions)
Added a new check and crash to prevent unencrypted Connect DB
Added non-fatal exception when error encountered opening any DB (before trySqlCipherDbUpdate)
Improved exception logging when crashing Connect DB to indicate source

## Feature Flag
PersonalID

### Safety story
Should be no noticeable changes to user, but simply reduced/improved logging when errors occur.
The new crash before opening an unencrypted DB is the biggest risk and could cause users to crash on upgrade.
If so, they would simply be prompted to reconfigure PersonalID when restarting the app.

### Automated test coverage
None

### QA Plan
Basic regression testing around PersonalID configuration and de-configuration, ensure no errors

### DIFF
--- a/app/src/org/commcare/connect/database/ConnectDatabaseHelper.java
+++ b/app/src/org/commcare/connect/database/ConnectDatabaseHelper.java
@@ -73,6 +73,9 @@ public class ConnectDatabaseHelper {
                     if (connectDatabase == null || !connectDatabase.isOpen()) {
                         try {
                             byte[] passphrase = ConnectDatabaseUtils.getConnectDbPassphrase(context, true);
+                            if(passphrase == null) {
+                                throw new IllegalStateException("Attempting to access Connect DB without a passphrase");
+                            }
 
                             String remotePassphrase = ConnectDatabaseUtils.getConnectDbEncodedPassphrase(context, false);
                             String localPassphrase = ConnectDatabaseUtils.getConnectDbEncodedPassphrase(context, true);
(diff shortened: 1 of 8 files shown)


app/src/org/commcare/models/database/app/DatabaseAppOpenHelper.java
```diff
@@ -44,6 +45,7 @@ public SQLiteDatabase getWritableDatabase() {
         try {
             return super.getWritableDatabase();
         } catch (SQLiteException sqle) {
+            Logger.exception("Opening database failed", sqle);
```
shubham1g5, line 48: nit: it might be useful to know what database is failing here from the messag like `"Opening App database failed"`

app/src/org/commcare/connect/database/ConnectDatabaseHelper.java
```diff
@@ -52,8 +52,7 @@ public static void handleReceivedDbPassphrase(Context context, String remotePass
             //Store the received passphrase as what's in use locally
             ConnectDatabaseUtils.storeConnectDbPassphrase(context, remotePassphrase, true);
         } catch (Exception e) {
-            Logger.exception("Handling received DB passphrase", e);
```
shubham1g5, line 55: curious why are not we logging the underlying exceptions now for `crashDb` calls.

app/src/org/commcare/models/database/app/DatabaseAppOpenHelper.java
OrangeAndGreen: I think we'll be able to see that pretty easily from the stack trace, no? Like whether it comes from DatabaseAppOpenHelper, DatabaseConnectOpenHelper, etc.

app/src/org/commcare/connect/database/ConnectDatabaseHelper.java
OrangeAndGreen: Ah, interesting. I removed them because they were essentially just duplicates of the fatal exception that gets logged immediately after when the app crashes, but didn't think about the lost crash-causing exception. I'll pass it through so the crashing exception can wrap it and we can preserve the full stack while avoiding duplicate events.

app/src/org/commcare/models/database/app/DatabaseAppOpenHelper.java
shubham1g5: I am concerned that Crashlytics will club them together into one single issue or not as it might be valuable to know if this issue is happening for all Dbs or just one by getting the issue count. Although not sure whats the behaviour here on Crashlytics really, so fine to ignore it atm

PR #3352, opened 2025-09-30
Logs all network requests to Analytics

shubham1g5:
## Product Description

Logging request urls and corresponding response codes for all requests mobile is making

## Labels and Review

- [] Does the PR introduce any major changes worth communicating ? If yes, the "Release Note" label is set and a "Release Note" is specified in PR description.


app/src/org/commcare/network/NetworkAnalyticsInterceptor.kt
```diff
@@ -0,0 +1,40 @@
...
+
+import okhttp3.Interceptor
+import okhttp3.Response
+import org.commcare.google.services.analytics.FirebaseAnalyticsUtil
+import org.javarosa.core.services.Logger
+
+/**
+ * OkHttp interceptor that logs analytics events for all HTTP requests/responses
+ */
+class NetworkAnalyticsInterceptor : Interceptor {
```
Jignesh-dimagi, line 11: @shubham1g5 We are trying to log all network calls, which might be available in AWS EC2 instance with corresponding delays. These might be too much data to handle and might not be useful at the end?

app/src/org/commcare/google/services/analytics/CCAnalyticsParam.java
```diff
@@ -40,4 +40,8 @@ public class CCAnalyticsParam {
     static final String PARAM_API_SUCCESS = "ccc_api_success";
     static final String PARAM_API_TOTAL_JOBS = "ccc_api_total_jobs";
     static final String PARAM_API_NEW_JOBS = "ccc_api_new_jobs";
+
+    static final String RESPONSE_CODE = "response_code";
+    static final String REQUEST_DURATION_MS = "request_duration_ms";
+    static final String URL = "url";
```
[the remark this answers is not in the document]
Jignesh-dimagi: +1, might need to remove PII or token

app/src/org/commcare/google/services/analytics/FirebaseAnalyticsUtil.java
```diff
@@ -586,4 +584,12 @@ public static void reportNavDrawerItemSelected(String selectedItem) {
                 new String[]{selectedItem});
     }
 
+    public static void reportNetworkRequest(String url, int responseCode, long durationMs) {
+        Bundle bundle = new Bundle();
+        bundle.putInt(CCAnalyticsParam.RESPONSE_CODE, responseCode);
+        bundle.putLong(CCAnalyticsParam.REQUEST_DURATION_MS, durationMs);
+        bundle.putString(CCAnalyticsParam.URL, url);
+        reportEvent(CCAnalyticsEvent.NETWORK_REQUEST, bundle);
+    }
```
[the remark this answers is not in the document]
Jignesh-dimagi: +1

app/src/org/commcare/network/NetworkAnalyticsInterceptor.kt
shubham1g5: I think we are making a shift to log more and more data to analytics (now that we don't have sumologic) , so I would not be worried about this data size untill it starts creating practical issues for us. 

I am also not sure how will we filter what requests we want to log and what not.

app/src/org/commcare/google/services/analytics/CCAnalyticsEvent.java
```diff
@@ -55,5 +55,6 @@ public class CCAnalyticsEvent {
     static final String PERSONAL_ID_CONFIGURATION_FAILURE = "personal_id_configuration_failure";
     static final String NAV_DRAWER_OPEN = "nav_drawer_open";
     static final String NAV_DRAWER_ITEM_SELECTED = "nav_drawer_item_selected";
+    static final String NETWORK_REQUEST = "network_request";
```
OrangeAndGreen, line 58: Nice addition

app/src/org/commcare/google/services/analytics/CCAnalyticsParam.java
```diff
@@ -40,4 +40,8 @@ public class CCAnalyticsParam {
     static final String PARAM_API_SUCCESS = "ccc_api_success";
     static final String PARAM_API_TOTAL_JOBS = "ccc_api_total_jobs";
     static final String PARAM_API_NEW_JOBS = "ccc_api_new_jobs";
```
OrangeAndGreen, line 42: Noting the overlap with the ccc_api_* events. These can log additional info in some cases though, such as number of new jobs received. I think fine to leave them all in for now, but I wonder if there's a future world we could get to where network_request events could be the single event and optionally include additional analytics info depending on the call? Recognizing this may be difficult with the more-generic NetworkAnalyticsInterceptor...

app/src/org/commcare/network/NetworkAnalyticsInterceptor.kt
```diff
@@ -0,0 +1,40 @@
...
+        var response: Response?
+        var responseCode = -1
+        
+        try {
+            response = chain.proceed(request)
+            responseCode = response.code
+            return response
+        } finally {
+            val duration = System.currentTimeMillis() - startTime
+            logNetworkAnalytics(request.url.toString(), responseCode, duration)
```
OrangeAndGreen, line 25: Noting that any exceptions sending the request will result in a -1 responseCode with no additional exception info. I wonder if we should attempt to distinguish the various possible exceptions with different negative error codes?

app/src/org/commcare/network/NetworkAnalyticsInterceptor.kt
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    private fun logNetworkAnalytics(
+        url: String,
+        responseCode: Int, 
+        duration: Long,
+    ) {
+        try {
+            FirebaseAnalyticsUtil.reportNetworkRequest(url, responseCode, duration)
+        } catch (e: Exception) {
```
OrangeAndGreen, line 36: +1 to adding safety to our analytics calls and silently reporting exceptions, I think we've seen a crash or two come from such calls in the past. But I think we should make it comprehensive, rather than the caller's responsibility as done here.

Looking in FirebaseAnalyticsUtil it seems we may be close to accomplishing this already.
* reportEvent(String,String[],String[]) is wrapped in a try-catch
* reportEvent(String,Bundle) is not

The first calls the second, but there are 16 additional usages of the second. Seems a little cleanup here would globally guarantee safety for analytics calls.

OrangeAndGreen (approved):
Left a few ideas but happy to save any/all for future work so we can get this in to 2.60

app/src/org/commcare/network/NetworkAnalyticsInterceptor.kt
shubham1g5: thanks for I can get that in as part of this work.

PR #3354, opened 2025-10-01
Cleaned up local DB passphrase code

OrangeAndGreen:
https://dimagi.atlassian.net/browse/CCCT-1574

## Product Description
No user-facing changes

## Technical Summary
This PR cleans up the migration code that was introduced earlier this year when we changed the Connect DB to start using a passphrase managed by PersonalID for encryption, instead of a locally-generated passphrase.

Brief description of the migration flow:
-Added an isLocal flag to ConnectKeyRecord
-All existing passphrases became "local" passphrases (isLocal=true) during the upgrade
-At app startup, we would issue a special API call if we detected the user didn't have the remote passphrase yet (isLocal=false)
-Once the remote passphrase was received, we would:
   -Store it as remote
   -Rekey the database if necessary (local != remote)
   -Store it as local

One additional caveat:
During the change described above we realized Connect was using the wrong overload for the DatabaseConnectOpenHelper constructor, and passing in a byte[] instead of an encoded string. So while supporting the upgrade/migration, we kept the ability for the app to keep accessing the DB via the legacy method until the migration was complete. The code supporting the legacy byte[] method has also been removed in this PR.
-Connect legacy used Base64.encodeToString
-Common CC code uses UserSandboxUtils.getSqlCipherEncodedKey

Changes to the code:
-Removed obsolete code for retrieving remote passphrase and re-keying DB.
-Removed legacy support for DBs using the old passphrase system (with different handling when getting DB handle).
-Deprecated the isLocal flag in ConnectKeyRecord

## Feature Flag
PersonalID

### Safety story
Dev tested:
1. Start with v2.60 and PersonalID configured, then upgrade to the latest and verify things work as expected
2. Restart the app after the above and verify again
3. Install the latest from scratch, then configure PersonalID and verify things work as expected
4. Restart the app after the above and verify again

For all 4 tests, "things work as expected" simply means that PersonalID is functional. This can be tested by:
-Viewing the side nav menu and verifying the PersonalID user info is displayed
-Account does not automatically get de-configured
-No crashes

### Automated test coverage
None

### QA Plan
See the dev tests conducted above with success criteria

### DIFF
--- a/app/src/org/commcare/android/database/global/models/ConnectKeyRecord.java
+++ b/app/src/org/commcare/android/database/global/models/ConnectKeyRecord.java
@@ -20,7 +20,7 @@ public class ConnectKeyRecord extends Persisted {
     @Deprecated
     @Persisting(2)
     @MetaField(IS_LOCAL)
-    boolean isLocal = true;
+    boolean isLocal;
 
     public ConnectKeyRecord() {
     }
--- a/app/src/org/commcare/android/database/global/models/ConnectKeyRecord.java
+++ b/app/src/org/commcare/android/database/global/models/ConnectKeyRecord.java
@@ -5,8 +5,6 @@ import org.commcare.models.framework.Persisting;
 import org.commcare.modern.database.Table;
 import org.commcare.modern.models.MetaField;
 
-import kotlin.Metadata;
-
 /**
  * DB model for storing the encrypted/encoded Connect DB passphrase
  *
@@ -19,16 +17,16 @@ public class ConnectKeyRecord extends Persisted {
     @Persisting(1)
     String encryptedPassphrase;
 
+    @Deprecated
     @Persisting(2)
     @MetaField(IS_LOCAL)
-    boolean isLocal;
+    boolean isLocal = true;
 
     public ConnectKeyRecord() {
     }
 
-    public ConnectKeyRecord(String encryptedPassphrase, boolean isLocal) {
+    public ConnectKeyRecord(String encryptedPassphrase) {
         this.encryptedPassphrase = encryptedPassphrase;
-        this.isLocal = isLocal;
     }
 
     public String getEncryptedPassphrase() {
@@ -37,11 +35,8 @@ public class ConnectKeyRecord extends Persisted {
     public void setEncryptedPassphrase(String passphrase) {
         encryptedPassphrase = passphrase;
     }
-    public boolean getIsLocal() {
-        return isLocal;
-    }
 
     public static ConnectKeyRecord fromV6(ConnectKeyRecordV6 oldVersion) {
-        return new ConnectKeyRecord(oldVersion.getEncryptedPassphrase(), true);
+        return new ConnectKeyRecord(oldVersion.getEncryptedPassphrase());
     }
 }
--- a/app/src/org/commcare/connect/database/ConnectDatabaseHelper.java
+++ b/app/src/org/commcare/connect/database/ConnectDatabaseHelper.java
@@ -1,18 +1,11 @@
 package org.commcare.connect.database;
 
 import android.content.Context;
-import android.os.Handler;
-import android.os.Looper;
-import android.widget.Toast;
-
 
 import org.commcare.android.database.connect.models.ConnectLinkedAppRecord;
 import org.commcare.android.database.connect.models.ConnectUserRecord;
-import org.commcare.android.database.global.models.ConnectKeyRecord;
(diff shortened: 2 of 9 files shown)


app/src/org/commcare/android/database/global/models/ConnectKeyRecord.java
```diff
@@ -19,16 +17,16 @@ public class ConnectKeyRecord extends Persisted {
     @Persisting(1)
     String encryptedPassphrase;
 
+    @Deprecated
     @Persisting(2)
     @MetaField(IS_LOCAL)
-    boolean isLocal;
+    boolean isLocal = true;
```
shubham1g5, line 23: Thoughts on setting it to `false` given there is no local passphrase anymore ? (everything comes from server)

app/src/org/commcare/connect/database/ConnectDatabaseUtils.java
```diff
@@ -42,39 +40,26 @@ record = new ConnectKeyRecord(encoded, isLocal);
         }
     }
 
-    public static ConnectKeyRecord getKeyRecord(boolean local) {
+    public static ConnectKeyRecord getKeyRecord() {
         Vector<ConnectKeyRecord> records = CommCareApplication.instance()
                 .getGlobalStorage(ConnectKeyRecord.class)
-                .getRecordsForValue(ConnectKeyRecord.IS_LOCAL, local);
+                .getRecordsForValue(ConnectKeyRecord.IS_LOCAL, true);
```
shubham1g5, line 46: can we not query for local field anymore and just get the first entry in this table ?

app/src/org/commcare/connect/database/ConnectDatabaseHelper.java
```diff
@@ -39,21 +30,8 @@ public class ConnectDatabaseHelper {
     public static IDatabase connectDatabase;
     static boolean dbBroken = false;
 
-    public static void handleReceivedDbPassphrase(Context context, String remotePassphrase) {
-        ConnectDatabaseUtils.storeConnectDbPassphrase(context, remotePassphrase, false);
-        try {
-            //Rekey the DB if the remote passphrase is different than local
-            String localPassphrase = ConnectDatabaseUtils.getConnectDbEncodedPassphrase(context, true);
-            if (connectDatabase != null && connectDatabase.isOpen() && !remotePassphrase.equals(localPassphrase)) {
-                DatabaseConnectOpenHelper.rekeyDB(connectDatabase, remotePassphrase);
```
shubham1g5, line 48: can we also clear out the `rekeyDB` from `DatabaseConnectOpenHelper` or is it still used elsewhere ?

app/src/org/commcare/android/database/global/models/ConnectKeyRecord.java
OrangeAndGreen: Ah, good idea. I had in my head that we needed to keep checking for the local passphrase to support users upgrading with a working database, but I tested it with isLocal defaulted to false and the query simply returning the first item and it works fine as well. [34bbc35](https://github.com/dimagi/commcare-android/pull/3354/commits/34bbc354efe9705f04d3328031df5da8435a45b8)

app/src/org/commcare/connect/database/ConnectDatabaseUtils.java
OrangeAndGreen: Part of this change: [34bbc35](https://github.com/dimagi/commcare-android/pull/3354/commits/34bbc354efe9705f04d3328031df5da8435a45b8)

app/src/org/commcare/connect/database/ConnectDatabaseUtils.java
```diff
@@ -42,39 +38,28 @@ record = new ConnectKeyRecord(encoded, isLocal);
...
-                .getGlobalStorage(ConnectKeyRecord.class)
-                .getRecordsForValue(ConnectKeyRecord.IS_LOCAL, local);
+    public static ConnectKeyRecord getKeyRecord() {
+        Iterable<ConnectKeyRecord> records = CommCareApplication.instance()
+                .getGlobalStorage(ConnectKeyRecord.class);
 
-        return records.size() > 0 ? records.firstElement() : null;
+        if (records.iterator().hasNext()) {
+            return records.iterator().next();
+        }
```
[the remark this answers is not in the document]
OrangeAndGreen: I worried the same but tested and the false case is working fine through an app upgrade

PR #3359, opened 2025-10-07
Fixes crash on clicking View Job Status

shubham1g5:
## Product Description

https://dimagi.atlassian.net/browse/QA-8141

[crash](https://console.firebase.google.com/u/1/project/commcare-a57e4/crashlytics/app/android:org.commcare.dalvik/issues/2453761d848f4bb7f52a49ae1554ea26?time=60m&sessionEventKey=68E4E50E033C000132D6FCAEBF1E4D76_2136864156345115017)

## Technical Summary

We were using incorrect data type for opportunity id which was resulting in the job to be set as null for `connectActivity` 

[This is where](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/connect/ConnectNavHelper.kt#L76) we add opportunity id to the extras as Integer

## Labels and Review

### DIFF
--- a/app/src/org/commcare/activities/connect/ConnectActivity.java
+++ b/app/src/org/commcare/activities/connect/ConnectActivity.java
@@ -81,9 +81,9 @@ public class ConnectActivity extends NavigationHostCommCareActivity<ConnectActiv
 
     private void initStateFromExtras() {
         redirectionAction = getIntent().getStringExtra(REDIRECT_ACTION);
-        String opportunityId = getIntent().getStringExtra(ConnectConstants.OPPORTUNITY_ID);
-        if(!TextUtils.isEmpty(opportunityId)){
-            job = ConnectJobUtils.getCompositeJob(this, Integer.parseInt(opportunityId));
+        int opportunityId = getIntent().getIntExtra(ConnectConstants.OPPORTUNITY_ID, -1);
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/activities/connect/ConnectActivity.java
```diff
@@ -81,9 +81,9 @@ private int getStartDestinationId(Bundle startArgs) {
 
     private void initStateFromExtras() {
         redirectionAction = getIntent().getStringExtra(REDIRECT_ACTION);
-        String opportunityId = getIntent().getStringExtra(ConnectConstants.OPPORTUNITY_ID);
-        if(!TextUtils.isEmpty(opportunityId)){
-            job = ConnectJobUtils.getCompositeJob(this, Integer.parseInt(opportunityId));
```
Jignesh-dimagi, line 86: @shubham1g5 Just trying to understand here, what is the reason that it was failing? code looks ok

shubham1g5: offline with Jignesh here and seems like we are sending it as String at a bunch of places in Push notifications code, so going to go with dual parsing here as both String and Int

Jignesh-dimagi (approved):

PR #3369, opened 2025-10-16
Fixed biometric config options when fingerprint not available

OrangeAndGreen:
https://dimagi.atlassian.net/browse/QA-8145

## Product Description
Fingerprint option will never be displayed on the Biometric Config page if the device doesn't have a fingerprint scanner.

## Technical Summary
Rearranged logic for showing fingerprint/PIN option on biometric config page.
Fixed issue where fingerprint option could be shown when device doesn't have a fingerprint scanner.

## Feature Flag
PersonalID

### Safety story
Tested manually by dev

### Automated test coverage
None

### QA Plan
Test the Biometric Config page under various scenarios of fingerprint/PIN being available and/or configured and Connect vs. non-Connect user. Make sure the appropriate options are shown to the user in every case.

### DIFF
--- a/app/src/org/commcare/fragments/personalId/PersonalIdBiometricConfigFragment.java
+++ b/app/src/org/commcare/fragments/personalId/PersonalIdBiometricConfigFragment.java
@@ -130,7 +130,7 @@ public class PersonalIdBiometricConfigFragment extends BasePersonalIdFragment {
         boolean hasFingerprintHardware = fingerprintStatus != BiometricsHelper.ConfigurationStatus.NoHardware;
 
         String title;
-        String message;
+        String message = "";
         String fingerprintButton = null;
         String pinButton = null;
 


app/src/org/commcare/fragments/personalId/PersonalIdBiometricConfigFragment.java
```diff
@@ -131,47 +131,43 @@ private void updateUiBasedOnMinSecurityRequired() {
...
-        } else if (fingerprintStatus == BiometricsHelper.ConfigurationStatus.Configured) {    // Fingerprint is configured so works for both PIN and BIOMETRIC_TYPE
-            title = getString(R.string.connect_verify_use_fingerprint_long);
+        if (fingerprintStatus == BiometricsHelper.ConfigurationStatus.Configured) {    // Fingerprint is configured so works for both PIN and BIOMETRIC_TYPE
             message = getString(R.string.connect_verify_fingerprint_configured);
             fingerprintButton = getString(R.string.connect_verify_agree);
-        } else if (BIOMETRIC_TYPE.equals(personalIdSessionDataViewModel.getPersonalIdSessionData().getRequiredLock())) {   //Fingerprint not configured but required for BIOMETRIC_TYPE
-            // Need at least Fingerprint configuration for BIOMETRIC_TYPE
+        } else {
+            message = getString(R.string.connect_verify_message);
             if(hasFingerprintHardware) {
```
Jignesh-dimagi: `hasFingerprintHardware` should check for other scenario also i.e. for `NotAvailable` and `NeedsUpdate`?

shubham1g5 (changes requested):

shubham1g5:
@OrangeAndGreen seems like build is not compiling here.

app/src/org/commcare/fragments/personalId/PersonalIdBiometricConfigFragment.java
shubham1g5: @Jignesh-dimagi Do you mean user gets blocked on using pin if fingerprint errors while doing a setup step ? If so, I agree that's not the behaviour we would want.

Jignesh-dimagi: @shubham1g5 Yes that's correct

PR #3370, opened 2025-10-17
Improve logging around resizing in QuestionsView

avazirna:
## Product Description
This is to improve visibility around [this](https://console.firebase.google.com/project/commcare-a57e4/crashlytics/app/android:org.commcare.dalvik/issues/1ceae08e29560910c0e182b24852e881?time=30d&types=crash&sessionEventKey=68F1FC52035D000159E86D9473069443_2140605356365267502) issue. It seems that an issue occurs during layout updates when returning to CommCare after launching a third-party app. 

## Labels and Review

### DIFF
--- a/app/res/values-es/strings.xml
+++ b/app/res/values-es/strings.xml
@@ -209,7 +209,9 @@
     <string name="connect_register_success_message">¡Tu cuenta de PersonalId se ha creado con éxito y ahora está lista para usar!</string>
     <string name="connect_register_success_button">DE ACUERDO</string>
     <string name="connect_verify_title">Opciones de desbloqueo</string>
-    <string name="connect_verify_message">Configura un método a continuación para desbloquear tanto tu dispositivo como tu cuenta de PersonalId</string>
+    <string name="connect_verify_message">Configura un método a continuación para desbloquear tanto tu dispositivo como tu cuenta de PersonalID</string>
+    <string name="connect_verify_message_pin">Configure un PIN o patrón a continuación para desbloquear su dispositivo y su cuenta PersonalID</string>
+    <string name="connect_verify_message_fingerprint">Configure su huella digital a continuación para desbloquear tanto su dispositivo como su cuenta PersonalID</string>
     <string name="connect_verify_use_fingerprint_long">Usar huella digital para iniciar sesión</string>
     <string name="connect_verify_fingerprint_configured">Tu huella digital ya se ha configurado y se utilizará para desbloquear PersonalId.</string>
     <string name="connect_verify_configure_fingerprint">Configurar huella digital</string>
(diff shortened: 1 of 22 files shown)


avazirna:

app/src/org/commcare/views/QuestionsView.java
```diff
@@ -347,7 +347,15 @@ public void setFocus(Context context, int indexOfLastChangedWidget) {
...
     private void scrollToWidget(final QuestionWidget widget) {
-        new Handler().post(() -> QuestionsView.this.scrollTo(0, widget.getTop()));
+        new Handler().post(() -> {
+            String widgetTextId = "None";
+            try {
+                widgetTextId = widget.getPrompt().getQuestion().getTextID();
+            } catch (Exception ignore) {}
+            Logger.log(LogTypes.SOFT_ASSERT, "Attempting to scroll to widget: " + widgetTextId);
+
+            QuestionsView.this.scrollTo(0, widget.getTop());
```
shubham1g5: @avazirna you can instead catch the exception in this call and rethrow with the additional details, that way we will only log this in case of errrors, otherwise I am afraid this is a rather high frequency log that we will want to avoid logging.

app/src/org/commcare/views/QuestionsView.java
```diff
@@ -346,6 +346,17 @@ public void setFocus(Context context, int indexOfLastChangedWidget) {
         }
     }
 
+    @Override
+    protected void onSizeChanged(int w, int h, int oldw, int oldh) {
+        try {
+            super.onSizeChanged(w, h, oldw, oldh);
+        } catch (IllegalArgumentException e) {
+            Logger.log(LogTypes.SOFT_ASSERT,
+                    "Resizing from "+oldw+"X"+oldh+" to "+w+"X"+h+" failed with focus on: " + findFocus());
```
shubham1g5: what will findFocus equate to as String ?

app/src/org/commcare/views/QuestionsView.java
avazirna: @shubham1g5 I haven't been able to reproduce, but one common factor between the forms where this issue is happening is the app callout question, so I'm trying to assess whether it's related. Within QuestionsView there is the `restoreFocusToQuestionThatCalledOut()` which triggers the logic to `scrollToWidget()`, so the previous log was to assess whether this is the cause and what's the widget causing the issue.

app/src/org/commcare/views/QuestionsView.java
```diff
@@ -346,6 +348,30 @@ public void setFocus(Context context, int indexOfLastChangedWidget) {
         }
     }
 
+    @Override
+    protected void onSizeChanged(int w, int h, int oldw, int oldh) {
+        try {
+            super.onSizeChanged(w, h, oldw, oldh);
+        } catch (IllegalArgumentException e) {
+            Logger.log(LogTypes.SOFT_ASSERT,
+                    "Resizing from " + oldw + "X" + oldh + " to " + w + "X" + h + " failed with focus on: " + getFocusedViewClassName());
```
Jignesh-dimagi: Log exception as it will gives exception details also?

PR #3382, opened 2025-10-27
Null-check when logging integrity analytics event

OrangeAndGreen:
https://dimagi.atlassian.net/browse/CCCT-1808

## Product Description
No user-facing change

## Technical Summary
Logging a non-fatal exception when trying to send integrity analytics event with a null (or "null") responseCode.

### Safety story
Simple change to log a non-fatal exception, not user-facing.

### Automated test coverage
None

### DIFF
--- a/app/src/org/commcare/google/services/analytics/FirebaseAnalyticsUtil.java
+++ b/app/src/org/commcare/google/services/analytics/FirebaseAnalyticsUtil.java
@@ -18,7 +18,8 @@ import org.commcare.utils.FormUploadResult;
 import org.javarosa.core.services.Logger;
 
 import java.util.Date;
-import java.util.Objects;
+
+import javax.annotation.Nullable;
 
 import androidx.navigation.NavController;
 import androidx.navigation.NavDestination;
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/google/services/analytics/FirebaseAnalyticsUtil.java
```diff
@@ -431,6 +431,10 @@ public static void reportPersonalIdHeartbeatIntegritySubmission(String requestId
     }
 
     public static void reportPersonalIdIntegritySubmission(String trigger, String requestId, String responseCode) {
+        if(responseCode == null || responseCode.equals("null")) {
+            Logger.exception("Integrity Check", new Exception("Null response code in integrity check: " + responseCode));
+        }
```
shubham1g5: Curious what additional info this will give us other than what we know already from anlaytics ? Think a log here should be able to determine from which specific pathway the null is coming from amongst the ones you examined in [this comment](https://dimagi.atlassian.net/browse/CCCT-1808?focusedCommentId=411945)

shubham1g5: I am a bit doubtful this exception will have a stacktrace attached to it by itself.

OrangeAndGreen: Thinking it should look like [this Firebase non-fatal](https://console.firebase.google.com/u/0/project/commcare-a57e4/crashlytics/app/android:org.commcare.dalvik/issues/3f0b668151a03b543813a1563dd4491d?time=7d&types=error&sessionEventKey=690230C80023000114CE8D38AD4B860C_2145110781515137080), corresponding to [this line of code](https://github.com/dimagi/commcare-android/blob/50bf05b7d753a5c8bc4f0d910f6a1d828594174f/app/src/org/commcare/network/CommcareRequestGenerator.java#L203).

Even though Logger.exception is called in CommcareRequestGenerator.checkForTokenError, we can see in the stack trace that the code came from i.e. LogSubmissionTask.submitReports

shubham1g5: nice that makes sense and thanks for double checking.

PR #3390, opened 2025-10-30
Fix work history click on Side Nav

shubham1g5:
## Product Description

https://dimagi.atlassian.net/browse/QA-8191

## Technical Summary

Adds click handling for work history

## Labels and Review

### DIFF
--- a/app/src/org/commcare/activities/StandardHomeActivity.java
+++ b/app/src/org/commcare/activities/StandardHomeActivity.java
@@ -280,7 +280,7 @@ public class StandardHomeActivity
                     ConnectNavHelper.INSTANCE.goToWorkHistory(this);
                     closeDrawer();
                 } else {
-                    navigateToCredential();
+                    navigateToWorkHistory();
                 }
             }
         }
--- a/app/src/org/commcare/navdrawer/BaseDrawerActivity.kt
+++ b/app/src/org/commcare/navdrawer/BaseDrawerActivity.kt
@@ -86,7 +86,7 @@ abstract class BaseDrawerActivity<T> : CommCareActivity<T>() {
             NavItemType.COMMCARE_APPS -> { /* No nav, expands/collapses menu */ }
             NavItemType.PAYMENTS -> {}
             NavItemType.MESSAGING -> { navigateToMessaging() }
-            NavItemType.WORK_HISTORY -> { navigateToCredential() }
+            NavItemType.WORK_HISTORY -> { navigateToWorkHistory() }
         }
     }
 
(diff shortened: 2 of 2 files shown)


app/src/org/commcare/activities/StandardHomeActivity.java
```diff
@@ -275,6 +275,14 @@ protected void handleDrawerItemClick(@NonNull BaseDrawerController.NavItemType i
...
                 }
             }
+            case WORK_HISTORY -> {
+                if(personalIdManagedLogin) {
+                    ConnectNavHelper.INSTANCE.goToWorkHistory(this);
+                    closeDrawer();
+                } else {
+                    navigateToCredential();
+                }
+            }
```
Jignesh-dimagi: @shubham1g5 BaseDrawerActivity is already handling this click, do we need this define again here? Also, workHistory is the final thing and we removing credentials ?

app/src/org/commcare/navdrawer/BaseDrawerActivity.kt
```diff
@@ -126,7 +125,7 @@ abstract class BaseDrawerActivity<T> : CommCareActivity<T>() {
     }
 
     protected fun navigateToCredential() {
```
Jignesh-dimagi: Should it be `navigateToWorkHistory`?

Jignesh-dimagi (changes requested):

app/src/org/commcare/activities/StandardHomeActivity.java
shubham1g5: > BaseDrawerActivity is already handling this click, do we need this define again here?

there is an additional check on this page which negates the need of unlocking Personal ID account given user is already logged into that app by using Personal ID unlock

> Also, workHistory is the final thing and we removing credentials ?

Yes.

shubham1g5: user can login into traditional CC apps with means of Personal ID OR with a username - password combo specific to that app. The check here says that if user logged into the app with Personal ID linking, don't ask them to unlock but otherwise go with the normal flow. Think we can only handle the special Personal ID linking scenario here and delegate the else part to super. Thoughts ?

Jignesh-dimagi: Yeah, I was thinking same but when seen this file, it was using what you have coded now but yeah better to delegate to super.

conroy-ricketts (approved):

OrangeAndGreen (approved):

app/src/org/commcare/activities/StandardHomeActivity.java
shubham1g5: @Jignesh-dimagi Think I am inclined to keep this as it is for now as I think we will need to change this code anyway with a different approach on unlocking account (based on yesterday's conversation with Product on removing these unlocks)

PR #3392, opened 2025-10-31
CCCT-1720 App Sidebar Usability Improvement

conroy-ricketts:
## Product Description

I added some logic to open the app's sidebar after a successful Personal ID registration.

https://github.com/user-attachments/assets/dc51a42a-5080-443c-bbde-f81aadc0bfa6

## Technical Summary

This one was pretty straightforward.

### Safety story

I verified that the app's sidebar automatically opens after successful Personal ID registration (for both recovery and a brand new user).

### QA Plan

Verify that the app's sidebar automatically opens after successful Personal ID registration. We should verify this for both account recovery **_and_** brand new user registration.

### DIFF
--- a/app/src/org/commcare/navdrawer/BaseDrawerActivity.kt
+++ b/app/src/org/commcare/navdrawer/BaseDrawerActivity.kt
@@ -16,6 +16,7 @@ import org.commcare.connect.ConnectNavHelper.unlockAndGoToWorkHistory
 import org.commcare.navdrawer.BaseDrawerController.NavItemType
 import org.commcare.pn.helper.NotificationBroadcastHelper
 import org.commcare.utils.FirebaseMessagingUtil
+import org.javarosa.core.services.Logger
 
 abstract class BaseDrawerActivity<T> : CommCareActivity<T>() {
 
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/connect/PersonalIdManager.java
```diff
@@ -200,6 +201,10 @@ public void completeSignin() {
         scheduleHeartbeat();
         NotificationsSyncWorkerManager.schedulePeriodicPushNotificationRetrieval(CommCareApplication.instance());
         CrashUtil.registerUserData();
+
+        if (parentActivity instanceof LoginActivity) {
```
OrangeAndGreen: Thinking we should check for BaseDrawerActivity here, which will cause the drawer to open anytime the parent activity hosts the drawer. Specifically, this will make it so the drawer opens if the user configures PersonalID from the CommCareSetupActivity

OrangeAndGreen (changes requested):

app/src/org/commcare/navdrawer/BaseDrawerActivity.kt
```diff
@@ -140,4 +140,8 @@ abstract class BaseDrawerActivity<T> : CommCareActivity<T>() {
     protected fun closeDrawer() {
         drawerController?.closeDrawer()
     }
+
+    fun openDrawer() {
+        drawerController?.openDrawer()
```
shubham1g5: can we soft assert using Logger.exception that drawerController is non null here. Just so that if we start calling this before `drawerController` is initialized, we should get to know about it from logs.

shubham1g5 (approved):

OrangeAndGreen (approved):

PR #3401, opened 2025-11-05
Fix app sandbox setup

shubham1g5:
## Product Description

Fixes https://dimagi.atlassian.net/browse/QA-8210

## Technical Summary

We were not restoring the last app sandbox state correctly leading to [this crash](https://console.firebase.google.com/u/0/project/commcare-a57e4/crashlytics/app/android:org.commcare.dalvik/issues/51aa7f6f5eb33e037c6fcb406e79cad7?time=12h&versions=2.61%20(477622);2.61%20(484255)&sessionEventKey=6909D11D036E000153B2B4A291C94671_2147282452912389058)

I added more sandbox handling to reset to the initial state and could verify the issue is fixed. 

- I also have put this code behind the pending flags check as this evaluation can be expensive and doesn't need to happen at the momenet when pending tabs are not turned on

## Labels and Review

### DIFF
--- a/app/src/org/commcare/activities/connect/viewmodel/PersonalIdWorkHistoryViewModel.kt
+++ b/app/src/org/commcare/activities/connect/viewmodel/PersonalIdWorkHistoryViewModel.kt
@@ -98,8 +98,8 @@ class PersonalIdWorkHistoryViewModel(
 
     private fun getWorkHistoryForAppRecord(record: ApplicationRecord): Iterable<PersonalIdWorkHistory> {
         val commcareApp = CommCareApp(record)
+        commcareApp.setupSandbox()
         try {
-            commcareApp.setupSandbox()
             val profile = commcareApp.initApplicationProfile()
             return profile.credentials.map { credential ->
                 PersonalIdWorkHistory().apply {


conroy-ricketts (commented):
LGTM - I don't have any concerns. I'd like another set of eyes though

OrangeAndGreen (dismissed):
I think this all looks good, but just making sure I understand:
The issue is that we need to load the sandbox for each app that we have locally in order to check for pending work history, so when we're done we need to make sure to put the previously seated app back. Is that correct?

I'm curious as to why this would cause a crash specifically when trying to download/install a new app (as described by QA)?

shubham1g5:
> The issue is that we need to load the sandbox for each app that we have locally in order to check for pending work history, so when we're done we need to make sure to put the previously seated app back. Is that correct?

Yes.


> I'm curious as to why this would cause a crash specifically when trying to download/install a new app (as described by QA)?

This is the [related stack trace for that](https://console.firebase.google.com/u/0/project/commcare-a57e4/crashlytics/app/android:org.commcare.dalvik/issues/fe7b63726e7f269e3ac358491c8be318), if we don't tear the sandbox and call setup on the same sandbox, it errors with this.

PR #3403, opened 2025-11-05
Fix textSizeSetting instrumentation test

avazirna:
## Product Description
Ticket: https://dimagi.atlassian.net/browse/SAAS-18098

## Labels and Review


avazirna:
@shubham1g5 Here are the changes to fix to the instrumentation test that is failing on the Samsung Galaxy Note9 device. I have only applied these to the test itself, but curious if we think we should also apply this to CommCare where these font values are used. To my knowledge we haven't received any feedback around this. The particularity with this device is that the default font scale seems to be `1.1`, which the others is `1`. In terms of the build time, I'm still monitoring, a [recent build](https://jenkins.dimagi.com/job/commcare-android-pr-job-builder/477711/) took 47min. which seems reasonable.

app/instrumentation-tests/src/org/commcare/utils/CustomMatchers.kt
```diff
@@ -116,7 +117,11 @@ object CustomMatchers {
                     return false
                 }
                 val pixels = target.textSize
-                val actualSize = pixels / target.getResources().displayMetrics.scaledDensity
+
+                val actualSize = (pixels / target.resources.displayMetrics.scaledDensity).roundToInt()
+                Logger.log(CustomMatchers::class.simpleName + ".withFontSize()", "View text size: $pixels")
+                Logger.log(CustomMatchers::class.simpleName + ".withFontSize()", "Actual size: $actualSize")
+                Logger.log(CustomMatchers::class.simpleName + ".withFontSize()", "Font scale: " + target.resources.configuration.fontScale)
```
shubham1g5: do we still need the logs ? (assuming they were for debugging)

conroy-ricketts: +1 on this comment

avazirna: Not needed, the plan was to remove them - [62b251f](https://github.com/dimagi/commcare-android/pull/3403/commits/62b251f9a00c885643060612478ab7db8ed1e047)

avazirna:

OrangeAndGreen (approved):

PR #3404, opened 2025-11-05
FCM issue after SignIn solved

Jignesh-dimagi:
## Technical Summary
https://dimagi.atlassian.net/browse/CCCT-1896

## Labels and Review

### DIFF
--- a/app/src/org/commcare/navdrawer/BaseDrawerController.kt
+++ b/app/src/org/commcare/navdrawer/BaseDrawerController.kt
@@ -120,7 +120,7 @@ class BaseDrawerController(
                 .getInstance()
                 .launchPersonalId(
                     activity,
-                    ConnectConstants.PERSONAL_ID_SIGN_UP_LAUNCH
+                    ConnectConstants.PERSONAL_ID_SIGN_UP_LAUNCH,
                 )
             closeDrawer()
         }


app/src/org/commcare/fragments/personalId/PersonalIdMessageFragment.java
```diff
@@ -168,7 +168,7 @@ private NavDirections navigateToBackupCode() {
     }
 
     private void successFlow(Activity activity) {
-        PersonalIdManager.getInstance().setStatus(PersonalIdManager.PersonalIdStatus.LoggedIn);
+        PersonalIdManager.getInstance().completeSignin();
         ConnectDatabaseHelper.setRegistrationPhase(getActivity(), ConnectConstants.PERSONALID_NO_ACTIVITY);
```
shubham1g5: I am quite confused now with this, I agree this should be the exact place from where `completeSignin` should be get called and a bit suprrised that it was not, but now I am confused about why is it getting called from other places in the code.

conroy-ricketts (dismissed):

OrangeAndGreen (approved):

app/src/org/commcare/fragments/personalId/PersonalIdMessageFragment.java
shubham1g5: Think that method has now developed more to be more of a method that should be called after the device configuration workflow. I see that @conroy-ricketts also added the [open drawer method there](https://github.com/dimagi/commcare-android/pull/3392/files) as part of first sign up process, that makes me think that method should already be getting called otherwise the functionality and demo in that PR would not really work. @Jignesh-dimagi Can you confirm this again.

Jignesh-dimagi: @shubham1g5 I am not sure if I understand your point but before this PR change, user personalId logged in status was set but FCM token was never sent to server by starting heartbeat. Drawer was working due to logged in status.

app/src/org/commcare/connect/PersonalIdManager.java
```diff
@@ -197,14 +197,17 @@ private void logBiometricInvalidations() {
...
 
     public void completeSignin() {
+        userUnlockedPersonalId();
+        if (parentActivity instanceof BaseDrawerActivity) {
+            ((BaseDrawerActivity<?>) parentActivity).openDrawer();
+        }
+    }
+
+    public void userUnlockedPersonalId(){
         personalIdSatus = PersonalIdStatus.LoggedIn;
```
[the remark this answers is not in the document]
Jignesh-dimagi: Done https://github.com/dimagi/commcare-android/pull/3404/commits/09776e0aa18635204c1a67e0d3ee9f133a2feeb1

PR #3409, opened 2025-11-06
CCCT-1840 Resend OTP With Fallback

conroy-ricketts:
## Technical Summary

I implemented a fallback to Twilio (via Personal ID) when the user reattempts to send the OTP code the first time. See the ticket linked above for more details. There are no user-visible changes for this ticket.

I feel like I'm still getting familiar with this area of the code, so let me know if I should approach this ticket from a different angle and I'd be happy to make that adjustment!

### Safety story

I verified that my changes work by following these testing steps:

1. Going through the Personal ID signup flow, verified that the **_very first_** OTP code sent to my phone number was from Firebase and that the OTP code was successfully verified in the app.
2. Going through the Personal ID signup flow, verified that the **_second_** OTP code sent to my phone number (after tapping on the "Resend Code" button exactly **_once_**) was from Twilio and that the OTP code was successfully verified in the app.
3. Going through the Personal ID signup flow, verified that the **_third_** OTP code sent to my phone number (after tapping on the "Resend Code" button exactly **_twice_**) was from Firebase and that the OTP code was successfully verified in the app.
4. Went over to BigQuery and verified that the new `otp_requested_event` Firebase analytics event was being logged properly and appropriately for my test device's ID.

### QA Plan

QA should probably follow the same steps I took but skip the last step. So here is the test plan for QA:

1. Going through the Personal ID signup flow, verify that the **_very first_** OTP code sent to your phone number was from Firebase and that the OTP code was successfully verified in the app. (you'll know it's from Firebase because the SMS message has different wording compared to Twilio!)
2. Going through the Personal ID signup flow, verify that the **_second_** OTP code sent to your phone number (after tapping on the "Resend Code" button exactly **_once_**) was from Twilio and that the OTP code was successfully verified in the app.
3. Going through the Personal ID signup flow, verify that the **_third_** OTP code sent to your phone number (after tapping on the "Resend Code" button exactly **_twice_**) was from Firebase and that the OTP code was successfully verified in the app.

### DIFF
--- a/app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
+++ b/app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
@@ -139,7 +139,9 @@ public class PersonalIdPhoneVerificationFragment extends BasePersonalIdFragment 
                 binding.connectPhoneVerifyButton.setEnabled(false);
             }
         };
-        setupOtpManager(false);
+        // Always fallback to Twilio (via Personal ID) if this is the second attempt in the session to send the user an OTP.
+        Boolean useOtpFallback = personalIdSessionData.getOtpAttempts() == 1;
+        setupOtpManager(useOtpFallback);
     }
 
     /**


app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
```diff
@@ -342,11 +351,38 @@ private void navigateToNameEntry() {
...
+            );
+        } else {
+            otpManager = new OtpManager(
+                    activity,
+                    personalIdSessionData,
+                    otpCallback
+            );
+        }
+
+        personalIdSessionData.setOtpReattempts(otpReattempts + 1);
```
Jignesh-dimagi: Because now otp re-attempt is stored in session data, whenever users navigate back and forth between screens, should app use directly Twilio if has pressed previously resend?

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
```diff
@@ -342,11 +351,38 @@ private void navigateToNameEntry() {
...
+        int otpReattempts = personalIdSessionData.getOtpReattempts();
+
+        // Fallback to Twilio (via Personal ID) if this is the first time the user reattempts to send the OTP.
+        if (otpReattempts == 0) {
+            otpManager = new OtpManager(
+                    activity,
+                    personalIdSessionData,
+                    otpCallback,
+                    SMS_METHOD_PERSONAL_ID
+            );
```
shubham1g5: - this also need to check on `personalIdSessionData.getOtpFallback()`  similar to `shouldAutoSwitchToPersonalIdAuth` . That will ensure that we are only switching to this when Personal ID fallback is allowed for this user

app/src/org/commcare/google/services/analytics/FirebaseAnalyticsUtil.java
```diff
@@ -577,6 +577,14 @@ public static void reportNavDrawerItemSelected(String selectedItem) {
                 new String[]{selectedItem});
     }
 
+    public static void reportOtpRequested(int numberOfReattempts) {
```
shubham1g5: we should ensure this gets called everytime we request otp and probably better to rephrase `OTP_REATTEMPTS` to `otp_attempts` and make it more transparent to analytics users by simply tracking nth user attempt rather (n-1)th.

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
shubham1g5: That's a good question whether we should track this per session vs restart the count every time user comes on a screen, I would think both are fine and but maybe siigltly nicer to track this per session given we already know user was not successful in their last attempt in this session.

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
conroy-ricketts: That's good to know, will do!

app/src/org/commcare/android/database/connect/models/PersonalIdSessionData.kt
```diff
@@ -41,6 +41,8 @@ data class PersonalIdSessionData(
     var otpFallback: Boolean = false,
     // the number of times the user hit the "resend code" button
     var otpReattempts: Int = 0,
```
shubham1g5: do we still need this ? (given it can be derived from `otpAttempts`)

app/src/org/commcare/google/services/analytics/FirebaseAnalyticsUtil.java
```diff
@@ -577,11 +577,11 @@ public static void reportNavDrawerItemSelected(String selectedItem) {
...
     }
 
-    public static void reportOtpRequested(int numberOfReattempts) {
+    public static void reportOtpRequested(int numberOfAttempts) {
         reportEvent(
                 CCAnalyticsEvent.OTP_REQUESTED_EVENT,
-                CCAnalyticsParam.OTP_REATTEMPTS,
-                String.format("%d reattempt(s)", numberOfReattempts)
+                CCAnalyticsParam.OTP_ATTEMPTS,
+                String.format("%d attempt(s)", numberOfAttempts)
```
shubham1g5: we should not really convert this into string, but directly use `bundle.putInt` to preserve the data type on this.

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
```diff
@@ -364,11 +366,12 @@ protected void navigateToMessageDisplay(
         Navigation.findNavController(binding.getRoot()).navigate(directions);
     }
 
-    private void setupOtpManagerForReattempt() {
-        int otpReattempts = personalIdSessionData.getOtpReattempts();
+    private void setupOtpManager(Boolean useOtpFallback) {
```
shubham1g5: nice abstraction!

app/src/org/commcare/android/database/connect/models/PersonalIdSessionData.kt
conroy-ricketts: I feel like we should so that we know to fallback to Twilio when there are currently "0" reattempts, which I think might be more idiomatic. But I'm open to changing this. What are you thinking? Use `otpAttempts` instead and fallback to Twilio when the current value of `otpAttempts` is "1"?

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
```diff
@@ -106,12 +106,8 @@ public void onFailure(OtpErrorType errorType, @Nullable String errorMessage) {
                 // Auto-switch from Firebase to PersonalId for non-recoverable errors
                 if (shouldAutoSwitchToPersonalIdAuth(errorType)) {
```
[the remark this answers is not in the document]
conroy-ricketts: Can you elaborate on this point more?

app/src/org/commcare/android/database/connect/models/PersonalIdSessionData.kt
conroy-ricketts: Cool cool!

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
```diff
@@ -126,34 +126,36 @@ public void onFailure(OtpErrorType errorType, @Nullable String errorMessage) {
...
-     * 
+     *
      * @param errorType The OTP error type from Firebase
      * @return true if we should switch to PersonalId auth, false otherwise
      */
     private boolean shouldAutoSwitchToPersonalIdAuth(OtpErrorType errorType) {
-        if(SMS_METHOD_PERSONAL_ID.equalsIgnoreCase(personalIdSessionData.getSmsMethod())) {
+        if (SMS_METHOD_PERSONAL_ID.equalsIgnoreCase(personalIdSessionData.getSmsMethod())) {
             return false;
         }
```
shubham1g5: think this can also be part of `setupOtpManager` to decide on `allowedToFallback`. The idea here is if the default method is already `SMS_METHOD_PERSONAL_ID` and then fall back logic is redundant as we are already using the fallback as default. Think we can now get rid of `shouldAutoSwitchToPersonalIdAuth` method and directly use `errorType.isNonRecoverable()` in it's place.

shubham1g5 (dismissed):
Left a suggestion but nothing blocking.

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
conroy-ricketts: I think this could work - I want to spend some more time testing this before I push changes for it

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
```diff
@@ -126,34 +126,36 @@ public void onFailure(OtpErrorType errorType, @Nullable String errorMessage) {
...
-                if(failureCode == BaseApiHandler.PersonalIdOrConnectApiErrorCodes.FAILED_AUTH_ERROR) {
+                if (failureCode == BaseApiHandler.PersonalIdOrConnectApiErrorCodes.FAILED_AUTH_ERROR) {
                     error = getString(R.string.personalid_incorrect_otp);
                 }
                 displayOtpError(error);
                 binding.connectPhoneVerifyButton.setEnabled(false);
             }
         };
-        otpManager = new OtpManager(activity, personalIdSessionData, otpCallback);
+        setupOtpManager(false);
```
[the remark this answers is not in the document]
conroy-ricketts: Ah I see what you mean. I think that both ideas are good honestly. I'll go ahead and tweak this very soon!

Jignesh-dimagi (approved):

PR #3411, opened 2025-11-07
Fix repeated crash at app startup on corrupt Connect DB.

OrangeAndGreen:
https://dimagi.atlassian.net/browse/CCCT-1868

## Product Description
Fixes a repeated crash at app startup after an initial crash due to a corrupt Connect DB.
Note this PR does not fix the original crash (i.e. why the DB got corrupted).
It just makes sure the app only crashes once when that happens.

## Technical Summary
Deleting Connect DB immediately when crash DB is called, so it's gone when the app restarts (to avoid crashing again repeatedly at app startup).

Also added an empty check to the passphrase when opening handle to Connect DB. If that were the case it would explain a corrupted DB error, so this could help determine how the app gets into this state to begin with.

## Feature Flag
PersonalID

### Safety story
I tested by starting with a clean PersonalID configuration, then temporarily overriding the passphrase to null in the code in order to cause the app to detect the DB as corrupt. Before applying the fix, I could observe the app crashing repeatedly on startup. After the fix, the app only crashes once, and when restarted the user sees PersonalID de-configured and a message explaining that something went wrong and they will need to re-configure PersonalID.

### Automated test coverage
None

### QA Plan
This will be difficult/impossible for QA to test since we're not sure how to cause the corrupt DB scenario that opens up this second bug.

### DIFF
--- a/app/src/org/commcare/connect/network/connectId/parser/RetrieveNotificationsResponseParser.kt
+++ b/app/src/org/commcare/connect/network/connectId/parser/RetrieveNotificationsResponseParser.kt
@@ -97,10 +97,9 @@ class RetrieveNotificationsResponseParser<T>(
         channelId: String,
     ): Boolean =
         isNotificationMessageType(notificationJsonObject) && notificationJsonObject
-            .getJSONObject("data")
             .get("channel") != null &&
             channelId.equals(
-                notificationJsonObject.getJSONObject("data").get("channel"),
+                notificationJsonObject.get("channel"),
             )
 
     private fun isNotificationMessageType(notificationJsonObject: JSONObject) =


app/src/org/commcare/utils/GlobalErrorUtil.java
```diff
@@ -15,7 +13,6 @@ public static void addError(GlobalErrorRecord error) {
 
     public static String handleGlobalErrors() {
```
Jignesh-dimagi: @OrangeAndGreen Changes look good, only minor nit: Rename this to `getGlobalErrorsIfAny` as it's not handling now anything?

OrangeAndGreen: Ah, makes sense. I went with just getGlobalErrors: [be2aeb7](https://github.com/dimagi/commcare-android/pull/3411/commits/be2aeb71406cf9d8a48c28bf78cbaafdf3936e27)

PR #3426, opened 2025-11-19
Correct message and authentication way whenever fingerprint not available

Jignesh-dimagi:
## Technical Summary
https://dimagi.atlassian.net/browse/CI-393

Corrected message when no biometric hardware present
Biometric prompt has authenticator set only for DEVICE_CREDENTIAL if no biometric hardware present

## Labels and Review


OrangeAndGreen (dismissed):
Noting that we want to send this to QA to test if the fix works for the device they have encountering the biometric error when trying to sign up with pattern (Android 10).

Assuming this works, we'll also need to make sure the no-fingerprint and PIN-not-allowed scenario errors correctly (this code change will cause an IllegalArgumentException in that case).

conroy-ricketts (dismissed):

Jignesh-dimagi:
Solutions done in this PR doesn't work, closing it.

PR #3434, opened 2025-11-25
Set side nav padding and margin top in percentage

jaypanchal-13:
## Product Description
Ticket -> https://dimagi.atlassian.net/browse/CCCT-1776

## Technical Summary
Set side nav margin and padding

ScreenShot ->
**Small Device**
left img -> before 
right img-> after
image

**Large Device**
left img -> before 
right img-> after
image

## Labels and Review


jaypanchal-13:
> > @Jignesh-dimagi yes with that there is change in horizontal padding
> 
> Ok got it. So, this can be handled by having `dp` in xml file only?

> @Jignesh-dimagi For different display's it will not be visualised properly. So I added percentage approch. But will be changing it to use `dimens.xml`

PR #3437, opened 2025-11-26
CCCT-1877 Phone Verification Crash (Empty Session Data)

conroy-ricketts:
## Technical Summary

This crash is happening because the user's session data is null when trying to get the phone number.

Here's why the session data is null for these users: They are putting the app in the background, and the Android OS, under memory pressure, decides to terminate the app's process to reclaim resources. Then, the entire `PersonalIdSessionDataViewModel` instance gets wiped from memory. When the users return to the app and the `PersonalIdPhoneVerificationFragment` is recreated, a new empty `PersonalIdSessionDataViewModel` is created because the old one was destroyed during process death and the `PersonalIdSessionData` inside the view model was never re-initialized.

I assume that the reason users are backgrounding the app in the first place is to read the OTP from the text message that gets sent to them.

Here is the easiest way I found to reproduce the crash:
1) Enable the "Developer options" in your phone's settings if you have not already.
2) In the "Developer options" settings, find the "Don't keep activities" setting under the "Apps" section and enable it.
3) In the CommCare app, go through the Personal ID signup flow until you get to the OTP verification screen, but do NOT enter the OTP - stay on that page.
4) Background the app.
5) Foreground the app.
6) Observe that the app crashes.

Here's a video demonstrating those reproduction steps above:

https://github.com/user-attachments/assets/876421f6-d481-4a03-a72f-ec1f57c36ec4

**I think the best way to fix this crash is to make`PersonalIdSessionDataViewModel` lifecycle-aware.**

I considered kicking the user back to the start of the Personal ID signup flow (or even the activity they were on just before they started the Personal ID signup flow) to avoid the crash, but if you really think about this, the user will likely see the crash again because they are likely to background the app again to read the OTP text message.

We could use the phone number that is saved to the `savedInstanceState` bundle in `PersonalIdPhoneVerificationFragment`, but the app will still crash because we also need the session data for getting/setting `otpAttempts`, getting the `otpFallback`, and getting the `smsMethod`.

Note that since the user is coming back to the same screen after process death, we also need to tweak some logic in the `PersonalIdPhoneVerificationFragment` because of our OTP fallback implementation or else we will have more logical errors.

Let me know what y'alls thoughts are on this!

### Safety story

I verified that, with these changes, the app no longer crashes.

I also verified that I'm still able to successfully complete the entire Personal ID signup flow with these changes - the app behaves as expected, especially when the OTP fallback is used.

### QA Plan

QA should follow the reproduction steps I listed above and ensure that the app does not crash. QA should also background and freground the app at various states of otp-recreations (before sending code, after sending code, after verifying otp , after resending code once and twice) to check that we restore successfully and able to verify otp in each of those scenarios. 

Also, QA should follow the test notes that I outlined [here](https://github.com/dimagi/commcare-android/pull/3409) to ensure that there are no regressions.

### DIFF
--- a/app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
+++ b/app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
@@ -196,7 +196,10 @@ public class PersonalIdPhoneVerificationFragment extends BasePersonalIdFragment 
         boolean verifyButtonEnabled = savedInstanceState.getBoolean(KEY_VERIFY_BUTTON_ENABLED);
         String otpRequestTimeString = savedInstanceState.getString(KEY_OTP_REQUEST_TIME_STRING);
 
-        otpRequestTime = DateTime.parse(otpRequestTimeString);
+        if (otpRequestTimeString != null) {
+            otpRequestTime = DateTime.parse(otpRequestTimeString);
+        }
+
         binding.connectPhoneVerifyButton.setEnabled(verifyButtonEnabled);
     }
 


app/src/org/commcare/activities/connect/viewmodel/PersonalIdSessionDataViewModel.java
```diff
@@ -2,24 +2,30 @@
...
 import androidx.lifecycle.ViewModel;
 
 public class PersonalIdSessionDataViewModel extends ViewModel {
-    private PersonalIdSessionData personalIdSessionData;
+    private static final String PERSONAL_ID_SESSION_DATA_KEY = "PERSONAL_ID_SESSION_DATA_KEY";
+    private final SavedStateHandle savedStateHandle;
+
+    public PersonalIdSessionDataViewModel(SavedStateHandle savedStateHandle) {
+        this.savedStateHandle = savedStateHandle;
+    }
```
shubham1g5: Does this constructor gets called by itself when activity gets re-created, resulting in re-creation of the `PersonalIdSessionDataViewModel` as it was saved ?

conroy-ricketts: Yes, when a `savedStateHandle` is in the viewmodel's constructor, the Android framework provides it to the viewmodel automatically, and the `savedStateHandle` will contain any saved data when the state is restored

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
```diff
@@ -375,12 +382,14 @@ private void setupOtpManager(Boolean useOtpFallback) {
                     otpCallback,
                     SMS_METHOD_PERSONAL_ID
             );
+            personalIdSessionData.setOtpFallbackUsedBeforeProcessDeath(true);
```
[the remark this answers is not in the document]
shubham1g5: > if the code they received was from either Twilio or Firebase. We need to cover both cases (Twilio & Firebase), and I think we cannot rely solely on the number of otpAttempts.

What if we restore all things required to restore the exact UI state on activity recreation -  `lastOtpMethod`, `otpRequestTime` and `otpAttempts` .

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
```diff
@@ -162,7 +169,13 @@ private boolean shouldAutoSwitchToPersonalIdAuth(OtpErrorType errorType) {
     @Override
     public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState) {
         binding = ScreenPersonalidPhoneVerifyBinding.inflate(inflater, container, false);
-        setupInitialState();
+
+        // If onCreateView() is being called again after process death, the user was already on this
+        // screen and receivied the initial OTP. In that case, we do not want to reset the initial state.
+        if (personalIdSessionData.getOtpAttempts() == 0) {
+            setupInitialState();
+        }
```
shubham1g5: > If onCreateView() is being called again after process death, the user was already on this screen and receivied the initial OTP. In that case, we do not want to reset the initial state.

Think a more concrete way of knowing that we are restoring after process death can be to check `savedInstanceState!=null`. In general this is a good pattern to follow - 

````
if(savedInstanceState!=null) {
 restoreState();
}else {
   setupInitialState();
}
````

File: app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java [hunk 4]
```diff
@@ -281,6 +294,7 @@ public void onDestroyView() {
     public void onSaveInstanceState(@NonNull Bundle outState) {
         super.onSaveInstanceState(outState);
         outState.putString(KEY_PHONE, primaryPhone);
+        outState.putBoolean(KEY_OTP_FALLBACK_USED, otpFallbackUsed);
```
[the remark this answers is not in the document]
conroy-ricketts: That would lead us back to the issue discussed in [this thread](https://github.com/dimagi/commcare-android/pull/3437#discussion_r2568925387)

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
```diff
@@ -57,6 +62,7 @@ public class PersonalIdPhoneVerificationFragment extends BasePersonalIdFragment
     private PersonalIdSessionData personalIdSessionData;
     OtpVerificationCallback otpCallback;
     private ActivityResultLauncher<Intent> smsConsentLauncher;
+    private String lastOtpMethod = SMS_METHOD_FIREBASE;
```
shubham1g5: we should not be giving it a default value here as it doesn't really have a default, the initial method can be firebase or twilio/personalId based on server response.

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
```diff
@@ -281,6 +306,9 @@ public void onDestroyView() {
     public void onSaveInstanceState(@NonNull Bundle outState) {
         super.onSaveInstanceState(outState);
         outState.putString(KEY_PHONE, primaryPhone);
+        outState.putString(KEY_LAST_OTP_METHOD, lastOtpMethod);
+        outState.putBoolean(KEY_VERIFY_BUTTON_ENABLED, binding.connectPhoneVerifyButton.isEnabled());
+        outState.putString(KEY_OTP_REQUEST_TIME_STRING, otpRequestTime.toString());
```
shubham1g5: Confirming `otpRequestTime` can't be null here ? Mostly wondering if the activity can get destroyed before we init it ?

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
```diff
@@ -139,8 +149,9 @@ public void onPersonalIdApiFailure(
                 binding.connectPhoneVerifyButton.setEnabled(false);
             }
         };
-        // Always fallback to Twilio (via Personal ID) if this is the second attempt in the session to send the user an OTP.
-        Boolean useOtpFallback = personalIdSessionData.getOtpAttempts() == 1;
```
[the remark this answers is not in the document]
conroy-ricketts: My thoughts were that checking for `otpAttempts` when initializing the OTP manager was unnecessary (and even wrong in some cases) because it will always be equal to `0` at this point in the code unless the user does something like background/foreground the app (regardless of whether or not they hit the "Resend Code" button) - but if the user ever does that then we want to be sure to setup the OTP manager the exact same way it was previously at this point in the code or else there may be an error verifying the OTP with the wrong service

For example, if the user goes to the phone verification screen for the first time, `otpAttempts` is 0 here, and **_then_** a new OTP is requested later in the code. If the user backgrounds the app at this point and the process is killed, and then the user comes back to the app, then the `otpAttempts` is now 1 when we check it and Twilio is used to setup the OTP manager which is incorrect because the user still has not entered their code from Firebase

For more context, we are still checking for `otpAttempts` in the `setupListeners()` function:
```
    private void setupListeners() {
        binding.connectResendButton.setOnClickListener(v -> {
            // Always fallback to Twilio (via Personal ID) if this is the first time the user reattempts to send the OTP.
            Boolean useOtpFallback = personalIdSessionData.getOtpAttempts() == 1;
            setupOtpManager(useOtpFallback);
            requestOtp();
        });
        ...
```

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
conroy-ricketts: Good question - it being null is very unlikely but technically, `otpRequestTime` will be null until we init it via `onCreateView()` (which is the worst case scenario I think).

Since it _is_ possible, I'll add a null check to `restoreState()` when we parse the time string to avoid an unnecessary crash

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
```diff
@@ -139,8 +149,9 @@ public void onPersonalIdApiFailure(
...
+
+        // The last OTP method may be Twilio (via Personal ID) after restoring this fragment.
+        Boolean useOtpFallback = SMS_METHOD_PERSONAL_ID.equalsIgnoreCase(lastOtpMethod);
```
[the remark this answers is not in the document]
conroy-ricketts: Oh that's interesting. So there are some cases where the OTP method is always `SMS_METHOD_PERSONAL_ID` and as such we would never use a "fallback"? If that's the case, we may want to rethink the entirety of our OTP fallback implementation because previously we were defaulting to Firebase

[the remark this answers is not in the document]
shubham1g5: yeah that makes sense to me.

shubham1g5 (approved):
Looks good to go and great work on troubleshooting and implementing the fix here, my sense is we might need to do this kind of background/re-creation testing with all our screens as I think we are not really storing state in Connect side of code anywhere, but it does feel specially important for OTP screen and I am glad to have this fixed.

PR #3446, opened 2025-12-02
CI-418 Error Message For Opportunities Failing To Refresh

conroy-ricketts:
### [CI-418](https://dimagi.atlassian.net/browse/CI-418) 

_(this is for a mobile action item)_

## Product Description

I updated the user-facing error message in the event of a 403 forbidden error when tapping the refresh button for the opportunities list (and for all 403 forbidden errors for that matter).

## Technical Summary

I noticed that we are already handling the forbidden error case with the correct message for the `PersonalIdPhoneFragment` [here](https://github.com/dimagi/commcare-android/blob/c81853a44d3b701e20d46cd05412cb1d84a52a94/app/src/org/commcare/fragments/personalId/PersonalIdPhoneFragment.java#L462), so there is no need to make any specific changes for that error message in Personal ID after generalizing the default error message for all 403 forbidden errors.

Also, when I was looking at our other error messages, I noticed that we had some missing string translations. So I went ahead and added them.

### Safety story

I verified this code will show an updated error message when the user taps on the refresh button and the API call fails with a 403 forbidden error by forcing it in the code.

### QA Plan

For anyone that is able, we should verify that the non-English translations are correct.

### DIFF
--- a/app/res/values-fr/strings.xml
+++ b/app/res/values-fr/strings.xml
@@ -185,7 +185,7 @@ License.
     <string name="recovery_network_outdated">Cette application est obsolète. Veuillez mettre à jour l\'application sur le Google Play Store pour continuer.</string>
     <string name="recovery_network_token_unavailable">Un problème de réseau est survenu. Veuillez réessayer.</string>
     <string name="recovery_network_unauthorized">Vous n\'êtes pas autorisé à faire cette demande.</string>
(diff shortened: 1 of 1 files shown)


conroy-ricketts:
@shubham1g5 Agreed, that makes sense to me! I'll update this and re-request your review when I pull this PR out of draft

OrangeAndGreen (approved):
Thanks for filling in the missing string translations!

shubham1g5 (approved):

PR #3448, opened 2025-12-03
Parsing additional map display info from case data

OrangeAndGreen:
https://dimagi.atlassian.net/browse/CCCT-1935

Cross-request: https://github.com/dimagi/commcare-core/pull/1513

## Product Description
No user-facing changes

## Technical Summary
New EntityMapDisplayInfo to hold additional map display info for entities.
Parsing additional map-related fields from case data and holding in memory (no display yet).
Validation for required data conditions.

## Feature Flag
Entity Map

### Safety story
Verified the existing functionality works (showing a single marker for each entity based on "address" field).
Verified additional data can be present or missing and everything still works.

Screenshot showing an example of the parsed EntityMapDisplayInfo object when all relevant fields are present:
image

Screenshot showing error message display (when error encountered un map case data):
image

### Automated test coverage
None

### QA Plan
Verify that the existing functionality in the Entity Map display still works (nothing should change from previous versions).
Try adding the following fields to a case and make sure everything still works:
* geo_boundary
* geo_boundary_color_hex
* geo_points
* geo_points_colors_hex

cross-request: https://github.com/dimagi/commcare-core/pull/1513

### DIFF
--- a/app/src/org/commcare/gis/EntityMapUtils.kt
+++ b/app/src/org/commcare/gis/EntityMapUtils.kt
@@ -13,8 +13,8 @@ import org.javarosa.core.model.data.GeoPointData
 import org.javarosa.core.model.data.UncastData
 import org.javarosa.core.model.instance.TreeReference
 import org.javarosa.core.model.utils.GeoPointUtils
-import org.javarosa.core.services.Logger
 import org.javarosa.core.model.utils.PolygonUtils
+import org.javarosa.core.services.Logger
 import java.util.Vector
 import javax.annotation.Nullable
 
@@ -108,7 +108,7 @@ object EntityMapUtils {
 
     @Nullable
     private fun parsePointListFromString(pointListString: String): List<LatLng>? {
-        if(pointListString.isEmpty()) {
+        if (pointListString.isEmpty()) {
             return null
         }
         val parts = pointListString.trim().split("\\s+".toRegex())
@@ -124,7 +124,7 @@ object EntityMapUtils {
 
     @Nullable
     private fun parseHexColorList(colorsString: String): List<Int>? {
-        if(colorsString.isEmpty()) {
+        if (colorsString.isEmpty()) {
             return null
         }
 
@@ -188,7 +188,7 @@ object EntityMapUtils {
             if (location == null) {
                 try {
                     location = getEntityLocation(entity, detail, i)
-                } catch(e: IllegalArgumentException) {
+                } catch (e: IllegalArgumentException) {
                     Logger.exception("Error parsing entity location for map display", e)
                     errorEncountered = true
                 }
@@ -197,7 +197,7 @@ object EntityMapUtils {
             if (boundary == null) {
                 try {
                     boundary = getEntityBoundary(entity, detail, i)
-                } catch(e: IllegalArgumentException) {
+                } catch (e: IllegalArgumentException) {
                     Logger.exception("Error parsing entity boundary for map display", e)
                     errorEncountered = true
                 }
@@ -206,7 +206,7 @@ object EntityMapUtils {
             if (boundaryColorHex == null) {
                 try {
                     boundaryColorHex = getEntityBoundaryColor(entity, detail, i)
-                } catch(e: IllegalArgumentException) {
+                } catch (e: IllegalArgumentException) {
                     Logger.exception("Error parsing entity boundary color for map display", e)
                     errorEncountered = true
                 }
@@ -215,7 +215,7 @@ object EntityMapUtils {
             if (points == null) {
                 try {
                     points = getEntityPoints(entity, detail, i)
-                } catch(e: IllegalArgumentException) {
+                } catch (e: IllegalArgumentException) {
                     Logger.exception("Error parsing entity points for map display", e)
                     errorEncountered = true
                 }
@@ -224,11 +224,22 @@ object EntityMapUtils {
             if (pointColorsHex == null) {
                 try {
                     pointColorsHex = getEntityPointColors(entity, detail, i)
-                } catch(e: IllegalArgumentException) {
+                } catch (e: IllegalArgumentException) {
                     Logger.exception("Error parsing entity point colors for map display", e)
                     errorEncountered = true
                 }
             }
+        }
+
+        if (pointColorsHex != null && points != null && pointColorsHex.size != points.size) {
+            Logger.exception(
+                "Mismatched point colors and points for entity map display",
+                IllegalArgumentException(
+                    "Number of point colors (${pointColorsHex.size}) does not match number of points (${points.size})"
+                )
+            )
+            pointColorsHex = null
+            errorEncountered = true
         }
 
         return if (errorEncountered || location != null || boundary != null || points != null) {


app/src/org/commcare/gis/EntityMapUtils.kt
```diff
@@ -25,24 +28,152 @@ object EntityMapUtils {
...
+            Logger.exception("Error parsing entity boundary for map display", e)
+            return null
+        }
+    }
+
+    @Nullable
+    private fun parseHexColor(colorString: String): Int? {
+        if (colorString.isEmpty()) {
+            return null
+        }
```
Jignesh-dimagi: Same, not required to check as will be handled by `IllegalArgumentException`exception. This will also log the exception.

app/src/org/commcare/gis/EntityMapUtils.kt
```diff
@@ -25,24 +28,152 @@ object EntityMapUtils {
...
+            Logger.exception("Error parsing hex color: $colorString", e)
+            return null
+        }
+    }
+
+    @Nullable
+    private fun parseHexColorList(colorsString: String): Array<Int>? {
+        if (colorsString.isEmpty()) {
+            return null
+        }
```
Jignesh-dimagi: Same, not required to check as will be handled by `Exception`. This will also log the exception.

app/src/org/commcare/gis/EntityMapUtils.kt
```diff
@@ -25,24 +28,152 @@ object EntityMapUtils {
...
+
+        try {
+            val parts = boundaryString.trim().split("\\s+".toRegex())
+
+            // Must have even number of values (pairs of lat/lng)
+//            if (parts.size < 2 || parts.size % 2 != 0) {
+//                return null
+//            }
```
Jignesh-dimagi: Remove this please if not required

app/src/org/commcare/gis/EntityMapUtils.kt
```diff
@@ -60,4 +191,75 @@ object EntityMapUtils {
...
+        return if (location != null || boundary != null || points != null) {
+            EntityMapDisplayInfo(
+                location = location,
+                boundary = boundary,
+                boundaryColorHex = boundaryColorHex,
+                points = points,
+                pointColorsHex = pointColorsHex
+            )
+        } else {
+            null
```
Jignesh-dimagi: App should log exception here

app/src/org/commcare/gis/EntityMapUtils.kt
```diff
@@ -25,24 +28,152 @@ object EntityMapUtils {
...
+        return null
+    }
+
+    @JvmStatic
+    fun getEntityBoundary(
+        entity: Entity<TreeReference>,
+        detail: Detail,
+        fieldIndex: Int
+    ): Array<LatLng>? {
+        if ("cc_geo_boundary" == detail.templateForms[fieldIndex]) {
```
shubham1g5: lint: would be nice to use constants for any such properties.

app/src/org/commcare/gis/EntityMapUtils.kt
```diff
@@ -25,24 +28,152 @@ object EntityMapUtils {
...
+
     @JvmStatic
     fun getNeededEntityDatum(): EntityDatum? {
         val datum = CommCareApplication.instance().currentSession.neededDatum
-        return if (datum is EntityDatum) {
-            datum
-        } else {
-            null
-        }
+        return datum as? EntityDatum
```
conroy-ricketts: Nice cleanup 👍

app/src/org/commcare/gis/EntityMapUtils.kt
```diff
@@ -25,24 +28,152 @@ object EntityMapUtils {
...
         return null
     }
 
+    @JvmStatic
+    fun getEntityPointColors(
+        entity: Entity<TreeReference>,
+        detail: Detail,
+        fieldIndex: Int
+    ): Array<Int>? {
+        if ("cc_geo_points_colors" == detail.templateForms[fieldIndex]) {
```
shubham1g5: lets remove any `_points` properties for now based on our discussion this evening.

shubham1g5: this is back in scope and hence we need to preserve these formats.

shubham1g5:
@OrangeAndGreen I have updated the template values in the ticket here - https://dimagi.atlassian.net/browse/CCCT-1935 given we are no longer operating on reserved case properties. Would you be able to update these in the PR as well, thanks!

app/src/org/commcare/gis/EntityMapUtils.kt
```diff
@@ -25,24 +28,152 @@ object EntityMapUtils {
...
+
+            return if (points.isNotEmpty()) {
+                points.toTypedArray()
+            } else {
+                null
+            }
+        } catch (e: NumberFormatException) {
+            Logger.exception("Error parsing entity boundary for map display", e)
+            return null
+        }
```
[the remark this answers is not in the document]
OrangeAndGreen: This is handled by the createPolygon function now

app/src/org/commcare/gis/EntityMapUtils.kt
```diff
@@ -60,4 +191,75 @@ object EntityMapUtils {
...
+            }
+
+            if (points == null) {
+                points = getEntityPoints(entity, detail, i)
+            }
+
+            if (pointColorsHex == null) {
+                pointColorsHex = getEntityPointColors(entity, detail, i)
+            }
+        }
```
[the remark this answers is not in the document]
OrangeAndGreen: This would be a bug if it happened, so should crash.

app/src/org/commcare/gis/EntityMapUtils.kt
```diff
@@ -25,24 +28,152 @@ object EntityMapUtils {
...
-    private fun getLatLngFromAddress(address: String): GeoPointData? {
+    private fun getLatLngFromAddress(address: String): LatLng? {
         try {
             if (!address.contentEquals("")) {
                 val data = GeoPointData().cast(UncastData(address))
-                return  data
+                return LatLng(data.latitude, data.longitude)
             }
-        } catch (ignored: IllegalArgumentException) { }
+        } catch (ignored: IllegalArgumentException) {
```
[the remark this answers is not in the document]
OrangeAndGreen: Nope! But I added logging to it now: [dba1093](https://github.com/dimagi/commcare-android/pull/3448/commits/dba1093292eaab8488d6a486c04b021d0e7fde44)

app/src/org/commcare/gis/EntityMapUtils.kt
```diff
@@ -10,39 +12,163 @@ import org.commcare.utils.AndroidInstanceInitializer
...
+    @Nullable
+    private fun parseBoundaryFromString(boundaryString: String): List<LatLng>? {
+        try {
+            val parts = boundaryString.trim().split("\\s+".toRegex())
+            val polygon = PolygonUtils.createPolygon(parts)
+            return polygon.map { LatLng(it.latitude, it.longitude) }
+                .toMutableList()
+        } catch (e: IllegalArgumentException) {
+            Logger.exception("Error parsing entity boundary for map display", e)
+            return null
```
shubham1g5: Think we want to throw any exceptions that might signify incorrect data from here and handle it in map screen by showing a user visible error that indicates something is off about that map they are seeing.

app/src/org/commcare/gis/EntityMapUtils.kt
```diff
@@ -60,4 +186,75 @@ object EntityMapUtils {
...
+            if (pointColorsHex == null) {
+                pointColorsHex = getEntityPointColors(entity, detail, i)
+            }
+        }
+
+        if (boundary != null && boundaryColorHex == null) {
+            boundary = null
+            Logger.exception(
+                "Entity has boundary but no boundary color specified for map display",
+                RuntimeException()
```
shubham1g5: think that's something we should expect and should just be using a default color in these cases.

app/src/org/commcare/gis/EntityMapUtils.kt
```diff
@@ -60,4 +186,75 @@ object EntityMapUtils {
...
+            )
+        }
+
+        if (points != null) {
+            if (pointColorsHex == null) {
+                points = null
+                Logger.exception(
+                    "Entity has points but no point colors specified for map display",
+                    RuntimeException()
```
shubham1g5: also a valid scenario where we should be using a default color.

shubham1g5 (changes requested):

shubham1g5:
Bdw think `cross-request` needs to be defined at the very end of PR description for Jenkins to pick it up correctly (just edited it)

app/src/org/commcare/gis/EntityMapUtils.kt
OrangeAndGreen: Removed

app/src/org/commcare/gis/EntityMapUtils.kt
OrangeAndGreen: Removed

app/src/org/commcare/gis/EntityMapDisplayInfo.kt
```diff
@@ -0,0 +1,11 @@
+package org.commcare.gis
+
+import com.google.android.gms.maps.model.LatLng
+
+data class EntityMapDisplayInfo(
+    val location: LatLng? = null,
+    val boundary: List<LatLng>? = null,
+    val boundaryColorHex: Int? = null,
+    val points: List<LatLng>? = null,
+    val pointColorsHex: List<Int>? = null
```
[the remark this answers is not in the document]
OrangeAndGreen: Easy enough to do, but I'm curious why we would want to? Unless I'm mistaken those headers will be used if the corresponding fields are displayed in the case list (via existing code), but to my knowledge the code here (specific to the usage in EntityMapActivity) has no need for those headers.

Noting that the title + snippet text that's displayed when a marker or polygon is clicked comes directly from the entity, and geo points are not clickable (if they were using this header would result in all geo points using the same text anyway).

OrangeAndGreen:
Merged in the code for drawing polygons and geo points

app/src/org/commcare/gis/EntityMapUtils.kt
shubham1g5: >  it most-likely indicates an error on the admin side.

Agree that this would be an app configuration error and the CommCare standard for any app errors is to communicate it out to user as an "Error" in App configuration. 


> How should we show the message? Would a snackbar with an OK button to dismiss be okay? Thinking similar to the new data refresh indicators being added in a different PR right now...

Think this needs to be sticky and non-dismissable error UI. As such a snack bar message that can't be dismissed should be ok here OR just an error text overlapping the map will be sufficient as well. 


> What should the message say?

`There was an error in app configuration due to which not all markers on the map are loaded correctly" .

app/src/org/commcare/gis/EntityMapDisplayInfo.kt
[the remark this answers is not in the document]
shubham1g5: yeah that sounds good

shubham1g5 (approved):

PR #3457, opened 2025-12-09
CI-423 Categorize Demo Users In Analytics (Phase 2)

conroy-ricketts:
## Technical Summary

After discussion with @OrangeAndGreen, we decided that more code changes are needed to reach the desired outcome for this ticket.

For context, in #3454 we added changes to report if a demo user is currently logged-in with Personal ID for all Firebase analytics events. The problem is that the user funnel tracks certain events _leading up to_ account creation or recovery, and currently we are only flagging demo users once they've completed the configuration workflow.

So, in other words, we also need a way of knowing if someone has entered a demo phone number as early as possible in the Personal ID setup flow - ideally as soon as they enter the phone number on the phone page and we create a new `PersonalIdSessionData`.

**_Edit:_** So we decided that the easiest way to achieve this is to flag demo users (with our `FirebaseAnalytics` instance) as soon as we get the flag from server and the `PersonalIdSessionData` is first created. The `FirebaseAnalytics` instance is used for all analytics events in our app, so the flag will be recorded with the next analytics event.

### Safety story

I verified that we are correctly identifying and reporting users that are using a demo phone number with these changes by actually using both a demo phone number and a real phone number.

### DIFF
--- a/app/src/org/commcare/gis/EntityMapActivity.java
+++ b/app/src/org/commcare/gis/EntityMapActivity.java
@@ -21,12 +21,14 @@ import com.google.android.gms.maps.model.Marker;
 import com.google.android.gms.maps.model.MarkerOptions;
 import com.google.android.gms.maps.model.Polygon;
 import com.google.android.gms.maps.model.PolygonOptions;
+import com.google.firebase.perf.metrics.Trace;
 
 import org.commcare.CommCareApplication;
 import org.commcare.activities.CommCareActivity;
 import org.commcare.activities.EntityDetailActivity;
 import org.commcare.cases.entity.Entity;
 import org.commcare.dalvik.R;
+import org.commcare.google.services.analytics.CCPerfMonitoring;
 import org.commcare.preferences.HiddenPreferences;
 import org.commcare.suite.model.Detail;
 import org.commcare.suite.model.DetailField;
@@ -41,6 +43,7 @@ import org.javarosa.xpath.XPathException;
 
 import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Vector;
 
 import androidx.annotation.NonNull;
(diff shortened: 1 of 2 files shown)


OrangeAndGreen (changes requested):

app/src/org/commcare/fragments/personalId/PersonalIdPhoneFragment.java
```diff
@@ -430,6 +433,11 @@ private void makeStartConfigurationCall(String requestHash,
             public void onSuccess(PersonalIdSessionData sessionData) {
                 personalIdSessionDataViewModel.setPersonalIdSessionData(sessionData);
                 personalIdSessionDataViewModel.getPersonalIdSessionData().setPhoneNumber(phone);
+
+                boolean isDemoUser = Boolean.TRUE.equals(sessionData.getDemoUser());
+                FirebaseAnalytics analyticsInstance = CommCareApplication.instance().getAnalyticsInstance();
```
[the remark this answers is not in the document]
conroy-ricketts: Good idea, I also did a very small refactor in `FirebaseAnalyticsUtil` so that we always use the helper function when it makes sense. Let me know what you think!

OrangeAndGreen (approved):

app/src/org/commcare/fragments/personalId/PersonalIdPhoneFragment.java
```diff
@@ -430,6 +429,10 @@ private void makeStartConfigurationCall(String requestHash,
             public void onSuccess(PersonalIdSessionData sessionData) {
                 personalIdSessionDataViewModel.setPersonalIdSessionData(sessionData);
                 personalIdSessionDataViewModel.getPersonalIdSessionData().setPhoneNumber(phone);
+
+                boolean isDemoUser = Boolean.TRUE.equals(sessionData.getDemoUser());
+                FirebaseAnalyticsUtil.flagPersonalIDDemoUser(isDemoUser);
```
Jignesh-dimagi: Can we directly use FirebaseAnalyticsUtil.flagPersonalIDDemoUser(sessionData.getDemoUser())?

app/src/org/commcare/google/services/analytics/FirebaseAnalyticsUtil.java
```diff
@@ -123,11 +128,6 @@ private static void setUserProperties(FirebaseAnalytics analyticsInstance) {
         }
 
         analyticsInstance.setUserProperty(CCAnalyticsParam.BUILD_NUMBER, String.valueOf(BuildConfig.VERSION_CODE));
-
-        Boolean isPersonalIDDemoUser = ReportingUtils.getIsPersonalIDDemoUser();
-        if (isPersonalIDDemoUser != null) {
-            analyticsInstance.setUserProperty(CCAnalyticsParam.IS_PERSONAL_ID_DEMO_USER, String.valueOf(isPersonalIDDemoUser));
-        }
```
shubham1g5: why did we decide to remove this from here and add it somewhere else ?

shubham1g5: >  I'm calling the wrapper function from inside reportEvent() instead of setUserProperties() because I think it makes more sense from a readability standpoint

This is what I am concerned about as it creates inconsistency with how other user properties are handled. Think user properties doesn't generally change properties across events, which is why we have not been setting them from reportEvent methods.

PR #3469, opened 2025-12-18
CCCT-1992 Tweak Delivery Completion Banner (Part 2)

conroy-ricketts:
## Product Description

I was a bit too quick to merge in #3467 😅. Turns out we are using a different message card in the `ConnectDeliveryProgressFragment`, so there is a second place to update the colors.

So, in #3467 I updated the colors in the app home screen. In this PR I updated the colors in the delivery screen.

See #3467 for more context.

https://github.com/user-attachments/assets/b84ab242-8953-418e-bf25-d329317b2119

### Safety story

I verified that the color schemes changed for the correct messages (**_on the delivery screen and not just the app home screen!_**).

### QA Plan

Test that we are able to see the updated color scheme for the two messages (**_on both the CommCare app home screen and the delivery screen_**):

- Finished/expired opportunity:
  - "The job has ended. You will not earn any progress for additional work."
- Maxed out visits/deliveries for an opportunity:
  - "You have completed the maximum number of visits. You will not earn any progress for additional work."

### DIFF
--- a/app/src/org/commcare/fragments/connect/ConnectDeliveryProgressFragment.java
+++ b/app/src/org/commcare/fragments/connect/ConnectDeliveryProgressFragment.java
@@ -6,6 +6,7 @@ import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
 
+import androidx.annotation.ColorRes;
 import androidx.annotation.NonNull;
 import androidx.core.content.ContextCompat;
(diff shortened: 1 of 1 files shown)


OrangeAndGreen:
Ah, interesting. Okay, could go either way here. It might be easy enough to just add the icon now since we have it in the linked Figma, but I'm also okay with saving it for a different ticket if you'd rather.

PR #3470, opened 2025-12-22
CCCT-1990 BaseConnectFragment NPE Crash

conroy-ricketts:
## Technical Summary

This crash happens when we try to update the UI after an asynchronous API call if the user navigates away from the screen we are trying to update.

In other words, there are cases where we are attempting to retrieve the `binding` (to update the UI) after it's set to `null` in `BaseConnectFragment.onDestroyView()` ([here](https://github.com/dimagi/commcare-android/blob/0dcdc43d805cc09bea36481be1247748f9c46feb/app/src/org/commcare/fragments/base/BaseConnectFragment.kt#L66)).

So, since we are incorrectly updating the UI, I searched through all of our Connect fragments and added checks to any fragment where we are updating the UI with that binding after an API call (if we weren't doing so already) so that this crash does not happen again. I actually fixed two crashes at the same time by doing this.

---

### Repro Path 1 (Example of the original crash)

1) After completing a learning assessment, go to the opportunity details page.
2) Tap the "DOWNLOAD" button ***and then immediately hit the back button.***
3) Observe the app crash.

https://github.com/user-attachments/assets/d01e38d2-cd76-4f50-ba02-eb0b14b59559

The app crashes here with `java.lang.NullPointerException at org.commcare.fragments.base.BaseConnectFragment.getBinding())` - this is the original crash I reported in the ticket.

With my changes, this is what the app looks like when it does ***not*** crash:

https://github.com/user-attachments/assets/4608ad99-e0ff-4463-82a2-5338050b1b51

---

### Repro Path 2 (A different but similar crash)

1) Go to the job intro screen for a brand new opportunity.
2) Tap the "GO TO LEARN APP" button ***and then immediately hit the back button.***
3) Observe the app crash.

https://github.com/user-attachments/assets/b9cde1d9-8c09-4aee-a891-af8d21b11db3

The app crashes here with `java.lang.IllegalStateException: Fragment ConnectJobIntroFragment... not attached to an activity.`

With my changes, this is what the app looks like when it does ***not*** crash:

https://github.com/user-attachments/assets/da9f7198-da54-42e7-9769-46b7460ebd39

---

I was not able to reproduce the app crashing when refreshing data ([here](https://github.com/dimagi/commcare-android/blob/0dcdc43d805cc09bea36481be1247748f9c46feb/app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java#L75) and [here](https://github.com/dimagi/commcare-android/blob/0dcdc43d805cc09bea36481be1247748f9c46feb/app/src/org/commcare/fragments/connect/ConnectDeliveryProgressFragment.java#L120)) or confirming payments ([here](https://github.com/dimagi/commcare-android/blob/0dcdc43d805cc09bea36481be1247748f9c46feb/app/src/org/commcare/fragments/connect/ConnectDeliveryProgressFragment.java#L150)) in both the `ConnectJobsListsFragment` and the `ConnectDeliveryProgressFragment`, which is actually a bit weird to me because the app should be crashing similarly to the two repro paths I outlined above unless I'm missing something. Regardless, I went ahead and added checks to those two fragments as well anyways because we are still using the binding to update the UI after an API call and it's good practice. Let me know if there are any objections to that.

### Safety story

I verified that I no longer see the app crash for the two repro paths above.

I verified that the app behaves as expected when not forcing a crash.

### QA Plan

We should ensure that the app never crashes in the two repro paths I outlined above.

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -142,7 +142,8 @@
             android:name="com.google.mlkit.vision.DEPENDENCIES"
             android:value="face" />
         <activity
-            android:name="org.commcare.activities.connect.ConnectMessagingActivity">
+            android:name="org.commcare.activities.connect.ConnectMessagingActivity"
+            android:theme="@style/CommonTheme.NoActionBar">
         </activity>
         <activity
             android:label="@string/application_name"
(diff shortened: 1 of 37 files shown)


conroy-ricketts:
And also +1 on the comment about `viewModelScope`

OrangeAndGreen:
Agreed on all the above, #1 being that long-term we should get these moved to view model scopes. Additional thoughts:
1. @Jignesh-dimagi Do you expect removing the `binding=null` in `onDestroyView` to result in fewer crashes? If it has no effect or could result in more crashes then I'd be wary of making that change in the short-term.
2. If I understand correctly, the remaining concern is an API call returning after `onDestroyView` was called but while `isAdded()` is still true. Would null-checking on `binding` instead of (or in addition to) relying on `isAdded` be better?

PR #3473, opened 2025-12-26
CCCT-2002 Opportunity Card Redesign

conroy-ricketts:
## Product Description

I redesigned the connect cards in the opportunity progress pages.

Here's some screenshots of the card in various different states:

Here's a video demonstrating the behavior of the new buttons:
 
https://github.com/user-attachments/assets/bf149dff-9cef-4e45-8028-5df18f558717

## Technical Summary

- When I added the new icon for the warning messages, I went back and added the new icon to the app home screens as well (the home screen messages were updated in #3467).
  - I chose to hide the warning icon when an opportunity is ready to transition to delivery: 

- Since the XML layout for the connect card is reused in several different place, I set the new UI elements I added to `android:visibility="gone"` by default so that we don't mess up the UI on other screens. 
  - The new redesign has us move the opportunity completion date to a different location on the card, so I added a 2nd text view element rather than deleting the old one so that we don't mess up the UI on the app home screens. That's why you see both `connect_job_end_date` and `connect_job_end_date_sub_heading` in `view_job_card.xml`.
  - That's also why I added a new utility function for getting the new date format in `ConnectDateUtils.kt` rather than tweaking the original utility function.
- The background of the message cards looked a bit wonky when there is a warning message on the delivery progress screen (where it meets the top of the "Progress" and "Payment" tabs, so I tweaked the background colors and added some margin to make it look better. Here are a couple of before & after photos of the issue:

### Safety story

- Besides the warning icon, made sure that the opportunity card remains unchanged on the app home screens.
- Verified that the new "View Info" and "Resume" buttons navigate the user to the correct places.
- Made sure that the new design appears like the Figma design.
- Tested the opportunity card in various different states by using different opportunities.

### QA Plan

- If possible, verify that the non-English translations on the opportunity card is correct.
- Test all the different states that the connect card could be in by starting an opportunity and seeing it through completion, ensuring that the card redesign appears correctly at all times.

### DIFF
--- a/app/src/org/commcare/android/logging/ReportingUtils.java
+++ b/app/src/org/commcare/android/logging/ReportingUtils.java
@@ -145,8 +145,8 @@ public class ReportingUtils {
             if (user.isEmpty()) {
                 return getPersonalID();
             }
-        } catch (Exception ignored) {
-        }
+            return user;
+        } catch (Exception ignored) {}
         return null;
     }
 


app/res/layout/view_job_card.xml
```diff
@@ -17,25 +17,38 @@
...
-            android:textStyle="bold"/>
+            app:layout_constraintTop_toBottomOf="@id/tv_job_title"
+            android:textStyle="bold" />
+
+        <TextView
+            android:id="@+id/connect_job_end_date_sub_heading"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_marginTop="8dp"
+            android:gravity="right"
```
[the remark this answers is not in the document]
conroy-ricketts: 23da255a24580a2987a8e6f58d2d33f71e99a414

app/src/org/commcare/connect/ConnectDateUtils.kt
```diff
@@ -20,6 +19,15 @@ object ConnectDateUtils {
...
     }
 
+    val completedJobDateFormat: DateFormat =
+        SimpleDateFormat("dd MMM, yyyy", Locale.getDefault())
+
+    fun formatDateForCompletedJob(date: Date): String {
+        synchronized(completedJobDateFormat) {
+            return completedJobDateFormat.format(date)
+        }
+    }
```
shubham1g5: I don't think there is a particular reason we should be using 2 different date formats here and instead the guiding product principle here should be _**consistency**_ .  

I think product just missed out on this inconsistency but will support us uisng the same date format across two very similar widgets.

app/src/org/commcare/fragments/connect/ConnectLearningProgressFragment.java
```diff
@@ -178,18 +179,12 @@ private void configureJobDetailsButton() {
...
+        getBinding().connectLearningButton.setOnClickListener(v -> navigateToLearnAppHome());
     }
 
     private void configureLaunchLearningButton() {
         getBinding().connectLearningButton.setText(getString(R.string.connect_learn_continue));
-        getBinding().connectLearningButton.setOnClickListener(v -> {
-            CommCareApplication.instance().closeUserSession();
-            ConnectAppUtils.INSTANCE.launchApp(requireActivity(), true, job.getLearnAppInfo().getAppId());
-        });
+        getBinding().connectLearningButton.setOnClickListener(v -> navigateToLearnAppHome());
```
OrangeAndGreen: Nit: It feels odd to me to have this function and configureGoToAssessmentButton be so similar, the only difference being the string ID used for the button text. No big deal to leave as is, but we could consider using one function that takes the string ID as input

OrangeAndGreen (approved):
Nice changes! All looks pretty good to me, left one nit but nothing blocking.

app/src/org/commcare/fragments/connect/ConnectLearningProgressFragment.java
conroy-ricketts: Oh snap, I just realized how oddly these functions are set up. These functions are all also very similar:

- `configureJobDetailsButton()`
- `configureDownloadButton()`

All these functions are doing is setting some text and/or a click listener on the same button. And I think the name `connectLearningButton` is a bit misleading too.

I'm going to leave these functions as is for now only because I don't want the changelist to get too big and I'd rather a larger refactor here later down the road

app/src/org/commcare/activities/StandardHomeActivityUIController.java
```diff
@@ -146,8 +149,12 @@ private void updateConnectJobMessage() {
 
             connectMessageCard.setCardBackgroundColor(ContextCompat.getColor(activity, backgroundColorRes));
             connectMessageCard.setVisibility(View.VISIBLE);
+            connectMessageWarningIcon.setVisibility(
+                    job.readyToTransitionToDelivery() ? View.GONE : View.VISIBLE
+            );
```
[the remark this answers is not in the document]
conroy-ricketts: The reason for not including `job.deliveryComplete()` is because we want to hide the icon only when the message looks something like this (I assume):



On the other hand, I like the idea of `isWarning`!

app/src/org/commcare/connect/ConnectDateUtils.kt
conroy-ricketts: I went ahead and did similar date formatting logic for the app home pages too for consistency with the progress pages - 4d776b18afd62e9bf4df8aae59dc91bf4dfa9652

app/src/org/commcare/activities/StandardHomeActivityUIController.java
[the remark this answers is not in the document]
shubham1g5: that makes sense to me, and fine to keep it as it is. Although I would slightly prefer for this logic to be part of the above `if-else` logic so that all the UI state calculation is part of single logic flow.

shubham1g5:
@conroy-ricketts I didn't realise there is an additional description there which is not present on the progress page cards, I was hoping we can standardize the 2 views better to make them more consistent with each other. I am not sure why we chose to show description on one card and not on another and think it might just be something that got overlooked by us and product. So think it's good to flag this to product but otherwise I am happy with the current state based on this PR.

shubham1g5 (approved):

app/src/org/commcare/activities/StandardHomeActivityUIController.java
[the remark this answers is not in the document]
conroy-ricketts: Ah I see what you mean. I agree!

app/src/org/commcare/activities/StandardHomeActivityUIController.java
```diff
@@ -100,8 +104,20 @@ private void setupConnectJobTile() {
             tvJobTitle.setText(job.getTitle());
             tvViewMore.setVisibility(View.GONE);
             tvJobDescription.setText(job.getShortDescription());
-            connectJobEndDate.setText(activity.getString(R.string.connect_learn_complete_by,
-                    ConnectDateUtils.INSTANCE.formatDate(job.getProjectEndDate())));
+
+            @StringRes int dateMessageStringRes;
+            if (job.deliveryComplete()) {
```
OrangeAndGreen: Nit: I have a slight preference for using ternary statements for this kind of thing, but not sure if we have a preference as a team. @shubham1g5 Your thoughts?

app/src/org/commcare/connect/ConnectDateUtils.kt
```diff
@@ -12,7 +11,7 @@ import java.util.Locale
 import java.util.TimeZone
 
 object ConnectDateUtils {
-    val dateFormat: DateFormat = SimpleDateFormat("MM/dd/yyyy", Locale.getDefault())
+    val dateFormat: DateFormat = SimpleDateFormat("dd MMM, yyyy", Locale.getDefault())
```
OrangeAndGreen: This has been on my mind in the past but since we're proposing changes I'll bring it up now. While this would make the date more clear than the previous code, we're still forcing an English-style display with "dd MMM" rather than using properly localized formats.
I wonder if it would be better to use `DateFormat.MEDIUM` here, which results in "dd MMM yyyy" for UK and "MMM dd, yyyy" for US as well as supporting other locales.

OrangeAndGreen (commented):
I added a couple thoughts but nothing blocking

OrangeAndGreen (approved):
Meant to approve...

app/src/org/commcare/connect/ConnectDateUtils.kt
shubham1g5: thats a good suggestion, didn't realise a constant like that exists. 

Think we might want to make that change across Connect screens and possibly good to scope that out as another ticket.

app/src/org/commcare/activities/StandardHomeActivityUIController.java
shubham1g5: I think historically in Dimagi, folks prefer to use if-else instead of ternary statements (I preferred to use ternary statements pre-dimagi but switched to if-else based on team's stylistic preference).

That said, I don't have strong thoughts on this and generally fine with either for simple assignments like these, but do think for complex statements if-else provide better readability.

app/src/org/commcare/connect/ConnectDateUtils.kt
OrangeAndGreen: Agreed this is better as a different ticket so it can be applied across pages and we can make sure nothing is negatively affected (thinking layout spacing specifically). I'll add a ticket to the backlog.

conroy-ricketts: Great callout!

shubham1g5 (approved):

PR #3475, opened 2025-12-29
Number hint not working issue

jaypanchal-13:
## Product Description
 Ticket -> https://dimagi.atlassian.net/browse/CCCT-1688

## Technical Summary
- Enhance log for failure

## Labels and Review

### DIFF
--- a/app/res/drawable/ic_connect_warning.xml
+++ b/app/res/drawable/ic_connect_warning.xml
@@ -1,0 +1,13 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="20dp"
+    android:height="20dp"
+    android:viewportWidth="20"
+    android:viewportHeight="20">
+  <group>
+    <clip-path
+        android:pathData="M0,0h20v20h-20z"/>
+    <path
+        android:pathData="M1.555,17.083L10,2.5L18.445,17.083H1.555ZM3.708,15.833H16.292L10,5L3.708,15.833ZM10,14.84C10.191,14.84 10.351,14.775 10.48,14.646C10.609,14.517 10.673,14.357 10.673,14.167C10.673,13.976 10.609,13.816 10.48,13.687C10.351,13.558 10.191,13.493 10,13.493C9.809,13.493 9.649,13.558 9.52,13.687C9.391,13.816 9.327,13.976 9.327,14.167C9.327,14.357 9.391,14.517 9.52,14.646C9.649,14.775 9.809,14.84 10,14.84ZM9.375,12.66H10.625V8.494H9.375V12.66Z"
+        android:fillColor="#1F2937"/>
+  </group>
+</vector>
--- a/app/res/layout-land/home_screen.xml
+++ b/app/res/layout-land/home_screen.xml
@@ -64,6 +64,17 @@
                     android:paddingTop="35dp"
                     android:paddingEnd="15dp">
 
+                    <ImageView
+                        android:id="@+id/ivConnectMessageWarningIcon"
+                        android:layout_width="wrap_content"
+                        android:layout_height="wrap_content"
+                        android:contentDescription="@null"
+                        android:src="@drawable/ic_connect_warning"
+                        app:layout_constraintTop_toTopOf="parent"
+                        app:layout_constraintStart_toStartOf="parent"
+                        app:layout_constraintBottom_toBottomOf="parent"
+                        android:paddingEnd="10dp"/>
+
                     <TextView
                         android:id="@+id/tvConnectMessage"
                         android:layout_width="0dp"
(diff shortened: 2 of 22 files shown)


app/src/org/commcare/utils/PhoneNumberHelper.java
```diff
@@ -134,9 +132,12 @@ public static void requestPhoneNumberHint(ActivityResultLauncher<IntentSenderReq
                         IntentSenderRequest intentSenderRequest = new IntentSenderRequest.Builder(pendingIntent).build();
                         phoneNumberHintLauncher.launch(intentSenderRequest);
                     } catch (Exception e) {
+                        FirebaseAnalyticsUtil.reportPhoneNumberHintFailure(e.getMessage());
```
shubham1g5: we don't want to log this to analytics but only with `Logger.exception` as non-fatals

app/src/org/commcare/utils/PhoneNumberHelper.java
```diff
@@ -82,31 +83,28 @@ public boolean isValidPhoneNumber(String phone) {
...
      */
     public long getNationalNumber(String phone) {
         try {
-            Phonenumber.PhoneNumber phoneNumber = phoneNumberUtil.parse(phone, null);
-            if (phoneNumberUtil.isValidNumber(phoneNumber)) {
-                return phoneNumber.getNationalNumber();
-            }
+            Phonenumber.PhoneNumber phoneNumber =
+                    phoneNumberUtil.parse(phone, "ZZ");
+            return phoneNumber.getNationalNumber();
```
shubham1g5: can we explain the two changes here and how do they solve the issue in technical description for PR.

app/src/org/commcare/utils/PhoneNumberHelper.java
```diff
@@ -82,31 +83,28 @@ public boolean isValidPhoneNumber(String phone) {
...
-            Phonenumber.PhoneNumber phoneNumber = phoneNumberUtil.parse(phone, null);
-            if (phoneNumberUtil.isValidNumber(phoneNumber)) {
-                return phoneNumber.getCountryCode();
-            }
+            Phonenumber.PhoneNumber phoneNumber =
+                    phoneNumberUtil.parse(phone, "ZZ");
+            return phoneNumber.getCountryCode();
         } catch (NumberParseException e) {
-            // Ignore
+            return -1;
```
shubham1g5: we should instead just be logging these using Logger.exception

conroy-ricketts (commented):
I have no other concerns besides what Shubham said

OrangeAndGreen (changes requested):
Generally seconding what Shubham pointed out in his comments, specifically:
1. We should keep the isValidNumber checks and just return early when the user hasn't typed a valid number yet (shouldn't report any exception)
2. When we encounter an exception trying to get the country code or national number, we should log it with Logger.exception (no need for analytics)

app/src/org/commcare/utils/PhoneNumberHelper.java
```diff
@@ -82,31 +83,28 @@ public boolean isValidPhoneNumber(String phone) {
      */
     public int getCountryCode(String phone) {
         try {
-            Phonenumber.PhoneNumber phoneNumber = phoneNumberUtil.parse(phone, null);
-            if (phoneNumberUtil.isValidNumber(phoneNumber)) {
-                return phoneNumber.getCountryCode();
-            }
+            Phonenumber.PhoneNumber phoneNumber =
+                    phoneNumberUtil.parse(phone, "ZZ");
```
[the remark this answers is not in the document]
conroy-ricketts: That makes sense to me. Can we add a short code comment here for using "ZZ" for future reference?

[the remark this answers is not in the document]
shubham1g5: @jaypanchal-13 Where is `+974` coming from in your video ? To me the issue seems to be that we are populating incorrect country code (as it should be +91) ?

> if google number hint does not contain '+' in number at that time isValidNumber returns -1.

Would adding `+` before passing this to `isValidNumber` resolve this ?


> I think the key here will be to get this working with the isValidNumber check still involved.

Agree with this.

jaypanchal-13:
> @jaypanchal-13 Sounds good for the issue described on the ticket.
> 
> Reg. the issue you discovered, it feels like we should be solving that in context of this ticket as users might be encountering that too. It feels like a simple fix to me to just add a `+` to the number when it's not present before passing it to `isValidNumber`. Are you able to confirm that ?

> @shubham1g5 + is already added at the time of getting country code. I am getting issue because I am getting +44 as country code and + is not added in sim number in setting(Every device does not have option to add and remove + in settings. I have so if I add + before number it works and if I do not add + it do not paste number in edit text) both together. Lastly adding + logic is already added while setting country code. So we are getting full number with +.

PR #3485, opened 2026-01-08
more robust view inflation checking in test

shubham1g5:
## Technical Summary

Trying to improve frequent failures on `CyclicCasesPurgeTest` on Jenkins with error -> 

````
java.lang.NullPointerException: Cannot invoke "android.widget.TextView.getText()" because the return value of "android.app.Dialog.findViewById(int)" is null
	at org.commcare.android.tests.formsave.CyclicCasesPurgeTest.testFormSaveResultingIntoCaseCycles_ShouldFail(CyclicCasesPurgeTest.java:80)
````

This happens on and off and think there is somewhere a race condition with Robolectric that is causing the test to assert before view gets inflated. I am not able to repro it locally, but the change should be harmless to make. 

## Labels and Review

### DIFF
--- a/app/unit-tests/src/org/commcare/android/tests/formsave/CyclicCasesPurgeTest.java
+++ b/app/unit-tests/src/org/commcare/android/tests/formsave/CyclicCasesPurgeTest.java
@@ -28,6 +28,8 @@ import org.robolectric.annotation.Config;
 import org.robolectric.shadows.ShadowDialog;
 import org.robolectric.shadows.ShadowLooper;
 
+import java.util.concurrent.TimeUnit;
+
 /**
  * Test Scenarios related to cyclic case relationships encountered during the case purge process.
  * These tests are only relevant when the case purge process is enabled with property `cc-auto-purge`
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by shubham1g5** [comment]

File: app/unit-tests/src/org/commcare/android/tests/formsave/CyclicCasesPurgeTest.java [hunk 1]
```diff
@@ -63,10 +63,14 @@ public void testFormSaveResultingIntoCaseCycles_ShouldFail() {
...
         // Retry mechanism to handle async timing issues
         Dialog latestDialog = null;
+        TextView messageView = null;
         for (int attempt = 0; attempt < 5; attempt++) {
             latestDialog = ShadowDialog.getLatestDialog();
-            if (latestDialog != null) {
-                break;
+            if (latestDialog != null && latestDialog.getWindow() != null) {
+                messageView = latestDialog.getWindow().getDecorView().findViewById(R.id.dialog_message);
+                if (messageView != null) {
```
OrangeAndGreen: Not sure what the speed of this loop is, but I wonder if it could be useful to put a small thread sleep in the else case here so we give a slight pause before trying again?

conroy-ricketts (approved):

PR #3487, opened 2026-01-13
Discard stale locations based on an App Setting

shubham1g5:
## Product Summary

https://dimagi.atlassian.net/browse/CCCT-2051

The PR adds a new app setting behind key `cc-discard-stale-locations` which apps can implement to discard any stale locations from GPS. This is an experimental setting at the moment and will be applied using the [custom properties](https://dimagi.atlassian.net/wiki/spaces/GS/pages/2146603845/Allow+users+to+add+arbitrary+custom+properties+to+their+application) section on the app. 

We are checking whether the location is older than 2 minutes old and discarding if so. The idea is for us to see whether adding this setting results in these apps getting more accurate location capture in the forms. 

## Technical Summary

We are also adding a non-fatal log everytime we are encountering the stale location to see how often scenarios like that happen. The check is not behind the custom app preperty as we want to get this data across our user base. Our current understanding is this is very rare and we want to validate this with the log. 

There is an addition log added "Location Received" which is meant to signify whether there was a location update before the stale update to confirm the hypothesis that the stale locations are always sent as the first location and not thereafter. 

## Safety Assurance

The functional change is behind a custom app property that will be tested by QA before we release. 

## Labels and Review

### DIFF
--- a/app/src/org/commcare/gis/EntityMapUtils.kt
+++ b/app/src/org/commcare/gis/EntityMapUtils.kt
@@ -19,7 +19,6 @@ import java.util.Vector
 import javax.annotation.Nullable
 
 object EntityMapUtils {
-
     // Field template form name constants
     private const val TEMPLATE_FORM_ADDRESS = "address"
     private const val TEMPLATE_FORM_GEO_BOUNDARY = "geo_boundary"
@@ -28,7 +27,10 @@ object EntityMapUtils {
     private const val TEMPLATE_FORM_GEO_POINTS_COLORS = "geo_points_colors_hex"
 
     @JvmStatic
-    private fun getHeaderText(detail: Detail, templateFormName: String): String? {
+    private fun getHeaderText(
+        detail: Detail,
+        templateFormName: String,
+    ): String? {
         for (i in 0 until detail.templateForms.size) {
             if (templateFormName == detail.templateForms[i]) {
                 return detail.fields[i].header.evaluate()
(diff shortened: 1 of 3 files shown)


shubham1g5:
@conroy-ricketts @OrangeAndGreen this work evolved a bit more into an functional change I am planning to deploy as hotfix soon as it's becoming critical for Connect to get this data asap. I haved added more details into PR description. 


> We could also consider logging the location age in analytics in a similar fashion, which would allow us to get a better view of typical location ages rather than just a flag of those greater than 2 minutes. It would then be trivial to query for location ages >2 minutes or any other threshold

Think at this moment, we don't really care how old are these locations and more interested in how often this happens. I also think it's very likely that we will get a really big spectrum of these values based on what was the last time user got a location fix on their device.

shubham1g5:
> I think the difference is between logging:

Was the first location received stale when the user entered the page?
OR
Was the final location stale when the page finished?

I could not find a easy solution here other than logging [another non-fatal](https://github.com/dimagi/commcare-android/pull/3487/changes/acc20406209a314dd9234d811e209e3f2ce113c4) for when we are saving this

conroy-ricketts (dismissed):

OrangeAndGreen (dismissed):

shubham1g5:
I am not able to satisfy KTlint on this PR, think the super-linter might need an upgrade to count for latest ktlint version.

PR #3490, opened 2026-01-14
Add perf monitoring trace for file encryption time

avazirna:
## Product Description
This PR is the first in a series related to the retirement of `CommCareSessionService`. One of the options is to encrypt files using Android keystore facilities. To ensure that this change won't negatively impact CommCare performance, we decided to start measuring files encryption times to better assess this impact.

Ticket: https://dimagi.atlassian.net/browse/SAAS-19065

### Safety story
This was designed to be a null-safe operation and has no impact to the user.

## Labels and Review

### DIFF
--- a/app/src/org/commcare/tasks/SaveToDiskTask.java
+++ b/app/src/org/commcare/tasks/SaveToDiskTask.java
@@ -1,11 +1,14 @@
 package org.commcare.tasks;
 
+import com.google.firebase.perf.metrics.Trace;
+
 import org.commcare.CommCareApplication;
 import org.commcare.activities.FormEntryActivity;
 import org.commcare.activities.components.ImageCaptureProcessing;
 import org.commcare.android.database.app.models.FormDefRecord;
 import org.commcare.android.database.user.models.FormRecord;
 import org.commcare.android.logging.ForceCloseLogger;
+import org.commcare.google.services.analytics.CCPerfMonitoring;
 import org.commcare.interfaces.FormSavedListener;
 import org.commcare.logging.XPathErrorLogger;
 import org.commcare.models.database.SqlStorage;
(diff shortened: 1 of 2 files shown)


OrangeAndGreen (approved):

app/src/org/commcare/models/database/HybridFileBackedSqlStorage.java
```diff
@@ -328,6 +335,8 @@ protected byte[] generateKeyAndAdd(ContentValues contentValues) {
 
     private void writeStreamToFile(ByteArrayOutputStream bos, String filename,
                                    byte[] key) throws IOException {
+        Trace trace = CCPerfMonitoring.INSTANCE.startTracing(CCPerfMonitoring.TRACE_FILE_ENCRYPTION_TIME);
```
shubham1g5: do we need this given we are putting the same tracing in `encryptFile` ? If so, should we try to distinguish between the 2 traces ?

app/src/org/commcare/tasks/SaveToDiskTask.java
```diff
@@ -257,11 +264,21 @@ private void exportData(boolean markCompleted)
     }
 
     private void writeXmlToStream(ByteArrayPayload payload, OutputStream output) throws IOException {
+        Trace trace = CCPerfMonitoring.INSTANCE.startTracing(CCPerfMonitoring.TRACE_FILE_ENCRYPTION_TIME);
```
shubham1g5: same comment as above to distinguish the traces for different code components with different Trace labels.

shubham1g5 (changes requested):
Are you able to test that the added `tracing` itself doesn't introduce a performance downside on low end phones ? Given this tracking is in critical code components, think we should be very careful here to test the performance downsides of added tracing here.

avazirna:
> Are you able to test that the added `tracing` itself doesn't introduce a performance downside on low end phones ? Given this tracking is in critical code components, think we should be very careful here to test the performance downsides of added tracing here.

I will request QA to include this in the testing.

PR #3497, opened 2026-01-16
CCCT-2035 Job Card Overlapping UI

conroy-ricketts:
## Product Description

This one is pretty straightforward from the ticket - I constrained the buttons to the bottom of delivery hours and added a bit of margin.

I might update this PR if Product wants me to adjust the margin.

Here is a before example:

And here is an after image:

If delivery hours are not included, this is what the UI looks like with these changes:

### Safety story

I tested these changes on different pages with **_and_** without delivery hours to make sure the UI looks solid.

### QA Plan

We should test to ensure that the delivery hours is no longer being overlapped in the UI.

### DIFF
--- a/app/src/org/commcare/utils/ConnectivityStatus.java
+++ b/app/src/org/commcare/utils/ConnectivityStatus.java
@@ -1,11 +1,24 @@
 package org.commcare.utils;
 
+import android.Manifest;
 import android.content.Context;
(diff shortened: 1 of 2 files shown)


conroy-ricketts:
I tweaked the margin and updated the screenshots a moment ago

PR #3498, opened 2026-01-16
CCCT-2017 Release Toggles Parser

conroy-ricketts:
## Technical Summary

I implemented a new network parser for the release toggles feature that utilizes the new `/toggles` endpoint while storing data in the Connect DB.

Note: I chose not to handle a `JSONException` directly inside the parser because that is already being handled [here](https://github.com/dimagi/commcare-android/blob/d4ad4c8dbfba82982890fb59c3ba11b600b486e5/app/src/org/commcare/connect/network/base/BaseApiHandler.kt#L126) in `BaseApiHandler`.

### Safety story

- Verified that calling the new API works by going to a random fragment (in my case, the `ConnectJobsListsFragment`) and calling the new API function like so:

```
new PersonalIdApiHandler<List<ConnectReleaseToggleRecord>>(false, this) {  
  
    @Override  
    public void onFailure(@NonNull PersonalIdOrConnectApiErrorCodes errorCode, @Nullable Throwable t) { }  
  
    @Override  
    public void onSuccess(List<ConnectReleaseToggleRecord> data) { }  

}.getReleaseToggles(requireContext(), user.getUserId(), user.getPassword());
```

- Verified that I received a successful API response (albeit the release toggles that came from server were empty because they are not configured for me yet).
- Verified that the parser correctly generates a list of `ConnectReleaseToggleRecord`'s from json by hardcoding some fake release toggle JSON that I fed to the parse function while debugging the code.
  - Also verified that the data is being stored correctly in the Connect DB by setting a debug point and verifying that I could retrieve my fake toggles from the database after the parser ran. Here is the exact expression I used from inside `ConnectAppDatabaseUtil.storeReleaseToggles()` for reference: 
  `toggleStorage.getRecordsForValues(new String[]{}, new Object[]{})` 

### Automated test coverage

I created a new unit test class to pair with the new parser, and I verified that they all run successfully.

### QA Plan

No action needed from QA on this ticket. I'm sure this will all be tested thoroughly via other tickets.

### DIFF
--- a/app/src/org/commcare/connect/database/ConnectAppDatabaseUtil.java
+++ b/app/src/org/commcare/connect/database/ConnectAppDatabaseUtil.java
@@ -82,13 +82,20 @@ public class ConnectAppDatabaseUtil {
             Context context,
             List<ConnectReleaseToggleRecord> toggles
     ) {
-        SqlStorage<ConnectReleaseToggleRecord> toggleStorage =
-                ConnectDatabaseHelper.getConnectStorage(context, ConnectReleaseToggleRecord.class);
+        try {
+            SqlStorage<ConnectReleaseToggleRecord> toggleStorage =
+                    ConnectDatabaseHelper.getConnectStorage(context, ConnectReleaseToggleRecord.class);
+            ConnectDatabaseHelper.connectDatabase.beginTransaction();
 
-        toggleStorage.removeAll();
+            toggleStorage.removeAll();
 
-        for (ConnectReleaseToggleRecord toggle : toggles) {
-            toggleStorage.write(toggle);
+            for (ConnectReleaseToggleRecord toggle : toggles) {
+                toggleStorage.write(toggle);
+            }
+
+            ConnectDatabaseHelper.connectDatabase.setTransactionSuccessful();
+        }  finally {
+            ConnectDatabaseHelper.connectDatabase.endTransaction();
         }
     }
 }


conroy-ricketts:
@shubham1g5 @OrangeAndGreen @jaypanchal-13 @Jignesh-dimagi  

I was just now able to verify the new unit tests run successfully from my terminal. So this is ready for a full review now 👍

I'm still investigating why all unit tests are failing for me in Android Studio though, but I'm sure it's just something wrong with my local environment.

shubham1g5 (changes requested):

app/unit-tests/src/org/commcare/connect/network/connectId/parser/ConnectReleaseTogglesParserTest.kt
```diff
@@ -0,0 +1,179 @@
...
+                }
+            }
+            """.trimIndent()
+
+        val expectedDate = DateUtils.parseDateTime(mockDateString)
+
+        val inputStream = ByteArrayInputStream(mockJson.toByteArray())
+
+        // Mock the DB storage call so it doesn't try to access a real DB.
+        every { ConnectAppDatabaseUtil.storeReleaseToggles(any(), any()) } returns Unit
```
Jignesh-dimagi: @conroy-ricketts Minor: This mocking of the DB storage call can be moved to `setUp` to remove repeating code.

Jignesh-dimagi: @conroy-ricketts LGTM, only minor comment

Jignesh-dimagi (approved):

PR #3509, opened 2026-01-22
Add support to sharing forms with media via WI-FI Direct

avazirna:
## Product Description
This PR addresses a limitation with the WI-FI Direct feature, in which form submissions would fail when they included multimedia files. The cause is related with the source device sharing encrypted files, which would then cause the recipient device to attempt to decrypt them with a different user's symmetric key, resulting in a failure.

Ticket: https://dimagi.atlassian.net/browse/SAAS-16844

## Technical Summary
The solution was to decrypt the files in the source device before sharing.

### Safety story
Successfully tested locally.

## Labels and Review

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,19 +93,9 @@ and view the results from the output file generated.
 
 ### Run unit-tests from Android Studio
 
-Create a new Android Studio JUnit Build configuration using the following steps.
+Unit tests can be ran from Android Studio via an auto-configured Gradle task by right-clicking on the test folder in the file explorer and then clicking on "Run 'org.commcare.tests'".
 
-- Click *Run -> Edit Configruations* and create a new JUnit configuration by pressing the green plus button.
-- Set *Name* to "commcare Android test suite"
-- Set *Test kind* to "All in directory"
-- set *Directory* to `/absolute/path/to/commcare-android/app/unit-tests/src/`
-- Right click on this directory and click the "Create 'All Tests'" option that should be listed more than half-way down the list.
-- Set *VM options* to `-ea -noverify`
-- Set *Working directory* to `/absolute/path/to/commcare-android/app/`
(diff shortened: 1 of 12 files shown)


avazirna:
> This PR beats the purpose of doing encryption in the first place as anyone will be able to share the file to a different device and get rid of encryption. Instead, we should just provide an app setting for projects to turn off the file encryption and update our wifi direct docs to mention the same

Yes, I battled with this too. The current workflow involves: (1) decrypting the files (2) transferring them to the recipient device unencrypted, and (3) having the recipient device upload the files unencrypted, which is how we do it now during normal form submissions. I've also realized that form submissions are currently sent and stored unencrypted, so I applied the same approach to multimedia files.

My concern is that an app setting wouldn't prevent loss of data which is what's happening now. The multimedia file is being transmitted in encrypted form and cannot be decrypted on the receiving device, causing the form submission to fail.  At that point, the file has already been deleted from the originating device. The app setting would work only if it were already set beforehand.

shubham1g5:
>  I've also realized that form submissions are currently sent and stored unencrypted, so I applied the same approach to multimedia files.

Form submissions happens over https so they are always encrypted at network layer. The encryption layer adds security at storage layer instead which is what we are compromising with this change. 

> The app setting would work only if it were already set beforehand.

Agree and we should make it clear in the docs that the wifi-direct will only work if that setting is enabled. 

And to solve the current issue, we should ask these devices to upload their data with a network connection as a one time thing.

shubham1g5:
@avazirna My sense is QA currently test for wifi direct during releases and this bug would have broke wifi direct for all projects, as such I am finding it suprising that QA missed this bug during release testing. Do you know if there is a caveat here that causes this bug to only happen in certain cases and not all ?

PR #3518, opened 2026-01-29
Not hiding app spinner when Login page loads with a preset app ID (to…

OrangeAndGreen:
https://dimagi.atlassian.net/browse/CCCT-2095

## Product Description
Fixes a bug where the app spinner on the Login page would sometimes be hidden when it shouldn't be.
The spinner is no longer hidden for pre-loaded apps.

## Technical Summary
The bug was related to when we decide to show/hide the app spinner. Generally, the logic is to show the spinner if there's more than one app installed. But when auto-logging in to Connect apps, we also added code to hide the spinner as an early attempt at "locking the page". However, with the addition of the side menu some of the logic relating to pre-loaded apps changed and caused the page to always configure that part of the UI with the spinner hidden.

Since the Login page is locked down during Login anyway (via modal dialogs while logging in), a safe fix seems to be just not hiding the spinner whenever there are multiple apps installed.

## Feature Flag
None

### Safety story
I tested various scenarios related to this part of the code (app switching, pre-loading, etc.):
* Initial page view
* Switching apps via the spinner
* Switching apps via the side menu
* Auto-logging into Connect apps

### Automated test coverage
None

### QA Plan
A video shared [here](https://dimagi.slack.com/archives/C0B1YKKU0/p1769708345886589) shows steps to reproduce the bug.
Verify that the above is fixed, and also test various app switching and login scenarios.


shubham1g5 (approved):

conroy-ricketts (approved):
Love the simple fixes!

PR #3522, opened 2026-02-02
CCCT-2053 - Store UUID into Connect Models 

Jignesh-dimagi:
## Technical Summary
https://dimagi.atlassian.net/browse/CCCT-2053

For all the connect modals, it is now required to have UUIDs instead of int ID keys. So now UUIDs are also stored in the respective connect modals, which are:

- ConnectJobDeliveryRecord

- ConnectJobLearningRecord

- ConnectJobPaymentRecord

- ConnectJobAssessmentRecord

- ConnectAppRecord

- ConnectLearnModuleSummaryRecord

- ConnectPaymentUnitRecord

Key factors taken into consideration while migrating to the new UUID:

- The default value of the UUID should be the ID value.

- If the response has a new UUID value, store that value in the DB; else, the UUID should default to ID.

- All of the connect modals have reference to `ConnectJobRecord`"ID", and now it is required to have reference to UUID also, so a separate reference to UUID was added to all connect modals. This separate column for UUID was required, as ID is an int key and UUID is a string.

- `ConnectJobLearningRecord` is not using its own ID anywhere, but still UUID has been added for uniformity of code across Connect modals.

- The names of the UUIDs field parameters are not explicitly mentioned in the mobile tech doc or in the ticket, so they are coded using [the web tech doc](https://docs.google.com/document/d/1Xpv5dqqaEKy4poH2aOj9utpSTjrErqSSVN9XgG-VL4U/edit?tab=t.0) naming convention that it will follow: x_id field, where x represents the model name. The server team needs to confirm the parameter name for UUIDs. 

### Safety story

- The application has been tested for DB migration for:

  - An already older version of the app is installed on the device and then upgraded to this new version. Application was able to fetch all details from DB.

  - Installing this new version directly on the device and testing the application in general.

### QA Plan
- Need to test the application in general after upgrading to this version and also from a fresh install.
- The version upgrade needs to happen from 2.57.0 to the latest version with a fully set up Personal ID login with opportunities and on upgrade, we should not need to re-login into Personal ID again and should see all the Connect related data as it was before the upgrade, this will need to be tested on all Connect screens.  

## Labels and Review

### DIFF
--- a/app/src/org/commcare/models/database/connect/ConnectDatabaseUpgrader.java
+++ b/app/src/org/commcare/models/database/connect/ConnectDatabaseUpgrader.java
@@ -681,14 +681,14 @@ public class ConnectDatabaseUpgrader {
     }
 
     private void upgradeEighteenNineteen(IDatabase db) {
-        addTableForNewModel(db, PushNotificationRecord.STORAGE_KEY, new PushNotificationRecord());
+        addTableForNewModel(db, PushNotificationRecordV21.STORAGE_KEY, new PushNotificationRecordV21());
     }
 
     private void upgradeNineteenTwenty(IDatabase db) {
         db.beginTransaction();
         try {
-            SqlStorage.dropTable(db, PushNotificationRecord.STORAGE_KEY);
-            addTableForNewModel(db, PushNotificationRecord.STORAGE_KEY, new PushNotificationRecord());
+            SqlStorage.dropTable(db, PushNotificationRecordV21.STORAGE_KEY);
+            addTableForNewModel(db, PushNotificationRecordV21.STORAGE_KEY, new PushNotificationRecordV21());
             db.setTransactionSuccessful();
         } finally {
             db.endTransaction();
--- a/app/src/org/commcare/android/database/connect/models/PushNotificationRecordV21.kt
+++ b/app/src/org/commcare/android/database/connect/models/PushNotificationRecordV21.kt
@@ -77,6 +77,6 @@ class PushNotificationRecordV21 :
     var acknowledged: Boolean = false
 
     companion object {
-        const val STORAGE_KEY = "push_notification_history"
+        const val STORAGE_KEY = PushNotificationRecord.STORAGE_KEY
     }
 }


app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
```diff
@@ -196,6 +202,10 @@ public static ConnectJobRecord fromJson(JSONObject json) throws JSONException {
         ConnectJobRecord job = new ConnectJobRecord();
 
         job.jobId = json.getInt(META_JOB_ID);
+        job.jobUUID = json.optString(META_JOB_UUID, "");
+        if (job.jobUUID.isEmpty()) {
+            job.jobUUID = String.valueOf(job.jobId);
+        }
```
conroy-ricketts: Nit: suggest using the string value of job ID as the fallback to `optString()` instead of an empty string to simplify code

app/src/org/commcare/models/database/connect/ConnectDatabaseUpgrader.java
```diff
@@ -685,6 +698,247 @@ private void upgradeTwentyTwentyOne(IDatabase db) {
...
+        for (Persistable r : oldStorage) {
+            ConnectJobAssessmentRecordV21 oldRecord = (ConnectJobAssessmentRecordV21)r;
+            ConnectJobAssessmentRecord newRecord = ConnectJobAssessmentRecordV21.Companion.fromV21(oldRecord);
+            newRecord.setID(oldRecord.getID());
+            newStorage.write(newRecord);
+        }
+    }
+
+    private void upgradeConnectJobPaymentRecordToV22(IDatabase db) {
+
```
conroy-ricketts: Nit: random blank line

```suggestion
```

app/src/org/commcare/models/database/connect/ConnectDatabaseUpgrader.java
```diff
@@ -685,6 +698,247 @@ private void upgradeTwentyTwentyOne(IDatabase db) {
...
+        for (Persistable r : oldStorage) {
+            ConnectJobPaymentRecordV21 oldRecord = (ConnectJobPaymentRecordV21)r;
+            ConnectJobPaymentRecord newRecord = ConnectJobPaymentRecordV21.Companion.fromV21(oldRecord);
+            newRecord.setID(oldRecord.getID());
+            newStorage.write(newRecord);
+        }
+    }
+
+    private void upgradeConnectAppRecordToV22(IDatabase db) {
+
```
conroy-ricketts: Nit: random blank line

```suggestion
```

app/src/org/commcare/android/database/connect/models/ConnectPaymentUnitRecord.java
```diff
@@ -50,16 +52,34 @@ public class ConnectPaymentUnitRecord extends Persisted implements Serializable
...
+
+            if (job.getJobUUID().isEmpty()) {
+                paymentUnit.jobUUID = Integer.toString(job.getJobId());
+            } else {
+                paymentUnit.jobUUID = job.getJobUUID();
+            }
+
             paymentUnit.unitId = json.getInt(META_ID);
+            String unitUUID = json.optString(META_PAYMENT_UNIT_UUID, "");
+            paymentUnit.unitUUID = unitUUID.isEmpty() ? Integer.toString(paymentUnit.unitId) : unitUUID;
```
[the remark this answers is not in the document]
Jignesh-dimagi: Same. I would have done this but wanted to make sure that the value of META_PAYMENT_UNIT_UUID should not be empty.

Jignesh-dimagi:
@shubham1g5 What will be the best way to get confirmation from web team about the UUID field name? I think we should take this to tech doc and web team can answer there? We need confirmation for the following:

META_DELIVERY_UUID
META_LEARNING_UUID
META_PAYMENT_UUID
META_JOB_UUID
META_PAYMENT_UNIT_UUID

app/src/org/commcare/android/database/connect/models/ConnectJobPaymentRecord.java
```diff
@@ -74,13 +84,21 @@ public static ConnectJobPaymentRecord fromV3(ConnectJobPaymentRecordV3 oldRecord
...
-        payment.jobId = jobId;
+        payment.jobId = job.getJobId();
+        if (job.getJobUUID().isEmpty()) {
+            payment.jobUUID = Integer.toString(job.getJobId());
+        } else {
+            payment.jobUUID = job.getJobUUID();
+        }
+        payment.paymentId = json.getString("id");
+        String paymentUUID = json.optString(META_PAYMENT_UUID, "");
+        payment.paymentUUID = paymentUUID.isEmpty() ? payment.paymentId : paymentUUID;
```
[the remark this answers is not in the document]
conroy-ricketts: I'm not sure I understand. I'm referring to doing something like this:

`payment.paymentUUID = json.optString(META_PAYMENT_UUID, payment.paymentId);`

Can you elaborate what you mean?

app/src/org/commcare/android/database/connect/models/ConnectJobPaymentRecord.java
```diff
@@ -93,7 +93,7 @@ public static ConnectJobPaymentRecord fromJson(JSONObject json, ConnectJobRecord
             payment.jobUUID = job.getJobUUID();
         }
         payment.paymentId = json.getString("id");
-        String paymentUUID = json.optString(META_PAYMENT_UUID, "");
+        String paymentUUID = json.optString(META_PAYMENT_ID, "");
```
conroy-ricketts: Confirming that this should be `META_PAYMENT_ID` and not `META_PAYMENT_UUID`?

app/src/org/commcare/models/database/connect/ConnectDatabaseUpgrader.java
```diff
@@ -685,6 +698,237 @@ private void upgradeTwentyTwentyOne(IDatabase db) {
...
+                new ConcreteAndroidDbHelper(c, db));
+
+        SqlStorage<Persistable> newStorage = new SqlStorage<>(
+                ConnectJobRecord.STORAGE_KEY,
+                ConnectJobRecord.class,
+                new ConcreteAndroidDbHelper(c, db));
+
+        for (Persistable r : oldStorage) {
+            ConnectJobRecordV21 oldRecord = (ConnectJobRecordV21)r;
+            ConnectJobRecord newRecord = ConnectJobRecordV21.Companion.fromV21(oldRecord);
```
OrangeAndGreen: The latest code breaks from the existing pattern and won't correctly upgrade the database in all situations.

First, the upgrade methods have traditionally been added to the new version of the class.
So in this case, the newest upgrade function should be ConnectJobRecord.fromV21 instead of ConnectJobRecordV21.fromV21.
Aside from the benefit of simply maintaining a consistent pattern, I think this convention makes sense: new classes know how to update themselves from old versions (old versions don't have to include code referencing the new version).

The broken part is that ConnectDatabaseUpgrader.upgradeTenEleven needs to be updated so it upgrades ConnectJobRecordV10 to ConnectJobRecordV21. Currently it skips V21 and upgrades directly to the modern version, which will cause a crash a short time later when trying to upgrade to V22 (upgradeConnectJobRecordtoV22 will encounter the modern version of the class when it expects V21).

The key is that for devices that need to upgrade through several schemas in a row (i.e. they haven't run in a long time), each upgrade needs to only do the incremental conversion so that the next upgrade encounters the correct version of the data and can perform the next conversion. The final conversion is the one that results in the modern version.

app/src/org/commcare/android/database/connect/models/ConnectLearnModuleSummaryRecordV21.kt
```diff
@@ -0,0 +1,63 @@
...
+    @MetaField(META_INDEX)
+    private var moduleIndex = 0
+
+    @Persisting(7)
+    private var lastUpdate: Date? = null
+
+    companion object {
+        const val STORAGE_KEY = ConnectLearnModuleSummaryRecord.STORAGE_KEY
+
+        fun fromV21(connectLearnModuleSummaryRecordV21: ConnectLearnModuleSummaryRecordV21): ConnectLearnModuleSummaryRecord {
```
OrangeAndGreen: This function should move to the modern ConnectLearnModuleSummaryRecord (new version knows how to upgrade from old version)

app/src/org/commcare/models/database/connect/ConnectDatabaseUpgrader.java
OrangeAndGreen: This upgrade needs the new storage to be ConnectJobDeliveryRecordV21, not the modern version yet.

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
```diff
@@ -196,6 +202,10 @@ public static ConnectJobRecord fromJson(JSONObject json) throws JSONException {
         ConnectJobRecord job = new ConnectJobRecord();
 
         job.jobId = json.getInt(META_JOB_ID);
+        job.jobUUID = json.optString(META_JOB_UUID, "");
+        if (job.jobUUID.isEmpty()) {
```
shubham1g5: this can't be empty in payload and we should enforce this.

app/src/org/commcare/models/database/connect/DatabaseConnectOpenHelper.java
```diff
@@ -63,7 +63,7 @@ public class DatabaseConnectOpenHelper extends SQLiteOpenHelper {
      * V.20 - Added acknowledged column in push_notification_history
      * V.21 - Added ConnectReleaseToggleRecord table
```
shubham1g5: we should also update the changelog here for v22.

app/src/org/commcare/models/database/connect/ConnectDatabaseUpgrader.java
```diff
@@ -520,32 +534,32 @@ private void upgradeTenEleven(IDatabase db) {
...
+                ConnectJobRecordV21 newRecord = ConnectJobRecordV21.Companion.fromV10(oldRecord);
                 //set this new record to have same ID as the old one
                 newRecord.setID(oldRecord.getID());
                 newStorage.write(newRecord);
             }
 
             //Next, migrate the old ConnectJobDeliveryRecord in storage to the new version
             oldStorage = new SqlStorage<>(
-                    ConnectJobDeliveryRecord.STORAGE_KEY,
+                    ConnectJobDeliveryRecordV2.STORAGE_KEY,
```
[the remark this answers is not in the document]
Jignesh-dimagi: Yeah I have seen both above but seems like should not attempt it now.

shubham1g5 (changes requested):
One blocking comment on defaulting UUID to old id, rest seems all good. 

Can we also test upgrading from an old CommCare version like 2.50 to this version to make sure all intermediate migrations are not broken as well.

conroy-ricketts (dismissed):

app/src/org/commcare/models/database/connect/ConnectDatabaseUpgrader.java
```diff
@@ -685,6 +699,256 @@ private void upgradeTwentyTwentyOne(IDatabase db) {
...
+
+        SqlStorage<Persistable> newStorage = new SqlStorage<>(
+                ConnectJobRecord.STORAGE_KEY,
+                ConnectJobRecord.class,
+                new ConcreteAndroidDbHelper(c, db));
+
+        for (Persistable r : oldStorage) {
+            ConnectJobRecordV21 oldRecord = (ConnectJobRecordV21)r;
+            ConnectJobRecord newRecord = ConnectJobRecord.fromV21(oldRecord);
+            newRecord.setID(oldRecord.getID());
```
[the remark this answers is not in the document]
Jignesh-dimagi: This is the internal record ID for DB, mostly used to determine if a record exists or not. Sorry, how does adding the UUID new field help here?

shubham1g5 (approved):

OrangeAndGreen (dismissed):

PR #3525, opened 2026-02-03
Fix background network calls on Android 15 and higher

avazirna:
## Product Description
This PR addresses an issue where network requests triggered by `AsyncTasks` are interrupted when CommCare goes to the background on devices running Android 15 and higher. This behavior is due the [background network access restrictions](https://developer.android.com/about/versions/15/behavior-changes-all#background-network-access) introduced in that version. As a result, the app installation feature was impacted as demonstrated in the video below:
<video src="https://github.com/user-attachments/assets/09ba1000-7a86-49f6-a748-15c9ec67b86a" width="300px"></video>

## Technical Summary
After some tests, it was noticed that the restrictions are not applied when `CommCareSessionService` is running, and that the affected features are those accessible when the service is not running. Based on this, the chosen approach was to implement a service, `NetworkNotificationService`, that could run in parallel to those features.

It's important to note that this is a temporary solution. The scope of these restrictions seem to be expanding, and as a result this approach mayb be rejected by the Google Play Store. Additionally, `AsyncTasks` were deprecated in Android 11, so the long term goal is to replace them with components align with modern Android development best practices. 

**Notes for reviewers:**
- Tasks are now being tracked using their TAG, this motivated changing the TAG to a _non-static_ field. It doesn't seem that there was a particular reason for it to be _static_, but curious if there are any other reasons I'm not considering.

### Safety story
This was successfully tested locally.

### QA Plan
I plan to send this to QA for regression testing.

## Labels and Review

### DIFF
--- a/app/res/values-es/strings.xml
+++ b/app/res/values-es/strings.xml
@@ -50,12 +50,11 @@
     <string name="install_from_list_prompt">Inicia sesión como usuario móvil o web para ver una lista de las aplicaciones que puedes instalar</string>
     <string name="legacy_failure">CommCare no admite la actualización desde la versión anterior. Para resolver este problema, desinstala CommCare de tu dispositivo y luego vuelve a instalarlo.</string>
     <string name="legacy_failure_title">Actualización de CommCare incompatible</string>
-    <string name="notificationtitle">Conectado a CommCare</string>
-    <string name="expirenotification">Conexión de CommCare caducada</string>
+    <string name="session_notification_title">Conectado a CommCare</string>
+    <string name="session_expire_notification">Conexión de CommCare caducada</string>
     <string name="submission_notification_title">Envío de datos</string>
     <string name="submission_logs_title">Envío de registros de CommCare</string>
     <string name="about_dialog">%s\n\n_Copyright 2016, Dimagi inc, Todos los derechos reservados_ \n\n---------------- \n\n__Agradecimientos:__ \n\n%sAlgunas imágenes de CommCare se utilizan amablemente bajo licencia Creative Commons de la colección thenounproject.com:\n\n* Símbolo «Puerta» por Tak Imoto \n* Símbolo «Lápiz» por Creologica \n* Símbolo «Cuaderno» por Carlo \n* Símbolo «Antena» por Dima \n* Símbolo «Marcador de mapa» por P.J. Onori \n* Lista por Landan Lloyd \n* Pantalla rota por Sam Martin de Noun Project \n\n\n\n"</string>
-    <string name="notification_message_title">Notificación de CommCare</string>
     <string name="hello_world">Introduce tus problemas a continuación</string>
     <string name="menu_settings">Ajustes</string>
     <string name="permission_content_provider_label">Leer datos seguros de CommCare</string>
@@ -183,7 +182,6 @@
     <string name="dependency_missing_dialog_message_plural">CommCare necesita que las aplicaciones a continuación estén instaladas en tu dispositivo antes de poder continuar.</string>
     <string name="dependency_missing_dialog_go_to_store">Instalar desde la tienda</string>
     <string name="dependency_missing_dialog_playstore_not_found">No se encontró Play Store en tu dispositivo</string>
-    <string name="fcm_notification">Notificación de FCM</string>
     <string name="fcm_default_notification_channel">canal-de-notificación-notificaciones-push</string>
     <string name="personalid_login_via">Vía PersonalId (o toca aquí)</string>
     <string name="personalid_welcome_user">¡Bienvenido %s!</string>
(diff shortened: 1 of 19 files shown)


app/src/org/commcare/tasks/templates/CommCareTask.java
```diff
@@ -93,6 +100,10 @@ protected void onPostExecute(Result result) {
                 connector.stopTaskTransition(taskId);
             }
         }
+
+        if (NetworkNotificationService.Companion.isServiceRunning()) {
+            CommCareApplication.instance().stopService(getNotificationIntent());
+        }
```
shubham1g5: This would not work anytime we have more than one Async tasks in parallel as any task terminating will also stop the service for the in progress task.

avazirna: @shubham1g5 I have opted to store the running tasks IDs in a List and call `stopSelf()` once the list is empty - 508fa39f8d5b062f29a6e19499b7976530720065

app/src/org/commcare/tasks/ResourceEngineTask.java
```diff
@@ -69,6 +69,8 @@ public ResourceEngineTask(CommCareApp app, int taskId, boolean shouldSleep, bool
         this.shouldSleep = shouldSleep;
         this.reinstall = reinstall;
         TAG = ResourceEngineTask.class.getSimpleName();
+        this.runNotificationService = true;
+        this.notificationServiceProgressTextKey = "network.notification.service.downloading.files";
```
shubham1g5: nitL think these 2 should be exposed with  over-ridable methods in `CommCareTask` like - 

````
fun requireNetworkNotification(): boolean
fun getNetworkNotificationKey(): String
````

File: app/src/org/commcare/services/NetworkNotificationService.kt [hunk 3]
```diff
@@ -0,0 +1,137 @@
...
+            .setContentTitle(Localization.get(notificationTitleKey))
+            .setOnlyAlertOnce(true)
+            .setSmallIcon(R.drawable.commcare_actionbar_logo)
+            .setContentIntent(pendingIntent)
+            .setOngoing(true)
+            .build()
+    }
+
+    @RequiresApi(Build.VERSION_CODES.N)
+    override fun onDestroy() {
```
shubham1g5: what's stopping this from getting called on < N devices ?

app/src/org/commcare/services/NetworkNotificationService.kt
```diff
@@ -0,0 +1,137 @@
...
+                    ),
+                )
+            }
+            STOP_NOTIFICATION_ACTION -> {
+                intent.getIntExtra(TASK_ID_INTENT_EXTRA, -1).let { taskId ->
+                    removeTaskId(taskId)
+                }
+            }
+        }
+        return super.onStartCommand(intent, flags, startId)
```
shubham1g5: think this should return `START_NOT_STICKY` and we should re-start the service if it gets killed from the code. This is because starting this service without an intent action like  `UPDATE_PROGRESS_NOTIFICATION_ACTION` would just start the service but it would not really know what to do in that case.

shubham1g5 (changes requested):
can we provide a detailed safety story here on what we tested along with QA notes. Think it's critical for us to provide evidence on the solution working with simultaneous tasks to as part of safety story here.

app/src/org/commcare/services/NetworkNotificationService.kt
```diff
@@ -0,0 +1,137 @@
...
+        const val NETWORK_NOTIFICATION_ID = "network_notification_service_id"
+        var isServiceRunning = false
+        const val UPDATE_PROGRESS_NOTIFICATION_ACTION = "update_progress_notification"
+        const val STOP_NOTIFICATION_ACTION = "stop_notification"
+        const val START_NOTIFICATION_ACTION = "start_notification"
+        const val PROGRESS_TEXT_KEY_INTENT_EXTRA = "progress_text_key"
+        const val TASK_ID_INTENT_EXTRA = "task_id"
+    }
+
+    private val _taskIds = MutableStateFlow<List<Int>>(emptyList())
```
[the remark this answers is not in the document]
avazirna: There is a `collector` based on this to decide when to stop the service. Using `StateFlow` seems appropriate here, as it handles safely concurrent reads and writes to the list, right?

app/src/org/commcare/services/NetworkNotificationService.kt
avazirna: yes, I missed this

avazirna:
> can we provide a detailed safety story here on what we tested along with QA notes. Think it's critical for us to provide evidence on the solution working with simultaneous tasks to as part of safety story here.

@shubham1g5 at the moment, this is to target the `EngineResourceTask`, which is user-blocking, so concurrent execution are not expected, unless I'm missing a user case. I'll describe the testing that was performed, but I'd defer testing the concurrent executions to a separate ticket to account for future tasks, unless you think it's critical to do this now.

app/src/org/commcare/services/NetworkNotificationService.kt
```diff
@@ -0,0 +1,137 @@
...
+        } else {
+            startForeground(NETWORK_NOTIFICATION_ID.hashCode(), buildNotification("network.notification.service.starting"))
+        }
+        serviceScope.launch {
+            taskIds.collect { list ->
+                if (list.isEmpty() && isServiceRunning) {
+                    stopSelf()
+                }
+            }
+        }
```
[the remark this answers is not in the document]
avazirna: `collect` is a suspending function, so it must be called form a coroutine. But moving away from the Coroutine + `StateFlow` design in favor of a more simpler approach

shubham1g5 (approved):
confirming that you have tested notification disappearing after service stops on a real device ?

PR #3552, opened 2026-02-19
Fail HeartbeatWorker cleanly when session is inactive

avazirna:
## Product Description
This PR addresses the most prominent non-fatal issue reported in Crashlytics. It's caused when the user's session expires and the `HeartbeatWorker` tries to access `CommCareApplication.instance().session` to verify if the session is active. This method throws a`SessionUnavailableException` exception when there is no active session, which is caught by the generic exception handler and logged as an "unexpected exception during heartbeat communications". Session expiry is expected behavior, and heartbeat requests are not relevant to CommCare when there is no active user session. 

 [Crashlytics report](https://console.firebase.google.com/project/commcare-a57e4/crashlytics/app/android:org.commcare.dalvik/issues/8fffd2dd42bdc34ab0be7ec7a73f97b8?time=7d&types=error&sessionEventKey=6997029B0054000169BFC87869125AE4_2186998611152130502). This is 

## Technical Summary
- Added an early `isSessionActive()` check in `HeartbeatWorker.doWork()` that returns `Result.failure()` when the session has expired. 

## Feature Flag
N/A

### Safety story
This change is a small guard clause. Existing tests continue to pass.

## Labels and Review


conroy-ricketts (dismissed):
Looks like we can delete an unused import as well?

PR #3554, opened 2026-02-20
Open keyboard with focus on name input on opening up name page

shubham1g5:
## Product Description

https://dimagi.atlassian.net/browse/CCCT-2103

https://github.com/user-attachments/assets/7180fd2f-db38-44ea-88d9-d307ed112a48

## Technical Summary

I tried the similar things listed in https://github.com/dimagi/commcare-android/pull/3532 to make the keyboard open without a delay but it didn't do the trick, so I ended up using the current approach to open the keyboard as well here that opens the keyboard after a 250 ms delay. 

## Safety Assurance

Verified locally

## Labels and Review


Jignesh-dimagi:
@shubham1g5 What is the screen condition shown in the video at 00:18 seconds? And also, it's showing the name screen again from the code screen. Are you pressing the back button?

PR #3566, opened 2026-02-24
Move unsent forms query off the main thread

avazirna:
## Product Description
The home screen sync button triggers a query to count unsent forms. This is a database query that is currently executed on the main thread, which can cause ANRs. Over the past 30 days, this ANR has been one of the most frequent, with over 3k occurences.
This PR moves the query to a background thread using a new \`LatestTaskExecutor\` utility, which posts the result back to the main thread via a callback once the query completes. Rapid successive calls cancel any previous query to avoid stale results.

## Technical Summary
Key changes:
- **`LatestTaskExecutor<T>`** — new Kotlin utility class wrapping a single-thread executor; cancels the previous task on each new submission and posts the result to the main thread via a interface `Callback<T>`
- **`SyncDetailCalculations.updateSubText`** — \`getNumUnsentForms()\` now runs off the main thread; UI update runs in the callback on the main thread with an activity lifecycle guard

### Safety story

- The UI update callback checks \`activity.isFinishing() || activity.isDestroyed()\` before touching any views
- Only one executor instance is used for the single sync button call site — cancel-previous behaviour is scoped correctly
- All other logic in \`updateSubText\` (sync time display, colour) is unchanged

### QA Plan

- Open the home screen and verify the unsent forms count displays correctly
- Submit forms and verify the count updates on next home screen refresh
- Rapidly navigate away and back to verify no stale count or crash

## Labels and Review


app/src/org/commcare/tasks/LatestTaskExecutor.kt
```diff
@@ -0,0 +1,41 @@
...
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.Executors
+import java.util.concurrent.Future
+import java.util.concurrent.atomic.AtomicReference
+
+/**
+ * Executor that ensures only the latest submitted task is executed, and any previous tasks are cancelled.
+ * The result is posted back to the main thread via the provided callback.
+ */
+class LatestTaskExecutor<T> {
```
shubham1g5: @avazirna I am curious why did you choose this way instead of going with the more standard Android patterns like Coroutines +  Flow/Live Data  to handle async processing. 

I think we would want to ensure that a solution here -

- is suited for long term and is aligned with Android standards patterns
- Integrates seamlessly with Android lifecycle and doesn't need extra lifecycle management from us (that is the main reason we want to move off from Async tasks as well)
 - Ensures that the undelying tasks are cancellable i.e. the DB query should not continue running if the ui component is not alive
- Ensures that Any Failures are not hidden from UI

avazirna: > I am curious why did you choose this way instead of going with the more standard Android patterns like Coroutines + Flow/Live Data to handle async processing.

I tried to balance the use case with the overhead of invoking coroutines from Java, especially given that `CompletableFuture` is not an option. I’m happy to revisit this if you have a different perspective.

I agree with all the points you highlighted. Absolutely, coroutines is the recommended approach, however, `ExecutorService` is still supported 

About these two aspects:
> Ensures that the undelying tasks are cancellable i.e. the DB query should not continue running if the ui component is not alive

SQLiteDatabase queries are blocking, so even with coroutines, if the ui gets destroyed they will continue running. Right? I can double-check this

> Ensures that Any Failures are not hidden from UI

Given the use case, I'm inclined to log exceptions as a non-fatals to avoid impacting the user.

avazirna: [b7be1fc](https://github.com/dimagi/commcare-android/pull/3566/commits/b7be1fcfbb8cf9571cc136d919af32f02d412e43)

app/src/org/commcare/utils/SyncDetailCalculations.java
```diff
@@ -34,44 +38,68 @@ public class SyncDetailCalculations {
...
+                    }
+                });
+    }
 
-        if (numUnsentForms != 0 || HiddenPreferences.shouldShowUnsentFormsWhenZero()) {
-            if (!TextUtils.isEmpty(syncStatus)) {
-                syncStatus += "\n\n";
-            }
-            syncStatus += syncIndicator;
+    public static LatestTaskExecutor<Integer> getUnsentFormsExecutor() {
```
shubham1g5: why public static ?

app/src/org/commcare/utils/SyncDetailCalculations.java
```diff
@@ -34,44 +38,68 @@ public class SyncDetailCalculations {
...
+                                squareButtonViewHolder.subTextView,
+                                numUnsentForms,
+                                lastSyncTimeAndMessage.first,
+                                activity.getResources().getColor(cardDisplayData.subTextColor),
+                                activity.getResources().getColor(R.color.cc_dark_warm_accent_color));
+                    }
+
+                    @Override
+                    public void onError(@NotNull Exception exception) {
+                        Logger.exception("LatestTaskExecutor task failed", exception);
```
shubham1g5: nit: `LatestTaskExecutor task failed` -> `Failed to get unsent forms`

app/src/org/commcare/tasks/LatestTaskExecutor.kt
```diff
@@ -0,0 +1,39 @@
...
+        task: Callable<T>,
+        callback: Callback<T>,
+    ) {
+        currentJob?.cancel()
+        currentJob = scope.launch {
+            try {
+                val result = withContext(Dispatchers.IO) { task.call() }
+                withContext(Dispatchers.Main) { callback.onResult(result) }
+            } catch (e: CancellationException) {
+                throw e
```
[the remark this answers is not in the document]
avazirna: Not really. This is a especial coroutines exception used to signal a cancellation to internal parts of coroutines, it needs to be rethrown to ensure that the cancellation propagates as expected. I'm exposing it to avoid being logged as an exception.

app/src/org/commcare/utils/SyncDetailCalculations.java
```diff
@@ -34,44 +38,68 @@ public class SyncDetailCalculations {
...
+                LifecycleOwnerKt.getLifecycleScope(activity),
+                SyncDetailCalculations::getNumUnsentForms,
+                new LatestTaskExecutor.Callback<>() {
+                    @Override
+                    public void onResult(Integer numUnsentForms) {
+                        if (activity.isFinishing() || activity.isDestroyed()) {
+                            return;
+                        }
+
+                        Pair<Long, String> lastSyncTimeAndMessage = getLastSyncTimeAndMessage();
```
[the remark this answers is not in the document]
avazirna: https://github.com/dimagi/commcare-android/pull/3566/changes/c1a1936582c19092e4c86a7a347d4d0b9f0c21a8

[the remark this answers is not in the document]
shubham1g5: I mean the complete logic into another method like - 

````
 getUnsentFormsExecutor().submit(
                LifecycleOwnerKt.getLifecycleScope(activity),
                SyncDetailCalculations::getNumUnsentForms,
                new LatestTaskExecutor.Callback<>() {
                    @Override
                    public void onResult(Integer numUnsentForms) {
                        if (activity.isFinishing() || activity.isDestroyed()) {
                            return;
                        }
                        updateUnsentFormsUI()
                    }

                    @Override
                    public void onError(@NotNull Exception exception) {
                        Logger.exception("Failed to retrieve unsent forms count ", exception);
                    }
````

this is to decouple the actual UI method from the boilerplate around this update cleanly.

shubham1g5 (approved):

PR #3569, opened 2026-02-25
Including user location in map zoom when available

OrangeAndGreen:
https://dimagi.atlassian.net/browse/CCCT-2142
https://dimagi.atlassian.net/browse/CCCT-2107

## Product Description
When the user goes to the Entity Map page, the map will now include their location when auto-zooming (in case it isn't already within the bounds of the provided case data points).

Screenshot showing 9 placemarks and user outside of the grid (Southwest corner) with 10% padding around calculated bounds.
Screenshot_20260302-162056

## Technical Summary
Implemented a one-time location request in CommCareFusedLocationController.
Calling the new function from EntityMapActivity when loading the page and map.
If the location is unavailable for any reason, the map zooms without it (based on case data points).

I also refactored the "added" logic that was used to ensure the bounds builder had at least one coordinate before attempting to build the bounds. It turns out it's just as good to see if any of the numMarkers/Polygons/Geopoints variables are >0, since anytime the helper methods returned true (i.e. "yes I added") they were also incrementing one of those counters. The logic is a bit cleaner now and more friendly to the async calls.

## Feature Flag
None

### Safety story
I tested with location services enabled and disabled.
When enabled, my location was included in the zoom.
When disabled, my location was not included in the zoom.

### Automated test coverage
None

### QA Plan
Test the map zooming when entering the Entity Map page, with and without location enabled on the device.
Zooming should occur either way.
When location is enabled, the user's location should be factored into the zoom calculation so the user is included.

### DIFF
--- a/app/src/org/commcare/gis/EntityMapActivity.java
+++ b/app/src/org/commcare/gis/EntityMapActivity.java
@@ -13,7 +13,6 @@ import android.view.View;
 import com.google.android.gms.maps.CameraUpdateFactory;
 import com.google.android.gms.maps.GoogleMap;
 import com.google.android.gms.maps.SupportMapFragment;
-import com.google.android.gms.tasks.TaskCompletionSource;
 import com.google.android.gms.maps.model.BitmapDescriptor;
 import com.google.android.gms.maps.model.BitmapDescriptorFactory;
 import com.google.android.gms.maps.model.Circle;
@@ -24,6 +23,7 @@ import com.google.android.gms.maps.model.Marker;
 import com.google.android.gms.maps.model.MarkerOptions;
 import com.google.android.gms.maps.model.Polygon;
 import com.google.android.gms.maps.model.PolygonOptions;
+import com.google.android.gms.tasks.TaskCompletionSource;
 import com.google.android.material.switchmaterial.SwitchMaterial;
 import com.google.common.base.Strings;
 import com.google.firebase.perf.metrics.Trace;
@@ -163,7 +163,7 @@ public class EntityMapActivity extends CommCareActivity implements
     }
 
     private String getToggleLabel(String headerValue, int defaultKey) {
-        if(!Strings.isNullOrEmpty(headerValue)) {
+        if (!Strings.isNullOrEmpty(headerValue)) {
             return headerValue;
         }
 
--- a/app/src/org/commcare/gis/EntityMapActivityExt.kt
+++ b/app/src/org/commcare/gis/EntityMapActivityExt.kt
@@ -1,12 +1,12 @@
 package org.commcare.gis
 
+import androidx.lifecycle.lifecycleScope
 import com.google.android.gms.maps.GoogleMap
 import com.google.android.gms.tasks.Task
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.async
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.suspendCancellableCoroutine
-import androidx.lifecycle.lifecycleScope
 import org.commcare.location.LocationHelper
 import org.commcare.views.UserfacingErrorHandling
 import org.javarosa.xpath.XPathException
@@ -38,7 +38,11 @@ fun EntityMapActivity.loadMap(mapReadyTask: Task<GoogleMap>) {
         setupMap(map, location)
 
         if (entityError != null) {
-            UserfacingErrorHandling<EntityMapActivity>().logErrorAndShowDialog(this@loadMap, entityError, true)
+            UserfacingErrorHandling<EntityMapActivity>().logErrorAndShowDialog(
+                this@loadMap,
+                entityError,
+                true
+            )
         }
     }
 }
--- a/app/src/org/commcare/location/CommCareFusedLocationController.kt
+++ b/app/src/org/commcare/location/CommCareFusedLocationController.kt
@@ -15,10 +15,10 @@ import com.google.android.gms.location.LocationServices
 import com.google.android.gms.location.LocationSettingsRequest
 import com.google.android.gms.location.Priority
 import kotlinx.coroutines.suspendCancellableCoroutine
-import kotlin.coroutines.resume
 import org.commcare.util.LogTypes
 import org.commcare.utils.GeoUtils.locationServicesEnabledGlobally
 import org.javarosa.core.services.Logger
+import kotlin.coroutines.resume
 
 /**
  * @author $|-|!˅@M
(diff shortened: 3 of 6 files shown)


app/src/org/commcare/location/CommCareLocationControllerFactory.kt
```diff
@@ -12,7 +12,7 @@ class CommCareLocationControllerFactory {
 
     companion object {
         @JvmStatic
-        fun getLocationController(context: Context, mListener: CommCareLocationListener): CommCareLocationController {
+        fun getLocationController(context: Context, mListener: CommCareLocationListener?): CommCareLocationController {
```
Jignesh-dimagi: @OrangeAndGreen I don't think we should allow null listener.

conroy-ricketts (dismissed):
I don't think I have anything else to add here. Looking good!

app/src/org/commcare/location/CommCareFusedLocationController.kt
```diff
@@ -59,6 +59,21 @@ class CommCareFusedLocationController(
...
+    fun getCurrentLocation(callback: CurrentLocationCallback) {
+        if (isLocationPermissionGranted(mContext)) {
+            mFusedLocationClient.getCurrentLocation(Priority.PRIORITY_HIGH_ACCURACY, null)
+                .addOnSuccessListener { location -> callback.onResult(location) }
+                .addOnFailureListener { callback.onResult(null) }
+        } else {
+            callback.onResult(null)
+        }
+    }
+
```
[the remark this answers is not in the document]
OrangeAndGreen: Ah, sure enough, somehow I missed the available getCurrentLocation call in the provider controller before, but it's there.
I added the matching implementation in that class and added the function to the interface definition so it can be called genericly. [1897909](https://github.com/dimagi/commcare-android/pull/3569/commits/1897909189e8c5c125cdda20dee42571de830615)

app/src/org/commcare/location/CommCareLocationControllerFactory.kt
OrangeAndGreen: Yeah, the no-action callbacks for the listener seemed awkward and I saw the constructors and implementations for both classes were ready for null listeners so thought it might be simpler to just make the listener nullable in the factory method also.

But instead, I created a LocationHelper class to provide a simpler way for callers to get a one-time location, and included a no-action listener there. I changed the factory back to requiring a non-null listener.

Noting though that not providing a listener does make sense if you're not going to use the controller for streaming, but only for a one-time request. When calling getCurrentLocation on either the fused or provider controllers, they take a different callback that's completely unrelated to the listener passed in to the constructor.

app/src/org/commcare/gis/EntityMapActivity.java
```diff
@@ -206,26 +205,10 @@ public void onMapReady(@NonNull final GoogleMap map) {
...
-                    });
-                } else {
-                    finishLoadingPerformanceTrace();
-                }
-            });
-
             setupMapToggles();
+
+            LocationHelper.getCurrentLocation(this).addOnCompleteListener(task ->
+                    proceedWithLocation(builder, task.isSuccessful() ? task.getResult() : null));
```
Jignesh-dimagi: The only concern here is that this might leak whenever activity gets destroyed before getting the result. Activity should cancel the task if results are not received in onDestroy.

app/src/org/commcare/location/CommCareLocationController.kt
```diff
@@ -13,13 +13,19 @@ import org.javarosa.core.services.Logger
  * @author $|-|!˅@M
  */
 interface CommCareLocationController {
+    fun interface CurrentLocationCallback {
+        fun onResult(location: Location?)
+    }
```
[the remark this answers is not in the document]
Jignesh-dimagi: This looks good way to proceed. Ideally, we should have `BaseLocationListener` activity with full fledge implementation, which can be re-usable. But seems ok for this implementation. 

I have only concern as stated [here](https://github.com/dimagi/commcare-android/pull/3569#discussion_r2871745644)

shubham1g5:
@OrangeAndGreen think the 2 seconds timeout may be quite short here and think would be nice to bump it to `5` seconds as a max threshold

app/src/org/commcare/location/LocationHelper.kt
```diff
@@ -0,0 +1,27 @@
...
+import com.google.android.gms.tasks.Task
+import kotlinx.coroutines.suspendCancellableCoroutine
+import kotlinx.coroutines.withTimeoutOrNull
+import kotlin.coroutines.resume
+
+object LocationHelper {
+    const val LOCATION_TIMEOUT_MS: Long = 5000L
+
+    @JvmStatic
+    fun getCurrentLocation(context: Context): Task<Location?> {
```
shubham1g5: can be private ?

app/src/org/commcare/gis/EntityMapActivity.java
```diff
@@ -104,15 +111,45 @@ protected void onCreate(Bundle savedInstanceState) {
...
+
+        showProgressDialog(MAP_LOAD_TASK_ID);
+        Tasks.whenAll(entitiesTask, locationTask, mapReadyTask)
+                .addOnSuccessListener(this, ignored -> {
+                    dismissProgressDialogForTask(MAP_LOAD_TASK_ID);
+                    setupMap(mapReadyTask.getResult(), locationTask.getResult());
+                })
+                .addOnFailureListener(this, e -> {
+                    dismissProgressDialogForTask(MAP_LOAD_TASK_ID);
+                    if (e instanceof XPathException xe) {
```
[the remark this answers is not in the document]
OrangeAndGreen: Obsolete

app/src/org/commcare/gis/EntityMapActivityExt.kt
```diff
@@ -0,0 +1,44 @@
...
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.suspendCancellableCoroutine
+import androidx.lifecycle.lifecycleScope
+import org.commcare.location.LocationHelper
+import org.commcare.views.UserfacingErrorHandling
+import org.javarosa.xpath.XPathException
+import kotlin.coroutines.resume
+import kotlin.coroutines.resumeWithException
+
+fun EntityMapActivity.loadMap(mapReadyTask: Task<GoogleMap>) {
```
shubham1g5: This seems like a good way to export needed kotlin functionality out of Java classes.

Jignesh-dimagi (approved):

PR #3576, opened 2026-02-27
CCCT-2199 Landing Page Bug Fixes

conroy-ricketts:
I'm linking these bugs to [CCCT-2199](https://dimagi.atlassian.net/browse/CCCT-2199), but these fixes were inspired by [QA-8393](https://dimagi.atlassian.net/browse/QA-8393).

## Product Description

The TLDR is that I fixed some bugs regarding incorrect UI state at various stages of an opportunity.

Here is a video showing what the entire workflow looks like with the changes in this PR (keep your eyes on "Test Opp"):

https://github.com/user-attachments/assets/63d00c35-e068-4af2-a09b-557490d57f8f

## Technical Summary

So, when I was trying to reproduce the QA ticket ([QA-8393](https://dimagi.atlassian.net/browse/QA-8393)), I noticed a bug where if you complete all learn modules for a new opportunity, but do **_not_** start the assessment, the "Proceed" button is shown on the landing page which takes you to the Delivery overview page (the correct UX here is that we show the "Resume" button and tapping on that takes the user to the Learn Progress screen which has the "Go to assessment" button).

Here is a video demonstrating the **_incorrect_** UX that I described above (the opportunity called "The Best Opportunity Ever" in my video has all learn modules complete but I had **_not_** attempted the assessment yet):

https://github.com/user-attachments/assets/a313a9ef-a198-4d36-b9f7-23fbd232592a

Here is a video demonstrating the **_correct_** UX for that same opportunity:

https://github.com/user-attachments/assets/733efb74-53f7-4198-b41f-33b00925b5c2

Also, an important note here is that when getting the `passedAssessment` value from a job, we need the "composite job" from `ConnectJobUtils.getCompositeJob()` to get the correct value. 

I feel like this composite job thing has caused the team quite a bit of unnecessary headache in the past, so I just went ahead and refactored the code to use the composite job whenever we create the `ConnectLoginJobListModel`'s in `ConnectJobsListsFragment.setJobListData()` since that is what is being passed around everywhere in the code. This allows us to cleanup the code a bit, and hopefully we no longer have anymore issues regarding this.

### Safety story

I ran multiple opportunities through the entire flow and checked the landing page every step of the way to verify that both the opportunity card tile and the "View Info" bottom sheet shows the correct UI given the current state of the opportunities.

### QA Plan

Run a brand new opportunity through the entire workflow and check the landing page every step of the way to verify that both the opportunity card tile and the "View Info" bottom sheet shows the correct UI given the current state of your opportunity.

[CCCT-2199]: https://dimagi.atlassian.net/browse/CCCT-2199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[QA-8393]: https://dimagi.atlassian.net/browse/QA-8393?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[QA-8393]: https://dimagi.atlassian.net/browse/QA-8393?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/res/layout/connect_job_list_item.xml
+++ b/app/res/layout/connect_job_list_item.xml
@@ -43,8 +43,8 @@
 
         <ImageView
             android:id="@+id/iv_completed_check"
-            android:layout_width="32dp"
-            android:layout_height="32dp"
+            android:layout_width="20dp"
+            android:layout_height="20dp"
             android:contentDescription="@null"
             android:importantForAccessibility="no"
             android:src="@drawable/ic_connect_check_circle"
--- a/app/res/values/colors.xml
+++ b/app/res/values/colors.xml
@@ -233,6 +233,7 @@
     <color name="lunar_mist">#E7E8F8</color>
     <color name="steel_dust">#ABABAB</color>
     <color name="sterling_ash">#959595</color>
+    <color name="burnt_amber">#C67E13</color>
 
     <color name="ic_backup">#14006C</color>
     <color name="icon_tint_dark_gray">#333333</color>
--- a/app/src/org/commcare/adapters/JobListConnectHomeAppsAdapter.java
+++ b/app/src/org/commcare/adapters/JobListConnectHomeAppsAdapter.java
@@ -257,6 +257,7 @@ public class JobListConnectHomeAppsAdapter extends RecyclerView.Adapter<Recycler
     ) {
         int progress = 0;
         int progressColor = 0;
+        boolean showPercentageText = true;
         ConnectLoginJobListModel.JobListEntryType jobType = item.getJobType();
         ConnectJobRecord job = item.getJob();
 
@@ -266,6 +267,11 @@ public class JobListConnectHomeAppsAdapter extends RecyclerView.Adapter<Recycler
         }
 
         switch (jobType) {
+            case NEW_OPPORTUNITY:
+                progress = 100;
+                progressColor = ContextCompat.getColor(context, R.color.burnt_amber);
+                showPercentageText = false;
+                break;
             case LEARNING:
                 progress = item.getLearningProgress();
                 progressColor = ContextCompat.getColor(context, R.color.connect_blue_color);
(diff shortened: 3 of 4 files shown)


OrangeAndGreen (approved):
One thing to consider but otherwise looks fine to me.

One thing to note: I think long-term the goal with the ConnectLoginJobListModel should be not to have the job be included in the object, but instead have any per-item state values be calculated in advance and included directly in the model. That way the adapter simply looks up answers, it doesn't calculate anything on its own.
The interim solution being used now isn't too bad since the calculations are pretty cheap or also just direct lookups on the job. Before we had each row retrieving the composite job from the DB, which had more potential to create performance issues when scrolling.

OrangeAndGreen:
Oh, might be worth fixing [this issue](https://dimagi.slack.com/archives/C09J9M8RTJ5/p1772215249370939?thread_ts=1772205967.643919&cid=C09J9M8RTJ5) Brendon brought up before merging this

conroy-ricketts:
@shubham1g5 @OrangeAndGreen Pushed another commit just now to address the thing Brendon brought up in Slack, so basically now we show the learning icon with a full 100% when an assessment is passed. As I understand it, this will be a temporary solution until we implement [CCCT-2138](https://dimagi.atlassian.net/browse/CCCT-2138)

[CCCT-2138]: https://dimagi.atlassian.net/browse/CCCT-2138?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3606, opened 2026-03-11
Offline First Implementation for Learn Progress and Jobs List Page

shubham1g5:
## Product Description

https://dimagi.atlassian.net/browse/CCCT-2262

Implements offline first behaviour with data refresh indicators as per the [design spec](https://docs.google.com/docume

Current UX:

https://github.com/user-attachments/assets/41ac9744-a2ed-4706-8df2-0cd04882c222

Error Screenshots:
WhatsApp Image 2026-05-08 at 17 32 24
WhatsApp Image 2026-05-08 at 17 32 39

There are a couple things from the design spec which are not part of this PR and will be handled in seperate tickets - 

- [Back Online Indicator when sync is successful after user regains connectivity](https://dimagi.atlassian.net/browse/CCCT-2354)
- Any UI update transitions to show updated data

In general, the PR focuses more on establishing undelying offline first architecture rather than UX of these pages. 

## Technical Summary

The [Tech Spec](https://docs.google.com/document/d/1AazP-tG5mRXpr5Cmto8GT9Ccbr5n6oACIdiQfY4j-kw/edit?tab=t.xd9ky82f5mtn) provides a great overview of the implementation and a more detailed Claude Plan this implementaion is based on is [available here](https://github.com/dimagi/commcare-android/blob/3ce3681ff5a218d8c319368e923d3fb4b2bb22f5/docs/claude/plans/2026-02-23-offline-first-connect-architecture.md)

### Safety story

- All new classes have extensive unit test coverage
- I have locally tested different states shown in the demo video. 
- Probably the riskiest downside of this PR is that we call out the server with multiple network requests to same endpoint on user going back and forth in a page, I have taken care of this in code by de-duplicating the requests on same endpoint and not making the same request again and have provided unit tests for the same. 
- Think the number of changes are high on this PR and therefore I am planning dedicated QA around this work once we complete initial set of implementation on this work. 

### QA Plan

Planning to get this QA'ed once the [delivery progress implementation](https://dimagi.atlassian.net/browse/CCCT-2263)  implementation is complete. 

## Labels and Review

## Review Notes

- Commits are structured well and shows AI commits and iterations over them sequentially. Although given high number of commits, It probably is not efficient to review this commit by commit but rather changes should be easier to review as a whole. 
- This PR seems quite big but around half the new lines added is tests on new classes, so hoping it's not a super big lift but would be happy to separate out base classes from their implementation if reviewer think it would be easier to review.


app/src/org/commcare/connect/network/ConnectNetworkHelper.kt
```diff
@@ -0,0 +1,129 @@
...
+                requestBody = ModernHttpRequester.getPostBody(multimap)
+                headers = getContentHeadersForXFormPost(requestBody)
+            } else {
+                val json = Gson().toJson(params)
+                requestBody = RequestBody.create("application/json".toMediaType(), json)
+                headers = outputHeaders
+            }
+
+            addVersionHeader(headers, version)
+            return requestBody
```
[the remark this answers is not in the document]
shubham1g5: Fixed in 2dc262545. Changed `addVersionHeader(headers, version)` to `addVersionHeader(outputHeaders, version)` so the version header is always written to the caller-supplied map regardless of encoding path.

docs/claude/plans/2026-03-09-offline-first-tdd-iteration-design.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Dependency Chain
+
+Each phase depends on all prior phases being complete (compile + tests passing):
+
+```
+Phase 0 → Phase 1a → Phase 1b → Phase 1.5a → Phase 1.5b
+       → Phase 2a → Phase 2b → Phase 3a → Phase 3b
+                             → Phase 4a → Phase 4b → Phase 5
+```
```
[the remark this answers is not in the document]
shubham1g5: Fixed in 2dc262545. Added `text` language tag to the dependency-chain code fence.

app/src/org/commcare/connect/repository/ConnectRepository.kt
```diff
@@ -0,0 +1,130 @@
...
+        ): Flow<DataState<List<ConnectJobRecord>>> =
+            offlineFirstFlow(
+                endpoint = ENDPOINT_OPPORTUNITIES,
+                forceRefresh = forceRefresh,
+                policy = policy,
+                loadCache = {
+                    ConnectJobUtils
+                        .getCompositeJobs(context, ConnectJobRecord.STATUS_ALL_JOBS, null)
+                        .takeIf { it.isNotEmpty() }
+                },
```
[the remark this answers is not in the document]
shubham1g5: The `.takeIf { it.isNotEmpty() }` is intentional: without it, an empty list returned on first install (before any sync) would reach `emit(DataState.Cached(cachedData, lastSyncTime\!\!))` with a null `lastSyncTime`, causing an NPE. To support a valid empty post-sync state as `Cached`, both the `takeIf` logic and the `lastSyncTime\!\!` assumption need to be revisited together. Leaving as a follow-up.

docs/plans/2026-03-09-claude-issue-implementer.md
```diff
@@ -0,0 +1,444 @@
...
+#   1 = exhausted retries (caller should escalate)
+
+set -euo pipefail
+
+: "${BRANCH_NAME:?}"
+: "${PR_NUMBER:?}"
+: "${MAX_RETRIES:?}"
+: "${ANTHROPIC_API_KEY:?}"
+: "${BROWSERSTACK_USERNAME:?}"
+: "${BROWSERSTACK_PASSWORD:?}"
```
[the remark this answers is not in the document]
shubham1g5: No occurrences of `BROWSERSTACK_ACCESS_KEY` found in the file — it already uses `BROWSERSTACK_PASSWORD` throughout.

app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java
```diff
@@ -69,44 +69,57 @@ public ConnectJobsListsFragment() {
...
+                String errorMsg = PersonalIdOrConnectApiErrorHandler.handle(
+                        requireActivity(), error.getErrorCode(), error.getThrowable());
+                if (!errorMsg.isEmpty()) {
+                    showError(errorMsg);
+                }
+                List<ConnectJobRecord> cachedData = error.getCachedData();
+                if (cachedData != null) {
+                    setJobListData(cachedData);
+                } else {
+                    navigateFailure();
```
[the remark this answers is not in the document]
shubham1g5: Already handled: offlineFirstFlow captures cachedData at the start of the flow and passes it to DataState.Error.from(throwable, cachedData). The fragment checks error.getCachedData() and calls setJobListData(cachedData) when non-null. navigateFailure() is only reached when there was genuinely no cached data (e.g. first load, no prior sync).

app/src/org/commcare/CommCareApplication.java
```diff
@@ -219,6 +220,8 @@ public class CommCareApplication extends Application implements LifecycleEventOb
     public void onCreate() {
         super.onCreate();
 
+        ConnectSyncPreferences.Companion.getInstance(this).markSessionStart();
```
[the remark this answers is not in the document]
shubham1g5: Resetting the session on Personal ID user login here - https://github.com/dimagi/commcare-android/pull/3606/commits/869276ac879bbb660d710725bb7895cfc95240f9 , we don't need to do the same on CommCare App Login as this preferene is only used for Connect Screens.

app/src/org/commcare/connect/viewmodel/ViewModelExtensions.kt
```diff
@@ -0,0 +1,24 @@
...
+import kotlinx.coroutines.flow.catch
+import kotlinx.coroutines.launch
+import org.commcare.connect.repository.DataState
+
+// viewModelScope uses Dispatchers.Main.immediate — liveData.value = is safe (no postValue).
+fun <T> ViewModel.collectInto(
+    flow: Flow<DataState<T>>,
+    liveData: MutableLiveData<DataState<T>>,
+) {
+    viewModelScope.launch {
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3606/commits/d35d6f05d78e8b6b0e5fb2fc8e5f42a3846d9f53

app/src/org/commcare/views/TopBarErrorViewController.kt
```diff
@@ -14,11 +15,20 @@ class TopBarErrorViewController(
 
     private var animator: ValueAnimator? = null
 
-    fun show(message: String) {
+    fun showError(message: String) {
+        errorView.setBackgroundColor(errorView.context.getColor(R.color.connect_warning_color))
```
OrangeAndGreen: Nit: This line is always paired with showBar, thinking that function could take the color as an additional input and make sure it always gets set(plus a little less code repetition).

conroy-ricketts (changes requested):
So far looking pretty good! I'm only about halfway through the changes. I'll have to come back and pick up where I left off later.

I also had a couple of suggestions that I dropped on the demo in Slack. I'll post those comments here too so that they don't get lost in the weeds:

> Something I noticed during code review - the landing page is sometimes fully blank on initial load (in my second video it happens very fast, but the first time I tried it the blank page lasted for a few seconds), and at other times it flashes the message telling my that I haven't been invited to any opportunities which is incorrect.

> What if we added some different UI there to the middle of the page during initial sync?

https://github.com/user-attachments/assets/dc8b6969-79c0-44aa-ab35-49567795e766

https://github.com/user-attachments/assets/c691a717-b93f-4697-87dd-ad1820321e7e

> Here's a better video showing what I mean:

https://github.com/user-attachments/assets/12e3074f-7280-4e08-8a5f-a843c813bb6e

> Another small thing I noticed:

> There is a gap between the status bar and the rest of the UI which causes some wonky UI. Can we eliminate that gap? (see right under the status bar as I scroll in this video)

https://github.com/user-attachments/assets/73320a01-eecc-4755-b037-b30fa05ceb49

OrangeAndGreen (commented):
Things look great so far! I made it through the main code changes, need to finish reviewing the test code tomorrow but wanted to submit comments in the meantime.

app/src/org/commcare/connect/network/connect/ConnectNetworkClient.kt
```diff
@@ -0,0 +1,105 @@
+package org.commcare.connect.network.connect
+
+import android.content.Context
+import androidx.annotation.VisibleForTesting
+import kotlinx.coroutines.CancellationException
+import okhttp3.ResponseBody
+import org.commcare.CommCareApplication
```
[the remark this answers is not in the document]
shubham1g5: yes, I plan to run lint on the whole PR before merge as per process, [this one](https://github.com/dimagi/commcare-android/pull/3606/commits/153decb88f6b96dadb97f633ce3ef4cc98a23e0d) addresses the import you pointed out

app/res/values/strings.xml
```diff
@@ -397,6 +397,8 @@
 
     <string name="connect_last_update">Updated: %s</string>
     <string name="connect_last_synced">Last synced: %s</string>
+    <string name="connect_sync_successful">Last synced: Just Now</string>
+    <string name="connect_sync_failed">Failed to sync updates, Last synced: %s</string>
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3606/commits/153decb88f6b96dadb97f633ce3ef4cc98a23e0d

app/src/org/commcare/connect/viewmodel/ConnectLearningProgressViewModel.kt
```diff
@@ -0,0 +1,36 @@
...
+    @VisibleForTesting
+    internal var repository: ConnectRepository = ConnectRepository.getInstance(application)
+
+    private val _learningProgress = MutableLiveData<DataState<ConnectJobRecord>>()
+    val learningProgress: LiveData<DataState<ConnectJobRecord>> = _learningProgress
+
+    private var loadLearnProgressJob: Job? = null
+
+    fun loadLearningProgress(
+        job: ConnectJobRecord,
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3606/commits/153decb88f6b96dadb97f633ce3ef4cc98a23e0d

app/src/org/commcare/views/TopBarErrorViewController.kt
shubham1g5: https://github.com/dimagi/commcare-android/pull/3606/commits/153decb88f6b96dadb97f633ce3ef4cc98a23e0d

app/src/org/commcare/views/TopBarErrorViewController.kt
```diff
@@ -14,11 +15,20 @@ class TopBarErrorViewController(
...
         showBar(message, showOffline = false, autoDismiss = true)
     }
 
+    fun showMessage(message: String) {
+        errorView.setBackgroundColor(errorView.context.getColor(R.color.connect_top_bar_success))
+        showBar(message, showOffline = false, autoDismiss = true)
+    }
+
+    fun isErrorShowing(): Boolean = errorView.visibility == View.VISIBLE
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3606/commits/153decb88f6b96dadb97f633ce3ef4cc98a23e0d

app/src/org/commcare/connect/network/connect/ConnectNetworkClient.kt
```diff
@@ -0,0 +1,105 @@
...
+    internal constructor(
+        private val apiService: ConnectApiService,
+    ) {
+        companion object {
+            private const val API_VERSION = "1.0"
+
+            @Volatile
+            private var instance: ConnectNetworkClient? = null
+
+            fun getInstance(context: Context): ConnectNetworkClient =
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3606/commits/153decb88f6b96dadb97f633ce3ef4cc98a23e0d

app/src/org/commcare/connect/repository/ConnectRepository.kt
```diff
@@ -0,0 +1,131 @@
+package org.commcare.connect.repository
+
+import android.content.Context
+import androidx.annotation.VisibleForTesting
+import kotlinx.coroutines.CancellationException
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3606/commits/153decb88f6b96dadb97f633ce3ef4cc98a23e0d

app/res/values/colors.xml
```diff
@@ -143,6 +143,7 @@
     <color name="connect_grey">#9A9A9A</color>
     <color name="connect_divider_color">#E5E7EB</color>
     <color name="connect_background_color">#f7f9fa</color>
+    <color name="connect_top_bar_success">#87BDDF</color>
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3606/commits/153decb88f6b96dadb97f633ce3ef4cc98a23e0d

app/src/org/commcare/connect/network/base/ConnectApiException.kt
```diff
@@ -0,0 +1,7 @@
+package org.commcare.connect.network.base
+
+// Carries typed error code through Result<T> chain; repository extracts it into DataState.Error.
+class ConnectApiException(
```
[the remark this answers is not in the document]
shubham1g5: curious what advantages you see of that ? I mostly use a `data` class for simple POJO kind of models.

app/src/org/commcare/connect/network/base/ConnectApiException.kt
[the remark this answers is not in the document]
shubham1g5: that method gets called from bunch of other network calls in `PersonalIdApiHandler` where we are not using `ConnectApiException` , moreover we are not utilizing `BaseApiHandler` in the refactored code at all to justify any changes there.

app/src/org/commcare/connect/network/connect/ConnectNetworkClient.kt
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/commit/153decb88f6b96dadb97f633ce3ef4cc98a23e0d

app/src/org/commcare/connect/network/connect/ConnectNetworkClient.kt
```diff
@@ -0,0 +1,105 @@
...
+import java.io.IOException
+import java.io.InputStream
+
+class ConnectNetworkClient
+    @VisibleForTesting
+    internal constructor(
+        private val apiService: ConnectApiService,
+    ) {
+        companion object {
+            private const val API_VERSION = "1.0"
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3606/commits/e2097c3d4d65c55082d328857b03d39352f2e266

app/src/org/commcare/connect/network/connect/ConnectNetworkClient.kt
```diff
@@ -0,0 +1,105 @@
...
+                        Result.failure(ConnectApiException(PersonalIdOrConnectApiErrorCodes.JSON_PARSING_ERROR, e))
+                    }
+                } else {
+                    val errorCode = mapHttpErrorCode(response.code(), response.errorBody()?.string())
+                    Result.failure(ConnectApiException(errorCode))
+                }
+            } catch (e: CancellationException) {
+                throw e
+            } catch (e: LoginInvalidatedException) {
+                throw e
```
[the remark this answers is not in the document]
shubham1g5: this was some left over cleanup from an earlier implementation - https://github.com/dimagi/commcare-android/pull/3606/commits/e2097c3d4d65c55082d328857b03d39352f2e266

app/src/org/commcare/connect/network/ConnectNetworkHelper.kt
```diff
@@ -0,0 +1,127 @@
...
+import org.json.JSONException
+import org.json.JSONObject
+import kotlin.coroutines.resume
+
+class ConnectNetworkHelper {
+    companion object {
+        @JvmStatic
+        fun addVersionHeader(
+            headers: HashMap<String, String>,
+            version: String?,
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3606/commits/e2097c3d4d65c55082d328857b03d39352f2e266

app/src/org/commcare/connect/network/ConnectNetworkHelper.kt
```diff
@@ -0,0 +1,127 @@
...
+            if (useFormEncoding) {
+                val multimap = ArrayListMultimap.create<String, String>()
+                for ((key, value) in params) {
+                    multimap.put(key, value.toString())
+                }
+                requestBody = ModernHttpRequester.getPostBody(multimap)
+                outputHeaders.putAll(getContentHeadersForXFormPost(requestBody))
+            } else {
+                val json = Gson().toJson(params)
+                requestBody = RequestBody.create("application/json".toMediaType(), json)
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3606/commits/e2097c3d4d65c55082d328857b03d39352f2e266

app/src/org/commcare/connect/network/ConnectNetworkHelper.kt
```diff
@@ -0,0 +1,127 @@
...
+            addVersionHeader(outputHeaders, version)
+            return requestBody
+        }
+
+        private fun getContentHeadersForXFormPost(postBody: RequestBody): HashMap<String, String> {
+            val headers = HashMap<String, String>()
+            headers["Content-Type"] = "application/x-www-form-urlencoded"
+            try {
+                headers["Content-Length"] = postBody.contentLength().toString()
+            } catch (e: Exception) {
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3606/commits/e2097c3d4d65c55082d328857b03d39352f2e266

app/src/org/commcare/connect/network/ConnectNetworkHelper.kt
```diff
@@ -0,0 +1,127 @@
...
+            return headers
+        }
+
+        @JvmStatic
+        fun checkForLoginFromDifferentDevice(errorBody: String?): Boolean {
+            if (errorBody == null) return false
+            return try {
+                val json = JSONObject(errorBody)
+                "LOGIN_FROM_DIFFERENT_DEVICE" == json.optStringSafe("error_code", null)
+            } catch (e: JSONException) {
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3606/commits/e2097c3d4d65c55082d328857b03d39352f2e266

app/src/org/commcare/connect/repository/ConnectRepository.kt
```diff
@@ -0,0 +1,131 @@
...
+        companion object {
+            const val ENDPOINT_OPPORTUNITIES = "/opportunities"
+            const val ENDPOINT_LEARNING_PREFIX = "/learning_progress/"
+
+            @Volatile
+            private var instance: ConnectRepository? = null
+
+            fun getInstance(context: Context): ConnectRepository =
+                instance ?: synchronized(this) {
+                    instance ?: ConnectRepository(
```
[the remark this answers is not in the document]
shubham1g5: This is actually a very common singleton pattern to ensure single unique instance for a singleton, pasting an AI response below on why it's desired - 

````
Why Two Null Checks?
First check instance ?: — before synchronized

Runs on every call, with no locking overhead
If instance already exists (the common case after initialization), it returns immediately
Avoids the expensive synchronized block entirely once initialized

Second check instance ?: — inside synchronized

Necessary because two threads could both pass the first check simultaneously
Thread A and Thread B both see instance == null
Thread A acquires the lock and creates the instance
Thread B is waiting. When B gets the lock, without this second check, it would create a second instance, breaking the singleton
````

File: app/src/org/commcare/connect/network/connect/ConnectNetworkClient.kt [hunk 24]
```diff
@@ -0,0 +1,105 @@
...
+            private var instance: ConnectNetworkClient? = null
+
+            fun getInstance(context: Context): ConnectNetworkClient =
+                instance ?: synchronized(this) {
+                    instance ?: ConnectNetworkClient(
+                        BaseApiClient
+                            .buildRetrofitClient(ConnectApiClient.BASE_URL)
+                            .create(ConnectApiService::class.java),
+                    ).also { instance = it }
+                }
```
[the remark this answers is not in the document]
shubham1g5: responded on similar comment below

app/src/org/commcare/connect/network/ConnectApiService.kt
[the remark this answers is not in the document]
shubham1g5: Think that is not a straightfroward change and will increase the PR scope quite a lot as it would mean changing a lot of network methods and  exposing all APIs via `ConnectNetworkClient` , we also can't really merge `getConnectOpportunities` from the two service classes as we need both suspend and callback implementations of this method at present unless we migrate all netowork api calls to use Coroutines.

shubham1g5:
> There is a gap between the status bar and the rest of the UI which causes some wonky UI. Can we eliminate that gap? (see right under the status bar as I scroll in this video)

You have a great eye! Resolved by https://github.com/dimagi/commcare-android/pull/3606/commits/4574bd438efc0365c0a61095688c290572c25494

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
```diff
@@ -115,47 +132,96 @@ abstract class BaseConnectFragment<B : ViewBinding> :
...
+                    onSuccess.accept(state.data)
+                }
+                is DataState.Error -> {
+                    hideLoading()
+                    if (state.errorCode == BaseApiHandler.PersonalIdOrConnectApiErrorCodes.NETWORK_ERROR &&
+                        !ConnectivityStatus.isNetworkAvailable(requireContext())
+                    ) {
+                        showOfflineIndicator()
+                    } else {
+                        showError(getString(R.string.connect_sync_failed, getRelativeLastSyncTime()))
```
Jignesh-dimagi, line 201: This is not handling TOKEN_UNAVAILABLE_ERROR and TOKEN_DENIED_ERROR. This just shows the error on the bar, but no action is taken based on error.

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
```diff
@@ -115,47 +132,96 @@ abstract class BaseConnectFragment<B : ViewBinding> :
...
+            if (!isAdded) return@observe
+            if (lastDataState == null && (state is DataState.Success || state is DataState.Error)) {
+                // terminal states should not be shown on initial load to avoid jarring UX
+                // this happens when LiveData emits a cached value immediately upon observation
+                return@observe
+            }
+            when (state) {
+                is DataState.Loading -> {
+                    hideError()
+                    showLoading()
```
Jignesh-dimagi, line 178: This will always show the loading on the top whenever the API is called. Do we have a use case where it's required to silently load the data?

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
```diff
@@ -115,47 +132,96 @@ abstract class BaseConnectFragment<B : ViewBinding> :
...
+                }
+                is DataState.Error -> {
+                    hideLoading()
+                    if (state.errorCode == BaseApiHandler.PersonalIdOrConnectApiErrorCodes.NETWORK_ERROR &&
+                        !ConnectivityStatus.isNetworkAvailable(requireContext())
+                    ) {
+                        showOfflineIndicator()
+                    } else {
+                        showError(getString(R.string.connect_sync_failed, getRelativeLastSyncTime()))
+                    }
```
Jignesh-dimagi, line 202: In case of error, this is always showing an error, and the main fragment is never aware that an error has occurred. I am thinking of a case where it's required to inform the main fragment to take some action. Should we create this mechanism?

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
```diff
@@ -115,47 +132,96 @@ abstract class BaseConnectFragment<B : ViewBinding> :
...
+     * Observes a LiveData of DataState and handle UI updates for loading, success, cached, and error states.
+     */
+    protected fun <T> observeDataState(
+        liveData: LiveData<DataState<T>>,
+        onCached: DataStateConsumer<T>,
+        onSuccess: DataStateConsumer<T>,
+    ) {
+        liveData.observe(viewLifecycleOwner) { state ->
+            if (!isAdded) return@observe
+            if (lastDataState == null && (state is DataState.Success || state is DataState.Error)) {
```
Jignesh-dimagi, line 170: There is no DataState.Cached in the condition mentioned, but the comment is referring to it. I am also not sure if we need this condition, as the cache is the first thing emitted before error and success.

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
```diff
@@ -115,47 +132,96 @@ abstract class BaseConnectFragment<B : ViewBinding> :
...
+    /**
+     * Observes a LiveData of DataState and handle UI updates for loading, success, cached, and error states.
+     */
+    protected fun <T> observeDataState(
+        liveData: LiveData<DataState<T>>,
+        onCached: DataStateConsumer<T>,
+        onSuccess: DataStateConsumer<T>,
+    ) {
+        liveData.observe(viewLifecycleOwner) { state ->
+            if (!isAdded) return@observe
```
Jignesh-dimagi: Because collectInto is viewModelScope, this condition should not be true anytime.

app/src/org/commcare/connect/repository/ConnectRepository.kt
```diff
@@ -0,0 +1,128 @@
...
+                        ConnectNetworkClient.getInstance(),
+                    ).also { instance = it }
+                }
+        }
+
+        fun getOpportunities(
+            forceRefresh: Boolean = false,
+            policy: RefreshPolicy = RefreshPolicy.SESSION_AND_TIME_BASED(),
+        ): Flow<DataState<List<ConnectJobRecord>>> =
+            offlineFirstFlow(
```
Jignesh-dimagi, line 44: Can it have other options in the future except for offlineFirstFlow?

app/src/org/commcare/connect/network/ConnectApiService.kt
```diff
@@ -0,0 +1,23 @@
...
+
+import okhttp3.ResponseBody
+import retrofit2.Response
+import retrofit2.http.GET
+import retrofit2.http.Header
+import retrofit2.http.HeaderMap
+import retrofit2.http.Path
+
+interface ConnectApiService {
+    @GET("/api/opportunity/")
```
Jignesh-dimagi: Can we use ApiEndPoints constants here?

Jignesh-dimagi:
First of all, +1 for the changes done; it looks like a lot of thought process has been given. 
Note: I was able to review only half of the PR, but it looks good so far; only a few comments.

Jignesh-dimagi (changes requested):

app/src/org/commcare/connect/repository/ConnectSyncPreferences.kt
```diff
@@ -0,0 +1,89 @@
...
+            is RefreshPolicy.SESSION_AND_TIME_BASED -> {
+                val lastSync = getLastSyncTime(endpoint) ?: return true
+                val sessionStart = getSessionStartTime()
+                val isNewSession = lastSync.before(sessionStart)
+                val ageMs = Date().time - lastSync.time
+                val isStale = ageMs >= policy.timeThresholdMs
+                isNewSession || isStale
+            }
+
+            else -> throw IllegalArgumentException("Unknown refresh policy: $policy")
```
Jignesh-dimagi: This will crash the application, is that intentional?

shubham1g5: definitely intentional as any other policy clearly signifies a programming error here.

app/src/org/commcare/connect/repository/ConnectRepository.kt
shubham1g5: I would think so depending on API.

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
shubham1g5: not at the moment, current we do want to always show the loading bar if the request is initiated from a UI component to let the user know that a refresh is in progress.

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
shubham1g5: can you give examples of use cases you are thinking of here ? Think it would be an easy add on to implement when we need it.

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
shubham1g5: > as the cache is the first thing emitted before error and success

This is a weird caveat of using view models and flows, the flow components immediately emits a "cached" state (which is the last emitted state by that flow)  when the fragment is resumed i.e if I open a fragment and that results into a `DataState.Success` as part of the network refresh and then I go out and go back in the fragment and re-subscribe to the liveData, it immediately emits the `DataState.Success` ahead of `DataState.Cached`

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
shubham1g5: https://github.com/dimagi/commcare-android/pull/3606/commits/090c7c5fa68fb25db18c1f14254b1b611fb7473c

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
shubham1g5: nice catch, https://github.com/dimagi/commcare-android/pull/3606/commits/2b3000ca416bb48dfba5949f134185bdb16af111

app/src/org/commcare/connect/network/ConnectApiService.kt
shubham1g5: https://github.com/dimagi/commcare-android/pull/3606/commits

app/src/org/commcare/connect/network/base/ConnectApiException.kt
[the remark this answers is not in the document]
conroy-ricketts: That's a good point, we could rename `ConnectApiException` to `PersonalIdOrConnectApiException` similarly to what we did with the current `PersonalIdOrConnectApiErrorCodes` class.

But also, I'm cool with this suggestion being optional. I figured the refactor would be good for consistency across our codebase since we are introducing a brand new class for exceptions.

app/src/org/commcare/connect/network/base/ConnectApiException.kt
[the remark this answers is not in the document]
conroy-ricketts: On second thought, I think there is no real benefit to making this a data class. Let's leave this as-is 👍🏾

app/src/org/commcare/connect/repository/ConnectRepository.kt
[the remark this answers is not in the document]
conroy-ricketts: Aah cool that makes sense 👍🏾

app/src/org/commcare/connect/network/ConnectApiService.kt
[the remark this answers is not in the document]
conroy-ricketts: I agree that this will be a significant scope change and I'm cool with leaving it as-is for now. But, how do you feel about creating a new backlog ticket to do this refactor? On first impression, it seems that we have two classes serving the same purpose at a high-level because they are both API services within Connect. Maybe we should also add a short code comment at the top of `ConnectApiService` explaining the difference between that and `ApiService`?

app/src/org/commcare/connect/network/ConnectNetworkHelper.kt
```diff
@@ -0,0 +1,127 @@
...
+import kotlin.coroutines.resume
+
+class ConnectNetworkHelper {
+    companion object {
+        @JvmStatic
+        fun addVersionHeader(
+            headers: HashMap<String, String>,
+            version: String,
+        ) {
+            if (version != null) {
```
conroy-ricketts: This null check is no longer necessary

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
```diff
@@ -186,4 +254,11 @@ abstract class BaseConnectFragment<B : ViewBinding> :
         networkCallback = null
         connectivityManager = null
     }
+
+    protected fun setWaitDialogEnabled(enabled: Boolean) {
+        val activity: Activity? = getActivity()
```
conroy-ricketts: Very minor nit: specifying the type here is unnecessary

  
```suggestion
        val activity = getActivity()
```

app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java
```diff
@@ -75,59 +73,32 @@ public ConnectJobsListsFragment() {
...
-                    return;
+    private void observeOpportunities() {
+        observeDataState(
+                viewModel.getOpportunities(),
+                cached -> {
+                    corruptJobs.clear();
+                    setJobListData(cached);
+                },
+                success -> {
+                    corruptJobs.clear();
```
conroy-ricketts: It seems that nothing is ever added to the corrupted jobs list but yet we still clear it. Is the `corruptJobs` list still necessary? Perhaps we can delete that variable from this class entirely?

app/unit-tests/src/org/commcare/connect/network/connect/ConnectNetworkClientTest.kt
conroy-ricketts, line 1: Can we add a couple new tests here for happy paths?

app/unit-tests/src/org/commcare/connect/repository/ConnectRepositoryTest.kt
```diff
@@ -0,0 +1,217 @@
...
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.annotation.Config
+import java.util.Date
+
+@Config(application = CommCareTestApplication::class)
+@RunWith(AndroidJUnit4::class)
+class ConnectRepositoryTest {
+    private val context = ApplicationProvider.getApplicationContext<CommCareTestApplication>()
```
conroy-ricketts: Looks like context is never used in this class


```suggestion
```

app/unit-tests/src/org/commcare/connect/viewmodel/ConnectJobsListViewModelTest.kt
```diff
@@ -0,0 +1,131 @@
...
+        val results = mutableListOf<DataState<List<ConnectJobRecord>>>()
+        viewModel.opportunities.observeForever { results.add(it) }
+
+        mainCoroutineRule.runBlockingTest {
+            viewModel.loadOpportunities()
+        }
+
+        assertEquals(2, results.size)
+        assertEquals(DataState.Loading, results[0])
+        assertTrue(results[1] is DataState.Success)
```
conroy-ricketts, line 67: Can we either use `assertEquals` both times here or use `assertTrue` both times for consistency

app/unit-tests/src/org/commcare/connect/viewmodel/ConnectJobsListViewModelTest.kt
```diff
@@ -0,0 +1,131 @@
...
+        assertEquals(jobs, (results[1] as DataState.Success).data)
+    }
+
+    @Test
+    fun testLoadOpportunities_postsError_onFailure() {
+        val cachedJobs = listOf(mockk<ConnectJobRecord>())
```
conroy-ricketts: Seems that this variable is never used

```suggestion
```

app/unit-tests/src/org/commcare/connect/viewmodel/ConnectJobsListViewModelTest.kt
```diff
@@ -0,0 +1,131 @@
...
+        mainCoroutineRule.runBlockingTest {
+            viewModel.loadOpportunities(forceRefresh = false)
+            viewModel.loadOpportunities(forceRefresh = true)
+        }
+
+        // Loading from first call, then Success from second — no stale Success from first
+        assertTrue(results.any { it is DataState.Loading })
+        assertTrue(results.last() is DataState.Success)
+        assertEquals(secondJobs, (results.last() as DataState.Success).data)
+        assertTrue(results.none { it is DataState.Success && (it as DataState.Success).data == firstJobs })
```
conroy-ricketts: Nit: the cast is redundant here

```suggestion
        assertTrue(results.none { it is DataState.Success && it.data == firstJobs })
```

conroy-ricketts (commented):
Just got done reviewing everything, it's looking great! Just had a few suggestions.

Also, I noticed that local builds are failing with these changes. It's an unresolved reference error:

```
...org/commcare/connect/ConnectJobHelper.kt:88:11 Unresolved reference 'getLearningAppProgress'.
```

app/unit-tests/src/org/commcare/connect/viewmodel/ConnectJobsListViewModelTest.kt
shubham1g5: Think using them interchangeably is pretty common and doesn't really introduce any readability concerns in the code, IMO we should be open to these sorts of small stylistic inconsistencies. Thoughts ?

conroy-ricketts: The example here is very minor and does not make a huge difference, but as a team we should be very cautious of making a habit of creating stylistic inconsistencies throughout the codebase. When our code follows the same conventions throughout, it requires a bit less cognitive load and readers are more focused on understanding what the code does vs how a particular dev writes code. There are more benefits to maintaining consistent coding style, but that reason is the biggest for me

With that being said, we don't need a coding standard or anything extreme like that for this example, but it makes sense to have back-to-back statements like this be consistent

shubham1g5: Think I agree with you in terms of using code design patterns but think putting the same sorts of limitations to be consitent for minute statement level things is often counter-productive and not a pragmatic choice. This is possibly a wider team discussion though and I will plan to bring this up in one of the team meetings.

shubham1g5: Actually while trying to edit this I realise this is more than a style inconsistency -> 

1. Using `assertEquals` instead of ` assertTrue(results[1] is DataState.Success)` is not really possible given we are not comparing 2 objects here.

2. Using `assertTrue` for ` assertEquals(DataState.Loading, results[0])` is possible to replace with `assertTrue(results[0] == DataState.Loading)`. But then we lose the failure message quality.

With `assertEquals`, when it fails we get:

Expected: DataState.Loading
Actual:   DataState.Error

With `assertTrue`, a failure just gives you:Expected: true but was false

So assertEquals is genuinely the better tool for the first assertion, and assertTrue is genuinely the better tool for the second.

app/src/org/commcare/connect/network/ConnectNetworkHelper.kt
shubham1g5: https://github.com/dimagi/commcare-android/pull/3606/commits/7bcd461c6a2adace51e916ccc4ae34a5edf1885c

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
shubham1g5: https://github.com/dimagi/commcare-android/pull/3606/commits/7bcd461c6a2adace51e916ccc4ae34a5edf1885c

app/unit-tests/src/org/commcare/connect/repository/ConnectRepositoryTest.kt
shubham1g5: https://github.com/dimagi/commcare-android/pull/3606/commits/7bcd461c6a2adace51e916ccc4ae34a5edf1885c

app/unit-tests/src/org/commcare/connect/viewmodel/ConnectJobsListViewModelTest.kt
shubham1g5: https://github.com/dimagi/commcare-android/pull/3606/commits/7bcd461c6a2adace51e916ccc4ae34a5edf1885c

app/unit-tests/src/org/commcare/connect/viewmodel/ConnectJobsListViewModelTest.kt
shubham1g5: https://github.com/dimagi/commcare-android/pull/3606/commits/7bcd461c6a2adace51e916ccc4ae34a5edf1885c

app/unit-tests/src/org/commcare/connect/network/connect/ConnectNetworkClientTest.kt
shubham1g5: https://github.com/dimagi/commcare-android/pull/3606/commits/b511c0b241666abb01f4885844d927d732b44a77

app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java
shubham1g5: https://github.com/dimagi/commcare-android/pull/3606/commits/d1069c2bd00f610d23621db28383f081f24a6670

app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java
```diff
@@ -75,61 +71,32 @@ public ConnectJobsListsFragment() {
...
-    }
-
-    private void navigateFailure() {
-        setJobListData(
-                ConnectJobUtils.getCompositeJobs(
-                        getActivity(),
-                        ConnectJobRecord.STATUS_ALL_JOBS,
-                        null
-                )
         );
```
conroy-ricketts, line 91: **_Optional_** nit: looks like we can use method references here now

```suggestion
        observeDataState(viewModel.getOpportunities(), this::setJobListData, this::setJobListData);
```

app/unit-tests/src/org/commcare/connect/network/connect/ConnectNetworkClientTest.kt
conroy-ricketts: Thanks, that looks good. Can we add just one more happy path test for `getLearningProgress()` as well (for completeness)?

app/unit-tests/src/org/commcare/connect/viewmodel/ConnectJobsListViewModelTest.kt
conroy-ricketts: Aaah got it, keeping this as-is sounds good to me 👍🏾

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
Jignesh-dimagi: I don't have anything at present but may be required to enable / disable buttons. I think we can add it to make it complete?

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
Jignesh-dimagi: Yeah, it's an issue with flow. This can go away if we split the data as a StateFlow and events as a Channel. Not sure how easy it can be with this architecture.

app/unit-tests/src/org/commcare/connect/network/connect/ConnectNetworkClientTest.kt
shubham1g5: https://github.com/dimagi/commcare-android/pull/3606/commits/4400ec6421c1cff85c2e3738676e642321a10045

app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java
shubham1g5: https://github.com/dimagi/commcare-android/pull/3606/commits/65a77d0413799a69bb7f6aec599f550895ea0c6b

conroy-ricketts (dismissed):

app/src/org/commcare/connect/repository/ConnectRepository.kt
```diff
@@ -0,0 +1,125 @@
...
+import org.commcare.connect.network.connect.models.applyToJob
+
+class ConnectRepository
+    @VisibleForTesting
+    internal constructor(
+        private val syncPrefs: ConnectSyncPreferences,
+        private val networkClient: ConnectNetworkClient,
+    ) {
+        companion object {
+            const val ENDPOINT_OPPORTUNITIES = "/opportunities"
```
OrangeAndGreen: Nit: The name and leading slash in these constants were a bit misleading to me and had me initially questioning why we're defining the URLs in a second place (other than ApiEndPoints). A couple suggestions:
1. Remove the leading slash in the constants
2. Maybe rename the constants something other than ENDPOINT? Alternatives could be API_CALL or SERVER_OP

OrangeAndGreen: A suggestion from Claude that also works pretty well: SYNC_KEY

app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java
```diff
@@ -75,61 +71,28 @@ public ConnectJobsListsFragment() {
...
-                        requireContext(),
-                        PersonalIdOrConnectApiErrorHandler.handle(requireActivity(), errorCode, t),
-                        Toast.LENGTH_LONG
-                ).show();
-                navigateFailure();
-            }
-
-            @Override
-            public void onSuccess(ConnectOpportunitiesResponseModel data) {
-                corruptJobs = data.getCorruptJobs();
```
OrangeAndGreen, line 111: Looks like the parser still parses and response model still holds corruptJobs, but we aren't displaying them anymore. This means if there's an error with the data for an opportunity that opp will not show up for the user and they won't be made aware of any problem. Actually, it doesn't look like we log any non-fatal exception to raise dev awareness of the issue either.

I'm thinking:
1. If we aren't going to show the user the corrupt jobs, should we go ahead and remove the parsing code that sorts them into a list?
2. Are we okay with this issue not being visible to the user? It's been quite a while since a corrupt job has come up but technically could happen anytime on a server bug. I'm worried that having a user's opportunity disappear without any indication might be distressing to the user.
3. Either way, seems like we should add better logging for us to detect if this happens (non-fatal exception or perhaps analytics event)

app/src/org/commcare/connect/repository/ConnectRepository.kt
shubham1g5: changed in https://github.com/dimagi/commcare-android/pull/3606/commits/72ccf2f084fe6968735de7e0f758bca90dbd0b34

app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java
shubham1g5: > If we aren't going to show the user the corrupt jobs, should we go ahead and remove the parsing code that sorts them into a list?

The only reason that's there at the moment is we are still using the opportunitities API at other places than Jobs List to decide on [connect access](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/activities/CommCareSetupActivity.java#L1009). I would think user should be allow Connect Access even if they have corrupt jobs but don't think it's super important to retain  if we toast a "Opportunities Refresh Failed" error in response when there is such an error. 


> Are we okay with this issue not being visible to the user? It's been quite a while since a corrupt job has come up but technically could happen anytime on a server bug. I'm worried that having a user's opportunity disappear without any indication might be distressing to the user.

Agree there should be some indication to the user here. I am inclined to show this as an error on Network status bar saying "Update Failed", 

> we should add better logging for us to detect if this happens (non-fatal exception or perhaps analytics event)

This already gets logged from [here](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/connect/network/connect/parser/ConnectOpportunitiesParser.kt#L38)

OrangeAndGreen: Ah, nice, I missed that non-fatal before the call to handleCorruptJob. The other two points sound good to me, agreed using the top-bar to show the user there was an error seems like the best way to go.

shubham1g5: great, addressed these in https://github.com/dimagi/commcare-android/pull/3606/commits/bd00b709f6a81dff7d3a778a4017de40eaaba9aa , https://github.com/dimagi/commcare-android/pull/3606/commits/c9de0c426d3c413e242eabc11e50994aa33fcb24 and a cleanup commit https://github.com/dimagi/commcare-android/pull/3606/commits/68b6f03354b6395df8917d3204157085245221e0 . Also confirming that I verified the changes by forcing a Json parsing error locally with code.

OrangeAndGreen: Looks great, nice getting that corrupt jobs code cleaned up and simplified again

OrangeAndGreen (dismissed):

Jignesh-dimagi:
Looks good. Hoping this is our base for other API calls also, as this is more systematic and has better UX experience.

Jignesh-dimagi (dismissed):

shubham1g5:
Updating that I did some minor UI tweaks in last commit - https://github.com/dimagi/commcare-android/pull/3606/commits/d393c80c70ec241dcbd8f75319f3b63e1495f24e and have added error screenshots in PR description to show the result of the change

OrangeAndGreen (approved):

PR #3653, opened 2026-04-08
Limit the number of attachments per form entry

avazirna:
## Product Description
Adds a per-form attachment limit (currently set to 50). When a user tries to add a new attachment beyond the limit — via camera, audio recording, file picker, or any media widget — they see an error message asking them to delete an existing attachment first:
image

JIRA ticket: https://dimagi.atlassian.net/browse/SAAS-19091

## Technical Summary
- Added `attachmentCount` tracker and `MAX_ATTACHMENTS` constant in `FormEntryActivity`
- Added `canAddAttachment()`, `incrementAttachmentCount()`, and `decrementAttachmentCount()` methods to `FormEntryActivity`
- Introduces a `MediaCapableWidget` interface in `QuestionWidget` with `getMediaName()`, `setMediaName()`, and `clearMediaData()` methods to standardize media attachment tracking across widget types
- `FormEntryActivity` maintains an in-memory attachment count, initialized from the instance folder on form load via `FileUtil.countNumberOfMediaFilesInDirectory()`
- Attachment limit checks are enforced in button click handlers (capture/choose) across all media widget types: `Audio`, `Video`, `Document`, `Image`, `Signature`
- Widgets that already have an attachment are allowed to replace it even at the limit (checked via getMediaName() == null)

### Safety story
  - The limit (50) is generous for normal use and only prevents extreme cases
  - The check happens before launching intents, so no work is lost — the user is informed before capturing/choosing
  - Attachment count is initialized from the file system on form load, ensuring correctness for resumed forms
  - Replacing existing attachments works at the limit since `isAttachmentLimitReached()` checks `getMediaName()` == null

### Automated test coverage
Unit tests needed for `canAddAttachment()` boundary conditions and widget rejection behavior.

### QA Plan
- Fill a form with multiple media questions, verify attachments work normally under the limit
- Attempt to exceed 50 attachments, verify the error message appears
- Verify replacing an existing attachment still works at the limit
- Verify audio recording (native CommCareAudioWidget) respects the limit

## Labels and Review

- [x] Do we need to enhance the manual QA test coverage ? If yes,  [RELEASES.md](../RELEASES.md) is updated accordingly
- [x] Does the PR introduce any major changes worth communicating ? If yes,  [RELEASES.md](../RELEASES.md) is updated accordingly


avazirna:
@coderabbitai full review

app/res/values/strings.xml
```diff
@@ -703,4 +703,6 @@
     <string name="network_notification_service_starting">Starting network service</string>
     <string name="network_notification_service_running">Running network requests...</string>
     <string name="network_notification_service_downloading_files">Downloading application files...</string>
+    <string name="form_attachment_limit_reached_title">Attachment limit reached</string>
+    <string name="form_attachment_limit_reached">This form has reached the maximum of %s attachments. To add a new one, remove an existing attachment first.</string>
```
conroy-ricketts: Should use `%d` here

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -429,6 +433,29 @@ public void showFileOversizeError() {
...
+    }
+
+    public void showFormAttachmentLimitReachedError() {
+        String title = StringUtils.getStringRobust(this, R.string.form_attachment_limit_reached_title);
+        String msg = StringUtils.getStringRobust(this, R.string.form_attachment_limit_reached,
+                String.valueOf(MAX_ATTACHMENTS));
+        CommCareAlertDialog dialog = StandardAlertDialog.getBasicAlertDialog(
+                title, msg, (dialog1, which) -> dialog1.dismiss());
+        showAlertDialog(dialog);
+    }
```
conroy-ricketts, line 460: Can we add a quick screenshot of what this UI looks like to the PR description

app/src/org/commcare/utils/FileUtil.java
```diff
@@ -992,4 +993,28 @@ public static boolean isFileTooLargeToUpload(ContentResolver contentResolver, Ur
...
+                                !file.getName().endsWith(XML_EXTENSION) &&
+                                (FormUploadUtil.isSupportedMultimediaFile(file.getName()) ||
+                                        file.getName().endsWith(MediaWidget.AES_EXTENSION))) {
+                            count++;
+                        }
+                    }
+                }
+            }
+        } catch (Exception e) {
+            // log non-fatal exception and return count as 0
```
conroy-ricketts: Is there a specific reason for doing it this way vs returning `null` for the count?

app/src/org/commcare/views/widgets/ImageWidget.java
```diff
@@ -219,6 +225,17 @@ public ImageWidget(final Context context, FormEntryPrompt prompt, PendingCallout
...
+    @Override
+    public String getMediaName() {
+        return mBinaryName;
+    }
+
+    @Override
+    public void setMediaName(String mediaName) {
+        mBinaryName = mediaName;
+        incrementAttachmentCount();
+    }
```
conroy-ricketts, line 237: This setter function should follow the single responsibility principle and only be responsible for setting the media name. Is there a better place where we can call `incrementAttachmentCount()`?

app/src/org/commcare/views/widgets/MediaWidget.java
```diff
@@ -208,14 +212,23 @@ private boolean ifMediaExtensionChecks(String binaryPath) {
...
+        clearMediaData();
         togglePlayButton(false);
     }
 
-    private void deleteMedia() {
+
+    @Override
+    public void setMediaName(String mediaName) {
+        mBinaryName = mediaName;
+        incrementAttachmentCount();
```
conroy-ricketts, line 223: Similar concern here

app/src/org/commcare/views/widgets/QuestionWidget.java
```diff
@@ -211,6 +233,30 @@ public void updateDrawState(TextPaint ds) {
...
 
+    protected boolean isAttachmentLimitReached() {
+        if (getContext() instanceof FormEntryActivity activity && this instanceof MediaCapableWidget mediaCapableWidget) {
+            if (mediaCapableWidget.getMediaName() == null && !activity.canAddAttachment()) {
+                activity.showFormAttachmentLimitReachedError();
+                return true;
+            }
+        }
+        return false;
+    }
```
conroy-ricketts: On first glance this function seems to only be responsible for checking if the limit is reached, but in reality it is also showing the error message. Maybe this simplest fix is to rename this function?

```suggestion
    protected boolean showAttachmentErrorIfLimitReached() {
        if (getContext() instanceof FormEntryActivity activity && this instanceof MediaCapableWidget mediaCapableWidget) {
            if (mediaCapableWidget.getMediaName() == null && !activity.canAddAttachment()) {
                activity.showFormAttachmentLimitReachedError();
                return true;
            }
        }
        return false;
    }
```

app/src/org/commcare/views/widgets/SignatureWidget.java
```diff
@@ -139,10 +143,26 @@ public SignatureWidget(Context context, FormEntryPrompt prompt, PendingCalloutIn
...
 
+    @Override
+    public String getMediaName() {
+        return mBinaryName;
+    }
+
+    @Override
+    public void setMediaName(String mediaName) {
+        mBinaryName = mediaName;
+        incrementAttachmentCount();
```
conroy-ricketts, line 163: Same concern here as mentioned above

app/src/org/commcare/views/widgets/ImageWidget.java
avazirna: Yeah, the name needs work. This method is from the `MediaCapableWidget` interface, it's not intended to be a _setter_, it's to enforce the need to increment the number of attachments whenever a new binary is registered.

conroy-ricketts: Aaah ok, yeah refactoring the function name here makes sense to me 👍🏾

app/res/values/strings.xml
avazirna: I'm calling this from `StringUtils.getStringRobust()` which is localization-aware, however this method only takes `String` args.

app/src/org/commcare/views/widgets/QuestionWidget.java
avazirna: [0a9b1ff](https://github.com/dimagi/commcare-android/pull/3653/commits/0a9b1ff7543068aeb7c145b974be7de58e37dfe6)

app/src/org/commcare/utils/FileUtil.java
avazirna: This is to act as a soft blocker when the counting errors for some reason. I've refactored this method to distinguish between an empty folder and errors by returning `-1`

app/res/values/strings.xml
conroy-ricketts: Ah sounds good 👍🏾

conroy-ricketts (dismissed):

app/res/values/strings.xml
shubham1g5: both of these should define `cc:translatable="true" as we would want these on HQ translations.

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -1501,6 +1536,9 @@ private boolean formHasLoaded() {
     private void loadStateFromBundle(Bundle savedInstanceState) {
         if (savedInstanceState != null) {
             instanceState.loadState(savedInstanceState);
+            if (savedInstanceState.containsKey(KEY_NUM_FORM_ATTACHMENTS)) {
```
shubham1g5, line 1542: nit: you do not need the containsKey check given the `getInt` provides a fallback below already

app/src/org/commcare/utils/FileUtil.java
```diff
@@ -992,4 +993,36 @@ public static boolean isFileTooLargeToUpload(ContentResolver contentResolver, Ur
...
+            }
+            File[] files = dir.listFiles();
+            if (files == null) {
+                return -1;
+            }
+            for (File file : files) {
+                if (file.isFile() &&
+                        !file.getName().endsWith(XML_EXTENSION) &&
+                        (FormUploadUtil.isSupportedMultimediaFile(file.getName()) ||
+                                file.getName().endsWith(MediaWidget.AES_EXTENSION))) {
```
shubham1g5: This is likely to break silently the moment we decide to encrypt a non media file and keep it's extension as `.aes`.

app/src/org/commcare/views/widgets/QuestionWidget.java
```diff
@@ -67,6 +68,25 @@
 import javax.annotation.Nullable;
 
 public abstract class QuestionWidget extends LinearLayout implements QuestionExtensionReceiver {
+
+    /**
+     * Interface for widgets that have media attachments, to allow the base QuestionWidget to check if the
+     * attachment limit has been reached
+     */
+    protected interface MediaCapableWidget {
```
shubham1g5, line 76: nice!

app/src/org/commcare/views/widgets/QuestionWidget.java
```diff
@@ -211,6 +231,30 @@ public void updateDrawState(TextPaint ds) {
         }
     }
 
+    protected boolean warnIfAttachmentLimitReached() {
+        if (getContext() instanceof FormEntryActivity activity && this instanceof MediaCapableWidget mediaCapableWidget) {
```
shubham1g5, line 235: think this would always be true - `getContext() instanceof FormEntryActivity activity`

shubham1g5 (changes requested):
Nothing too major although the PR must be complemented with tests as it's unreasonable to ask QA to test a form by filling 50 attachements to a page as part of our releases.

app/res/values/strings.xml
avazirna: [c8b7d35](https://github.com/dimagi/commcare-android/pull/3653/commits/c8b7d3576b3dc349939a612fc0d719b3bf3dd651)[c8b7d35](https://github.com/dimagi/commcare-android/pull/3653/commits/c8b7d3576b3dc349939a612fc0d719b3bf3dd651)

app/src/org/commcare/utils/FileUtil.java
avazirna: [fe764d3](https://github.com/dimagi/commcare-android/pull/3653/commits/fe764d3868a91d235375a425fe496143f87216ef)[fe764d3](https://github.com/dimagi/commcare-android/pull/3653/commits/fe764d3868a91d235375a425fe496143f87216ef)

avazirna:
> 0e84a18



> Nothing too major although the PR must be complemented with tests as it's unreasonable to ask QA to test a form by filling 50 attachements to a page as part of our releases.

I added a test around this feature, and reduced the limit to 5. My QA note was in the sense that QA would do a once-off test around this feature.

shubham1g5 (dismissed):

shubham1g5 (approved):

PR #3655, opened 2026-04-10
CCCT-2204 - Add email - tech spec, implementation plan and its tickets

Jignesh-dimagi:
## Product Description
Users are presented with a screen to add / validate their email during the sign-up and recovery flow. These screens are optional, and users can skip them. For the legacy users, they are prompt to add the emails.

## Technical Summary
**### [CCCT-2204](https://dimagi.atlassian.net/browse/CCCT-2204)**

This task was taken during AI week to have a tech spec and implementation plan for adding email during the user sign-up / recovery flow. 

Design Doc [here](https://docs.google.com/document/d/1Vc3DV_44XsgIVgrMf1daarpqg2nxXB6tDG0i0UaodsU/edit?tab=t.0#heading=h.4h92hp66c5wz).

This plan was created with Claude code after working with it in iterative mode. Enough context was provided to ensure the implementation plan is realistic, and if wanted, we can ask Claude to write code to directly implement it without the dev writing any code. 

This plan has also created the sub-tickets to break down the task. We can also ask Claude to work on each ticket and give a corresponding PR.

The server-side endpoints `/users/send_email_otp` and `/users/verify_email_otp` need to be confirmed with the backend team before finalising.

### Safety story
Unit tests are included in the implementation plan

## Labels and Review

[CCCT-2204]: https://dimagi.atlassian.net/browse/CCCT-2204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/superpowers/plans/superpowers-add-email-to-personalid-signup.md
+++ b/docs/superpowers/plans/superpowers-add-email-to-personalid-signup.md
@@ -1,0 +1,1962 @@
+# Add Email to PersonalID Signup / Recovery Flow — Implementation Plan
+
+> **For agentic workers:** REQUIRED: Use superpowers:subagent-driven-development (if subagents available) or superpowers:executing-plans to implement this plan. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Insert an optional email-entry + OTP-verification step into the PersonalID signup flow (between Name and Backup Code), and offer a non-blocking email-collection prompt to existing users who completed signup without an email.
+
+**Architecture:** New Kotlin fragments (`PersonalIdEmailFragment`, `PersonalIdEmailVerificationFragment`) extend the existing `BasePersonalIdFragment` pattern. Email state is stored in `PersonalIdSessionData` during signup and persisted to `ConnectUserRecord` (new DB columns) upon verification. The legacy prompt uses a two-offer-with-30-day-gap policy tracked via `emailOfferCount` (int) and `lastEmailOfferDate` (Date) on `ConnectUserRecord`.
+
+**Tech Stack:** Kotlin (new files), Java (existing files modified), AndroidX Navigation Safe Args, Retrofit 2 / OkHttp, SQLCipher (versioned migration), Robolectric unit tests.
+
+---
+
+## File Structure
+
+### Files to Create
+| File | Purpose |
+|------|---------|
+| `app/src/org/commcare/android/database/connect/models/ConnectUserRecordV24.java` | Snapshot of v24 ConnectUserRecord (for DB migration) |
+| `app/src/org/commcare/fragments/personalId/PersonalIdEmailFragment.kt` | Email entry UI fragment |
+| `app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt` | Email OTP verification UI fragment |
+| `app/res/layout/fragment_personalid_email.xml` | Layout for email entry screen |
+| `app/res/layout/fragment_personalid_email_verification.xml` | Layout for email OTP screen |
+| `app/src/org/commcare/connect/network/connectId/parser/SendEmailOtpResponseParser.kt` | Parses send-email-OTP server response |
+| `app/src/org/commcare/connect/network/connectId/parser/VerifyEmailOtpResponseParser.kt` | Parses verify-email-OTP server response |
+| `app/unit-tests/src/org/commcare/connect/PersonalIdEmailOfferTest.kt` | Unit tests for evaluateEmailOffer logic |
+| `app/unit-tests/src/org/commcare/connect/PersonalIdEmailValidationTest.kt` | Unit tests for client-side email format validation |
+
+### Files to Modify
+| File | Change |
+|------|--------|
+| `app/src/org/commcare/android/database/connect/models/ConnectUserRecord.java` | Add email, emailVerified, emailOfferCount, lastEmailOfferDate fields |
+| `app/src/org/commcare/android/database/connect/models/PersonalIdSessionData.kt` | Add email and emailVerified fields |
+| `app/src/org/commcare/fragments/personalId/PersonalIdPhotoCaptureFragment.java` | Read email/emailVerified from session data when constructing ConnectUserRecord |
+| `app/src/org/commcare/fragments/personalId/PersonalIdBackupCodeFragment.java` | Read email/emailVerified from session data in handleSuccessfulRecovery() |
+| `app/src/org/commcare/models/database/connect/DatabaseConnectOpenHelper.java` | Bump CONNECT_DB_VERSION to 25 |
+| `app/src/org/commcare/models/database/connect/ConnectDatabaseUpgrader.java` | Add upgradeTwentyFourTwentyFive() |
+| `app/src/org/commcare/connect/network/ApiEndPoints.java` | Add sendEmailOtp and verifyEmailOtp endpoints |
+| `app/src/org/commcare/connect/network/ApiService.java` | Add sendEmailOtp and verifyEmailOtp service methods |
+| `app/src/org/commcare/connect/network/ApiPersonalId.java` | Add sendEmailOtp() and verifyEmailOtp() static methods |
+| `app/src/org/commcare/connect/network/connectId/PersonalIdApiHandler.java` | Add sendEmailOtpCall() and verifyEmailOtpCall() |
+| `app/res/navigation/nav_graph_personalid.xml` | Add email and email-OTP destinations; reroute Name → Email |
+| `app/src/org/commcare/fragments/personalId/PersonalIdNameFragment.java` | Update navigate action to go to email instead of backup code |
+| `app/src/org/commcare/activities/connect/PersonalIdActivity.java` | Handle EXTRA_LEGACY_EMAIL_FLOW intent extra |
+| `app/src/org/commcare/connect/PersonalIdManager.java` | Add checkEmailCollection() and evaluateEmailOffer() |
+| `app/src/org/commcare/activities/LoginActivity.java` | Call checkEmailCollection() after PersonalID login |
+
+---
+
+## Chunk 1: Database Layer
+
+### Task 1: Create ConnectUserRecordV24 (migration snapshot)
+
+**Files:**
+- Create: `app/src/org/commcare/android/database/connect/models/ConnectUserRecordV24.java`
+
+- [ ] **Step 1.1: Write the failing unit test**
+
+  Create `app/unit-tests/src/org/commcare/connect/ConnectUserRecordMigrationV24Test.kt`:
+
+  ```kotlin
+  package org.commcare.connect
+
+  import org.junit.Assert.assertEquals
+  import org.junit.Assert.assertNull
+  import org.junit.Test
+
+  class ConnectUserRecordMigrationV24Test {
+
+      @Test
+      fun `fromV24 copies all fields and sets email defaults`() {
+          val old = ConnectUserRecordV24().apply {
+              // Verify V24 compiles with fields 1-16 only (no email fields)
+          }
+          val new = ConnectUserRecord.fromV24(old)
+          assertNull(new.email)
+          assertEquals(false, new.emailVerified)
+          assertNull(new.emailOfferDate1)
+          assertNull(new.emailOfferDate2)
+      }
+  }
+  ```
+
+  Run: `./gradlew testCommcareDebug --tests "org.commcare.connect.ConnectUserRecordMigrationV24Test"`
+  Expected: FAIL — `ConnectUserRecordV24` and `ConnectUserRecord.fromV24` don't exist yet.
+
+- [ ] **Step 1.2: Create ConnectUserRecordV24.java**
+
+  This is a verbatim copy of the current `ConnectUserRecord.java` with the class name changed and `STORAGE_KEY` kept. It will only ever be read, not written.
+
+  ```java
+  package org.commcare.android.database.connect.models;
+
+  import org.commcare.android.storage.framework.Persisted;
+  import org.commcare.connect.ConnectConstants;
+  import org.commcare.models.framework.Persisting;
+  import org.commcare.modern.database.Table;
+  import org.commcare.modern.models.MetaField;
+  import java.util.Date;
+
+  @Table(ConnectUserRecordV24.STORAGE_KEY)
+  public class ConnectUserRecordV24 extends Persisted {
+      public static final String STORAGE_KEY = "user_info";
+      public static final String META_PIN = "pin";
+
+      @Persisting(1) private String userId;
+      @Persisting(2) private String password;
+      @Persisting(3) private String name;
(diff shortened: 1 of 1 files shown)


shubham1g5:
> The server-side endpoints /users/send_email_otp and /users/verify_email_otp need to be confirmed with the backend team before finalising.

@Jignesh-dimagi are you able to add a web ticket for these 2 in the current epic and define the payload and request data for these in that ticket.

docs/superpowers/plans/superpowers-add-email-to-personalid-signup.md
```diff
@@ -0,0 +1,1962 @@
...
+          }.sendEmailOtpCall(requireActivity(), email, personalIdSessionData)
+      }
+
+      private fun skipEmail() {
+          FirebaseAnalyticsUtil.reportPersonalIDContinueClicked(javaClass.simpleName, "skip")
+          // Mark that the email step was explicitly shown and declined during signup.
+          // createAndSaveConnectUser() will read this to initialise emailOfferCount = 1
+          // so the post-login dialog does not fire again immediately.
+          personalIdSessionData.emailSkippedDuringSignup = true
+          if (isLegacyFlow) {
```
shubham1g5: Why is this getting skipped for legacy users ? If so, why do even open this fragment for these users ?

docs/superpowers/plans/superpowers-add-email-to-personalid-signup.md
```diff
@@ -0,0 +1,1962 @@
...
+          binding.emailTextValue.requestFocus()
+      }
+
+      private fun createEmailWatcher() = object : TextWatcher {
+          override fun beforeTextChanged(s: CharSequence?, start: Int, count: Int, after: Int) {}
+          override fun onTextChanged(s: CharSequence?, start: Int, before: Int, count: Int) {
+              enableContinueButton(isValidEmail(s?.toString()))
+          }
+          override fun afterTextChanged(s: Editable?) {}
+      }
```
shubham1g5, line 1082: can we define error state here for when the email is not valid, think we would like to show an error text to the user in those cases saying "email is invalid" ?

docs/superpowers/plans/superpowers-add-email-to-personalid-signup.md
```diff
@@ -0,0 +1,1962 @@
...
+              personalIdSessionData.getOauthPassword(),
+              personalIdSessionData.getUserName(),
+              String.valueOf(personalIdSessionData.getBackupCode()),
+              new Date(), photoAsBase64,
+              personalIdSessionData.getDemoUser(),
+              personalIdSessionData.getRequiredLock(),
+              personalIdSessionData.getInvitedUser());
+      // Carry over email verification state from the signup flow
+      user.setEmail(personalIdSessionData.getEmail());
+      user.setEmailVerified(personalIdSessionData.getEmailVerified());
```
shubham1g5: How about setting these in user preferences instead given these are programmatical states and resets on every login

docs/superpowers/plans/superpowers-add-email-to-personalid-signup.md
```diff
@@ -0,0 +1,1962 @@
...
+      if (personalIdSessionData.getEmailSkippedDuringSignup()) {
+          // User was shown the email step and declined. Count it as the first offer so the
+          // post-login dialog only fires after the 30-day gap, not on the very next login.
+          user.setEmailOfferCount(1);
+          user.setLastEmailOfferDate(new Date());
+      }
```
shubham1g5: think we can just store the last login timstamp here instead and can use it to compare the 30 day gap when prompting. We should also remove the `setEmailOfferCount` as we can assume it to be `1` post-login if it's not there.

shubham1g5 (commented):
Looks good overall, Think we need to create implementation tickets on this and ensure smaller PRs. Probably a good functional split of tickets on this work would be - 

Mobile:
- DB changes to store email
- network layer changes 
- Email Page along with validations and keyboard handling
- Email verification page along with account creation and navigation flow
- Email Prompts reminders

Web:
- Probably create a spec ticket for web to dig deeper on this. We should provide the API payloads for this and ask web to spec out the 
1. API implementations
2. Actual Implmentation to generate and send the otp to user email.

docs/superpowers/plans/superpowers-add-email-to-personalid-signup.md
Jignesh-dimagi: @shubham1g5 You mean the app should show the error message when the user is still typing?

docs/superpowers/plans/superpowers-add-email-to-personalid-signup.md
Jignesh-dimagi: @shubham1g5 It's easier to handle when the user logs out instead of clearing individual preferences.

docs/superpowers/plans/superpowers-add-email-to-personalid-signup.md
Jignesh-dimagi: It's like the user has agreed to open the email fragment from the reminder prompt for the legacy user but doesn't want to complete it, so they pressed skip.

shubham1g5: where are they pressing skip in this workflow ?

docs/superpowers/plans/superpowers-add-email-to-personalid-signup.md
shubham1g5: yeah that's what I meant, thoughts ?

docs/superpowers/plans/superpowers-add-email-to-personalid-signup.md
shubham1g5: are not we resetting these properties anyway individually on re-login ?

Also you can have a new user specific preference class `PersonalIDPreferences` that we clear as a whole on user signing out.

docs/superpowers/plans/superpowers-add-email-to-personalid-signup.md
conroy-ricketts: How about adding a debounce here that only shows the error after the user stopped typing for something like 3 seconds? So that way the message is not intrusive when the user knows what to type, but also serves as a hint if the user gets stuck

conroy-ricketts (dismissed):

conroy-ricketts (commented):
Just to confirm, this plan is **_not_** going to be merged into `master` right?

Jignesh-dimagi:
> Just to confirm, this plan is **_not_** going to be merged into `master` right?

Yes, that's correct.

docs/superpowers/plans/superpowers-add-email-to-personalid-signup.md
Jignesh-dimagi: I think we are already taking care of it by not enabling the continue button in case there is a wrong email format. This is the same as the phone number screen.

conroy-ricketts: It may still be beneficial to show the user why the continue button is disabled, but I have no strong opinions here and I'm cool with whatever 👍🏾

docs/superpowers/plans/superpowers-add-email-to-personalid-signup.md
Jignesh-dimagi: Users can skip while coming from the sign-up / recovery flow or legacy user prompt.

docs/superpowers/plans/superpowers-add-email-to-personalid-signup.md
Jignesh-dimagi: That can be a possibility by creating a class, but I'm not sure what the advantage is of using a preference rather than having it in a DB. I was thinking of having everything at a single source. Your thoughts?

shubham1g5: The major advantage is to have the DB layer clean with only the persistent properties, something we are able to recover from server. Plus, Putting app UI states into the same table as server data can make this table very populated and un-readable over the course of time.

docs/superpowers/plans/superpowers-add-email-to-personalid-signup.md
shubham1g5: @Jignesh-dimagi Got it, though I am still struggling to understand this flow with this context, can you please provide a more detailed and wholistic response here on what the user flow is coming from Sign up  vs the Reminder Prompt after sign in, and how does `isLegacyFlow` gets determined for each of those flows in combination with user skip functionality, what's the UI way to skip this for user etc.  Feel free to point me to relevant section in the plan if you think that info is already there.

docs/superpowers/plans/superpowers-add-email-to-personalid-signup.md
shubham1g5: Added a guidance file on the same here - https://github.com/dimagi/open-source/pull/21/files that is open to discussion.

docs/superpowers/plans/superpowers-add-email-to-personalid-signup.md
Jignesh-dimagi: @shubham1g5 you can find [here](https://github.com/dimagi/commcare-android/blob/ccct-2204-add-email-in-signup-recovery/docs/superpowers/plans/superpowers-add-email-to-personalid-signup.md#chunk-6-legacy-user-email-prompt). Basically, users will be prompt to add the email, if not already, every 30 days.

shubham1g5: that doesn't detail the "Skip" functionality you are mentioning here

Jignesh-dimagi: Aha, ok. Let me put it here. The user has chosen to add email from the prompt and lands on the 'add email' fragment. The user now changed their mind to add it later, so it can be done by pressing skip, and it will close that activity.

shubham1g5: Is that different from Back navigation ? I don't think we want to give user a "Skip Button" on UI. Also `isLegacyFlow` seems like a very confusing term for that workflow

Jignesh-dimagi: Some more details [here](https://github.com/dimagi/commcare-android/blob/ccct-2204-add-email-in-signup-recovery/docs/superpowers/plans/superpowers-add-email-to-personalid-signup.md#ticket-3--email-entry-fragment).

shubham1g5: Don't think that answers the feedback on not having "Skip button". Do you know where is the skip button coming from ? Was it mentioned in the design ?

Jignesh-dimagi: I think I was just trying to provide you more details before seeing your latest response (we both typed at the same time), which was adding to my previous comment.

Now, coming to the skip button, it has come as email is optional for both, i.e., new sign-ups or legacy users as defined in the [design doc](https://docs.google.com/document/d/1Vc3DV_44XsgIVgrMf1daarpqg2nxXB6tDG0i0UaodsU/edit?tab=t.0) under the headings `Goals` and  `Solution Summary`. In order to keep the design the same for both cases, the Skip button is always present on the email entry fragment.

I am not sure what your query is here. Email entry fragments shouldn't have the `Skip` button at all?

docs/superpowers/plans/superpowers-add-email-to-personalid-signup.md
Jignesh-dimagi: Ok, I think this makes sense. I will update the plan for using the shared preferences instead of DB.

docs/superpowers/plans/superpowers-add-email-to-personalid-signup.md
shubham1g5: > I am not sure what your query is here. Email entry fragments shouldn't have the Skip button at all?

Yes that was  a followup based on your responses, It makes sense to have a skip button based on the design doc linked. 

Although my initial query was why do we need `isLegacyFlow` ? That I think is still un-answered as that button doesn't really related to user clicking "Skip" as `isLegacyFlow` seems to be being determined by fragment arguments not by skip click.

Jignesh-dimagi: > Although my initial query was why do we need `isLegacyFlow` ? That I think is still un-answered as that button doesn't really related to user clicking "Skip" as `isLegacyFlow` seems to be being determined by fragment arguments not by skip click.

Yeah, it's driven by fragment arguments because it defines where the user is coming from.

If coming from a personalid name fragment → email entry fragment → skip press → backup code fragment.

If coming from a legacy user prompt → email entry fragment → skip press → close the activity.

shubham1g5: I see, it's finally making sense to me. Thanks!

docs/superpowers/plans/superpowers-add-email-to-personalid-signup.md
```diff
@@ -0,0 +1,2045 @@
...
+  ```
+
+- [ ] **Step 8.2: Run the unit test to verify it passes**
+
+  Run: `./gradlew testCommcareDebug --tests "org.commcare.connect.PersonalIdEmailOfferTest"`
+  Expected: PASS
+
+- [ ] **Step 8.3: Call checkEmailCollection in LoginActivity**
+
+  Find the location in `LoginActivity` where `PersonalIdManager.getInstance().handleFinishedActivity(...)` is called. Immediately after that call, add:
```
shubham1g5: It doesn't seem right to prompt user just after sign up completes, given the user probably just skipped this in the signup flow.

Jignesh-dimagi: No, it will not show the prompt immediately to users who have just done the sign-up. The dialog fires 30 days later.

shubham1g5: make sense but the bigger concern I wanted to highlight with that stemement is this dialog only fires after user comes from the signup i.e we are requiring user to complete the sign up flow at the moment to fire this dialog when we want to fire this dialog independenly of user sign up just based on the 30-day period.

Jignesh-dimagi: Yeah, good catch. That was the bug in the implementation plan, and I am working on it. I was not able to decide the best place to show this add email dialogue. Should it be a standard home activity or a connected activity (opportunity listing page)?

shubham1g5: Think App home should work well here, but we should update the design doc to reflect that decision.

conroy-ricketts:
@shubham1g5 @Jignesh-dimagi We're good to close this PR now right?

PR #3658, opened 2026-04-15
Add mobile UI styleguide

mmmary314:
## Summary
- Adds a self-contained HTML styleguide (`docs/mobile-styleguide.html`) covering colors, typography, components, and usage guidelines for new UI screens.

## Test plan
- [ ] Open `docs/mobile-styleguide.html` in a browser and verify all sections render correctly

### DIFF
--- a/docs/mobile-styleguide.html
+++ b/docs/mobile-styleguide.html
@@ -1,0 +1,777 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+<meta charset="UTF-8">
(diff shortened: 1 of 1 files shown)


shubham1g5:
@mmmary314 I doubt we will be able to use this, mobile layouts are xml layout and we will instead need to built these in into our Android styles.

conroy-ricketts:
The style guide itself looks great! But it may be best to keep any style guide contained to a place that is more accessible to UI/UX designers which will be more maintainable in the future. The mocks in Figma may be adequate for us

PR #3659, opened 2026-04-15
Move incomplete forms count off the main thread

avazirna:
## Product Description

No user-facing changes. The home screen incomplete forms button now loads its count asynchronously, preventing potential UI jank from the database query.

## Technical Summary

The `StorageUtils.getNumIncompleteForms()` database query in `HomeButtons` was running synchronously on the main thread during home screen button setup. This moves it off the main thread using `LatestTaskExecutor` (coroutine-based) tied to the activity's lifecycle scope.

- Extract `updateIncompleteFormsUI()` helper method for updating the button text
- Add lazy-initialized `LatestTaskExecutor<Integer>` for async execution
- Replace synchronous `getNumIncompleteForms()` call with async `submit()` via `Dispatchers.IO`
- `SessionUnavailableException` is silently ignored in `onError` since it indicates expected session expiry, not a real error

### Safety story
- The coroutine is scoped to the activity's lifecycle via `LifecycleOwnerKt.getLifecycleScope()`, so it's automatically cancelled on activity destruction
- Additional `isFinishing() || isDestroyed()` guard in the callback as a belt-and-suspenders check
- `LatestTaskExecutor` cancels any previous in-flight task on re-submit, preventing stale results
- Button text defaults to the plain "Incomplete Forms" label from `cardDisplayData.text` while the async count loads
- This change was successfully tested locally. Steps taken were:
  - Verified that the incomplete forms button shows the correct count on the home screen
  - Submitted and deleted incomplete forms and returned to home screen, verified that the count updated accordingly
  - Rotated the device on the home screen, and verified that there were no crashes and count still appears

## Labels and Review

- [ ] Do we need to enhance the manual QA test coverage? If yes, [RELEASES.md](../RELEASES.md) is updated accordingly
- [ ] Does the PR introduce any major changes worth communicating? If yes, [RELEASES.md](../RELEASES.md) is updated accordingly

### DIFF
--- a/app/src/org/commcare/activities/HomeButtons.java
+++ b/app/src/org/commcare/activities/HomeButtons.java
@@ -206,23 +206,26 @@ public class HomeButtons {
 
     private static TextSetter getIncompleteButtonTextSetter(final StandardHomeActivity activity) {
         return (cardDisplayData, squareButtonViewHolder, context, notificationText) -> {
-            int numIncompleteForms;
-            try {
-                numIncompleteForms = StorageUtils.getNumIncompleteForms();
-            } catch (SessionUnavailableException e) {
-                // stop button setup, since redirection to login is imminent
-                return;
-            }
+            getIncompleteFormsExecutor().submit(
+                    LifecycleOwnerKt.getLifecycleScope(activity),
+                    StorageUtils::getNumIncompleteForms,
+                    new LatestTaskExecutor.Callback<>() {
+                        @Override
+                        public void onResult(Integer numIncompleteForms) {
+                            if (activity.isFinishing() || activity.isDestroyed()) {
+                                return;
+                            }
+                            updateIncompleteFormsUI(activity, squareButtonViewHolder.textView, numIncompleteForms);
+                        }
 
-            if (numIncompleteForms > 0) {
-                Spannable incompleteIndicator =
-                        (activity.localize("home.forms.incomplete.indicator",
-                                new String[]{String.valueOf(numIncompleteForms),
-                                        Localization.get("home.forms.incomplete")}));
-                squareButtonViewHolder.textView.setText(incompleteIndicator);
-            } else {
-                squareButtonViewHolder.textView.setText(activity.localize("home.forms.incomplete"));
-            }
+                        @Override
+                        public void onError(@NonNull Exception exception) {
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/activities/HomeButtons.java
```diff
@@ -173,25 +181,51 @@ private static View.OnClickListener getIncompleteButtonListener(final StandardHo
...
-                        (activity.localize("home.forms.incomplete.indicator",
-                                new String[]{String.valueOf(numIncompleteForms),
-                                        Localization.get("home.forms.incomplete")}));
-                squareButtonViewHolder.textView.setText(incompleteIndicator);
-            } else {
-                squareButtonViewHolder.textView.setText(activity.localize("home.forms.incomplete"));
-            }
             squareButtonViewHolder.textView.setTextColor(context.getResources()
                     .getColor(cardDisplayData.textColor));
             squareButtonViewHolder.subTextView.setVisibility(View.GONE);
```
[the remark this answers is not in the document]
avazirna: This is already the `default` value. I'd rather leave the previous value before getting a new value one, instead of setting the default value.

conroy-ricketts (approved):
LGTM but good to have a second set of eyes before merge

app/src/org/commcare/activities/HomeButtons.java
```diff
@@ -173,25 +181,51 @@ private static View.OnClickListener getIncompleteButtonListener(final StandardHo
...
-                return;
-            }
+            getIncompleteFormsExecutor().submit(
+                    LifecycleOwnerKt.getLifecycleScope(activity),
+                    StorageUtils::getNumIncompleteForms,
+                    new LatestTaskExecutor.Callback<>() {
+                        @Override
+                        public void onResult(Integer numIncompleteForms) {
+                            if (activity.isFinishing() || activity.isDestroyed()) {
+                                return;
```
shubham1g5, line 216: can this be true given you are scoping this to activity lifecycle ?

shubham1g5 (approved):

app/src/org/commcare/activities/HomeButtons.java
avazirna: Yes, if `onResult` is called while the activity is finishing. I believe the `kotlin` cancelation logic is triggered at the activity's onDestroy.

PR #3676, opened 2026-04-23
Upgraded Firebase Auth dependency to 23.2.0

OrangeAndGreen:
https://dimagi.atlassian.net/browse/CI-636

## Product Description
No user-facing change

## Technical Summary
Upgraded the Firebase Auth dependency to 23.2.0.
This was done to resolve a deprecated dependency to reCAPTCHA, loaded transitively via the old Firebase Auth.
Starting with this one upgrade to minimize changes while solving the deprecation issue.
Note we also discussed upgrading the Play Integrity dependency but that doesn't seem necessary so I left it out.

Checking the new dependency for recaptcha (Windows PowerShell syntax):
image

### Safety story
I verified I can still recover my PersonalID account after the change.
I also tested with a new demo user who was not pre-invited and got all the way to the backup code page with no issues.
I also searched the dependencies after build and verified that recaptcha will now use v18.6.1 instead of the old 18.1.2
All unit and integration tests still pass with no additional changes.

### Automated test coverage
No new testing added, functionality should be identical to callers.

### QA Plan
Verify that both Connect and non-Connect users can still register and recover their accounts after the change.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -34,7 +34,7 @@ configurations {
 
 dependencies {
     implementation 'androidx.exifinterface:exifinterface:1.3.6'
-    implementation 'com.google.android.play:integrity:1.4.0'
+    implementation 'com.google.android.play:integrity:1.6.0'
     testImplementation 'junit:junit:4.13.2'
     testImplementation('org.robolectric:robolectric:4.8.2') {
         exclude(group: 'org.bouncycastle', module: 'bcprov-jdk15on')
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -163,7 +163,7 @@ dependencies {
     implementation "androidx.camera:camera-camera2:$cameraX_version"
     implementation "androidx.camera:camera-lifecycle:$cameraX_version"
     implementation 'com.google.android.gms:play-services-mlkit-face-detection:17.1.0'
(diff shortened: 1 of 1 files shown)


OrangeAndGreen:
@shubham1g5 Looks like this is all we need, but should I go ahead and do the Play Integrity upgrade as well?

conroy-ricketts (dismissed):
LGTM assuming that we don't need to bump Play Integrity

shubham1g5:
@OrangeAndGreen Think we also need to update the `com.google.android.play:integrity:1.4.0` to latest as it depends on reCaptcha as well. 

> I verified I can still recover my PersonalID account after the change.

We should confirm this for a non pre-invited users with no opportunity access as well.

OrangeAndGreen:
@shubham1g5 I also tested with a new demo user who was not pre-invited and got all the way to the backup code page with no issues. Updated the safety story

OrangeAndGreen:
I bumped Play Integrity from 1.4.0 to 1.6.0, waiting for the build to pass now. Noting that this had no effect on the transitive recaptcha version after the previous upgrade, it's still 18.6.1.

conroy-ricketts (approved):

PR #3677, opened 2026-04-23
CCCT-2294 Parse Relearn Tasks From Delivery Progress Response

conroy-ricketts:
## Product Description

No user-visible change in isolation. This PR wires up data from Server to the new UI that was created in https://github.com/dimagi/commcare-android/pull/3639.
- `ConnectJobRecord.isRelearnTaskPending()` previously always returned `false`. It now reflects real server state, so once Server starts returning `assigned_tasks` and the downstream UI gets this data, delivery progress can be blocked for jobs with at least one assigned (incomplete) relearn task.
- `shouldShowRelearnTasksCompletedMessage()` will now return `true` for up to 6 hours after the last assigned relearn task flips to completed.

I recorded a demo video, but it's too big to upload here in GitHub, so please refer to the demo video I posted in Slack [here](https://dimagi.slack.com/archives/C09H7D1V1NK/p1777053385296949).

## Technical Summary

- Parses the new `assigned_tasks` array on the `/delivery_progress` response into a transient `List<ParsedConnectTask>` carried on `DeliveryAppProgressResponseModel`. Tasks are not persisted to the Connect DB.
- Per-task JSON deserialization lives on `ParsedConnectTask.fromJson(JSONObject)` (moved out of `DeliveryAppProgressResponseParser`) so the model owns its own parsing.
- Adds a new `ConnectJobPreferences` class that owns a per-job `SharedPreferences` file (`connect_job_prefs_<jobUUID>`), instantiated once per `ConnectJobRecord` via a lazy `getJobPreferences()` accessor. Keeps relearn-task state out of the global preferences file and out of `ConnectConstants`.
- Adds `ConnectJobRecord.syncRelearnTasksPrefs(tasks)` (instance method), invoked from `ConnectJobHelper.updateDeliveryProgress` on every response:
  - Writes a per-job boolean pref (`true` when any task is `assigned`, else `false`).
  - Resets the per-job relearn-tasks-completed timestamp whenever at least one task is still assigned.
  - When all tasks are completed and the timestamp pref is unset, sets it to the latest task `date_modified` (or the current time as a fallback).
- The sync helper runs even when the response omits `assigned_tasks`, so the per-job pending pref clears correctly if the server stops including the key.
- `isRelearnTaskPending()` now reads the per-job pref instead of returning the placeholder `false`.
- Removes the now-unused `RELEARN_TASKS_COMPLETED_TIME` constant from `ConnectConstants` (the key now lives privately in `ConnectJobPreferences`). Note: this pref was previously written as a single global value; it is now per-job, which is the intended scope.

### Safety story

- I tested this end-to-end with relearn tasks from Web to make sure the data is synced properly on Mobile.
- Storage is shared-prefs only — no schema change, no migration, no Connect DB writes. Each job gets its own SharedPreferences file (`connect_job_prefs_<jobUUID>`), so there is no cross-job pref pollution and no collision with global app prefs. The default boolean pref value is `false` (not pending), so existing installs read as "no pending tasks" until the server first returns `assigned_tasks`.
- `ParsedConnectTask` is intentionally transient (lives only between parse and sync) to keep the blast radius contained to prefs.
- `ConnectJobPreferences` resolves `Context` via `CommCareApplication.instance()`, matching the pattern used by `NotificationPrefs`. The cached instance on `ConnectJobRecord` is `transient` so it is rebuilt lazily after deserialization.

### Automated test coverage

No new tests in this PR in order to keep the scope small and minimize conflicts with in-flight Test Improver PRs.

### QA Plan

The QA notes in `RELEASES.md` was updated with https://github.com/dimagi/commcare-android/pull/3639, please refer to that file.

[CCCT-2294]: https://dimagi.atlassian.net/browse/CCCT-2294?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


conroy-ricketts:
Converting this to a draft for now until I'm able to dev test with an opportunity that has relearn tasks

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
```diff
@@ -859,29 +861,59 @@ public void setClaimed(boolean claimed) {
...
         return relearnTasksCompletedTimeMs != -1 && timeElapsedSinceTasksCompleted < DateUtils.HOUR_IN_MS * 6;
     }
+
+    public static void syncRelearnTasksPrefs(String jobUUID, List<ParsedConnectTask> tasks) {
+        ICommCarePreferenceManager preferenceManager = CommCarePreferenceManagerFactory.getCommCarePreferenceManager();
+
+        if (tasks == null || tasks.isEmpty()) {
+            preferenceManager.putLong(getPendingRelearnTasksKey(jobUUID), 0);
+            return;
+        }
```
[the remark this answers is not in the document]
conroy-ricketts: Yes, "no `assigned_tasks` in response" means "no relearn concept for this job at all". I think that it's unnecessary to add a code comment here for this, so we should probably leave this as-is.

app/src/org/commcare/connect/network/connect/parser/DeliveryAppProgressResponseParser.kt
```diff
@@ -84,12 +83,36 @@ class DeliveryAppProgressResponseParser<T> : BaseApiResponseParser<T> {
...
+    private fun parseTask(json: JSONObject): ParsedConnectTask {
+        val assigned = json.getString("status") == "assigned"
+        var dateModified: Date? = null
+
+        if (json.has("date_modified")) {
+            dateModified = DateUtils.parseDate(json.getString("date_modified"))
+        }
+
+        return ParsedConnectTask(assigned, dateModified)
     }
```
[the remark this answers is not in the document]
conroy-ricketts: I think that Server will never send `null` for `"date_modified"`. Also, yes, `status` is guaranteed present and non-null in every `assigned_tasks` entry.

conroy-ricketts:
> Converting this to a draft for now until I'm able to dev test with an opportunity that has relearn tasks

Fully tested now 👍🏾

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
```diff
@@ -859,29 +861,61 @@ public void setClaimed(boolean claimed) {
...
+                anyAssigned = true;
+            }
+
+            Date modified = task.getDateModified();
+            if (modified != null && (latestModified == null || modified.after(latestModified))) {
+                latestModified = modified;
+            }
+        }
+
+        preferenceManager.putLong(getPendingRelearnTasksKey(jobUUID), anyAssigned ? 1 : 0);
```
shubham1g5: think `boolean` might be a more apt data type here instead of `Long` ?

shubham1g5 (dismissed):

app/src/org/commcare/connect/network/connect/parser/DeliveryAppProgressResponseParser.kt
```diff
@@ -84,12 +83,36 @@ class DeliveryAppProgressResponseParser<T> : BaseApiResponseParser<T> {
...
                         job.payments = payments
                     }
+
+                    if (json.has("assigned_tasks")) {
+                        val array = json.getJSONArray("assigned_tasks")
+                        for (i in 0 until array.length()) {
+                            val obj = array[i] as JSONObject
+                            parsedTasks.add(parseTask(obj))
+                        }
+                    }
```
Jignesh-dimagi, line 96: @conroy-ricketts Can we put this function inside `ParsedConnectTask` like others i.e. `ConnectJobDeliveryRecord.fromJson` and `ConnectJobPaymentRecord.fromJson`

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
```diff
@@ -859,29 +861,61 @@ public void setClaimed(boolean claimed) {
...
+        if (anyAssigned) {
+            preferenceManager.putLong(RELEARN_TASKS_COMPLETED_TIME, -1);
+            return;
+        }
+
+        long currentTasksCompletedTime = preferenceManager.getLong(RELEARN_TASKS_COMPLETED_TIME, -1);
+        if (currentTasksCompletedTime == -1) {
+            // Set the completion time for all tasks to the latest date any task was modified, or
+            // fallback to the current date if there is no latest modified date.
+            long newTasksCompletedTime = latestModified != null ? latestModified.getTime() : new Date().getTime();
```
Jignesh-dimagi: @conroy-ricketts It should set only if no tasks are pending? 

Does this logic will work whenever there are multiple parsed tasks as RELEARN_TASKS_COMPLETED_TIME preference is shared among all?

conroy-ricketts: Yup that's correct! We only want to show the green completed message in the UI if all tasks were completed (i.e. no tasks are pending). I've successfully tested this with a demo user on Staging that's had 3 assigned relearn tasks.

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
conroy-ricketts: The reason for this originally was because `ICommCarePreferenceManager` only supported long values. But looking back at this, I see no reason why we shouldn't extend the interface!

New commit - 42e5aa1ea98c0fd87720106a799d28577313a2db

Cross-requested PR for core changes - https://github.com/dimagi/commcare-core/pull/1536

app/src/org/commcare/connect/network/connect/parser/DeliveryAppProgressResponseParser.kt
conroy-ricketts: Good idea! 7545136a0951ec43e903f2ba794324fa3dad680e

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
```diff
@@ -859,29 +861,61 @@ public void setClaimed(boolean claimed) {
...
-        //  or the latest date_modified from the Server response. Whenever there is at least one
-        //  pending task, always reset the RELEARN_TASKS_COMPLETED_TIME shared pref back to -1.
-
         ICommCarePreferenceManager preferenceManager = CommCarePreferenceManagerFactory.getCommCarePreferenceManager();
-        assert preferenceManager != null;
         long relearnTasksCompletedTimeMs = preferenceManager.getLong(RELEARN_TASKS_COMPLETED_TIME, -1);
         long timeElapsedSinceTasksCompleted = new Date().getTime() - relearnTasksCompletedTimeMs;
 
-        return relearnTasksCompletedTimeMs != -1 && timeElapsedSinceTasksCompleted < DateUtils.HOUR_IN_MS * 6;
+        return relearnTasksCompletedTimeMs != -1 && timeElapsedSinceTasksCompleted < DateUtils.HOUR_IN_MS * 6
```
OrangeAndGreen: Think it's worth extracting 6 to a defined constant like CompletedTasksMessageTimeoutHours

OrangeAndGreen (dismissed):

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
conroy-ricketts: Agreed! 73cee720e9c2d02238d715c3bebdded10af0b1cd

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
Jignesh-dimagi: Just curious over here: Were all 3 relearn tasks on different opportunities?

shubham1g5: @Jignesh-dimagi the delivery progress API is job specific so these tasks can't belong to different opportunities

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
```diff
@@ -859,29 +863,61 @@ public void setClaimed(boolean claimed) {
...
         long relearnTasksCompletedTimeMs = preferenceManager.getLong(RELEARN_TASKS_COMPLETED_TIME, -1);
         long timeElapsedSinceTasksCompleted = new Date().getTime() - relearnTasksCompletedTimeMs;
 
-        return relearnTasksCompletedTimeMs != -1 && timeElapsedSinceTasksCompleted < DateUtils.HOUR_IN_MS * 6;
+        return relearnTasksCompletedTimeMs != -1 && timeElapsedSinceTasksCompleted < RELEARN_TASKS_COMPLETED_MESSAGE_WINDOW_MS
+                && status == STATUS_DELIVERING;
+    }
+
+    public static void syncRelearnTasksPrefs(String jobUUID, List<ParsedConnectTask> tasks) {
+        ICommCarePreferenceManager preferenceManager = CommCarePreferenceManagerFactory.getCommCarePreferenceManager();
```
shubham1g5: @conroy-ricketts we don't really need to route through `CommCarePreferenceManagerFactory` here, that's a way for us to expose preferences to commcare-core classes, Instead you can directly use the `CommCareApplication.instance().getSharedPreferences` for all Connect and Personal ID code. It would be nice to add a class `ConnectJobPreferences` as well which is Connect opportunity specific so that we don't need to key in job id in these preferences as well. Think it would be quite a small change that this PR can benefit from.

shubham1g5 (dismissed):
Left one suggestion which would be quite good to have on this PR, but keeping it non-blocking.

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
Jignesh-dimagi: @shubham1g5 Yeah, this is where I am confused. SharedPreference 'RELEARN_TASKS_COMPLETED_TIME' is common between all opportunities. So I am not sure. How does it react when it has to relearn tasks across different opportunities?

shubham1g5: the shared preference takes in the opportunity id and use it to determine the key for the preference, so we are maintaining a different preference per opp

shubham1g5: @conroy-ricketts Jignesh flagged that `RELEARN_TASKS_COMPLETED_TIME`  is not per opportunity and is not keying the opportunity id, so that does seem to be a real issue here

conroy-ricketts: @Jignesh-dimagi @shubham1g5 

Aaaah, yes this is a known issue and it affects multiple keys in our SharedPrefs, we have a ticket to resolve these issues here: [CCCT-2341](https://dimagi.atlassian.net/browse/CCCT-2341)

Do you guys think we should prioritize CCCT-2341 for the current Sprint?

conroy-ricketts: I'm now realizing that `RELEARN_TASK_PENDING_PREFIX` should be refactored and migrated out of the global prefs as well

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
conroy-ricketts: Ah yes, responded to similar comment above [here](https://github.com/dimagi/commcare-android/pull/3677#discussion_r3154288632)

conroy-ricketts:
@coderabbitai full review

conroy-ricketts:
@coderabbitai full review

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
conroy-ricketts: Offlined with @shubham1g5. We will prioritize [CCCT-2341](https://dimagi.atlassian.net/browse/CCCT-2341), but I also created a new job-specific shared prefs for this PR

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
conroy-ricketts: PR updated to reflect this now

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
```diff
@@ -188,6 +184,8 @@ public class ConnectJobRecord extends Persisted implements Serializable {
 
     private boolean claimed;
 
+    private transient ConnectJobPreferences jobPreferences;
```
shubham1g5, line 187: curious if there is a specific reason to define this as `transient` ?

shubham1g5 (approved):

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
conroy-ricketts: Claude added this so that the prefs object would be ignored during Java's serialization process. It seemed like a good idea (which I hadn't thought of at the time) but I'm open to thoughts here! I realize that it may be unnecessary given how we currently use `ConnectJobRecord`

conroy-ricketts: Update: I ended up removing this code while addressing the feedback [here](https://github.com/dimagi/commcare-android/pull/3682#discussion_r3156687056)

shubham1g5 (dismissed):

Jignesh-dimagi (dismissed):

conroy-ricketts:
@Jignesh-dimagi @shubham1g5 My latest commit was to optimize imports and format code

shubham1g5 (approved):

PR #3678, opened 2026-04-27
Fix NullPointerException in ComboboxWidget item-click handler

avazirna:
## Technical Summary
Crashlytics is reporting a `NullPointerException` from `ComboboxWidget.lambda$addListeners$0`:
```
java.lang.NullPointerException: Attempt to invoke virtual method
'java.lang.Object android.widget.AdapterView.getItemAtPosition(int)'
on a null object reference
  at ComboboxWidget.lambda$addListeners$0(ComboboxWidget.java:84)
  at AutoCompleteTextView.performCompletion(AutoCompleteTextView.java:1086)
  at AutoCompleteTextView$DropDownItemClickListener.onItemClick(...)
```
It seems that a race condition causes the dropdown list to be dismissed between when the user selects an item and when the onClick event is processed. This change retrieves the selected from the adapter instead.

Crashlytics report [Link](https://console.firebase.google.com/u/0/project/commcare-a57e4/crashlytics/app/android:org.commcare.dalvik/issues/2fdea36c5289476e677c8f301f5a1d53?time=30d&versions=2.62.0%20(486560)&types=crash,ANR&sessionEventKey=69EEFEBE0242000172A2FF27CB470392_2211768961625711250)

### Safety story
- New code preserves existing semantics: `parent.getItemAtPosition(position)` delegates to `adapter.getItem(position)` internally, so a successful click still produces the same `ComboItem`.
- The previous code crashed in the null-`parent` scenario; the new code returns the correct item even if the dropdown list has been dismissed.

## Labels and Review

- [ ] Do we need to enhance the manual QA test coverage? — No
- [ ] Does the PR introduce any major changes worth communicating? — No
- [ ] The set of people pinged as reviewers is appropriate for the level of risk

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -17,6 +17,7 @@ These are published publically on Playstore, Github Releases and CommCare Forums
 - [6-box Backup Codes] Using 6-box numeric inputs to collect backup codes from the user
 
 #### Important Bug Fixes
+- Fixed a crash triggered during combobox item selection when the dropdown list had already been dismissed
 
 #### Internal Release Notes
 
@@ -43,6 +44,7 @@ we would like to communicate to QA as part of the release testing
   - Text cursor functionality (i.e. backspacing, clicking an earlier box to jump back)
 
 - Test the new offline status indicator at the top of refreshable Connect pages (Connect Home, Learning Progress, Delivery Progress). Verify that the error message appears when entering these pages while offline, and that it disappears once the device comes back online.
+- Verify that the combobox widget is working as expected when selecting an item that is used to filter another combobox widget and also determines the visibility of some other unrelated question whose relevance condition depends on the selection.
 
 ## CommCare 2.62
 
--- a/app/src/org/commcare/views/widgets/ComboboxWidget.java
+++ b/app/src/org/commcare/views/widgets/ComboboxWidget.java
@@ -81,9 +81,9 @@ public class ComboboxWidget extends QuestionWidget {
 
     private void addListeners() {
         comboBox.setOnItemClickListener((parent, view, position, id) -> {
-            selectedComboItem = (ComboItem) parent.getItemAtPosition(position);
-            widgetEntryChanged();}
-        );
+            selectedComboItem = (ComboItem) comboBox.getAdapter().getItem(position);
+            widgetEntryChanged();
+        });
 
         // Note that Combobox has an OnFocusChangeListener defined in its own class, so when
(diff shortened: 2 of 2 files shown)


shubham1g5:
> It seems that a race condition causes the dropdown list to be dismissed between when the user selects an item and when the onClick event is processed. This change retrieves the selected from the adapter instead.

Think more context around this would be helpful here explicitly what are the 2 code pathways that causes a race condition and have you been able to reproduce this by introducing a wait in one of those pathways.

avazirna:
> > It seems that a race condition causes the dropdown list to be dismissed between when the user selects an item and when the onClick event is processed. This change retrieves the selected from the adapter instead.
> 
> Think more context around this would be helpful here explicitly what are the 2 code pathways that causes a race condition and have you been able to reproduce this by introducing a wait in one of those pathways.

I couldn't replicate as such. I tried a few things: item selection during configuration changes, different combinations of display conditions and filters, but no success. That's why I decided to retrieve the item from the adapter instead, which is what the `DropDownListView.getItemAtPosition()` internally does. The tricky part is to understand which `ComboboxWidget` is actually triggering this. I will add a debugging log.

conroy-ricketts (dismissed):
+1 on adding a debugging log

avazirna:
> > > It seems that a race condition causes the dropdown list to be dismissed between when the user selects an item and when the onClick event is processed. This change retrieves the selected from the adapter instead.
> > 
> > 
> > Think more context around this would be helpful here explicitly what are the 2 code pathways that causes a race condition and have you been able to reproduce this by introducing a wait in one of those pathways.
> 
> I couldn't replicate as such. I tried a few things: item selection during configuration changes, different combinations of display conditions and filters, but no success. That's why I decided to retrieve the item from the adapter instead, which is what the `DropDownListView.getItemAtPosition()` internally does. The tricky part is to understand which `ComboboxWidget` is actually triggering this. I will add a debugging log.

Done 91d1b58a68b9e998c26907eb17ee6b5344c683f7

shubham1g5 (dismissed):
Should we add a QA note for this to ensure we test impacted areas as part of the release.

conroy-ricketts (dismissed):

OrangeAndGreen (dismissed):

avazirna:
> Should we add a QA note for this to ensure we test impacted areas as part of the release.

I've updated RELEASES.md. Regression testing around the widget should suffice.

shubham1g5 (approved):

PR #3681, opened 2026-04-28
CCCT-2334 - Hide the pending filter in the payment unit detail screen if there are no pending visits for that payment unit

Jignesh-dimagi:
## Product Description
It has visible changes; it will hide the pending filter in the payment unit detail screen if there are no pending visits for that payment unit.

### [CCCT-2334](https://dimagi.atlassian.net/browse/CCCT-2334) ###

The application is checking for the pending visits in the payment unit detail screen. If there are no pending visits for that payment unit, it will hide the pending filter else will keep it visible.

The check is done at two places, while loading initial data and after data refresh.

Screenshot_20260428_100544

Note: I have applied a temporarily patch in the code to get a screenshot to show that buttons are aligned uniformly after removing the pending button.

## Labels and Review

[CCCT-2334]: https://dimagi.atlassian.net/browse/CCCT-2334?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/org/commcare/fragments/connect/ConnectDeliveryListFragment.java
+++ b/app/src/org/commcare/fragments/connect/ConnectDeliveryListFragment.java
@@ -85,7 +85,26 @@ public class ConnectDeliveryListFragment extends ConnectJobFragment<FragmentConn
             filterCards[i].setOnClickListener(v -> onFilterSelected(filter, filterCards, filterLabels));
         }
 
+        updatePendingFilterVisibility();
         setFilterHighlight(filterCards[0], filterLabels[0], true);
+    }
+
+    private void updatePendingFilterVisibility() {
+        boolean hasPending = hasPendingDeliveries();
+        getBinding().pendingFilterButton.setVisibility(hasPending ? View.VISIBLE : View.GONE);
+        if (!hasPending && PENDING_IDENTIFIER.equals(currentFilter)) {
+            getBinding().allFilterButton.performClick();
+        }
+    }
+
+    private boolean hasPendingDeliveries() {
+        for (ConnectJobDeliveryRecord delivery : job.getDeliveries()) {
+            if (delivery.getUnitName().equalsIgnoreCase(unitName)
+                    && PENDING_IDENTIFIER.equalsIgnoreCase(delivery.getStatus())) {
+                return true;
+            }
+        }
+        return false;
     }
 
     private void onFilterSelected(String selectedFilter, CardView[] filterCards, TextView[] filterLabels) {
@@ -126,6 +145,7 @@ public class ConnectDeliveryListFragment extends ConnectJobFragment<FragmentConn
     private void refreshData() {
         ConnectJobHelper.INSTANCE.updateDeliveryProgress(getContext(), job, true, this, (success,error) -> {
             if (success) {
+                updatePendingFilterVisibility();
                 adapter.updateDeliveries(getFilteredDeliveries());
             }
         });
--- a/app/src/org/commcare/fragments/connect/ConnectDeliveryListFragment.java
+++ b/app/src/org/commcare/fragments/connect/ConnectDeliveryListFragment.java
@@ -99,8 +99,7 @@ public class ConnectDeliveryListFragment extends ConnectJobFragment<FragmentConn
 
     private boolean hasPendingDeliveries() {
         for (ConnectJobDeliveryRecord delivery : job.getDeliveries()) {
-            if (delivery.getUnitName().equalsIgnoreCase(unitName)
-                    && PENDING_IDENTIFIER.equalsIgnoreCase(delivery.getStatus())) {
+            if (matchesFilter(delivery, PENDING_IDENTIFIER)) {
                 return true;
             }
         }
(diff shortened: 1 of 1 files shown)


Jignesh-dimagi:
> ```diff
>  private void refreshData() {
>          ConnectJobHelper.INSTANCE.updateDeliveryProgress(getContext(), job, true, this, (success,error) -> {
>              if (success) {
> -                updatePendingFilterVisibility();
> -                adapter.updateDeliveries(getFilteredDeliveries());
> +                boolean filterChangedViaFallback = updatePendingFilterVisibility();
> +                if (!filterChangedViaFallback) {
> +                    adapter.updateDeliveries(getFilteredDeliveries());
> +                }
>              }
>          });
>      }
> ```

@coderabbit
This seems wrong, as the application needs to update the deliveries on the UI irrespective of whether it has pending deliveries or not in the refresh. As per your logic, it will refresh UI only in some conditions, which is false.

app/src/org/commcare/fragments/connect/ConnectDeliveryListFragment.java
```diff
@@ -85,9 +85,28 @@ private void setupFilterControls() {
...
 
+    private void updatePendingFilterVisibility() {
+        boolean hasPending = hasPendingDeliveries();
+        getBinding().pendingFilterButton.setVisibility(hasPending ? View.VISIBLE : View.GONE);
+        if (!hasPending && PENDING_IDENTIFIER.equals(currentFilter)) {
+            getBinding().allFilterButton.performClick();
+        }
+    }
+
+    private boolean hasPendingDeliveries() {
```
shubham1g5, line 100: we can probably reuse `getFilteredDeliveries` method here by passing the filter as a method parameter instead

app/src/org/commcare/fragments/connect/ConnectDeliveryListFragment.java
```diff
@@ -85,9 +85,28 @@ private void setupFilterControls() {
...
+        updatePendingFilterVisibility();
         setFilterHighlight(filterCards[0], filterLabels[0], true);
     }
 
+    private void updatePendingFilterVisibility() {
+        boolean hasPending = hasPendingDeliveries();
+        getBinding().pendingFilterButton.setVisibility(hasPending ? View.VISIBLE : View.GONE);
+        if (!hasPending && PENDING_IDENTIFIER.equals(currentFilter)) {
+            getBinding().allFilterButton.performClick();
+        }
```
shubham1g5, line 97: how can `PENDING_IDENTIFIER.equals(currentFilter)` be `true` at this state given we are setting up the fragment and initial filter is `all` ?

Jignesh-dimagi: Yeah, it cannot be at the entry / initial but can be while refreshing the data, so care is taken for the same.

shubham1g5: is not this only called on fragment creation at the moment ?

app/src/org/commcare/fragments/connect/ConnectDeliveryListFragment.java
Jignesh-dimagi: `getFilteredDeliveries` already has some logic around the `currentFilter`, which I was not thinking to combine with this, but with your comment, I have abstracted the core functionality and routed both through it. Also, `hasPendingDeliveries()` short-circuits on the first match only.
Changes done [here](https://github.com/dimagi/commcare-android/pull/3681/changes/be6e0139344629b6796e31e0b9a072b09af96672)

app/src/org/commcare/fragments/connect/ConnectDeliveryListFragment.java
Jignesh-dimagi: User can [refresh data](https://github.com/dimagi/commcare-android/blob/ae80bce89f55c7e7e2e8df64e740b95a2dea1129/app/src/org/commcare/fragments/connect/ConnectDeliveryListFragment.java#L118) from menu item.

shubham1g5: right but that's not ending up calling `updatePendingFilterVisibility` which seems to be an issue with this implementation, think we should keep this method as it is but also call `updatePendingFilterVisibility` after the data has been refreshed.

Jignesh-dimagi: It is getting called [here](https://github.com/dimagi/commcare-android/blob/be6e0139344629b6796e31e0b9a072b09af96672/app/src/org/commcare/fragments/connect/ConnectDeliveryListFragment.java#L147), after data is refreshed.

shubham1g5 (approved):

conroy-ricketts (approved):

OrangeAndGreen (approved):

PR #3682, opened 2026-04-28
CCCT-2341 Migrate Job-Specific Shared Prefs

conroy-ricketts:
### [CCCT-2341](https://dimagi.atlassian.net/browse/CCCT-2341)

> **Note for reviewers:** This branch was forked from the branch in https://github.com/dimagi/commcare-android/pull/3677 (CCCT-2294) so that it could reuse the new `ConnectJobPreferences` class introduced there. As a result, every commit from PR #3677 is included here. To keep review focused, **please review only the last four commits on this branch**:
> - `92a43537a` — CCCT-2341 Migrate Job-Specific Shared Prefs
> - `7565f6048` — CCCT-2341 Optimized imports
> - `8e2f9fe9b` — CCCT-2341 Decoupled job prefs from `ConnectJobRecord`
> - `908a3a674` — CCCT-2341 Consolidated "timestamp not set" constants
>
> The earlier commits will land in master via PR #3677 and should not be re-reviewed here.

## Technical Summary

Moves the `payment_confirmation_hidden_since_time` value out of the global `CommCarePreferenceManager` and into the per-job `ConnectJobPreferences` file (`connect_job_prefs_<jobUUID>`).

The previous implementation stored a single global timestamp for "user has hidden the payment confirmation tile". With multiple Connect jobs/opportunities, hiding the tile on one job would silently suppress it on all others. The value is conceptually job-specific, so it now lives in the per-job prefs file alongside the relearn-task prefs introduced in PR #3677.

Changes:
- `ConnectJobPreferences.kt` — adds `getPaymentConfirmationHiddenSinceTime` / `setPaymentConfirmationHiddenSinceTime` / `resetPaymentConfirmationHiddenSinceTime` / `paymentConfirmationHiddenSinceTimeNotSet`. The two per-key "not set" sentinels were collapsed into a single `TIMESTAMP_NOT_SET = -1L`.
- `ConnectJobUtils.getJobPreferences(jobUUID)` — new static accessor so any caller with a `jobUUID` can reach a job's prefs without going through a `ConnectJobRecord` instance. The cached `ConnectJobPreferences` field and `getJobPreferences()` getter on `ConnectJobRecord` were removed; the model's remaining pref-reading methods (`isRelearnTaskPending`, `shouldShowRelearnTasksCompletedMessage`, `syncRelearnTasksPrefs`) call the static helper.
- `ConnectJobUtils.storePayments` — drops the `ConnectJobPreferences` parameter and looks prefs up via the new static accessor when resetting the hidden-since timestamp on new payments.
- `ConnectDeliveryProgressFragment` — read/write paths now go through `ConnectJobUtils.getJobPreferences(job.getJobUUID())`.
- `ConnectConstants.PAYMENT_CONFIRMATION_HIDDEN_SINCE_TIME` constant removed.
- Stale `CommCarePreferenceManagerFactory` / `ICommCarePreferenceManager` imports removed from the two touched files.

### Codebase audit — no other migration candidates

I searched the rest of the codebase for additional job-specific values still living in global shared prefs and found none. Reasoning:

- No `preferenceManager.put*/get*` calls remain anywhere in `org.commcare.connect.*`, `org.commcare.fragments.connect.*`, or `org.commcare.android.database.connect.*`.
- No `HiddenPreferences.put*/get*/set*` calls in connect-related code (the lone reference, `HiddenPreferences.getUserDomain()` in `ApiPersonalId.java`, is user-level and correctly app-global).
- No ad-hoc `getSharedPreferences(...)` calls anywhere in the connect packages.
- `ConnectConstants.java` — every remaining constant is an intent extra, request code, navigation destination, auth token, or notification field. None are persisted shared-prefs keys.
- `NotificationPrefs.kt` is the only other custom prefs file; it stores a single global FCM-notification-read boolean, which is not per-job.
- The only persistent prefs keys with job-flavored names (`relearn_task_pending`, `relearn_tasks_completed_time_ms`, `payment_confirmation_hidden_since_time`) all live in `ConnectJobPreferences`.

### Safety story

- I tested the app locally and confirmed the payment-confirmation tile is still hidden correctly when the user dismisses it in the UI.
- Behavior change for existing users: any previously-stored global `payment_confirmation_hidden_since_time` value is effectively reset on first launch after this change (the new per-job key starts unset), so users will see the tile re-appear once if they had previously dismissed it. If they hide it again, that hide is now correctly scoped to the active job.
- No schema/migration changes; the affected storage is a SharedPreferences file.

### Automated test coverage

No new unit tests added — the change is a straightforward storage-location swap.

### QA Plan

QA should regression-test the payment confirmation tile flow to confirm there are no regressions, in particular:
- The tile appears when there are unconfirmed payments and disappears after the user dismisses ("Ask Later"), then reappears after the documented hidden window.
- With multiple opportunities, dismissing the tile on one opportunity should not suppress it on others.
- A brand-new payment received from the server should clear the hidden state and re-show the tile.

[CCCT-2341]: https://dimagi.atlassian.net/browse/CCCT-2341?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/org/commcare/connect/ConnectJobHelper.kt
+++ b/app/src/org/commcare/connect/ConnectJobHelper.kt
@@ -116,7 +116,7 @@ object ConnectJobHelper {
                     if (job.payments.isNotEmpty()) {
                         events.add(PAID_DELIVERY)
                     }
-                    ConnectJobUtils.storePayments(context, job.payments, job.jobUUID, true, job.jobPreferences)
+                    ConnectJobUtils.storePayments(context, job.payments, job.jobUUID, true)
                 }
 
                 job.syncRelearnTasksPrefs(deliveryAppProgressResponseModel.parsedTasks)
--- a/app/src/org/commcare/connect/ConnectJobHelper.kt
+++ b/app/src/org/commcare/connect/ConnectJobHelper.kt
@@ -116,7 +116,7 @@ object ConnectJobHelper {
                     if (job.payments.isNotEmpty()) {
                         events.add(PAID_DELIVERY)
                     }
-                    ConnectJobUtils.storePayments(context, job.payments, job.jobUUID, true)
+                    ConnectJobUtils.storePayments(context, job.payments, job.jobUUID, true, job.jobPreferences)
                 }
 
                 job.syncRelearnTasksPrefs(deliveryAppProgressResponseModel.parsedTasks)
--- a/app/src/org/commcare/preferences/ConnectJobPreferences.kt
+++ b/app/src/org/commcare/preferences/ConnectJobPreferences.kt
@@ -26,10 +26,24 @@ class ConnectJobPreferences(jobUUID: String) {
         setRelearnTasksCompletedTime(RELEARN_TASKS_COMPLETED_TIME_NOT_SET)
     }
 
+    fun getPaymentConfirmationHiddenSinceTime(): Long = prefs.getLong(PAYMENT_CONFIRMATION_HIDDEN_SINCE_TIME, PAYMENT_CONFIRMATION_HIDDEN_SINCE_TIME_NOT_SET)
+
+    fun paymentConfirmationHiddenSinceTimeNotSet(): Boolean = getPaymentConfirmationHiddenSinceTime() == PAYMENT_CONFIRMATION_HIDDEN_SINCE_TIME_NOT_SET
+
+    fun setPaymentConfirmationHiddenSinceTime(hiddenSinceTimeMs: Long) {
+        prefs.edit { putLong(PAYMENT_CONFIRMATION_HIDDEN_SINCE_TIME, hiddenSinceTimeMs) }
+    }
+
+    fun resetPaymentConfirmationHiddenSinceTime() {
+        setPaymentConfirmationHiddenSinceTime(PAYMENT_CONFIRMATION_HIDDEN_SINCE_TIME_NOT_SET)
+    }
+
     companion object {
         private const val PREF_NAME_PREFIX = "connect_job_prefs_"
         private const val KEY_RELEARN_TASK_PENDING = "relearn_task_pending"
         private const val KEY_RELEARN_TASKS_COMPLETED_TIME_MS = "relearn_tasks_completed_time_ms"
         private const val RELEARN_TASKS_COMPLETED_TIME_NOT_SET = -1L
+        private const val PAYMENT_CONFIRMATION_HIDDEN_SINCE_TIME = "payment_confirmation_hidden_since_time"
+        private const val PAYMENT_CONFIRMATION_HIDDEN_SINCE_TIME_NOT_SET = -1L
     }
 }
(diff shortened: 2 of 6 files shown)


shubham1g5:
@conroy-ricketts I just changed the base branch to `CCCT-2294-parse-relearn-tasks-for-UI` here so that only the additional commits appear in the PR.

app/src/org/commcare/connect/ConnectJobHelper.kt
```diff
@@ -116,7 +116,7 @@ object ConnectJobHelper {
                     if (job.payments.isNotEmpty()) {
                         events.add(PAID_DELIVERY)
                     }
-                    ConnectJobUtils.storePayments(context, job.payments, job.jobUUID, true)
+                    ConnectJobUtils.storePayments(context, job.payments, job.jobUUID, true, job.jobPreferences)
```
shubham1g5: think we should make these preferences accessible from everywhere by making a `static getJobPreferences(jobId)` method possibly in `ConnectJobUtils` itself. That will get rid of the coupling between the model and Preference class as well.

app/src/org/commcare/preferences/ConnectJobPreferences.kt
```diff
@@ -26,10 +26,24 @@ class ConnectJobPreferences(jobUUID: String) {
...
+        setPaymentConfirmationHiddenSinceTime(PAYMENT_CONFIRMATION_HIDDEN_SINCE_TIME_NOT_SET)
+    }
+
     companion object {
         private const val PREF_NAME_PREFIX = "connect_job_prefs_"
         private const val KEY_RELEARN_TASK_PENDING = "relearn_task_pending"
         private const val KEY_RELEARN_TASKS_COMPLETED_TIME_MS = "relearn_tasks_completed_time_ms"
         private const val RELEARN_TASKS_COMPLETED_TIME_NOT_SET = -1L
+        private const val PAYMENT_CONFIRMATION_HIDDEN_SINCE_TIME = "payment_confirmation_hidden_since_time"
+        private const val PAYMENT_CONFIRMATION_HIDDEN_SINCE_TIME_NOT_SET = -1L
```
shubham1g5: think we can use a single constant for all timestamps not set (it'd always be `-1`)

app/src/org/commcare/connect/ConnectJobHelper.kt
conroy-ricketts: 8e2f9fe9bf1fc70a2b515b440446ff2d22614803

I also cleaned up the creation of the prefs object given that Android already caches shared pref instances internally

app/src/org/commcare/preferences/ConnectJobPreferences.kt
conroy-ricketts: Good catch!

908a3a6743465da317e3daa9f4f23fee92eeff7c

shubham1g5 (dismissed):

Jignesh-dimagi (dismissed):

conroy-ricketts:
@Jignesh-dimagi @shubham1g5 the latest commit here was to optimize imports and format code

shubham1g5 (approved):

PR #3684, opened 2026-04-30
CCCT-2217 Session Endpoint Navigation From Connect Notifications

shubham1g5:
## Product Description

When a Connect push notification with a `session_endpoint_id` is clicked, the user is now navigated directly to the relevant session endpoint in their CommCare app (optionally after a blocking sync), rather than landing on the Connect activity. A special `cc_app_home` endpoint is also supported, which simply leaves the user on the app home screen.

Video showing redirect based on user state - 

https://github.com/user-attachments/assets/faaf758a-52b9-45c1-adb4-279379dd6f45

Video showing app switching from notification redirect- 

https://github.com/user-attachments/assets/73be0468-ab93-4bca-a639-2a7fbc899288

Video showing notification redirect directly into a form - 

https://github.com/user-attachments/assets/0e311888-c298-4b55-836a-2e3b95d0005f

## Technical Summary

This PR wires up session endpoint navigation from Connect push notifications end-to-end:

- **`PushNotificationRecord` (DB v24 → v25):** Added `sessionEndpointId` and `requireAppSync` fields. `requireAppSync` defaults to `true` so older server payloads always trigger a sync. Includes a versioned snapshot class (`PushNotificationRecordV24`) and an additive migration that adds the new columns.
- **`ConnectJobUtils`:** New `getAppIdForOpportunity()` helper that resolves the CommCare app ID from an opportunity UUID and status (learn vs. delivery).
- **`FirebaseMessagingUtil`:** `ccc_generic_opportunity` notifications that carry a `session_endpoint_id` are now routed to `DispatchActivity` (with the endpoint ID, app ID, and `requireSync` flag) instead of `ConnectActivity`. Falls back to `ConnectActivity` routing if the opportunity or app ID cannot be resolved.
- **`HomeScreenBaseActivity`:** Handles the `CC_LAUNCH_REQUIRE_SYNC` flag — when set, triggers a blocking sync before navigating to the session endpoint. Adds support for the `cc_app_home` endpoint, which no-ops navigation and leaves the user on the home screen.
- **`DispatchActivity`:** Propagates `CC_LAUNCH_REQUIRE_SYNC` to `HomeScreenBaseActivity`. Fixes a lifecycle race condition: when a session endpoint notification is clicked while the login screen is showing, `FLAG_ACTIVITY_CLEAR_TOP` causes `LoginActivity` to send `RESULT_CANCELED` to `DispatchActivity`, which was calling `finish()` in `onResume` before `onNewIntent` could deliver the new notification intent. The fix posts `finish()` via `Handler.post()` so `onNewIntent` (delivered synchronously in the same Looper transaction) can cancel it and dispatch instead.

### Safety story

- DB migration is purely additive — two new columns with safe defaults (`""` and `true`). Existing records are migrated with those defaults applied.
- Notification routing falls back to the existing `ConnectActivity` path if the opportunity ID is missing or the app ID cannot be resolved, so no regression for malformed payloads.
- The `DispatchActivity` lifecycle fix is guarded by a re-check of `shouldFinish` inside the posted lambda, so the normal "user pressed back from login → close app" flow is unaffected.
- Local Testing:
-  notification click from login screen, notification click while app is backgrounded, and normal login-cancel flow. 
- Test the behaviour from different screens on top of activity stack when user clicks on notification
- Update Testing for Notifications Migration
- Tests session endpoints other than App home

### Automated test coverage

`SessionEndpointNotificationTest` covers:
- `ccc_generic_opportunity` with `session_endpoint_id` routes to `DispatchActivity` with correct extras

[CCCT-2217]: https://dimagi.atlassian.net/browse/CCCT-2217?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CCCT-2329]: https://dimagi.atlassian.net/browse/CCCT-2329?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -28,7 +28,7 @@ These notes are only published internally in [CommCare Change log wiki](https://
 along with the public release notes above
 -->
 
-- - Session endpoint navigation from Connect notifications: clicking a notification with a `session_endpoint_id` now navigates the user directly to the specified CommCare session endpoint (after a sync if required), instead of opening the Connect activity.
+- Session endpoint navigation from Connect notifications: clicking a notification with a `session_endpoint_id` now navigates the user directly to the specified CommCare session endpoint (after a sync if required), instead of opening the Connect activity.
 
 
 ### QA Notes
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -42,7 +42,7 @@ we would like to communicate to QA as part of the release testing
     - On clicking, Notification should take user to the relevant CommCare App Home page and auto-login and auto-syncs the user with a blocking dialog. 
     - Click the notification while logged out
     - Click the notification while the app is backgrounded
-    - Cancel login normally (back button, no notification) → app should still close as expected.
+    - Verify that notifications redirect work as expected from various places in the app - Opp Screen, App Home, Login Screen, Form Entry etc and back navigation works correctly after the notification redirect
     - Verify no regression on existing Connect notification types (payments, messaging, delivery/learn progress).
 
 - Verify that the existing opportunity card UI is unchanged when there are no relearn tasks.
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -14,6 +14,8 @@ These are published publically on Playstore, Github Releases and CommCare Forums
 - Offline status shown on refreshable Connect pages when applicable
 
 - [Relearn Tasking] Added relearn task notification UI to Connect opportunity cards
+- [Relearn Tasking] Implements a new notification when a relearn task is assigned to a user
+- [Work Area Assignment] Implements a new notification when a new work area is assigned to a user
 - [6-box Backup Codes] Using 6-box numeric inputs to collect backup codes from the user
 
 #### Important Bug Fixes
@@ -26,6 +28,8 @@ These notes are only published internally in [CommCare Change log wiki](https://
 along with the public release notes above
 -->
 
+- - Session endpoint navigation from Connect notifications: clicking a notification with a `session_endpoint_id` now navigates the user directly to the specified CommCare session endpoint (after a sync if required), instead of opening the Connect activity.
+
 
 ### QA Notes
 
@@ -33,6 +37,13 @@ along with the public release notes above
 These are for internal use and for us to keep track of important notes that
 we would like to communicate to QA as part of the release testing
 -->
+
+- **Task and Work Area Assignment Notifications (Connect):**
+    - On clicking, Notification should take user to the relevant CommCare App Home page and auto-login and auto-syncs the user with a blocking dialog. 
+    - Click the notification while logged out
+    - Click the notification while the app is backgrounded
+    - Cancel login normally (back button, no notification) → app should still close as expected.
+    - Verify no regression on existing Connect notification types (payments, messaging, delivery/learn progress).
 
 - Verify that the existing opportunity card UI is unchanged when there are no relearn tasks.
 - Verify that the opportunity card updates as expected when there are either pending relearn tasks or completed relearn tasks.
--- a/app/src/org/commcare/activities/DispatchActivity.java
+++ b/app/src/org/commcare/activities/DispatchActivity.java
@@ -171,7 +171,6 @@ public class DispatchActivity extends AppCompatActivity {
 
         Intent pnIntent = checkIfAnyPNIntentPresent();
         if (pnIntent != null) {
-            String actionType = pnIntent.getStringExtra(REDIRECT_ACTION);
             FirebaseAnalyticsUtil.reportNotificationEvent(
                     AnalyticsParamValue.NOTIFICATION_EVENT_TYPE_CLICK,
                     AnalyticsParamValue.REPORT_NOTIFICATION_CLICK_NOTIFICATION_TRAY,
--- a/app/src/org/commcare/activities/DispatchActivity.java
+++ b/app/src/org/commcare/activities/DispatchActivity.java
@@ -55,6 +55,7 @@ public class DispatchActivity extends AppCompatActivity {
     // Args to session endpoints can be passed as a name to value bundle or more loosely as a list
     public static final String SESSION_ENDPOINT_ARGUMENTS_BUNDLE = "ccodk_session_endpoint_arguments_bundle";
     public static final String SESSION_ENDPOINT_ARGUMENTS_LIST = "ccodk_session_endpoint_arguments_list";
+    public static final String CC_LAUNCH_REQUIRE_SYNC = "ccodk_require_sync";
     public static final String WAS_EXTERNAL = "launch_from_external";
     public static final String EXIT_AFTER_FORM_SUBMISSION = "ccodk_exit_after_form_submission";
     public static final Boolean EXIT_AFTER_FORM_SUBMISSION_DEFAULT = true;
@@ -441,6 +442,8 @@ public class DispatchActivity extends AppCompatActivity {
                         i.putExtra(SESSION_ENDPOINT_ID, sessionEndpointId);
                         i.putExtra(SESSION_ENDPOINT_ARGUMENTS_BUNDLE, args);
                         i.putStringArrayListExtra(SESSION_ENDPOINT_ARGUMENTS_LIST, argsList);
+                        i.putExtra(CC_LAUNCH_REQUIRE_SYNC,
+                                getIntent().getBooleanExtra(CC_LAUNCH_REQUIRE_SYNC, false));
 
                         // Session Endpoint extra is no longer needed. If not removed, it triggers
                         // the external launch logic in subsequent logins
--- a/app/src/org/commcare/activities/DispatchActivity.java
+++ b/app/src/org/commcare/activities/DispatchActivity.java
@@ -12,8 +12,6 @@ import static org.commcare.utils.FirebaseMessagingUtil.getNotificationActionFrom
 
 import android.content.Intent;
 import android.os.Bundle;
-import android.os.Handler;
-import android.os.Looper;
 
 import androidx.annotation.NonNull;
 import android.util.Log;
@@ -141,29 +139,11 @@ public class DispatchActivity extends AppCompatActivity {
         return false;
     }
 
-
-    @Override
-    protected void onNewIntent(@NonNull Intent intent) {
-        super.onNewIntent(intent);
-        setIntent(intent);
-        if (shouldFinish) {
-            // onResume already ran and posted a finish();
-            // but a new intent has arrived so cancel that decision and dispatch instead.
-            shouldFinish = false;
-            dispatch();
-        }
-    }
-
     @Override
     protected void onResume() {
         super.onResume();
         if (shouldFinish) {
-            // Post finish() so that onNewIntent() — which Android may deliver synchronously
-            // after onResume() in the same Looper transaction — can cancel it by resetting
-            // shouldFinish before the posted runnable executes.
-            new Handler(Looper.getMainLooper()).post(() -> {
-                if (shouldFinish) finish();
-            });
+            finish();
         } else {
             dispatch();
         }
--- a/app/src/org/commcare/activities/DispatchActivity.java
+++ b/app/src/org/commcare/activities/DispatchActivity.java
@@ -12,6 +12,10 @@ import static org.commcare.utils.FirebaseMessagingUtil.getNotificationActionFrom
 
 import android.content.Intent;
 import android.os.Bundle;
+import android.os.Handler;
+import android.os.Looper;
+
+import androidx.annotation.NonNull;
 import android.util.Log;
 import android.widget.Toast;
 
@@ -139,10 +143,27 @@ public class DispatchActivity extends AppCompatActivity {
 
 
     @Override
+    protected void onNewIntent(@NonNull Intent intent) {
+        super.onNewIntent(intent);
+        setIntent(intent);
+        if (shouldFinish) {
+            // onResume already ran and posted a finish();
+            // but a new intent has arrived so cancel that decision and dispatch instead.
+            shouldFinish = false;
+            dispatch();
+        }
+    }
+
+    @Override
     protected void onResume() {
         super.onResume();
         if (shouldFinish) {
-            finish();
+            // Post finish() so that onNewIntent() — which Android may deliver synchronously
+            // after onResume() in the same Looper transaction — can cancel it by resetting
+            // shouldFinish before the posted runnable executes.
+            new Handler(Looper.getMainLooper()).post(() -> {
+                if (shouldFinish) finish();
+            });
         } else {
             dispatch();
         }
--- a/app/src/org/commcare/activities/DispatchActivity.java
+++ b/app/src/org/commcare/activities/DispatchActivity.java
@@ -465,6 +465,8 @@ public class DispatchActivity extends AppCompatActivity {
                         i.putStringArrayListExtra(SESSION_ENDPOINT_ARGUMENTS_LIST, argsList);
                         i.putExtra(CC_LAUNCH_REQUIRE_SYNC,
                                 getIntent().getBooleanExtra(CC_LAUNCH_REQUIRE_SYNC, false));
+                        i.putExtra(IS_LAUNCH_FROM_CONNECT,
+                                getIntent().getBooleanExtra(IS_LAUNCH_FROM_CONNECT, false));
 
                         // Session Endpoint extra is no longer needed. If not removed, it triggers
                         // the external launch logic in subsequent logins
--- a/app/src/org/commcare/activities/DispatchActivity.java
+++ b/app/src/org/commcare/activities/DispatchActivity.java
@@ -445,8 +445,6 @@ public class DispatchActivity extends AppCompatActivity {
                         i.putStringArrayListExtra(SESSION_ENDPOINT_ARGUMENTS_LIST, argsList);
                         i.putExtra(CC_LAUNCH_REQUIRE_SYNC,
                                 getIntent().getBooleanExtra(CC_LAUNCH_REQUIRE_SYNC, false));
-                        i.putExtra(IS_LAUNCH_FROM_CONNECT,
-                                getIntent().getBooleanExtra(IS_LAUNCH_FROM_CONNECT, false));
 
                         // Session Endpoint extra is no longer needed. If not removed, it triggers
                         // the external launch logic in subsequent logins
--- a/app/src/org/commcare/activities/HomeScreenBaseActivity.java
+++ b/app/src/org/commcare/activities/HomeScreenBaseActivity.java
@@ -182,6 +182,9 @@ public abstract class HomeScreenBaseActivity<T> extends SyncCapableCommCareActiv
     private FirebaseMessagingDataSyncer dataSyncer;
     private boolean isVisible;
 
+    // Set when a session endpoint launch requires a blocking sync before navigating to the endpoint
+    private boolean pendingEndpointNavigationAfterSync = false;
+
     {
         dataSyncer = new FirebaseMessagingDataSyncer(this);
 
@@ -226,8 +229,15 @@ public abstract class HomeScreenBaseActivity<T> extends SyncCapableCommCareActiv
     private void processFromExternalLaunch(Bundle savedInstanceState) {
         if (savedInstanceState == null && getIntent().hasExtra(DispatchActivity.WAS_EXTERNAL)) {
             wasExternal = true;
-            if (processSessionEndpoint()) {
-                sessionNavigator.startNextSessionStep();
+            if (getIntent().getBooleanExtra(DispatchActivity.CC_LAUNCH_REQUIRE_SYNC, false)) {
+                getIntent().removeExtra(DispatchActivity.CC_LAUNCH_REQUIRE_SYNC);
+                pendingEndpointNavigationAfterSync = true;
+                redirectedInOnCreate = true;
+                triggerSync(false);
+            } else {
+                if (processSessionEndpoint()) {
+                    sessionNavigator.startNextSessionStep();
+                }
             }
         }
     }
@@ -1322,6 +1332,14 @@ public abstract class HomeScreenBaseActivity<T> extends SyncCapableCommCareActiv
 
     @Override
     public void onResumeSessionSafe() {
+        if (pendingEndpointNavigationAfterSync) {
+            pendingEndpointNavigationAfterSync = false;
+            if (processSessionEndpoint()) {
+                sessionNavigator.startNextSessionStep();
+            }
+            return;
+        }
+
         if (!redirectedInOnCreate && !sessionNavigationProceedingAfterOnResume) {
             refreshActionBar();
             attemptDispatchHomeScreen();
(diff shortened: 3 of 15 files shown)


app/src/org/commcare/models/database/connect/ConnectDatabaseUpgrader.java
```diff
@@ -1012,12 +1018,42 @@ private void upgradeTwentyThreeTwentyFour(IDatabase db) {
...
+    }
+
+    private void upgradeTwentyFourTwentyFive(IDatabase db) {
+        db.beginTransaction();
+        try {
+            db.execSQL(DbUtil.addColumnToTable(
+                    PushNotificationRecord.STORAGE_KEY,
+                    PushNotificationRecord.META_SESSION_ENDPOINT_ID,
+                    "TEXT"));
+
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3684/commits/f9eb89a010b35192d9cd5f1bad00a0d25842d44e

app/src/org/commcare/android/database/connect/models/PushNotificationRecordV24.kt
```diff
@@ -0,0 +1,123 @@
...
+
+    @Persisting(17)
+    @MetaField(META_OPPORTUNITY_STATUS)
+    var opportunityStatus: String = ""
+
+    companion object {
+        const val STORAGE_KEY = PushNotificationRecord.STORAGE_KEY
+
+        fun fromV23(pushNotificationRecordV23: PushNotificationRecordV23): PushNotificationRecord =
+            PushNotificationRecord().apply {
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3684/commits/f9eb89a010b35192d9cd5f1bad00a0d25842d44e

app/src/org/commcare/activities/HomeScreenBaseActivity.java
```diff
@@ -181,6 +182,9 @@ public abstract class HomeScreenBaseActivity<T> extends SyncCapableCommCareActiv
     private FirebaseMessagingDataSyncer dataSyncer;
     private boolean isVisible;
 
+    // Set when a session endpoint launch requires a blocking sync before navigating to the endpoint
+    private boolean pendingEndpointNavigationAfterSync = false;
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3684/commits/1722434403c35b58d0bf10d918fe625d7bd4aa07

app/src/org/commcare/activities/HomeScreenBaseActivity.java
```diff
@@ -1320,6 +1332,14 @@ private void triggerSync(boolean triggeredByAutoSyncPending) {
 
     @Override
     public void onResumeSessionSafe() {
+        if (pendingEndpointNavigationAfterSync) {
+            pendingEndpointNavigationAfterSync = false;
+            if (processSessionEndpoint()) {
+                sessionNavigator.startNextSessionStep();
+            }
+            return;
+        }
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3684/commits/97cd24187250a3e30fcd9380c2cd374e3782cca0

app/src/org/commcare/utils/FirebaseMessagingUtil.java
```diff
@@ -350,11 +355,53 @@ private static Intent handleResumeLearningOrDeliveryJobPushNotification(Boolean
...
+        String sessionEndpointId = fcmMessageData.getPayloadData()
+                .get(PushNotificationRecord.META_SESSION_ENDPOINT_ID);
+        Intent intent;
+        if (!TextUtils.isEmpty(sessionEndpointId)) {
+            intent =  handleSessionEndpointNotification(context, fcmMessageData, showNotification, sessionEndpointId);
+            if (intent != null) {
+                return intent;
+            }
+        }
+        intent = getConnectActivityNotification(context, fcmMessageData);
```
Jignesh-dimagi: nit: This can be simplified further by
intent = intent==null?getConnectActivityNotification(context, fcmMessageData):intent; 
and having the code to show notification here only, thus this will allow removing showNotification from handleSessionEndpointNotification.

app/src/org/commcare/utils/FirebaseMessagingUtil.java
```diff
@@ -350,11 +355,53 @@ private static Intent handleResumeLearningOrDeliveryJobPushNotification(Boolean
...
+            Context context, FCMMessageData fcmMessageData, boolean showNotification,
+            String sessionEndpointId) {
+        Map<String, String> payloadData = fcmMessageData.getPayloadData();
+        String opportunityID = payloadData.get(OPPORTUNITY_UUID);
+        String opportunityStatus = payloadData.get(OPPORTUNITY_STATUS);
+        String appId = null;
+        try {
+            appId = ConnectJobUtils.getAppIdForOpportunity(context, opportunityID, opportunityStatus);
+        } catch (IllegalArgumentException e) {
+            Logger.exception("Could not resolve appId for opportunity", e);
```
Jignesh-dimagi, line 385: return null here only

app/src/org/commcare/utils/FirebaseMessagingUtil.java
```diff
@@ -350,11 +355,53 @@ private static Intent handleResumeLearningOrDeliveryJobPushNotification(Boolean
...
+
+    private static Intent handleSessionEndpointNotification(
+            Context context, FCMMessageData fcmMessageData, boolean showNotification,
+            String sessionEndpointId) {
+        Map<String, String> payloadData = fcmMessageData.getPayloadData();
+        String opportunityID = payloadData.get(OPPORTUNITY_UUID);
+        String opportunityStatus = payloadData.get(OPPORTUNITY_STATUS);
+        String appId = null;
+        try {
+            appId = ConnectJobUtils.getAppIdForOpportunity(context, opportunityID, opportunityStatus);
```
Jignesh-dimagi, line 383: This is rare, but just trying to understand what will happen if the user has still not downloaded the app, like after account recovery.

app/src/org/commcare/activities/DispatchActivity.java
```diff
@@ -137,11 +142,28 @@ private boolean finishIfNotRoot() {
...
+        super.onNewIntent(intent);
+        setIntent(intent);
+        if (shouldFinish) {
+            // onResume already ran and posted a finish();
+            // but a new intent has arrived so cancel that decision and dispatch instead.
+            shouldFinish = false;
+            dispatch();
+        }
+    }
+
```
Jignesh-dimagi: Theoretically, onNewIntent is called before onResume, but I agree it cannot be guaranteed. But in case dispatch() is called twice?

app/src/org/commcare/activities/HomeScreenBaseActivity.java
```diff
@@ -225,20 +232,29 @@ private void loadInstanceState(Bundle savedInstanceState) {
...
             wasExternal = true;
-            if (processSessionEndpoint()) {
-                sessionNavigator.startNextSessionStep();
+            if (getIntent().getBooleanExtra(DispatchActivity.CC_LAUNCH_REQUIRE_SYNC, false)) {
+                getIntent().removeExtra(DispatchActivity.CC_LAUNCH_REQUIRE_SYNC);
+                pendingEndpointNavigationAfterSync = true;
+                redirectedInOnCreate = true;
+                triggerSync(false);
+            } else {
+                if (processSessionEndpoint()) {
```
conroy-ricketts: Nit: this can be combined into a single `else if` branch

app/src/org/commcare/activities/HomeScreenBaseActivity.java
```diff
@@ -1320,6 +1337,14 @@ private void triggerSync(boolean triggeredByAutoSyncPending) {
 
     @Override
     public void onResumeSessionSafe() {
+        if (pendingEndpointNavigationAfterSync) {
+            if (processSessionEndpoint()) {
```
conroy-ricketts: Nit: can combine these into one branch with `&&` (unless this was intentional for performance gains?)

RELEASES.md
```diff
@@ -26,6 +28,8 @@ These notes are only published internally in [CommCare Change log wiki](https://
 along with the public release notes above
 -->
 
+- - Session endpoint navigation from Connect notifications: clicking a notification with a `session_endpoint_id` now navigates the user directly to the specified CommCare session endpoint (after a sync if required), instead of opening the Connect activity.
```
conroy-ricketts: Nit: there are two bullet points here for one

```suggestion
- Session endpoint navigation from Connect notifications: clicking a notification with a `session_endpoint_id` now navigates the user directly to the specified CommCare session endpoint (after a sync if required), instead of opening the Connect activity.
```

app/unit-tests/src/org/commcare/utils/SessionEndpointNotificationTest.kt
```diff
@@ -0,0 +1,168 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.MockedStatic
+import org.mockito.Mockito
+import org.robolectric.annotation.Config
+
+@Config(application = CommCareTestApplication::class)
+@RunWith(AndroidJUnit4::class)
+class SessionEndpointNotificationTest {
+    private val context: Context = CommCareTestApplication.instance()
```
conroy-ricketts: Nit: type here is unnecessary/redundant

```suggestion
    private val context = CommCareTestApplication.instance()
```

app/unit-tests/src/org/commcare/utils/SessionEndpointNotificationTest.kt
```diff
@@ -0,0 +1,168 @@
...
+    private lateinit var connectJobUtilsMock: MockedStatic<ConnectJobUtils>
+    private lateinit var savedStatus: PersonalIdManager.PersonalIdStatus
+
+    @Before
+    fun setUp() {
+        // Set PersonalIdManager status to LoggedIn directly to satisfy cccCheckPassed().
+        // init() only touches the DB when status == NotIntroduced; once LoggedIn it's a no-op.
+        val manager = PersonalIdManager.getInstance()
+        savedStatus = manager.getStatus()
+        manager.setStatus(PersonalIdManager.PersonalIdStatus.LoggedIn)
```
conroy-ricketts: Nit: can treat these as properties with Kotlin

```suggestion
        savedStatus = manager.status
        manager.status = PersonalIdManager.PersonalIdStatus.LoggedIn
```

app/unit-tests/src/org/commcare/utils/SessionEndpointNotificationTest.kt
```diff
@@ -0,0 +1,168 @@
...
+                    Mockito.eq(opportunityUuidValue),
+                    Mockito.eq(ConnectConstants.OPPORTUNITY_STATUS_DELIVERY),
+                )
+            }.thenReturn(deliveryAppId)
+    }
+
+    @After
+    fun tearDown() {
+        connectJobUtilsMock.close()
+        PersonalIdManager.getInstance().setStatus(savedStatus)
```
conroy-ricketts: Nit: can use Kotlin's property access syntax here too

app/unit-tests/src/org/commcare/utils/SessionEndpointNotificationTest.kt
```diff
@@ -0,0 +1,168 @@
...
+            map["opportunity_uuid"] = opportunityUuid
+        }
+        if (opportunityStatus != null) {
+            map["opportunity_status"] = opportunityStatus
+        }
+        return map
+    }
+
+    @Test
+    fun `ccc_generic_opportunity with session_endpoint_id routes to DispatchActivity`() {
```
conroy-ricketts, line 96: Think this is great, but this probably warrants a larger team discussion. There is an inconsistency with the format of our unit test names and it's worth being aligned on it.

For example, `PhoneNumberHelperTest.kt` uses this format:
`fun testPhoneNumberValidation_invalidShortNumber()`

Here's a good summary from Claude if that doesn't make sense:
```
  - PhoneNumberHelperTest uses camelCase with a test prefix and underscore-separated scenario (e.g. testPhoneNumberValidation_invalidShortNumber). This is the older JUnit 3 / Java-style convention (sometimes called             
  testMethodName_scenario or "Roy Osherove's UnitOfWork_Scenario_ExpectedBehavior" naming). 
                                                                                                            
  - SessionEndpointNotificationTest uses Kotlin backtick-quoted identifiers containing spaces to write the test name as a natural-language sentence (e.g. `ccc_generic_opportunity with session_endpoint_id routes to              
  DispatchActivity`). This is commonly called "backtick test names", "spec-style", or "sentence-style" test naming — a Kotlin idiom enabled by the language allowing arbitrary identifiers between backticks.  
```

conroy-ricketts (dismissed):
LGTM! Think it's worth at least one more pair of eyes though

OrangeAndGreen (commented):
Something that jumps out to me early, are we requiring the user to unlock PersonalId before auto-logging into the app? Seems like this could be a security concern if someone can click on a notification and be taken directly into the app.

app/src/org/commcare/utils/FirebaseMessagingUtil.java
OrangeAndGreen, line 385: To consider: We'll hit this if a user clicks on a notification with a session endpoint but is not currently signed in to PersonalId (i.e. the Connect DB doesn't exist). The PersonalId server hangs on to FCM tokens even if a user logs out of PersonalId (or reinstalls the CommCare app), so it's possible for users to receive FCM messages on a device that is no longer signed in to PersonalId.

Perhaps we should only log the non-fatal exception if the user is logged in to PersonalId, and just let the non-logged in case fall through silently (I think ConnectActivity will show a message to the user telling them they need to login?).

This is probably not extremely common in the field, but Dimagi users testing things across devices could run into it (I know I have) and clutter up the Firebase log to make "real" issues more difficult to catch.

shubham1g5:
> Something that jumps out to me early, are we requiring the user to unlock PersonalId before auto-logging into the app? Seems like this could be a security concern if someone can click on a notification and be taken directly into the app.

Ahh great callout and definitely a gap in implementation. I am afraid this limitation exists for other notifications as well although the app login does make it more critical. I would do a different PR on that soon in scope of 2.63  itself. Also think this will probably get into weeds of https://dimagi.atlassian.net/browse/CCCT-2078 as well.

shubham1g5:
>  I would do a different PR on that soon in scope of 2.63 itself. Also think this will probably get into weeds of https://dimagi.atlassian.net/browse/CCCT-2078 as well.

that didn't turned out to be the case, I chose for an easier implementation here  (video updated) and is modified by these 2 commits - https://github.com/dimagi/commcare-android/pull/3684/commits/be30dd7cf678090d7fab34d90e07e20e5538e1b2 and https://github.com/dimagi/commcare-android/pull/3684/commits/3761fdde758d0b7686c8f594a21b400b0bfde6e9

app/src/org/commcare/activities/DispatchActivity.java
```diff
@@ -442,6 +463,10 @@ private void handleExternalLaunch() {
                         i.putExtra(SESSION_ENDPOINT_ID, sessionEndpointId);
                         i.putExtra(SESSION_ENDPOINT_ARGUMENTS_BUNDLE, args);
                         i.putStringArrayListExtra(SESSION_ENDPOINT_ARGUMENTS_LIST, argsList);
+                        i.putExtra(CC_LAUNCH_REQUIRE_SYNC,
+                                getIntent().getBooleanExtra(CC_LAUNCH_REQUIRE_SYNC, false));
+                        i.putExtra(IS_LAUNCH_FROM_CONNECT,
+                                getIntent().getBooleanExtra(IS_LAUNCH_FROM_CONNECT, false));
```
Jignesh-dimagi: These intents need to be removed after being consumed. Just like [getLaunchedFromConnect()](https://github.com/dimagi/commcare-android/blob/38607d335cc2555f433d552e82a68411baf677e8/app/src/org/commcare/activities/DispatchActivity.java#L352)

app/unit-tests/src/org/commcare/utils/SessionEndpointNotificationTest.kt
shubham1g5: Agree on inconsistency and think I am flexible in general to use any of these styles, I only found about it recently and think it's easier to understand than the underscored version, but I am in general open to have both styles in the code base as I don't think that adds an overhead for us in terms of how we write or analyze tests. But agree that it would be good to have thoughts from the wider team here.

app/src/org/commcare/activities/HomeScreenBaseActivity.java
shubham1g5: https://github.com/dimagi/commcare-android/pull/3684/commits/98bdd789a580981c9e5884bdd4fb76445a814245

app/src/org/commcare/activities/HomeScreenBaseActivity.java
shubham1g5: https://github.com/dimagi/commcare-android/pull/3684/commits/98bdd789a580981c9e5884bdd4fb76445a814245

RELEASES.md
shubham1g5: https://github.com/dimagi/commcare-android/pull/3684/commits/98bdd789a580981c9e5884bdd4fb76445a814245

app/unit-tests/src/org/commcare/utils/SessionEndpointNotificationTest.kt
shubham1g5: https://github.com/dimagi/commcare-android/pull/3684/commits/98bdd789a580981c9e5884bdd4fb76445a814245

app/unit-tests/src/org/commcare/utils/SessionEndpointNotificationTest.kt
shubham1g5: https://github.com/dimagi/commcare-android/pull/3684/commits/98bdd789a580981c9e5884bdd4fb76445a814245

app/unit-tests/src/org/commcare/utils/SessionEndpointNotificationTest.kt
shubham1g5: https://github.com/dimagi/commcare-android/pull/3684/commits/98bdd789a580981c9e5884bdd4fb76445a814245

app/src/org/commcare/utils/FirebaseMessagingUtil.java
shubham1g5: we will still have the `appId` but would not be able to find it while launching notification's pending intent, so we will just end up on the default launch screen for the user depending on their state (setup or login)

app/src/org/commcare/utils/FirebaseMessagingUtil.java
shubham1g5: https://github.com/dimagi/commcare-android/pull/3684/commits/515ed80284f69af95b71f616765fcacefa89e253

app/src/org/commcare/utils/FirebaseMessagingUtil.java
shubham1g5: https://github.com/dimagi/commcare-android/pull/3684/commits/515ed80284f69af95b71f616765fcacefa89e253

app/src/org/commcare/activities/DispatchActivity.java
shubham1g5: yeah that seems to be the case, let me add another flag here :/

shubham1g5: actually we don't need onNewIntent override anymore as I decided to launch notification in a clean task due to these sort of issues - https://github.com/dimagi/commcare-android/pull/3684/commits/465a376fdc78f188ef7ee30ea4451f6f40ee889c , I have reverted this change in https://github.com/dimagi/commcare-android/pull/3684/commits/e4f5b1436302197cdc7848ab62e7014cd1fd692a

app/src/org/commcare/activities/DispatchActivity.java
shubham1g5: we no longer need this extra, removed here - https://github.com/dimagi/commcare-android/pull/3684/commits/97da34aac35129622040d3593e4033856bee8f52

Jignesh-dimagi: Ok. I think we remove `CC_LAUNCH_REQUIRE_SYNC` also like done [here](https://github.com/dimagi/commcare-android/blob/38607d335cc2555f433d552e82a68411baf677e8/app/src/org/commcare/activities/DispatchActivity.java#L448)

shubham1g5: it's getting removed [here](https://github.com/dimagi/commcare-android/blob/ccct-2217_ccct-2329_notifications_with_session_endpoint/app/src/org/commcare/activities/HomeScreenBaseActivity.java#L236)

Jignesh-dimagi (dismissed):

RELEASES.md
conroy-ricketts: This may be the wrong commit?

app/unit-tests/src/org/commcare/utils/SessionEndpointNotificationTest.kt
conroy-ricketts: Tagging @OrangeAndGreen @Jignesh-dimagi here as well. This could be something that we add to our coding standards

I much prefer the way Shubham named the tests here

conroy-ricketts: And also, I think it's fine to keep our current tests as-is, but move forward with a shared standard

Jignesh-dimagi: I am ok with any, but this is much easier to understand. I used to write in my previous company with this format only but am surprised why Claude is not recommending it.

conroy-ricketts: Yeah I used this format at my last company too

conroy-ricketts (dismissed):

shubham1g5:
@conroy-ricketts @Jignesh-dimagi the last 4 commits are some more fixes that I found testing with a form's session endpoint, I have also added a video to reflect that testing in PR description.

shubham1g5:

conroy-ricketts (approved):

Jignesh-dimagi (approved):

OrangeAndGreen (approved):

app/src/org/commcare/utils/FirebaseMessagingUtil.java
shubham1g5: skipping to get the merge in for code freeze, but curious what's the failure mode when user is not logged into PersonalID ? Does the querying for job failed because the DB is locked or something simiar ?

PR #3685, opened 2026-04-30
CCCT-2052 Track OTP Request/Verify Outcome And Method

Jignesh-dimagi:
## Technical Summary

Extends the existing `otp_requested` Firebase Analytics event with four new
parameters — `event_type`, `outcome`, `method`, and `reason` — so OTP
success/failure rates can be broken down by request vs verify and by Firebase
vs PersonalID. The Firebase event name and the existing `value` (attempts)
parameter are preserved for dashboard backwards compatibility.

Key changes:

- `AnalyticsParamValue.java` — adds `OTP_EVENT_TYPE_REQUEST/VERIFY`,
  `OTP_OUTCOME_SUCCESS/FAILURE`, `OTP_METHOD_FIREBASE/PERSONAL_ID` value
  constants.
- `CCAnalyticsParam.java` — adds `OTP_EVENT_TYPE`, `OTP_OUTCOME`, `OTP_METHOD`
  param-key constants. Widens the existing `REASON` constant from
  package-private to `public` so it can be referenced from the fragment.
- `FirebaseAnalyticsUtil.java` — replaces `reportOtpRequested(int)` with
  `reportOtpEvent(eventType, outcome, method, reason, attempts)`. Null
  `reason` is omitted from the bundle (Firebase rejects null string params).
- `OtpAnalyticsMapper.kt` (new) — normalizes the `OtpManager` SMS-method
  string (`firebase` / `personal_id`, case-insensitive, defaults to
  `firebase`) and the `OtpErrorType` / `PersonalIdOrConnectApiErrorCodes`
  enums into stable lowercase analytics strings (e.g. `INVALID_CREDENTIAL` →
  `invalid_credential`).
- `PersonalIdPhoneVerificationFragment.java` — adds an `OtpOp { REQUEST,
  VERIFY }` enum + `currentOtpOp` field, set immediately before each kick-off,
  so the next `OtpVerificationCallback` terminal can attribute itself
  correctly. Emission moves out of the kick-off (`requestOtp`) and into the
  four terminals (`onCodeSent`, `onSuccess`, `onFailure`,
  `onPersonalIdApiFailure`). The Firebase auto-verify `onCodeVerified` path is
  intentionally silent — the server-confirmed `onSuccess` is the authoritative
  verify-success terminal, and emitting on both would double-count.

Design rationale: the helper uses `lastOtpMethod` (the active manager's
method) rather than `personalIdSessionData.getSmsMethod()` (the server-
assigned value), because `setupOtpManager` may fall back or auto-switch
between Firebase and PersonalID independently of the server-assigned value.
In the auto-switch path inside `onFailure`, the original-method failure is
emitted before `setupOtpManager(true)` flips `lastOtpMethod`, and the
follow-up request emits its own event under the new method.

### Safety story

The change is additive and back-compat-preserving from a Firebase Analytics
perspective:

- The event NAME (`otp_requested`) is unchanged. Existing dashboards keyed on
  the event name continue to fire.
- The existing `value` parameter (attempts count) is still emitted with the
  same semantics. Dashboards keyed only on `value` continue to work.
- New breakdowns become available via the four added params; consumers must
  opt in.

The four new fragment callback emissions all early-return when `otpCallback`
is null (cleared in `onDestroyView`), preventing post-destroy NPEs from
late-arriving callbacks. The mapper defaults to `firebase` for null/unknown
method strings so events always carry a method value.

### Automated test coverage

`OtpAnalyticsMapperTest.kt` covers the mapper with 18 JUnit 4 tests across
`methodFromSmsMethod`, `reasonFrom(OtpErrorType)`, and
`reasonFrom(PersonalIdOrConnectApiErrorCodes)` — including null inputs,
case-insensitivity, and the unrecognized-method fallback. The fragment-level
emission wiring is not covered by an automated test (no test exists for this
fragment today); regression coverage relies on manual verification.

[CCCT-2052]: https://dimagi.atlassian.net/browse/CCCT-2052?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
+++ b/app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
@@ -31,9 +31,11 @@ import org.commcare.connect.network.base.BaseApiHandler;
 import org.commcare.connect.network.PersonalIdOrConnectApiErrorHandler;
 import org.commcare.dalvik.R;
 import org.commcare.dalvik.databinding.ScreenPersonalidPhoneVerifyBinding;
+import org.commcare.google.services.analytics.AnalyticsParamValue;
 import org.commcare.google.services.analytics.FirebaseAnalyticsUtil;
 import org.commcare.util.LogTypes;
 import org.commcare.utils.KeyboardHelper;
+import org.commcare.utils.OtpAnalyticsMapper;
 import org.commcare.utils.OtpErrorType;
 import org.commcare.utils.OtpManager;
 import org.commcare.utils.OtpVerificationCallback;
@@ -60,6 +62,9 @@ public class PersonalIdPhoneVerificationFragment extends BasePersonalIdFragment 
     private SMSBroadcastReceiver smsBroadcastReceiver;
     private ScreenPersonalidPhoneVerifyBinding binding;
     private final Handler resendTimerHandler = new Handler();
+    /** Which OTP call is currently in flight, used to attribute the next callback. */
+    private enum OtpOp { REQUEST, VERIFY }
+    private OtpOp currentOtpOp;
     private OtpManager otpManager;
     private PersonalIdSessionData personalIdSessionData;
     OtpVerificationCallback otpCallback;
@@ -97,23 +102,34 @@ public class PersonalIdPhoneVerificationFragment extends BasePersonalIdFragment 
             @Override
             public void onCodeSent(String verificationId) {
                 if (otpCallback == null) return;
+                emitOtpAnalytics(AnalyticsParamValue.OTP_OUTCOME_SUCCESS, /* reason= */ null);
                 Toast.makeText(requireContext(), getString(R.string.connect_otp_sent), Toast.LENGTH_SHORT).show();
             }
 
             @Override
             public void onCodeVerified(String code) {
                 if (otpCallback == null) return;
+                // No analytics emission here: this is Firebase's local auto-verification step.
+                // Verify success is recorded from the server-confirmed onSuccess callback so
+                // we don't double-count, and verify failure is recorded from onFailure /
+                // onPersonalIdApiFailure.
                 Toast.makeText(requireContext(), getString(R.string.connect_otp_verified), Toast.LENGTH_SHORT).show();
             }
 
             @Override
             public void onSuccess() {
+                if (otpCallback == null) return;
+                emitOtpAnalytics(AnalyticsParamValue.OTP_OUTCOME_SUCCESS, /* reason= */ null);
                 navigateToNameEntry();
             }
 
             @Override
             public void onFailure(OtpErrorType errorType, @Nullable String errorMessage) {
                 if (otpCallback == null) return;
+
+                emitOtpAnalytics(
+                        AnalyticsParamValue.OTP_OUTCOME_FAILURE,
+                        OtpAnalyticsMapper.reasonFrom(errorType));
 
                 // Auto-switch from Firebase to PersonalId for non-recoverable errors
                 if (shouldAutoSwitchToPersonalIdAuth(errorType)) {
@@ -140,6 +156,12 @@ public class PersonalIdPhoneVerificationFragment extends BasePersonalIdFragment 
                     @NonNull BaseApiHandler.PersonalIdOrConnectApiErrorCodes failureCode,
                     Throwable t
             ) {
+                if (otpCallback == null) return;
+
+                emitOtpAnalytics(
+                        AnalyticsParamValue.OTP_OUTCOME_FAILURE,
+                        OtpAnalyticsMapper.reasonFrom(failureCode));
+
                 if (handleCommonSignupFailures(failureCode)) {
                     return;
                 }
@@ -329,12 +351,32 @@ public class PersonalIdPhoneVerificationFragment extends BasePersonalIdFragment 
         }
     }
 
+    private void emitOtpAnalytics(String outcome, @Nullable String reason) {
+        if (currentOtpOp == null) {
+            // Defensive — a callback fired without a tracked op. Don't emit.
+            return;
+        }
+        String eventType = currentOtpOp == OtpOp.REQUEST
+                ? AnalyticsParamValue.OTP_EVENT_TYPE_REQUEST
+                : AnalyticsParamValue.OTP_EVENT_TYPE_VERIFY;
+        // Use lastOtpMethod (the active manager's method) rather than
+        // personalIdSessionData.getSmsMethod(), which is the server-assigned
+        // value and does not reflect setupOtpManager's fallback/auto-switch logic.
+        String method = OtpAnalyticsMapper.methodFromSmsMethod(lastOtpMethod);
+        FirebaseAnalyticsUtil.reportOtpEvent(
+                eventType,
+                outcome,
+                method,
+                reason,
+                personalIdSessionData.getOtpAttempts());
+    }
+
     private void requestOtp() {
         clearOtpError();
         otpRequestTime = new DateTime();
+        currentOtpOp = OtpOp.REQUEST;
         otpManager.requestOtp(primaryPhone);
         personalIdSessionData.setOtpAttempts(personalIdSessionData.getOtpAttempts() + 1);
-        FirebaseAnalyticsUtil.reportOtpRequested(personalIdSessionData.getOtpAttempts());
     }
 
     private void verifyOtp() {
@@ -345,6 +387,7 @@ public class PersonalIdPhoneVerificationFragment extends BasePersonalIdFragment 
         if (otpCode.length() != 6) {
             Toast.makeText(requireContext(), getString(R.string.connect_enter_otp), Toast.LENGTH_SHORT).show();
         } else {
+            currentOtpOp = OtpOp.VERIFY;
             otpManager.verifyOtp(otpCode);
         }
     }
--- a/app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
+++ b/app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
@@ -46,6 +46,7 @@ import org.joda.time.DateTime;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import static org.commcare.utils.OtpAnalyticsMapper.getEventType;
 import static org.commcare.utils.OtpManager.SMS_METHOD_FIREBASE;
 import static org.commcare.utils.OtpManager.SMS_METHOD_PERSONAL_ID;
 
@@ -62,9 +63,7 @@ public class PersonalIdPhoneVerificationFragment extends BasePersonalIdFragment 
     private SMSBroadcastReceiver smsBroadcastReceiver;
     private ScreenPersonalidPhoneVerifyBinding binding;
     private final Handler resendTimerHandler = new Handler();
-    /** Which OTP call is currently in flight, used to attribute the next callback. */
-    private enum OtpOp { REQUEST, VERIFY }
-    private OtpOp currentOtpOp;
+    private OtpAnalyticsMapper.OtpOp currentOtpOp;
     private OtpManager otpManager;
     private PersonalIdSessionData personalIdSessionData;
     OtpVerificationCallback otpCallback;
@@ -102,7 +101,7 @@ public class PersonalIdPhoneVerificationFragment extends BasePersonalIdFragment 
             @Override
             public void onCodeSent(String verificationId) {
                 if (otpCallback == null) return;
-                emitOtpAnalytics(AnalyticsParamValue.OTP_OUTCOME_SUCCESS, /* reason= */ null);
+                reportOtpAnalytics(AnalyticsParamValue.OTP_OUTCOME_SUCCESS, null);
                 Toast.makeText(requireContext(), getString(R.string.connect_otp_sent), Toast.LENGTH_SHORT).show();
             }
 
@@ -119,17 +118,17 @@ public class PersonalIdPhoneVerificationFragment extends BasePersonalIdFragment 
             @Override
             public void onSuccess() {
                 if (otpCallback == null) return;
-                emitOtpAnalytics(AnalyticsParamValue.OTP_OUTCOME_SUCCESS, /* reason= */ null);
+                reportOtpAnalytics(AnalyticsParamValue.OTP_OUTCOME_SUCCESS, null);
                 navigateToNameEntry();
             }
 
             @Override
             public void onFailure(OtpErrorType errorType, @Nullable String errorMessage) {
-                if (otpCallback == null) return;
-
-                emitOtpAnalytics(
+                reportOtpAnalytics(
                         AnalyticsParamValue.OTP_OUTCOME_FAILURE,
                         OtpAnalyticsMapper.reasonFrom(errorType));
+
+                if (otpCallback == null) return;
 
                 // Auto-switch from Firebase to PersonalId for non-recoverable errors
                 if (shouldAutoSwitchToPersonalIdAuth(errorType)) {
@@ -158,7 +157,7 @@ public class PersonalIdPhoneVerificationFragment extends BasePersonalIdFragment 
             ) {
                 if (otpCallback == null) return;
 
-                emitOtpAnalytics(
+                reportOtpAnalytics(
                         AnalyticsParamValue.OTP_OUTCOME_FAILURE,
                         OtpAnalyticsMapper.reasonFrom(failureCode));
 
@@ -351,14 +350,11 @@ public class PersonalIdPhoneVerificationFragment extends BasePersonalIdFragment 
         }
     }
 
-    private void emitOtpAnalytics(String outcome, @Nullable String reason) {
-        if (currentOtpOp == null) {
-            // Defensive — a callback fired without a tracked op. Don't emit.
-            return;
-        }
-        String eventType = currentOtpOp == OtpOp.REQUEST
-                ? AnalyticsParamValue.OTP_EVENT_TYPE_REQUEST
-                : AnalyticsParamValue.OTP_EVENT_TYPE_VERIFY;
+    private void reportOtpAnalytics(String outcome, @Nullable String reason) {
+
+        String eventType = getEventType(currentOtpOp);
+        if (eventType == null) return;
+
         // Use lastOtpMethod (the active manager's method) rather than
         // personalIdSessionData.getSmsMethod(), which is the server-assigned
         // value and does not reflect setupOtpManager's fallback/auto-switch logic.
@@ -374,7 +370,7 @@ public class PersonalIdPhoneVerificationFragment extends BasePersonalIdFragment 
     private void requestOtp() {
         clearOtpError();
         otpRequestTime = new DateTime();
-        currentOtpOp = OtpOp.REQUEST;
+        currentOtpOp = OtpAnalyticsMapper.OtpOp.REQUEST;
         otpManager.requestOtp(primaryPhone);
         personalIdSessionData.setOtpAttempts(personalIdSessionData.getOtpAttempts() + 1);
     }
@@ -387,7 +383,7 @@ public class PersonalIdPhoneVerificationFragment extends BasePersonalIdFragment 
         if (otpCode.length() != 6) {
             Toast.makeText(requireContext(), getString(R.string.connect_enter_otp), Toast.LENGTH_SHORT).show();
         } else {
-            currentOtpOp = OtpOp.VERIFY;
+            currentOtpOp = OtpAnalyticsMapper.OtpOp.VERIFY;
             otpManager.verifyOtp(otpCode);
         }
     }
--- a/app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
+++ b/app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
@@ -351,13 +351,8 @@ public class PersonalIdPhoneVerificationFragment extends BasePersonalIdFragment 
     }
 
     private void reportOtpAnalytics(String outcome, @Nullable String reason) {
-
-        String eventType = getEventType(currentOtpOp);
+      String eventType = getEventType(currentOtpOp);
         if (eventType == null) return;
-
-        // Use lastOtpMethod (the active manager's method) rather than
-        // personalIdSessionData.getSmsMethod(), which is the server-assigned
-        // value and does not reflect setupOtpManager's fallback/auto-switch logic.
         String method = OtpAnalyticsMapper.methodFromSmsMethod(lastOtpMethod);
         FirebaseAnalyticsUtil.reportOtpEvent(
                 eventType,
--- a/app/src/org/commcare/google/services/analytics/AnalyticsParamValue.java
+++ b/app/src/org/commcare/google/services/analytics/AnalyticsParamValue.java
@@ -211,4 +211,12 @@ public class AnalyticsParamValue {
             "click_resubscribed_confirm";
     public final static String CCC_MESSAGING_EVENT_TYPE_CONSENT_API_RESULT =
             "consent_api_success_result";
+
+    // Param values for OTP analytics (CCCT-2052)
+    public static final String OTP_METHOD_FIREBASE = "firebase";
+    public static final String OTP_METHOD_PERSONAL_ID = "personal_id";
+    public static final String OTP_OUTCOME_SUCCESS = "success";
+    public static final String OTP_OUTCOME_FAILURE = "failure";
+    public static final String OTP_EVENT_TYPE_REQUEST = "request";
+    public static final String OTP_EVENT_TYPE_VERIFY = "verify";
 }
--- a/app/src/org/commcare/google/services/analytics/AnalyticsParamValue.java
+++ b/app/src/org/commcare/google/services/analytics/AnalyticsParamValue.java
@@ -212,7 +212,7 @@ public class AnalyticsParamValue {
     public final static String CCC_MESSAGING_EVENT_TYPE_CONSENT_API_RESULT =
             "consent_api_success_result";
 
-    // Param values for OTP analytics (CCCT-2052)
+    // Param values for OTP analytics
     public static final String OTP_METHOD_FIREBASE = "firebase";
     public static final String OTP_METHOD_PERSONAL_ID = "personal_id";
     public static final String OTP_OUTCOME_SUCCESS = "success";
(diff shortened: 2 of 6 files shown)


app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
```diff
@@ -329,12 +351,32 @@ private void updateVerificationMessage() {
...
+        // personalIdSessionData.getSmsMethod(), which is the server-assigned
+        // value and does not reflect setupOtpManager's fallback/auto-switch logic.
+        String method = OtpAnalyticsMapper.methodFromSmsMethod(lastOtpMethod);
+        FirebaseAnalyticsUtil.reportOtpEvent(
+                eventType,
+                outcome,
+                method,
+                reason,
+                personalIdSessionData.getOtpAttempts());
+    }
```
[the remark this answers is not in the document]
Jignesh-dimagi: @coderabbit

The user will receive the OTP on their phone, so it will definitely have a larger delay, so the chances of happening both at same is almost nil.

app/src/org/commcare/google/services/analytics/AnalyticsParamValue.java
```diff
@@ -211,4 +211,12 @@ public class AnalyticsParamValue {
             "click_resubscribed_confirm";
     public final static String CCC_MESSAGING_EVENT_TYPE_CONSENT_API_RESULT =
             "consent_api_success_result";
+
+    // Param values for OTP analytics (CCCT-2052)
```
shubham1g5: we should not be adding ticket numbers in the code unless absolutely necessary

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
```diff
@@ -97,24 +102,35 @@ private void initOtpManager() {
...
+                if (otpCallback == null) return;
+                emitOtpAnalytics(AnalyticsParamValue.OTP_OUTCOME_SUCCESS, /* reason= */ null);
                 navigateToNameEntry();
             }
 
             @Override
             public void onFailure(OtpErrorType errorType, @Nullable String errorMessage) {
                 if (otpCallback == null) return;
 
+                emitOtpAnalytics(
```
shubham1g5: should these emits come before the null `otpCallback ` null check ?

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
```diff
@@ -329,12 +351,32 @@ private void updateVerificationMessage() {
         }
     }
 
+    private void emitOtpAnalytics(String outcome, @Nullable String reason) {
+        if (currentOtpOp == null) {
+            // Defensive — a callback fired without a tracked op. Don't emit.
```
shubham1g5: think good to be defensive in analytics code, although would be nice to log this to crashlytics so that we know it's not happening often and screwing analytics.

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
```diff
@@ -329,12 +351,32 @@ private void updateVerificationMessage() {
         }
     }
 
+    private void emitOtpAnalytics(String outcome, @Nullable String reason) {
```
shubham1g5: nit: rename to `reportOtpAnalytics`, more consistent with existing naming plus `emit` makes it feel like a event based stream implemenation which it is not.

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
```diff
@@ -329,12 +351,32 @@ private void updateVerificationMessage() {
...
+        if (currentOtpOp == null) {
+            // Defensive — a callback fired without a tracked op. Don't emit.
+            return;
+        }
+        String eventType = currentOtpOp == OtpOp.REQUEST
+                ? AnalyticsParamValue.OTP_EVENT_TYPE_REQUEST
+                : AnalyticsParamValue.OTP_EVENT_TYPE_VERIFY;
+        // Use lastOtpMethod (the active manager's method) rather than
+        // personalIdSessionData.getSmsMethod(), which is the server-assigned
+        // value and does not reflect setupOtpManager's fallback/auto-switch logic.
```
shubham1g5: optional: think we can omit the comment as it's not adding value.

app/src/org/commcare/utils/OtpAnalyticsMapper.kt
```diff
@@ -0,0 +1,42 @@
...
+ */
+object OtpAnalyticsMapper {
+    /**
+     * Maps the OtpManager SMS method string ({@link OtpManager#SMS_METHOD_FIREBASE} /
+     * {@link OtpManager#SMS_METHOD_PERSONAL_ID}) to the analytics method constant.
+     * Falls back to {@link AnalyticsParamValue#OTP_METHOD_FIREBASE} when the input is
+     * null or unrecognized so events always carry a method value.
+     */
+    @JvmStatic
+    fun methodFromSmsMethod(smsMethod: String?): String =
```
shubham1g5: rename: `smsMethod` -> `otpMethod`

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
```diff
@@ -329,12 +351,32 @@ private void updateVerificationMessage() {
...
     }
 
+    private void emitOtpAnalytics(String outcome, @Nullable String reason) {
+        if (currentOtpOp == null) {
+            // Defensive — a callback fired without a tracked op. Don't emit.
+            return;
+        }
+        String eventType = currentOtpOp == OtpOp.REQUEST
+                ? AnalyticsParamValue.OTP_EVENT_TYPE_REQUEST
+                : AnalyticsParamValue.OTP_EVENT_TYPE_VERIFY;
```
shubham1g5: optional: should be part of the `OtpAnalyticsMapper` as well ?

app/src/org/commcare/utils/OtpAnalyticsMapper.kt
```diff
@@ -0,0 +1,42 @@
...
+        when {
+            smsMethod == null -> AnalyticsParamValue.OTP_METHOD_FIREBASE
+            smsMethod.equals(OtpManager.SMS_METHOD_PERSONAL_ID, ignoreCase = true) ->
+                AnalyticsParamValue.OTP_METHOD_PERSONAL_ID
+            smsMethod.equals(OtpManager.SMS_METHOD_FIREBASE, ignoreCase = true) ->
+                AnalyticsParamValue.OTP_METHOD_FIREBASE
+            else -> AnalyticsParamValue.OTP_METHOD_FIREBASE
```
shubham1g5: how about we just return the `smsMethod` as it is if it's neither of those 2 values  ?

app/unit-tests/src/org/commcare/utils/OtpAnalyticsMapperTest.kt
```diff
@@ -0,0 +1,157 @@
+package org.commcare.utils
+
+import org.commcare.connect.network.base.BaseApiHandler.PersonalIdOrConnectApiErrorCodes
+import org.commcare.google.services.analytics.AnalyticsParamValue
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNull
+import org.junit.Test
+
+class OtpAnalyticsMapperTest {
```
shubham1g5, line 9: optional: I would not really add these tests as they are using simple if-else conditions and mostly `errorType?.name?.lowercase()` which I don't think is adding any real value or safety to the PR.

shubham1g5 (changes requested):

app/src/org/commcare/google/services/analytics/FirebaseAnalyticsUtil.java
```diff
@@ -737,9 +737,37 @@ public static void reportNavDrawerItemSelected(String selectedItem) {
...
         Bundle bundle = new Bundle();
-        bundle.putLong(FirebaseAnalytics.Param.VALUE, numberOfAttempts);
+        bundle.putString(CCAnalyticsParam.OTP_EVENT_TYPE, eventType);
+        bundle.putString(CCAnalyticsParam.OTP_OUTCOME, outcome);
+        bundle.putString(CCAnalyticsParam.OTP_METHOD, method);
+        if (reason != null) {
+            bundle.putString(CCAnalyticsParam.REASON, reason);
+        }
+        bundle.putLong(FirebaseAnalytics.Param.VALUE, attempts);
         reportEvent(CCAnalyticsEvent.OTP_REQUESTED, bundle);
```
OrangeAndGreen, line 771: Noting this event name "otp_requested" isn't really accurate anymore since we do additional OTP related logging. I could see leaving it as is for continuity with old events, but given how much we're changing how these events are logged I wonder if continuity is useful? Changing the name now could create a good separation between the old and new reporting structures.

On the other hand, there are also concerns to consider about using additional event names (toward the limit of 500 total). Could go either way, but worth considering

app/src/org/commcare/utils/OtpAnalyticsMapper.kt
```diff
@@ -0,0 +1,42 @@
...
+
+import org.commcare.connect.network.base.BaseApiHandler.PersonalIdOrConnectApiErrorCodes
+import org.commcare.google.services.analytics.AnalyticsParamValue
+
+/**
+ * Normalizes Firebase {@link OtpErrorType}, {@link PersonalIdOrConnectApiErrorCodes},
+ * and the OtpManager SMS-method string into the stable analytics strings emitted with
+ * the {@code otp_requested} event (CCCT-2052).
+ */
+object OtpAnalyticsMapper {
```
OrangeAndGreen, line 13: Non-blocking, but maybe a bit more than a nit?

This feels like a lot of overhead to me for converting an analytics string. Since PersonalIdPhoneVerificationFragment is the only place that does the conversion, seems like methodFromSmsMethod could be a local helper method and the two calls to reasonFrom could just be local calls to errorType.name?.lowerCase().

I'm okay with keeping as is if that's what the team prefers, but it feels like it reduces readability for me to have the extra boilerplate to navigate through.

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
```diff
@@ -329,12 +351,32 @@ private void updateVerificationMessage() {
         }
     }
 
+    private void emitOtpAnalytics(String outcome, @Nullable String reason) {
+        if (currentOtpOp == null) {
```
OrangeAndGreen: Question: Should we be nulling out currentOtpOp when we know it's been logged to completion for a given op, i.e. to handle extra callbacks and avoid double-logging later in the UX for the page? Seems like this check will only help if the callback occurs before requestOtp is called. It's not clear to me how any of these stray callbacks could occur, so depending on what points we expect them to happen we may want to improve the safe-guarding

OrangeAndGreen (dismissed):
Left a question and a few nits but generally looks good to me pending the answer to that question.

I also was unsure about all the commenting that was added as it seems like much more than we usually include, but Shubham left comments about the more important ones and otherwise this might be good for a live discussion later (how much commenting do we want? Does it help Claude with future development?)

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
```diff
@@ -97,24 +102,35 @@ private void initOtpManager() {
             @Override
             public void onCodeSent(String verificationId) {
                 if (otpCallback == null) return;
+                emitOtpAnalytics(AnalyticsParamValue.OTP_OUTCOME_SUCCESS, /* reason= */ null);
```
conroy-ricketts: Suggest removing the `/* reason= */` comments throughout this class because it's a bit excessive and is not a pattern followed in our other Java code

conroy-ricketts: "emit" is a strong word here. Maybe `reportOtpAnalytics` is more suitable?

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
```diff
@@ -97,24 +102,35 @@ private void initOtpManager() {
...
                 Toast.makeText(requireContext(), getString(R.string.connect_otp_sent), Toast.LENGTH_SHORT).show();
             }
 
             @Override
             public void onCodeVerified(String code) {
                 if (otpCallback == null) return;
+                // No analytics emission here: this is Firebase's local auto-verification step.
+                // Verify success is recorded from the server-confirmed onSuccess callback so
+                // we don't double-count, and verify failure is recorded from onFailure /
+                // onPersonalIdApiFailure.
```
conroy-ricketts, line 114: The addition of this comment is concerning. It's justifying why some code is _**not**_ there rather than justifying code that is there. This will confuse future readers more than help because it pertains to a very specific context. Suggest removing it

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
conroy-ricketts: This is contradictory to the team's fail-fast approach. Thoughts on either crashing here instead or tackling this from a different approach where we pass in the OTP Op? This function should never be called without an OTP Op

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
conroy-ricketts: It may be helpful to keep the comment but make it more concise (e.g. `Use lastOtpMethod given that personalIdSessionData.getSmsMethod() does not reflect the OTP fallback/auto-switch mechanism.`)

app/src/org/commcare/google/services/analytics/CCAnalyticsParam.java
```diff
@@ -49,4 +49,9 @@ public class CCAnalyticsParam {
     static final String CCC_MESSAGING_CHANNEL_ID = "channel_id";
     static final String CCC_MESSAGING_CONSENT_API_RESULT = "consent_api_success";
     static final String CCC_MESSAGING_DESIRED_CONSENT_STATUS = "consent_api_desired_consent_status";
+
+    // Param keys for OTP analytics (CCCT-2052)
```
conroy-ricketts: Can we remove the ticket number from here as well

app/src/org/commcare/google/services/analytics/FirebaseAnalyticsUtil.java
```diff
@@ -737,9 +737,37 @@ public static void reportNavDrawerItemSelected(String selectedItem) {
         );
     }
 
-    public static void reportOtpRequested(int numberOfAttempts) {
+    /**
+     * Reports the otp_requested event with the outcome and reason of an OTP request
+     * or verify call (CCCT-2052).
```
conroy-ricketts: Should remove the ticket number from here as well

app/src/org/commcare/google/services/analytics/FirebaseAnalyticsUtil.java
conroy-ricketts: Agreed. Seems worth it to at least change the event name here

app/src/org/commcare/utils/OtpAnalyticsMapper.kt
```diff
@@ -0,0 +1,42 @@
+package org.commcare.utils
+
+import org.commcare.connect.network.base.BaseApiHandler.PersonalIdOrConnectApiErrorCodes
+import org.commcare.google.services.analytics.AnalyticsParamValue
+
+/**
+ * Normalizes Firebase {@link OtpErrorType}, {@link PersonalIdOrConnectApiErrorCodes},
+ * and the OtpManager SMS-method string into the stable analytics strings emitted with
+ * the {@code otp_requested} event (CCCT-2052).
```
conroy-ricketts: Flagging the ticket number here too

app/src/org/commcare/utils/OtpAnalyticsMapper.kt
```diff
@@ -0,0 +1,42 @@
+package org.commcare.utils
+
+import org.commcare.connect.network.base.BaseApiHandler.PersonalIdOrConnectApiErrorCodes
+import org.commcare.google.services.analytics.AnalyticsParamValue
+
+/**
+ * Normalizes Firebase {@link OtpErrorType}, {@link PersonalIdOrConnectApiErrorCodes},
```
conroy-ricketts, line 9: What does `@link` and `@code` do in the comments here? It seems that Android Studio is not applying special formatting to these in the pop-up windows

app/src/org/commcare/utils/OtpAnalyticsMapper.kt
conroy-ricketts: > seems like methodFromSmsMethod could be a local helper method and the two calls to reasonFrom could just be local calls to errorType.name?.lowerCase()

Agreed

app/unit-tests/src/org/commcare/utils/OtpAnalyticsMapperTest.kt
conroy-ricketts, line 1: This also ties into the suggestion brought up in `OtpAnalyticsMapper.kt`, but suggest deleting both that object and these tests as there seems to be little benefit

conroy-ricketts (commented):
Looking great so far! Just had a few suggestions

conroy-ricketts:
@Jignesh-dimagi the QA Plan may not be helpful for our QA team. I don't believe they are expected to use Android Studio?

app/src/org/commcare/google/services/analytics/AnalyticsParamValue.java
Jignesh-dimagi: Removed [here](https://github.com/dimagi/commcare-android/pull/3685/changes/e60ca8991ee53c21dc35ee8548a3d08f3b27ebde)

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
Jignesh-dimagi: That makes sense, as it's required to log errors, but Claude is not agreeing and asking to make it uniform like other places. Changed [here](https://github.com/dimagi/commcare-android/pull/3685/changes/e60ca8991ee53c21dc35ee8548a3d08f3b27ebde)

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
Jignesh-dimagi: Changed [here](https://github.com/dimagi/commcare-android/pull/3685/changes/e60ca8991ee53c21dc35ee8548a3d08f3b27ebde)

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
Jignesh-dimagi: Changed [here](https://github.com/dimagi/commcare-android/pull/3685/changes/e60ca8991ee53c21dc35ee8548a3d08f3b27ebde)

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
Jignesh-dimagi: Changed [here](https://github.com/dimagi/commcare-android/pull/3685/changes/e60ca8991ee53c21dc35ee8548a3d08f3b27ebde)

app/src/org/commcare/utils/OtpAnalyticsMapper.kt
Jignesh-dimagi: Changed [here](https://github.com/dimagi/commcare-android/pull/3685/changes/e60ca8991ee53c21dc35ee8548a3d08f3b27ebde)

app/src/org/commcare/google/services/analytics/FirebaseAnalyticsUtil.java
Jignesh-dimagi: I agree that the event name is not fully matching, but as we have limitations of adding total events in Firebase, it was mentioned in the ticket to reuse that.

app/src/org/commcare/utils/OtpAnalyticsMapper.kt
Jignesh-dimagi: I was thinking to keep it separate from Fragment, as it's not a UI-related method. I think eventually, we should move other misc or business logic to the helper/viewmodel from the UI fragment. Your thoughts?

app/src/org/commcare/google/services/analytics/CCAnalyticsParam.java
Jignesh-dimagi: Changed [here](https://github.com/dimagi/commcare-android/pull/3685/changes/e60ca8991ee53c21dc35ee8548a3d08f3b27ebde)

app/src/org/commcare/google/services/analytics/FirebaseAnalyticsUtil.java
Jignesh-dimagi: Changed [here](https://github.com/dimagi/commcare-android/pull/3685/changes/e60ca8991ee53c21dc35ee8548a3d08f3b27ebde)

app/src/org/commcare/utils/OtpAnalyticsMapper.kt
Jignesh-dimagi: Changed [here](https://github.com/dimagi/commcare-android/pull/3685/changes/e60ca8991ee53c21dc35ee8548a3d08f3b27ebde)

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
Jignesh-dimagi: Changed [here](https://github.com/dimagi/commcare-android/pull/3685/changes/e60ca8991ee53c21dc35ee8548a3d08f3b27ebde)

Jignesh-dimagi: Changed [here](https://github.com/dimagi/commcare-android/pull/3685/changes/e60ca8991ee53c21dc35ee8548a3d08f3b27ebde)

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
Jignesh-dimagi: Removed [here](https://github.com/dimagi/commcare-android/pull/3685/changes/1aa759a67fe77808e74937cfe67a0b579b56dd8b)

app/src/org/commcare/utils/OtpAnalyticsMapper.kt
Jignesh-dimagi: Changed [here](https://github.com/dimagi/commcare-android/pull/3685/changes/1aa759a67fe77808e74937cfe67a0b579b56dd8b)

app/unit-tests/src/org/commcare/utils/OtpAnalyticsMapperTest.kt
Jignesh-dimagi: Yeah, I agree. Maybe this is an advantage of using Claude, giving more testable code, so I prefer to keep it but am open to your suggestions.

app/src/org/commcare/utils/OtpAnalyticsMapper.kt
shubham1g5: I do like the mapper class and find it more readable and cleaner for long term code maintenance, is the overhead we are talking about is just that it's in a separate class ?

shubham1g5 (approved):

shubham1g5:
> the QA Plan may not be helpful for our QA team. I don't believe they are expected to use Android Studio?

+1 on this, we do not expect QA to test analytics in general and it's a dev responsibility which should be baked in safety story for the PR. I understand that this is coming from Claude Skill but authorr should still revise the PR after it's been created with the help of agent.

Jignesh-dimagi:
> +1 on this, we do not expect QA to test analytics in general and it's a dev responsibility which should be baked in safety story for the PR. I understand that this is coming from Claude Skill but authorr should still revise the PR after it's been created with the help of agent.

Yeah, I think one more point to be added in create pr skill to not use Android studio as debug tool anytime for QA.

shubham1g5:
@Jignesh-dimagi are you able to address kotlin lint on this

conroy-ricketts (commented):
Don't think there's anything blocking, but flagging that there seems to be a few comments still open above

app/src/org/commcare/utils/OtpAnalyticsMapper.kt
conroy-ricketts: I don't want to speak for Dave here, but at least from my point of view the `reasonFrom()` functions seem to over-complicate the code a bit and favoring simplicity in calling `.name.lowercase()` directly will reduce cognitive load for readers

conroy-ricketts: I personally don't have strong opinions on the location of `methodFromSmsMethod()`

Jignesh-dimagi:
> @Jignesh-dimagi are you able to address kotlin lint on this

It's giving somehow error on file `ConnectJobPreferences` which is not modified in this PR.

Jignesh-dimagi:
> Don't think there's anything blocking, but flagging that there seems to be a few comments still open above

@conroy-ricketts If there are any specific comments which need addressing, please let me know. I replied to or changed the code at all places.

app/src/org/commcare/utils/OtpAnalyticsMapper.kt
OrangeAndGreen: Yeah, it just seems like unnecessary extra code for something that isn't user-facing or expected to grow, the thought being that sometimes it's okay (and easier to read) not to use 50 lines when 5 will do. But I was basing that on the idea that it isn't going to grow, and if we think there's additional code that will eventually live here than I have no complaints.

conroy-ricketts:
> If there are any specific comments which need addressing, please let me know. I replied to or changed the code at all places.

@Jignesh-dimagi That sounds good to me. I don't think anything left is blocking so I'm fine with this being merged in the interest of time (I assume we still want to get this in before the code freeze?)

Side note - the code freeze hasn't happened yet, right? cc @shubham1g5

app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java
Jignesh-dimagi: Yeah, I think it's more about handling and playing defensively here, as we don't want to break anything just in case something is not right while logging it.

OrangeAndGreen (approved):

conroy-ricketts (approved):

PR #3686, opened 2026-05-01
CCCT-2330 Edit Profile Photo From Nav Drawer

conroy-ricketts:
## Product Description

PersonalID users can now update their profile photo directly from the navigation drawer. The drawer image is wrapped in a circular white frame with a small camera icon overlay along the bottom edge to indicate it is tappable. Tapping the image opens an "Update Profile Photo" confirmation dialog (Continue / Cancel, dismissible by tapping outside). Choosing Continue launches the same camera capture flow used during PersonalID signup, and on a successful upload the captured photo replaces the existing image in the drawer and a success toast is shown.

If the user taps Continue while the device is offline, a "no network" toast is shown and the camera is not launched. If the upload fails after the camera flow (e.g. connectivity drops mid-upload), the drawer keeps showing the previous photo, an error toast is shown, and the camera overlay icon switches to a yellow warning triangle.

https://github.com/user-attachments/assets/59a92196-d42a-415b-bfb8-b3a09e521e49

## Technical Summary

- Reuses the existing `MicroImageActivity` capture flow (front camera, face detection, 160px max dimension, 100KB max size).
- Reuses the existing `update_profile` endpoint via `ApiPersonalId.updateUserProfile()`, which gains an optional `photoAsBase64` parameter. `PersonalIdApiHandler.updateProfile()` is a new wrapper that exposes this to drawer callers.
- Routes upload errors through the shared `PersonalIdOrConnectApiErrorHandler.handle()` so non-blocking errors surface as toasts and blocking errors crash, matching the PersonalID signup flow.
- Tracks upload-failure state via an in-memory `lastPhotoUploadFailed` flag on `BaseDrawerController`. The flag is read on every `refreshDrawerContent()` call to drive the camera-vs-warning overlay icon, and is reset by the next successful upload or when a new controller instance is created.
- Defers showing the new photo until after the upload succeeds (no optimistic load + revert-on-failure), so users on slow connections do not see the photo briefly appear and then disappear.
- Adds a Continue-time network availability check using `ConnectivityStatus.isNetworkAvailable`; offline users see the existing `recovery_network_unavailable` toast instead of launching the camera.
- Re-reads the `ConnectUserRecord` from the database inside `refreshDrawerContent()` and `uploadUserPhoto()` rather than caching it at `setupDrawer()` time, so the drawer reflects the currently signed-in user after a sign-out/sign-in.
- Updates `nav_drawer_header.xml` to use a 72dp `MaterialCardView` for the circular image frame, with a translucent overlay container at the bottom containing the camera/warning icon. The previously-unused `manage_profile` `TextView` and the old `image_user_profile` ID are removed; `DrawerViewRefs` now exposes `userImage` and `userImageOverlayIcon`.
- New drawables: `ic_personalid_camera`, `ic_personalid_warning`. New color: `black_60`.
- New strings: `personalid_user_photo_update_dialog_title`, `_message`, `_continue`, `_cancel`, and `personalid_user_photo_update_success` (success toast). The existing `micro_image_activity_title` is updated from "Capture Face Picture" to "Take Profile Photo" (and translations updated to match).
- Architecture is documented in `app/src/org/commcare/activities/connect/personalid_photo_management.md`.

### Safety story

**What gives confidence:**

- I manually exercised the full happy path on a device: opened the drawer, tapped the image, confirmed the dialog, captured a new photo, and verified it appeared in the drawer with the camera overlay icon.
- I manually verified that the confirmation dialog is dismissible by tapping outside of it (in addition to the Cancel button).
- I verified the offline pre-check by tapping Continue with airplane mode on: the "no network" toast appears and the camera is not launched. I also exercised the post-capture failure path by toggling connectivity between camera capture and upload completion; the drawer keeps showing the previous photo, an error toast appears, and the warning triangle replaces the camera icon.
- The change reuses the already-shipped `MicroImageActivity` capture pipeline and `update_profile` HTTP endpoint, so the camera capture and network layer are not new code paths.
- Error handling delegates to the shared `PersonalIdOrConnectApiErrorHandler.handle()` used elsewhere in the PersonalID flows, so the blocking/non-blocking error categorization matches existing behavior.

**Risks to review:**

- I did not run unit tests locally, so any regressions in `BaseDrawerController`, `ApiPersonalId.updateUserProfile()` callers, or `PersonalIdApiHandler` would only be caught by CI on this PR.
- `ApiPersonalId.updateUserProfile()` gained a new `photoAsBase64` positional parameter. Existing callers needed to be updated; a missed caller would fail to compile rather than misbehave at runtime, but it is worth a reviewer confirming all signup-flow callers were updated correctly.
- The `lastPhotoUploadFailed` flag now lives on the `BaseDrawerController` instance, so the warning overlay disappears whenever a new controller is created (e.g. activity recreation, navigation to another screen with a drawer). This is a deliberate trade-off vs. the prior persisted-flag design and is worth a reviewer's eye.
- No automated coverage was added for the new drawer code paths (dialog wiring, photo result handler, upload success/failure branches, overlay icon state).
- The blocking-error path re-throws as a `RuntimeException` to match the signup flow. The drawer is a more general-purpose surface than signup, so a server-side blocking error (e.g. file too large) will now crash the app from anywhere the drawer is reachable. This is intentional but worth a reviewer's eye.
- `MicroImageActivity`'s title string was renamed from "Capture Face Picture" to "Take Profile Photo" globally (including all locale files). Any other entry point into `MicroImageActivity` will see the new title.

### Automated test coverage

No new automated tests are added in this PR. Existing tests cover unrelated drawer functionality and the underlying `MicroImageActivity` and API layers, but do not exercise the new dialog wiring, upload success/failure branches, or overlay-icon state in `BaseDrawerController`. Regressions in those new paths would not be caught by automation today.

### QA Plan

QA notes for this change have been added to the QA Notes section of [RELEASES.md](../RELEASES.md) under CommCare 2.64. The 2.64 section also has a "Profile Photo Update" entry under What's New.

[CCCT-2330]: https://dimagi.atlassian.net/browse/CCCT-2330?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/org/commcare/activities/connect/personalid_photo_management.md
```diff
@@ -0,0 +1,78 @@
...
+9. **On blocking failure** (e.g. file too large): the app crashes via
+   `PersonalIdOrConnectApiErrorHandler.handle()` to mirror the signup flow's behavior.
+
+## Components
+
+* **`BaseDrawerController`** (`navdrawer/BaseDrawerController.kt`)
+  * Wires the avatar tap → confirmation dialog → camera launch.
+  * Reads `PersonalIDUserPreferences.didLastPhotoUploadFail()` on every drawer refresh and
+    swaps `avatar_overlay_icon` between the camera and warning drawables accordingly.
+  * Calls `openDrawer()` after a successful capture to keep the sidebar visible while the
```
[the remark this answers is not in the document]
conroy-ricketts: 46bbc734317c103c1b94b499bd8157a22a777bca

app/src/org/commcare/navdrawer/BaseDrawerController.kt
```diff
@@ -144,24 +163,47 @@ class BaseDrawerController(
...
+        dialog.makeCancelable()
+        dialog.showNonPersistentDialog(activity)
     }
 
     fun refreshDrawerContent() {
         if (PersonalIdManager.getInstance().isloggedIn()) {
             setSignedInState(true)
             binding.ivNotification.setImageResource(getNotificationIcon(activity))
+            binding.userName.text = user!!.name
+            loadUserPhoto(user!!.photo!!)
```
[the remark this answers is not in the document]
conroy-ricketts: 1b7f8c8c44c3b43ed26e51b0ff92fce98c4418a3

app/res/values/strings.xml
```diff
@@ -543,7 +543,12 @@
         <item>save</item>
     </string-array>
 
-    <string name="micro_image_activity_title">Capture Face Picture</string>
+    <string name="micro_image_activity_title">Take Profile Photo</string>
+
+    <string name="nav_drawer_user_photo_update_dialog_title">Update Profile Photo</string>
+    <string name="nav_drawer_user_photo_update_dialog_message">Would you like to take a new profile photo?</string>
+    <string name="nav_drawer_user_photo_update_dialog_continue">Continue</string>
+    <string name="nav_drawer_user_photo_update_dialog_cancel">Cancel</string>
```
shubham1g5: nit: replace `nav_drawer_*` -> `personid_*`

app/src/org/commcare/activities/connect/personalid_photo_management.md
```diff
@@ -0,0 +1,76 @@
+# PersonalID Photo Management
```
shubham1g5, line 1: Are you intending for this to live in repo ?  

Also thoughts on moving it to a new  `org.comcare.personalId.docs` package.

app/src/org/commcare/CommCareApplication.java
```diff
@@ -278,6 +279,8 @@ public void onCreate() {
         ProcessLifecycleOwner.get().getLifecycle().addObserver(this);
         NotificationsSyncWorkerManager.schedulePeriodicPushNotificationRetrievalChecked(
                 CommCareApplication.instance());
+
+        PersonalIDUserPreferences.Companion.resetLastPhotoUploadFailed();
```
shubham1g5: We should not need a preference for something that we don't need to persist across application lifecycle, you can instead have a static variable attached to a class that's independent of activity lifecyle to track things like these.

app/src/org/commcare/navdrawer/BaseDrawerController.kt
```diff
@@ -291,4 +333,75 @@ class BaseDrawerController(
...
+                ConnectUserDatabaseUtil.storeUser(activity, user)
+                PersonalIDUserPreferences.setLastPhotoUploadFailed(false)
+                binding.userImageOverlayIcon.setImageResource(R.drawable.ic_personalid_camera)
+            }
+
+            override fun onFailure(
+                errorCode: PersonalIdOrConnectApiErrorCodes,
+                t: Throwable?,
+            ) {
+                revertUserPhoto()
```
shubham1g5: Optional: It might be easier or simpler if we only call `loadUserPhoto` after the upload has been successful rather than trying to revert on failures. Also on slower connections, it might seem weird if the taken photo appears for a little bit only to disappear after 5 seconds.

shubham1g5 (changes requested):
Looks quite good, seems like only 1 blocking comment around use of preferences.

app/src/org/commcare/navdrawer/BaseDrawerController.kt
```diff
@@ -144,24 +163,47 @@ class BaseDrawerController(
...
+
+        dialog.makeCancelable()
+        dialog.showNonPersistentDialog(activity)
     }
 
     fun refreshDrawerContent() {
         if (PersonalIdManager.getInstance().isloggedIn()) {
             setSignedInState(true)
             binding.ivNotification.setImageResource(getNotificationIcon(activity))
+            binding.userName.text = user!!.name
```
OrangeAndGreen: I'm seeing some sort of issue just after recovering an account, and I think the cause is that we aren't refreshing the "user" object from the DB here anymore.

First I tried recovering my account and hit a crash here just after completing recovery because "user" was null:
java.lang.NullPointerException
at org.commcare.navdrawer.BaseDrawerController.refreshDrawerContent(BaseDrawerController.kt:197)
at org.commcare.navdrawer.BaseDrawerController$setupActionBarDrawerToggle$1.onDrawerSlide(BaseDrawerController.kt:103)
at androidx.drawerlayout.widget.DrawerLayout.dispatchOnDrawerSlide(DrawerLayout.java:962)
at androidx.drawerlayout.widget.DrawerLayout.setDrawerViewOffset(DrawerLayout.java:974)
at androidx.drawerlayout.widget.DrawerLayout$ViewDragCallback.onViewPositionChanged(DrawerLayout.java:2264)
at androidx.customview.widget.ViewDragHelper.continueSettling(ViewDragHelper.java:779)
at androidx.drawerlayout.widget.DrawerLayout.computeScroll(DrawerLayout.java:1367)

Then I forgot my account and tried again with a demo number, and when I completed recovery (no crash this time) the side nav display showed the info for the previously logged in user (the one I'd forgotten). I then killed and restarted the app and the correct user info was displayed.

app/src/org/commcare/navdrawer/BaseDrawerController.kt
OrangeAndGreen: Agreed here, seems better not to swap the image until the update is complete rather than risk swapping it and then having to swap back a few seconds later.

RELEASES.md
```diff
@@ -55,7 +43,22 @@ we would like to communicate to QA as part of the release testing
   - Text cursor functionality (i.e. backspacing, clicking an earlier box to jump back)
 
 - Test the new offline status indicator at the top of refreshable Connect pages (Connect Home, Learning Progress, Delivery Progress). Verify that the error message appears when entering these pages while offline, and that it disappears once the device comes back online.
-- Verify that the combobox widget is working as expected when selecting an item that is used to filter another combobox widget and also determines the visibility of some other unrelated question whose relevance condition depends on the selection.
+
+- **PersonalID profile photo update from nav drawer:**
```
OrangeAndGreen, line 35: A couple notes here:
1. I think we need to start a section for v2.64 since this work is going in the next release (and move the new text here to the QA Notes for 2.64)
2. Also add a Release Note about the new functionality.

OrangeAndGreen (changes requested):
One bug to solve and a couple other comments to consider, but generally looks good. This will be a big milestone when I finally get to add a photo to the account linked to my real phone number (created before profile photos were a thing)! :-)

app/src/org/commcare/activities/connect/personalid_photo_management.md
conroy-ricketts: Yeah having this here was intentional. Let's move this conversation over to the thread Dave started [here](https://dimagi.slack.com/archives/C080JGPUGG1/p1777998533835299?thread_ts=1777996692.365609&cid=C080JGPUGG1). I'll make changes to this PR based on the team's consensus.

app/src/org/commcare/CommCareApplication.java
conroy-ricketts: Great catch! I'm also realizing that we do not have to "reset" this variable here and just default it to false in the `BaseDrawerController` to achieve the same effect

app/src/org/commcare/navdrawer/BaseDrawerController.kt
conroy-ricketts: Agreed!

app/src/org/commcare/navdrawer/BaseDrawerController.kt
conroy-ricketts: Great catch!! I did not mean to remove the user refresh here and this was an oversight by me

app/res/values/strings.xml
conroy-ricketts: 83bd47d8dbb50eeb8ae84e5d4216a4d95d6811f0

app/src/org/commcare/CommCareApplication.java
conroy-ricketts: eb121cb59f7d2d64061f572472eaf745e9a9f906

app/src/org/commcare/navdrawer/BaseDrawerController.kt
conroy-ricketts: 95fad6e4516e246da7f3579ac109d188f9c95ec5

app/src/org/commcare/navdrawer/BaseDrawerController.kt
conroy-ricketts: ac6703cfdc119827609daf534a14d7b3be90a4fe

RELEASES.md
conroy-ricketts: ac77f66510c17b094709beab8d61cbc66af7954e

shubham1g5 (dismissed):

app/src/org/commcare/navdrawer/BaseDrawerController.kt
```diff
@@ -291,4 +338,71 @@ class BaseDrawerController(
...
+                RequestOptions()
+                    .placeholder(R.drawable.nav_drawer_person_avatar)
+                    .error(R.drawable.nav_drawer_person_avatar),
+            ).into(binding.userImage)
+    }
+
+    companion object {
+        private const val USER_PHOTO_MAX_DIMENSION_PX = 160
+        private const val USER_PHOTO_MAX_SIZE_BYTES = 100 * 1024 // 100 KB
+    }
```
Jignesh-dimagi, line 409: Nit: I think it is better to define these in MicroImageActivity, as the same is repeated [here](https://github.com/dimagi/commcare-android/blob/3daf86162b5317e11da7cf1f7274a1eed3e4fdb7/app/src/org/commcare/fragments/personalId/PersonalIdPhotoCaptureFragment.java#L43) also.

app/src/org/commcare/navdrawer/BaseDrawerController.kt
```diff
@@ -291,4 +338,71 @@ class BaseDrawerController(
...
+                }
+            }
+    }
+
+    private fun launchCameraForPhotoEdit() {
+        val intent = Intent(activity, MicroImageActivity::class.java).apply {
+            putExtra(MicroImageActivity.MICRO_IMAGE_MAX_DIMENSION_PX_EXTRA, USER_PHOTO_MAX_DIMENSION_PX)
+            putExtra(MicroImageActivity.MICRO_IMAGE_MAX_SIZE_BYTES_EXTRA, USER_PHOTO_MAX_SIZE_BYTES)
+        }
+        takePhotoLauncher.launch(intent)
```
Jignesh-dimagi, line 367: Maybe not related to this PR, but the back button on this image capture activity is not working.

Jignesh-dimagi:
No blocking comments, but small things are mentioned.

Jignesh-dimagi (dismissed):

app/src/org/commcare/navdrawer/BaseDrawerController.kt
conroy-ricketts: I actually thought about doing this, but I saw that `MicroImageActivity` defines defaults for these two values around [here](https://github.com/dimagi/commcare-android/blob/3daf86162b5317e11da7cf1f7274a1eed3e4fdb7/app/src/org/commcare/fragments/MicroImageActivity.java#L65).

So I figured that the original intention was for the caller (i.e. wherever the `MicroImageActivity` is launched from) to pass in custom values and `MicroImageActivity` has shared defaults to fall back to. I feel like moving the shared constants here might defeat the purpose. I know that this is a nit, but what are your thoughts on that?

@OrangeAndGreen @shubham1g5 do you guys know what the original intention was with the [default values](https://github.com/dimagi/commcare-android/blob/3daf86162b5317e11da7cf1f7274a1eed3e4fdb7/app/src/org/commcare/fragments/MicroImageActivity.java#L65) in `MicroImageActivity` ?

conroy-ricketts: Well I guess the better question is why are these intent extras in the first place and not local variables?

app/src/org/commcare/navdrawer/BaseDrawerController.kt
conroy-ricketts: That's a good catch! I'll log a new ticket for this to avoid potentially increasing the scope here (I was able to reproduce for normal PersonalID signup too)

app/src/org/commcare/navdrawer/BaseDrawerController.kt
OrangeAndGreen: I don't have much context there but I think @avazirna may have written that code or at least been the one to most recently work on it.

OrangeAndGreen (dismissed):

app/src/org/commcare/navdrawer/BaseDrawerController.kt
conroy-ricketts: https://dimagi.atlassian.net/browse/CI-650

app/src/org/commcare/navdrawer/BaseDrawerController.kt
shubham1g5: The idea here is definitely to keep this activity verstatile and configurable and easy to be plugged in for different features, there is a very strong use case for this activity to be called out from a CommCare callout question type for example to allow someone to take user pictures inside a CC form.

app/src/org/commcare/activities/connect/personalid_photo_management.md
conroy-ricketts: Resolving this given that I'll make a new PR to address this very soon

conroy-ricketts:
@dimagi/connect-mobile-devs I just pushed another commit for linting changes

shubham1g5 (approved):

PR #3692, opened 2026-05-06
Bump androidx.appcompat to 1.7.0 

avazirna:
## Technical Summary

The motivation behind this PR is the Firebase BOM migration, which transitively upgrades AppCompat to version 1.7.0. This upgrade affects one of the unit tests, `DemoUserRestoreTest`, so I decided to open a separate PR to bump AppCompat to version 1.7.0 and address the issue. See the details below.

- `DemoUserRestoreTest` started failing with `IllegalStateException: Iteration already started` from `ViewTreeObserver$CopyOnWriteArray`. 
  - **Root cause:**
    - The home grid's `OnGlobalLayoutListener` (`StandardHomeActivityUIController.setupGridView`) calls `requestLayout()` from inside the global-layout dispatch. Under the test's `@LooperMode(LEGACY)`, Robolectric runs the resulting traversal synchronously, re-entering `dispatchOnGlobalLayout` while the outer iteration is still active.
    -  This was latent on appcompat 1.2: the grid listener was the only one registered, and since it removes itself first, the nested dispatch saw an empty listener list and short-circuited.
    - Starting with appcompat 1.4.0, `ToolbarWidgetWrapper` (created by `setSupportActionBar()`) calls `ViewCompat.setAccessibilityPaneTitle()` on the toolbar. On sdk < 28 (our Robolectric config pins `sdk=23`), androidx.core implements this by registering a *persistent* `OnGlobalLayoutListener` on the same window ViewTreeObserver. The listener list is now non-empty during the re-entrant dispatch, so the non-reentrant iteration guard trips.
  - **Fix:** avoid calling `visible()` when launching `StandardHomeActivity` in the test (use `create().start().postCreate(null).resume()` instead of `setup()`), so no layout traversal runs. This matches every other test in the suite that launches `StandardHomeActivity` (`RecoveryMeasuresTest`, `ExternalLaunchTests`, `PersonalIdDrawerVisibilityTest`, `FormRecordListActivityTest` all stop at `create()`/`resume()`). All assertions, including options-menu visibility, still pass. A comment in the test explains why `visible()` is deliberately omitted.

Not observable on devices: `Choreographer` always defers the traversal to the next frame there; the synchronous re-entry is specific to Robolectric's deprecated LEGACY looper mode.

### Safety story
- No production code change; the only functional edit is unit-test-only which is now running successfully.

## Labels and Review


Jignesh-dimagi (approved):

app/unit-tests/src/org/commcare/android/tests/DemoUserRestoreTest.java
```diff
@@ -73,9 +73,13 @@ private static ShadowActivity launchHomeActivityForDemoUser() {
         Intent homeActivityIntent =
                 new Intent(ApplicationProvider.getApplicationContext(), StandardHomeActivity.class);
         homeActivityIntent.putExtra(DispatchActivity.START_FROM_LOGIN, true);
+        // Deliberately not calling visible(): it attaches the view hierarchy and runs a
+        // layout traversal, and the home grid's OnGlobalLayoutListener calls requestLayout()
+        // from within the dispatch. Under LEGACY looper mode that traversal runs re-entrantly
+        // and crashes with "IllegalStateException: Iteration already started".
```
shubham1g5, line 79: nit: comment is confusing here with reference to `visible()` as think there is no such method on `Robolectric.buildActivity`, also think we can simply remove the whole comment given the other pattern is a standard pattern we follow in other tests.

shubham1g5 (approved):

app/unit-tests/src/org/commcare/android/tests/DemoUserRestoreTest.java
avazirna: It's part of `ActivityController.setup()`. This method initializes the controller by calling the method chain `create().start().postCreate(null).resume().visible()`. In this case, the method was replaced with its individual calls, excluding `visible()`.

PR #3698, opened 2026-05-07
CCCT-2332 Migrate Android Startup Strings to strings.xml

conroy-ricketts:
## Product Description

The strings that appear during CommCare's pre-app-sandbox flows (initial install, App Manager, Install From List, etc.) now resolve through the standard Android resource framework (`R.string`) instead of through the JavaRosa `Localization` framework backed by `app/assets/locales/android_startup_strings.txt`. Visible behavior should be unchanged for end users on supported languages, with translations added for Spanish, French, Hausa, Hindi, Lithuanian, Norwegian, Portuguese, Swahili, and Tigrinya.

## Technical Summary

- These strings come out of the app sandbox and are not exported to CC app translations, so there is no longer a reason to maintain them in a separate `*.txt` asset file. Migrating them to `strings.xml` consolidates them with the rest of the platform / Connect strings and lets us drop the separate Localization registration.
- 49 new entries added to `app/res/values/strings.xml`. Per `app/docs/translations.md`, the 3 update-flow error strings (`update_version_mismatch`, `update_major_mismatch`, `update_minor_mismatch`) are marked `cc:translatable="true"` because they can fire while a user is logged in to an app (i.e. inside the CC HQ app sandbox); the remaining 46 entries are not marked translatable because they only fire pre-login. 2 entries (`install_barcode_top`, `install_barcode_bottom`) already existed in `strings.xml` and were left in place.
- Translations added for all 9 supported non-English language locales: `values-es`, `values-fr`, `values-ha`, `values-hi`, `values-lt`, `values-no`, `values-pt`, `values-sw`, `values-ti`. Translations are machine-quality and would benefit from native-speaker review for less common locales (Tigrinya, Hausa, Lithuanian, Norwegian).
- All `Localization.get("install.…")` / `"menu.…"` / `"updates.resources.…"` / `"archive.install.…"` / etc. callers for the migrated keys were converted to `getString(R.string.<underscored_key>)`. Format-arg keys use Android positional args (`%1$s`, `%2$s`). The 3 update-flow error lookups in `UpdateHelper` use `StringUtils.getStringRobust(...)` so HQ-supplied translations override at runtime when the user is logged in.
- Three indirection patterns were refactored so they no longer go through `Localization.get`:
    - **Dynamic key concat** in `InstallFromListActivity` (`"get.app.list.user.error." + (mode)`) was replaced with a conditional `R.string` lookup.
    - **Preference `keyToTitleMap` entry** for `app.manager.advanced.developer.options` in `AppManagerAdvancedPreferences` was removed; the developer-options preference title is now declared in `app/res/xml/app_manager_preferences.xml` via `android:title="@string/app_manager_advanced_developer_options"`.
    - **`@UiElement(locale = …)`** annotations on the archive prompt / button views in `InstallArchiveActivity` had their `locale` attribute removed; the install-state default text is now declared in the layout XML (`screen_multimedia_inflater.xml`) via `android:text=…`. The existing `if (fromUpdateActivity)` block continues to override these for the update-launch case.
- `CommCareApplication.initializeDefaultLocalizerData()` no longer registers `android_startup_strings.txt` as a JavaRosa locale reference, and the asset file itself was deleted.

### Safety story

**What gives confidence:**
- I built the debug APK locally (`./gradlew :app:assembleCommcareDebug`) — BUILD SUCCESSFUL.
- Unit tests for `org.commcare.{activities,fragments,preferences,update,engine}.*` packages passed.
- I ran the build on a device/emulator and walked the migrated screens end-to-end: welcome screen, manual URL install, `.ccz` install, App Manager → Advanced Settings → Developer Options row, and Install From List error states (mobile vs web). Strings render correctly.
- The migration is mostly mechanical 1:1 swaps from `Localization.get("…")` to `getString(R.string.…)`, with `grep` verification that zero migrated dotted keys remain in `*.java` / `*.kt` and zero references to `android_startup_strings` remain in source.

**Risks to review:**
- Translations for Hausa, Hindi, Lithuanian, Norwegian, Swahili, and Tigrinya were produced via machine translation and have not been reviewed by native speakers. Spanish, French, and Portuguese should be reasonably accurate but a quick pass is still worthwhile. If any translation is awkward or wrong, it ships as-is until someone catches it.
- I did not exhaustively walk every locale. The non-English translations should ideally be verified by native speakers.
- The `keyToTitleMap` framework in `CommCarePreferenceFragment.setupLocalizedText()` is still used by other preferences (e.g. `clear-user-data`, `data-change-logs`, `enable-mobile-privilege`). Those still go through `Localization.get(...)` and are unaffected by this change.
- `@UiElement(locale = …)` is still used elsewhere in the codebase (e.g. `PinAuthenticationActivity`, `ConnectionDiagnosticActivity`, `KeyAccessRequestActivity`). Only the two annotations in `InstallArchiveActivity` were touched.
- The migration removes the dotted Localization keys for these strings from the runtime. Any third-party code or undiscovered call site that was using `Localization.get("install.…")` for one of these keys will now throw `NoLocalizedTextException` instead of returning the string. I greped the project source and found none, but the risk is non-zero for code that constructs keys dynamically without a literal that grep would match.
- The 3 newly-`cc:translatable` `update_*_mismatch` strings will be picked up by the next translation export run (mobile-deploy `update_translations.py`); until then, runtime falls back to the device-language entries in `values-*` via `StringUtils.getStringRobust`'s `c.getString(...)` fallback path.

### Automated test coverage

No new automated tests were added — this is a refactor of how existing strings are looked up, not a behavior change. Existing unit tests for the touched packages (`org.commcare.{activities,fragments,preferences,update,engine}.*`) continue to pass. Resource-resolution failures would surface at build time (`AAPT` errors) or as a `Resources.NotFoundException` at runtime.

### QA Plan

QA notes for this change have been added to the QA Notes section of [RELEASES.md](../RELEASES.md) under the CommCare 2.64 release.

[CCCT-2332]: https://dimagi.atlassian.net/browse/CCCT-2332?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


conroy-ricketts:
@coderabbitai full review

app/res/values-fr/strings.xml
```diff
@@ -535,6 +535,56 @@ License.
...
+    <string name="install_manual">Veuillez saisir l\'URL</string>
+    <string name="install_ready_top">Presque terminé !</string>
+    <string name="install_ready_bottom">Appuyez sur le bouton ci-dessous pour installer votre application.</string>
+    <string name="install_netwarn">Si vous tentez d\'installer depuis un serveur, assurez-vous d\'avoir une connexion Internet.</string>
+    <string name="install_button_enter">Saisir le code</string>
+    <string name="install_button_startover">Recommencer</string>
+    <string name="install_button_retry">Réessayer l\'installation</string>
+    <string name="install_button_start">Démarrer l\'installation</string>
+    <string name="install_bad_ref">Vous n\'avez pas scanné une URL valide. Veuillez réessayer.</string>
+    <string name="install_bad_login">Se connecter à une application existante</string>
```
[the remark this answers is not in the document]
conroy-ricketts: (Claude AI)

Will skip. The English source — both the migrated text and the resource entry at app/res/values/strings.xml:113 — reads Login to Existing App, a neutral CTA. All other locales          
  translated it the same way (es: "Iniciar Sesión en Aplicación Existente", pt: "Entrar em Aplicativo Existente"). The reviewer's interpretation is keyed off the resource name containing "bad", not the actual displayed text.   
  Changing French to imply an error would diverge from the English source and the eight peer translations. The key has no code references either, so there's no UX context contradicting the literal English. Reverting would      
  require changing the English first.

RELEASES.md
```diff
@@ -2,6 +2,41 @@
...
+#### What's New
+
+#### Important Bug Fixes
+
+#### Internal Release Notes
+
+<!--
+Release notes that are not applicable for wider CommCare users but only for a specific projects. 
+These notes are only published internally in [CommCare Change log wiki](https://dimagi.atlassian.net/wiki/spaces/internal/pages/2145058874/CommCare+Mobile+Changelog)
+along with the public release notes above
```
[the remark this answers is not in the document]
conroy-ricketts: 2435329ae6a96d4d970b092a818f77cb83447323

app/src/org/commcare/update/UpdateHelper.java
```diff
@@ -101,12 +100,13 @@ public ResultAndError<AppInstallStatus> update(String profileRef, ResourceInstal
...
+                Context appContext = CommCareApplication.instance();
+                error = appContext.getString(R.string.update_version_mismatch, e.getRequiredVersionString(), e.getAvailableVesionString());
                 error += " ";
                 if (e.getRequirementType() == UnfullfilledRequirementsException.RequirementType.MAJOR_APP_VERSION) {
-                    error += Localization.get("update.major.mismatch");
+                    error += appContext.getString(R.string.update_major_mismatch);
                 } else if (e.getRequirementType() == UnfullfilledRequirementsException.RequirementType.MINOR_APP_VERSION) {
-                    error += Localization.get("update.minor.mismatch");
+                    error += appContext.getString(R.string.update_minor_mismatch);
                 }
```
shubham1g5: think for these 2 strings, it makes more sense to make them `translatable = true` and use `getStringRobust` here (as these errors can happen while a user is logged in into an app)

app/src/org/commcare/activities/InstallArchiveActivity.java
```diff
@@ -67,6 +67,9 @@ protected void onCreate(Bundle savedInstanceState) {
         if (fromUpdateActivity) {
             txtDisplayPrompt.setText(Localization.get("archive.update.prompt"));
             btnInstallArchive.setText(Localization.get("archive.update.button"));
+        } else {
+            txtDisplayPrompt.setText(R.string.archive_install_prompt);
+            btnInstallArchive.setText(R.string.archive_install_button);
```
Jignesh-dimagi: @conroy-ricketts Why is the 'else' condition required here?

app/src/org/commcare/preferences/AppManagerAdvancedPreferences.java
```diff
@@ -78,6 +76,7 @@ protected void setupPrefClickListeners() {
         });
 
         Preference devOptionsButton = findPreference(DEVELOPER_OPTIONS);
+        devOptionsButton.setTitle(requireContext().getString(R.string.app_manager_advanced_developer_options));
```
Jignesh-dimagi: @conroy-ricketts This is required? as it was not present previously.

Jignesh-dimagi (changes requested):

OrangeAndGreen (dismissed):
Looks good on my end, nice cleanup

app/src/org/commcare/activities/InstallArchiveActivity.java
avazirna: +1

avazirna (dismissed):

app/src/org/commcare/update/UpdateHelper.java
conroy-ricketts: Fixed in e064632b5. Marked `update_version_mismatch`, `update_major_mismatch`, and `update_minor_mismatch` as `cc:translatable="true"` per `app/docs/translations.md`, and switched the lookups in `UpdateHelper` to `StringUtils.getStringRobust(...)`. Extended to `update_version_mismatch` for consistency since it's part of the same error message and the same "can fire while logged in" reasoning applies to it.

app/src/org/commcare/activities/InstallArchiveActivity.java
conroy-ricketts: Fixed in e064632b5. The `else` was needed because removing `locale=` from the `@UiElement` annotation also removed the framework path that previously set the install-state default text — without an `else`, the prompt and button would have been blank in the (default) install case. Cleaner solution: I moved the default text into the layout XML via `android:text="@string/archive_install_prompt"` / `android:text="@string/archive_install_button"`. The layout now provides the install-state defaults, and the existing `if (fromUpdateActivity)` block overrides them only when launched from the update flow, so the `else` branch is no longer required.

app/src/org/commcare/preferences/AppManagerAdvancedPreferences.java
conroy-ricketts: Fixed in e064632b5. The line was needed because previously the title came from the framework's `keyToTitleMap` mechanism (`CommCarePreferenceFragment.setupLocalizedText` reads the map and calls `Localization.get(value)` → `setTitle(...)`); removing the map entry as part of this migration meant the title had to be set somewhere else. Cleaner solution: I moved `android:title="@string/app_manager_advanced_developer_options"` directly into `app/res/xml/app_manager_preferences.xml` and dropped the runtime `setTitle` call. The other two entries (`clear-user-data`, `data-change-logs`) still use `keyToTitleMap` since their titles are translatable strings handled by the existing framework — they're out of scope for this migration.

conroy-ricketts:
@Jignesh-dimagi @shubham1g5 

I decided to experiment with the `dev-utils:resolve-pr-comments` skill in Claude Code to address your comments 👀 (the commit and comments were entirely generated by AI)

Let me know how it did!

OrangeAndGreen (approved):

shubham1g5 (approved):

Jignesh-dimagi (approved):

PR #3700, opened 2026-05-08
CCCT-2354 Add Back Online Indicator

OrangeAndGreen:
## Product Description
When a sync succeeds after a previous offline failure, the network status bar now displays a green "Back Online" indicator with a WiFi icon alongside the standard "Last synced" message. The regular sync-success bar background was also changed from sky blue to green to align with the new indicator design.

Demo:

https://github.com/user-attachments/assets/da0651c7-511b-4ed4-9a06-d960d996fabd

## Technical Summary
- `NetworkStatusBarViewController` now manages left-icon visibility and dynamically swaps the right-side icon/label so the single status bar layout can render error, offline, success, and back-online states. Adds `showBackOnline(message)` (green background, `ic_online` drawable, "Back Online" label) and switches `showMessage` to use `connect_green` (#16A085).
- `BaseConnectFragment` tracks a `wasOffline` flag set by `showOfflineIndicator()`. On the next `DataState.Success`, if the flag is set, the Back Online bar is shown instead of the regular sync-success bar; the flag is then cleared.
- Adds `connect_back_online` string with translations for all locales that already translate `connect_offline` (es, fr, ha, hi, lt, no, pt, sw, ti). The translations were AI-generated and should be verified by the translation team.
- Reuses the existing `connect_green` color constant; no new color tokens introduced.

### Safety story
- The change is additive — the new code path only triggers when the previous data state observed by `BaseConnectFragment` was an offline-network error. Existing error and offline flows are unchanged.
- No persistent storage or migration impact.
- Translations should be validated by the translation team before release; English fallback applies for any unlisted locale.

### Automated test coverage
No new automated tests; existing test coverage for `BaseConnectFragment` continues to exercise the `DataState` observer.

[CCCT-2354]: https://dimagi.atlassian.net/browse/CCCT-2354?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/res/drawable/ic_online.xml
conroy-ricketts, line 1: Out of curiosity, where did this icon originate from? This isn't from a Figma mock right?

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
```diff
@@ -210,9 +215,14 @@ abstract class BaseConnectFragment<B : ViewBinding> :
         mNetworkStatusBarViewController!!.showMessage(getString(R.string.connect_sync_successful))
     }
 
+    private fun showBackOnline() {
+        mNetworkStatusBarViewController!!.showBackOnline(getString(R.string.connect_sync_successful))
+    }
```
conroy-ricketts, line 220: Thoughts on doing a small refactor here?

Now realizing that we have 4 different functions showing the bar in `NetworkStatusBarViewController`: 
- `showError()`
- `showMessage()`
- `showOfflineStatus()`
- `showBackOnline()`

Not only is each function only called once, but it seems that both `showMessage()` and `showBackOnline()` are doing the same thing but +/- a couple of parameters for the right icon/label. Also, both `showError()` and `showOfflineStatus()` are very similar.

What if we delete these four functions and simply call `showBar()` directly, or at the very least combine the error-related functions into one and the "happy" functions into one? The current functions in `NetworkStatusBarViewController` seem to be an unnecessary abstraction, especially because the callers inside `BaseConnectFragment` already abstract this logic further.

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
```diff
@@ -210,9 +215,14 @@ abstract class BaseConnectFragment<B : ViewBinding> :
...
+
     private fun shouldMonitorNetwork(): Boolean = this is RefreshableFragment
 
     private fun showOfflineIndicator() {
+        wasOffline = true
```
conroy-ricketts, line 225: Nit: can we move this one-level up to the caller function where we do the check for the network connectivity

app/src/org/commcare/views/NetworkStatusBarViewController.kt
```diff
@@ -17,29 +19,67 @@ class NetworkStatusBarViewController(
...
+
+        val rightVisibility = if (rightIcon != null && rightLabel != null) View.VISIBLE else View.GONE
+        statusBarBinding.ivOffline.visibility = rightVisibility
+        statusBarBinding.tvOfflineLabel.visibility = rightVisibility
+        if (rightIcon != null) {
+            statusBarBinding.ivOffline.setImageResource(rightIcon)
+        }
+        if (rightLabel != null) {
+            statusBarBinding.tvOfflineLabel.setText(rightLabel)
+        }
```
conroy-ricketts: Nit (for readability): it seems like we are taking an all-or-nothing approach with showing the right icon and label here. Can we have just one big if statement and put all of this in there?

```
if (rightIcon != null && rightLabel != null) {

} else {

}
```

conroy-ricketts (commented):
Looks great! I don't have anything blocking

app/res/drawable/ic_online.xml
OrangeAndGreen: Yeah, first it was just a screenshot that someone put together that made it into a Figma [here](https://www.figma.com/design/MVp8U7IQuXFK6S4rtxcS4h/Inline-Error---Data-Refresh-Indicator?node-id=179-53&t=5XTjxYu68Dqlc29z-0), but then Brendon was nice enough to [formalize the design](https://www.figma.com/design/MVp8U7IQuXFK6S4rtxcS4h/Inline-Error---Data-Refresh-Indicator?node-id=546-36&t=piTxOpHhEWpNhqPf-4) and provide me with a proper asset. The second part is also when we decided to change the bar to green for success.

conroy-ricketts: Ah cool, also confirming that the copyright comment here was intentional?

shubham1g5 (dismissed):

Jignesh-dimagi (dismissed):

app/res/drawable/ic_online.xml
OrangeAndGreen: It was not! [dffdeb9](https://github.com/dimagi/commcare-android/pull/3700/commits/dffdeb9eacd6ab381a8ae18b3f10bc505bb2578f)

app/src/org/commcare/views/NetworkStatusBarViewController.kt
OrangeAndGreen: [c8ac2ec](https://github.com/dimagi/commcare-android/pull/3700/commits/c8ac2ec7c1ddf08512656eedeb74430c679622bf)

conroy-ricketts (dismissed):
Looking great, I have nothing blocking. Flagging there are a couple of open comments above, feel free to resolve if not addressing

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
OrangeAndGreen: Decided to leave this one as is, the thinking being that there should never be a call to showOfflineIndicator that doesn't also updated the variable (i.e. in case of future reuse)

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
OrangeAndGreen: I decided to leave this for now since there's more work coming related to this functionality as we implement additional pages and it seems like something we can easily clean up later when all the work is complete. In the meantime it seems likely enough that we may need to make some small additional changes in the near future, so feels like it's better to save this cleanup for the end.

conroy-ricketts (dismissed):

shubham1g5 (dismissed):

shubham1g5 (approved):

conroy-ricketts (approved):

PR #3702, opened 2026-05-11
CCCT-2078 Extract Personal ID Unlock Logic Into PersonalIdUnlocker

shubham1g5:
### Product Description

Early No-Op refactor for [CCCT-2078](https://dimagi.atlassian.net/browse/CCCT-2078) without making any user visible changes yet.  

## Technical Summary

- Extracts biometric/PIN unlock logic from `PersonalIdManager.unlockConnect()` into a new `PersonalIdUnlocker` Kotlin singleton.
- Introduces an `UnlockPolicy` enum with two values:
  - `ALWAYS` — always prompts (preserves existing behaviour); used by all current callers.
  - `SESSION_WITH_TIME_THRESHOLD` — skips the prompt if the user unlocked within the last 10 minutes (groundwork for future use).
- Migrates all callers (`LoginActivity`, `ConnectMessagingActivity`, `ConnectNavHelper`, `ConnectUnlockFragment`, `PersonalIdManager` internals) to `PersonalIdUnlocker.unlock()`.
- Adds `resetSession()`, called on app start and Personal ID logout, so stale unlock timestamps are never inherited across sessions.

### Safety story

This is a pure refactor — no user-visible behaviour changes. All existing callers are migrated to `UnlockPolicy.ALWAYS`, which replicates the previous unconditional-prompt logic exactly. The new `SESSION_WITH_TIME_THRESHOLD` policy is introduced but not yet invoked in any production path. `resetSession()` is called at app start and on logout to ensure unlock state is never stale.

### Automated test coverage

`PersonalIdUnlockerTest` covers all branches of the session-bypass logic:
- No prior unlock → requires unlock
- Unlocked within threshold → bypasses prompt
- Unlocked beyond threshold (11 min) → requires unlock
- Unlocked at boundary (9 min) → bypasses prompt
- After `resetSession()` → requires unlock

[CCCT-2078]: https://dimagi.atlassian.net/browse/CCCT-2078?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -58,6 +58,15 @@ we would like to communicate to QA as part of the release testing
   - In order to achieve this functionality, DB migrations are done to accommodate the new email address field. QA should start testing with the previous version of the app, having PersonalID login already, and then upgrade to this new version. The app should work without crashing.
   - QA should also test with a fresh installation of this new version, going through PersonalID signup/recovery.
 
+
+- Verify that biometric/PIN unlock still triggers correctly in all existing entry points:
+  - Login screen auto-unlock (`LoginActivity`)
+  - Connect messaging notification tap (`ConnectMessagingActivity`)
+  - Connect nav helper flows (`ConnectNavHelper`)
+  - Connect unlock fragment (`ConnectUnlockFragment`)
+  - App-linking dialogs (`PersonalIdManager`)
+
+
 ## CommCare 2.63
 
 ### Release Notes
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -59,12 +59,7 @@ we would like to communicate to QA as part of the release testing
   - QA should also test with a fresh installation of this new version, going through PersonalID signup/recovery.
 
 
-- Verify that biometric/PIN unlock still triggers correctly in all existing entry points:
-  - Login screen auto-unlock (`LoginActivity`)
-  - Connect messaging notification tap (`ConnectMessagingActivity`)
-  - Connect nav helper flows (`ConnectNavHelper`)
-  - Connect unlock fragment (`ConnectUnlockFragment`)
-  - App-linking dialogs (`PersonalIdManager`)
+- Verify that biometric/PIN unlock still triggers correctly in all existing entry points
 
 
 ## CommCare 2.63
--- a/app/src/org/commcare/connect/ConnectNavHelper.kt
+++ b/app/src/org/commcare/connect/ConnectNavHelper.kt
@@ -21,7 +21,6 @@ object ConnectNavHelper {
         listener: ConnectActivityCompleteListener,
         navigationAction: (Context) -> Unit,
     ) {
-        PersonalIdManager.getInstance().init(activity)
         PersonalIdUnlocker.unlock(activity, UnlockPolicy.ALWAYS) { success ->
             if (success) {
                 navigationAction(activity)
--- a/app/src/org/commcare/connect/ConnectNavHelper.kt
+++ b/app/src/org/commcare/connect/ConnectNavHelper.kt
@@ -12,6 +12,8 @@ import org.commcare.connect.ConnectConstants.GO_TO_JOB_STATUS
 import org.commcare.connect.ConnectConstants.OPPORTUNITY_UUID
 import org.commcare.connect.ConnectConstants.SHOW_LAUNCH_BUTTON
 import org.commcare.connect.database.ConnectUserDatabaseUtil
+import org.commcare.personalId.PersonalIdUnlocker
+import org.commcare.personalId.UnlockPolicy
 
 object ConnectNavHelper {
     private fun unlockAndGoTo(
@@ -19,11 +21,8 @@ object ConnectNavHelper {
         listener: ConnectActivityCompleteListener,
         navigationAction: (Context) -> Unit,
     ) {
-        val personalIdManager: PersonalIdManager = PersonalIdManager.getInstance()
-        personalIdManager.init(activity)
-        personalIdManager.unlockConnect(
-            activity,
-        ) { success: Boolean ->
+        PersonalIdManager.getInstance().init(activity)
+        PersonalIdUnlocker.unlock(activity, UnlockPolicy.ALWAYS) { success ->
             if (success) {
                 navigationAction(activity)
             }
--- a/app/src/org/commcare/personalId/PersonalIdUnlocker.kt
+++ b/app/src/org/commcare/personalId/PersonalIdUnlocker.kt
@@ -16,8 +16,9 @@ import org.commcare.google.services.analytics.FirebaseAnalyticsUtil
 import org.commcare.utils.BiometricsHelper
 import org.commcare.utils.EncryptionKeyProvider
 import org.javarosa.core.services.Logger
+import kotlin.time.Duration.Companion.minutes
 
-private const val SESSION_UNLOCK_THRESHOLD_MS = 10 * 60 * 1000L // 10 minutes
+private val SESSION_UNLOCK_THRESHOLD_MS = 10.minutes.inWholeMilliseconds
 internal const val BIOMETRIC_INVALIDATION_KEY = "biometric-invalidation-key"
 
 /** Middleware to manage Personal ID unlock prompts with session-based bypass logic. */
@@ -58,16 +59,15 @@ object PersonalIdUnlocker {
         activity: CommCareActivity<*>,
         callback: PersonalIdManager.ConnectActivityCompleteListener,
     ) {
+        val personalIdManager = PersonalIdManager.getInstance()
         if (BuildConfig.IS_QA_AUTOMATION) {
             lastUnlockTime = SystemClock.elapsedRealtime()
-            PersonalIdManager.getInstance().userUnlockedPersonalId()
+            personalIdManager.userUnlockedPersonalId()
             callback.connectActivityComplete(true)
             return
         }
 
         logBiometricInvalidations(activity)
-
-        val personalIdManager = PersonalIdManager.getInstance()
         val bioManager = personalIdManager.getBiometricManager(activity)
         val user = ConnectUserDatabaseUtil.getUser(activity)
 
@@ -119,7 +119,7 @@ object PersonalIdUnlocker {
     private fun logBiometricInvalidations(activity: Context) {
         if (!AndroidKeyStore.doesKeyExist(BIOMETRIC_INVALIDATION_KEY)) return
         val provider = EncryptionKeyProvider(activity, true, BIOMETRIC_INVALIDATION_KEY)
-        if (!provider.isKeyValid()) {
+        if (!provider.isKeyValid) {
             FirebaseAnalyticsUtil.reportBiometricInvalidated()
             provider.deleteKey()
             provider.getKeyForEncryption()
--- a/app/src/org/commcare/personalId/PersonalIdUnlocker.kt
+++ b/app/src/org/commcare/personalId/PersonalIdUnlocker.kt
@@ -6,6 +6,7 @@ import org.commcare.connect.PersonalIdManager
 
 private const val SESSION_UNLOCK_THRESHOLD_MS = 10 * 60 * 1000L // 10 minutes
 
+/** Middleware to manage Personal ID unlock prompts with session-based bypass logic. */
 object PersonalIdUnlocker {
     @VisibleForTesting
     internal var lastUnlockTime: Long? = null
@@ -40,7 +41,9 @@ object PersonalIdUnlocker {
         activity: CommCareActivity<*>,
         callback: PersonalIdManager.ConnectActivityCompleteListener,
     ) {
-        PersonalIdManager.getInstance().unlockConnect(activity) { success ->
+        val personalIdManager = PersonalIdManager.getInstance()
+        personalIdManager.init(activity)
+        personalIdManager.unlockConnect(activity) { success ->
             if (success) lastUnlockTime = System.currentTimeMillis()
             callback.connectActivityComplete(success)
         }
--- a/app/src/org/commcare/personalId/PersonalIdUnlocker.kt
+++ b/app/src/org/commcare/personalId/PersonalIdUnlocker.kt
@@ -17,7 +17,6 @@ import org.commcare.utils.EncryptionKeyProvider
 import org.javarosa.core.services.Logger
 
 private const val SESSION_UNLOCK_THRESHOLD_MS = 10 * 60 * 1000L // 10 minutes
-private const val BIOMETRIC_INVALIDATION_KEY = "biometric-invalidation-key"
 
 /** Middleware to manage Personal ID unlock prompts with session-based bypass logic. */
 object PersonalIdUnlocker {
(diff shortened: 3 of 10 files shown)


app/src/org/commcare/personalId/PersonalIdUnlocker.kt
```diff
@@ -0,0 +1,152 @@
...
+                ) = callback.connectActivityComplete(false)
+
+                override fun onAuthenticationSucceeded(result: BiometricPrompt.AuthenticationResult) {
+                    lastUnlockTime = System.currentTimeMillis()
+                    personalIdManager.userUnlockedPersonalId()
+                    callback.connectActivityComplete(true)
+                }
+
+                override fun onAuthenticationFailed() = callback.connectActivityComplete(false)
+            }
```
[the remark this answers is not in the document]
shubham1g5: not taking any action here at the moment to keep this PR strictly a no-op (no user changes)

app/unit-tests/src/org/commcare/personalId/PersonalIdUnlockerTest.kt
```diff
@@ -0,0 +1,45 @@
...
+        PersonalIdUnlocker.lastUnlockTime = elevenMinutesAgo
+        assertTrue(PersonalIdUnlocker.requiresUnlockForSession())
+    }
+
+    @Test
+    fun `SESSION_WITH_TIME_THRESHOLD bypasses unlock when last unlock is within threshold boundary`() {
+        val nineMinutesAgo = System.currentTimeMillis() - (9 * 60 * 1000L)
+        PersonalIdUnlocker.lastUnlockTime = nineMinutesAgo
+        assertFalse(PersonalIdUnlocker.requiresUnlockForSession())
+    }
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3702/commits/a921c2b2a906a2cf1c951d325a0c24c327e6d5d2

app/src/org/commcare/personalId/PersonalIdUnlocker.kt
```diff
@@ -0,0 +1,152 @@
...
+import org.commcare.utils.EncryptionKeyProvider
+import org.javarosa.core.services.Logger
+
+private const val SESSION_UNLOCK_THRESHOLD_MS = 10 * 60 * 1000L // 10 minutes
+internal const val BIOMETRIC_INVALIDATION_KEY = "biometric-invalidation-key"
+
+/** Middleware to manage Personal ID unlock prompts with session-based bypass logic. */
+object PersonalIdUnlocker {
+    @VisibleForTesting
+    internal var lastUnlockTime: Long? = null
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3702/commits/a921c2b2a906a2cf1c951d325a0c24c327e6d5d2

app/src/org/commcare/personalId/PersonalIdUnlocker.kt
```diff
@@ -0,0 +1,153 @@
...
+import org.commcare.connect.PersonalIdManager
+import org.commcare.connect.database.ConnectUserDatabaseUtil
+import org.commcare.dalvik.BuildConfig
+import org.commcare.dalvik.R
+import org.commcare.google.services.analytics.FirebaseAnalyticsUtil
+import org.commcare.utils.BiometricsHelper
+import org.commcare.utils.EncryptionKeyProvider
+import org.javarosa.core.services.Logger
+
+private const val SESSION_UNLOCK_THRESHOLD_MS = 10 * 60 * 1000L // 10 minutes
```
conroy-ricketts: Can use Kotlin's duration API here (e.g. `SESSION_UNLOCK_THRESHOLD_MS = 10.minutes`)

app/src/org/commcare/personalId/PersonalIdUnlocker.kt
```diff
@@ -0,0 +1,153 @@
...
+        if (BuildConfig.IS_QA_AUTOMATION) {
+            lastUnlockTime = SystemClock.elapsedRealtime()
+            PersonalIdManager.getInstance().userUnlockedPersonalId()
+            callback.connectActivityComplete(true)
+            return
+        }
+
+        logBiometricInvalidations(activity)
+
+        val personalIdManager = PersonalIdManager.getInstance()
```
conroy-ricketts: Nit: this variable can be moved to the first line of the function so that it can be used inside `if (BuildConfig.IS_QA_AUTOMATION) {  }` as well

app/src/org/commcare/personalId/PersonalIdUnlocker.kt
```diff
@@ -0,0 +1,153 @@
...
+                        Toast.LENGTH_SHORT,
+                    ).show()
+            }
+        }
+    }
+
+    private fun logBiometricInvalidations(activity: Context) {
+        if (!AndroidKeyStore.doesKeyExist(BIOMETRIC_INVALIDATION_KEY)) return
+        val provider = EncryptionKeyProvider(activity, true, BIOMETRIC_INVALIDATION_KEY)
+        if (!provider.isKeyValid()) {
```
conroy-ricketts: Nit: with Kotlin we can use this like a property


```suggestion
        if (!provider.isKeyValid) {
```

app/unit-tests/src/org/commcare/personalId/PersonalIdUnlockerTest.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    @Test
+    fun `SESSION_WITH_TIME_THRESHOLD bypasses unlock when unlocked within threshold`() {
+        PersonalIdUnlocker.lastUnlockTime = SystemClock.elapsedRealtime()
+        assertFalse(PersonalIdUnlocker.requiresUnlockForSession())
+    }
+
+    @Test
+    fun `SESSION_WITH_TIME_THRESHOLD requires unlock when last unlock at exactly the threshold`() {
+        val tenMinutesAgo = SystemClock.elapsedRealtime() - (10 * 60 * 1000L)
```
conroy-ricketts: Kotlin's Duration API can be used throughout these tests as well (e.g. `10.minutes`, `11.minutes`, `9.minutes`)

conroy-ricketts (commented):
This looks pretty good, but why do this before publishing the Claude plan for review?

shubham1g5:
> This looks pretty good, but why do this before publishing the Claude plan for review?

Mostly because I didn't see the need to get a design review on this given how straightforward the refactor is and followed my gut on it to be more efficient to have review on code directly for the team.

app/src/org/commcare/personalId/PersonalIdUnlocker.kt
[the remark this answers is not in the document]
OrangeAndGreen: Noting that I think this may be a bug I've encountered, i.e. when my first attempt to scan my fingerprint fails but an additional attempt succeeds. I think I've seen an error toast appear after the first failure indicating a more serious error than what actually happened (the user just needs to try scanning their fingerprint again). Okay with not fixing this here, but we should put in a ticket to make sure this gets done

app/src/org/commcare/connect/ConnectNavHelper.kt
```diff
@@ -12,18 +12,16 @@ import org.commcare.connect.ConnectConstants.GO_TO_JOB_STATUS
...
+import org.commcare.personalId.UnlockPolicy
 
 object ConnectNavHelper {
     private fun unlockAndGoTo(
         activity: CommCareActivity<*>,
         listener: ConnectActivityCompleteListener,
         navigationAction: (Context) -> Unit,
     ) {
-        val personalIdManager: PersonalIdManager = PersonalIdManager.getInstance()
-        personalIdManager.init(activity)
```
OrangeAndGreen, line 23: We're dropping the init call to PersonalIdManager here, is that going to be okay? It doesn't seem like the unlocker is going to do it

RELEASES.md
```diff
@@ -58,6 +58,15 @@ we would like to communicate to QA as part of the release testing
   - In order to achieve this functionality, DB migrations are done to accommodate the new email address field. QA should start testing with the previous version of the app, having PersonalID login already, and then upgrade to this new version. The app should work without crashing.
   - QA should also test with a fresh installation of this new version, going through PersonalID signup/recovery.
 
+
+- Verify that biometric/PIN unlock still triggers correctly in all existing entry points:
+  - Login screen auto-unlock (`LoginActivity`)
+  - Connect messaging notification tap (`ConnectMessagingActivity`)
+  - Connect nav helper flows (`ConnectNavHelper`)
```
OrangeAndGreen: Is this going to be meaningful to QA? We may need to be more explicit about how they can achieve these flows

conroy-ricketts (dismissed):

app/src/org/commcare/personalId/PersonalIdUnlocker.kt
```diff
@@ -0,0 +1,153 @@
...
+     * whether the prompt is actually shown. Always invokes [callback] with the result.
+     *
+     * [UnlockPolicy.ALWAYS] — always prompts (existing behaviour, no session bypass).
+     * [UnlockPolicy.SESSION_WITH_TIME_THRESHOLD] — skips prompt if unlocked within
+     *     the last 10 minutes; caller is responsible for first checking that the app
+     *     is Personal ID managed before calling this.
+     */
+    fun unlock(
+        activity: CommCareActivity<*>,
+        policy: UnlockPolicy,
```
OrangeAndGreen, line 40: Design discussion: I'm wondering whether we should have each caller specify the unlock policy independently, or whether that should be more centrally managed?

It feels like each caller shouldn't need to worry about the unlock policy, and instead it could be maintained internally by the new PersonalIdUnlocker. This would allow centralized decisions about the policy (similar to how the 10-minute timeout is centrally managed).

The only caveat I can think of is if we want different callers to be able to apply different policies, like always unlock in certain situations and use the timeout in others. But right now it doesn't seem like that applies anywhere, and even in the long-term it might make more sense to add an override that supports that functionality when we need it (like a forceUnlock function).

OrangeAndGreen (dismissed):
I left a few comments worth considering but nothing I consider blocking if you prefer it as is. The most concerning would be about the removed call to PersonalIdManager.init.

app/src/org/commcare/personalId/PersonalIdUnlocker.kt
```diff
@@ -0,0 +1,153 @@
...
+        if (!provider.isKeyValid) {
+            FirebaseAnalyticsUtil.reportBiometricInvalidated()
+            provider.deleteKey()
+            provider.getKeyForEncryption()
+        }
+    }
+
+    @VisibleForTesting
+    internal fun requiresUnlockForSession(): Boolean {
+        val last = lastUnlockTime ?: return true
```
Jignesh-dimagi, line 132: 'lastUnlockTime' is the timestamp when the user unlocked the app the last time. Are we handling background retriggered re-auth with this variable only, or will it be separate, like measuring the background timestamp and taking action?

Jignesh-dimagi:
Looks good as a refactor, assuming that in the next PR, it will have an implementation with UnlockPolicy. SESSION_WITH_TIME_THRESHOLD, is that the correct understanding?

Jignesh-dimagi (dismissed):

app/src/org/commcare/personalId/PersonalIdUnlocker.kt
shubham1g5: > Are we handling background retriggered re-auth with this variable only

what do you mean by this ? Are you able to give an example or use case where we do this today ?

shubham1g5:
> Looks good as a refactor, assuming that in the next PR, it will have an implementation with UnlockPolicy. SESSION_WITH_TIME_THRESHOLD, is that the correct understanding?

yes

app/src/org/commcare/personalId/PersonalIdUnlocker.kt
shubham1g5: > The only caveat I can think of is if we want different callers to be able to apply different policies, 

Think that's quite true with the `Login with Personal ID` where the unlock policy will need to be `ALWAYS` as if user is logged out of the app, we will require an explicit auth for them to login everytime they logout of the app.

app/src/org/commcare/personalId/PersonalIdUnlocker.kt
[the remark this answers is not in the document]
shubham1g5: Agree, I will see if I can repro and get to it in another PR, otherwise will create a ticket on this.

RELEASES.md
shubham1g5: those seems like a waste for QA, I kept it simple in - https://github.com/dimagi/commcare-android/pull/3702/commits/8cfa508da3434152f23730d7ce49bbe464958af4 and will refine those with the next PR on user visible changes for this feature.

app/src/org/commcare/connect/ConnectNavHelper.kt
shubham1g5: added it back in unlocker here - https://github.com/dimagi/commcare-android/pull/3702/commits/d392cb043a86678a0dfc9a963929df92de578382

conroy-ricketts (approved):

app/src/org/commcare/personalId/PersonalIdUnlocker.kt
Jignesh-dimagi: There can be 3 types of timeout, viz. 
 - Time elapsed between 2 consecutive unlocks
 - In activity timeout (considering that it's out of scope for now)
 - App in background timeout

Do we have differentiation between 1 and 3?

shubham1g5: thanks for additional context, we are not differentiating here between 1 and3 and asking for re-auth every 10 mins, no matter what.

app/src/org/commcare/personalId/PersonalIdUnlocker.kt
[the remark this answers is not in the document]
shubham1g5: @OrangeAndGreen Do you know where in the app the bug you mentioned happens ? I was not able to repro it on just trying at random places. Are you open to take a look here and would be great if you can file an issue if you are able to repro what you meant.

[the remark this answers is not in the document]
OrangeAndGreen: Yes! I just hit it again on Friday, and verified these steps now:
* With PersonalId configured, login to a trad'l app using user/pass
* When prompted to link to PersonalId, proceed
* Now scan the wrong finger

You'll be prompted to try again, but you'll also see the page behind immediately transition to app home. After that it doesn't matter whether you succeed or fail the biometric, your app doesn't get linked to PersonalId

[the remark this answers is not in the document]
shubham1g5: Ahh damn, agree that's bad and worth fixing. Although think we might need to look at this behaviour at all places we do unlock.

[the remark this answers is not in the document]
shubham1g5: Added an issue here for this - https://dimagi.atlassian.net/browse/CCCT-2427. I was able to repro this although think it's not as bad as I was imagining as user has entered the username and password already so technically they should be able to move to the App Home and the auth we require is only for linking to Personal ID. 

Does the desired behaviour outlined on ticket seems right to you ?

PR #3707, opened 2026-05-11
CCCT-2396 Warn Before Form Exit On Push Notification Tap

OrangeAndGreen:
## Product Description

When a user has a partially completed form open and taps a navigation push notification (e.g. a Connect message, payment, opportunity, or session-endpoint deep link), they are now shown the same quit-form warning dialog the back button shows: **STAY IN FORM / EXIT WITHOUT SAVING / SAVE INCOMPLETE**. Choosing *STAY IN FORM* keeps them in the form; *EXIT WITHOUT SAVING* discards changes and navigates to the notification target; *SAVE INCOMPLETE* saves the form as incomplete and then navigates to the notification target. Today, tapping such a notification mid-form silently tears down the form and discards any unsaved data. SYNC-only notifications and notifications tapped when no form is open behave exactly as before.

Demo #1: SAVE INCOMPLETE

https://github.com/user-attachments/assets/99766855-30bd-43cc-9474-a42b7f22919d

Demo #2: STAY IN FORM

https://github.com/user-attachments/assets/b193e230-face-453e-b727-fd739716db94

Demo #3: EXIT WITHOUT SAVING

https://github.com/user-attachments/assets/705eaef7-ae63-4fe7-ab7d-acb999f68904

## Technical Summary

The notification `PendingIntent` built in `FirebaseMessagingUtil.buildNotification` previously used `FLAG_ACTIVITY_CLEAR_TASK | FLAG_ACTIVITY_NEW_TASK` and pointed directly at `DispatchActivity` / `ConnectActivity`, which destroyed `FormEntryActivity` with no prompt. Decisions about whether to interrupt or pass through could not be made at notification-build time (the form-active state is only known at tap time), so a transparent dispatcher activity was introduced.

Key pieces:

- **New `PushNotificationLaunchActivity`** (Kotlin, `Theme.NoDisplay`, `singleInstance`, `taskAffinity=""`, `excludeFromRecents`, `exported=false`). Every navigation notification's `PendingIntent` now targets this activity. It reads the original target intent from `EXTRA_WRAPPED_NAV_INTENT` and at tap time queries `FormEntryActivity.isFormEntryInProgress()`. If no form is open, it adds the original `CLEAR_TASK | NEW_TASK` flags to a defensive copy of the wrapped intent and starts it — identical to today's behavior. If a form is open, it re-enters `FormEntryActivity` in `singleTop` with `CLEAR_TOP`, carrying the wrapped intent as `EXTRA_PENDING_NAV_INTENT`.
- **`FormEntryActivity.onNewIntent`** override routes the wrapped extra to a new sibling method `triggerUserQuitInputForExternalNav(Intent)`. It mirrors the existing `triggerUserQuitInput()` short-circuits (in-flight save → drop tap; form not loaded → start nav and finish; read-only → start nav and finish via `finishReturnInstance(false)`) but, in the default case, shows a new overload `FormEntryDialogs.createQuitDialog(activity, isIncompleteEnabled, pendingNav)`.
- **The dialog overload** reuses the existing `quit_form_title` and choice strings. Stay drops `pendingNav`. Discard routes through a new `FormEntryActivity.discardChangesAndExitToPendingNav` that calls `startPendingNavSafely` (wraps `startActivity` with an `ActivityNotFoundException` guard) before `discardChangesAndExit`, matching the read-only branch's pattern. Save stashes `pendingNav` on the activity and triggers `saveFormToDisk(EXIT)`; `savingComplete`'s `SAVED_AND_EXIT` branch starts the pending nav before `finishReturnInstance`. `mPendingNavAfterSave` is persisted via `onSaveInstanceState` so the save-and-navigate sequence survives process death after the user has picked Save.
- **Analytics:** added `AnalyticsParamValue.PUSH_NOTIFICATION_TAP`; `triggerUserQuitInputForExternalNav` calls `FirebaseAnalyticsUtil.reportFormQuitAttempt` with that source (after the in-flight-save guard so silently-dropped taps don't pollute the event stream).

`FirebaseMessagingUtil.buildNotification` no longer applies `CLEAR_TASK | NEW_TASK` to the wrapped intent — those flags are re-applied only inside the launch activity's no-form-active branch. The direct-launch paths (`getIntentForPNIfAny`, `getIntentForPNClick`) never went through `buildNotification`, so their behavior is unchanged.

### Safety story

**What gives confidence:**

- I built the debug APK locally and manually exercised the feature on a device, walking through the QA scenarios from `RELEASES.md` — the dialog appears with the three expected choices, each choice produces the expected outcome (stay / discard-and-navigate / save-and-navigate), and notifications tapped outside form entry still open their target directly.
- I reviewed the full diff. The flow is narrow: one new dispatcher activity, an extra branch in `onNewIntent`, a new `createQuitDialog` overload that reuses the existing dialog construction, and one change inside `savingComplete`'s `SAVED_AND_EXIT` branch.
- Unit tests cover the dispatcher's two dispatch branches and the missing-extra case, the new dialog overload's Discard and Save listener behaviors, the `onNewIntent` no-op path, the `mPendingNavAfterSave` persistence round-trip, and the `buildNotification` wrap-and-target change.
- SYNC payloads are completely untouched — they never reach `buildNotification`.
- `ActivityNotFoundException` paths are caught and logged via the new `startPendingNavSafely` helper (and the dispatcher's own try/catch), so a missing target activity won't strand the user.

**Risks to review:**

- I did not separately re-run the regression scenarios for the existing Connect notification types (messages, payments, learn/delivery progress, opportunity summary, session-endpoint deep links). The QA notes call these out, but my own manual testing focused on the new dialog flow.
- The dispatcher's `singleInstance` launch mode means a second tap arriving while the activity is mid-dispatch could in theory be dropped (the activity overrides `onCreate` only, not `onNewIntent`). The dispatch path is microseconds, so this is unlikely in practice, but it is not exercised by tests.
- Process death while the dialog is showing but **before** the user picks Save will drop the pending intent — `mPendingNavAfterSave` is only set after the Save choice. The user must re-tap the notification in that case. This is an accepted limitation noted in the QA notes' coverage of process death, but it is a behavioral edge case.
- `getIntentForPNIfAny` and the in-app notification list (`PushNotificationActivity`) bypass the dispatcher entirely — tapping a notification *in those surfaces* while a form is open will still tear down the form with no dialog. The scope is the tray PendingIntent only; widening it is a separate change.
- The `onSaveInstanceState` / `loadStateFromBundle` calls use the deprecated single-argument `getParcelable*` forms (no API-33 guard). Consistent with the rest of the file but worth noting.
- This PR targets the `commcare_2.63` release branch, not `master`. The same change should be ported forward to `master` as a separate PR.

### Automated test coverage

Five new unit-test classes in `app/unit-tests/src/org/commcare/`:

- `activities/PushNotificationLaunchActivityTest` — three tests: form-inactive branch dispatches with `CLEAR_TASK | NEW_TASK`; form-active branch re-enters `FormEntryActivity` with `SINGLE_TOP | CLEAR_TOP` and the wrapped extra preserved; missing-wrapped-extra case finishes without starting anything.
- `activities/FormEntryActivityExternalNavTest` — `mPendingNavAfterSave` survives an `onSaveInstanceState` round trip; `onNewIntent` without the extra does not interact with `FormEntryDialogs`.
- `activities/components/FormEntryDialogsExternalNavTest` — Discard routes through `discardChangesAndExitToPendingNav`; Save stashes the intent then triggers an EXIT-flavored save; no helpers fire when the user chooses Stay.
- `utils/FirebaseMessagingUtilBuildNotificationTest` — the constructed `PendingIntent` targets `PushNotificationLaunchActivity` and preserves the original target as `EXTRA_WRAPPED_NAV_INTENT`, including `NOTIFICATION_ID` on the wrapped intent.

The dialog-shown short-circuits inside `triggerUserQuitInputForExternalNav` (form-not-loaded and read-only) depend on a fully initialized `FormController` and are not feasible in unit form; they are covered by the QA plan.

[CCCT-2396]: https://dimagi.atlassian.net/browse/CCCT-2396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -338,9 +342,21 @@ protected void onSaveInstanceState(Bundle outState) {
...
     }
 
+    @Override
+    protected void onNewIntent(@NonNull Intent intent) {
+        super.onNewIntent(intent);
+        Intent pendingNav = intent.getParcelableExtra(EXTRA_PENDING_NAV_INTENT);
+        if (pendingNav != null) {
+            triggerUserQuitInputForExternalNav(pendingNav);
+        }
+    }
```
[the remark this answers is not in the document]
OrangeAndGreen: Fixed in 69b697d35. Wrapped the `getParcelableExtra` in a `Build.VERSION.SDK_INT >= TIRAMISU` check, mirroring the pattern in `PushNotificationLaunchActivity.kt`.

*— Claude Code*

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -1586,6 +1648,8 @@ private void loadStateFromBundle(Bundle savedInstanceState) {
                 this.instanceIsReadOnly = savedInstanceState.getBoolean(KEY_IS_READ_ONLY);
             }
 
+            mPendingNavAfterSave = savedInstanceState.getParcelable(EXTRA_PENDING_NAV_INTENT);
```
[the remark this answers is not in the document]
OrangeAndGreen: Fixed in 69b697d35. Same TIRAMISU-aware pattern applied to the `Bundle.getParcelable` call for `mPendingNavAfterSave` restoration.

*— Claude Code*

app/src/org/commcare/utils/FirebaseMessagingUtil.java
```diff
@@ -547,16 +548,19 @@ private static Intent getConnectActivityNotification(Context context, FCMMessage
...
 
         int flags = Build.VERSION.SDK_INT >= Build.VERSION_CODES.S
                 ? PendingIntent.FLAG_UPDATE_CURRENT | PendingIntent.FLAG_IMMUTABLE
                 : PendingIntent.FLAG_UPDATE_CURRENT;
 
-        Bundle bundleExtras = new Bundle();
-        intent.putExtra(NOTIFICATION_ID, fcmMessageData.getPayloadData().get(NOTIFICATION_ID));
-
-        PendingIntent contentIntent = PendingIntent.getActivity(context, 0, intent, flags);
+        PendingIntent contentIntent = PendingIntent.getActivity(context, 0, launchIntent, flags);
```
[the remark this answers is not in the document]
OrangeAndGreen: Fixed in 69b697d35. The `PendingIntent` now uses a unique request code derived from `NOTIFICATION_ID` via `NotificationIdentifiers.generateNotificationIdFromString` (falling back to `FCM_NOTIFICATION_ID`), and the launch Intent action is tagged with the same request code so each `PendingIntent` has distinct identity. Matches the existing pattern in `showNotification()`.

*— Claude Code*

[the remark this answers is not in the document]
OrangeAndGreen: [69b697d](https://github.com/dimagi/commcare-android/pull/3707/commits/69b697d35516c779e9188f59d6437a790201a8ce)

app/src/org/commcare/activities/FormEntryActivity.java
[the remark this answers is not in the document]
OrangeAndGreen: [69b697d](https://github.com/dimagi/commcare-android/pull/3707/commits/69b697d35516c779e9188f59d6437a790201a8ce)

app/src/org/commcare/activities/FormEntryActivity.java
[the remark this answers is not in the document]
OrangeAndGreen: [69b697d](https://github.com/dimagi/commcare-android/pull/3707/commits/69b697d35516c779e9188f59d6437a790201a8ce)

conroy-ricketts:
@OrangeAndGreen Question about the PR description that seemed a bit confusing:

Three options are mentioned for the exit dialog: "Don't exit / Don't save / Keep changes", but the current exit dialog only has two options: Stay in form / Exit without saving.

Was the current exit dialog itself modified at all?

OrangeAndGreen:
@conroy-ricketts Ah, that's existing behavior built-in to the dialog. The option to "Keep changes" only appears if the app is configured to support incomplete form storage. The app I tested with isn't configured for that, but nothing with that logic changed in this PR so it should work as expected.

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -1235,6 +1256,51 @@ protected void triggerUserQuitInput() {
...
+    /**
+     * Variant of {@link #triggerUserQuitInput()} for the case where the user is being
+     * pulled away from the form by an external navigation event (currently: tapping a push
+     * notification). On Save / Discard the form's existing exit handling runs and then
+     * pendingNav is started so the user reaches the notification's target.
+     */
+    protected void triggerUserQuitInputForExternalNav(Intent pendingNav) {
+        if (mSaveToDiskTask != null) {
+            return;
+        }
```
shubham1g5, line 1272: what is this check for meant to protect ? And seems like users will loose push notification tap in these cases ? It would be nice to add a log here nevertheless.

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -1235,6 +1256,51 @@ protected void triggerUserQuitInput() {
...
+        FirebaseAnalyticsUtil.reportFormQuitAttempt(
+                AnalyticsParamValue.PUSH_NOTIFICATION_TAP,
+                getCurrentFormXmlnsFailSafe());
+
+        if (!formHasLoaded()) {
+            startPendingNavSafely(pendingNav);
+            finish();
+            return;
+        }
+        if (mFormController != null && mFormController.isFormReadOnly()) {
```
shubham1g5: `mFormController != null ` is not necessary given it's the same check done by `formHasLoaded`

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -1235,6 +1256,51 @@ protected void triggerUserQuitInput() {
...
+        FormEntryDialogs.createQuitDialog(this, mIncompleteEnabled, pendingNav);
+    }
+
+    private void startPendingNavSafely(Intent pendingNav) {
+        try {
+            startActivity(pendingNav);
+        } catch (android.content.ActivityNotFoundException e) {
+            Logger.exception(
+                    "Push notification target activity not found", e);
+        }
```
shubham1g5: lets just crash here as it means there is definitely an obvious bug in our implementation.

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -1354,6 +1420,7 @@ public void savingComplete(SaveToDiskTask.SaveStatus saveStatus, String errorMes
                         Toast.makeText(this,
                                 Localization.get("form.entry.complete.save.success"), Toast.LENGTH_SHORT).show();
                     }
+                    consumePendingNavAfterSave();
```
shubham1g5: what workflow does this relates to ? This can interfere with [End of form navigation](https://dimagi.atlassian.net/wiki/spaces/commcarepublic/pages/2143955342/End+of+Form+Navigation) and would avoid unless it's achieving a strict requirement for this PR.

app/src/org/commcare/activities/PushNotificationLaunchActivity.kt
```diff
@@ -0,0 +1,74 @@
...
+
+    private fun dispatchWithoutForm(wrapped: Intent) {
+        val toStart =
+            Intent(wrapped)
+                .addFlags(Intent.FLAG_ACTIVITY_CLEAR_TASK or Intent.FLAG_ACTIVITY_NEW_TASK)
+        try {
+            startActivity(toStart)
+        } catch (e: ActivityNotFoundException) {
+            Logger.exception("Push notification target activity not found", e)
+        }
```
shubham1g5: think good to hard crash here and above for ActivityNotFoundException

app/src/org/commcare/utils/FirebaseMessagingUtil.java
```diff
@@ -547,16 +548,24 @@ private static Intent getConnectActivityNotification(Context context, FCMMessage
...
 
-        Bundle bundleExtras = new Bundle();
-        intent.putExtra(NOTIFICATION_ID, fcmMessageData.getPayloadData().get(NOTIFICATION_ID));
-
-        PendingIntent contentIntent = PendingIntent.getActivity(context, 0, intent, flags);
+        String notificationId = fcmMessageData.getPayloadData().get(NOTIFICATION_ID);
+        int requestCode = !TextUtils.isEmpty(notificationId)
+                ? NotificationIdentifiers.generateNotificationIdFromString(notificationId)
+                : FCM_NOTIFICATION_ID;
+        launchIntent.setAction("org.commcare.notifications.tap." + requestCode);
```
shubham1g5, line 567: what are we trying to achieve here ? It feels very sketchy that we are setting an action with a request code referring to the notiifcation id.

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -1767,10 +1840,24 @@ public SecretKeySpec getSymetricKey() {
...
+
     public static boolean isFormEntryInProgress() {
         return isFormEntryActive;
     }
 
+    @VisibleForTesting
+    public static void setFormEntryInProgressForTest(boolean inProgress) {
+        isFormEntryActive = inProgress;
+    }
+
```
shubham1g5: This should not really need to be visible but instead can be determined by activity lifecycle itself, i.e. it should be set automatically by `onCreate` and `onDestroy` and we need to handle the activity lifecycle correctly in tests.

app/unit-tests/src/org/commcare/activities/FormEntryActivityExternalNavTest.kt
```diff
@@ -0,0 +1,72 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.MockedStatic
+import org.mockito.Mockito
+import org.robolectric.Robolectric
+import org.robolectric.annotation.Config
+
+@Config(application = CommCareTestApplication::class)
+@RunWith(AndroidJUnit4::class)
+class FormEntryActivityExternalNavTest {
```
shubham1g5: nit: rename to `FormEntryActivityPendingNavTest`

app/unit-tests/src/org/commcare/activities/PushNotificationLaunchActivityTest.kt
```diff
@@ -0,0 +1,107 @@
...
+@RunWith(AndroidJUnit4::class)
+class PushNotificationLaunchActivityTest {
+    @After
+    fun tearDown() {
+        // Ensure no in-progress form state leaks between tests.
+        FormEntryActivity.setFormEntryInProgressForTest(false)
+    }
+
+    @Test
+    fun whenFormInProgress_reentersFormEntryWithPendingNavExtraAndSingleTop() {
```
shubham1g5: This and the test below are not achieving much by testing these flags as it feels like they are testing more Android framework code to reliably pass intents rather than application logic.

app/unit-tests/src/org/commcare/activities/components/FormEntryDialogsExternalNavTest.kt
```diff
@@ -0,0 +1,60 @@
...
+import org.mockito.Mockito.mock
+import org.mockito.Mockito.never
+import org.mockito.Mockito.times
+import org.mockito.Mockito.verify
+import org.mockito.kotlin.any
+import org.robolectric.annotation.Config
+
+@Config(application = CommCareTestApplication::class)
+@RunWith(AndroidJUnit4::class)
+class FormEntryDialogsExternalNavTest {
```
shubham1g5: these tests seems quite unhelpful to me with mostly saying call a method that call method B and verify that method B was called.

shubham1g5 (changes requested):
Can we correct PR description here to provide exit dialog options exactly as they are (the PR descrioption seems to refer to non exisiting options like - `Don't exit`) . 

Also can we have a demo video here and on slack showing workflows for all 3 exit dialog options. 

Reg. testing, think the PR is better off not having any of the new tests as they hardly give confidence into the complete workflow we want to test, a better test here can be  one that test for final outcomes, for eg. Verify that user is still inside Form Entry or on a new re-directed screen based on the dialog option chosen by the user during a form entry.

app/src/org/commcare/activities/FormEntryActivity.java
OrangeAndGreen: The problem is that we don't want to immediately try to navigate away if a save is in progress. But a line was missing here, storing the intent so we can follow it once the save completes: [6fc5e1c](https://github.com/dimagi/commcare-android/pull/3707/commits/6fc5e1c2b59039372bfe5153fa06d7054d94c9c1)

app/src/org/commcare/activities/FormEntryActivity.java
OrangeAndGreen: Ah, so this relates to the latest change [here](https://github.com/dimagi/commcare-android/pull/3707#discussion_r3232859708), with the goal being to let the save complete and then do the navigation from the clicked notification.

But I see the issue with the end-of-form navigation as this would definitely interrupt that if a user clicks on a notification while a form save is in progress. But now the question is, what should the behavior be in that case? Sounds like we're in a race condition between two user actions with different navigation targets: completing and saving a form, and clicking on a push notification.

Just spit-balling, but if they click on a notification while waiting for a form to save that seems like the most recent action by the user and therefore the one we should prioritize. But alternatively maybe we just show a toast explaining that we can't navigate away during a form save?

shubham1g5: > maybe we just show a toast explaining that we can't navigate away during a form save?

yeah agree that seems a good option, although good to test that we are still able to redirect user after choosing to save form as incomplete.

app/src/org/commcare/activities/FormEntryActivity.java
OrangeAndGreen: [a97001f](https://github.com/dimagi/commcare-android/pull/3707/commits/a97001fa08680cb61921cc0833b226be6034cc48)

app/src/org/commcare/activities/FormEntryActivity.java
OrangeAndGreen: [f0503f1](https://github.com/dimagi/commcare-android/pull/3707/commits/f0503f103308dd06433878aef16fe41c94daf188)

app/src/org/commcare/activities/PushNotificationLaunchActivity.kt
OrangeAndGreen: [3ab47aa](https://github.com/dimagi/commcare-android/pull/3707/commits/3ab47aa170ee73d9d80a4ea789c58ed934d84108)

app/unit-tests/src/org/commcare/activities/components/FormEntryDialogsExternalNavTest.kt
OrangeAndGreen: Deleted [3301cc9](https://github.com/dimagi/commcare-android/pull/3707/commits/3301cc961ca7723251d044ac8c4f004bc32aafcc)

app/unit-tests/src/org/commcare/activities/PushNotificationLaunchActivityTest.kt
OrangeAndGreen: Deleted [3301cc9](https://github.com/dimagi/commcare-android/pull/3707/commits/3301cc961ca7723251d044ac8c4f004bc32aafcc)

app/unit-tests/src/org/commcare/activities/FormEntryActivityExternalNavTest.kt
OrangeAndGreen: Deleted [3301cc9](https://github.com/dimagi/commcare-android/pull/3707/commits/3301cc961ca7723251d044ac8c4f004bc32aafcc)

app/src/org/commcare/activities/FormEntryActivity.java
OrangeAndGreen: Deleted [3301cc9](https://github.com/dimagi/commcare-android/pull/3707/commits/3301cc961ca7723251d044ac8c4f004bc32aafcc)

app/src/org/commcare/activities/components/FormEntryDialogs.java
```diff
@@ -72,6 +86,24 @@ public static void createQuitDialog(final FormEntryActivity activity, boolean is
...
+            activity.discardChangesAndExit();
+        }
+    }
+
+    static void handleSaveChoice(FormEntryActivity activity, Intent pendingNav) {
+        if (pendingNav != null) {
+            activity.setPendingNavAfterSave(pendingNav);
+        }
+        activity.saveFormToDisk(FormEntryConstants.EXIT);
+    }
```
Jignesh-dimagi: Do we need these methods? Ultimately Flow is moving back to 'FormEntryActivity' via methods discardChangesAndExit and saveFormToDisk, where, depending upon the action, the app can check for EXTRA_PENDING_NAV_INTENT and redirect accordingly if required.

If the user has discarded the changes, `discardChangesAndExit()` can startActivity if there is any EXTRA_PENDING_NAV_INTENT?

If the user has chosen to save the changes, the 'case SAVED_AND_EXIT' of the 'savingComplete' method can startActivity if there is any EXTRA_PENDING_NAV_INTENT.

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -1235,6 +1256,44 @@ protected void triggerUserQuitInput() {
...
+    /**
+     * Variant of {@link #triggerUserQuitInput()} for the case where the user is being
+     * pulled away from the form by an external navigation event (currently: tapping a push
+     * notification). On Save / Discard the form's existing exit handling runs and then
+     * pendingNav is started so the user reaches the notification's target.
+     */
+    protected void triggerUserQuitInputForExternalNav(Intent pendingNav) {
+        if (mSaveToDiskTask != null) {
+            //Store the pending nav intent so it can be followed once the save is complete
+            mPendingNavAfterSave = pendingNav;
```
Jignesh-dimagi: I think this variable will not be required if we check EXTRA_PENDING_NAV_INTENT as mentioned in [this](https://github.com/dimagi/commcare-android/pull/3707/changes#r3234804261) after every discard or save action.

app/src/org/commcare/activities/PushNotificationLaunchActivity.kt
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * Transparent, no-UI launch activity for push-notification PendingIntents.
+ *
+ * Decides at tap time whether to:
+ *   - dispatch the wrapped Intent directly (no form in progress), or
+ *   - re-enter [FormEntryActivity] in singleTop carrying the wrapped Intent so the
+ *     user is prompted before navigation occurs (form in progress).
+ */
+class PushNotificationLaunchActivity : Activity() {
```
Jignesh-dimagi, line 19: I am not sure if this task has that scope or not, but as we are now introducing the activity between push notification and target screen, can we take advantage of it by having the following?
 - Checking if the user has personalId login or not.
 - If any biometric security is required before opening the screen.

app/AndroidManifest.xml
```diff
@@ -357,6 +357,13 @@
             android:name="org.commcare.activities.FormEntryActivity"
             android:launchMode="singleTop"
             android:windowSoftInputMode="adjustResize" />
+        <activity
+            android:name="org.commcare.activities.PushNotificationLaunchActivity"
+            android:theme="@android:style/Theme.NoDisplay"
+            android:taskAffinity=""
+            android:excludeFromRecents="true"
+            android:launchMode="singleInstance"
+            android:exported="false" />
```
Jignesh-dimagi: nit: default is false only, can be removed

app/src/org/commcare/utils/FirebaseMessagingUtil.java
OrangeAndGreen: (warning: not an area I know a lot about personally)

My understanding is that we need to embed the request code in the action to make the pending intent appear unique to Android's Intent.filterEquals function. That function doesn't look at extras when determining whether two pending intents are unique, so without this extra code the system could see two pending intents from notifications and mistakenly combine them. The result for the user would be clicking on Notification A but navigating to the destination for Notification B.

Note similar existing behavior can be found in [showNotification](https://github.com/dimagi/commcare-android/blob/ed81450acba5615de8aeb7bf1da0951eb586f331/app/src/org/commcare/utils/FirebaseMessagingUtil.java#L595)

app/src/org/commcare/activities/FormEntryActivity.java
OrangeAndGreen: First noting that this line has changed and we now just show the user a toast in this scenario.

But regarding the mPendingNavAfterSave, I think this is still necessary in order to determine whether the user actually chose to save and navigate or just stay on the page. I.e.:
* User clicks push notification while in form entry, chooses STAY IN FORM when prompted
* User completes and submits the form

The EXTRA_PENDING_NAV_INTENT from the launching intent would still be present at this point, so the app would navigate to the notification target when the user completes the form. By specifically storing the "committed" intent only when the user chooses SAVE INCOMPLETE, we have a reliable way to determine whether to navigate once the save completes.

app/src/org/commcare/activities/components/FormEntryDialogs.java
OrangeAndGreen: Ah, yeah that works: [8360c8a](https://github.com/dimagi/commcare-android/pull/3707/commits/8360c8a3f0a7f72589a4b42ff4bd184a0202e958)

app/AndroidManifest.xml
OrangeAndGreen: [52d3198](https://github.com/dimagi/commcare-android/pull/3707/commits/52d3198038cd8179b51dfb90cf8c8bbcdf481066)

app/src/org/commcare/activities/FormEntryActivity.java
OrangeAndGreen: Showing the toast instead: [e4d2156](https://github.com/dimagi/commcare-android/pull/3707/commits/e4d215664b2f8d3aa3a9ddda359f94852efa4800)

app/src/org/commcare/activities/FormEntryActivity.java
OrangeAndGreen: Changed this to just show a toast to the user when this scenario occurs and do nothing (also logging a non-fatal so we can be aware of it): [e4d2156](https://github.com/dimagi/commcare-android/pull/3707/commits/e4d215664b2f8d3aa3a9ddda359f94852efa4800)

app/src/org/commcare/activities/FormEntryActivity.java
Jignesh-dimagi: Yeah I was thinking to remove intent when user has chosen to stay on page.

I was just trying to avoid variables and instead use intent directly.

app/src/org/commcare/activities/PushNotificationLaunchActivity.kt
OrangeAndGreen: Open to additional opinions on this but I'm leaning toward leaving it out of this PR, since the new activity is currently no-UI and showing dialogs (like biometric unlock or an error message) wouldn't fit well. It also seems like the wrong place for that logic, like this activity should just make the routing decision and let ConnectActivity/ConnectMessagingActivity/etc handle security.

There's also a bit of an ordering concern, especially if the user is in form entry, i.e.:
* User receives and clicks on push notification while in form entry
* They see the form exit dialog and make a decision
* If the decision leads to proceeding with the navigation (drop form or save incomplete), THEN the user is prompted to unlock PersonalId before accessing the target destination


In the scenario above, if the user chooses to STAY IN FORM they're never prompted to unlock PersonalId, which makes sense to me. But if the PushNotificationLaunchActivity required the unlock before proceeding, the user would have to unlock before being shown the dialog that allows them to abort the navigation and stay in the form.

OrangeAndGreen:
@shubham1g5 Note the description is updated with verbatim dialog options and demos for all 3 dialog choices

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -151,6 +154,7 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
     private static volatile boolean isFormEntryActive = false;
 
     private boolean mIncompleteEnabled = true;
+    private Intent mPendingNavAfterSave;
```
conroy-ricketts, line 157: Out of curiosity, why prefix this with `m`?

RELEASES.md
conroy-ricketts, line 1: Flagging the "Don't exit", "Don't save", and "Keep changes" verbiage here as well. QA may be confused as to what exactly these correlate to

conroy-ricketts: Also, the QA notes read "Quit form?" instead of "Exit Form?"

app/src/org/commcare/activities/FormEntryActivity.java
OrangeAndGreen: That's an often-used convention for indicating private variables. I saw Claude chose that name for this variable, but when looking through other variables in the class it seemed like the general pattern in this file so I decided to leave the name and be consistent within the file.

conroy-ricketts: Ah makes sense!

RELEASES.md
OrangeAndGreen: [76497d9](https://github.com/dimagi/commcare-android/pull/3707/commits/76497d9d16acc64a72f8377d2e0925aa43c2884a)

RELEASES.md
```diff
@@ -60,6 +61,18 @@ we would like to communicate to QA as part of the release testing
 - Verify that the combobox widget is working as expected when selecting an item that is used to filter another combobox widget and also determines the visibility of some other unrelated question whose relevance condition depends on the selection.
 - Open the Connect notification history screen (the list of push notifications) and verify the screen title reads "Notifications" (or the localized PersonalID notification title) and that no secondary breadcrumb/title strip is shown above or below it. Confirm the back arrow and the cloud-sync menu action both still work, and that opening a notification still routes correctly.
 
+- **Form exit warning on push notification tap:**
+  - **Editable form, dialog appears:** Open any Connect app and enter a form. Trigger a navigation push notification (e.g. a Connect message, payment notification, or any `ccc_*` notification that opens a screen). Tap the notification.
+    - Verify a "Quit form?" dialog appears with three choices: "STAY IN FORM", "EXIT WITHOUT SAVING", and "SAVE INCOMPLETE" (the same dialog the back button shows).
```
conroy-ricketts: Optional nit (since this is in quotes and doesn't exactly match the dialog):

```suggestion
    - Verify a "Exit Form?" dialog appears with three choices: "STAY IN FORM", "EXIT WITHOUT SAVING", and "SAVE INCOMPLETE" (the same dialog the back button shows).
```

conroy-ricketts (approved):

app/src/org/commcare/activities/PushNotificationLaunchActivity.kt
Jignesh-dimagi: Yes, I agree that it should be in a separate task. We can decide the logic for PersonalId login and biometrics at that time.

app/src/org/commcare/utils/FirebaseMessagingUtil.java
Jignesh-dimagi: This is not mandatory, but it's a defensive guard against mixing cached pending intents. With an introduction to this, cache is maintained per notification.

nit (optional): This can be abstracted in a single method `resolveNotificationKey(fcmMessageData)` and can be used in buildNotification and showNotification methods.

app/src/org/commcare/activities/FormEntryActivity.java
shubham1g5: Do we still need this call here given we are now just showing the toast and do not want to re-direct user on save as complete ?

app/src/org/commcare/utils/FirebaseMessagingUtil.java
shubham1g5: > Note similar existing behavior can be found in [showNotification](https://github.com/dimagi/commcare-android/blob/ed81450acba5615de8aeb7bf1da0951eb586f331/app/src/org/commcare/utils/FirebaseMessagingUtil.java#L595)

Is that the right link, I don't see where we are embedding a notification ID into an Intent action. I am quite down for using an unique notification ID when showing a notification using `notifiy` but am perplexed about using it in the intent action.

OrangeAndGreen: Ah, my mistake, the linked code is just using the same logic for retrieving the notification ID.

My understanding is that we need something distinguishable in the PendingIntent to make the caching work correctly if there are two similar notifications in the tray. That being said, if I remove this code I'm not able to reproduce the error that would supposedly occur. Although I'm also not sure exactly how it would be triggered, I just tested sending two messages to a user and verifying they navigate correctly. Also worth noting that with this code in place everything works as expected.

Would you prefer that we just use the FCM_NOTIFICATION_ID for the action?

Here's Claude's explanation for why we should set a unique action:

>   Why also set a unique action on the launchIntent?
> 
>   launchIntent.setAction("org.commcare.notifications.tap." + requestCode);
>   PendingIntent contentIntent = PendingIntent.getActivity(context, requestCode, launchIntent, flags);
> 
>   Android caches PendingIntents by (package, requestCode, Intent.filterEquals). filterEquals only compares action / data / type / component / categories — it ignores extras. The real navigation
>    target here lives entirely in an extra (EXTRA_WRAPPED_NAV_INTENT, set at line 555–557), so two launchIntents for different pushes look identical to filterEquals.
> 
>   With FLAG_UPDATE_CURRENT, if two PendingIntents collide under that key, Android keeps a single PendingIntent and just refreshes its extras. The risk this guards against: a stale cached
>   PendingIntent firing with the wrong wrapped nav intent — i.e., tapping notification A launches the destination of notification B.
> 
>   Embedding the requestCode in the action makes each launchIntent distinct under filterEquals, so the PendingIntent cache key is unique on both axes (requestCode and intent identity). It's
>   defense-in-depth — requestCode alone is technically part of the key, but the unique action ensures the underlying Intent is also distinguishable, which matters for any code path that compares
>    intents directly (e.g., PendingIntent.getActivity(..., FLAG_NO_CREATE) lookups, or system components doing intent matching).
> 
>   Note the fallback case is weaker: when notificationId is empty, every such push reuses requestCode=1005 and action "...tap.1005", so concurrent payload-less pushes will overwrite each other's
>    PendingIntent. That's intentional — without a server-supplied id there's nothing to disambiguate them anyway.

app/src/org/commcare/activities/FormEntryActivity.java
OrangeAndGreen: Oh, definitely! My mistake: [fb4d54f](https://github.com/dimagi/commcare-android/pull/3707/commits/fb4d54f0bf0dbb6b3422143262b7f87ac1ef6961)

app/src/org/commcare/utils/FirebaseMessagingUtil.java
OrangeAndGreen: Looking for references other than Claude:
* [Medium article from 2022](https://piotrmadry.medium.com/pending-intent-caching-on-android-16e892c6307)
* [SO post where both the OP and top answer are setting unique actions](https://stackoverflow.com/questions/3140072/android-keeps-caching-my-intents-extras-how-to-declare-a-pending-intent-that-ke)

app/src/org/commcare/activities/FormEntryActivity.java
shubham1g5: Ahhh I would have thought we still need a call for Incomplete forms save but not for all saves. Does the save as incomplete flow still works after this change ?

app/src/org/commcare/utils/FirebaseMessagingUtil.java
shubham1g5: Thanks for providing detailed reasoning here, that explanation resonates to me and highlights the need of setting an unique action better in light of this particular change. 

One other request, can we validate if it's possible for `TextUtils.isEmpty(notificationId)` to be `true` here, I doubt so given we are not doing any similar checking a few lines above - ` intent.putExtra(NOTIFICATION_ID, fcmMessageData.getPayloadData().get(NOTIFICATION_ID));`

shubham1g5: Actually think you can directly do `launchIntent.setAction("org.commcare.notifications.tap." + notificationId);` , the code to stringify an int notification ID was only earlier there because `notificationManager.notify()` only accept an integer.

OrangeAndGreen: I can remove the empty string check if you prefer although I don't see how it hurts anything as is and still seems like good defensive guarding in case the server doesn't include a "notification_id" field in a message payload (now or in the future).

Regarding whether to put requestCode or notificationId into the action string, I can make that change as well but noting that we still need the int value requestCode to pass in to PendingIntent.getActivity on the next line.

I leave it up to you though, if you prefer I can change the code to this:
```
String notificationId = fcmMessageData.getPayloadData().get(NOTIFICATION_ID);
launchIntent.setAction("org.commcare.notifications.tap." + notificationId);

int requestCode = NotificationIdentifiers.generateNotificationIdFromString(notificationId);
PendingIntent contentIntent = PendingIntent.getActivity(context, requestCode, launchIntent, flags);
```

shubham1g5: > I can remove the empty string check if you prefer although I don't see how it hurts anything as is and still seems like good defensive guarding in case the server doesn't include a "notification_id" field in a message payload (now or in the future).

I am ok with us being defensive here but right now it feels like code is ambiguous around whether we are being defensive or not. I am worried that code  may throw an NPE on `fcmMessageData.getPayloadData().get(NOTIFICATION_ID)` itself if the notification ID is not there in the payload making it meaningless to do a null check later on.  


Also think we are not doing anythng with the `requestCode` here and can simpify all of this to -

````
String notificationId = fcmMessageData.getPayloadData().get(NOTIFICATION_ID);
launchIntent.setAction("org.commcare.notifications.tap." + notificationId);
PendingIntent contentIntent = PendingIntent.getActivity(context, 0, launchIntent, flags);
````

Also tagging this thread as `Optional` given we have resolved the original question of this thread.

OrangeAndGreen: Yeah, [if the key isn't in the map it just returns null](https://docs.oracle.com/javase/6/docs/api/java/util/HashMap.html#get%28java.lang.Object%29), it doesn't crash.

So the call to TextUtils.isEmpty guards against all of these:
* notification_id not included in message payload
* notification_id included but set to null explicitly
* notification_id included but set to empty string

If you're okay with it I'll leave it as is.

app/src/org/commcare/activities/FormEntryActivity.java
OrangeAndGreen: I decided to leave this one as is, with the thinking that the additional variable changes the default behavior to cancelling the navigation, rather than the default being to navigate.

In other words:
* The variable starts null and only gets assigned when the user explicitly makes the choice to leave the page (abandoning the form or saving as incomplete)
* If we use the extra instead, the default behavior will be to follow the pending nav UNLESS code clears the extra from the launching intent

So in any corner-case that needs to exit out of the standard workflow, we'll have to remember to clear the extra. One example of that is [here](https://github.com/dimagi/commcare-android/blob/fb4d54f0bf0dbb6b3422143262b7f87ac1ef6961/app/src/org/commcare/activities/FormEntryActivity.java#L1266), but there could be more in the future.

Conversely, we aren't likely to add more greenlight paths, and if we do remembering to add the line that grabs the EXTRA_PENDING_NAV_INTENT into mPendingNavAfterSave seems easier (more in the dev's current context) than in the other scenario.

PR #3708, opened 2026-05-12
CCCT-2376 Add Email Parsing And Email OTP API Scaffolding

Jignesh-dimagi:
## Technical Summary

Lays down the backend plumbing for the upcoming PersonalID email step (parent feature: adding an optional email entry / OTP verification screen after Backup Code, gated by the server-driven `email_otp_verification` toggle). No UI is wired up in this PR — the new screens land in follow-up PRs.

What changes:

- **Session data:** `PersonalIdSessionData` gains an `email: String?` field, an `emailSkippedDuringSignup: Boolean` flag, and an `isEmailOtpVerificationToggleActive()` helper that scans the existing `featureReleaseToggles` list for the `email_otp_verification` slug. The plan explicitly designates `email != null` as the in-session source of truth for "verified" — there is intentionally no separate `emailVerified` field.
- **Parser:** `StartConfigurationResponseParser` now reads the new optional top-level `email` returned by `/users/start_configuration`. The value is treated as already-verified, but is only stored when it survives a `JsonExtensions.optNonBlankStringSafe(...)` + `StringUtils.isValidEmail(...)` check. Missing key, JSON-null, empty, whitespace-only, and malformed addresses all collapse to `null` so downstream consumers can rely on the "non-null means verified" contract.
- **Helpers:** `JsonExtensions.optNonBlankStringSafe(key)` (Kotlin) and `StringUtils.isValidEmail(String)` (Java, backed by `android.util.Patterns.EMAIL_ADDRESS`). The latter is unused for now and will be called by the email-entry fragment in a follow-up PR.
- **API layer:** new `/users/send_email_otp` and `/users/verify_email_otp` endpoints in `ApiEndPoints`, matching Retrofit methods in `ApiService`, static wrappers in `ApiPersonalId`, and handler methods (`sendEmailOtpCall` / `verifyEmailOtpCall`) in `PersonalIdApiHandler`. Both handler methods use `NoParsingResponseParser` since the server responses for these endpoints carry no body data we need to surface.
- **Error handling:** new `RATE_LIMITED_ERROR` value in `BaseApiHandler.PersonalIdOrConnectApiErrorCodes`, a `"RATE_LIMITED"` case in `PersonalIdApiHandler.handleErrorCodeIfPresent`, and a mapping in `PersonalIdOrConnectApiErrorHandler.handle(...)` to the existing `R.string.personalid_too_many_otp_attempts` string. This covers the HTTP 429 response the server returns when a user requests too many email OTPs in a short window.

### Safety story

**What gives confidence:**
- The change is purely additive at the API/parsing layer. No existing code paths were rewritten; new methods are simply new entry points that nothing in the app calls yet.
- The new top-level `email` field in the start-configuration response is gated through `optNonBlankStringSafe(...)` plus `StringUtils.isValidEmail(...)`, so any unexpected server payload (null, empty, whitespace, malformed) collapses to `null` rather than propagating.
- Automated coverage was extended in the same PR: the parser test now exercises six branches (missing key, JSON null, empty string, whitespace, malformed, well-formed) and a new `StringUtilsTest` covers the email validator.
- The `RATE_LIMITED_ERROR` enum value is only reachable via the new `"RATE_LIMITED"` code path; existing error-handling switches retain their original behavior.

**Risks to review:**
- The new `sendEmailOtpCall` / `verifyEmailOtpCall` methods and the email parsing path are unreachable from any UI in this PR, so any latent bug in this code will not surface until the follow-up email fragments land.
- The plan called out using the private `createCallback(PersonalIdSessionData, PersonalIdApiResponseParser)` overload, but `NoParsingResponseParser` is a `BaseApiResponseParser`, so the new handler methods invoke the parent `BaseApiHandler.createCallback` overload instead. This matches the existing pattern used for `heartbeatRequest` and `updateChannelConsent`, but it does mean `sessionData` is not auto-populated by the callback for these two calls — the future email fragments will need to mutate `sessionData.email` themselves on success (the plan already restricts session-data writes to the verification fragment, so this is the expected design).
- The new `personalid_too_many_otp_attempts` string is reused from existing copy; if product wants different wording for the email-OTP rate-limit case specifically, this mapping will need to change.

### Automated test coverage

- `StartConfigurationResponseParserTest`: `testParseSetsEmailWhenServerReturnsIt`, `testParseLeavesEmailNullWhenServerOmitsEmail`, `testParseLeavesEmailNullWhenServerReturnsBlank`, `testParseLeavesEmailNullWhenServerReturnsEmptyString`, `testParseLeavesEmailNullWhenServerReturnsMalformedAddress`, `testParseLeavesEmailNullWhenServerReturnsJsonNull`.
- `StringUtilsTest` (new): well-formed addresses, surrounding whitespace, `null`, empty/blank, and malformed forms (`not-an-email`, `missing@domain`, `@no-local.com`, `spaces in@example.com`).
- No integration / instrumentation coverage was added for the new API endpoints in this PR — they're unreachable until the email fragments exist.

[CCCT-2376]: https://dimagi.atlassian.net/browse/CCCT-2376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/unit-tests/src/org/commcare/utils/StringUtilsTest.kt
```diff
@@ -0,0 +1,43 @@
...
+    @Test
+    fun isValidEmail_returnsFalseForEmptyOrBlank() {
+        assertFalse(StringUtils.isValidEmail(""))
+        assertFalse(StringUtils.isValidEmail("   "))
+    }
+
+    @Test
+    fun isValidEmail_returnsFalseForMalformedAddress() {
+        assertFalse(StringUtils.isValidEmail("not-an-email"))
+        assertFalse(StringUtils.isValidEmail("missing@domain"))
```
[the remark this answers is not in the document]
Jignesh-dimagi: I just ran the test and found out that it didn't fail. Please review on your side.

app/src/org/commcare/connect/network/base/BaseApiHandler.kt
```diff
@@ -104,7 +104,10 @@ abstract class BaseApiHandler<T>(
 
         // The photo uploaded for the user is too large for server to handle. Note that, currently,
         // the user does not have much control over the photo size.
-        FILE_TOO_LARGE_ERROR, ;
+        FILE_TOO_LARGE_ERROR,
+
+        // The user requested too many OTPs in a short window (HTTP 429 / RATE_LIMITED).
+        RATE_LIMITED_ERROR, ;
```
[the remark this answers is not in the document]
Jignesh-dimagi: This will compile, its basically style which allowed by kotlin grammar

app/unit-tests/src/org/commcare/utils/StringUtilsTest.kt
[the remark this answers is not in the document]
Jignesh-dimagi: Mark as resolved

RELEASES.md
```diff
@@ -55,6 +55,7 @@ we would like to communicate to QA as part of the release testing
   - **Email entry screen:** (will be added here)
   - **Email OTP screen:** (will be added here)
   - **Legacy logged-in users prompt:** (will be added here)
+  - **API/parsing scaffolding (CCCT-2376):** this change only adds backend plumbing — the new email entry / OTP screens are not yet reachable. For this slice specifically, QA should confirm that **PersonalID signup and recovery continue to work exactly as they did in the previous build** (no new screens appear, no unexpected errors, the user lands on the same destination at the end of each flow). Any regression in those flows would indicate a problem with the new start_configuration parsing or error-code handling.
```
shubham1g5: this seems quite un-relevant for QA given it doesn't have any user visible component, I would prefer removing this completely.

app/src/org/commcare/android/database/connect/models/PersonalIdSessionData.kt
```diff
@@ -60,5 +64,14 @@ data class PersonalIdSessionData(
...
+        const val EMAIL_OTP_VERIFICATION_SLUG = "email_otp_verification"
     }
+
+    /**
+     * True when the `email_otp_verification` slug is present in the parsed
+     * featureReleaseToggles list AND its `active` flag is true. Returns false when the
+     * slug is missing OR explicitly false.
+     */
+    fun isEmailOtpVerificationToggleActive(): Boolean =
+        featureReleaseToggles?.firstOrNull { it.slug == EMAIL_OTP_VERIFICATION_SLUG }?.active == true
```
shubham1g5: thoughts on having a helper class that hosts any checks like this ? I would avoid having a lots of these methods in this class as the list of toggles grow.

app/src/org/commcare/android/database/connect/models/PersonalIdSessionData.kt
```diff
@@ -33,6 +33,10 @@ data class PersonalIdSessionData(
     var attemptsLeft: Int? = null,
     // name of the user
     var userName: String? = null,
+    // email of the user; non-null means the address has been verified server-side or via OTP this session
+    var email: String? = null,
+    // tracks when the user explicitly declines the email offer during signup
+    var emailSkippedDuringSignup: Boolean = false,
```
shubham1g5: what will this be used for ?

app/src/org/commcare/connect/network/PersonalIdOrConnectApiErrorHandler.java
```diff
@@ -63,6 +63,8 @@ public static String handle(
                 return context.getString(R.string.recovery_network_unknown);
             case INCORRECT_OTP_ERROR:
                 return context.getString(R.string.personalid_incorrect_otp);
+            case RATE_LIMITED_ERROR:
```
shubham1g5: how is this different from existing `RATE_LIMIT_EXCEEDED_ERROR` ? we should re-use the existing error codes here.

app/src/org/commcare/android/database/connect/models/PersonalIdSessionData.kt
```diff
@@ -60,5 +64,14 @@ data class PersonalIdSessionData(
...
         const val PIN = "pin"
         const val BIOMETRIC_TYPE = "biometric"
+        const val EMAIL_OTP_VERIFICATION_SLUG = "email_otp_verification"
     }
+
+    /**
+     * True when the `email_otp_verification` slug is present in the parsed
+     * featureReleaseToggles list AND its `active` flag is true. Returns false when the
+     * slug is missing OR explicitly false.
+     */
```
conroy-ricketts: Nit: this comment seems redundant because reading the function is self-explanatory 

```suggestion
```

app/src/org/commcare/connect/network/base/BaseApiHandler.kt
```diff
@@ -104,7 +104,10 @@ abstract class BaseApiHandler<T>(
 
         // The photo uploaded for the user is too large for server to handle. Note that, currently,
         // the user does not have much control over the photo size.
-        FILE_TOO_LARGE_ERROR, ;
+        FILE_TOO_LARGE_ERROR,
+
+        // The user requested too many OTPs in a short window (HTTP 429 / RATE_LIMITED).
```
conroy-ricketts: Nit: will this error only apply to OTPs related to email verification? If so, can we specify that here to prevent confusion between that and phone verification

```suggestion
        // The user requested too many OTPs for email verification in a short window (HTTP 429 / RATE_LIMITED).
```

app/src/org/commcare/connect/network/connectId/parser/StartConfigurationResponseParser.java
```diff
@@ -30,6 +31,11 @@ public void parse(JSONObject json, PersonalIdSessionData sessionData) throws JSO
         sessionData.setSessionFailureSubcode(JsonExtensions.optStringSafe(json, "failure_subcode", null));
         sessionData.setOtpFallback(json.optBoolean("otp_fallback", false));
 
+        // Email is returned ONLY when the server already has a verified address for this user.
+        // Ignore blank/malformed values so downstream consumers can rely on `email != null` meaning verified.
+        String email = JsonExtensions.optNonBlankStringSafe(json, "email");
+        sessionData.setEmail(StringUtils.isValidEmail(email) ? email : null);
```
conroy-ricketts: Is the `isValidEmail()` check necessary here? If the email from Server has already been verified for the user (indicated by`email != null`), then it should already be a valid email

```suggestion
        sessionData.setEmail(JsonExtensions.optNonBlankStringSafe(json, "email"));
```

app/src/org/commcare/connect/network/connectId/PersonalIdApiHandler.java
```diff
@@ -329,6 +332,26 @@ public void validateOtp(Activity activity, String otp, PersonalIdSessionData ses
         );
     }
 
+    public void sendEmailOtpCall(Activity activity, String email, PersonalIdSessionData sessionData) {
```
conroy-ricketts: Nit: can remove "Call" from the function name

```suggestion
    public void sendEmailOtp(Activity activity, String email, PersonalIdSessionData sessionData) {
```

app/src/org/commcare/connect/network/connectId/PersonalIdApiHandler.java
```diff
@@ -329,6 +332,26 @@ public void validateOtp(Activity activity, String otp, PersonalIdSessionData ses
...
+        ApiPersonalId.sendEmailOtp(
+                activity,
+                email,
+                sessionData.getToken(),
+                createCallback(new NoParsingResponseParser<>(), null)
+        );
+    }
+
+    public void verifyEmailOtpCall(Activity activity, String email, String otp,
```
conroy-ricketts: Nit: can remove "Call" from the function name

```suggestion
    public void verifyEmailOtp(Activity activity, String email, String otp,
```

app/src/org/commcare/connect/network/connectId/PersonalIdApiHandler.java
```diff
@@ -329,6 +332,26 @@ public void validateOtp(Activity activity, String otp, PersonalIdSessionData ses
         );
     }
```
conroy-ricketts: Nit: thoughts on refactoring these to add the word "Phone" to the function names now that we distinguish them from email verification?

```suggestion
public void sendPhoneOtp(Activity activity, PersonalIdSessionData sessionData) {
ApiPersonalId.sendPhoneOtp(
activity,
sessionData.getToken(),
createCallback(sessionData, null)
);
}

public void validatePhoneOtp(Activity activity, String otp, PersonalIdSessionData sessionData) {
ApiPersonalId.validatePhoneOtp(
activity,
sessionData.getToken(),
otp,
createCallback(sessionData, null)
);
}
```

app/src/org/commcare/connect/network/connectId/parser/StartConfigurationResponseParser.java
OrangeAndGreen: Empty string seems reasonable but I think a malformed email address should be considered a bug and reported (I'd say via non-fatal, not crashing)

app/unit-tests/src/org/commcare/utils/StringUtilsTest.kt
conroy-ricketts, line 1: Since this is a new test class, we should have the test names here follow our [guidelines](https://github.com/dimagi/open-source/blob/master/docs/mobile_standards.md#testing-criteria) to utilize backticks to write test names as natural-language sentences

app/src/org/commcare/connect/network/connectId/parser/StartConfigurationResponseParser.java
OrangeAndGreen: Ah, but the empty string case isn't possible from the previous line, so we should only expect null or a valid email and anything else should be reported as an error.

app/unit-tests/src/org/commcare/utils/StringUtilsTest.kt
```diff
@@ -0,0 +1,43 @@
...
+    @Test
+    fun isValidEmail_returnsFalseForNull() {
+        assertFalse(StringUtils.isValidEmail(null))
+    }
+
+    @Test
+    fun isValidEmail_returnsFalseForEmptyOrBlank() {
+        assertFalse(StringUtils.isValidEmail(""))
+        assertFalse(StringUtils.isValidEmail("   "))
+    }
```
conroy-ricketts, line 34: Nit: can we make this two separate tests, one for `empty` and another for `blank`, following the single responsibility principle

RELEASES.md
```diff
@@ -61,6 +55,7 @@ we would like to communicate to QA as part of the release testing
...
```
OrangeAndGreen: I think we're avoiding putting ticket numbers in the release/QA notes

conroy-ricketts: Should we remove this? When this feature is fully complete and QA is ready to test `2.64`, will this still be relevant?

conroy-ricketts (changes requested):
Looks great! Just had some minor comments

OrangeAndGreen (changes requested):

app/src/org/commcare/connect/network/connectId/parser/StartConfigurationResponseParser.java
shubham1g5: Should we log a non-fatal if we receive a non-valid user email from server ?

shubham1g5 (changes requested):

app/src/org/commcare/connect/network/base/BaseApiHandler.kt
Jignesh-dimagi: Actually, it was discussed while reviewing the web design doc and decided to keep it generic only.

app/src/org/commcare/android/database/connect/models/PersonalIdSessionData.kt
Jignesh-dimagi: It is required to show the dialog prompt to the user to enter an email address, after 30 days, if the user has skipped during sign-up.

If this flag is true, it means that the user has been presented with email but has skipped it. Treating the signup screen as the first offer, the dialog prompt to enter email should appear only after 30 days later.

In general, this flag will be used to update the `PersonalIDPreferences` accordingly at the end of the sign-up / recovery flow. This will be implemented in upcoming PRs.

shubham1g5: can't we use the `email` field itself, if `email` is present in Connect User record at the end of workflow, we know that email verification has happened already and otherwise not.

app/src/org/commcare/android/database/connect/models/PersonalIdSessionData.kt
Jignesh-dimagi: Yeah, I am always in favour of helper classes for clean implementation. This has only one slug right now, so I didn't create it. Maybe we should whenever it has more than one?

But I agree and have in mind to create the helper for reading toggles from the DB. 

Open to your suggestions.

RELEASES.md
Jignesh-dimagi: Done. This is the issue with `create-mobile-pr` SKILL, it doesn't allow developer to read RELEASES.md but commit directly. I think [PR](https://github.com/dimagi/dimagi-claude-workflows/pull/38) should address this issue.

app/src/org/commcare/android/database/connect/models/PersonalIdSessionData.kt
shubham1g5: think best to create one now given we know we will need it soon.

app/src/org/commcare/connect/network/connectId/parser/StartConfigurationResponseParser.java
Jignesh-dimagi: @conroy-ricketts This is required in case of any server bug.

@shubham1g5 Agree, added [here](https://github.com/dimagi/commcare-android/pull/3708/changes/ee24149f4c0bfcbb8e1ec01d1109a36d315d6607)

app/src/org/commcare/connect/network/connectId/PersonalIdApiHandler.java
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3708/changes/ee24149f4c0bfcbb8e1ec01d1109a36d315d6607)

app/src/org/commcare/connect/network/connectId/PersonalIdApiHandler.java
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3708/changes/ee24149f4c0bfcbb8e1ec01d1109a36d315d6607)

app/src/org/commcare/connect/network/connectId/PersonalIdApiHandler.java
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3708/changes/ee24149f4c0bfcbb8e1ec01d1109a36d315d6607)

app/unit-tests/src/org/commcare/utils/StringUtilsTest.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3708/changes/ee24149f4c0bfcbb8e1ec01d1109a36d315d6607)

RELEASES.md
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3708/changes/ee24149f4c0bfcbb8e1ec01d1109a36d315d6607)

app/src/org/commcare/connect/network/connectId/parser/StartConfigurationResponseParser.java
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3708/changes/ee24149f4c0bfcbb8e1ec01d1109a36d315d6607)

app/src/org/commcare/connect/network/PersonalIdOrConnectApiErrorHandler.java
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3708/changes/ee24149f4c0bfcbb8e1ec01d1109a36d315d6607)

RELEASES.md
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3708/changes/ee24149f4c0bfcbb8e1ec01d1109a36d315d6607)

app/unit-tests/src/org/commcare/utils/StringUtilsTest.kt
Jignesh-dimagi: I think it's fine, as it's testing all cases for white spaces, the same as malformed email cases.

app/src/org/commcare/android/database/connect/models/PersonalIdSessionData.kt
Jignesh-dimagi: Yeah, I was initially thinking the same, but email can be blank due to the toggle also, so this extra flag gives confirmation that email has been shown but the user skipped it.

app/src/org/commcare/android/database/connect/models/PersonalIdSessionData.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3708/changes/bdc3208acefbd2f20ac1d78662fe5d9c9474a60d)

app/src/org/commcare/android/database/connect/models/PersonalIdSessionData.kt
Jignesh-dimagi: @shubham1g5 Removed as offlined with you to not consider signup prompt as email offer count but always prompting 2 times when there is not an email in the account

app/unit-tests/src/org/commcare/connect/ReleaseToggleHelperTest.kt
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    @Test
+    fun `isToggleActive with context returns false when DB returns null`() {
+        every { ConnectAppDatabaseUtil.getReleaseToggles(context) } returns null
+
+        assertFalse(ReleaseToggleHelper.isToggleActive(context, "feature_a"))
+    }
+
+    // --- isEmailOtpVerificationActive(sessionData) -----------------------------------------
```
shubham1g5: I would remove email specific methods here and only test for generic methods, otherwise we are adding 10 tests for each toggle which seems like a nightmare.

Jignesh-dimagi: Makes sense; I was somewhat biased to not have these tests specific to each toggle but kept them as coverage might be an issue someday, but removed now.

app/src/org/commcare/connect/network/connectId/parser/StartConfigurationResponseParser.java
conroy-ricketts: @Jignesh-dimagi Can we change the code comment here? It is very misleading. If there is a server bug, then the email was not verified correctly.

e.g. this:

```
// The email returned SHOULD have already been verified by Server.
```

instead of this:

```
// Email is returned ONLY when the server already has a verified address for this user.
```

Jignesh-dimagi: I think it is a correct comment, as the 'email' key will be present only when the server already has a verified email address for this user; otherwise, the key will be absent.

app/src/org/commcare/connect/network/base/BaseApiHandler.kt
conroy-ricketts, line 1: Nit: we can revert all changes to this file

app/src/org/commcare/connect/ReleaseToggleHelper.kt
```diff
@@ -0,0 +1,30 @@
...
+
+    fun isToggleActive(
+        sessionData: PersonalIdSessionData?,
+        slug: String,
+    ): Boolean = evaluate(sessionData?.featureReleaseToggles, slug)
+
+    fun isToggleActive(
+        context: Context,
+        slug: String,
+    ): Boolean = evaluate(ConnectAppDatabaseUtil.getReleaseToggles(context), slug)
```
conroy-ricketts, line 19: Nit: these two functions can either be private or just marked visible for testing

Jignesh-dimagi: It has been kept public if any usage directly required it by just passing the toggle slug.

app/src/org/commcare/connect/network/base/BaseApiHandler.kt
Jignesh-dimagi: What do you mean here?

app/src/org/commcare/connect/network/connectId/parser/StartConfigurationResponseParser.java
conroy-ricketts: I'm getting a bit confused here. Will there be a scenario in the future where there is a sever bug where the email address is verified _**and**_ not valid? That seems weird

shubham1g5 (approved):

app/src/org/commcare/connect/ReleaseToggleHelper.kt
conroy-ricketts: That sounds good to me, but then why have separate `isEmailOtpVerificationActive()` functions when callers can just use `isToggleActive()` with the slug?

We could either continue adding more functions here for readability, or alternatively, we could remove `isEmailOtpVerificationActive()` and avoid extending this object in the future for other release toggles.

conroy-ricketts: Either way, this is definitely just a nit for me

app/src/org/commcare/connect/network/base/BaseApiHandler.kt
conroy-ricketts: With the latest commits, these changes are not functionally different from `master` and just adds a blank line

conroy-ricketts (approved):
Updating my review to non-blocking!

app/src/org/commcare/connect/network/connectId/PersonalIdApiHandler.java
```diff
@@ -312,23 +315,43 @@ public void retrieveWorkHistory(Context context, String userId, String password)
...
+        ApiPersonalId.sendPhoneOtp(
                 activity,
                 sessionData.getToken(),
                 createCallback(sessionData, null)
         );
     }
 
-    public void validateOtp(Activity activity, String otp, PersonalIdSessionData sessionData) {
-        ApiPersonalId.validateOtp(
+    public void validatePhoneOtp(Activity activity, String otp, PersonalIdSessionData sessionData) {
```
OrangeAndGreen, line 326: Nit: Noting that the corresponding function for email is called verifyEmailOtp. We should name these consistently, and personally I like "verify" over "validate"

app/src/org/commcare/connect/ReleaseToggleHelper.kt
```diff
@@ -0,0 +1,30 @@
...
+
+    private fun evaluate(
```
OrangeAndGreen, line 21: Nit: "evaluate" seems unclear as to what this function does, maybe "checkForActiveToggle" or just "isToggleActiveInList"?

OrangeAndGreen (approved):

PR #3709, opened 2026-05-12
Shift PR comment handler to run once a day

shubham1g5:
I was able to pin-point this workflow being the driver of the token usage and burning around 8million tokens per day, I am therefore shifting the schedule here to be once a day instead from 6 times a day. 

This also reduces the github toolset used for the job, I want to see if that helps reducing tokens used by this job.

### DIFF
--- a/.github/workflows/pr-comment-handler.lock.yml
+++ b/.github/workflows/pr-comment-handler.lock.yml
@@ -30,7 +30,7 @@
 # - Skips ambiguous or contested comments rather than guessing
 # Always transparent, never merges PRs, defers to human judgement on contested changes.
 #
-# gh-aw-metadata: {"schema_version":"v2","frontmatter_hash":"8d2433b41cad2962ad79c96ad6461d068bfe47d86619f866447e78f84212f310","compiler_version":"v0.57.2","strict":true}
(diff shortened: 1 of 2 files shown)


Jignesh-dimagi:
@shubham1g5  The changes look good. 
Just curious, how did you find RCA that this is the issue for consuming the 8 million tokens a day? Does Claude provide any interface which can be used to find such things?
Also, if this runs every day, roughly 13 million tokens will still be consumed. Are we ok with that?

shubham1g5:
> Does Claude provide any interface which can be used to find such things?

Right, we use a different account for Github workflows which charge us based on tokens and one can see token usage in that account. 

> Also, if this runs every day, roughly 13 million tokens will still be consumed. Are we ok with that?

yeah for now (that roughly comes to 1 usd per run in terms of costing), although I do want to continue experimenting and see if we are able to reduce the token usage further. Also did another change here to see whether that results into any optimizations - https://github.com/dimagi/commcare-android/pull/3709/commits/5059920133d8e31dc759a25171e5f2f0c7d84a4a.

Jignesh-dimagi (approved):

PR #3711, opened 2026-05-12
CCCT-2164 Add Plan To Decouple Login From Connect App Launches

conroy-ricketts:
## Technical Summary

Creates the Claude plan for removing the Login Screen for Connect App Launches.

The plan covers extracting the login work currently fused to `LoginActivity` into a headless `LoginController` so the Connect-initiated app launch path can skip both `LoginActivity` and `SeatAppActivity`. It breaks the work into five dependency-ordered phases (one JIRA ticket each, with Phase 5 deferred pending the planned Connect redesign), defines new classes and wrappers, enumerates failure modes and their UX, and lists out-of-scope items.

No code changes — this PR is purely the planning document.

### Safety story

**What gives confidence:**
- Diff is a single new Markdown file under `docs/superpowers/plans/` — zero code paths affected, zero runtime behavior changes.
- I reviewed the plan end-to-end and cross-checked its file/line references against the current branch to make sure the citations point at real code.

**Risks to review:**
- The plan describes a multi-PR refactor that touches a high-risk auth surface (`LoginActivity`, `DispatchActivity`, Connect launch). The implementation PRs that follow this plan will need careful review — the design decisions captured here are worth scrutiny now while it's cheap to change.
- The plan assumes some current behaviors. Anything subtly wrong here only surfaces during implementation, not during plan review.

[CCCT-2164]: https://dimagi.atlassian.net/browse/CCCT-2164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
```diff
@@ -0,0 +1,363 @@
...
+
+**Acceptance:** Every call above runs on a successful login in both the `LoginActivity` and silent paths. Covered by a table-driven unit test.
+
+### C9. AsyncTask suspend wrappers
+
+Required by C1 (suspend controller).
+
+`KeyRecordOperations` and `SyncOperations` (`org.commcare.login`) wrap `ManageKeyRecordTask` / `DataPullTask` in `suspendCancellableCoroutine { ... }` with `invokeOnCancellation { task.cancel(true) }`.
+
+**Cancellation is best-effort.** The wrapper does not await task termination. The underlying task may continue executing in the background and complete normally (writing local state) even after the coroutine is cancelled. Document this in `docs/connect/connect-app-launcher.md` — it's acceptable for analytics fidelity but worth knowing.
```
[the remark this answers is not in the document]
conroy-ricketts: [Claude Code AI] Partially valid. The plan already declares this "acceptable for analytics fidelity" deliberately. Mitigation (1) is already implemented by C4 step 1 (closeUserSession() runs before every new silent launch). Mitigation (2) requires modifying ManageKeyRecordTask/DataPullTask, which the "Out Of Scope" section explicitly defers. Mitigation (3) speculates a session-validity check in the routing layer that doesn't currently exist anywhere. I'll add one sentence noting how step 1 absorbs orphans, but skip the three-mitigation framework as scope creep.

435589e855039bc306625590487aff4148933430

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
```diff
@@ -0,0 +1,363 @@
...
+    object PersonalIdUnlockNeeded : PostLoginDestination()                                // PersonalIdUnlockRequired
+}
+
+enum class FallbackReason { BadCredentials, KeyRecordMissing, ConnectLinkageInvalid, SandboxCorrupted, AppNotSeated }
+enum class RetryableReason { Sync, Network }
+```
+
+**New extension:** `fun PostLoginDestination.toIntent(context, launchContext): Intent`. Defined only on `Home` and `ConnectOpportunityInfo` — the failure variants have no Intent (the orchestrator emits `ShowError`; the activity wrapper stays on-screen).
+
+This is the **single source of truth** for the extras `RootMenuHomeActivity` / `StandardHomeActivity` / `ConnectNavHelper` consume. Both the silent path (C4) and `DispatchActivity.onActivityResult` use it. The string extras themselves are unchanged — those activities are out of scope to migrate.
```
OrangeAndGreen: This isn't clear to me. What is Claude considering out of scope?

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
```diff
@@ -0,0 +1,363 @@
...
+
+`LoginActivity` itself stays — manual launches, MDM auto-login, shortcut launches, and `needAnotherAppLogin()` re-prompts all still need it. We're only short-circuiting the *Connect-initiated* path.
+
+## Required Changes
+
+Each change below is a discrete commit-sized unit, listed in dependency order.
+
+### C1. Extract a headless `LoginController`
+
+The prerequisite for everything else: a non-UI caller must be able to perform a login.
```
OrangeAndGreen: Non-UI? Isn't it more a matter of getting additional UIs to support login functionality? Maybe it doesn't matter if this plan gets around it anyway, but I wonder if it would simplify the design if we defined an interface (like ILoginHost) that login-able UIs could implement

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
```diff
@@ -0,0 +1,363 @@
...
+5. Translate the result via `PostLoginRouter` → `PostLoginDestination`:
+   - `Home` / `ConnectOpportunityInfo` → `toIntent(...)` → `SilentLaunchOutcome.StartActivity`.
+   - `FallbackToLoginActivity(appId, reason)` → `ShowError` with `OpenLoginActivityManualMode`. Covers `BadCredentials`, `KeyRecordMissing`, `ConnectLinkageInvalid`, `SandboxCorrupted`, `AppNotSeated` — all "user must see the full login screen" cases.
+   - `RetryableFailure` → `ShowError` with `RetryAction.RelaunchSame`. Covers `SyncFailed` / `NetworkUnavailable`.
+   - `PersonalIdUnlockNeeded` → invoke `host.requestUnlock()`. On `Granted`, call `performLogin` exactly one more time; if that retry also returns `PersonalIdUnlockNeeded`, emit `ShowError` (with retry) and do **not** call `host.requestUnlock()` again. On `Cancelled` / `Failed` / `HostUnavailable`, emit `ShowError` (with retry) immediately.
+
+`CancellationException` propagates from `launch`; the caller's scope handles it. The orchestrator does not emit UI itself — it returns an outcome.
+
+**Decision: always escalate on `BadCredentials`.** There is no rotate-password API to retry against, so silent rotation isn't an option.
```
OrangeAndGreen: I haven't heard much mention of local-login vs remote via the API, and this line is a bit concerning in that respect. If the local login fails it should try the remote login, is that Claude's plan?

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
```diff
@@ -0,0 +1,363 @@
...
+sealed class ManualEscalation {
+    data class OpenLoginActivityManualMode(val appId: String, val reason: FallbackReason) : ManualEscalation()
+}
+```
+
+**Orchestration order:**
+
+1. Close the current session.
+2. Seat the app via `AppSeater.seatIfNeeded(appId, progressSink)` — caller renders a modal progress dialog (no `SeatAppActivity`).
+3. Resolve credentials via `ConnectCredentialResolver.resolve(appId, createIfNeeded = true)`. If null → `ShowError` with `OpenLoginActivityManualMode(appId, KeyRecordMissing)`.
```
OrangeAndGreen: Uh-oh, but we aren't defining ConnectCredentialResolver until C7...

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
```diff
@@ -0,0 +1,363 @@
...
+4. Call `LoginController.performLogin(request)`.
+5. Translate the result via `PostLoginRouter` → `PostLoginDestination`:
+   - `Home` / `ConnectOpportunityInfo` → `toIntent(...)` → `SilentLaunchOutcome.StartActivity`.
+   - `FallbackToLoginActivity(appId, reason)` → `ShowError` with `OpenLoginActivityManualMode`. Covers `BadCredentials`, `KeyRecordMissing`, `ConnectLinkageInvalid`, `SandboxCorrupted`, `AppNotSeated` — all "user must see the full login screen" cases.
```
OrangeAndGreen: Some of these don't feel right to me. For example, KeyRecordMissing shouldn't be an outright failure should it? Instead the code should try to retrieve the key record and proceed.

The AppNotSeated error also feels odd to encounter here as the app should definitely be seated along the way if necessary. But maybe this is to handle error situations, like where the app failed to seat?

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
```diff
@@ -0,0 +1,363 @@
...
+   - `PersonalIdUnlockNeeded` → invoke `host.requestUnlock()`. On `Granted`, call `performLogin` exactly one more time; if that retry also returns `PersonalIdUnlockNeeded`, emit `ShowError` (with retry) and do **not** call `host.requestUnlock()` again. On `Cancelled` / `Failed` / `HostUnavailable`, emit `ShowError` (with retry) immediately.
+
+`CancellationException` propagates from `launch`; the caller's scope handles it. The orchestrator does not emit UI itself — it returns an outcome.
+
+**Decision: always escalate on `BadCredentials`.** There is no rotate-password API to retry against, so silent rotation isn't an option.
+
+**Caller wiring in `ConnectJobsListsFragment`:**
+
+- Launch from `viewLifecycleOwner.lifecycleScope` — backgrounding cancels cleanly; the user re-taps on return.
+- Disable all opportunity rows while a launch is in flight (prevents double-launch; UI-level lock is the only lock).
```
OrangeAndGreen: Noting a bit of a short-term/long-term issue here.

Short-term, I think there are several places that a learn/deliver app could be launched from, not just the opportunity list. For example, for an opp that is completed the user is taken to the Delivery Progress page first and must click the Resume button on that page to download/launch the delivery app. So for now, locking down the UI will be a page-by-page task ("opportunity rows" only applies to one page).

Long-term (like next few months?), the Connect UX redesign is going to change these mechanics and which pages are capable of launching learn/deliver apps. I think it at that point may all happen through the (new) Opportunity Home page.

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
```diff
@@ -0,0 +1,363 @@
...
+        is ShowError -> renderSnackbar(outcome.message, outcome.manualEscalation, outcome.retry)
+    }
+} finally {
+    requireActivity().requestedOrientation = previousOrientation
+    restoreRowsEnabled()
+    dismissSeatProgressDialog()
+}
+```
+
+- On `OpenLoginActivityManualMode`, the snackbar shows "Enter password manually". Tapping it calls:
```
OrangeAndGreen: I don't think we want to show a snackbar message here

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
```diff
@@ -0,0 +1,363 @@
...
+CommCareLauncher.launchCommCareForAppId(
+    activity,
+    appId,
+    mapOf(IS_LAUNCH_FROM_CONNECT to true, SKIP_AUTO_LOGIN to true),
+)
+```
+
+The existing `LoginActivity → DispatchActivity → ConnectNavHelper.goToActiveInfoForJob` route lands the user back on Connect opp info once they enter credentials.
+
+- The seat-progress dialog is a **non-dismissable Material alert** with spinner + text. The fragment shows it when seating begins and dismisses on every termination path.
```
OrangeAndGreen: Re: "Material alert": Sounds like Claude intends to build a new dialog for this, but we could probably stick with our common CommCare dialog

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
```diff
@@ -0,0 +1,363 @@
...
+
+`ConnectAppUtils.launchApp(...)` delegates to `ConnectAppLauncher`. `CommCareLauncher.launchCommCareForAppId(...)` and `IS_LAUNCH_FROM_CONNECT` remain — but the only live post-migration use is the manual-escalation re-entry above.
+
+**Acceptance:** Tapping an opportunity for a healthy account never shows `LoginActivity` or `SeatAppActivity`. Time-to-home drops by one Activity transition. Manual flow (app drawer → CommCare → login screen) is unaffected.
+
+### C5. Add `SKIP_AUTO_LOGIN` gate in `LoginActivity` for manual-escalation entry
```
OrangeAndGreen: I'm having some trouble understanding the whole manual-escalation re-entry concept. Maybe we could add a "Proposed Flow" section to the top with a condensed overview of how the new code would go.

I'm a little worried this may be suggesting a flow like:
* Try to auto-login to learn/deliver app
* If something goes wrong, dump the user on Login and ask for credentials

That's only slightly worse than what the app does today, which is to stay on the Login page but keep the credential fields hidden. Either way, if the scenario above is encountered we should make sure our reaction is to show an error message to the user, not suggest they try a user/pass (they won't have one).

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
```diff
@@ -0,0 +1,363 @@
...
+  - Pre-populate username from the Connect linked-app record.
+  - Leave the password field blank (pre-populating with the known-bad password just re-fails).
+  - Optionally surface a small inline hint based on the `FallbackReason`.
+- Post-login routing is unchanged — `IS_LAUNCH_FROM_CONNECT=true` still drives the route to Connect opp info.
+
+**Acceptance:** Launching with both `IS_LAUNCH_FROM_CONNECT=true` and `SKIP_AUTO_LOGIN=true` shows the login form (username pre-filled, password blank), does not invoke `initiateLoginAttempt()` on resume, and routes back to Connect opp info on manual submission.
+
+### C6. Inline `AppSeater` (skip `SeatAppActivity` on silent path)
+
+Removes the second Activity transition. Independent of `LoginActivity` work but required by C4.
```
OrangeAndGreen: Required by C4? I may be missing the point of the ordering for these tasks...

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
```diff
@@ -0,0 +1,363 @@
...
+
+### C9. AsyncTask suspend wrappers
+
+Required by C1 (suspend controller).
+
+`KeyRecordOperations` and `SyncOperations` (`org.commcare.login`) wrap `ManageKeyRecordTask` / `DataPullTask` in `suspendCancellableCoroutine { ... }` with `invokeOnCancellation { task.cancel(true) }`.
+
+**Cancellation is best-effort.** The wrapper does not await task termination. The underlying task may continue executing in the background and complete normally (writing local state) even after the coroutine is cancelled. An orphaned session written this way is absorbed by step 1 of C4's orchestration — `closeUserSession()` runs before every new silent launch, so the next launch starts from a clean slate. Document this in `docs/connect/connect-app-launcher.md` — it's acceptable for analytics fidelity but worth knowing.
+
+Progress reported via a `(Progress) -> Unit` lambda. `LoginActivity` supplies a real callback; `ConnectAppLauncher` supplies a no-op (the seat modal is dismissed before sync begins).
```
OrangeAndGreen: Does this mean "silently launched" apps won't report progress while they're attempting to auto-login? I think we'll still have UI to show progress, so may need to plumb it through ConnectAppLauncher

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
OrangeAndGreen: Ah, second bullet [here](https://github.com/dimagi/commcare-android/blob/435589e855039bc306625590487aff4148933430/docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md#out-of-scope) helped

OrangeAndGreen (commented):
This is a big one! I'll start by saying that it was a bit hard to piece it all together and understand the purpose of all the proposed components and changes (see comment about adding a Proposed Flow section).

I initially really liked the idea that the work could be separated into smaller functional chunks like C1, C2, etc. If that leads to a ticket for each chunk I think it would really help when it comes time to review all these changes! But as I went through the document that logic doesn't seem to hold up, and early sections are relying on things that wouldn't be implemented until later sections. Maybe we can get Claude to fix that? Otherwise maybe we can put some thought into how we could break things up well.

One final note is that it wasn't clear to me exactly how to approach a review like this. In the end I think I mostly just put up notes for consideration, or spots where I think the design could be either wrong or not completely fleshed out. All in all, I'd say these are more topics of discussion then "requested changes".

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
OrangeAndGreen: After finishing the doc, I get a sense that the "silent login" Claude has in mind is more silent than we actually want. We're still going to want progress reporting along the way (during seating, login, key record retrieval, data pull, etc.)

conroy-ricketts: I interpreted "non-UI" here as meaning this new `LoginController` is never actually shown to the user (hopefully I understood your question correctly).

If I understand correctly, the idea you're suggesting is to implement a progress spinner dialog while all the login related stuff happens in the background for the user? That seems like a good idea to me (we could have that shown to the user while on the opportunities list page).

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: Yup, I figured that refactoring those activities to use `PostLoginDestination` directly should be out of scope for this. Let me know if I should edit this line to be more clear about that for others.

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: This is a great catch. Think I should rewrite/reorder this plan into larger "phases" that we can make tickets out of, so this will change very soon.

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: "KeyRecordMissing" might be a bit misleading, it's more like a "LinkedAppRecordMissing". What triggers this is `ConnectCredentialResolver` returning `null`. I'll rename this.

I think that "AppNotSeated" might be misleading too. You're correct - the idea here is that this error would happen when the app fails to seat. I'll rename this to something like "AppSeatFailed".

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: Yeah, the idea here is that the way this currently works will be preserved (i.e. we try a local login first, and fall back to a remote login if the local login was not successful).

`BadCredentials` would only show after both the local and remote login attempts were unsuccessful.

This line of the plan specifically is saying that there currently is no option to ask HQ for a new password (to retry with) if the password was wrong. This functionality does not exist at the moment.

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: Good point!! I'll tweak this.

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: Yeah that's true, I agree. Maybe an error dialog would be much better here? We can tweak the dialog message based on the reason. For example:
- `BadCredentials` -> "We couldn't sign you in. Please try again."
- `LinkedAppRecordMissing` -> "Your PersonalID credentials aren't available right now."
- `AppSeatFailed` -> "The app couldn't be set up. Check your connection and try again."
- `SyncFailed / NetworkUnavailable` -> "The server couldn't be reached. Please try again and ensure you have an internet connection."
- `ConnectLinkageInvalid / SandboxCorrupted` -> "A reset is required to continue. Please forget your PersonalID user and try again."

Thoughts?

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: Agreed, will tweak this.

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: Ooh that's true. When iterating on this plan I wasn't thinking about how user's never see their password and can't type it on retries. Think I'll have to rework this plan substantially. Great catch!

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: Yup this was an oversight. This will change soon.

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: Yup, addressed this in my reply to another comment above

conroy-ricketts (commented):
> One final note is that it wasn't clear to me exactly how to approach a review like this. In the end I think I mostly just put up notes for consideration, or spots where I think the design could be either wrong or not completely fleshed out. All in all, I'd say these are more topics of discussion then "requested changes".

@OrangeAndGreen My thought was that we should approach these "Claude plan" reviews as if we were reviewing a normal tech spec. So I think your perspective is correct here! But I'm definitely open to thoughts from the rest of the team

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
OrangeAndGreen: Yeah, not just a generic wait dialog but something that can provide extra details in addition to generally letting the user know that an action is in progress. Today a large modal dialog appears and initially says "Logging in", but if a data pull is required (i.e. on initial login) it shows a series of additional messages to the user indicating the status of the pull. Data pulls can take a while if there's a lot to download so it becomes important to show the user more detailed status (that way they have an idea how much longer they'll be waiting).

I think in some of our Connect UX redesign work we've discussed showing the progress spinner and status text directly on the page instead of using a modal dialog. However, we might need to lockdown the UI during this process so the user can't abort (the behavior today). @shubham1g5 Thoughts on this (modal vs. on-page)?

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
OrangeAndGreen: Yeah, although a few notes:
* Some of these errors are the kind of thing that should be invisible to the user and just "never break". If they occur it probably means we messed something up, and in that case we probably need to show a more significant error and take steps we think will get things back up and running. I think this might be the case for:
  * BadCredentials (if we mean the credentials that PersonalId had stored for the login are no longer working)
  * LinkedAppRecordMissing (either a non-error if it just isn't configured, or something that should never happen since we shouldn't have let the user try to login without a linked app record)
  * ConnectLinkageInvalid
  * As you mentioned for the last one, I think in all these cases we probably need to tell the user something went wrong with PersonalId, then lead them to recover their account
* AppSeatFailed and SandboxCorrupted may have good error-handling in place already which we should keep (like navigating to an app recovery workflow)

conroy-ricketts:
@OrangeAndGreen Please let me know if I missed anything! The plan went through some major revisions

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
```diff
@@ -0,0 +1,269 @@
...
+   - **Damaged-app routing** — for seat failures. `CommCareApplication.initializeAppResources()` already marks the app `STATE_CORRUPTED` and reports the exception via `ForceCloseLogger` / `CrashUtil`. Preserve [`DispatchActivity.handleDamagedApp()`](https://github.com/dimagi/commcare-android/blob/684512e2662996c9855e35af306da3e311e56c80/app/src/org/commcare/activities/DispatchActivity.java#L292) by starting `DispatchActivity` with `FLAG_ACTIVITY_CLEAR_TOP`; its existing `STATE_CORRUPTED` check at [`DispatchActivity.java:211`](https://github.com/dimagi/commcare-android/blob/684512e2662996c9855e35af306da3e311e56c80/app/src/org/commcare/activities/DispatchActivity.java#L211) routes to `RecoveryActivity`.
```
shubham1g5, line 27: Does this routing happens today on Login Screen ?

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
```diff
@@ -0,0 +1,269 @@
...
+   - **Retry / Cancel error dialog** — for transient sync or network failures.
+
+## Current Flow
+
+1. [`ConnectJobsListsFragment.launchAppForJob()`](https://github.com/dimagi/commcare-android/blob/684512e2662996c9855e35af306da3e311e56c80/app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java#L154) → [`ConnectAppUtils.launchApp()`](https://github.com/dimagi/commcare-android/blob/684512e2662996c9855e35af306da3e311e56c80/app/src/org/commcare/connect/ConnectAppUtils.kt#L117) → [`closeUserSession()`](https://github.com/dimagi/commcare-android/blob/684512e2662996c9855e35af306da3e311e56c80/app/src/org/commcare/CommCareApplication.java#L390) → [`CommCareLauncher.launchCommCareForAppId(...)`](https://github.com/dimagi/commcare-android/blob/684512e2662996c9855e35af306da3e311e56c80/commcare-support-library/src/main/java/org/commcare/commcaresupportlibrary/CommCareLauncher.java#L20) with [`IS_LAUNCH_FROM_CONNECT=true`](https://github.com/dimagi/commcare-android/blob/684512e2662996c9855e35af306da3e311e56c80/app/src/org/commcare/connect/ConnectAppUtils.kt#L18).
```
shubham1g5, line 32: Just flagging that we launch App for multiple pages today including learn and delivery progress page and in general we should assume it to be something that we might want to do from multiple connect pages.

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
```diff
@@ -0,0 +1,269 @@
...
+
+  `LoginController` is unlock-agnostic: callers are expected to have already unlocked PersonalID before invoking `performLogin`. The Connect silent path satisfies this via the [`ConnectActivity.onCreate`](https://github.com/dimagi/commcare-android/blob/684512e2662996c9855e35af306da3e311e56c80/app/src/org/commcare/activities/connect/ConnectActivity.java#L73) gate. `LoginActivity`'s manual PersonalID-managed flow continues to call [`PersonalIdUnlocker.INSTANCE.unlock(...)`](https://github.com/dimagi/commcare-android/blob/684512e2662996c9855e35af306da3e311e56c80/app/src/org/commcare/activities/LoginActivity.java#L243) before invoking the controller — unchanged from today.
+
+- **`LoginRequest`** — `appId`, `username`, `password`, `mode` (`Manual`, `AutoFromConnect`, `MdmManaged`, `Demo`), `restoreSession`.
+
+- **`LoginProgressSink`** — single-method interface: `fun onProgress(progress: LoginProgress)`. `LoginProgress` carries a phase tag (`Seating`, `SigningIn`, `Syncing`) plus an optional percentage and message. Phase 1's `LoginActivity` refactor wires this to the activity's existing progress UI; Phase 3's `ConnectAppLauncher` wires it to a `CustomProgressDialog`.
```
shubham1g5, line 74: For `LoginActivity`, confirming that this would not cause any UI changes for dialog progress ? Given the `seating` happens before user presses login on Login Screen.

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
```diff
@@ -0,0 +1,269 @@
...
+
+- **`LoginResult`** — sealed class. Variants carry the raw data the routing layer needs; they do not produce destinations directly (`PostLoginRouter` does that in Phase 2):
+  - `Success(loginMode: LoginMode, restoreSession: Boolean, manualSwitchToPwMode: Boolean, personalIdManagedLogin: Boolean)`
+  - `BadCredentials` — both local and remote auth failed
+  - `LinkedAppRecordMissing` — `ConnectCredentialResolver` returned null
+  - `ConnectLinkageInvalid` — auth token unavailable or denied; PersonalID linkage is stale
+  - `SandboxCorrupted(requiresIntervention: Boolean)` — `requiresIntervention` is `true` for `PullTaskResult.BAD_DATA_REQUIRES_INTERVENTION`, `false` for `BAD_DATA`. Phase 3's fragment uses the flag to pick the `StockMessages` variant
+  - `SyncFailed(reason)`
+  - `NetworkUnavailable`
```
shubham1g5: Leaving as a suggestion but fine to skip - should we encapsulate all of these in a `Failed(error: LoginError)` where `Error` can be another sealed class ? I am thinking about various classes that will implement login and would love if all of them don't have to handle all of the errror pathways in switch case blocks.

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
```diff
@@ -0,0 +1,269 @@
...
+
+- **`LoginProgressSink`** — single-method interface: `fun onProgress(progress: LoginProgress)`. `LoginProgress` carries a phase tag (`Seating`, `SigningIn`, `Syncing`) plus an optional percentage and message. Phase 1's `LoginActivity` refactor wires this to the activity's existing progress UI; Phase 3's `ConnectAppLauncher` wires it to a `CustomProgressDialog`.
+
+- **`LoginResult`** — sealed class. Variants carry the raw data the routing layer needs; they do not produce destinations directly (`PostLoginRouter` does that in Phase 2):
+  - `Success(loginMode: LoginMode, restoreSession: Boolean, manualSwitchToPwMode: Boolean, personalIdManagedLogin: Boolean)`
+  - `BadCredentials` — both local and remote auth failed
+  - `LinkedAppRecordMissing` — `ConnectCredentialResolver` returned null
+  - `ConnectLinkageInvalid` — auth token unavailable or denied; PersonalID linkage is stale
+  - `SandboxCorrupted(requiresIntervention: Boolean)` — `requiresIntervention` is `true` for `PullTaskResult.BAD_DATA_REQUIRES_INTERVENTION`, `false` for `BAD_DATA`. Phase 3's fragment uses the flag to pick the `StockMessages` variant
```
shubham1g5: Don't think `BAD_DATA` means `SandboxCorrupted` but it's more of a parsing error for sync response. Think we can map it to `SyncFailed` as well.

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
```diff
@@ -0,0 +1,269 @@
...
+  - `NetworkUnavailable`
+
+  **Mapping from existing task outcomes** (use this verbatim — implementers should not invent new mappings). Source enums: [`HttpCalloutOutcomes`](https://github.com/dimagi/commcare-android/blob/684512e2662996c9855e35af306da3e311e56c80/app/src/org/commcare/network/HttpCalloutTask.java#L37) (from [`ManageKeyRecordTask`](https://github.com/dimagi/commcare-android/blob/684512e2662996c9855e35af306da3e311e56c80/app/src/org/commcare/tasks/ManageKeyRecordTask.java)) and [`PullTaskResult`](https://github.com/dimagi/commcare-android/blob/684512e2662996c9855e35af306da3e311e56c80/app/src/org/commcare/tasks/DataPullTask.java#L716).
+
+  | Source | `LoginResult` |
+  |---|---|
+  | `HttpCalloutOutcomes.Success` (+ pull success) | `Success` |
+  | `HttpCalloutOutcomes.AuthFailed`, `PullTaskResult.AUTH_FAILED` | `BadCredentials` |
+  | `HttpCalloutOutcomes.NetworkFailure`, `NetworkFailureBadPassword`, `CaptivePortal` | `NetworkUnavailable` |
+  | `HttpCalloutOutcomes.TokenUnavailable`, `TokenRequestDenied`, `IncorrectPin`, `AuthOverHttp` | `ConnectLinkageInvalid` |
```
shubham1g5: `TokenRequestDenied` needs to fall through to the `CommCareExceptionHandler` as today in order to trigger a global errror.

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
```diff
@@ -0,0 +1,269 @@
...
+
+**Goal:** A single authority translates a `LoginResult` into a destination. Seating moves off the activity stack.
+
+**New files** (`org.commcare.login`):
+
+- **`PostLoginRouter`** — pure Kotlin, no Android dependencies.
+  - Input: `LoginResult` + `LaunchContext` (Connect-initiated vs not, target `appId`, `restoreSession`, `jobId`).
+  - Output: `PostLoginDestination`:
+    - `Home(loginMode, startFromLogin, manualSwitchToPwMode, personalIdManagedLogin)`
+    - `ConnectOpportunityInfo(jobId, loginMode)`
```
shubham1g5: Reg. `ConnectOpportunityInfo `, Is this a new UX change we are introducing or an existing workflow ? If later, can we reference what existing routing in code it's supposed to replace.

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
```diff
@@ -0,0 +1,269 @@
...
+   - Any failure → `Failed(<corresponding reason>)`.
+7. Release the in-flight lock in `finally`.
+
+`CancellationException` propagates from `launch`; the caller's scope handles it.
+
+**Caller wiring — [`ConnectJobsListsFragment`](https://github.com/dimagi/commcare-android/blob/684512e2662996c9855e35af306da3e311e56c80/app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java):**
+
+- Launch from `viewLifecycleOwner.lifecycleScope`. Backgrounding cancels cleanly; user re-taps on return.
+- Show [`CustomProgressDialog`](https://github.com/dimagi/commcare-android/blob/684512e2662996c9855e35af306da3e311e56c80/app/src/org/commcare/views/dialogs/CustomProgressDialog.java) before calling `launch()`. The fragment provides a `LoginProgressSink` that updates the dialog's message text.
+- Lock orientation on entry, restore on exit via `try/finally`. The lock guards in-flight coroutine state across configuration changes.
```
shubham1g5: This might not be necessary given the `ConnectActivity` itself locks orientation on a permanent basis

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
```diff
@@ -0,0 +1,269 @@
...
+- Routing per outcome:
+
+  | Outcome | Action |
+  |---|---|
+  | `StartActivity(intent)` | `startActivity(intent)`, dismiss dialog |
```
shubham1g5: is startActivity being an outcome correct here ? (agree with it being Action)

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
```diff
@@ -0,0 +1,269 @@
...
+  | `Failed(BadCredentials)` | Log via `Logger` + Firebase (reason `"connect_login_bad_credentials"`); show a `StandardAlertDialog` titled "Sign in again" with body "Your saved PersonalID credentials are no longer valid. Sign in again from the app sidebar to recover your account." Primary button "Sign Out" calls [`personalIdManager.forgetUser("connect_login_bad_credentials")`](https://github.com/dimagi/commcare-android/blob/684512e2662996c9855e35af306da3e311e56c80/app/src/org/commcare/connect/PersonalIdManager.java#L165) and then calls `requireActivity().finish()` to exit `ConnectActivity`. The user lands on the screen they entered Connect from; they re-enter PersonalID via the sidebar's "Sign-in/Reconfigure" option, which launches `PersonalIdActivity` → backup-code recovery |
+  | `Failed(LinkedAppRecordMissing)` | Same as `BadCredentials`, reason `"connect_login_linked_record_missing"`. Should not occur in normal use |
+  | `Failed(ConnectLinkageInvalid)` | Same as `BadCredentials`, reason `"connect_login_linkage_invalid"` |
```
shubham1g5: For these three, I would not expect a scenario here where things fail due to Personal ID login /linkage failures which is not either a network failure OR a token denied scenario which is handled through `CommCareExceptionHandler` more centrally in the app. 

If it's a network error, I think we should just show a network error message -something similar to ->  `Unable to login due to network issues, please try again`  and if it's token denied or a more concrete global error, we should let it fall through to `CommCareExceptionHandler`. 

I would also just map any network errors directly to `Failed(NetworkUnavailable)`

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
```diff
@@ -0,0 +1,269 @@
...
+- No biometric / PIN prompt appears at any point in the silent launch.
+- `BadCredentials` / `LinkedAppRecordMissing` / `ConnectLinkageInvalid`: dialog text matches the spec above; "Sign Out" triggers `forgetUser` then `finish()` of `ConnectActivity`; user is dropped back to the previous screen with no Connect entry on the stack.
+- `AppSeatFailed`: `STATE_CORRUPTED` is set on the `ApplicationRecord`, `DispatchActivity` is launched, `RecoveryActivity` appears.
+- `SandboxCorrupted(requiresIntervention = true)` shows the `Remote_BadRestoreRequiresIntervention` stock message; `requiresIntervention = false` shows `Remote_BadRestore`.
+- Backgrounding mid-launch cancels cleanly; re-tap on return starts fresh.
+- JVM unit tests: happy path, each failure branch, in-flight lock rejection, both `SandboxCorrupted` variants, `Demo` mode rejected.
+- Robolectric tests: dialog show/dismiss on every termination path; orientation lock/restore on every termination path; `forgetUser` + `finish()` chain on the forget-PersonalID path; `DispatchActivity` start intent on the seat-failure path.
+
+### Phase 4 — Roll out to remaining entry points
```
shubham1g5: think we can move out the `ConnectJobIntroFragment` and `ConnectDownloadingFragment` into another phase 5 that we don't prioritize at the moment. This is because we might be getting rid of these screens in the planned redesign.

shubham1g5 (dismissed):
Looks great overall, left some comments around error handling that might need another deeper look.

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: This damaged-app routing does happen today, but not on the `LoginActivity`. The chain is more like `SeatAppActivity` → `DispatchActivity` → `RecoveryActivity`.

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: Yes I believe so because `LoginActivity` would only emit `SigningIn` and `Syncing` in this scenario, and seating would happen outside of the `LoginController`

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: Sounds like a good idea!

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: Good catch!

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: It's an existing workflow - I'll definitely make this more clear

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: Yeah that seems right, I'll revise this

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: Yes, these are two different things here, but this raises a good point on readability (the outcome is `SilentLaunchOutcome.StartActivity(intent)` and the action is a `startActivity(intent)` function call). I'll tweak this.

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: That's a good point, I think I should make a larger revision here

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: That sounds good to me

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: That's a good point, I'll see how I can revise this to use `SyncFailed` without diverting the existing UX

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: [7dfed51](https://github.com/dimagi/commcare-android/pull/3711/changes/7dfed51ec87479697aaf3c94e829b80363929915)

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: baf2b528afbfdb9ab83d7f3db944e6086bf309ca

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: 6120dc4e92ea5093310b23f5e79ea0fe918dbbbb

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: cadffeea7b67e01eb7b08ae4823ad446e7307444

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: e2a1918a25caeea42cec55e40b98742976c767ef

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: 749aab9e55f416b98c8b7c863a8652a2aa9d5731

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: aba32d8d742212536a22b3303710775d3406d50a

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: 89f752043d324c3ac77b6f502a91fc914e1ac8b6

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
shubham1g5: Ahh I think this workflow is just trying to establish a back behaviour rather than post login redirect i.e. it's not that we are directing to Opportunity info on user login but mostly just making sure when user presses back, they go back to the original source, I think it should not be a `PostLoginDestination` as there is no scenario in which we will not go to Home post login.

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
```diff
@@ -0,0 +1,284 @@
...
+
+  | Source | `LoginResult` |
+  |---|---|
+  | `HttpCalloutOutcomes.Success` (+ pull success) | `Success` |
+  | `HttpCalloutOutcomes.AuthFailed`, `PullTaskResult.AUTH_FAILED` | `BadCredentials` |
+  | `HttpCalloutOutcomes.NetworkFailure`, `NetworkFailureBadPassword`, `CaptivePortal` | `NetworkUnavailable` |
+  | `HttpCalloutOutcomes.TokenRequestDenied` | `TokenDenied` |
+  | `HttpCalloutOutcomes.TokenUnavailable`, `IncorrectPin`, `AuthOverHttp` | `ConnectLinkageInvalid` |
```
shubham1g5: - `IncorrectPin`, `AuthOverHttp` doesn't seem to Connect Linkage issues. `IncorrectPin` signifies user entered an incorrect password and think should be in the same buckets as `BadCredentials`. While `AuthOverHttp` is a programming safety guard for us to not do any auth requests over `http` (and only on `https`). Think it probably should go to it's stand alone bucket. 

Also think we can put `HttpCalloutOutcomes.TokenUnavailable` as `NetworkUnavailable` itself as it just means a network error for the token request.

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
```diff
@@ -0,0 +1,284 @@
...
+
+  **Damaged-state contract.** On `SeatResult.Failed`, `AppSeater` must guarantee that the seated app's resource state is `CommCareApplication.STATE_CORRUPTED`. `initializeAppResources()` already flips it on the failure paths SeatAppActivity sees today, but `AppSeater` should defensively set it via `CommCareApp.setAppResourceState(STATE_CORRUPTED)` inside the catch block as well. Phase 3's seat-failure routing assumes this state is set when it bounces back to `DispatchActivity`.
+
+  `LoginActivity` keeps using `SeatAppActivity` via the existing `seatAppIfNeeded()` — migrating that path is out of scope.
```
shubham1g5: think that's fine as long as we don't duplicate the code of app seating in 2 places i.e. `SeatAppActivity` should still in some way rely on the new `AppSeater`

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
```diff
@@ -0,0 +1,284 @@
...
+  | `Failed(BadCredentials)`, `Failed(LinkedAppRecordMissing)`, `Failed(ConnectLinkageInvalid)` | Log via `Logger` + Firebase with a specific reason string (`"connect_login_bad_credentials"`, `"connect_login_linked_record_missing"`, `"connect_login_linkage_invalid"`), then show the same retry/cancel dialog used for `SyncFailed` / `NetworkUnavailable`. These outcomes are not expected in normal use — `LinkedAppRecordMissing` shouldn't happen, and the realistic credential-side failure modes are either network errors or token denials, both routed through their own outcomes. If a user genuinely gets stuck after retrying, the existing "Forget PersonalID" flow is the manual recovery surface                                                                                                                                                                                                                       |
```
shubham1g5: think both ``Failed(LinkedAppRecordMissing)`, `Failed(ConnectLinkageInvalid)`  signifies grave errors where user should no longer continue to work on this setup. Do you know what happens today when we encounter these scenarios in the app ?

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: Agreed, that makes sense to me

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: True, looking back at it, I think this should _not_ be "out of scope". I'll revise this

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: Ohh I think that's right! I'll update this - Opp Info should not be a destination here.

conroy-ricketts: 36bbc30e1beaa54611ccdf7c2776eb1d83716d67

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: ff124efe3c525947ce32326790a56794915e75c1

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: 55bfc715aff11d310b7d0244c95e4a0da227e856

docs/superpowers/plans/2026-05-11-ccct-2164-decouple-login-from-connect-launch.md
conroy-ricketts: I removed `ConnectLinkageInvalid` from the plan while addressing the other comments above

As for `LinkedAppRecordMissing` (if I understand correctly), what happens today is either one of two scenarios:
1.  No record exists which leads to a random password being generated and stored in a fresh, new record. In this case, the issue is effectively resolved on its own.
2. The app simply crashes.

I'll revise the plan to more closely mirror the existing behavior

conroy-ricketts: I ended up dropping `LinkedAppRecordMissing` from the plan and made `ConnectCredentialResolver` non-null with the possibility of throwing an `IllegalStateException` - a322f8c361c2980c5f63f60b385ffae2a1339b29

shubham1g5 (approved):
It's looking good to me, thanks for the iterations. Think this would be a good time to add implementation tickets on this.

conroy-ricketts:
[Claude Code AI] Closing this PR — the work has been broken into five tickets:

- [CCCT-2437](https://dimagi.atlassian.net/browse/CCCT-2437) — Remove Login Screen Phase 1 — Headless Login Engine
- [CCCT-2438](https://dimagi.atlassian.net/browse/CCCT-2438) — Remove Login Screen Phase 2 — Routing Extraction and Inline Seating
- [CCCT-2439](https://dimagi.atlassian.net/browse/CCCT-2439) — Remove Login Screen Phase 3 — Silent Launch Path (Connect Opportunities List)
- [CCCT-2440](https://dimagi.atlassian.net/browse/CCCT-2440) — Remove Login Screen Phase 4 — Roll Out Silent Launch to Remaining Priority Entry Points
- [CCCT-2441](https://dimagi.atlassian.net/browse/CCCT-2441) — Remove Login Screen Phase 5 — Migrate or Delete Redesign-Bound Entry Points

[CCCT-2437]: https://dimagi.atlassian.net/browse/CCCT-2437?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CCCT-2438]: https://dimagi.atlassian.net/browse/CCCT-2438?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CCCT-2439]: https://dimagi.atlassian.net/browse/CCCT-2439?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CCCT-2440]: https://dimagi.atlassian.net/browse/CCCT-2440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CCCT-2441]: https://dimagi.atlassian.net/browse/CCCT-2441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3713, opened 2026-05-13
Disable Android keystore + bucket log encryption traces separately (perf measurement)

avazirna:
## Product Description 
In a previous work, we switched to encrypting device reports using Android Keystore, with the goal of evaluating its performance impact before potentially applying the same approach to form submissions and large files. While we have now solid data on encryption using the keystore, the baseline data intended for comparison is incomplete due to the 5-attribute limit in Firebase Performance Tracing. This PR reverts to the previous model to allow for a proper baseline data collection and a proper comparison.

## Technical Summary

Disables android keystore-backed encryption to gather Firebase Performance traces to compare encryption durations across two dimensions:

1. **Keystore-backed vs. `SecretKeySpec`-backed encryption paths.**
2. **Device-log encryption (`DeviceReportWriter`) vs. form/data encryption (`SaveToDiskTask` / `EncryptionIO` / `HybridFileBackedSqlStorage`).**

### Safety story
- **Newly-created records** while running this build will store their AES key inline in the device report record's `aesKey` field, instead of being keystore-wrapped.
- **Reads of records created BEFORE this build** continue to work — `DeviceReportRecord.shouldUseKeystoreKey()` is `getKey().length == 0`, so each record retains its own provenance.
- **Reads of records created during this build remain readable** on builds that re-enable the keystore — they're plain AES keys, not keystore handles.
- Each record's encryption path is fixed at creation time, so there's no mid-life decryption inconsistency for records created on this build. This was successfully tested locally. The steps taken were:
  - Airplane mode was enabled
  - Records were created using the current version in the Playstore, 2.62.0
  - CommCare was then updated to this version, and new records created
  - Airplane mode was disabled
  - Forms and Device reports were uploaded and monitored using HTTP Toolkit

## Labels and Review

### DIFF
--- a/app/src/org/commcare/android/security/AndroidKeyStore.kt
+++ b/app/src/org/commcare/android/security/AndroidKeyStore.kt
@@ -22,10 +22,15 @@ object AndroidKeyStore {
 
     fun doesKeyExist(alias: String): Boolean = instance.containsAlias(alias)
 
+    /**
+     * The main purpose of this method is to determine if the Android Keystore is supported on the device. It does
+     * this by trying to access the singleton instance of the KeyStore. It's currently returning false for
+     * comparison purposes between keystore-backed vs SecretKeySpec-backed encryption times.
+     */
     fun isKeystoreAvailable(): Boolean =
         try {
             instance
-            true
+            false
         } catch (e: Exception) {
             Logger.exception("Android keystore is not supported on this device", e)
             false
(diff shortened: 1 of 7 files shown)


shubham1g5 (commented):
@avazirna looks good to me, are you able to revise the safety story here and make sure it's all relevant and also mention any local testing you have done for the same. I would like to ensure that you have tested this fully as it's a last min change we are including in the release after all QA has been completed.

avazirna:
> @avazirna looks good to me, are you able to revise the safety story here and make sure it's all relevant and also mention any local testing you have done for the same. I would like to ensure that you have tested this fully as it's a last min change we are including in the release after all QA has been completed.

@shubham1g5 I've added the steps I took to test this changes locally to the safety story section.

shubham1g5 (approved):
great, thanks for confirming.

PR #3714, opened 2026-05-13
Report attachment-count vs filesystem mismatch at form save

avazirna:
https://dimagi.atlassian.net/browse/SAAS-19082

## Technical Summary

Adds a diagnostic check to compares the in-memory form attachments count against the actual number of media files in the form's instance folder when the form is being saved. If the number of media files is lower than what's expected, a non-fatal exception is logged, surfacing the mismatch in Crashlytics. The save itself is unaffected — this is purely diagnostic.

This is part of a work to identify and diagnose reported missing media files when there is evidence in the form that a media was capture. 

### Safety story
This just adds a diagnostic log, it doesn't block, retry, or alter the save.

## Labels and Review

### DIFF
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -450,6 +450,10 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
         }
     }
 
+    public int getFormAttachmentCount() {
+        return formAttachmentCount;
+    }
+
     public void showFormAttachmentLimitReachedError() {
         String title = StringUtils.getStringRobust(this, R.string.form_attachment_limit_reached_title);
         String msg = StringUtils.getStringRobust(this, R.string.form_attachment_limit_reached,
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -879,6 +879,8 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
             return;
         }
 
+        reportAttachmentCountMismatch();
+        
         if (complete) {
             HiddenPreferences.clearInterruptedFormState();
             HiddenPreferences.clearInterruptedSSD();
@@ -893,6 +895,19 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
         }
         mSaveToDiskTask.setFormSavedListener(this);
         mSaveToDiskTask.executeParallel();
+    }
+
+    private void reportAttachmentCountMismatch() {
+        int numAttachmentsOnDisk = FileUtil.countMediaFiles(FormEntryInstanceState.getInstanceFolder());
+        if (numAttachmentsOnDisk == -1) {
+            return;
+        }
+        if (numAttachmentsOnDisk < formAttachmentCount) {
+            Logger.exception(
+                    "Form attachment count mismatch at save — "
+                            + "in-memory=" + formAttachmentCount + " on-disk=" + numAttachmentsOnDisk,
+                    new Throwable());
+        }
     }
 
     public void discardChangesAndExit() {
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -895,6 +897,19 @@ private void saveDataToDisk(boolean exit, boolean complete, String updatedSaveNa
...
+    private void reportAttachmentCountMismatch() {
+        int numAttachmentsOnDisk = FileUtil.countMediaFiles(FormEntryInstanceState.getInstanceFolder());
+        if (numAttachmentsOnDisk == -1) {
+            return;
+        }
+        if (numAttachmentsOnDisk < formAttachmentCount) {
+            Logger.exception(
+                    "Form attachment count mismatch at save — "
+                            + "in-memory=" + formAttachmentCount + " on-disk=" + numAttachmentsOnDisk,
+                    new Throwable());
```
shubham1g5: +1 to code rabbit comment above, and think the exception should have a message as well.

shubham1g5 (changes requested):
think if you add a `Logger.log` statements every time we increase/decarease attachment count, that would really help with actually tracking the count per addition deletion to provide much more visibility in the issue.

Although that will definitely make the sumo logs bloated, so maybe we can just directly use `Firebase.log` statements in this case.

avazirna:
> think if you add a `Logger.log` statements every time we increase/decarease attachment count, that would really help with actually tracking the count per addition deletion to provide much more visibility in the issue.

From the reports we have received, it seems that this is a rare issue so the goal here is to log whenever a difference is detected to analyze those scenarios. Considering that we are now sampling logs, sending these to Sumologic won't allow for proper debugging the next time this issue occurs.

avazirna:
> Although that will definitely make the sumo logs bloated, so maybe we can just directly use `Firebase.log` statements in this case.

`Filebase.log`? I might not be familiar with this option. Is this to send normal app usage logs to Firebase?

conroy-ricketts (commented):
+1 for coderabbit

shubham1g5:
> Filebase.log? I might not be familiar with this option. Is this to send normal app usage logs to Firebase?

https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/utils/CrashUtil.java#L54 , that's where the logs you see attached to a crash come from.

avazirna:
> > Filebase.log? I might not be familiar with this option. Is this to send normal app usage logs to Firebase?
> 
> https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/utils/CrashUtil.java#L54 , that's where the logs you see attached to a crash come from.

I see what you mean. This way these logs would only show up associated with the non-fatal and not in Sumologic

app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -895,6 +901,19 @@ private void saveDataToDisk(boolean exit, boolean complete, String updatedSaveNa
         mSaveToDiskTask.executeParallel();
     }
 
+    private void reportAttachmentCountMismatch() {
+        int numAttachmentsOnDisk = FileUtil.countMediaFiles(FormEntryInstanceState.getInstanceFolder());
+        if (numAttachmentsOnDisk == -1) {
+            return;
+        }
```
shubham1g5, line 908: should we return here ? If the `formAttachmentCount` is > 0 at this point, that would be a mismatch too ?

shubham1g5 (changes requested):
For thoroughness, are you able to confirm that image save and deletion are working as usual after this change

app/src/org/commcare/activities/FormEntryActivity.java
avazirna: `-1` here indicates that the files in the form instance forder couldn't be read, so the return is to cancel the report given that we are unable to make a comparison.

shubham1g5: makes sense.

shubham1g5 (approved):

avazirna:
> For thoroughness, are you able to confirm that image save and deletion are working as usual after this change

We can wait and see if any of the instrumentation tests will fail. But I also did a test locally and didn't find any issues.

PR #3716, opened 2026-05-15
CCCT-2078 Use Session Unlock Policy For In-App Connect Navigation

shubham1g5:
### [CCCT-2078](https://dimagi.atlassian.net/browse/CCCT-2078)

https://github.com/user-attachments/assets/c71bdac7-997c-4d08-9edc-cd022370698a

We currently do the unlock at multiple different places. We divided all these places into 2 sets - 

1. Places where we always need user to unlock before entering. These are 
 - User logging into a CC app using Personal ID (Login with Personal ID option). Note that any notification redirect into app will only ask for unlock if user is logged out of the app. 
- While Linking/Delinking a traditional CC app to Personal ID login
2. Places Where we only require an unlock if there has not been one in the current app session or in last 10 minutes of the same app session ex. Opportunities, Messaging and Work History menus and notifications redirects to these screens

Also note that we have defined an app session as strictly a foreground session i.e. any time user gets out of the app and resume back into it, we will consider it a different session and will require user to unlock. 

## Technical Summary

This PR refines the `UnlockPolicy` used in `PersonalIdUnlocker.unlock()` calls across the app. Each call site was reviewed and assigned a policy based on whether the unlock is a cold/external entry point (requiring explicit re-authentication every time) or an in-app navigation action (where a session-based bypass is appropriate).

**Decisions per call site:**

- **LoginActivity (Personal ID Login):** `ALWAYS` — primary session authentication gate.
- **PersonalIdManager.unlockAndLinkConnect:** `ALWAYS` — one-time, high-stakes operation that stores credentials.
- **PersonalIdManager.promptToDelinkPersonalIdApp:** `ALWAYS` — irreversible operation; explicit re-auth required regardless of session state.
- **ConnectNavHelper (Messaging / Work History / Connect Jobs List):** Changed to `SESSION_WITH_TIME_THRESHOLD` — these are nav drawer taps from an already-authenticated CommCare session. Prompting on every tap adds friction with minimal security benefit within a 10-minute window.
- **ConnectUnlockFragment (notification redirects):** `SESSION_WITH_TIME_THRESHOLD` -> If a new session, requires unlock otherwise not. Aligns with "ConnectActivity" related navigations which all default to "SESSION_WITH_TIME_THRESHOLD". 
- **ConnectMessagingActivity (notification redirect):** `SESSION_WITH_TIME_THRESHOLD` — same as above. Aligns with policy on Messaging Menu. 

### Safety story

Minimal change — one line in `ConnectNavHelper.kt`. The `SESSION_WITH_TIME_THRESHOLD` path has existing coverage in `PersonalIdUnlockerTest`. All sensitive operations (login, link/unlink, notification entry points) retain `ALWAYS`.

### Automated test coverage

`PersonalIdUnlockerTest` covers the `SESSION_WITH_TIME_THRESHOLD` bypass logic for both within-threshold and expired-threshold cases.

[CCCT-2078]: https://dimagi.atlassian.net/browse/CCCT-2078?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RELEASES.md
```diff
@@ -13,6 +13,7 @@ These are published publicly on Playstore, Github Releases and CommCare Forums
 #### What's New
 
 - [Profile Photo Update] PersonalID users can now update their profile photo directly from the side navigation drawer
+- Reduce frequency of required Biometric or Pin unlocks for PersonalID and Connect
```
conroy-ricketts: Nit: 


```suggestion
- Reduced frequency of required biometric or pin unlocks for PersonalID and Connect  
```

conroy-ricketts (dismissed):

app/src/org/commcare/activities/connect/ConnectMessagingActivity.java
```diff
@@ -135,7 +135,7 @@ private void handleRedirectIfAny() {
                     getNotificationActionFromIntent(getIntent()),
                     getIntent().getStringExtra(NOTIFICATION_ID)
             );
-            PersonalIdUnlocker.INSTANCE.unlock(this, UnlockPolicy.ALWAYS, success -> {
+            PersonalIdUnlocker.INSTANCE.unlock(this, UnlockPolicy.SESSION_WITH_TIME_THRESHOLD, success -> {
```
Jignesh-dimagi, line 138: This is from a notification click, so should it be 'UnlockPolicy.ALWAYS' as coming from external?

app/src/org/commcare/connect/ConnectNavHelper.kt
```diff
@@ -21,7 +21,7 @@ object ConnectNavHelper {
         listener: ConnectActivityCompleteListener,
         navigationAction: (Context) -> Unit,
     ) {
-        PersonalIdUnlocker.unlock(activity, UnlockPolicy.ALWAYS) { success ->
+        PersonalIdUnlocker.unlock(activity, UnlockPolicy.SESSION_WITH_TIME_THRESHOLD) { success ->
```
Jignesh-dimagi: This is a helper function; it might be used from different parts of the code in the future also. Is it not better if this function has 'UnlockPolicy' as a parameter instead of hardcoding here?

app/src/org/commcare/fragments/connect/ConnectUnlockFragment.java
```diff
@@ -63,7 +63,7 @@ public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
     private final Runnable unlockRunnable = new Runnable() {
         @Override
         public void run() {
-            PersonalIdUnlocker.INSTANCE.unlock((CommCareActivity<?>) requireActivity(), UnlockPolicy.ALWAYS, success -> {
+            PersonalIdUnlocker.INSTANCE.unlock((CommCareActivity<?>) requireActivity(), UnlockPolicy.SESSION_WITH_TIME_THRESHOLD, success -> {
```
Jignesh-dimagi, line 66: Same. This can be from a notification click, so it can be 'UnlockPolicy.ALWAYS' as coming from external?

Jignesh-dimagi (changes requested):

app/src/org/commcare/activities/connect/ConnectMessagingActivity.java
shubham1g5: I think it should be same as the underlying component "messaging", given it's a new launch `SESSION_WITH_TIME_THRESHOLD` will default to asking for authentication here the first time user enters in an app session.

app/src/org/commcare/fragments/connect/ConnectUnlockFragment.java
shubham1g5: Given `ConnectActivity` itself is behind `SESSION_WITH_TIME_THRESHOLD`, I don't see a reason why a notification redirect should be higher auth, although it would not matter if user is coming from an external place as in that case even with `SESSION_WITH_TIME_THRESHOLD` user will be prompted for auth.

app/src/org/commcare/connect/ConnectNavHelper.kt
shubham1g5: https://github.com/dimagi/commcare-android/pull/3716/commits/ed9067fd446979ba8757f83e8ce0315221feb36a

RELEASES.md
shubham1g5: https://github.com/dimagi/commcare-android/pull/3716/commits/24094090d851ddd94074c757266400d1e98b23f4

app/src/org/commcare/activities/connect/ConnectMessagingActivity.java
Jignesh-dimagi: Yeah, I agree with you that if a user is coming from an external place and even with 'SESSION_WITH_TIME_THRESHOLD', the user will be prompted for auth with a new launch of the app. But if the app is in the background and the user clicks on the notification, it should re-authenticate, as it's an external entry point. I was thinking of isolating the notification entry as 'always' only. Your thoughts?

shubham1g5: >  But if the app is in the background and the user clicks on the notification, it should re-authenticate

1. That's the behaviour with `SESSION_WITH_TIME_THRESHOLD`  today as well (given app is in background, resuming into it counts as a new session)

2. If Opening Messaging itself (launch App and click on Messaging Menu), doesn't require re-auth, a notificaiton redirect to the same screen must follow the same behaviour as well given in both cases users started outside the app. So the unlock policy here should be same for both launch points.

Jignesh-dimagi: Aha, I see that change; you moved the session reset from onCreate to onStart. That works!

Jignesh-dimagi (dismissed):

RELEASES.md
```diff
@@ -64,9 +65,13 @@ we would like to communicate to QA as part of the release testing
   - In order to achieve this functionality, DB migrations are done to accommodate the new email address field. QA should start testing with the previous version of the app, having PersonalID login already, and then upgrade to this new version. The app should work without crashing.
   - QA should also test with a fresh installation of this new version, going through PersonalID signup/recovery.
 
-
-- Verify that biometric/PIN unlock still triggers correctly in all existing entry points
-
+- **[PersonalID] Session-based unlock for in-app Connect navigation:**
+  - An app session is a foreground app session, i.e. user existing the app and resuming into it counts as a new session.
```
conroy-ricketts: Nit: typo here


```suggestion
  - An app session is a foreground app session, i.e. user exiting the app and resuming into it counts as a new session. 
```

conroy-ricketts (dismissed):

RELEASES.md
```diff
@@ -64,9 +65,13 @@ we would like to communicate to QA as part of the release testing
   - In order to achieve this functionality, DB migrations are done to accommodate the new email address field. QA should start testing with the previous version of the app, having PersonalID login already, and then upgrade to this new version. The app should work without crashing.
   - QA should also test with a fresh installation of this new version, going through PersonalID signup/recovery.
 
-
-- Verify that biometric/PIN unlock still triggers correctly in all existing entry points
-
+- **[PersonalID] Session-based unlock for in-app Connect navigation:**
+  - An app session is a foreground app session, i.e. user exiting the app and resuming into it counts as a new session.
```
OrangeAndGreen: "Exiting the app" seems like it could be unclear to a tester. We could say "sending the app to background" to be more clear.

RELEASES.md
```diff
@@ -64,9 +65,13 @@ we would like to communicate to QA as part of the release testing
...
 
-
-- Verify that biometric/PIN unlock still triggers correctly in all existing entry points
-
+- **[PersonalID] Session-based unlock for in-app Connect navigation:**
+  - An app session is a foreground app session, i.e. user exiting the app and resuming into it counts as a new session. 
+  - Tapping Connect Jobs, Messaging, or Work History from the nav drawer no longer prompts for biometric/PIN if the user already unlocked within the last 10 minutes in the same app session.
+  - Notification redirects to these screens follow same rules as the menu itself. 
+  - Opening any of below destinations from a push notification still always prompts:
+    - Sensitive operations (login, link/unlink app, notification deep links) still require explicit re-authentication every time.
```
OrangeAndGreen: Will QA know what "notification deep links" means and how to test them?

OrangeAndGreen (dismissed):
Just a couple small suggestions for the QA notes to make things clearer. Otherwise, looking forward to fewer unlocks!

conroy-ricketts (approved):

PR #3726, opened 2026-05-19
Unit tests for PersonalId Photo page

OrangeAndGreen:
https://dimagi.atlassian.net/browse/CCCT-2303

## Product Description
No user-facing changes

## Technical Summary
Added 12 unit tests around the photo page:
* 4x test initial state: user name text, two button states, error view hidden
* 1x tests clicking the Take Photo button click
* 2x test success and cancel behavior when taking a photo
* 1x tests Save Photo button click
* 4x test the possible return states when attempting to complete the page

Some additional scope came out of this work, and common code was extracted to the BasePersonalIdConfigurationTest class.
* Helper methods for launching PersonalIdActivity, seeding the session, and navigating to a specific fragment in the nav graph
* Helper method for installing a TestNavHostController to test navigation components
* Helpers for setting up, using, and tearing down a MockWebServer to mock API calls from a fake server

Existing test classes were updated to use the new helpers for the following fragments:
* Phone
* Biometric
* Name
* Email
* Email Verification

### Safety story
All existing tests as well as the 12 new tests pass.

### Automated test coverage
Yup! This PR adds exactly that


app/unit-tests/src/org/commcare/fragments/personalId/BasePersonalIdPhotoCaptureFragmentTest.kt
```diff
@@ -0,0 +1,267 @@
...
+
+        errorHandlerMock = Mockito.mockStatic(PersonalIdOrConnectApiErrorHandler::class.java)
+        errorHandlerMock
+            .`when`<String> {
+                PersonalIdOrConnectApiErrorHandler.handle(
+                    any(),
+                    any(),
+                    Mockito.nullable(Throwable::class.java),
+                )
+            }.thenReturn("Network error occurred")
```
shubham1g5: Any reason we are mocking this ? Seems like this class doesn't require any setup for a test to need a mock for this.

app/unit-tests/src/org/commcare/fragments/personalId/BasePersonalIdPhotoCaptureFragmentTest.kt
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    protected fun setUpPhotoCaptureFragment(sessionData: PersonalIdSessionData = testSessionData) {
```
shubham1g5, line 97: nit: would be nice to generalise this for any fragment and put it in [BasePersonalIdConfigurationTest](https://github.com/dimagi/commcare-android/blob/master/app/unit-tests/src/org/commcare/fragments/personalId/BasePersonalIdConfigurationTest.kt)

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdPhotoCaptureFragmentTest.kt
```diff
@@ -0,0 +1,313 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.ArgumentCaptor
+import org.mockito.Mockito
+import org.robolectric.annotation.Config
+import org.robolectric.shadows.ShadowLooper
+
+@Config(application = CommCareTestApplication::class)
+@RunWith(AndroidJUnit4::class)
+class PersonalIdPhotoCaptureFragmentTest : BasePersonalIdPhotoCaptureFragmentTest() {
```
shubham1g5, line 36: nit: Any new test classes should have names in natural language as per guidance here - https://github.com/dimagi/open-source/pull/22

app/unit-tests/src/org/commcare/fragments/personalId/BasePersonalIdPhotoCaptureFragmentTest.kt
```diff
@@ -0,0 +1,267 @@
...
+/**
+ * Testable subclass of PersonalIdPhotoCaptureFragment that exposes:
+ *  - an injectable NavController (matches PersonalIdBiometricConfigFragment pattern)
+ *  - reflection helpers to drive the private callback methods
+ *  - a helper to replace the takePhotoLauncher with a mock for intent-capture tests
+ *  - a setter for the private photoAsBase64 field
+ */
+class TestablePersonalIdPhotoCaptureFragment(
+    private val testNavController: NavController? = null,
+) : PersonalIdPhotoCaptureFragment() {
```
shubham1g5, line 197: Think a lot of the methods in this class can be mocked using Intent level mocking similar to done [here](https://github.com/dimagi/commcare-android/blob/master/app/unit-tests/src/org/commcare/android/tests/formentry/IdentityCalloutTests.kt#L138), this helps avoid duplicating the callback pattern from the production class.

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdPhotoCaptureFragmentTest.kt
```diff
@@ -0,0 +1,313 @@
...
+            errorView.text = "stale error"
+        }
+        ShadowLooper.idleMainLooper()
+
+        // Act
+        activity.runOnUiThread { saveButton.performClick() }
+        ShadowLooper.idleMainLooper()
+
+        // Verify
+        apiPersonalIdMock.verify {
```
shubham1g5: To expand coverage here,  we can use a `MockWebServer` and verify whether the request was received by the server and that the params to the request is correct.

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdPhotoCaptureFragmentTest.kt
```diff
@@ -0,0 +1,313 @@
...
+        assertEquals(View.GONE, errorView.visibility)
+        assertEquals("", errorView.text.toString())
+    }
+
+    @Test
+    fun testCompleteProfile_onSuccess_storesUserAndNavigatesToSuccess() {
+        // Act
+        activity.runOnUiThread {
+            fragment.invokePhotoUploadSuccess("fake-base64-photo")
+        }
```
shubham1g5: Don't think we should need to do this but it can be natural continuation for the method above `testSavePhotoClick_callsCompleteProfileWithCorrectArgs` if you mock the server itself.

app/unit-tests/src/org/commcare/fragments/personalId/BasePersonalIdPhotoCaptureFragmentTest.kt
```diff
@@ -0,0 +1,267 @@
...
+            activityController
+                .create()
+                .start()
+                .resume()
+                .get()
+
+        // Seed the ViewModel before swapping the fragment in; onCreateView reads it.
+        activity.runOnUiThread {
+            val viewModel = ViewModelProvider(activity)[PersonalIdSessionDataViewModel::class.java]
+            viewModel.setPersonalIdSessionData(sessionData)
```
conroy-ricketts: Nit: can treat this as a property in Kotlin

```suggestion
            viewModel.personalIdSessionData = sessionData
```

app/unit-tests/src/org/commcare/fragments/personalId/BasePersonalIdPhotoCaptureFragmentTest.kt
```diff
@@ -0,0 +1,267 @@
...
+    override fun getNavController(): NavController = testNavController ?: super.getNavController()
+
+    fun replaceTakePhotoLauncher(launcher: ActivityResultLauncher<Intent>) {
+        val field = PersonalIdPhotoCaptureFragment::class.java.getDeclaredField("takePhotoLauncher")
+        field.isAccessible = true
+        field.set(this, launcher)
+    }
+
+    @Suppress("UNCHECKED_CAST")
+    fun getTakePhotoLauncher(): ActivityResultLauncher<Intent> {
```
conroy-ricketts: Is this function still being used?

shubham1g5 (changes requested):
The variations around tests look great but think we can mock the code better to avoid calling the fragment methods and duplicate the network / photo callbacks in tests.

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdPhotoCaptureFragmentTest.kt
conroy-ricketts, line 1: Can we change these test names to use natural language sentences with backticks ( ` )

conroy-ricketts (changes requested):

app/unit-tests/src/org/commcare/fragments/personalId/BasePersonalIdPhotoCaptureFragmentTest.kt
OrangeAndGreen: Ah, I wondered about this one too. The goal was to bypass two aspects of the error handling code:
1. Error messages are localized, so the two tests ([example](https://github.com/dimagi/commcare-android/blob/9396724c790d4d6c4f447ff9f023b1d48fab4e38/app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdPhotoCaptureFragmentTest.kt#L277)) that want to assert on the error message would need to know which error message and translation to assert. By mocking, we ensure that these tests won't break just because the text of an error message gets changed. In other words, these tests only want to verify that an error is shown (along with button states), not the contents of that error.
2. Not an issue now but could be in the future: Some error codes result in crashing behavior, but these tests aren't testing that. Specifically, the TOKEN_INVALID_ERROR used in [this test](https://github.com/dimagi/commcare-android/blob/9396724c790d4d6c4f447ff9f023b1d48fab4e38/app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdPhotoCaptureFragmentTest.kt#L300) cause the error handler to fall through to a default case that will throw an exception if a throwable is included. Right now the calling code doesn't include a throwable, but if the error handler ever changes in that default case it would cause this unrelated test to fail.

All that being said, I'm still getting used to what should and shouldn't be mocked in the name of test stability! The first point seemed reasonable to me particularly, but maybe instead we could just let the natural code run and verify that the returned error string is not empty (rather than asserting on its exact contents)?

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdPhotoCaptureFragmentTest.kt
OrangeAndGreen: [bea4a49](https://github.com/dimagi/commcare-android/pull/3726/commits/bea4a499b69f8c869cc9e5bbe1121ab5eadb1e3c)

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdPhotoCaptureFragmentTest.kt
OrangeAndGreen: [bea4a49](https://github.com/dimagi/commcare-android/pull/3726/commits/bea4a499b69f8c869cc9e5bbe1121ab5eadb1e3c)

app/unit-tests/src/org/commcare/fragments/personalId/BasePersonalIdPhotoCaptureFragmentTest.kt
OrangeAndGreen: Removed: [24047c2](https://github.com/dimagi/commcare-android/pull/3726/commits/24047c2e191e65bafb64a990c5354968ef0f2192)

app/unit-tests/src/org/commcare/fragments/personalId/BasePersonalIdPhotoCaptureFragmentTest.kt
OrangeAndGreen: [24047c2](https://github.com/dimagi/commcare-android/pull/3726/commits/24047c2e191e65bafb64a990c5354968ef0f2192)

app/unit-tests/src/org/commcare/fragments/personalId/BasePersonalIdPhotoCaptureFragmentTest.kt
OrangeAndGreen: I did some refactoring to move common code to the base class. There are a couple combinations of how tests launch the activity and possibly to navigate to other fragments, hence some stacking of methods for different callers. [d8de09c](https://github.com/dimagi/commcare-android/pull/3726/commits/d8de09c5395318a62779e7da618e91a660b37862)

app/unit-tests/src/org/commcare/fragments/personalId/BasePersonalIdPhotoCaptureFragmentTest.kt
shubham1g5: > verify that the returned error string is not empty (rather than asserting on its exact contents)?

yeah that sounds quite good to me although think it's equally easy to just reference the `strings.xml` constant to  assert with. 


>  that will throw an exception if a throwable is included. Right now the calling code doesn't include a throwable, but if the error handler ever changes in that default case it would cause this unrelated test to fail.

Our approach for those tests has been to assert on exception with `assertThrows(exception)`

app/unit-tests/src/org/commcare/fragments/personalId/BasePersonalIdPhotoCaptureFragmentTest.kt
shubham1g5: looks great!

app/unit-tests/src/org/commcare/fragments/personalId/BasePersonalIdPhotoCaptureFragmentTest.kt
OrangeAndGreen: Ah, this was new to me but good to learn about and a nice improvement: [d94f66d](https://github.com/dimagi/commcare-android/pull/3726/commits/d94f66d621df8708bff340d494a88e4754bfa823)

shubham1g5 (commented):
changing my review to non-blocking as I will be un-available to review follow ups on this.

shubham1g5 (dismissed):

conroy-ricketts (dismissed):

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdPhotoCaptureFragmentTest.kt
OrangeAndGreen: Another nice improvement to learn about and implement! I also took the opportunity to extract common code related to MockWebServer and put it in the base test class, which did some cleanup in some of the existing test classes. [1c07eb2](https://github.com/dimagi/commcare-android/pull/3726/commits/1c07eb2c88302cb7a091cb34b9d9f32dfaf0feaa)

app/unit-tests/src/org/commcare/fragments/personalId/BasePersonalIdPhotoCaptureFragmentTest.kt
OrangeAndGreen: Removed the error handler mock when changing to use MockWebServer instead of mocking the API: [1c07eb2](https://github.com/dimagi/commcare-android/pull/3726/commits/1c07eb2c88302cb7a091cb34b9d9f32dfaf0feaa)

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdPhotoCaptureFragmentTest.kt
OrangeAndGreen: Mocking the web server now: [1c07eb2](https://github.com/dimagi/commcare-android/pull/3726/commits/1c07eb2c88302cb7a091cb34b9d9f32dfaf0feaa)

app/unit-tests/src/org/commcare/fragments/personalId/BasePersonalIdConfigurationTest.kt
```diff
@@ -29,7 +55,22 @@ const val TEST_INTEGRITY_TOKEN: String = "test_integrity_token_12345"
...
  */
-abstract class BasePersonalIdConfigurationTest {
+abstract class BasePersonalIdConfigurationTest<T : BasePersonalIdFragment> {
+    protected lateinit var activityController: ActivityController<PersonalIdActivity>
+    protected lateinit var navHostFragment: NavHostFragment
+    protected lateinit var activity: PersonalIdActivity
+    protected lateinit var fragment: T
+    protected lateinit var navController: TestNavHostController
+
+    // Only initialized for tests that opt in via setupMockWebServer()
```
conroy-ricketts, line 65: Thoughts on calling `setupMockWebServer()` from this class's `setUp()` (and `tearDownMockWebServer()` from this class's `tearDown()`), then the other test classes can optionally use them as needed?

OrangeAndGreen: Yeah, that cleans things up a bit and most of the fragments make web calls anyway. [e5c98ec](https://github.com/dimagi/commcare-android/pull/3726/commits/e5c98ecda8ac0971edad7a54293b23c12ef1df16)

conroy-ricketts (dismissed):

conroy-ricketts (dismissed):

conroy-ricketts (approved):

PR #3729, opened 2026-05-20
CCCT-2377 Add Email Entry Fragment To PersonalID Flow

Jignesh-dimagi:
## Product Description

Adds a new "Email entry" screen to the PersonalID configuration flow. The screen offers the user an optional email address, branches differently for signup, recovery, and a future legacy-add-email flow, and confirms a "Skip for now" tap with a yes/no dialog. The screen is not yet reachable from production code in this PR — the entry point (BackupCode → Email navigation) lands in a [follow-up PR](https://dimagi.atlassian.net/browse/CCCT-2407).

Screenshot_20260528_120741

Screenshot_20260528_120924

Screenshot_20260528_120949

## Technical Summary

- New Kotlin fragment `PersonalIdEmailFragment` + layout `fragment_personalid_email.xml` (banner + divider + icon + 55dp outlined input + side-by-side Material buttons). Outline-email vector drawable added.
- Nav graph: new `personalid_email` destination with mandatory `isLegacyFlow` / `isRecovery` boolean args; new `action_personalid_backupcode_to_personalid_email` exposed for the upstream PR.
- Three exit modes selected at runtime when pressing skip:
  - **Signup** → navigate to Photo Capture.
  - **Recovery** → call new `PersonalIdRecoveryCompleter.finalizeAccountRecovery(...)` (writes the DB passphrase, stores `ConnectUserRecord` with email, logs recovery analytics, fires second-device notification) and route to the existing recovery-success message destination.
  - **Legacy add-email** → `activity.finish()`.
- `ApiPersonalId.sendEmailOtp(token, user, ...)` now unified: uses `AuthInfo.TokenAuth(token)` when a session token exists, otherwise falls back to `AuthInfo.ProvidedAuth(user.userId, user.password)`.
- Skip button shows a `StandardAlertDialog` confirmation ("Skip email? / Are you sure you want to skip?") — Yes runs the existing skip logic, No is a no-op.
- Created EmailHelper: Encapsulated email-related logic for reusability across task-related operations (e.g., editing or adding emails from the profile screen).Key responsibilities include: Generating an auth token based on the application flow (configuration or existing user). Sending and verifying email OTPs. Redirecting the user based on the specific flow.
- 7 new English strings + translations in all 9 locales (es, fr, ha, hi, lt, no, pt, sw, ti); appbar title localized as "Email".
- Test scaffolding: extracted the existing IntegrityToken mock injection out of `BasePersonalIdPhoneFragmentTest` into a shared `BasePersonalIdConfigurationTest`. New `BasePersonalIdEmailFragmentTest` extends it and boots PersonalIdActivity, seeds an empty `PersonalIdSessionData`, and uses the real NavController to navigate to the email destination with the mandatory args.

### Safety story

What gives confidence:
- I ran the app, walked the email screen, and verified the layout matches the name-fragment style (banner, divider, icon, input, button row).
- I verified the action bar title reads "Email".
- 8 Robolectric unit tests pass — initial-state assertions (continue disabled, error hidden, skip enabled, input empty) and the email-input → continue-button enable/disable contract (valid / invalid / blank / valid-then-invalid).
- `./gradlew assembleCommcareDebug` and `ktlintFile` on every changed Kotlin file pass clean.
- The 12 existing `PersonalIdPhoneFragmentTest` cases still pass after the shared-base refactor.
- The screen is not yet reachable from production code, so the blast radius on shipped users is effectively zero until a follow-up PR wires the entry.

Risks to review:
- I did not exercise the legacy flow (`isLegacyFlow=true`) end-to-end on a device — these branches have no production entry point in this PR.

### Automated test coverage

- `PersonalIdEmailFragmentTest` (Robolectric, 8 cases): initial UI state + the email-validation → continue-button-state contract.
- Existing PersonalID-fragment tests (phone fragment, biometric config) continue to pass after the shared `BasePersonalIdConfigurationTest` extraction.
- Not covered: skip-confirm dialog button taps, recovery branch (`PersonalIdRecoveryCompleter` side effects), legacy branch (`activity.finish()`), `sendEmailOtp` auth-header selection. These will be exercised by the follow-up PRs that wire the entry points.

[CCCT-2377]: https://dimagi.atlassian.net/browse/CCCT-2377?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Jignesh-dimagi:
## Suggested Review Order

- `app/src/org/commcare/connect/network/ApiPersonalId.java` — the unified `sendEmailOtp(token, user)` is the key contract change; read first to anchor the auth-selection logic.
- `app/src/org/commcare/connect/network/connectId/PersonalIdApiHandler.java` — thin handler wrapper consumed by the fragment.
- `app/src/org/commcare/personalId/PersonalIdRecoveryCompleter.kt` — new helper invoked by the recovery branch in the fragment.
- `app/src/org/commcare/fragments/personalId/PersonalIdEmailFragment.kt` — main consumer; signup / recovery / legacy branching lives here.
- `app/res/layout/fragment_personalid_email.xml` — UI layout the fragment binds; mirrors `screen_personalid_name.xml`.
- `app/res/navigation/nav_graph_personalid.xml` — new destination with mandatory `isLegacyFlow` / `isRecovery` args.
- `app/unit-tests/src/org/commcare/fragments/personalId/BasePersonalIdConfigurationTest.kt` — new shared test base extracted from the phone test.
- `app/unit-tests/src/org/commcare/fragments/personalId/BasePersonalIdPhoneFragmentTest.kt` — refactored to extend the new base; confirms no phone-test regression.
- `app/unit-tests/src/org/commcare/fragments/personalId/BasePersonalIdEmailFragmentTest.kt` — email-specific test base (boots activity, seeds session data, navigates).
- `app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdEmailFragmentTest.kt` — 8 tests for initial state + email-input → continue-button contract.
- `app/res/drawable/ic_outline_email_24.xml` — Material outline-email vector, mirrors `ic_outline_person_24.xml`.
- `app/res/values/strings.xml` (+ 9 `values-*` locale files) — new email screen + skip-confirm strings and "Email" appbar title.
- `RELEASES.md` — fills the `Email entry screen:` QA placeholder under CommCare 2.64.

app/src/org/commcare/connect/network/connectId/PersonalIdApiHandler.java
```diff
@@ -332,11 +332,22 @@ public void validatePhoneOtp(Activity activity, String otp, PersonalIdSessionDat
         );
     }
 
-    public void sendEmailOtp(Activity activity, String email, PersonalIdSessionData sessionData) {
+    /**
+     * Sends an email OTP. Uses the PersonalID session token when present
+     * (signup / recovery); otherwise authenticates with the supplied
+     * {@code user}'s credentials (legacy flow).
+     */
+    public void sendEmailOtp(
```
OrangeAndGreen, line 340: Nit: This isn't new in this PR but it's the first time I noticed it.

sendPhoneOtp and sendEmailOtp both seem less than ideal as names since "send OTP" is ambiguous between the server sending an OTP to SMS/email vs. mobile sending the OTP to the server (i.e. for validation). Both are sending OTPs.

Suggest either requestEmailOtp or sendEmailOtpRequest

app/src/org/commcare/personalId/PersonalIdRecoveryCompleter.kt
```diff
@@ -0,0 +1,89 @@
...
+ * Mirrors the logic previously inlined in PersonalIdBackupCodeFragment.handleSuccessfulRecovery().
+ * This will be used from:
+ *  PersonalIdBackupCodeFragment: whenever the user already has a valid email and there is no need to show the email entry fragment (TODO https://dimagi.atlassian.net/browse/CCCT-2407).
+ *  PersonalIdEmailFragment: whenever the user presses skip and is recovering the account
+ *  PersonalIdEmailValidationFragment: whenever it validates OTP / skip and the user is recovering the account (TODO https://dimagi.atlassian.net/browse/CCCT-2378).
+ * Comments here will be updated as and when ToDo tickets are done.
+ */
+object PersonalIdRecoveryCompleter {
+    @JvmStatic
+    fun finalizeAccountRecovery(
```
OrangeAndGreen, line 25: I see there's a lot of overlap with the code in the Backup Code fragment that handles [successful recovery](https://github.com/dimagi/commcare-android/blob/5ee05b7c7573bd0b6c93229dad613f567865d53a/app/src/org/commcare/fragments/personalId/PersonalIdBackupCodeFragment.java#L213). Although I see in the other place we're retrieving the backup code from the UI when constructing the ConnectUserRecord.

Would it be difficult to rework that code so it can call this function instead of duplicating the logic?

app/src/org/commcare/personalId/PersonalIdRecoveryCompleter.kt
```diff
@@ -0,0 +1,89 @@
...
+                sessionData.userName,
+                sessionData.backupCode,
+                Date(),
+                sessionData.photoBase64,
+                sessionData.demoUser ?: false,
+                sessionData.requiredLock,
+                sessionData.invitedUser,
+            )
+        // Email (DB column)
+        user.email = sessionData.email
```
OrangeAndGreen, line 44: I suggest we add email to the constructor for ConnectUserRecord. That constructor is only called in 3 places:
1. Here
2. Backup code fragment (see my previous comment about having that code use this function instead)
3. Photo fragment (can pass in sessionData.email in case user configured email on the previous page)

app/src/org/commcare/personalId/PersonalIdRecoveryCompleter.kt
```diff
@@ -0,0 +1,89 @@
...
+        ConnectUserDatabaseUtil.storeUser(activity, user)
+
+        logRecoveryResult(true)
+        notifySecondDeviceLoginIfApplicable(activity, sessionData)
+    }
+
+    private fun logRecoveryResult(success: Boolean) {
```
OrangeAndGreen: Noting success will always be true since we're in the "recovery completer", I think we can hard-code true instead of passing it in.

Alternatively, we could expose the function publicly and have the Backup Code fragment use it instead of duplicating the analytics call (in that fragment it can be called with success either true or false)

app/src/org/commcare/fragments/personalId/PersonalIdEmailFragment.kt
```diff
@@ -0,0 +1,244 @@
...
+        inflater: LayoutInflater,
+        container: ViewGroup?,
+        savedInstanceState: Bundle?,
+    ): View {
+        binding = FragmentPersonalidEmailBinding.inflate(inflater, container, false)
+        personalIdSessionData =
+            ViewModelProvider(requireActivity())
+                .get(PersonalIdSessionDataViewModel::class.java)
+                .personalIdSessionData
+        activity = requireActivity()
```
OrangeAndGreen: No need for a class member here, "activity" can be a local variable

app/src/org/commcare/fragments/personalId/PersonalIdEmailFragment.kt
```diff
@@ -0,0 +1,244 @@
...
+class PersonalIdEmailFragment : BasePersonalIdFragment() {
+    private lateinit var binding: FragmentPersonalidEmailBinding
+    private lateinit var activity: Activity
+    private lateinit var personalIdSessionData: PersonalIdSessionData
+
+    /**
+     * True when launched for a legacy user who is adding email post-registration.
+     * Passed as a nav argument (isLegacyFlow: Boolean = false).
+     */
+    private var isLegacyFlow: Boolean = false
```
OrangeAndGreen: I think the naming could be misleading here. This isn't just legacy users who can get here, right? I think users who chose to skip the email during registration would also fall into this bucket.

Suggestion: isConfigurationWorkflow (this name would negate the boolean so true means use config token).
Or perhaps just userConfigured (no negation, true means use user credentials from DB)

app/src/org/commcare/fragments/personalId/PersonalIdEmailFragment.kt
```diff
@@ -0,0 +1,244 @@
...
+
+    /**
+     * True when entered from the recovery path (existing user who just validated backup code
+     * on a new device). On skip or successful OTP, the fragment must finalize account recovery
+     * instead of navigating to Photo Capture.
+     */
+    private var isRecovery: Boolean = false
```
OrangeAndGreen: Ah, so I think this bool combined with the one above dictate three possible states:
* Registering new account: legacy=false, recovery=false
* Recovering existing account: legacy=false, recovery=true
* Editing email with account already configured (i.e. from profile page): legacy=true, recovery=false

That leaves a fourth option that is possible but not valid: legacy=true, recovery=true.

Looking at the code, I see the recovery boolean will be ignored when legacy=true, so nothing breaks. But for the sake of making those parameters clear to the caller, I wonder if an enum would be more clear?

enum class EmailWorkflow {
   REGISTRATION,
   RECOVERY,
   UPDATE
}

app/src/org/commcare/fragments/personalId/PersonalIdEmailFragment.kt
```diff
@@ -0,0 +1,244 @@
...
+    override fun onResume() {
+        super.onResume()
+        binding.emailTextValue.requestFocus()
+    }
+
+    override fun keyboardEnterPressed() {
+        if (StringUtils.isValidEmail(binding.emailTextValue.text?.toString())) {
+            submitEmail()
+        } else {
+            KeyboardHelper.hideVirtualKeyboard(requireActivity())
```
OrangeAndGreen, line 71: This could be confusing to the user. If they think they entered a valid email but made a mistake, when they press Enter the keyboard will just go away and nothing will happen. Since we have the check on valid email inside submitEmail, I think it would be better for this code to always call through.

So if the user mistypes the email (resulting in an invalid result) and then presses Enter, they'll see the error message telling them that what they typed isn't valid so they know to go back and take a closer look.

OrangeAndGreen (changes requested):
Left a bunch of suggestions but only [one that I consider blocking](https://github.com/dimagi/commcare-android/pull/3729/changes#r3274860135)

app/src/org/commcare/connect/network/ApiPersonalId.java
```diff
@@ -360,13 +360,21 @@ public static void validatePhoneOtp(
         BaseApi.Companion.callApi(context, call, callback, ApiEndPoints.validateSessionOtp);
     }
 
+    /**
+     * Sends an email OTP. Auth source: PersonalID session token when available
+     * (signup / recovery), else basic auth from the supplied {@code user} (legacy
+     * flow where the user is already signed up and no session token exists).
+     */
```
conroy-ricketts, line 367: Confirming the intention here is that the fallback would only be used in the future (e.g. for the "manage/edit profile" feature)? The token should always be present during PersonalID signup/recovery right?

app/src/org/commcare/personalId/PersonalIdRecoveryCompleter.kt
conroy-ricketts, line 44: Can we move this to the `ConnectUserRecord` constructor

app/src/org/commcare/fragments/personalId/PersonalIdEmailFragment.kt
conroy-ricketts: Agreed, the term "legacy" implies deprecated code which is a bit confusing.

More suggestions: `isPostSignupEntry` or `isExistingUserFlow` to be more descriptive.

Also flagging that "legacy" is used in other places.

app/src/org/commcare/fragments/personalId/PersonalIdEmailFragment.kt
```diff
@@ -0,0 +1,244 @@
...
+                t: Throwable?,
+            ) {
+                showError(
+                    PersonalIdOrConnectApiErrorHandler.handle(
+                        requireActivity(),
+                        failureCode,
+                        t,
+                    ),
+                )
+                enableContinueButton(true)
```
conroy-ricketts, line 145: Should we check if the email is valid for enabling the button here?

app/res/layout/fragment_personalid_email.xml
```diff
@@ -0,0 +1,129 @@
...
+                android:background="@drawable/connect_side_icon_bg"
+                android:gravity="center">
+
+                <ImageView
+                    android:layout_width="42dp"
+                    android:layout_height="42dp"
+                    android:padding="1dp"
+                    android:src="@drawable/ic_outline_email_24" />
+
+            </LinearLayout>
```
conroy-ricketts: Can we remove the redundant nested LinearLayout here since it only has one image view?

conroy-ricketts (changes requested):
Looking really good so far! I wasn't able to finish the entire review today so I'll come back to this tomorrow

app/src/org/commcare/connect/network/connectId/PersonalIdApiHandler.java
Jignesh-dimagi: It's an interesting thing you pointed out. After some thoughts, it came as everything is being in an API request/handler; it's not wrong to have 'sendEmailOtp', as, by default, it's always requesting the server.

app/src/org/commcare/personalId/PersonalIdRecoveryCompleter.kt
Jignesh-dimagi: Basically, account recovery can end at three different places:

- PersonalIdBackupCodeFragment: whenever the user already has a valid email, as we are not showing the email entry screen to them.
- PersonalIdEmailFragment: whenever the user presses skip while recovering the account.
- PersonalIdEmailValidationFragment: whenever the user validates the email OTP while the user is recovering the account.

In order to handle the recovery from all these places, this recovery completer is created and will be the one handling it. I have already mentioned the same [here](https://github.com/dimagi/commcare-android/blob/e9645a68fd7d45d0907c93a896377eb49143ecfc/app/src/org/commcare/personalId/PersonalIdRecoveryCompleter.kt#L19) in the comment.

app/src/org/commcare/personalId/PersonalIdRecoveryCompleter.kt
Jignesh-dimagi: I was going to add that to the constructor, but it already has 10 parameters, which is really hard to read and follow. As I have stated in the above response, backup code will eventually be using this recovery completer only. For the photo capture page, it will be a similar method to the sign-up completer. So setting email should not be a problem.

app/src/org/commcare/fragments/personalId/PersonalIdEmailFragment.kt
Jignesh-dimagi: The term 'legacy' has propagated from the design doc. I think 'postRegistrationFlow' can be used. I am open to suggestions here.

app/src/org/commcare/fragments/personalId/PersonalIdEmailFragment.kt
Jignesh-dimagi: Here, I have tried to keep it the same as the phone number screen. The 'Continue' button is not enabled until the user enters a valid email address. This should signal to the user that email is not correct.

app/src/org/commcare/connect/network/ApiPersonalId.java
Jignesh-dimagi: Yes, that's correct

app/src/org/commcare/fragments/personalId/PersonalIdEmailFragment.kt
Jignesh-dimagi: Already The API can get called only when the email is correct. Here, an error occurs from an API call, and it should change the email address so we can enable the continue button in order that the user can request again.

app/src/org/commcare/personalId/PersonalIdRecoveryCompleter.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3729/changes/76398e4afb9cd1f37e0d01b67ab79249057e2797)

app/src/org/commcare/fragments/personalId/PersonalIdEmailFragment.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3729/changes/76398e4afb9cd1f37e0d01b67ab79249057e2797)

app/src/org/commcare/fragments/personalId/PersonalIdEmailFragment.kt
Jignesh-dimagi: Removed the term due to introduction of EmailWorkFlow enum, done [here](https://github.com/dimagi/commcare-android/pull/3729/changes/76398e4afb9cd1f37e0d01b67ab79249057e2797)

app/src/org/commcare/fragments/personalId/PersonalIdEmailFragment.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3729/changes/76398e4afb9cd1f37e0d01b67ab79249057e2797)

app/res/layout/fragment_personalid_email.xml
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3729/changes/2e16409c3d86caac13f78b694d1c36946746bcb6)

app/src/org/commcare/connect/network/connectId/PersonalIdApiHandler.java
OrangeAndGreen: Yeah, I definitely see your point as well. Okay leaving this as is.

app/src/org/commcare/personalId/PersonalIdRecoveryCompleter.kt
OrangeAndGreen: Agreed, but it can be used in PersonalIdBackupCodeFragment now, correct? Looks like right now only PersonalIdEmailFragment is actually using the new helper

app/src/org/commcare/fragments/personalId/PersonalIdEmailFragment.kt
OrangeAndGreen: Ah okay, I'm on board with staying consistent with other fragments in this PR.

Separately though, I wonder if we should create a backlog ticket to provide more direct feedback to users if they press the Enter key (rather than directly dismissing the keyboard) while the input is invalid on any of these pages. It seems like that action is direct enough that simply dismissing the keyboard may not be a clear response. Thoughts from others?

OrangeAndGreen (dismissed):
No blockers from me

app/src/org/commcare/connect/network/ApiPersonalId.java
conroy-ricketts: Sounds good, can we remove the word "legacy" from this comment

app/src/org/commcare/fragments/personalId/EmailWorkFlow.kt
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Launch context for the PersonalID email entry / verification screens.
+ *
+ *  - [REGISTRATION]: brand-new signup — the user is going through the full PersonalID flow
+ *    (phone → biometric → OTP → name → backup code → email → photo).
+ *  - [RECOVERY]: existing user recovering their account on a new device after validating the
+ *    backup code. On skip / successful OTP, the email step finalizes account recovery instead
+ *    of moving to photo capture.
+ *  - [EXISTING_USER]: a legacy logged-in user adding (or verifying) their email
```
conroy-ricketts: Can we remove the word "legacy" here

```suggestion
 *  - [EXISTING_USER]: a logged-in user adding (or verifying) their email
```

app/src/org/commcare/fragments/personalId/PersonalIdEmailFragment.kt
```diff
@@ -0,0 +1,233 @@
...
+import org.commcare.utils.StringUtils
+import org.commcare.views.dialogs.StandardAlertDialog
+
+class PersonalIdEmailFragment : BasePersonalIdFragment() {
+    private lateinit var binding: FragmentPersonalidEmailBinding
+    private lateinit var personalIdSessionData: PersonalIdSessionData
+
+    /**
+     * Launch context for this screen — distinguishes brand-new signup, account recovery,
+     * and the legacy "existing user adding email" entry point. Read from a required nav arg.
```
conroy-ricketts: Suggest removing the word "legacy"

```suggestion
     * and the "existing user adding email" entry point. Read from a required nav arg.
```

app/src/org/commcare/fragments/personalId/PersonalIdEmailFragment.kt
```diff
@@ -0,0 +1,233 @@
...
+            email,
+            if (workflow == EmailWorkFlow.EXISTING_USER) null else personalIdSessionData.token,
+            if (workflow == EmailWorkFlow.EXISTING_USER) ConnectUserDatabaseUtil.getUser(requireActivity()) else null,
+        )
+    }
+
+    private fun skipEmail() {
+        FirebaseAnalyticsUtil.reportPersonalIDContinueClicked(javaClass.simpleName, "skip")
+        when (workflow) {
+            // Existing user adding email post-registration — finish the legacy entry activity.
```
conroy-ricketts: Same here, the activity is not deprecated 

```suggestion
            // Existing user adding email post-registration — finish the entry activity.
```

conroy-ricketts (commented):
LGTM! Just think we should remove the word "legacy" from a few more places and that's it

app/src/org/commcare/fragments/personalId/PersonalIdEmailFragment.kt
shubham1g5: No strong thoughts on either behaviour, I find the current behaviour to dismiss the keyboard as acceptable here mostly because in my mind it's a softer action than clicking the Continue button but also would not oppose us making a change here.

app/src/org/commcare/personalId/PersonalIdRecoveryCompleter.kt
```diff
@@ -0,0 +1,89 @@
...
+                sessionData.oauthPassword,
+                sessionData.userName,
+                sessionData.backupCode,
+                Date(),
+                sessionData.photoBase64,
+                sessionData.demoUser ?: false,
+                sessionData.requiredLock,
+                sessionData.invitedUser,
+            )
+        // Email (DB column)
```
shubham1g5: nit: remove comment

app/src/org/commcare/personalId/PersonalIdRecoveryCompleter.kt
shubham1g5: +1 on removing duplicate code.

shubham1g5 (changes requested):

app/src/org/commcare/personalId/PersonalIdRecoveryCompleter.kt
Jignesh-dimagi: That is the whole plan for creating a recovery completer. As mentioned [here](https://github.com/dimagi/commcare-android/blob/e9645a68fd7d45d0907c93a896377eb49143ecfc/app/src/org/commcare/personalId/PersonalIdRecoveryCompleter.kt#L19), the duplicate code will be removed during [this ticket](https://dimagi.atlassian.net/browse/CCCT-2407) while implementing several navigation flow changes in PersonalIdBackupCodeFragment. 

As we can see, this code is already reused [here](https://github.com/dimagi/commcare-android/blob/c8c4c1a60575adbbb2c148cd189827c7e083db8f/app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt#L227) in [this PR](https://github.com/dimagi/commcare-android/pull/3730); the same change is to come for 'PersonalIdBackupCodeFragment'.

shubham1g5 (dismissed):

app/src/org/commcare/personalId/PersonalIdRecoveryCompleter.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3729/changes/ae879851f10b5caa5c0d8ba0c97e8124dbff77fe)

app/src/org/commcare/fragments/personalId/PersonalIdEmailFragment.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3729/changes/ae879851f10b5caa5c0d8ba0c97e8124dbff77fe)

app/src/org/commcare/fragments/personalId/PersonalIdEmailFragment.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3729/changes/ae879851f10b5caa5c0d8ba0c97e8124dbff77fe)

app/src/org/commcare/fragments/personalId/EmailWorkFlow.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3729/changes/ae879851f10b5caa5c0d8ba0c97e8124dbff77fe)

shubham1g5 (approved):

conroy-ricketts (approved):

app/src/org/commcare/personalId/PersonalIdRecoveryCompleter.kt
OrangeAndGreen: Okay, I may be getting a bit mixed up between the different PRs but as long as the function gets called from all the intended places when all the PRs are merged I have no complaints

OrangeAndGreen (approved):

Jignesh-dimagi:
@shubham1g5 @OrangeAndGreen @conroy-ricketts 

The latest commit includes the following major changes:

Created EmailHelper: Encapsulated email-related logic for reusability across task-related operations (e.g., editing or adding emails from the profile screen).
Key responsibilities include: 
- Generating an auth token based on the application flow (configuration or existing user). 
- Sending and verifying email OTPs. 
- Redirecting the user based on the specific flow.
- Validating email address

conroy-ricketts (dismissed):
The new helper object is going to be super helpful!

app/src/org/commcare/fragments/personalId/EmailHelper.kt
```diff
@@ -0,0 +1,146 @@
...
+     * navigate to the recovery-success message screen.
+     */
+    fun finalizeRecoveryAndShowSuccess(
+        activity: Activity,
+        sessionData: PersonalIdSessionData,
+        onSuccessNavigate: () -> Unit,
+    ) {
+        PersonalIdRecoveryCompleter.finalizeAccountRecovery(activity, sessionData)
+        onSuccessNavigate()
+    }
```
shubham1g5: this doesn't look related to email and should not be here.

shubham1g5 (changes requested):
one thing that we should sort out, but otherwise looks great.

app/src/org/commcare/fragments/personalId/EmailHelper.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3729/changes/78061f162d4b58c972cdc593d2926ec57217cf39)

shubham1g5 (dismissed):

shubham1g5 (dismissed):

shubham1g5 (approved):

PR #3730, opened 2026-05-22
CCCT-2378 Add Email OTP Verification Fragment To PersonalID Flow

Jignesh-dimagi:
## Product Description

Screenshot_20260521-111424_CommCare

Screenshot_20260521-111444_CommCare

Adds the Email OTP verification step to the PersonalID configuration flow. It is reached from the Email entry screen once the user submits an address. The screen accepts a 6-digit OTP, auto-submits on completion, supports resend with a 2-minute cooldown, shows a "Verification unsuccessful" dialog after 3 failed attempts (with retry or proceed-without-email), and branches its post-success behavior by flow:
- **Signup**: advances to Photo Capture.
- **Recovery**: finalizes the new device's account-recovery and shows the recovery-success screen.
- **Existing user**: writes the email onto the `ConnectUserRecord`, and closes the entry activity after the user acknowledges a success dialog.

This PR is stacked on top of #3729 (`ccct-2377-email-entry-fragment`) and will not compile against `master` until that lands.

## Technical Summary

- New Kotlin fragment `PersonalIdEmailVerificationFragment` + layout `fragment_personalid_email_verification.xml`.
- Resend timer runs on a `Handler` with 1-second ticks. The runnable computes `remaining` each tick and re-posts itself only while the cooldown hasn't elapsed; on expiry the UI flips the resend button visible and the runnable stops re-arming. `onDestroyView` removes any pending callback to scrub on teardown.
- Nav graph: adds `personalid_email_verification` destination with mandatory args `email: String` + `workflow: EmailWorkFlow`, plus `action_personalid_email_to_personalid_email_verification` from the existing email entry destination. The email entry fragment's `navigateToEmailVerification(email)` is now wired with Safe Args (passing `enteredEmail` + the current `workflow`).
- `EmailWorkFlow` enum (`REGISTRATION` / `RECOVERY` / `EXISTING_USER`) is consumed throughout. Both fragments read the workflow via the generated `…Args.fromBundle(...)`.
- API layer: `ApiPersonalId.verifyEmailOtp(context, email, otp, token, user, callback)` and the handler wrapper `PersonalIdApiHandler.verifyEmailOtp(activity, email, otp, token, user)` now accept a `ConnectUserRecord`. Auth header is `TokenAuth(token)` when a PersonalID session token is present (signup / recovery), else `ProvidedAuth(user.userId, user.password)` (existing user flow without a session). Mirrors the `sendEmailOtp` existing user fallback added in #3729.
- Success acknowledgement for the existing user flow: `showEmailAddedSuccessDialog()` builds a non-cancellable `StandardAlertDialog` ("Email Added" / "Your email has been added successfully.") whose OK button calls `requireActivity().finish()`. Pattern matches the existing `showProceedWithoutEmailDialog` in this fragment.
- 8 new English strings (`personalid_email_verification_appbar_title`, `_title`, `_description`, `personalid_email_verify_button`, `personalid_email_otp_failed_{title,message,retry,skip}`, `personalid_email_added_{title,message}`) plus translations across all 9 supported locales (es, fr, ha, hi, lt, no, pt, sw, ti).

### Safety story

What gives confidence:
- I walked the verification screen on a device: entered a valid OTP, an invalid OTP, watched the resend countdown, hit the "Verification unsuccessful" dialog after 3 failed attempts, and saw the success dialog for the `EXISTING_USER` flow.
- 6 Robolectric tests pass for the verification fragment — initial state (verify disabled, OTP empty, error hidden), description text contains the email nav-arg, partial code keeps button disabled, and complete 6-digit code triggers submission (button stays disabled while in flight).
- The 8 existing `PersonalIdEmailFragmentTest` cases still pass after the `EmailWorkFlow` enum refactor.
- `./gradlew assembleCommcareDebug` and `ktlintFile` clean on every changed Kotlin file.
- The new screen is still unreachable from production code — the BackupCode fragment's wiring to navigate into this stack lands in a [later PR](https://dimagi.atlassian.net/browse/CCCT-2407).

Risks to review:
- API endpoints need to be tested whenever available from the backend
- I was unable to test the EXISTING_USER flow; it will be verified in [upcoming ticket](https://dimagi.atlassian.net/browse/CCCT-2379).

### Automated test coverage

- `PersonalIdEmailVerificationFragmentTest` (Robolectric, 6 cases): initial UI state, description-email substitution, partial-code button-state, and complete-code auto-submit cascade.
- Existing `PersonalIdEmailFragmentTest` (8 cases) still passes after the `EmailWorkFlow` refactor.
- Existing `PersonalIdPhoneFragmentTest` (12 cases) still passes — the shared `BasePersonalIdConfigurationTest` is reused.

[CCCT-2378]: https://dimagi.atlassian.net/browse/CCCT-2378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Jignesh-dimagi:
## Suggested Review Order

- `app/src/org/commcare/connect/network/ApiPersonalId.java` — `verifyEmailOtp` now takes a `ConnectUserRecord`; auth header is token-or-basic. Mirror of the `sendEmailOtp` change in #3729.
- `app/src/org/commcare/connect/network/connectId/PersonalIdApiHandler.java` — thin handler wrapper forwarding the legacy fallback.
- `app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt` — main fragment; covers all three workflows + the resend timer + the success/failure dialogs.
- `app/res/layout/fragment_personalid_email_verification.xml` — UI the fragment binds against; matches `screen_personalid_phone_verify` styling.
- `app/res/navigation/nav_graph_personalid.xml` — new `personalid_email_verification` destination with `email` + `workflow` args, plus the action from `personalid_email`.
- `app/src/org/commcare/fragments/personalId/PersonalIdEmailFragment.kt` — only `navigateToEmailVerification` was wired up (previously a TODO).
- `app/unit-tests/src/org/commcare/fragments/personalId/BasePersonalIdEmailVerificationFragmentTest.kt` — new shared test base for the verification fragment; explains the seeded session token and SDK pin.
- `app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragmentTest.kt` — 6 tests for initial state + OTP-input contract.
- `app/res/values/strings.xml` (+ 9 `values-*` locale files) — 8 new strings for the verification screen and the success / failure dialogs.
- `RELEASES.md` — fills the `Email OTP screen:` QA placeholder under CommCare 2.64.

conroy-ricketts:
@Jignesh-dimagi do you have screenshots for this fragment

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
```diff
@@ -0,0 +1,334 @@
...
+
+class PersonalIdEmailVerificationFragment : BasePersonalIdFragment() {
+    private lateinit var binding: FragmentPersonalidEmailVerificationBinding
+    private lateinit var activity: Activity
+
+    /**
+     * Activity-scoped session data populated by upstream PersonalID fragments
+     * (phone / biometric / OTP / name / backup-code). The existing user add-email entry point
+     * does NOT go through those, so this is null in the existing user flow.
+     */
```
conroy-ricketts, line 35: Nit: verbose comment


```suggestion
    /**
     * Null for the existing user flow.
     */
```

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
```diff
@@ -0,0 +1,334 @@
...
+    private var personalIdSessionData: PersonalIdSessionData? = null
+
+    /**
+     * Email the user typed on the previous screen, threaded in as a nav arg. This is the
+     * canonical "pending" address — `sessionData.email` is NOT touched until OTP verification
+     * succeeds (see [onEmailVerified]).
+     */
```
conroy-ricketts: Nit: redundant comment

```suggestion
```

Jignesh-dimagi:
> @Jignesh-dimagi do you have screenshots for this fragment

Just added. Thanks for reminding.

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
```diff
@@ -0,0 +1,334 @@
...
+        destroyKeyboardScrollListener(binding.personalidEmailVerificationScrollView)
+    }
+
+    private fun enableVerifyButton(enabled: Boolean) {
+        binding.personalidEmailVerifyButton.isEnabled = enabled
+    }
+
+    private fun updateResendButtonState() {
+        val elapsed = System.currentTimeMillis() - otpRequestTime
+        val remaining = resendCooldownMillis - elapsed
```
conroy-ricketts: These two variables can be passed in as arguments

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
```diff
@@ -0,0 +1,334 @@
...
+            }
+        }.sendEmailOtp(
+            requireActivity(),
+            enteredEmail,
+            if (workflow == EmailWorkFlow.EXISTING_USER) null else personalIdSessionData?.token,
+            if (workflow == EmailWorkFlow.EXISTING_USER) user else null,
+        )
+    }
+
+    private fun submitOtp() {
```
conroy-ricketts, line 149: The existence of both `submitOtp()` and `resendOtp()` will lead to confusion. Thoughts on renaming `resendOtp()` to "`requestOtp()`"?

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragmentTest.kt
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    // The "complete 6-digit code" path is not tested at the fragment level. Entering 6 digits
+    // triggers an immediate submit chain (on-code-changed flips the button enabled, then the
+    // code-complete listener calls submitOtp() which disables it, then the API call may fail
+    // synchronously under Robolectric and call enableVerifyButton(true) again via onFailure
+    // → shouldAllowRetry). The post-state depends on whether the network resolves before the
+    // assertion runs, which is non-deterministic across CI environments. The on-code-changed
+    // listener wiring is already verified by `partial code keeps verify button disabled` and
+    // submitOtp's behavior should be covered by a dedicated handler test if needed.
```
conroy-ricketts: Nit: this comment would be better at the top of the class for future readers

RELEASES.md
```diff
@@ -68,7 +68,17 @@ we would like to communicate to QA as part of the release testing
         - Tap No: the dialog dismisses and the email screen stays put with any typed value preserved.
         - Tap Yes: the dialog dismisses and the flow advances to the next step (Photo Capture during signup).
     - Switch the device language to a supported locale (e.g. French, Spanish, Hindi, Swahili) and re-walk the screen. Verify the screen title, banner area, header copy, description, input hint, both button labels, and the skip-confirm dialog all render in the selected language.
-  - **Email OTP screen:** (will be added here)
+  - **Email OTP screen:**
+    - Reach the screen by entering a valid email on the previous step and tapping Continue (currently reachable only once the upstream PR wires Backup Code → Email → Email OTP). Verify the CommCare-by-Dimagi banner is shown at the top, the action bar title reads "Verify Email", a lock icon sits to the left of the 6-digit OTP field, and the description below the divider reads "Enter the 6-digit code sent to <email>" with the address the user typed.
```
conroy-ricketts: Nit: these notes can assume that 2.64 is complete


```suggestion
    - Reach the screen by entering a valid email on the previous step and tapping Continue. Verify the CommCare-by-Dimagi banner is shown at the top, the action bar title reads "Verify Email", a lock icon sits to the left of the 6-digit OTP field, and the description below the divider reads "Enter the 6-digit code sent to <email>" with the address the user typed.
```

conroy-ricketts (commented):
Looks fantastic so far!

@Jignesh-dimagi do you have a screenshot for when there is an error on this fragment

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
```diff
@@ -0,0 +1,334 @@
...
+        navigateToMessageDisplay(
+            getString(R.string.connect_recovery_success_title),
+            getString(R.string.connect_recovery_success_message),
+            isCancellable = false,
+            phase = ConnectConstants.PERSONALID_RECOVERY_SUCCESS,
+            buttonText = R.string.ok,
+        )
+    }
+
+    private fun showProceedWithoutEmailDialog() {
```
shubham1g5, line 219: I think once user have failed to provide an otp thrice, we should just move the user directly to the next page and toast a message saying "Unable to add your email" without adding an intermediary dialog, the worflow anyway provides back navigation support so user should be able to press back through workflow if they want to retry.

app/unit-tests/src/org/commcare/fragments/personalId/BasePersonalIdEmailVerificationFragmentTest.kt
```diff
@@ -0,0 +1,90 @@
...
+                .start()
+                .resume()
+                .get()
+
+        // The verification fragment reads PersonalIdSessionData from an activity-scoped ViewModel
+        // for registration/recovery workflows. In production the upstream fragments populate it; in tests we
+        // skip those, so seed an instance here. NumericCodeView fires its code-complete listener
+        // synchronously when setCode("123456") is called, which triggers submitOtp() → API call.
+        // We seed a non-null token so the token-auth header builds without NPE; the actual HTTP
+        // call enqueues asynchronously and does not affect the synchronous UI assertions.
```
shubham1g5: think the whole commit is un-ncessary here, if you think otherwise you should simplify it  and make it conscise to a single line.

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragmentTest.kt
shubham1g5: We should be able to mock the API call and make the behaviour deterministic to test here, that is the approach we already take to test API dependent feature set and think the comment here is not justifiable and we must add coverage for all worflows in the fragment.

shubham1g5 (changes requested):

shubham1g5:
Are you planning to add analytics on this feature as part of a separate PR ?

Jignesh-dimagi:
> Are you planning to add analytics on this feature as part of a separate PR ?

@shubham1g5 I originally planned to add the analytics to the current PR, but it might be clearer to do it in a separate PR once the major work is done (including existing user flow). Let me know if you're okay with that. I can create ticket for the same.

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
Jignesh-dimagi: This dialog will appear for all three use cases, i.e., sign up, recovery and exiting user flow. The user will be able to navigate back during the sign-up flow, but for recovery and existing user flow, they will not be able, as next page during recovery is account recovery success message , and for existing user flow, it will close the activity. The user will never have a chance to navigate back.

shubham1g5:
> Let me know if you're okay with that. I can create ticket for the same.

ok with that, please post the ticket here once you have added it.

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
shubham1g5: >  The user will never have a chance to navigate back.

I still think having a dialog after every 3 tires that doesn't do anything specific other than resetting the failure count seems like a poor UX here. Are you able to propose alternate options for all the 3 flows here if we don't show a dialog.

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
Jignesh-dimagi: I have not removed it completely but kept the brief description.  Done [here](https://github.com/dimagi/commcare-android/pull/3730/changes/e19de14f3e67695b9d779b1606f45e1356962d10).

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3730/changes/e19de14f3e67695b9d779b1606f45e1356962d10).

RELEASES.md
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3730/changes/e19de14f3e67695b9d779b1606f45e1356962d10).

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragmentTest.kt
Jignesh-dimagi: Removed it [here](https://github.com/dimagi/commcare-android/pull/3730/changes/e19de14f3e67695b9d779b1606f45e1356962d10).

app/unit-tests/src/org/commcare/fragments/personalId/BasePersonalIdEmailVerificationFragmentTest.kt
Jignesh-dimagi: Removed it, (https://github.com/dimagi/commcare-android/pull/3730/changes/e19de14f3e67695b9d779b1606f45e1356962d10).

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
Jignesh-dimagi: The dialog features two buttons: 'Try again' and 'Proceed without email'. 

After every three failed attempts, it does mainly two things:
- Notifies the user that they have had multiple unsuccessful attempts.
- Allows the user to proceed without an email address in case of access issues or other problems.

I believe this approach will help users to decide. Your thoughts?

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
```diff
@@ -0,0 +1,330 @@
...
+     * does NOT go through those, so this is null in the existing user flow.
+     */
+    private var personalIdSessionData: PersonalIdSessionData? = null
+
+    /**
+     * Email the user typed on the previous screen, threaded in as a nav arg.
+     */
+    private lateinit var enteredEmail: String
+
+    /**
```
OrangeAndGreen, line 40: I'd vote that the comments on all these member variables aren't necessary, that's not something we usually have elsewhere in the code. But maybe we should have a discussion on whether we think descriptive comments like these are helpful, for us and/or for Claude?

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
```diff
@@ -0,0 +1,330 @@
...
+            requireActivity(),
+            enteredEmail,
+            if (workflow == EmailWorkFlow.EXISTING_USER) null else personalIdSessionData?.token,
+            if (workflow == EmailWorkFlow.EXISTING_USER) user else null,
+        )
+    }
+
+    private fun submitOtp() {
+        val otp = binding.otpCodeView.codeValue
+        if (otp.length != 6) return
```
OrangeAndGreen: When might we hit this "if"? Seems like this function should only be called once the OTP is valid, but if we somehow get here I'm wondering whether we should log an error and/or inform the user. But maybe the silent return is okay if there's a case I'm not thinking of.

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
```diff
@@ -0,0 +1,330 @@
...
+        FirebaseAnalyticsUtil.reportPersonalIDContinueClicked(javaClass.simpleName, null)
```
OrangeAndGreen, line 155: Nit: In code I think we prefer PersonalId instead of PersonalID. Noting though that the function wasn't created in this PR, could just be a nice cleanup

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
```diff
@@ -0,0 +1,330 @@
...
+            enteredEmail,
+            otp,
+            if (workflow == EmailWorkFlow.EXISTING_USER) null else personalIdSessionData?.token,
+            if (workflow == EmailWorkFlow.EXISTING_USER) user else null,
+        )
+    }
+
+    private fun onEmailVerified() {
+        when (workflow) {
+            // Existing user: write straight to the existing ConnectUserRecord; session data is not
```
OrangeAndGreen: The comments here don't feel necessary to me

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
OrangeAndGreen: My two cents: I agree the UX is a bit odd here, but in the end I think the best approach is to prompt the user after every few failures to nudge them to skip if they're having trouble (as implemented now). In reality I suspect few users will fail to enter the correct code three times (they'll either get it right the first time or make a mistake and then fix it). But if they are having trouble, then they're likely to skip before failing three times, or otherwise choose to do so when given the prompt after three failures.

But giving them the decision makes it clear that this may not be a permanent failure, and if they really want to keep trying they can choose to do so. Maybe we could improve the message in the dialog to say something like "Looks like you're having trouble verifying your email. You can skip this for now and try again later, or keep trying".

OrangeAndGreen (approved):
Nothing blocking from me

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
conroy-ricketts: I'd vote for not having these comments at all. The code should be (and IMO is) readable enough to not need them

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragmentTest.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3730/changes/4aeb08241986f3d031c89938294ad84518cb857c).

Jignesh-dimagi:
> ok with that, please post the ticket here once you have added it.

@shubham1g5 https://dimagi.atlassian.net/browse/CCCT-2447

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
Jignesh-dimagi: @conroy-ricketts Only having a comment for 'NULL' doesn't describe it, so I appended that. Done [here](https://github.com/dimagi/commcare-android/pull/3730/changes/38bba7862be3b35ed93a6551fe429eeaa117877f)

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3730/changes/38bba7862be3b35ed93a6551fe429eeaa117877f)

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3730/changes/38bba7862be3b35ed93a6551fe429eeaa117877f)

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
shubham1g5: Strongly against silent returns here, we should at very least log this and show an error to the user but also preferably just throw an Illegal State exception from here if this scenario is not practically possible.

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
shubham1g5: >  prompt the user after every few failures to nudge them to skip if they're having trouble (as implemented now).

That framing makes sense to me. 

> aybe we could improve the message in the dialog to say something like "Looks like you're having trouble verifying your email. You can skip this for now and try again later, or keep trying".

+1

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragmentTest.kt
```diff
@@ -0,0 +1,236 @@
...
+
+        repeat(3) {
+            enterCode("123456")
+            mockWebServer.takeRequest()
+            ShadowLooper.runUiThreadTasksIncludingDelayedTasks()
+        }
+
+        val dialog = ShadowDialog.getLatestDialog() as? AlertDialog
+        assertNotNull("Verification-unsuccessful dialog should be shown after 3 failed attempts", dialog)
+        assertTrue("Dialog should be visible", dialog!!.isShowing)
```
shubham1g5, line 214: can we expand this test to verify the behaviour on the 2 CTAs present on this dialog.

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
conroy-ricketts: @Jignesh-dimagi Bumping this comment so it does not get lost in the weeds

Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3730/changes/e51915a6527da01d6fa6a928d684a737ce868edc#diff-b52a28c2bfe3b06160db63e45a92f721d7fdb54c26026b065b3e18af173d09bf)

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
Jignesh-dimagi: Changed dialog title [here](https://github.com/dimagi/commcare-android/pull/3730/changes/e51915a6527da01d6fa6a928d684a737ce868edc#diff-b52a28c2bfe3b06160db63e45a92f721d7fdb54c26026b065b3e18af173d09bf)

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3730/changes/e51915a6527da01d6fa6a928d684a737ce868edc#diff-b52a28c2bfe3b06160db63e45a92f721d7fdb54c26026b065b3e18af173d09bf)

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragmentTest.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3730/changes/e51915a6527da01d6fa6a928d684a737ce868edc#diff-b52a28c2bfe3b06160db63e45a92f721d7fdb54c26026b065b3e18af173d09bf)

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
Jignesh-dimagi: That’s a good suggestion. I plan to tackle this after the email changes are merged and will create a separate PR for it. Including it in this current PR would clutter too many files and confuse the reviewers

conroy-ricketts: IMO PersonalID is better because it matches the branding. I think at some point in the past I made a few changes to make the "D" capital

conroy-ricketts (approved):

shubham1g5 (approved):

Jignesh-dimagi:
@shubham1g5 @conroy-ricketts :
The last minor commit made the code compatible with the EmailHelper function removal.

shubham1g5 (approved):

PR #3731, opened 2026-05-22
CCCT-2355 Handle Missing Integrity headers Error on Start Configuration

shubham1g5:
## Product Description

Fixes a PersonalID configuration crash where the server rejects the start-configuration request because the integrity token was missing from the request headers. 

Some Useful notes to review:

- We make the call to the server without token in the first place as token is not always mandatory depending on whether the user is pre-invited or not. 
- I have also enhanced logging around this issue for us to be sure that this issue only happens due to token being empty. 

### Safety story

Minimum risks, Error handling mostly with existing code pathway

### Automated test coverage

Added tests in `PersonalIdPhoneFragmentStartConfigurationTest` 

[CCCT-2355]: https://dimagi.atlassian.net/browse/CCCT-2355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/org/commcare/fragments/personalId/PersonalIdPhoneFragment.java
+++ b/app/src/org/commcare/fragments/personalId/PersonalIdPhoneFragment.java
@@ -61,6 +61,7 @@ import org.javarosa.core.services.Logger;
 import org.jetbrains.annotations.NotNull;
 
 import java.util.HashMap;
+import java.util.function.Consumer;
 
 import static com.google.android.play.core.integrity.model.IntegrityDialogResponseCode.DIALOG_SUCCESSFUL;
 import static org.commcare.utils.Permissions.shouldShowPermissionRationale;
@@ -341,30 +342,11 @@ public class PersonalIdPhoneFragment extends BasePersonalIdFragment implements C
             body.put("device", model);
         }
 
-        integrityTokenApiRequestHelper.withIntegrityToken(
-                body,
-                new IntegrityTokenViewModel.IntegrityTokenCallback() {
-                    @Override
-                    public void onTokenReceived(
-                            @NotNull String requestHash,
-                            @NotNull StandardIntegrityManager.StandardIntegrityToken integrityTokenResponse
-                    ) {
-                        makeStartConfigurationCall(requestHash, body, integrityTokenResponse);
-                    }
-
-                    @Override
-                    public void onTokenFailure(@NotNull Exception exception) {
-                        String errorCode = IntegrityTokenApiRequestHelper.Companion.getCodeForException(
-                                exception
-                        );
-                        FirebaseAnalyticsUtil.reportPersonalIdConfigurationIntegritySubmission(
-                                errorCode
-                        );
-
-                        makeStartConfigurationCall(null, body, null);
-                    }
-                }
-        );
+        fetchIntegrityTokenAndStartConfiguration(body, exception -> {
+            String errorCode = IntegrityTokenApiRequestHelper.Companion.getCodeForException(exception);
+            FirebaseAnalyticsUtil.reportPersonalIdConfigurationIntegritySubmission(errorCode);
+            makeStartConfigurationCall(null, body, null);
+        });
     }
 
     @Override
@@ -554,6 +536,21 @@ public class PersonalIdPhoneFragment extends BasePersonalIdFragment implements C
                                 personalIdSessionDataViewModel.getPersonalIdSessionData().getSessionFailureSubcode()
                         );
                         break;
+                    case MISSING_DATA_ERROR: {
+                        String subCode = personalIdSessionDataViewModel.getPersonalIdSessionData().getSessionFailureSubcode();
+                        boolean isIntegrityHeadersMissing = BaseApiHandler.PersonalIdApiSubErrorCodes.INTEGRITY_HEADERS.name().equals(subCode);
+                        if (isIntegrityHeadersMissing && token.isEmpty()) {
+                            retryWithNewIntegrityToken(body);
+                        } else if (isIntegrityHeadersMissing) {
+                            onConfigurationFailure(
+                                    AnalyticsParamValue.START_CONFIGURATION_INTEGRITY_CHECK_FAILURE,
+                                    getString(R.string.personalid_configuration_process_failed_subtitle)
+                            );
+                        } else {
+                            navigateFailure(failureCode, t);
+                        }
+                        break;
+                    }
                     default:
                         navigateFailure(failureCode, t);
                         break;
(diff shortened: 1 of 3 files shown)


shubham1g5:
## Suggested Review Order

Review commit by commit

app/src/org/commcare/fragments/personalId/PersonalIdPhoneFragment.java
```diff
@@ -608,6 +606,34 @@ private void showIntegrityCheckDialog(
...
+                            @NotNull String requestHash,
+                            @NotNull StandardIntegrityManager.StandardIntegrityToken integrityTokenResponse
+                    ) {
+                        makeStartConfigurationCall(requestHash, body, integrityTokenResponse);
+                    }
+
+
+                    @Override
+                    public void onTokenFailure(@NotNull Exception exception) {
+                        onTokenFailure.accept(exception);
```
conroy-ricketts, line 635: Flagging that it seems this requires API level 24 but our current minimum is API level 23. Should we have a fallback here?

shubham1g5: Personal ID signup is hidden before Android N (api 24), so think we are good without taking an action here.

conroy-ricketts (approved):

app/src/org/commcare/fragments/personalId/PersonalIdPhoneFragment.java
```diff
@@ -543,17 +525,26 @@ public void onFailure(
...
                         handleIntegritySubError(
                                 integrityTokenResponse,
                                 personalIdSessionDataViewModel.getPersonalIdSessionData().getSessionFailureSubcode()
                         );
                         break;
+                    case MISSING_DATA_ERROR: {
+                        String subCode = personalIdSessionDataViewModel.getPersonalIdSessionData().getSessionFailureSubcode();
+                        boolean isIntegrityHeadersMissing = BaseApiHandler.PersonalIdApiSubErrorCodes.INTEGRITY_HEADERS.name().equals(subCode);
+                        if (isIntegrityHeadersMissing && token.isEmpty()) {
+                            retryWithNewIntegrityToken(body);
```
Jignesh-dimagi: We are already retrying the integrity token fetch [here](https://github.com/dimagi/commcare-android/blob/3d86f7b442d9d24edd74e128137e61cbdaa887b9/app/src/org/commcare/android/integrity/IntegrityTokenViewModel.kt#L89) before sending a blank / null integrity token. Do we still need another retry?

shubham1g5: That retry doesn't cover all failure modes for example if provider itself is not initialised at the time when user presses continue, the philosophy here is that after server enforces user to get the token, we make one last try before showing the error to the user.

Jignesh-dimagi: I think if the provider itself is not initialised, it has a 'pendingRequests' queue to respond back whenever it is. But, in general, I agree with your philosophy to try one last time. Another question here: will the user see the loading dialog whenever the app is requesting the token again? If it's not, the user might be confused that nothing is happening and might keep pressing the continue button. Your thoughts?

shubham1g5: > I think if the provider itself is not initialised, it has a 'pendingRequests' queue to respond back whenever it is.

Ahh that's a good point, I think we should not need a retry in that case, but going to re-look the code here if there are other cases that might need us to retry and rework this PR accordingly.

shubham1g5: ended up removing retry here - https://github.com/dimagi/commcare-android/pull/3731/commits/ad5c47659d84c7d814841b14f73e490a5e745c84

app/src/org/commcare/fragments/personalId/PersonalIdPhoneFragment.java
```diff
@@ -543,17 +525,31 @@ public void onFailure(
...
                         );
                         break;
+                    case MISSING_DATA_ERROR: {
+                        String subCode = personalIdSessionDataViewModel.getPersonalIdSessionData().getSessionFailureSubcode();
+                        boolean isIntegrityHeadersMissing = BaseApiHandler.PersonalIdApiSubErrorCodes.INTEGRITY_HEADERS.name().equals(subCode);
+                        if (isIntegrityHeadersMissing) {
+                            if (!token.isEmpty()) {
+                                Logger.exception("Missing Data error related to ingerity headers",
+                                        new Exception("Missing integrity headers when token is present"));
+                            }
```
Jignesh-dimagi, line 546: @shubham1g5 Do we need to log exception even when token is empty?

app/src/org/commcare/fragments/personalId/PersonalIdPhoneFragment.java
```diff
@@ -543,17 +525,31 @@ public void onFailure(
...
                                 integrityTokenResponse,
                                 personalIdSessionDataViewModel.getPersonalIdSessionData().getSessionFailureSubcode()
                         );
                         break;
+                    case MISSING_DATA_ERROR: {
+                        String subCode = personalIdSessionDataViewModel.getPersonalIdSessionData().getSessionFailureSubcode();
+                        boolean isIntegrityHeadersMissing = BaseApiHandler.PersonalIdApiSubErrorCodes.INTEGRITY_HEADERS.name().equals(subCode);
+                        if (isIntegrityHeadersMissing) {
+                            if (!token.isEmpty()) {
+                                Logger.exception("Missing Data error related to ingerity headers",
```
Jignesh-dimagi: Minor: Spelling mistake `ingerity` -> 'integrity'

shubham1g5: https://github.com/dimagi/commcare-android/pull/3731/commits/1a377386f764e86cf0e02466d265695e6f9a6ec0

app/src/org/commcare/fragments/personalId/PersonalIdPhoneFragment.java
shubham1g5: It was not the objective of the logs I added and think token being empty is expected in some cases, but added the log guessing it's harmless - https://github.com/dimagi/commcare-android/pull/3731/commits/1a377386f764e86cf0e02466d265695e6f9a6ec0

Jignesh-dimagi (approved):

conroy-ricketts (approved):

PR #3732, opened 2026-05-22
Remove unused DOTS tracker feature

avazirna:
## Product Description
This PR removes the DOTS (Directly Observed Therapy) tracker feature, as it is no longer in use. The feature was originally developed for the UATBC project to track medication intake by program beneficiaries. Since it is obsolete, the associated functionality and related screens have been removed from the application. The main screen that were part of this feature include:
<table>
<tr>
<th>Calendar</th>
<th>Day period</th>
<th>Medication intake</th>
</tr>
<tr>
<td>
dots_calendar
</td>
<td>
dot_period
</td>
<td>
dots_medication_intake
</td>
</tr>
</table>

## Technical Summary
Retires `DotsEntryActivity` and its supporting classes/resources. The activity was launchable only via the `org.commcare.dalvik.action.DotsEntry` intent action, declared `exported="false"` so only same-app callers could
fire it — and **no in-tree caller fires that action**. Neither `commcare-android` nor the `libs/commcare-core` submodule contains any reference to the action string or to the deleted classes outside the DOTS files themselves. Last substantive change to any of these sources was a 2025-09 lint pass; no feature work in years.

### Safety story
- Static verification on the branch tip: `git grep` for every removed class name, every removed resource ID, and the intent action string `org.commcare.dalvik.action.DotsEntry` returns zero matches in both `app/` and the `libs/commcare-core` submodule.
- Commit ordering keeps the manifest valid at every intermediate step (manifest dropped first, then classes, then resources — never a manifest pointing at a missing class).
- Data impact: none expected. The activity serialized state into its own bundle key (`DOTS_DATA`); since nothing on master launches it, there are no in-flight payloads to migrate.

## Labels and Review
- [ ] Manual QA: not required — feature is unreachable on master.
- [ ] RELEASES.md: no update needed (no behavior change).
- [x] Risk: low (pure deletion, statically verified).
- [ ] Reviewers: ping someone familiar with the historical DOTS tracker
      and the XForm intent-callout dispatch path.


avazirna:
@ctsims I've been planning to remove this feature for some time. Any concerns with this removal?

conroy-ricketts (approved):

PR #3733, opened 2026-05-22
CCCT-2301 Add Unit Tests For PersonalIdNameFragment

conroy-ricketts:
## Technical Summary

Adds Robolectric unit tests for `PersonalIdNameFragment`. The fragment previously had no unit coverage; this PR backfills that gap with two test classes plus a shared base.

### Safety story

What gives confidence:
- I ran the tests locally and saw all 17 tests pass.
- Test-only change: no production code was touched, so the diff cannot regress runtime behavior.
- New tests follow the patterns already in use for the sibling phone and biometric fragments, including the `MockWebServer` + `PersonalIdApiClient.apiService` field-injection approach.

Risks to review:
- The `MockWebServer` setup mutates a static `apiService` field on `PersonalIdApiClient` via reflection and restores it in `tearDown`. If a future test forgets to restore it (or the field name changes), unrelated tests can flake — same risk that already exists in `PersonalIdPhoneFragmentStartConfigurationTest`.
- The enter-key "empty input" test asserts that no HTTP request is made, but relies on `TextView.onEditorAction(IME_ACTION_DONE)` actually invoking the editor-action listener under Robolectric. If a Robolectric upgrade changes that behavior the test could pass trivially.

### Automated test coverage

This PR *is* the automated test coverage. The 17 new tests exercise: initial UI state, the name-input → continue-button gating, request payload + auth header for `/users/check_name`, the success path (navigation + session-data write), retryable failure UI (500), non-retryable failure UI (403 `PHONE_NOT_VALIDATED`), routed failures (401 `LOCKED_ACCOUNT` → message screen), and the enter-key shortcut.

[CCCT-2301]: https://dimagi.atlassian.net/browse/CCCT-2301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdNameFragmentAddOrVerifyNameTest.kt
+++ b/app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdNameFragmentAddOrVerifyNameTest.kt
@@ -9,6 +9,7 @@ import androidx.test.ext.junit.runners.AndroidJUnit4
 import com.google.android.material.button.MaterialButton
 import okhttp3.mockwebserver.MockResponse
 import okhttp3.mockwebserver.MockWebServer
+import okhttp3.mockwebserver.RecordedRequest
 import org.commcare.CommCareTestApplication
 import org.commcare.activities.connect.viewmodel.PersonalIdSessionDataViewModel
 import org.commcare.connect.network.ApiService
@@ -25,6 +26,7 @@ import org.junit.Test
 import org.junit.runner.RunWith
 import org.robolectric.annotation.Config
 import org.robolectric.shadows.ShadowLooper
+import java.util.concurrent.TimeUnit
 
 /**
  * Tests the `addOrVerifyName` API integration of [PersonalIdNameFragment]. Uses MockWebServer to
@@ -96,10 +98,18 @@ class PersonalIdNameFragmentAddOrVerifyNameTest : BasePersonalIdNameFragmentTest
         ShadowLooper.idleMainLooper()
     }
 
+    /**
+     * Reads the next request with a bounded wait so a missing dispatch fails fast instead of
+     * hanging the suite, then drains delayed UI work so callbacks can run before assertions.
+     */
     private fun drainHttp() {
-        mockWebServer.takeRequest()
+        takeRequestOrFail()
         ShadowLooper.runUiThreadTasksIncludingDelayedTasks()
     }
+
+    private fun takeRequestOrFail(timeoutSeconds: Long = 5): RecordedRequest =
+        mockWebServer.takeRequest(timeoutSeconds, TimeUnit.SECONDS)
+            ?: throw AssertionError("Expected an HTTP request within ${timeoutSeconds}s but none arrived")
 
     // ========== Request payload ==========
 
@@ -109,8 +119,7 @@ class PersonalIdNameFragmentAddOrVerifyNameTest : BasePersonalIdNameFragmentTest
 
         typeAndContinue("  Margaret Hamilton  ")
 
-        val request = mockWebServer.takeRequest()
-        assertNotNull(request)
+        val request = takeRequestOrFail()
         assertEquals("/users/check_name", request.path)
         assertEquals("POST", request.method)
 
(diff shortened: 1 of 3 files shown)


conroy-ricketts:
## Suggested Review Order

- `app/unit-tests/src/org/commcare/fragments/personalId/BasePersonalIdNameFragmentTest.kt` — shared setup: activity + session ViewModel + TestNavHostController; both test classes depend on it.
- `app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdNameFragmentTest.kt` — straightforward UI-state tests; easiest read and documents the TextWatcher contract.
- `app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdNameFragmentAddOrVerifyNameTest.kt` — heavier MockWebServer-backed tests; pairs naturally with the existing `PersonalIdPhoneFragmentStartConfigurationTest`.

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdNameFragmentAddOrVerifyNameTest.kt
```diff
@@ -0,0 +1,279 @@
...
+        setName(name)
+        activity.runOnUiThread { continueButton().performClick() }
+        ShadowLooper.idleMainLooper()
+    }
+
+    private fun drainHttp() {
+        mockWebServer.takeRequest()
+        ShadowLooper.runUiThreadTasksIncludingDelayedTasks()
+    }
+
```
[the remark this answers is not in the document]
conroy-ricketts: [Claude Code AI] Verified against current code and addressed in 267032f5d:

- `drainHttp()` and the two tests that read `request.path` / `request.headers` (formerly lines 100, 112, 128) — **valid**. Both call sites now go through a new `takeRequestOrFail(timeoutSeconds = 5)` helper that calls `mockWebServer.takeRequest(5, TimeUnit.SECONDS)` and throws `AssertionError` if it returns null, so a missing dispatch fails fast instead of hanging the suite.
- Looper drain after `continueButton().performClick()` in the "immediately disables" test — **valid in spirit**, but I used `ShadowLooper.idleMainLooper()` rather than the suggested `runUiThreadTasksIncludingDelayedTasks()`. The stronger drain would advance time far enough for the enqueued mock response to settle, which would weaken the "immediately disabled" guarantee the test exists to defend (success doesn't re-enable the button — it navigates away, so the assert would still pass but for the wrong reason). `idleMainLooper()` runs only the immediately-posted UI work, matching the pattern already used by `typeAndContinue` in the same file.

All 10 tests in this class still pass after the changes.

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdNameFragmentAddOrVerifyNameTest.kt
```diff
@@ -0,0 +1,292 @@
...
+        typeAndContinue("  Margaret Hamilton  ")
+
+        val request = takeRequestOrFail()
+        assertEquals("/users/check_name", request.path)
+        assertEquals("POST", request.method)
+
+        val body = request.body.readUtf8()
+        assertTrue("Body should carry the trimmed name", body.contains("\"name\":\"Margaret Hamilton\""))
+        assertFalse("Body should not include the surrounding whitespace", body.contains("  Margaret Hamilton  "))
```
shubham1g5: nit: think the first assert automatically verify the second assert here.

shubham1g5 (dismissed):

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdNameFragmentAddOrVerifyNameTest.kt
conroy-ricketts: Agreed 621778183e4e08ca0f9fdd67ef180aa425f3b45f

shubham1g5 (approved):

OrangeAndGreen (approved):

PR #3736, opened 2026-05-25
SAAS-19802 Fix Combobox.autoCorrectCapitalization choice match check

avazirna:
## Product Description
Fixes a bug causing string/numeric questions to lose focus while users are entering data. For each character/digit typed, the form controller performs a pass through the form's triggerables. During this process, existing questions are re-evaluated. Because the `ComboboxWidget` current value within `autoCorrectCapitalization` was always `false`, a new value was being set, causing the input focus to be reassigned:

https://github.com/user-attachments/assets/d3f7c374-152b-474e-b935-4ba3ac7dfe29

Ticket: https://dimagi.atlassian.net/browse/SAAS-19802

Related PRs:
- https://github.com/dimagi/commcare-android/pull/3515 
- https://github.com/dimagi/commcare-core/pull/1516

## Technical Summary
`autoCorrectCapitalization` had a latent bug in its guard:

```java
if (enteredText != null && !choices.contains(enteredText) && ...)
```

`choices` is `Vector<ComboItem>` and `enteredText` is `String`. `ComboItem.equals(Object)` (commcare-core) only returns `true` for another `ComboItem`, so `choices.contains(enteredText)` was **always false**. The intended "skip the case-fix when the user already typed the canonical form" branch never fired, so a `setText(...)` ran every
time the field lost focus on a valid value — silent extra work plus an unnecessary TextWatcher pass.

### Safety story
- This was successfully tested locally.
- Added new tests to assess the correct behavior of the `autoCorrectCapitalization()` method

### Automated test coverage
`app/unit-tests/src/org/commcare/views/ComboboxTest.kt` — four tests,
all passing locally 

## Labels and Review

### DIFF
--- a/app/unit-tests/src/org/commcare/views/ComboboxTest.kt
+++ b/app/unit-tests/src/org/commcare/views/ComboboxTest.kt
@@ -1,0 +1,87 @@
+package org.commcare.views
+
+import androidx.test.core.app.ApplicationProvider
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import org.commcare.CommCareTestApplication
+import org.commcare.adapters.ComboboxAdapter
+import org.javarosa.core.model.ComboItem
+import org.javarosa.core.model.ComboboxFilterRule
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.annotation.Config
+import java.util.Vector
+
+@Config(application = CommCareTestApplication::class)
+@RunWith(AndroidJUnit4::class)
+class ComboboxTest {
+    @Test
+    fun autoCorrectCapitalization_rewritesWrongCaseToCanonical() {
+        val combobox = newCombobox("Apple", "Banana", "Cherry")
+        combobox.setText("apple")
+
+        combobox.autoCorrectCapitalization()
+
+        assertEquals("Apple", combobox.text.toString())
+    }
+
+    @Test
+    fun autoCorrectCapitalization_leavesCanonicalCasingUntouched() {
+        val combobox = newCombobox("Apple", "Banana")
+        combobox.setText("Apple")
+        val originalEditable = combobox.text
+
+        combobox.autoCorrectCapitalization()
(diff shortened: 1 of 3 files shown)


app/unit-tests/src/org/commcare/views/ComboboxTest.kt
```diff
@@ -0,0 +1,87 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.annotation.Config
+import java.util.Vector
+
+@Config(application = CommCareTestApplication::class)
+@RunWith(AndroidJUnit4::class)
+class ComboboxTest {
+    @Test
+    fun autoCorrectCapitalization_rewritesWrongCaseToCanonical() {
```
shubham1g5, line 19: we should utilize natual language format for test names based on https://github.com/dimagi/open-source/pull/22/files

shubham1g5 (commented):
Code looks good, although a few questions - 

1. Is this a recent regression, if so , can we link the PR it was introduced in the description. Also think this would need a hotfix if it's a recent regression. 
2. When auto-capitialization does happen, we would still loose focus from the numeric widget. Is that intended/acceptable ?

avazirna:
> 1. Is this a recent regression, if so , can we link the PR it was introduced in the description. Also think this would need a hotfix if it's a recent regression.

@shubham1g5 we have recently migrated from `String` based dropdown list to [ComboItem](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/model/ComboItem.java) to improve handling of duplicates. During that work, this method was not updated based on the new structure. I will add the PR.

> When auto-capitialization does happen, we would still loose focus from the numeric widget. Is that intended/acceptable ?

This scenario is improbable. Auto-capitalization is also triggered when the Combobox widget loses focus, so once focus has been lost, there should no longer be a need to apply auto-capitalization.

shubham1g5 (approved):

avazirna:
Superseded by #3749

PR #3737, opened 2026-05-25
Override Connect SMS Links

OrangeAndGreen:
https://dimagi.atlassian.net/browse/CCCT-2395

## Product Description
Clicking on opportunity invite links in received SMS messages from Connect will navigate directly into the app where the user can view and accept the invitation and get started. If the opportunity is already in progress (learning or delivery status) at the time the link is clicked, the user is navigated to the learning/delivery progress page.

If the CommCare app is not installed when the user clicks the link, the device will naturally navigate to a webpage on Connect that immediately redirects them to the Google Play listing for CommCare so they can install the app. The webpage and redirect are not part of this work.

If the opportunity can not be found (including after requesting updated opportunities from the server), a toast is shown to the user saying "Opportunity not found" and they are navigated to Connect home (opportunity list).

Demos:
* [Available opportunity opens to Job Intro](https://drive.google.com/file/d/1oz0jZlUHAE_EcxMqUgrtVVm-5Z08FPBH/view?usp=drive_link)
* [Opp in learning opens to Learning Progress](https://drive.google.com/file/d/1meoRKG1pDb92eMZbXsLXamZM7hbJTOZt/view?usp=drive_link)
* [Opp in delivery opens to Delivery Progress](https://drive.google.com/file/d/1v_2NmOUaA9tfI4FaYccUv3IZd1oisdCH/view?usp=drive_link)
* [Opportunity not found shows toast and goes to Connect Home (opportunity list)](https://drive.google.com/file/d/1TddOGk9ZaUVzIomI5HAhZZh-KDIbmS5t/view?usp=drive_link)
* [PersonalID not configured opens app to landing page (Setup or Login)](https://drive.google.com/file/d/1oKk0oBMW3aUXB2blZkZwqQKYeK4aYn-Z/view?usp=drive_link)
* [User-cancelled biometric unlock navigates to landing page](https://drive.google.com/file/d/1h-yg1BISduddJ8RHPjjbety4_UzItTkv/view?usp=drive_link)
* [App not installed navigates to web page (currently a 404 pending server work)](https://drive.google.com/file/d/1kBKW35qCF3dmz1zP2Iqw_2ir15d3bQXN/view?usp=drive_link)

## Technical Summary
* Updates manifest to override links when clicked (Connect prod or staging, URL like /users/invite_redirect/<uuid>)
* Updates DispatchActivity and ConnectActivity to handle routing when launched with an opportunity UUID
* Updates ConnectUnlockFragment to receive the opportunity UUID and handle routing after unlock
* Adds analytics for success and failure scenarios

### Safety story
Since the server changes are not yet in place to support the new SMS links with "invite_redirect", I crafted invite links for the various tests manually and texted them to myself. That way I could open an SMS message containing the link and click it to navigate into the app.
Greenlight paths tested: Navigating to opportunity in available/learning/delivery status
Redlight paths tested: App not installed, PersonalID not configured, biometric unlock cancelled, opportunity not found

### Automated test coverage
None


app/src/org/commcare/activities/DispatchActivity.java
```diff
@@ -262,6 +279,53 @@ private boolean isExternalLaunch() {
...
+
+        // Expect exactly ["users", "invite_redirect", "<uuid>"]. Reject anything longer
+        // (e.g. /users/invite_redirect/<uuid>/extra) or shorter so the UUID we pass on
+        // is unambiguously a single segment.
+        List<String> segments = data.getPathSegments();
+        if (segments.size() != 3
+                || !"users".equals(segments.get(0))
+                || !"invite_redirect".equals(segments.get(1))) {
+            return false;
+        }
```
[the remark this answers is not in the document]
OrangeAndGreen: Checking scheme and the specific host configured (prod or staging) when retrieving the intent: [6efd23c](https://github.com/dimagi/commcare-android/pull/3737/commits/6efd23c7b3afd4f3b09cd0bd5454551c5ae7ec7b)

app/AndroidManifest.xml
```diff
@@ -159,6 +159,22 @@
 
                 <category android:name="android.intent.category.DEFAULT" />
             </intent-filter>
+            <intent-filter android:autoVerify="true">
```
shubham1g5, line 162: Curious if it's possible to combine the 2 intent filters here into one ?

app/src/org/commcare/activities/DispatchActivity.java
```diff
@@ -262,6 +284,50 @@ private boolean isExternalLaunch() {
                 this.getIntent().hasExtra(SESSION_ENDPOINT_ID);
     }
 
+    private Intent retrieveSmsInviteIntentIfPresent() {
```
shubham1g5: Can we move this out to a helper class given it doesn't use lot of the state from the activity and will help us keep this `pure business` logic separate from the activity itself.

app/src/org/commcare/activities/DispatchActivity.java
```diff
@@ -262,6 +284,50 @@ private boolean isExternalLaunch() {
...
+
+        // Clear the URI immediately so future dispatch() doesn't reprocess this link
+        intent.setData(null);
+
+        PersonalIdManager personalIdManager = PersonalIdManager.getInstance();
+        personalIdManager.init(this);
+        if (!personalIdManager.isloggedIn()) {
+            FirebaseAnalyticsUtil.reportSmsInviteLinkEvent(SMS_INVITE_LINK_PERSONAL_ID_NOT_CONFIGURED);
+            return null;
+        }
```
shubham1g5: Curious what are your thoughts here on having a generic event  like `external_app_launch` that can possibly track all such external launches and takes - 

1. `type` -> `OPP_INVITE_LINK`
2.  `sucess` -> `1` or `0`
3.  `error` -> `personalid_not_configured` , `opp_not_found`

app/src/org/commcare/activities/DispatchActivity.java
```diff
@@ -178,7 +190,16 @@ private void dispatch() {
                     pnIntent.getStringExtra(NOTIFICATION_ID)
             );
             startActivity(pnIntent);
-        }else if (currentApp == null) {
+            return;
+        }
+
+        Intent smsIntent = retrieveSmsInviteIntentIfPresent();
```
shubham1g5: Naming can be improved here to make it explicit that we are talking about Connect Opp Invite given we are in shared app context here- 

`smsIntent` -> `connectOppInviteIntent` 
`retrieveSmsInviteIntentIfPresent` -> `retrieveConnectOppInviteIntentIfPresent`

app/src/org/commcare/activities/connect/ConnectActivity.java
```diff
@@ -157,10 +159,17 @@ private int handleSecureRedirect(Bundle startArgs) {
                 redirectionAction = CCC_DEST_DELIVERY_PROGRESS; //  Generic push notification for delivery progress related
             }else if(job!=null && job.getStatus() == ConnectJobRecord.STATUS_LEARNING){
                 redirectionAction = CCC_DEST_LEARN_PROGRESS;    // Generic push notification for learning progress related
+            }else {
+                redirectionAction = CCC_DEST_OPPORTUNITY_SUMMARY_PAGE;    // Generic push notification for learning progress related
             }
```
shubham1g5: the comment seems incorrect here ? 

Also curious if it's better to be more explicit around what exact statuses map to this page instead of treating this as a fallback. I am mostly worried that the current code might cause un-intended redirects to this page.

app/src/org/commcare/fragments/connect/ConnectUnlockFragment.java
```diff
@@ -85,21 +95,55 @@ private void retrieveOpportunities() {
         new ConnectApiHandler<List<ConnectJobRecord>>() {
 
             @Override
-            public void onFailure(@NonNull PersonalIdOrConnectApiErrorCodes errorCode, @androidx.annotation.Nullable Throwable t) {
+            public void onFailure(@NonNull PersonalIdOrConnectApiErrorCodes errorCode,
+                                  @androidx.annotation.Nullable Throwable t) {
+                if (!isAdded()) { return; }
+                if (fromSmsInviteLink) {
+                    handleSmsLinkFailure(AnalyticsParamValue.SMS_INVITE_LINK_NETWORK_FAILURE);
+                }
```
shubham1g5: Should we do the `findRequestedJob` check here as well ? (given a network error here doesn't really guarantee that the opportunity is not present already in the DB)

app/src/org/commcare/fragments/connect/ConnectUnlockFragment.java
```diff
@@ -85,21 +95,55 @@ private void retrieveOpportunities() {
...
                 setFragmentRedirection();
             }
 
             @Override
             public void onSuccess(List<ConnectJobRecord> jobs) {
+                if (!isAdded()) { return; }
                 if (!jobs.isEmpty()) {
                     ConnectUserDatabaseUtil.turnOnConnectAccess(requireContext());
                 }
+                if (fromSmsInviteLink) {
```
shubham1g5: think it'd be great to move this to another method  - `handleOppInviteLink` or similar given there is a clear separation of logic in this block from the rest of the `onSuccess` block.

app/src/org/commcare/fragments/connect/ConnectUnlockFragment.java
```diff
@@ -85,21 +95,55 @@ private void retrieveOpportunities() {
...
+                    ConnectJobRecord requested = findRequestedJob(jobs);
+                    if (requested == null) {
+                        handleSmsLinkFailure(AnalyticsParamValue.SMS_INVITE_LINK_OPPORTUNITY_NOT_FOUND);
+                    } else {
+                        FirebaseAnalyticsUtil.reportSmsInviteLinkEvent(
+                                AnalyticsParamValue.SMS_INVITE_LINK_SUCCESS);
+                        ((ConnectActivity) requireActivity()).setActiveJob(requested);
```
shubham1g5: I would be surprised if that doesn't already happen elsewhere given notifications redirect use the same pathway and think we might be duplicating code here.

shubham1g5 (changes requested):
Think it'd be great to add a few unit tests to the PR that can verify the final activity based on the launching intent and Opp Status.

conroy-ricketts (commented):
I have no additional comments to add. Looks great!

app/src/org/commcare/fragments/connect/ConnectUnlockFragment.java
OrangeAndGreen: Yeah, it can also happen [here](https://github.com/dimagi/commcare-android/blob/26d0694a629cd454a4aa264ee4dbbfa1f43d1089/app/src/org/commcare/activities/connect/ConnectActivity.java#L124) when ConnectActivity starts up and retrieves extras. But since the job may not have been available at that point (i.e. if we just retrieved it for the first time in this API call), we may still need to populate it now. Even if the job did exist, this makes sure we're getting the most up-to-date version in memory (in case the API call updated the status of this opp)

app/AndroidManifest.xml
OrangeAndGreen: Ah, we can but it could create an issue. If we list both connect and connect-staging hosts in the same entry, then BOTH have to be confirmed (via hosted assetlinks.json files on each server) in order for the app links to work. So if bundled together, it would be possible for a change on staging to break prod.

In practice though, the assetlinks files are already hosted on both servers and are unlikely to change on their own.

I'm leaning toward staying safe and keeping the two entries apart, do you agree?

app/src/org/commcare/activities/DispatchActivity.java
OrangeAndGreen: Yeah I like that. One issue with the change is that we don't have a good way to log the initial "user tapped on SMS link" event, but as long as we cover all the end states from that click with a success or failure event then I think it's okay not to have that additional event. [5a62a96](https://github.com/dimagi/commcare-android/pull/3737/commits/5a62a9673c76358d8eab8ee35efaa1e9fa7fade4)

app/src/org/commcare/activities/connect/ConnectActivity.java
OrangeAndGreen: Ah, yup. Those comments didn't seem very useful in general so I just removed them.

Re: the logic for routing, I made the final else more explicit (for AVAILABLE and AVAILABLE_NEW statuses) but noting that all statuses are now covered so there never should be an "else" fall through. Specifically, we handle:
* Available (x2)
* Learning
* Delivery (plus a special case to go to payments)

[a4f4307](https://github.com/dimagi/commcare-android/pull/3737/commits/a4f43075b228048ead79674054602b11887d6e46)

app/src/org/commcare/activities/DispatchActivity.java
OrangeAndGreen: [3365863](https://github.com/dimagi/commcare-android/pull/3737/commits/3365863a8c232d9a3c971c6bbcc616296e478283)

app/src/org/commcare/activities/DispatchActivity.java
OrangeAndGreen: [cfda1b2](https://github.com/dimagi/commcare-android/pull/3737/commits/cfda1b216f2b5ff0ca4dd0e6ce39f071ae14be35)

app/src/org/commcare/fragments/connect/ConnectUnlockFragment.java
OrangeAndGreen: [c5a31ee](https://github.com/dimagi/commcare-android/pull/3737/commits/c5a31ee5acf469004ae46c078134753453d00437)

app/src/org/commcare/fragments/connect/ConnectUnlockFragment.java
OrangeAndGreen: Ah, good call. I also changed the code to just retrieve the job by UUID from the DB (if the API call succeeded it already updated the DB before calling back). That way we use a single source of truth for getting the data (the DB) regardless of whether the API call succeeded.
[c5a31ee](https://github.com/dimagi/commcare-android/pull/3737/commits/c5a31ee5acf469004ae46c078134753453d00437)

shubham1g5:
@OrangeAndGreen the new changes looks great to me although flagging that there are three comments from last review which have not been addressed.  Puting them here for easy reference - 

- https://github.com/dimagi/commcare-android/pull/3737#discussion_r3308572138
- https://github.com/dimagi/commcare-android/pull/3737#discussion_r3308855959
- https://github.com/dimagi/commcare-android/pull/3737#discussion_r3308887709

app/src/org/commcare/fragments/connect/ConnectUnlockFragment.java
```diff
@@ -85,21 +96,55 @@ private void retrieveOpportunities() {
...
+                    requireContext(), requestedOpportunityUuid);
+            if (requested == null) {
+                handleOppInviteLinkFailure(AnalyticsParamValue.OPP_INVITE_LINK_OPPORTUNITY_NOT_FOUND);
+            } else {
+                FirebaseAnalyticsUtil.reportExternalAppLaunchEvent(
+                        AnalyticsParamValue.OPP_INVITE_LINK, true, null);
+                ((ConnectActivity) requireActivity()).setActiveJob(requested);
+            }
+        }
+    }
```
Jignesh-dimagi: Already the job is set in ConnectActivity [here](https://github.com/dimagi/commcare-android/blob/47847dc775b2082a0d31d58fe7c527f4d8ee244c/app/src/org/commcare/activities/connect/ConnectActivity.java#L121) before coming to this fragment. Can we try to load only if job == null?

app/src/org/commcare/fragments/connect/ConnectUnlockFragment.java
```diff
@@ -85,21 +96,55 @@ private void retrieveOpportunities() {
...
-                setFragmentRedirection();
 
+                tryToLoadInvitedOpp();
+                setFragmentRedirection();
             }
         }.getConnectOpportunities(requireContext(), user);
     }
 
+    private void tryToLoadInvitedOpp() {
+        if (fromOppInviteLink) {
```
Jignesh-dimagi: I think it should not check this condition; I feel like it might be useful for the users coming from a new opportunity push notification as well.

app/src/org/commcare/connect/ConnectJobHelper.kt
```diff
@@ -174,4 +177,87 @@ object ConnectJobHelper {
...
     }
+
+    fun resolveGenericOpportunityDestination(
+        currentAction: String?,
+        job: ConnectJobRecord?,
+        paymentUuid: String?,
+    ): String? {
+        if (ConnectConstants.CCC_GENERIC_OPPORTUNITY != currentAction || job == null) {
+            return currentAction
+        }
```
Jignesh-dimagi, line 184: For the new opportunity invite, this will be true most of the time, as users might not have synced opportunities. Whenever 'fromOppInviteLink' is true in 'ConnectUnlockFragment' and if ConnectActivity's job is null, I think the app should try to find the redirection again after syncing the opportunity list.

app/src/org/commcare/connect/ConnectJobHelper.kt
```diff
@@ -174,4 +177,87 @@ object ConnectJobHelper {
...
+
+        // Require /users/invite_redirect/<opp_uuid>
+        val segments = data.pathSegments
+        if (segments.size != 3 ||
+            ("users" != segments[0]) ||
+            ("invite_redirect" != segments[1])
+        ) {
+            return null
+        }
+        val uuid = segments[2]
```
Jignesh-dimagi: This works well for this SMS, but to make it more modular and maintainable, can we create an SMSHelper that can parse different SMS messages and extract their relevant intents?

app/src/org/commcare/fragments/connect/ConnectUnlockFragment.java
OrangeAndGreen: See my response [here](https://github.com/dimagi/commcare-android/pull/3737#discussion_r3320380539), but I think it's worth getting an updated version of the job from the DB even if we already had it before, since the successful API call may have updated the status of the job (i.e. from learning to delivery if the user passed the assessment since the last data refresh).

app/AndroidManifest.xml
shubham1g5: interesting, that makes sense to me and thanks for investigating it.

app/src/org/commcare/fragments/connect/ConnectUnlockFragment.java
shubham1g5: the refactor seems good to me although think we should only log `SMS_INVITE_LINK_NETWORK_FAILURE` if the opportunity was not found in local DB.

app/src/org/commcare/fragments/connect/ConnectUnlockFragment.java
shubham1g5: that makes sense to me.

app/unit-tests/src/org/commcare/connect/SmsInviteLinkFlowTest.kt
```diff
@@ -0,0 +1,235 @@
...
+            "Dispatch did not route SMS link to ConnectActivity",
+            ConnectActivity::class.java.name,
+            connectIntent.component?.className,
+        )
+
+        assertTrue(connectIntent.getBooleanExtra(ConnectConstants.FROM_SMS_INVITE_LINK, false))
+        assertTrue(connectIntent.getBooleanExtra(ConnectConstants.SHOW_LAUNCH_BUTTON, false))
+        assertEquals(uuid, connectIntent.getStringExtra(ConnectConstants.OPPORTUNITY_UUID))
+
+        return ConnectJobHelper.resolveGenericOpportunityDestination(
```
shubham1g5: think we should not need to call this method here but instead just verify the destination based on the resulting fragment.

app/src/org/commcare/fragments/connect/ConnectUnlockFragment.java
OrangeAndGreen: My understanding is that the code called here isn't necessary for the push notification workflow, since the API call to update opportunities is performed synchronously and the two related variables (active job and redirectionAction) that this function would update should already have been set previously (in ConnectActivity.initStateFromExtras).

I could allow the code to run anyway since technically retrieving these items again won't hurt anything, but if we do that there are some considerations:
1. Analytics is probably the biggest question. Do we want the new analytics events related to "opportunity links" to fire for both SMS and push notification clicks? If so, thinking we may want to include another parameter to the event indicating which workflow we're in.
2. Since launches from push notifications would come through this flow even if the notification doesn't include an opportunity UUID, I think we'd want to change the existing "if" to check that requestedOpportunityUuid is not empty, otherwise we'll incorrectly fire the OPP_INVITE_LINK_OPPORTUNITY_NOT_FOUND error and clear the redirectionAction.

Your thoughts?

app/src/org/commcare/connect/ConnectJobHelper.kt
OrangeAndGreen: Ah, great catch! Agreed, we should try updating the redirectionAction again if we're able to retrieve the job after data refresh in case we didn't have the job before (or job status changed with the latest data refresh). [45dc794](https://github.com/dimagi/commcare-android/pull/3737/commits/45dc794519d01654c62fc2fffc87bce33cd1a9ff)

app/src/org/commcare/connect/ConnectJobHelper.kt
OrangeAndGreen: Yeah that sounds good, although I decided to call it DeepLinkHelper since the links are not specific to SMS (they can be clicked from anywhere). [9290ca0](https://github.com/dimagi/commcare-android/pull/3737/commits/9290ca0a9f28cef34af1a761622f4eb9b0ec8756)

app/src/org/commcare/fragments/connect/ConnectUnlockFragment.java
OrangeAndGreen: Ah, ok, I thought it might be nice to report since a failure here could result in navigating to an outdated state for the app but otherwise wouldn't be clearly reported. But it also results in a possible double-logging of the failure, i.e. once saying we couldn't refresh and a possible second saying the job couldn't be found.

No issue on my end making the change if that's what you prefer: [1ab5771](https://github.com/dimagi/commcare-android/pull/3737/commits/1ab5771f049f4fbcc28c83426b76419dff3220c0)

shubham1g5: thanks, the commit looks good to me, I was mostly concerned that we would report `SMS_INVITE_LINK_NETWORK_FAILURE` even though there is no user-facing failure in the corresponding workflow.

app/src/org/commcare/fragments/connect/ConnectUnlockFragment.java
Jignesh-dimagi: 1. I think we can log analytics based on the variable 'fromOppInviteLink'.
2. When a push notification navigates to ConnectActivity, it must have an opportunity UUID. Additionally, to log a failure, we can reuse the 'fromOppInviteLink' variable.

app/unit-tests/src/org/commcare/connect/SmsInviteLinkFlowTest.kt
OrangeAndGreen: [f5b8dfd](https://github.com/dimagi/commcare-android/pull/3737/commits/f5b8dfd1523eef52d3d53a8743dea4f47a49eca7)

app/unit-tests/src/org/commcare/connect/SmsInviteLinkFlowTest.kt
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    @Test
+    @Config(sdk = [Build.VERSION_CODES.Q])
+    fun `SMS link with unknown job status falls back to generic opportunity destination`() {
+        assertEquals(
+            ConnectConstants.CCC_GENERIC_OPPORTUNITY,
+            launchSmsInvite(validUri, jobWithStatus(ConnectJobRecord.STATUS_ALL_JOBS)),
+        )
+    }
```
conroy-ricketts, line 126: Confirming that a job with the status `STATUS_ALL_JOBS` is meant to be an "unkown" status?

app/unit-tests/src/org/commcare/connect/SmsInviteLinkFlowTest.kt
```diff
@@ -0,0 +1,301 @@
...
+
+        val navHost =
+            connect.supportFragmentManager.findFragmentById(R.id.nav_host_fragment_connect)
+                as NavHostFragment
+        return navHost.navController.currentBackStackEntry
+            ?.arguments
+            ?.getString(ConnectConstants.REDIRECT_ACTION)
+    }
+
+    private fun assertRoutedToLandingPage(intent: Intent) {
```
conroy-ricketts: Nit: thoughts on renaming this function? On first read, "landing page" could be mistaken for either the CC app home screen or the Connect opportunities list

conroy-ricketts (dismissed):

app/unit-tests/src/org/commcare/connect/SmsInviteLinkFlowTest.kt
OrangeAndGreen: [5abba70](https://github.com/dimagi/commcare-android/pull/3737/commits/5abba703c107bcd361a15e17841222344e8a4213)

app/unit-tests/src/org/commcare/connect/SmsInviteLinkFlowTest.kt
OrangeAndGreen: A job will never have STATUS_ALL_JOBS for its status, that value is only used as an additional option when retrieving jobs from the DB (i.e. to retrieve all jobs as opposed to only jobs that match a given status). This test uses the value just as a way to test that an unrecognized status navigates correctly

app/src/org/commcare/fragments/connect/ConnectUnlockFragment.java
OrangeAndGreen: [b28b203](https://github.com/dimagi/commcare-android/pull/3737/commits/b28b2038c205f8f7ad0844336eccacdd0ec81d74)

Jignesh-dimagi (dismissed):

app/unit-tests/src/org/commcare/connect/SmsInviteLinkFlowTest.kt
shubham1g5: nit: we can use something hardcoded like a `unknown-status` to make it more clear

app/unit-tests/src/org/commcare/connect/SmsInviteLinkFlowTest.kt
```diff
@@ -0,0 +1,302 @@
...
+    @Config(sdk = [Build.VERSION_CODES.Q])
+    fun `SMS link with no matching job in DB falls back to generic opportunity destination`() {
+        assertEquals(
+            ConnectConstants.CCC_GENERIC_OPPORTUNITY,
+            launchSmsInvite(validUri, jobInDb = null),
+        )
+    }
+
+    @Test
+    fun `SMS link redirection is re-resolved against synced job after opportunities load`() {
```
shubham1g5: @OrangeAndGreen this test seems like is using old `resolveGenericOpportunityDestination` based approach rather than letting the natural UI flow happen with help of mocks.

shubham1g5 (dismissed):
Nothing blocking, left a comment around testing strategy for one test.

conroy-ricketts (dismissed):

app/unit-tests/src/org/commcare/connect/SmsInviteLinkFlowTest.kt
OrangeAndGreen: Ah, good catch, I was only looking at the other test that called that function before.

The refactor for this one essentially just resulted in calling launchSmsInvite twice and verifying the correct destination each time: first with a null BD object for the opp, then with a valid opp in delivery status. Since all of those flows are tested elsewhere already I decided this test wasn't going to add any value, so deleted it. [b6b13d6](https://github.com/dimagi/commcare-android/pull/3737/commits/b6b13d6682a799936a4366fb5cce6ebd6916c65e)

conroy-ricketts (dismissed):

shubham1g5 (dismissed):

OrangeAndGreen:
This is blocked on instrumentation tests, which are failing due to a deprecated device (Samsung S10)

conroy-ricketts (approved):

PR #3739, opened 2026-05-27
Ccct 2321:  integrity delay UX Issue

shubham1g5:
## Product Description
https://dimagi.atlassian.net/browse/CCCT-2321

Before (with a hardcoded delay into Integirty Token request)

https://github.com/user-attachments/assets/1bca483e-bd9b-4107-99b2-bbfb2fd979af

After (with a hardcoded delay into Integirty Token request)

https://github.com/user-attachments/assets/1343b17f-be28-4f51-a05a-cdfc37dd2346

## Technical Summary

1. User clicks Continue → onContinueClicked() disables the button and calls startConfigurationRequest()
2. While the integrity token request is pending, onLocationResult() / onLocationServiceChange() / the phone text watcher / consent checkbox can all fire updateContinueButtonState() → which re-enables the button if allowedToContinue() returns true

The fix is to track isRequestInProgress and guard updateContinueButtonState() against it.

### Safety story

- Verified that only one dialog appears at a moment even if we called `showProgressDialogIfNeeded` twice with the same task id
- Verfied that the dialog shows up correctly for rest of the fragments in the Personal ID flow. 

## Labels and Review

Best Reviewed Commit by Commit

### DIFF
--- a/app/src/org/commcare/fragments/personalId/PersonalIdPhoneFragment.java
+++ b/app/src/org/commcare/fragments/personalId/PersonalIdPhoneFragment.java
@@ -81,6 +81,7 @@ public class PersonalIdPhoneFragment extends BasePersonalIdFragment implements C
     private ActivityResultLauncher<IntentSenderRequest> resolutionLauncher;
     private String playServicesError;
     private ActivityResultLauncher<IntentSenderRequest> playServicesResolutionLauncher;
+    private boolean isRequestInProgress = false;
 
     private static final String[] REQUIRED_PERMISSIONS = new String[]{
             Manifest.permission.ACCESS_COARSE_LOCATION, Manifest.permission.ACCESS_FINE_LOCATION
@@ -279,7 +280,9 @@ public class PersonalIdPhoneFragment extends BasePersonalIdFragment implements C
     }
 
     private void updateContinueButtonState() {
-        enableContinueButton(allowedToContinue());
+        if (!isRequestInProgress) {
+            enableContinueButton(allowedToContinue());
+        }
     }
 
     private boolean allowedToContinue() {
@@ -325,6 +328,7 @@ public class PersonalIdPhoneFragment extends BasePersonalIdFragment implements C
     }
 
     private void startConfigurationRequest() {
+        isRequestInProgress = true;
         clearError();
         phone = PhoneNumberHelper.buildPhoneNumber(
                 binding.countryCode.getText().toString(),
@@ -493,6 +497,7 @@ public class PersonalIdPhoneFragment extends BasePersonalIdFragment implements C
         new PersonalIdApiHandler<PersonalIdSessionData>() {
             @Override
             public void onSuccess(PersonalIdSessionData sessionData) {
+                isRequestInProgress = false;
                 personalIdSessionDataViewModel.getPersonalIdSessionData().setPhoneNumber(phone);
 
                 FirebaseAnalyticsUtil.flagPersonalIDDemoUser(sessionData.getDemoUser());
@@ -519,6 +524,7 @@ public class PersonalIdPhoneFragment extends BasePersonalIdFragment implements C
                     @NonNull PersonalIdOrConnectApiErrorCodes failureCode,
                     @Nullable Throwable t
             ) {
+                isRequestInProgress = false;
                 if (handleCommonSignupFailures(failureCode)) {
                     return;
                 }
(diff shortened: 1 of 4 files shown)


conroy-ricketts (dismissed):

Jignesh-dimagi:
@shubham1g5 
The changes look good. One question: The mentioned ticket has also asked for the following:
There should be a timer running for the integrity token fetch; if it exceeds that limit, the app should call the start configuration API without an integrity token.

I am not sure if we want to implement that timer; I mean, is it worth putting efforts there? Though, I have seen Aathithya's mobile was really taking a long time just to fetch the token.

app/src/org/commcare/fragments/personalId/PersonalIdPhoneFragment.java
```diff
@@ -561,6 +570,15 @@ public void onFailure(
         }.makeStartConfigurationCall(requireActivity(), body, token, requestHash, newSessionData);
     }
 
+    private void setRequestProgressState(boolean inProgress) {
+        isRequestInProgress = inProgress;
+        if (isRequestInProgress) {
+            ((CommCareActivity<?>)requireActivity()).showProgressDialogIfNeeded(NETWORK_ACTIVITY_ID);
+        } else {
+            ((CommCareActivity<?>)requireActivity()).dismissProgressDialogForTask(NETWORK_ACTIVITY_ID);
```
Jignesh-dimagi: Do we need this dismissal here? It is already getting dismissed in BaseApi [here](https://github.com/dimagi/commcare-android/blob/26d0694a629cd454a4aa264ee4dbbfa1f43d1089/app/src/org/commcare/connect/network/base/BaseApi.kt#L35) and [here](https://github.com/dimagi/commcare-android/blob/26d0694a629cd454a4aa264ee4dbbfa1f43d1089/app/src/org/commcare/connect/network/base/BaseApi.kt#L61).

shubham1g5: I wanted to cover all possible scenarios here and be symmetric in execution i.e. if this fragment is starting something it should also clean it up  without relying on other classes.

shubham1g5:
> There should be a timer running for the integrity token fetch; if it exceeds that limit, the app should call the start configuration API without an integrity token.

@Jignesh-dimagi I think I am quite comfortable not implementing a timer here given we are showing a loading bar now and we have only really heard it taking long for one user. I would also be surprised if the API internally doesn't implement a timeout although i was not able to find anything in docs reg. that.

Jignesh-dimagi:
> @Jignesh-dimagi I think I am quite comfortable not implementing a timer here given we are showing a loading bar now and we have only really heard it taking long for one user. I would also be surprised if the API internally doesn't implement a timeout although i was not able to find anything in docs reg. that.

The delay was solely caused by the integrity token fetch. We didn't see a progress dialog during this time, but one appeared after a long delay, indicating the API request had started. However, since we've only seen this behavior with a single user, we can avoid implementing a timer.

Jignesh-dimagi (dismissed):

shubham1g5:
@Jignesh-dimagi Curious what do you mean by super long here -  less than 2 min or more like 5-10 min ?

Jignesh-dimagi:
> @Jignesh-dimagi Curious what do you mean by super long here - less than 2 min or more like 5-10 min ?

@shubham1g5 
The delay was definitely in the range of 1-2 mins and not more, but it was enough to break the patience of the user, as there was no visual loading, and on top of that, the 'continue' button getting enabled in between, tempting them to press again.

conroy-ricketts (approved):

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdPhoneFragmentStartConfigurationTest.kt
```diff
@@ -159,7 +159,7 @@ class PersonalIdPhoneFragmentStartConfigurationTest : BasePersonalIdPhoneFragmen
 
         // Assert
         mockWebServer.takeRequest()
-        ShadowLooper.runUiThreadTasksIncludingDelayedTasks()
+        ShadowLooper.idleMainLooper()
```
OrangeAndGreen, line 162: Can you address this change and why it was made in the PR description?

OrangeAndGreen (approved):
All looks good to me, just curious to learn more about the change in test code to use idleMainLooper

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdPhoneFragmentStartConfigurationTest.kt
shubham1g5: it's mostly a nit/no-op change given `runUiThreadTasksIncludingDelayedTasks` is un-ncessary here given we don't have any `delayed` tasks at the moment and can just run a ligter `idleMainLooper` version that achieves what we want to do here is to just complete any exisitng async processes. 

I don't think it fits well with PR story but can add to the description if you thing otherwise.

PR #3743, opened 2026-05-29
CCCT-2438 Extract Post-Login Routing And Inline App Seating

conroy-ricketts:
## Technical Summary

Phase 2 of the login/Connect decoupling (CCCT-2164 plan), stacked on the `CCCT-2437-headless-login-engine` branch.

- Introduces `PostLoginRouter` / `PostLoginDestination`: a pure, Android-free mapping from a `LoginResult` + `LaunchContext` to a navigation destination. `DispatchActivity`'s `LOGIN_USER` result handler now delegates to it instead of unpacking intent extras inline.
- Extracts app seating out of `SeatAppActivity`'s `Thread`/`Handler` into `AppSeater`, a lifecycle-scoped coroutine wrapper with injectable record-lookup / seat / dispatcher; `SeatAppActivity` becomes a thin UI host.

### Safety story

What gives me confidence:
- I ran the login → home flow on an emulator/device and confirmed the app still seats and reaches the home screen.
- New unit coverage for the extracted logic: `PostLoginRouter` mapping (including negative cases) and `AppSeater` success / not-found / corrupted / exception paths.
- The diff is a contained restructure of the login → seat → dispatch path, with no schema or data-model changes.

Risks to review:
- The `DispatchActivity` routing integration and the `SeatAppActivity` host are exercised manually, not by automated tests.
- `SeatAppActivity` drops the old `savedInstanceState` (`KEY_IN_PROGRESS`) rotation guard; seating now relies on `lifecycleScope` cancellation, so re-seating behaviour on config change / process recreation is worth a look.
- This is an incremental phase: `PostLoginRouter` does not yet own the full Connect-vs-home routing decision (that still lives in `dispatch()`), so today it maps only success → Home and failure → TerminalFailure.

### Automated test coverage

- `PostLoginRouterTest` — success routes to `Home` carrying login mode and launch context; asserts `redirectToConnectOpportunityInfo` does not leak into the `Home` destination; each `LoginError` maps to `TerminalFailure`.
- `AppSeaterTest` — missing record → `APP_NOT_FOUND` without seating; ready state → `Success`; corrupted state → `CORRUPTED`; thrown exceptions from both `seatApp` and `recordLookup` propagate (fail-fast); emits `Seating` progress.

[CCCT-2438]: https://dimagi.atlassian.net/browse/CCCT-2438?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


conroy-ricketts:
## Suggested Review Order

- `app/src/org/commcare/login/PostLoginDestination.kt` — the output contract (sealed `Home` / `TerminalFailure`); read first.
- `app/src/org/commcare/login/PostLoginRouter.kt` — pure mapping that produces a destination from `LoginResult` + `LaunchContext`.
- `app/unit-tests/src/org/commcare/login/PostLoginRouterTest.kt` — documents the routing behaviour, including what must *not* leak into `Home`.
- `app/src/org/commcare/login/AppSeater.kt` — extracted seating core: injectable deps, coroutine wrapper, exception → `SEAT_ERROR` guard.
- `app/unit-tests/src/org/commcare/login/AppSeaterTest.kt` — covers the seating outcomes incl. the new exception paths.
- `app/src/org/commcare/activities/SeatAppActivity.java` — now a thin host consuming `AppSeater`; note the dropped `KEY_IN_PROGRESS` guard.
- `app/src/org/commcare/activities/DispatchActivity.java` — integrates `PostLoginRouter` at the `LOGIN_USER` result boundary; read last.

conroy-ricketts:
@coderabbitai full review

app/src/org/commcare/activities/DispatchActivity.java
```diff
@@ -515,12 +519,8 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
                 if (resultCode == RESULT_CANCELED) {
                     shouldFinish = true;
                 } else if (intent != null) {
-                    lastLoginMode = (LoginMode)intent.getSerializableExtra(LoginActivity.LOGIN_MODE);
-                    userManuallyEnteredPasswordMode =
-                            intent.getBooleanExtra(LoginActivity.MANUAL_SWITCH_TO_PW_MODE, false);
-                    personalIdManagedLogin = intent.getBooleanExtra(PERSONALID_MANAGED_LOGIN, false);
                     connectManagedLogin = intent.getBooleanExtra(CONNECT_MANAGED_LOGIN, false);
```
shubham1g5, line 573: should we also move this and expression for `redirectToConnectOpportunityInfo`   inside `routeLoginResult` ? (to localise these to where they are used and be more consistent around where are we processing the intent for extras)

app/src/org/commcare/activities/DispatchActivity.java
```diff
@@ -540,4 +540,32 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
+                restoreSession,
+                personalIdManaged,
+                connectManagedLogin,
+                new PostLoginOutcome(redirectToConnectOpportunityInfo)
+        );
+
+        return PostLoginRouter.route(success, new LaunchContext(true, manualSwitchToPwMode));
+    }
+
+    private void applyPostLoginDestination(PostLoginDestination destination) {
```
shubham1g5: I am a bit concerned that we might be over-engineering or over-abstracting things here.  What `PostLoginDestination` and `PostLoginRouter` components seems to achieve in the end is just the assignement of the variables below  after passing each other all the context around and think this code was much simpler and straightforward without use of these classes. 

Do you know if there was a particular reason we wanted to abstract these out like this for eg. do we want to use the same components elsewhere ?

app/src/org/commcare/login/AppSeater.kt
```diff
@@ -0,0 +1,76 @@
...
+    APP_NOT_FOUND,
+    CORRUPTED,
+    SEAT_ERROR,
+}
+
+class AppSeater
+    @JvmOverloads
+    constructor(
+        private val recordLookup: (String) -> ApplicationRecord? = { MultipleAppsUtil.getAppById(it) },
+        private val seatApp: (ApplicationRecord) -> Int = { record ->
```
shubham1g5, line 32: Curious if these 2 methods are defined in constructor on purpose ? The only reason I would think of defining these 2 methods in constructor is if we want to switch these lambda implementation by something else.  

As such, It might be cleaner to just have these 2 separate methods in this class instead but curious if you feel otherwise and if you think if I am just bringing up something stylistic.

app/src/org/commcare/activities/SeatAppActivity.java
```diff
@@ -20,86 +17,31 @@
...
+        setResult(treatAsSeated ? RESULT_OK : RESULT_CANCELED, new Intent(getIntent()));
+        finish();
     }
 
     @Override
     public void onBackPressed() {
         // Make it impossible to quit in the middle of this activity
     }
-
-    private void setInProgress(boolean b) {
```
shubham1g5, line 68: wow, this code was written with a seriously old Java pattern, good riddance!

app/src/org/commcare/login/AppSeater.kt
```diff
@@ -0,0 +1,76 @@
...
+                    SeatResult.Failed(SeatFailure.CORRUPTED)
+                } else {
+                    SeatResult.Success
+                }
+            } catch (e: CancellationException) {
+                throw e
+            } catch (e: Exception) {
+                Logger.exception("Unexpected error while seating app $appId", e)
+                SeatResult.Failed(SeatFailure.SEAT_ERROR)
+            }
```
shubham1g5: I would default to not catching a generic exception here and have the same fail-fast behaviour as earlier code.

app/src/org/commcare/activities/SeatAppActivity.java
```diff
@@ -20,86 +17,31 @@
...
-        super.onSaveInstanceState(outState);
-        outState.putBoolean(KEY_IN_PROGRESS, inProgress);
+    private void finishWithResult(SeatResult result) {
+        // A corrupted seat still returns RESULT_OK; DispatchActivity detects
+        // STATE_CORRUPTED and routes to recovery.
+        boolean treatAsSeated = result instanceof SeatResult.Success
+                || (result instanceof SeatResult.Failed
+                && ((SeatResult.Failed)result).getReason() == SeatFailure.CORRUPTED);
+        setResult(treatAsSeated ? RESULT_OK : RESULT_CANCELED, new Intent(getIntent()));
+        finish();
```
shubham1g5: think `RESULT_CANCELED` doesn't really convey to the caller (dispatch) that an error has occurred, instead I would think we should pass `RESULT_OK` irrespective of the result and caller can verify the app state and take actions based on the app state ?

shubham1g5 (changes requested):
Have not reviewed tests yet as wanted to get some clarification on design before diving into those.

app/src/org/commcare/activities/SeatAppActivity.java
conroy-ricketts: Yeah I think that's a better approach

app/src/org/commcare/login/AppSeater.kt
conroy-ricketts: Agreed

app/src/org/commcare/activities/SeatAppActivity.java
conroy-ricketts: Also noting here that I think this function was an oversight on my part. The caller would actually be the `LoginActivity`, and it does nothing with `RESULT_CANCELED`

app/src/org/commcare/login/AppSeater.kt
conroy-ricketts: https://github.com/dimagi/commcare-android/commit/d165331abc56d60f248b1dcee26625bc7bb66173

app/src/org/commcare/activities/SeatAppActivity.java
conroy-ricketts: https://github.com/dimagi/commcare-android/commit/d165331abc56d60f248b1dcee26625bc7bb66173

app/src/org/commcare/activities/DispatchActivity.java
conroy-ricketts: Yeah I agree that it looks over-engineered at this point, but the reason for this is to set the code up for phase 3

In phase 3, `ConnectAppLauncher` will route the real `LoginResult` (rather than the fake one seen here) through the same `PostLoginRouter`

app/src/org/commcare/activities/DispatchActivity.java
conroy-ricketts: I'm hesitant to move the expression for `redirectToConnectOpportunityInfo` because that is not only used for logins, but also for when the user taps the "View Job Status" button on a CC app home screen

But, I like the idea of moving `connectManagedLogin`!

conroy-ricketts: https://github.com/dimagi/commcare-android/commit/94d61c1a75cdc2dd04c23f65a9628fc70b557bd9

app/src/org/commcare/login/AppSeater.kt
conroy-ricketts: My only reason for this is unit testing

So yes, the idea here is that the tests can easily swap these method implementations for fakes (no need for a real CC app)

shubham1g5 (dismissed):

app/src/org/commcare/activities/SeatAppActivity.java
```diff
@@ -20,86 +16,26 @@
...
+        String appId = getIntent().getStringExtra(KEY_APP_TO_SEAT);
+        new AppSeater().start(this, appId, progress -> {}, this::finishWithResult);
     }
 
-    @Override
-    protected void onSaveInstanceState(Bundle outState) {
-        super.onSaveInstanceState(outState);
-        outState.putBoolean(KEY_IN_PROGRESS, inProgress);
+    private void finishWithResult(SeatResult result) {
+        setResult(RESULT_OK, new Intent(getIntent()));
```
OrangeAndGreen, line 39: Should the result be applied to the result code to pass a seating failure back to the caller?

OrangeAndGreen (dismissed):
All looks good to me, just one question to consider

app/src/org/commcare/activities/SeatAppActivity.java
conroy-ricketts: This was intentional and there's some additional context in [this](https://github.com/dimagi/commcare-android/pull/3743#discussion_r3334907055) comment thread, but callers currently do not handle a failure result directly

I'm going to make a note to revisit this in phase 3 to see if there's a better way to handle this

OrangeAndGreen (approved):

OrangeAndGreen (changes requested):
While reviewing the next PR after this one (3753) I noticed we remove two of the classes introduced in this PR. Given that this one is still open, it seems like it would make more sense to remove those classes here rather than in a follow-up PR

OrangeAndGreen (approved):

PR #3744, opened 2026-06-01
Logs phone number when there is an error parsing

shubham1g5:
No User visible change.

Adding a log to get more info on the [errors here](https://console.firebase.google.com/u/0/project/commcare-a57e4/crashlytics/app/android:org.commcare.dalvik/issues/920fecd71166d9c2a6363b9c5b336fdd)

### DIFF
--- a/app/src/org/commcare/utils/PhoneNumberHelper.java
+++ b/app/src/org/commcare/utils/PhoneNumberHelper.java
@@ -62,14 +62,14 @@ public class PhoneNumberHelper {
      * Validates whether the given phone number is valid.
      */
     public boolean isValidPhoneNumber(String phone) {
+        if (isEmpty(phone)) {
+            return false;
+        }
         try {
             Phonenumber.PhoneNumber phoneNumber = phoneNumberUtil.parse(phone, null);
             return phoneNumberUtil.isValidNumber(phoneNumber);
         } catch (NumberParseException e) {
-            String redactedPhone = "";
-            if (!isEmpty(phone)) {
-                redactedPhone = phone.length() < 4 ? "<redacted>" : "***" + phone.substring(phone.length() - 4);
-            }
+            String redactedPhone = phone.length() < 4 ? "<redacted>" : "***" + phone.substring(phone.length() - 4);
             Logger.log(LogTypes.TYPE_MAINTENANCE, "Parse exception for phone " + redactedPhone);
             Logger.exception("Exception occurred while verifying phone number", e);
             return false;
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/utils/PhoneNumberHelper.java
```diff
@@ -63,6 +64,7 @@ public boolean isValidPhoneNumber(String phone) {
             Phonenumber.PhoneNumber phoneNumber = phoneNumberUtil.parse(phone, null);
             return phoneNumberUtil.isValidNumber(phoneNumber);
         } catch (NumberParseException e) {
+            Logger.log(LogTypes.TYPE_MAINTENANCE, "Parse exception for phone " + phone);
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3744/commits/ecc29b09044bcd0cbabb0a6d411253bdf13a75c4

app/src/org/commcare/utils/PhoneNumberHelper.java
```diff
@@ -63,6 +66,11 @@ public boolean isValidPhoneNumber(String phone) {
             Phonenumber.PhoneNumber phoneNumber = phoneNumberUtil.parse(phone, null);
             return phoneNumberUtil.isValidNumber(phoneNumber);
         } catch (NumberParseException e) {
+            String redactedPhone = "";
+            if (!isEmpty(phone)) {
+                redactedPhone = phone.length() < 4 ? "<redacted>" : "***" + phone.substring(phone.length() - 4);
+            }
+            Logger.log(LogTypes.TYPE_MAINTENANCE, "Parse exception for phone " + redactedPhone);
             Logger.exception("Exception occurred while verifying phone number", e);
```
Jignesh-dimagi: I don't think we should log an error here when the phone number is empty. Since this function is called for every location result, and users may not have started entering a phone number yet, an empty field triggers a NumberParseException.

shubham1g5: agree - https://github.com/dimagi/commcare-android/pull/3744/commits/bb7030cef5e3681e69790c419741844dcd05b961

Jignesh-dimagi (approved):

app/src/org/commcare/utils/PhoneNumberHelper.java
avazirna: It seems that this exception is thrown when the [length of the phone number is less than 2](https://github.com/MichaelRocks/libphonenumber-android/blob/38f04b163b7ab95d114b93686fab82211ebdfa84/library/src/main/java/io/michaelrocks/libphonenumber/android/PhoneNumberUtil.java#L3360). Is the goal to know which numbers are these?

shubham1g5: I mostly wanted to see why there are so many of them, think what Jignesh pointed out was probably the main culprit here but think I mostly wanted to verify if we are not hitting this error for anything that's genuine.

conroy-ricketts (approved):

PR #3745, opened 2026-06-02
CCCT-2447 Add Firebase Analytics For Add-Email Flows

Jignesh-dimagi:
## Technical Summary

Instruments the PersonalID add-email flows (signup / recovery / existing-user) with Firebase analytics. OTP request/verify reporting is centralized in `EmailHelper.sendEmailOtp`/`verifyEmailOtp`; the `attempts` count is threaded into `verifyEmailOtp` as a parameter. New email-specific OTP event types/method (`OTP_EVENT_TYPE_REQUEST_EMAIL` / `VERIFY_EMAIL`, `OTP_METHOD_EMAIL`) and `OtpOp` enum values reuse the existing `reportOtpEvent` path. Continue/skip events use `reportPersonalIDContinueClicked` for the existing-user offer prompt and the proceed-without-email path.

### Safety story

What gives me confidence:
- I ran the unit tests: the `OtpAnalyticsMapper` cases for the new email event types and the new `EmailHelper` emission tests (success/failure for both request and verify) all pass.
- Narrow, additive diff — analytics calls only; no change to email send/verify behavior or control flow.
- Request/verify reporting lives in a single chokepoint (`EmailHelper`), reducing the chance of missed or duplicated events.

### Automated test coverage

- `OtpAnalyticsMapperTest`: new cases asserting `REQUEST_EMAIL` → `request_email` and `VERIFY_EMAIL` → `verify_email`.
- `EmailHelperTest`: new cases that capture the API callback and assert `reportEmailOtpEvent` is emitted with the correct op/outcome/reason/attempts on success and failure for both `sendEmailOtp` and `verifyEmailOtp` (including the threaded `failedAttempts` value).

[CCCT-2447]: https://dimagi.atlassian.net/browse/CCCT-2447?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Jignesh-dimagi:
## Suggested Review Order

- `app/src/org/commcare/google/services/analytics/AnalyticsParamValue.java` — new event-type/method constants the rest depend on
- `app/src/org/commcare/utils/OtpAnalyticsMapper.kt` — new `OtpOp` values and their event-type mapping
- `app/src/org/commcare/google/services/analytics/FirebaseAnalyticsUtil.java` — `reportEmailOtpEvent` wrapper over `reportOtpEvent`
- `app/src/org/commcare/fragments/personalId/EmailHelper.kt` — core integration: request/verify reporting centralized here
- `app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt` — consumer; threads `failedAttempts`, reports proceed-without-email
- `app/src/org/commcare/connect/EmailOfferHelper.kt` — consumer; offer-prompt add/skip events
- `app/unit-tests/src/org/commcare/utils/OtpAnalyticsMapperTest.kt` — covers the new mappings
- `app/unit-tests/src/org/commcare/fragments/personalId/EmailHelperTest.kt` — covers the emission behavior

app/src/org/commcare/google/services/analytics/FirebaseAnalyticsUtil.java
```diff
@@ -771,6 +775,29 @@ public static void reportOtpEvent(String eventType,
...
+     * @param attempts attempt counter; logged to {@link FirebaseAnalytics.Param#VALUE}.
+     */
+    public static void reportEmailOtpEvent(OtpAnalyticsMapper.OtpOp op,
+                                           String outcome,
+                                           @Nullable String reason,
+                                           int attempts) {
+        String eventType = OtpAnalyticsMapper.getEventType(op);
+        if (eventType == null) {
+            return;
+        }
```
shubham1g5, line 792: can we log a non-fatal here, or just remove this statement and put the whole metod in a try-catch that logs a non-fatal from catch.

app/src/org/commcare/google/services/analytics/FirebaseAnalyticsUtil.java
```diff
@@ -771,6 +775,29 @@ public static void reportOtpEvent(String eventType,
...
+        reportOtpEvent(eventType, outcome, AnalyticsParamValue.OTP_METHOD_EMAIL, reason, attempts);
```
shubham1g5: can we also log what email workflow we are in

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
```diff
@@ -162,6 +162,7 @@ class PersonalIdEmailVerificationFragment : BasePersonalIdFragment() {
             otp = otp,
             workflow = workflow,
             sessionData = personalIdSessionData,
+            failedAttempts = failedOtpAttempts,
```
shubham1g5: given this gets reset after the "Seems like you are having issues verifying" , think we need to implement another count here to log into the analytics the absolute number for failed attempt in the session

app/src/org/commcare/connect/EmailOfferHelper.kt
```diff
@@ -62,10 +64,12 @@ object EmailOfferHelper {
                 activity.getString(R.string.personalid_email_offer_message),
             )
         dialog.setPositiveButton(activity.getString(R.string.personalid_email_offer_yes)) { _, _ ->
+            FirebaseAnalyticsUtil.reportPersonalIDContinueClicked(EMAIL_OFFER_SCREEN, null)
```
shubham1g5: `reportPersonalIDContinueClicked` should only be used for `Continue` buttons on the Personal ID Sign up pages, I am afraid this will interfere with the existing funnel setup otherwise. 

I would suggest logging this into a new event instead - `email_offer_dialog` and use it for both the email offer dialog and having trouble with email dialog.

shubham1g5 (changes requested):

app/src/org/commcare/utils/OtpAnalyticsMapper.kt
```diff
@@ -47,14 +56,28 @@ object OtpAnalyticsMapper {
     enum class OtpOp {
         REQUEST,
         VERIFY,
```
conroy-ricketts: Thoughts on renaming these to `REQUEST_PHONE` and `VERIFY_PHONE`?

app/src/org/commcare/fragments/personalId/EmailHelper.kt
conroy-ricketts, line 47: I know this function was created in a different PR, but can we rename this to `requestEmailOtp()`

app/src/org/commcare/fragments/personalId/EmailHelper.kt
```diff
@@ -52,6 +55,12 @@ object EmailHelper {
         val (token, user) = buildAuthArgs(activity, workflow, sessionData)
         object : PersonalIdApiHandler<Boolean>() {
             override fun onSuccess(status: Boolean) {
+                FirebaseAnalyticsUtil.reportEmailOtpEvent(
+                    OtpAnalyticsMapper.OtpOp.REQUEST_EMAIL,
+                    AnalyticsParamValue.OTP_OUTCOME_SUCCESS,
+                    null,
+                    0,
```
conroy-ricketts: Defaulting this to 0 is not entirely accurate. Can we set this to a variable that tracks the total number of times we attempted to send/request an email OTP

app/src/org/commcare/fragments/personalId/EmailHelper.kt
```diff
@@ -79,19 +94,32 @@ object EmailHelper {
...
         onFailure: (PersonalIdOrConnectApiErrorCodes, Throwable?) -> Unit,
     ) {
         val (token, user) = buildAuthArgs(activity, workflow, sessionData)
         object : PersonalIdApiHandler<Boolean>() {
             override fun onSuccess(status: Boolean) {
+                FirebaseAnalyticsUtil.reportEmailOtpEvent(
+                    OtpAnalyticsMapper.OtpOp.VERIFY_EMAIL,
+                    AnalyticsParamValue.OTP_OUTCOME_SUCCESS,
+                    null,
+                    failedAttempts,
```
conroy-ricketts: The number of attempts passed here is not meant to count the number of failed attempts but rather the total number of times we send/request the OTP (i.e. we should also be counting a success in this analytics event)

conroy-ricketts (changes requested):
Looks great! Just had a few things to add

app/src/org/commcare/google/services/analytics/FirebaseAnalyticsUtil.java
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3745/changes/f567ad2c1061d7d23028b33fdd554f25aa3883ef)

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3745/changes/f567ad2c1061d7d23028b33fdd554f25aa3883ef)

app/src/org/commcare/utils/OtpAnalyticsMapper.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3745/changes/f567ad2c1061d7d23028b33fdd554f25aa3883ef)

app/src/org/commcare/fragments/personalId/EmailHelper.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3745/changes/f567ad2c1061d7d23028b33fdd554f25aa3883ef)

app/src/org/commcare/fragments/personalId/EmailHelper.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3745/changes/f567ad2c1061d7d23028b33fdd554f25aa3883ef)

app/src/org/commcare/connect/EmailOfferHelper.kt
Jignesh-dimagi: To stay within Firebase event limits, we reuse a single "continue button clicked" event. Depending on the user's action, it logs the following payload:

- Current Screen Name as SCREEN_NAME and null as PERSONAL_ID_CONTINUE_CLICKED_INFO when the user taps continue.
- Current Screen Name as SCREEN_NAME and 'skip' as PERSONAL_ID_CONTINUE_CLICKED_INFO when the user taps skip.
- 'EmailOfferDialog' as SCREEN_NAME and null as PERSONAL_ID_CONTINUE_CLICKED_INFO when an existing user chooses to add an email from the prompt.
- 'EmailOfferDialog' as SCREEN_NAME and 'skip' as PERSONAL_ID_CONTINUE_CLICKED_INFO when an existing user chooses to skip adding an email from the prompt.
- Current Screen Name as SCREEN_NAME and 'proceed_without_email' as PERSONAL_ID_CONTINUE_CLICKED_INFO when the user taps Proceed without email on the dialog.

Your thoughts?

app/src/org/commcare/fragments/personalId/AttemptTracker.kt
conroy-ricketts, line 1: This seems like a bit of over-engineering for a couple of counters. Thoughts on moving the counter variables to `EmailHelper` directly?

app/src/org/commcare/connect/EmailOfferHelper.kt
shubham1g5: it definitely come across as weird to me to use "Continue" button event to log "non" continue like events. Also this event is dedicated to track the Personal ID signup flow at the moment, I would highly encourage using a different event here on dialog confirmation CTAs otherwise it might be a mess to write a query tracking Personal ID funnel (for eg. when a continue button event appears in mid of a non-signup workflow. 


Curious if @OrangeAndGreen  have thoughts on this from writing a query perspective ?

Otherwise I would say that we should just add another generalised event here to track the email feature and log it with that.

app/src/org/commcare/google/services/analytics/FirebaseAnalyticsUtil.java
shubham1g5: @Jignesh-dimagi bumping this

Jignesh-dimagi: 'getEventType' already logs the non-fatal error while retrieving the event type, so it is not logged here.

shubham1g5: I see, sounds good in that case.

app/src/org/commcare/fragments/personalId/EmailHelper.kt
shubham1g5: +1 to this

shubham1g5 (dismissed):
Changing my review to non-blocking.

app/src/org/commcare/fragments/personalId/EmailHelper.kt
Jignesh-dimagi: All API handlers follow a naming convention that directs the server to take action. For example, "send" is named from the server's perspective—meaning the server sends the OTP to the user's email. Client methods simply pass the endpoint name up through every layer unchanged. Your thoughts?

shubham1g5: Good point although think that might be true for `PersonalIdApiHandler` methods but not for it's callers.

conroy-ricketts: Yeah that is a good point

To play devil's advocate, the caller of `PersonalIdApiHandler.sendPhoneOtp()` uses the terminology "requestOtp" [here](https://github.com/dimagi/commcare-android/blob/52f05db6cbe12dd13f53652848e5836af3495408/app/src/org/commcare/utils/PersonalIdAuthService.kt#L12), and that terminology is then reused going up through the `OtpManager` [here](https://github.com/dimagi/commcare-android/blob/52f05db6cbe12dd13f53652848e5836af3495408/app/src/org/commcare/utils/OtpManager.java#L35), so there is an inconsistency there

But, definitely downgrading this point to a nit, especially because the Kdoc comment would help clear any potential confusion anyways

I'm cool with wherever you take this 👍🏾

app/src/org/commcare/fragments/personalId/AttemptTracker.kt
Jignesh-dimagi: Since 'EmailHelper' is stateless, it cannot hold that data. The only remaining option was to provide the tracker from the source consuming these APIs. Now, it will work across all workflows: personal ID configuration, email offer prompt, and manage profile.

conroy-ricketts: That makes sense to me. I'm cool with keeping this as-is

conroy-ricketts (dismissed):

app/src/org/commcare/connect/EmailOfferHelper.kt
OrangeAndGreen: As far as querying goes, specifically related to the PersonalID funnel, I don't think this would cause issues. We aren't relying on this event in the funnel right now, and even if we start to we would pair it with the screen_name to know what page the user clicked Continue on, so the proposed new events could be easily filtered out since they wouldn't come from a page in the configuration workflow.

All in all I'm a bit torn. On the one side I can see Shubham's point that the usage of the event seems odd here, particularly in the "skip" case. But on the other hand I know we want to limit our usage of event_names in order to avoid the Firebase limit, so as long as we can reliably separate the data in queries then the proposed approach makes sense.

In the end, I think we're better off using a new event_name for these events. We have plenty of capacity for additional event_names, so while generally we should avoid creating new ones I think in this case it would be okay. When I think about it in terms of "Will we ever want to track these events together?" the answer seems to be no, i.e. the events for tracking Continue presses in the configuration workflow vs. other parts of the app would never be needed at the same time. So if the events are distinct that way, I think it makes sense for them to have different names.

Jignesh-dimagi: Thank you for your inputs @OrangeAndGreen @shubham1g5

I have added a new event, user_prompt. It is currently used in two places: email verification failure retries and existing user email reminders. However, because it is generic, it can also be used wherever a user's consent or decision is required in a prompt. One upcoming use case is in the profile management section, where a prompt will ask to send an OTP to an email address. 

Changes done [here](https://github.com/dimagi/commcare-android/pull/3745/changes/cf8f6ab1b25a30372083cbeda3924989d52512d3)

app/src/org/commcare/google/services/analytics/FirebaseAnalyticsUtil.java
```diff
@@ -771,6 +781,46 @@ public static void reportOtpEvent(String eventType,
...
+     *                       {@link AnalyticsParamValue#OTP_OUTCOME_FAILURE}.
+     * @param reason         failure reason from {@link OtpAnalyticsMapper}; pass null for success.
+     * @param requestCount   total number of times an email OTP was sent/requested this session. Logged to {@link FirebaseAnalytics.Param#VALUE}.
+     * @param failedAttempts absolute number of failed email OTP verifications this session. Logged
+     *                       to {@link CCAnalyticsParam#EMAIL_OTP_FAILED_ATTEMPTS}.
+     * @param workflow       the email workflow string from
+     *                       {@link OtpAnalyticsMapper#workflowParam}. Logged to
+     *                       {@link CCAnalyticsParam#EMAIL_OTP_WORKFLOW}.
+     */
+    public static void reportEmailOtpEvent(OtpAnalyticsMapper.OtpOp op,
```
OrangeAndGreen: This function along with reportPhoneOtpEvent feel like they could use some cleanup:
* Both usages rely on OtpAnalyticsMapper.getEventType to convert the OtpOp to a string, but the phone version relies on the fragment to call the helper (passing a string to analytics) while the email version passes the OtpOp to analytics and lets it call the helper. I think it makes sense for both version to pass the enum in and have analytics always be the one to call the helper function.
* Generally (and after making the change above), I see a lot of overlap between the two functions, with the only real difference being that one logs a couple extra params (failed attempts count and email workflow). I'm thinking these could share a common base function: reportOtpEvent
* Also to consider: Should the phone OTP workflow also log failed attempts? Maybe that's something we just didn't consider when implementing the phone version, and could be a nice improvement
* Similarly, I think we could change EMAIL_OTP_WORKFLOW to just OTP_WORKFLOW, and for now hard-code REGISTRATION for phone-related calls since that's the only time a user does the phone verification. If/when we support allowing users to change their phone number, this code will be ready to support the analytics for the new workflow.

app/src/org/commcare/google/services/analytics/FirebaseAnalyticsUtil.java
```diff
@@ -671,6 +672,16 @@ public static void reportPersonalIDContinueClicked(String screenName, @Nullable
         reportEvent(CCAnalyticsEvent.PERSONAL_ID_CONTINUE_CLICKED, params);
     }
 
+    public static void reportUserPromptEvent(String type, String action, @Nullable String info) {
```
OrangeAndGreen: Nit: the info can be non-nullable (no one ever passes null in)

app/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragment.kt
```diff
@@ -224,13 +233,23 @@ class PersonalIdEmailVerificationFragment : BasePersonalIdFragment() {
                 getString(R.string.personalid_email_otp_failed_message),
             )
         dialog.setPositiveButton(getString(R.string.personalid_email_otp_failed_retry)) { _, _ ->
+            FirebaseAnalyticsUtil.reportUserPromptEvent(
```
OrangeAndGreen, line 236: Nit: Looking at these calls to reportUserPrompt, the ordering of the inputs feels odd. Seems like the "info" parameter would make more sense coming before the action, since it's kind of a sub-category of the workflow.
Like:
* Email workflow, retrying after failure, user chooses to try again
* Email workflow, retrying after failure, user chooses to proceed without email

app/src/org/commcare/google/services/analytics/CCAnalyticsParam.java
```diff
@@ -54,4 +54,11 @@ public class CCAnalyticsParam {
     public static final String OTP_OUTCOME = "outcome";
     public static final String OTP_EVENT_TYPE = "event_type";
     public static final String OTP_METHOD = "method";
+    public static final String EMAIL_OTP_FAILED_ATTEMPTS = "failed_attempts";
+    public static final String EMAIL_OTP_WORKFLOW = "workflow";
+
+    // Param keys for user prompt analytics
+    static final String USER_PROMPT_TYPE = "type";
```
OrangeAndGreen: Can we make this string "event_type"? I'll start a convo with the team about this, but there's some benefit to re-using strings for similar types of fields across different events. We already have several similar parameters that map to "event_type", so making this new one do the same would be nice.

app/res/navigation/nav_graph_personalid.xml
```diff
@@ -188,6 +188,10 @@
         <argument
             android:name="workflow"
             app:argType="org.commcare.fragments.personalId.EmailWorkFlow" />
+        <argument
+            android:name="emailOtpRequestCount"
+            app:argType="integer"
+            android:defaultValue="0" />
```
OrangeAndGreen: Curious why this argument is optional? I see it means we have to set it separately in PersonalIdEmailFragment (rather than including it with the other two args), and it doesn't seem like there's ever a case where we wouldn't pass the arg in

app/src/org/commcare/fragments/personalId/EmailOtpAttemptTracker.kt
```diff
@@ -0,0 +1,32 @@
+package org.commcare.fragments.personalId
+
+/**
+ * Counters for a email-OTP collection session
+ *
+ *  - [requestCount]: total email OTP send/request calls this session.
+ *  - [failedAttempts]: absolute number of failed email OTP verifications this session.
+ */
+class EmailOtpAttemptTracker(
```
OrangeAndGreen: Nit: This could just be AttemptTracker, it doesn't need to be OTP and/or email specific.

OrangeAndGreen (changes requested):

app/src/org/commcare/google/services/analytics/FirebaseAnalyticsUtil.java
Jignesh-dimagi: Alright, I have updated the phone OTP logging to match the email OTP logging, which tracks the number of times the OTP is sent and the number of times verification fails. Additionally, I have created a common function to report these metrics.

Change done [here](https://github.com/dimagi/commcare-android/pull/3745/changes/616dc7d9f92a1a5534f47cd362126deb7e62b553)

app/src/org/commcare/google/services/analytics/FirebaseAnalyticsUtil.java
Jignesh-dimagi: Change done [here](https://github.com/dimagi/commcare-android/pull/3745/changes/616dc7d9f92a1a5534f47cd362126deb7e62b553)

app/src/org/commcare/google/services/analytics/CCAnalyticsParam.java
Jignesh-dimagi: Change done [here](https://github.com/dimagi/commcare-android/pull/3745/changes/616dc7d9f92a1a5534f47cd362126deb7e62b553)

app/res/navigation/nav_graph_personalid.xml
Jignesh-dimagi: I was initially thinking of keeping it decoupled, but I agree with you that the user should have tried at least once before coming to this screen

Change done [here](https://github.com/dimagi/commcare-android/pull/3745/changes/616dc7d9f92a1a5534f47cd362126deb7e62b553)

app/src/org/commcare/fragments/personalId/EmailOtpAttemptTracker.kt
Jignesh-dimagi: Change done [here](https://github.com/dimagi/commcare-android/pull/3745/changes/616dc7d9f92a1a5534f47cd362126deb7e62b553)

app/src/org/commcare/fragments/personalId/AttemptTracker.kt
conroy-ricketts, line 1: We can also remove any mention of "OTP" from all the code comments

app/src/org/commcare/fragments/personalId/AttemptTracker.kt
```diff
@@ -0,0 +1,32 @@
...
+    var failedAttempts: Int = initialFailedAttempts
+        private set
+
+    /** Records one OTP send/request, regardless of outcome. */
+    fun recordRequest() {
+        requestCount++
+    }
+
+    /** Records one failed OTP verification. */
+    fun recordFailedVerification() {
```
conroy-ricketts: Since this class was generalized to "AttemptTracker", we could also rename this function

```suggestion
    fun recordFailedAttempt() {
```

OrangeAndGreen (dismissed):

app/src/org/commcare/fragments/personalId/AttemptTracker.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3745/changes/5de6077cf712fa0c2a268f9eb54661f875e03cc7)

app/src/org/commcare/fragments/personalId/AttemptTracker.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3745/changes/5de6077cf712fa0c2a268f9eb54661f875e03cc7)

conroy-ricketts (approved):

PR #3746, opened 2026-06-02
CCCT-2427 Keep Unlock Prompt Open On Recoverable Biometric Failure

Jignesh-dimagi:
## Product Description

When linking a traditional (username/password) app to PersonalID, an unrecognized fingerprint no longer skips the user past the login screen. The biometric prompt stays open so the user can retry, and linking completes once the unlock succeeds. Cancelling or dismissing the prompt still proceeds to the app home, unchanged.

## Technical Summary

`onAuthenticationFailed` in the PersonalID unlock callback (`PersonalIdUnlocker.performUnlock`) was treating a recoverable biometric rejection as a terminal failure — it called `connectActivityComplete(false)`, which propagated through `unlockAndLinkConnect` → `LoginActivity.setResultAndFinish(false)` and dismissed the login screen while the system prompt was still open, also short-circuiting the link. Per the AndroidX `BiometricPrompt` contract, `onAuthenticationFailed` is non-terminal (prompt stays open for retry); only `onAuthenticationError` (cancel/dismiss/lockout) and `onAuthenticationSucceeded` end the session. The fix makes `onAuthenticationFailed` a no-op, matching the existing pattern in `PersonalIdBiometricConfigFragment`.

### Safety story

What gives confidence:
- I manually reproduced the bug on a device and verified the fix: scanning the wrong finger now keeps the prompt open with no background transition to app home, and a subsequent correct scan links the app.

Risks to review:
- No automated test covers this behavior — the callback is an inline anonymous object and not unit-testable without extracting it. Verification was manual.
- Behavior change for existing users on the login→link flow; reviewers should confirm the cancel/dismiss path still transitions to app home as expected.

[CCCT-2427]: https://dimagi.atlassian.net/browse/CCCT-2427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/org/commcare/personalId/PersonalIdUnlocker.kt
+++ b/app/src/org/commcare/personalId/PersonalIdUnlocker.kt
@@ -85,7 +85,9 @@ object PersonalIdUnlocker {
                     callback.connectActivityComplete(true)
                 }
 
-                override fun onAuthenticationFailed() = callback.connectActivityComplete(false)
+                // A recoverable failure (e.g. an unrecognized fingerprint). The system
+                // prompt stays open for the user to retry.
+                override fun onAuthenticationFailed() = Unit
             }
 
         when {
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -20,6 +20,7 @@ These are published publicly on Playstore, Github Releases and CommCare Forums
 #### Important Bug Fixes
 
 - Fixed the back arrow on the camera capture screen so it correctly returns to the previous screen
+- Fixed PersonalID app linking so a failed fingerprint scan no longer skips you past the login screen; you can retry the unlock and linking completes once it succeeds
 
 #### Internal Release Notes
 


Jignesh-dimagi:
## Suggested Review Order

- `app/src/org/commcare/personalId/PersonalIdUnlocker.kt` — the one-line behavior fix; read the three `AuthenticationCallback` overrides together to see why only `onAuthenticationFailed` changed
- `RELEASES.md` — user-facing release note for the fix

conroy-ricketts (approved):

app/src/org/commcare/personalId/PersonalIdUnlocker.kt
```diff
@@ -85,7 +85,9 @@ object PersonalIdUnlocker {
                     callback.connectActivityComplete(true)
                 }
 
-                override fun onAuthenticationFailed() = callback.connectActivityComplete(false)
```
shubham1g5, line 88: seems like with this change, we no longer need to pass a boolean flag to `connectActivityComplete` and can always assume it to be `true` ?

Also this change seems to contradict the change in https://github.com/dimagi/commcare-android/pull/3748/files given the `else -> finish` condition will never be true in that PR. 

Also think it would be critical for us to test all the impacted places with this change (anywhere we are handling for `false` result)

Jignesh-dimagi: 'BiometricPrompt.AuthenticationCallback' has 2 error callbacks:

onAuthenticationFailed: 
- Fired when a biometric didn't match an enrolled one — e.g., the wrong finger or someone else's face.
- The prompt stays open, and the user can immediately retry, so it can fire multiple times in one session.

onAuthenticationError:
- Fired when the authentication session ends without success, and the prompt is dismissed
- The user pressed the back button.
- If no biometric hardware is present
- Too many 'onAuthenticationFailed', i.e., tried too many times with the wrong finger or face 

I have removed the 'connectActivityComplete' callback from 'onAuthenticationFailed' only in order to allow the user to scan biometrics again in case of wrong finger or face. This change also prevents the application from proceeding while the prompt remains open.

The application still has 'connectActivityComplete' callbacks from 'onAuthenticationError', which will be fired whenever the errors, mentioned above, occur.

shubham1g5: Ahh that makes a lot of sense, I didn't realise there is another `onAuthenticationError` up there.

shubham1g5 (approved):

PR #3749, opened 2026-06-05
[Hotfix] SAAS-19802 Fix Combobox.autoCorrectCapitalization choice match check

avazirna:
## Product Description
Fixes a bug causing string/numeric questions to lose focus while users are entering data. For each character/digit typed, the form controller performs a pass through the form's triggerables. During this process, existing questions are re-evaluated. Because the `ComboboxWidget` current value within `autoCorrectCapitalization` was always `false`, a new value was being set, causing the input focus to be reassigned:

https://github.com/user-attachments/assets/d3f7c374-152b-474e-b935-4ba3ac7dfe29

Ticket: https://dimagi.atlassian.net/browse/SAAS-19802

Related PRs:
- https://github.com/dimagi/commcare-android/pull/3515 
- https://github.com/dimagi/commcare-core/pull/1516

## Technical Summary
`autoCorrectCapitalization` had a latent bug in its guard:

```java
if (enteredText != null && !choices.contains(enteredText) && ...)
```

`choices` is `Vector<ComboItem>` and `enteredText` is `String`. `ComboItem.equals(Object)` (commcare-core) only returns `true` for another `ComboItem`, so `choices.contains(enteredText)` was **always false**. The intended "skip the case-fix when the user already typed the canonical form" branch never fired, so a `setText(...)` ran every
time the field lost focus on a valid value — silent extra work plus an unnecessary TextWatcher pass.

### Safety story
- This was successfully tested locally.
- Added new tests to assess the correct behavior of the `autoCorrectCapitalization()` method

### Automated test coverage
`app/unit-tests/src/org/commcare/views/ComboboxTest.kt` — four tests,
all passing locally 

## Labels and Review

### DIFF
--- a/app/unit-tests/src/org/commcare/views/ComboboxTest.kt
+++ b/app/unit-tests/src/org/commcare/views/ComboboxTest.kt
@@ -1,5 +1,6 @@
 package org.commcare.views
 
+import android.content.Context
 import androidx.test.core.app.ApplicationProvider
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import org.commcare.CommCareTestApplication
@@ -59,7 +60,7 @@ class ComboboxTest {
     }
 
     private fun newCombobox(vararg displayTexts: String): Combobox {
-        val context = ApplicationProvider.getApplicationContext<android.content.Context>()
+        val context = ApplicationProvider.getApplicationContext<Context>()
         val items = displayTexts.mapIndexed { i, label -> ComboItem(label, label, i) }
         val choicesVector = Vector(items)
         val adapter = ComboboxAdapter(context, items.toTypedArray(), permissiveFilterRule())
--- a/app/unit-tests/src/org/commcare/views/ComboboxTest.kt
+++ b/app/unit-tests/src/org/commcare/views/ComboboxTest.kt
@@ -1,0 +1,87 @@
+package org.commcare.views
+
+import androidx.test.core.app.ApplicationProvider
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import org.commcare.CommCareTestApplication
+import org.commcare.adapters.ComboboxAdapter
+import org.javarosa.core.model.ComboItem
+import org.javarosa.core.model.ComboboxFilterRule
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.annotation.Config
+import java.util.Vector
+
+@Config(application = CommCareTestApplication::class)
+@RunWith(AndroidJUnit4::class)
+class ComboboxTest {
+    @Test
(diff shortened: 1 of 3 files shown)


avazirna:
@shubham1g5 can you re-approve this. It's a copy of https://github.com/dimagi/commcare-android/pull/3736 plus QA notes

app/unit-tests/src/org/commcare/views/ComboboxTest.kt
conroy-ricketts, line 1: Can we rename these tests to be natural language sentences with backticks as per the guidance [here](https://github.com/dimagi/open-source/blob/master/docs/mobile_standards.md#testing-criteria)

app/unit-tests/src/org/commcare/views/ComboboxTest.kt
```diff
@@ -0,0 +1,87 @@
...
+        val combobox = newCombobox("Apple", "Banana")
+        combobox.setText("")
+
+        combobox.autoCorrectCapitalization()
+
+        assertEquals("", combobox.text.toString())
+    }
+
+    private fun newCombobox(vararg displayTexts: String): Combobox {
+        val context = ApplicationProvider.getApplicationContext<android.content.Context>()
```
conroy-ricketts: Nit: should import this

```suggestion
        val context = ApplicationProvider.getApplicationContext<Context>()
```

app/unit-tests/src/org/commcare/views/ComboboxTest.kt
```diff
@@ -0,0 +1,87 @@
...
+        object : ComboboxFilterRule {
+            override fun choiceShouldBeShown(
+                choice: ComboItem,
+                textEntered: CharSequence,
+            ): Boolean = true
+
+            override fun shouldRestrictTyping(): Boolean = false
+        }
+
+    private fun assertSameEditable(
```
conroy-ricketts, line 80: Why create a helper vs using `assertEquals()`?

avazirna: This method is checking for referential inequality with `!==`, not comparing content, to check if a new object was created.

OrangeAndGreen (approved):

app/unit-tests/src/org/commcare/views/ComboboxTest.kt
conroy-ricketts: Aaah I see, sounds good!

app/unit-tests/src/org/commcare/views/ComboboxTest.kt
avazirna: 7317e6607d4cefe750fc66b71e2c233c26e42240

conroy-ricketts (approved):

PR #3750, opened 2026-06-05
[Hotfix] MOB-171 Fix GPS Here() race condition crash on back Navigation

avazirna:
## Product Description

Fixed a fatal crash affecting app configurations that use the `here()` XPath function in case lists. When a user pressed Back from such a case list, a GPS Binder IPC race condition could cause the entity loader to run with a `null` session command, resulting in an empty `EvaluationContext` that could not resolve the `casedb` instance.

## Technical Summary

**Root Cause (GPS Binder IPC Race Condition):**

1. `EntitySelectActivity` evaluates `here()` during entity loading → GPS location updates start via `LocationManager.requestLocationUpdates(..., mainLooper)`
2. Entity load completes → `deliverLoadResult()` sets `loader = null` (enabling a future re-trigger)
3. User presses Back → `onPause()` → `forbidGpsUse()` → `removeLocationUpdates()` — GPS deregistered, but a Binder callback already dispatched to the main Looper queue cannot be recalled
4. Parent activity's `onActivityResult()` fires (this happens between `onPause` and `onStop` per Android lifecycle) → `processCanceledGetCommandOrCase()` → `stepBack()` → `session.currentCmd = null`
5. In-flight GPS Binder callback fires on main thread → `onLocationChanged()` → `listener != null` (not yet unregistered — that was in `onStop`) → `onEvalLocationChanged()` → `loadEntities()`
6. `loader == null` (from step 2) → creates new `EntityLoaderTask` with `evalContext()`
7. `evalContext()` → `session.getEvaluationContext(iif, null, null)` → returns `new EvaluationContext(null)` with no form instances
8. Background task calls `expandReferenceList` → `retrieveInstance("casedb")` → **RuntimeException** → rethrown fatally

**Why only `here()` apps are affected:** Without `here()`, GPS is never started, so no Binder callbacks are ever dispatched.

**Fix:** Move `hereFunctionHandler.unregisterListener()` from `onStop()` to `onPause()`, immediately after `forbidGpsUse()`. This nulls the listener before `onActivityResult()` can modify session state, so the in-flight GPS callback sees `listener == null` and never calls `onEvalLocationChanged()`. But in case it gets called, there is now a safeguard that prevents the `onEvalLocationChanged()` to be called when activity is finishing or session is `null`.

Symmetrically, `hereFunctionHandler.registerListener(this)` is moved from `onStart()` to `onResumeSessionSafe()` (the valid-session branch, before `allowGpsUse()`), keeping registration and deregistration in the same `onResume/onPause` lifecycle pair.

A `CrashUtil.log` diagnostic is also added in `loadEntities()` to surface the null-command condition non-fatally in Crashlytics, confirming the hypothesis if it recurs.

### Safety story
- The crash was reproducible after commenting out the code that stops location updates, disabling mobile data to force the provider to rely on GPS hardware, and moving around to generate additional location updates that could trigger the issue. 
  - With the changes in this PR, the crash was no longer observed under the same conditions, suggesting the fix is effective.
  - The case list refresh on location updates was also verified and continues to work as expected.
- Another aspect worth noting is that the Android lifecycle guarantee (`onActivityResult` fires between `onPause` and `onStop`) directly validates the fix
- The `here()` GPS flow is only active for app configurations that explicitly use the `here()` XPath function in case lists, limiting the blast radius

### Automated test coverage

No automated test coverage for this specific lifecycle/GPS interaction. The race condition is timing-dependent and involves Android Binder IPC, making it difficult to unit test. The `CrashUtil.log` diagnostic will surface non-fatally in Crashlytics if the condition is ever triggered again.

[MOB-171]: https://dimagi.atlassian.net/browse/MOB-171?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -2,6 +2,37 @@
 This file is meant as an easy way for us to collate notes and change logs across releases. 
 -->
 
+## CommCare 2.63.1
+
+### Release Notes
+
+<!--
+These are published publicly on Playstore, Github Releases and CommCare Forums
+-->
+
+#### What's New
+
+
+#### Important Bug Fixes
+
+- Fixed a crash that could occur when navigating back from a case list that uses the GPS `here()` function
+
+#### Internal Release Notes
+
+<!--
+Release notes that are not applicable for wider CommCare users but only for specific projects.
+These notes are only published internally in [CommCare Change log wiki](https://dimagi.atlassian.net/wiki/spaces/internal/pages/2145058874/CommCare+Mobile+Changelog)
+along with the public release notes above
+-->
+
+### QA Notes
+
+<!--
+These are for internal use and for us to keep track of important notes that
+we would like to communicate to QA as part of the release testing
+-->
+
+
 ## CommCare 2.63
 
 ### Release Notes
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -32,6 +32,8 @@ These are for internal use and for us to keep track of important notes that
 we would like to communicate to QA as part of the release testing
 -->
 
+- **GPS `here()` crash regression:** Using an app configuration that has `here()` in a case list (e.g. a case list that shows distance to GPS coordinates), navigate to that case list and wait for it to fully load. Press Back. Verify the app returns to the previous screen without crashing.
+- **GPS `here()` case list refresh:** Using the same app configuration, navigate to the case list and allow GPS to acquire a location fix. Move to a different location (or simulate a location change) while the list is displayed. Verify the case list automatically refreshes and distance values update accordingly.
 
 ## CommCare 2.63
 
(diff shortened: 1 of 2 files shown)


avazirna:
@shubham1g5 this ready for review. I've added a safeguard to prevent the 596038ceceef133f78a8d1828086f12df481e04a `loadEntities()` to be called when the session is in a stale state.

conroy-ricketts (commented):
> I did not personally test these changes — verification is needed by another developer or QA against an app config that uses here() in a case list

Even though the diff is narrow, this is a big risk. Can we manually test this before handing it off to QA

OrangeAndGreen (approved):

avazirna:
> > I did not personally test these changes — verification is needed by another developer or QA against an app config that uses here() in a case list
> 
> Even though the diff is narrow, this is a big risk. Can we manually test this before handing it off to QA

My bad, I updated the ticket description but may not have saved it properly. It’s updated now.

conroy-ricketts (approved):

PR #3753, opened 2026-06-05
CCCT-2439 Launch Connect Apps Without Showing The Login Screen

conroy-ricketts:
## Product Description

When a worker launches an installed app from a Connect opportunity, it now opens directly behind a single progress dialog instead of briefly flashing the `LoginActivity` and app-seating screens. Backing out of the app's Home screen returns to the Connect opportunities list, as before.

Here is the _**new**_ login flow (I intentionally slowed it down for this demo so every phase is shown):

https://github.com/user-attachments/assets/6edfa635-766a-4cdb-bdb7-4cd94ab3922c

Here is the _**old**_ login flow (I intentionally slowed it down for this demo so every phase is shown):

https://github.com/user-attachments/assets/778c24d6-cdc1-4499-ac8e-b720fff68a10

## Technical Summary

Phase 3 of the headless-login work (builds on CCCT-2437/CCCT-2438). `ConnectAppLauncher` runs the seat + PersonalID login headlessly off the opportunities list and reports a `LaunchOutcome`; `LaunchOutcomeRouter`/`LaunchActions` is a thin seam (unit-tested without a fragment) that maps each outcome to a UI/nav action. `ConnectJobsListsFragment` drives it on `viewLifecycleOwner.lifecycleScope` behind a modal `CustomProgressDialog`.

A few decisions worth knowing before the diff:

- **Failures reuse existing surfaces**: `TokenDenied` → `TokenExceptionHandler`; a seat failure → `DispatchActivity` (which reaches recovery for a corrupted app); credential/other failures → fall back to the legacy `ConnectAppUtils.launchApp`, which shows `LoginActivity` with its usual error handling (no new error UI/strings).
- **Reverted phase-2 over-engineering**: this branch also inlines `DispatchActivity`'s post-login intent handling back to its master form and deletes the now-unused `PostLoginRouter`/`PostLoginDestination`/`LaunchContext` (+ test). They were added in CCCT-2438 on the expectation phase 3 would consume them, but phase 3 doesn't — `route()` only ever produced `Home` there, so the round-trip was behavior-identical to the inline version.
- **Trimmed the seating result API**: `SeatResult` is now just `Success`/`Failed`, dropping the plan's per-reason `SeatFailure` enum and the `SeatResultCallback` seam. No caller acts on the reason — a seat failure is already surfaced through the app's `STATE_CORRUPTED` state that `DispatchActivity` detects — so the extra detail was unused.

## Open Question

Open question for the team (not addressed here): the progress dialog shows the same text in both its title and body during the seating and login phases (pre-existing for the login phase; introduced for seating to match). Is this worth tweaking?

### Safety story

What gives me confidence:
- I ran the change on a device and launched apps from the Connect opportunities list end-to-end, confirming the single-dialog open, the seating/login dialog text, and back-from-Home returning to the opportunities list.
- The launcher and outcome routing are covered by JVM unit tests.
- Diff is scoped to the Connect launch path; failures route into pre-existing handlers rather than new code.

Risks to review:
- I did **not** manually exercise the failure paths (token denied, no network, seat failure, missing Connect user). They're unit-tested at the launcher/router level and delegate to existing handlers, but haven't been run on a device.
- Home is launched directly (not via `DispatchActivity`), so the session stays active with the Connect screen on the back stack and the `HOME_SCREEN`-for-result logout/back contract is bypassed for this entry point — worth confirming logout/back behavior is acceptable.
- No Robolectric fragment test for `ConnectJobsListsFragment` (no fragment-Robolectric harness exists in the suite); the routing decision is covered via the `LaunchActions` seam instead.

### Automated test coverage

- `ConnectAppLauncherTest` — phase/credential resolution, seat-failure short-circuit, and `LoginResult` → `LaunchOutcome` mapping for each failure variant.
- `LaunchOutcomeRouterTest` — every outcome dismisses progress first and dispatches the expected `LaunchActions` call.

[CCCT-2439]: https://dimagi.atlassian.net/browse/CCCT-2439?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


conroy-ricketts:
## Suggested Review Order

- `app/src/org/commcare/connect/ConnectAppLauncher.kt` — new core: the headless seat+login flow and the `LaunchOutcome` result type
- `app/src/org/commcare/connect/LaunchOutcomeRouter.kt` — the `LaunchActions` seam mapping each outcome to a nav/UI action
- `app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java` — the caller: progress dialog, lifecycle scope, and the action implementations
- `app/src/org/commcare/activities/HomeScreenBaseActivity.java` — `launchHome`/`buildHomeIntent` used on success
- `app/src/org/commcare/activities/DispatchActivity.java` — Connect home routing plus the inlined post-login handling (phase-2 revert)
- `app/src/org/commcare/login/PostLoginRouter.kt`, `PostLoginDestination.kt`, `app/unit-tests/.../PostLoginRouterTest.kt` (deleted) — the reverted, now-unused routing abstraction
- `app/unit-tests/src/org/commcare/connect/ConnectAppLauncherTest.kt` — outcome/credential/seat-failure coverage
- `app/unit-tests/src/org/commcare/connect/LaunchOutcomeRouterTest.kt` — outcome→action dispatch coverage
- `docs/commcare/login-engine.md` — high-level doc for phases 1–3

conroy-ricketts:
@coderabbitai full review

app/src/org/commcare/connect/ConnectAppLauncher.kt
```diff
@@ -0,0 +1,127 @@
...
+            appId,
+        )
+
+        if (seatApp(appId, sink) is SeatResult.Failed) {
+            return LaunchOutcome.AppSeatFailed
+        }
+
+        val username =
+            connectUsername(context)?.trim()?.lowercase()
+        if (username.isNullOrEmpty()) {
```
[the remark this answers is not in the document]
conroy-ricketts: [Claude Opus 4.8] Fixed in 10082ee — `lowercase()` now uses `Locale.ROOT` to avoid locale-sensitive case folding when normalizing the Connect username.

app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java
```diff
@@ -180,6 +203,134 @@ private void launchAppForJob(ConnectJobRecord job, boolean isLearning) {
...
+        } else {
+            launchDialog = CustomProgressDialog.newInstance(
+                    Localization.get("seating.app"),
+                    Localization.get("seating.app"),
+                    LAUNCH_DIALOG_TASK_ID
+            );
+        }
+        showingSyncDialog = syncing;
+        launchDialog.showNow(getChildFragmentManager(), LAUNCH_DIALOG_TAG);
+    }
```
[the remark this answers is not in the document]
conroy-ricketts: [Claude Opus 4.8] Fixed in 10082ee — `showLaunchDialog()` now returns early when `getChildFragmentManager().isStateSaved()`, before the dismiss, so a launch callback delivered between `onSaveInstanceState` and `onDestroyView` can no longer trigger an `IllegalStateException`.

app/src/org/commcare/activities/DispatchActivity.java
```diff
@@ -519,7 +508,12 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
                 if (resultCode == RESULT_CANCELED) {
                     shouldFinish = true;
                 } else if (intent != null) {
-                    applyPostLoginDestination(routeLoginResult(intent));
+                    lastLoginMode = (LoginMode)intent.getSerializableExtra(LoginActivity.LOGIN_MODE);
+                    userManuallyEnteredPasswordMode =
+                            intent.getBooleanExtra(LoginActivity.MANUAL_SWITCH_TO_PW_MODE, false);
+                    personalIdManagedLogin = intent.getBooleanExtra(PERSONALID_MANAGED_LOGIN, false);
+                    connectManagedLogin = intent.getBooleanExtra(CONNECT_MANAGED_LOGIN, false);
+                    startFromLogin = true;
```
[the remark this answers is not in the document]
conroy-ricketts: [Claude Opus 4.8] Skipped. The three exits this flags (`shouldFinish()`, `recoveryMeasuresPending()`, `handleConnectButtonPress()`) are pre-existing and unchanged by this PR. DispatchActivity reads these extras only inside `if (intent != null)` (DispatchActivity.java:510), so a null/partial result Intent does not produce an NPE — the fields simply retain their defaults. No PR-introduced regression here, so leaving as-is to keep the change scoped.

app/unit-tests/src/org/commcare/connect/ConnectAppLauncherTest.kt
```diff
@@ -0,0 +1,162 @@
...
+class ConnectAppLauncherTest {
+    private val context = mockk<Context>(relaxed = true)
+    private val app = mockk<CommCareApplication>(relaxed = true)
+    private val sink = LoginProgressSink { }
+
+    private var seatResult: SeatResult = SeatResult.Success
+    private var loginAnswer: suspend () -> LoginResult = { success() }
+    private var username: String? = "Alice "
+    private val seatedApps = mutableListOf<String>()
+    private val capturedRequests = mutableListOf<LoginRequest>()
```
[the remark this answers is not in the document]
conroy-ricketts: [Claude Opus 4.8] Skipped. JUnit4 instantiates a fresh test-class instance per `@Test` method, so `seatedApps`, `capturedRequests`, `seatResult`, `loginAnswer`, `username`, and `launcher` are already re-initialized for every test. There is no cross-test pollution to guard against. (Did remove the unused `reportCccAppFailedAutoLogin` mock stub in 10082ee.)

app/src/org/commcare/activities/SeatAppActivity.java
```diff
@@ -35,7 +34,7 @@ protected void onCreate(Bundle savedInstanceState) {
         );
     }
 
-    private void finishWithResult(SeatResult result) {
+    private void finishWithResult() {
```
OrangeAndGreen: Nit: This function no longer takes a result so the name is misleading. Maybe rename it to setResultAndFinish instead?

OrangeAndGreen (dismissed):

app/src/org/commcare/activities/SeatAppActivity.java
conroy-ricketts: Good eye!

conroy-ricketts: https://github.com/dimagi/commcare-android/commit/8153e70a706aaacbac7daf2c806bc4ba39023290

OrangeAndGreen (approved):

app/src/org/commcare/connect/ConnectAppLauncher.kt
```diff
@@ -0,0 +1,132 @@
...
+        )
+
+        if (seatApp(appId, listener) is SeatResult.Failed) {
+            return LaunchOutcome.AppSeatFailed
+        }
+
+        val username =
+            connectUsername(context)?.trim()?.lowercase(Locale.ROOT)
+        if (username.isNullOrEmpty()) {
+            return LaunchOutcome.CredentialResolutionFailed
```
shubham1g5, line 93: I would think that we should crash from here instead of replying with an error code.

app/src/org/commcare/connect/ConnectAppLauncher.kt
```diff
@@ -0,0 +1,132 @@
...
+            )
+
+        return when (val result = performLogin(context, request, listener)) {
+            is LoginResult.Success -> {
+                LaunchOutcome.Launched
+            }
+
+            is LoginResult.Failed -> {
+                when (result.error) {
+                    is LoginError.TokenDenied -> LaunchOutcome.TokenDenied
```
shubham1g5, line 116: we earlier agree'd on Token denied error handled through with the current pathway in `CommCareExceptionHandler` and curious if there is a reason that we decided to divert here

app/src/org/commcare/connect/ConnectAppLauncher.kt
```diff
@@ -0,0 +1,132 @@
...
+                    else -> LaunchOutcome.Retryable(result.error)
+                }
+            }
+        }
+    }
+
+    private fun hasMultipleMatchingUsers(username: String): Boolean {
```
shubham1g5, line 123: can we log a non-fatal if this is ever true, I don't see how can there be multiple users with same username for connect.

app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java
```diff
@@ -180,6 +203,137 @@ private void launchAppForJob(ConnectJobRecord job, boolean isLearning) {
...
+                LogTypes.TYPE_ERROR_WORKFLOW,
+                "Connect launch failed for app " + appId + ": " + reason
+        );
+        FirebaseAnalyticsUtil.reportCccAppFailedAutoLogin(appId);
+    }
+
+    private void showLaunchDialog(boolean syncing) {
+        if (getChildFragmentManager().isStateSaved()) {
+            return;
+        }
```
shubham1g5, line 302: what is this meant to achieve ? Think would be nice to leave a in-code comment here as it's not obvious to me what is this check for

app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java
```diff
@@ -180,6 +203,137 @@ private void launchAppForJob(ConnectJobRecord job, boolean isLearning) {
...
+                getViewLifecycleOwner(),
+                activity,
+                appId,
+                isLearning,
+                progress -> activity.runOnUiThread(() -> updateLaunchProgress(progress)),
+                outcome -> handleLaunchOutcome(outcome, activity, isLearning, appId)
+        );
+    }
+
+    private void updateLaunchProgress(LoginProgress progress) {
```
shubham1g5, line 219: think we should put all dialod handling to a `base` fragment/activity level so that we can utilise it for all connect screens we want to launch the CC app from.

app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java
```diff
@@ -180,6 +203,137 @@ private void launchAppForJob(ConnectJobRecord job, boolean isLearning) {
...
+            );
+            launchDialog.addProgressBar();
+        } else {
+            launchDialog = CustomProgressDialog.newInstance(
+                    Localization.get("seating.app"),
+                    Localization.get("seating.app"),
+                    LAUNCH_DIALOG_TASK_ID
+            );
+        }
+        showingSyncDialog = syncing;
```
shubham1g5, line 318: we should define new strings in `strings.xml` with language translations,  instead of reusing the existing ones preserved for app level translations (to be consistent that all strings on connect screens come from strings.xml)

shubham1g5 (commented):
> Open question for the team (not addressed here): the progress dialog shows the same text in both its title and body during the seating and login phases (pre-existing for the login phase; introduced for seating to match). Is this worth tweaking?

Agree this seems worth changing so that we show a more conscise title with a details phase specific message, for eg. the title can remain "Launching App"  for all phases, and message can provide further phase specific info based on the internal phase.

app/src/org/commcare/connect/ConnectAppLauncher.kt
conroy-ricketts: I'm struggling to remember what the reason was for diverting here. I'll change this to route through `CommCareExceptionHandler` for now

app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java
conroy-ricketts: Sounds good, I'll add a comment here.

The purpose of this is to avoid an `IllegalStateException` given that we are showing the dialog asynchronously outside of the activity lifecycle.

Calling `showNow()` on the dialog after the host fragment has saved its instance state will throw an `IllegalStateException`

app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java
conroy-ricketts: Skipping this because it changed quite a bit in the next PR

app/src/org/commcare/connect/ConnectAppLauncher.kt
conroy-ricketts: https://github.com/dimagi/commcare-android/commit/c89d8f5d9e3e49f3c35303b5d9a5a000b7177b4a

app/src/org/commcare/connect/ConnectAppLauncher.kt
conroy-ricketts: https://github.com/dimagi/commcare-android/commit/62775897e1247d63409027fa4795253fd5ccb539

app/src/org/commcare/connect/ConnectAppLauncher.kt
conroy-ricketts: https://github.com/dimagi/commcare-android/commit/a019136ec55b436e8de41756b246692dfde3117e

app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java
conroy-ricketts: https://github.com/dimagi/commcare-android/commit/d4b444090e61bc8bffe3b98b499a27a14bfb1b31

app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java
conroy-ricketts: https://github.com/dimagi/commcare-android/commit/6780b799010072122fb69201f7e4088d41e403b0

PR #3755, opened 2026-06-09
Legacy Apache fix for Maps usage

OrangeAndGreen:
https://dimagi.atlassian.net/browse/CI-709

## Product Description
No visible change to user, but fixes a crash for some users running SDK 28 (Android 9) or newer when entering the entity map page.

## Technical Summary
Specifying legacy Apache library usage in the manifest.
This addresses a crash that can happen when accessing a Google Maps interface inside CommCare.
The issue is that devices running older than SDK 28 automatically include the legacy Apache code, so Maps (which may use that code) works. But on newer devices, the legacy code is only included if the manifest explicitly says to do so, which is what our change does.

Stack trace: ([example](https://console.firebase.google.com/u/0/project/commcare-a57e4/crashlytics/app/android:org.commcare.dalvik/issues/ec5d385398404bc36c8e5978d93a268d?time=7d&types=crash&sessionEventKey=6A27E8DA013A0001430ED806322C4548_2227791755130443448&sessionTab=data&userInfoQuery=b376c4af07ec7cef43ab))
Fatal Exception: java.lang.NoClassDefFoundError: Failed resolution of: Lorg/apache/http/ProtocolVersion;
       at m6.cv.a(:com.google.android.gms.dynamite_mapsdynamite@252431035@25.24.31 (260400-0):217)
       at m6.cc.a(:com.google.android.gms.dynamite_mapsdynamite@252431035@25.24.31 (260400-0):202)
       at m6.cc.run(:com.google.android.gms.dynamite_mapsdynamite@252431035@25.24.31 (260400-0):6)

### Safety story
Verified the maps interface still works for me (I wasn't crashing before).
This is a pretty safe change for two reasons:
* It only includes code, nothing has to call that code
* The required flag is set to false, so nothing breaks if the code can't be imported (but it should always work anyway)

### Automated test coverage
No changes

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -411,6 +411,10 @@
             android:name="com.google.android.maps"
             android:required="false" />
 
+        <uses-library
+            android:name="org.apache.http.legacy"
+            android:required="false" />
+
         <activity android:name="org.commcare.activities.MessageActivity" />
 
         <receiver


conroy-ricketts (approved):
LGTM

Should this PR target the release branch for the hotfix?

avazirna:
@OrangeAndGreen can you confirm if this is only relevant to Connect or CommCare in general? If only to Connect, do we know which Google Maps part/module is trying to load this class?

OrangeAndGreen:
Technically this feature is open to all CommCare apps, but I think so far only Connect is using it. It's configurable on an app basis, and when configured provides an option when selecting a case to show cases on a map. [EntityMapActivity](https://github.com/dimagi/commcare-android/blob/574d90c6239d216b62858506b8b6bc2c6f0ba61b/app/src/org/commcare/gis/EntityMapActivity.java#L188) specifically is the one that loads the map.

PR #3756, opened 2026-06-09
CCCT-2440 Connect App Launch For Remaining Pages And Back Navigation

conroy-ricketts:
### [CCCT-2440](https://dimagi.atlassian.net/browse/CCCT-2440)

> Stacked on top of `CCCT-2439-connect-login-silent-launch-path` — review/merge that first.

## Product Description

Extends the no-login-screen Connect app launch to the remaining priority entry points and makes back navigation between the app and Connect consistent:

- Launching a CommCare app from the **Delivery Progress** and **Learning Progress** pages now opens it directly (single loading dialog) instead of flashing the login and app-setup screens.
- From a Connect-launched app's home screen, **back** returns to the Connect screen the app was launched from; the app session stays alive (so relaunching the same app skips re-login).
- **Logging out** from the app home falls back to the Connect screen the app was launched from, the same as backing out.
- **View Job Status** on the app home opens the job's progress page on top of the live home; backing out of it returns to the app home, then back again returns to the Connect screen the app was launched from.
- A job progress page opened directly from the opportunities list still backs out to the opportunities list.

https://github.com/user-attachments/assets/9a8212d4-0b22-453b-a99a-484a40a838f2

## Technical Summary

- Extracts the launch/dialog/outcome-routing wiring that lived in `ConnectJobsListsFragment` into a shared `ConnectAppLaunchController`, and routes all three Connect launch surfaces (jobs list, delivery progress, learning progress) through it.
- Home is launched with `FLAG_ACTIVITY_CLEAR_TOP` (not for result): both a back-out and an in-app logout let the back stack return the worker to the launching screen. The app session is kept alive on back-out, so relaunching an already-signed-in app skips seating and login. A retryable launch failure shows a Retry/Cancel dialog instead of falling back to `LoginActivity`.
- `userPressedOpportunityStatus` ("View Job Status") now launches the job-status page on top of the still-live home instead of finishing it, preserving back-to-home.
- The progress dialog is driven by a small pure `LaunchProgressMapper` → `LaunchDialogState`; dialog cleanup is handled by a view-lifecycle observer so callers don't manage it.
- Removes dead code surfaced in review (the unused static `HomeScreenBaseActivity.launchHome` and `ConnectAppUtils.getPasswordOverride`/`shouldOverridePassword`).

### Safety story

What gives confidence:
- I ran the app and exercised the launch and back-navigation flows on a device.
- The diff is scoped to the Connect launch surfaces; the shared controller centralizes logic that was previously duplicated.
- `LaunchProgressMapper`, `ConnectAppLauncher`, and `LaunchOutcomeRouter` have JVM unit-test coverage.

Risks to review:
- The multi-activity back-stack behavior (home ↔ job-status page ↔ opportunities list) is not exercised by the JVM test suite — it relies on the on-device verification above.
- `userPressedOpportunityStatus` is shared with the legacy `LoginActivity` launch path; it now keeps home alive and launches the job page on top for that path too.
- Removed `ConnectAppUtils.getPasswordOverride`/`shouldOverridePassword` after confirming no remaining source callers.

### Automated test coverage

- `LaunchProgressMapperTest` covers the per-phase dialog state mapping (seating / signing-in / syncing, percent handling, message override).
- Existing `ConnectAppLauncherTest` and `LaunchOutcomeRouterTest` cover the launcher sequencing and outcome routing the controller delegates to.

[CCCT-2440]: https://dimagi.atlassian.net/browse/CCCT-2440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


conroy-ricketts:
## Suggested Review Order

- `app/src/org/commcare/connect/ConnectAppLaunchController.kt` — new shared launch orchestrator; the heart of the change
- `app/src/org/commcare/connect/ConnectNavHelper.kt` — adds the `clearTop` option used by the back-out path
- `app/src/org/commcare/activities/HomeScreenBaseActivity.java` — "View Job Status" + the Home launch intent; dead `launchHome` removed
- `app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java` — refactored onto the shared controller (boilerplate removed)
- `app/src/org/commcare/fragments/connect/ConnectDeliveryProgressFragment.java` — new caller of the controller
- `app/src/org/commcare/fragments/connect/ConnectLearningProgressFragment.java` — new caller of the controller
- `app/src/org/commcare/connect/ConnectAppUtils.kt` — dead-code removal
- `app/unit-tests/src/org/commcare/connect/LaunchProgressMapperTest.kt` — dialog-state mapping tests
- `docs/commcare/login-engine.md` — doc update for the launch/back-nav behavior
- `RELEASES.md` — QA notes

app/src/org/commcare/activities/HomeScreenBaseActivity.java
```diff
@@ -586,10 +587,12 @@ protected void userTriggeredLogout() {
...
-        i.putExtra(REDIRECT_TO_CONNECT_OPPORTUNITY_INFO, true);
-        setResult(RESULT_OK, i);
-        finish();
+        // Launch the seated app's job status page on top of this (still-live) Home so the app
+        // session is preserved and backing out of the status page returns here.
+        ConnectJobRecord job = ConnectJobHelper.INSTANCE.getJobForSeatedApp(this);
+        if (job != null) {
+            ConnectNavHelper.INSTANCE.goToActiveInfoForJob(this, job, true);
+        }
     }
```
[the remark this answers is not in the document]
conroy-ricketts: 2b239938797e2e738bddfeaa91c465a14dc1f582

OrangeAndGreen (changes requested):
I noticed a couple behavioral changes with the new approach that I think we'll need to address.

1. Back nav after launching app from learning/delivery progress pages
Old behavior: navigating back from App Home (after successful login) returned to opportunity list
New behavior: Back nav now returns to learning/delivery progress instead

2. Logout button on App Home
Old behavior: Logs out all the way to the Login page
New behavior: Returns to previous page (opportunity list, learning/delivery progress, etc.)

conroy-ricketts:
@OrangeAndGreen 

> 2. Logout button on App Home

Great catch! I didn't think to test that

> 1. Back nav after launching app from learning/delivery progress pages

I'm having trouble reproducing this point. Are you able to provide repro steps?

OrangeAndGreen:
No problem, these steps will show the different behavior:
* In the opportunity list, press Resume on an opportunity in learning or one where delivery is complete (i.e. to navigate to the learning progress or delivery progress page instead of logging into the app directly)
* On the learning or delivery progress page, press the Resume button to login in to the app
* Swipe back on the App Home page (or press the back button if your device has one)

Results:
* Master: Navigate back to opportunity list
* This branch: Navigate back to learning/delivery progress page

OrangeAndGreen:
Did a pull on the latest code and the back nav is working as expected now, so only the Logout button behavior still to address.

app/src/org/commcare/activities/HomeScreenBaseActivity.java
```diff
@@ -586,10 +587,15 @@ protected void userTriggeredLogout() {
...
 
     protected void userPressedOpportunityStatus() {
-        Intent i = new Intent();
-        i.putExtra(REDIRECT_TO_CONNECT_OPPORTUNITY_INFO, true);
-        setResult(RESULT_OK, i);
-        finish();
+        // Launch the seated app's job status page on top of this (still-live) Home so the app
+        // session is preserved and backing out of the status page returns here.
+        ConnectJobRecord job = ConnectJobHelper.INSTANCE.getJobForSeatedApp(this);
+        if (job == null) {
```
OrangeAndGreen: Blocker: This seems crash-worthy to me, we should never show the Job Status button if not in a job context so the user clicking that button would indicate something is quite wrong.

app/src/org/commcare/fragments/connect/ConnectDeliveryProgressFragment.java
```diff
@@ -56,6 +55,7 @@ public class ConnectDeliveryProgressFragment extends ConnectJobFragment<Fragment
     private int initialTabPosition = 0;
     private boolean isProgrammaticTabChange = false;
     private ConnectDeliveryProgressViewModel viewModel;
+    private final ConnectAppLaunchController launchController = new ConnectAppLaunchController(this);
```
OrangeAndGreen: Seems like this could be a local variable in navigateToDeliverAppHome, or maybe use a static function instead?

app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java
```diff
@@ -204,136 +182,13 @@ private void launchAppForJob(ConnectJobRecord job, boolean isLearning) {
...
-        dismissLaunchDialog();
-        if (syncing) {
-            launchDialog = CustomProgressDialog.newInstance(
-                    Localization.get("sync.communicating.title"),
-                    Localization.get("sync.progress.starting"),
-                    LAUNCH_DIALOG_TASK_ID
-            );
-            launchDialog.addProgressBar();
+        if (isLearning) {
+            launchController.launchLearningApp(appId);
```
OrangeAndGreen: Nit: Feels like we're splitting a boolean just to combine it back together, i.e. here we switch on isLearning to either call launchLearningApp or launchDeliveryApp, and those two functions just hard-code true and false for isLearning before calling `launch(LaunchTarget(appId, isLearning))`. A simple `launchApp(appId, isLearning)` seems like it would be more straight-forward.

app/src/org/commcare/connect/ConnectAppLaunchController.kt
```diff
@@ -0,0 +1,231 @@
...
+        private var observedLifecycle: Lifecycle? = null
+
+        // Registered at construction (fragment init, the only valid time) so a back-out of the launched
+        // app's Home is delivered back here and can return the worker to the opportunities list.
+        private val homeResultLauncher: ActivityResultLauncher<Intent> =
+            fragment.registerForActivityResult(ActivityResultContracts.StartActivityForResult()) { result ->
+                onHomeResult(result)
+            }
+
+        fun launchLearningApp(appId: String) = launch(LaunchTarget(appId, isLearning = true))
```
OrangeAndGreen: Referenced in another comment but adding more detail here:
In all 3 usages of this class, we construct the class at fragment startup and then only use it in one place (technically two for he jobs list fragment but they're right next to each other and logically are like a single usage).

Is there a reason for the class to construct the launch controller before it's needed and hold a reference to it in the fragment? Otherwise I think a static launch method would be simpler and cleaner for callers.

OrangeAndGreen (changes requested):

app/src/org/commcare/activities/HomeScreenBaseActivity.java
conroy-ricketts: Great point

app/src/org/commcare/fragments/connect/ConnectDeliveryProgressFragment.java
conroy-ricketts: The reason for this not being a local variable or static is because `ConnectAppLaunchController` calls ` fragment.registerForActivityResult()` when the fragment is initialized

So keeping it global here avoids an `IllegalStateException` crash (i.e. the fragment attempting to `registerForActivityResult` **_after_** being created)

app/src/org/commcare/connect/ConnectAppLaunchController.kt
conroy-ricketts: Moving convo to earlier thread [here](https://github.com/dimagi/commcare-android/pull/3756#discussion_r3404612918)

app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java
conroy-ricketts: Yeah I was a little torn about this at first, and looking back at it again I agree your suggestion is better

app/src/org/commcare/activities/HomeScreenBaseActivity.java
conroy-ricketts: https://github.com/dimagi/commcare-android/commit/dab2abbabd3f2355653259dd70102e8e641f5c94

app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java
conroy-ricketts: https://github.com/dimagi/commcare-android/commit/4a1414e7c0c25bf03922d8ebcd061df93bc28657

app/src/org/commcare/connect/ConnectAppLaunchController.kt
```diff
@@ -0,0 +1,232 @@
...
+    constructor(
+        private val fragment: Fragment,
+        private val launcher: ConnectAppLauncher = ConnectAppLauncher(),
+    ) {
+        private var launchDialog: CustomProgressDialog? = null
+        private var showingSyncDialog = false
+        private var observedLifecycle: Lifecycle? = null
+
+        // Registered at construction (fragment init, the only valid time) so a back-out of the launched
+        // app's Home is delivered back here and can return the worker to the opportunities list.
```
shubham1g5: > can return the worker to the opportunities list

it should be whereever we launched the app from.

app/src/org/commcare/connect/ConnectAppLaunchController.kt
```diff
@@ -0,0 +1,232 @@
...
+
+                    override fun recoverFromSeatFailure() {
+                        val intent =
+                            Intent(activity, DispatchActivity::class.java)
+                                .addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP)
+                        fragment.startActivity(intent)
+                        activity.finish()
+                    }
+
+                    override fun fallBackToLegacyLaunch() {
```
shubham1g5: I probably missed it earlier, why do we ever want to fallback to legacy launch ?

app/src/org/commcare/connect/ConnectAppLaunchController.kt
```diff
@@ -0,0 +1,232 @@
...
+                        ConnectAppUtils.launchApp(activity, target.isLearning, target.appId)
+                    }
+
+                    override fun reportFailure(reason: String) {
+                        Logger.log(
+                            LogTypes.TYPE_ERROR_WORKFLOW,
+                            "Connect launch failed for app ${target.appId}: $reason",
+                        )
+                        FirebaseAnalyticsUtil.reportCccAppFailedAutoLogin(target.appId)
```
shubham1g5, line 159: think this should be logged for other failure modes as well

app/src/org/commcare/connect/ConnectAppLaunchController.kt
```diff
@@ -0,0 +1,232 @@
...
+        }
+
+        private fun onHomeResult(result: ActivityResult) {
+            if (!fragment.isAdded) {
+                return
+            }
+
+            val activity = fragment.requireActivity()
+            if (result.resultCode == Activity.RESULT_OK) {
+                // Route to the login screen via DispatchActivity on the user logging-out.
```
shubham1g5: think we should not do any special handling here for user logout and let user fallback naturally to the parent screen where user came back from (I realise this might be a  change from current behaviour but think that's the way we are going anyway with the redesign work that won't have any "Logout" action for connect launch.

app/src/org/commcare/connect/ConnectAppLaunchController.kt
```diff
@@ -0,0 +1,232 @@
...
+                    Intent(activity, DispatchActivity::class.java)
+                        .addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP or Intent.FLAG_ACTIVITY_SINGLE_TOP)
+                        .putExtra(LoginActivity.USER_TRIGGERED_LOGOUT, true)
+                        .putExtra(ConnectAppUtils.IS_LAUNCH_FROM_CONNECT, true)
+                activity.startActivity(intent)
+                return
+            }
+
+            // Back-out (RESULT_CANCELED): end the app session, return to the opportunities list.
+            CommCareApplication.instance().closeUserSession()
```
shubham1g5: This should already happen by code in `HomeScreenBaseActivity.userTriggeredLogout` ?

app/src/org/commcare/activities/DispatchActivity.java
```diff
@@ -116,6 +118,16 @@ protected void onCreate(Bundle savedInstanceState) {
         }
     }
 
+    @Override
+    protected void onNewIntent(@NonNull Intent intent) {
+        super.onNewIntent(intent);
+        // Adopt the new intent so dispatch() routes a reused (CLEAR_TOP | SINGLE_TOP) launch.
+        setIntent(intent);
```
shubham1g5: This seems a bit risky to me override the original intent without actually processing it, this can cause Dispatch to be in a state that doesn't correspond to the current intent set.

shubham1g5 (changes requested):
left some comments on https://github.com/dimagi/commcare-android/pull/3753 some of which are resolved in this PR but would be great if you can glance over and resolve the rest as part of this.

app/src/org/commcare/connect/ConnectAppLaunchController.kt
conroy-ricketts: Sounds good to me, yeah this was addressing an earlier comment that we were diverting from current behavior

conroy-ricketts: Actually, we may want to keep this as-is for now. I'm seeing some weird UX when I let the user naturally return to the screen they launched from. Here's an example where logging out would take the user back to the learn progress screen: 

https://github.com/user-attachments/assets/cfe3baf2-4517-4e4c-bfe4-47a2d07e8bd9

app/src/org/commcare/connect/ConnectAppLaunchController.kt
conroy-ricketts: This was added so that it also happens for the user on backing out of the app

app/src/org/commcare/connect/ConnectAppLaunchController.kt
conroy-ricketts: I think that would lead to poor UX. I know that this would be an uncommon scenario, but here's an example of us always returning the user to where they launched the app from (vs just taking them straight back to the opp list):

https://github.com/user-attachments/assets/de768847-8164-4217-bfe0-7e6728bda275

app/src/org/commcare/connect/ConnectAppLaunchController.kt
conroy-ricketts: The reason for this is so that we can reuse the existing error handling in `LoginActivity`

app/src/org/commcare/activities/DispatchActivity.java
conroy-ricketts: Can you elaborate more on this for me?

From what I understand, overriding `onNewIntent()` is a standard Android pattern for when we use `Intent.FLAG_ACTIVITY_SINGLE_TOP` to start an activity, and `onResume()` always runs after `onNewIntent()`. So, `DispatchActivity` ultimately re-reads `getIntent()` via `DispatchActivity.dispatch()`

app/src/org/commcare/connect/ConnectAppLaunchController.kt
conroy-ricketts: https://github.com/dimagi/commcare-android/commit/73ccd2668c93fed18e22b9da06f79a01ccb5822c

app/src/org/commcare/connect/ConnectAppLaunchController.kt
shubham1g5: think what we need to do here instead is to launch the Home Screen with a clear top flag so that we don't create duplicate instances of home screen in the app stack. 

Another thing we should do here is -> If user is already logged into the app, we should not perform the login process again but just do the post login routing directly.

app/src/org/commcare/connect/ConnectAppLaunchController.kt
shubham1g5: I would prefer to avoid introducing any dependency on the login screen workflow for connect launches from code maintenance perspective unless there is a good reason for us to do so. Curious what scenarios are you thinking of where we would need legacy error handling here and if it's not possible to add that in the new login workflow itself ?

app/src/org/commcare/connect/ConnectAppLaunchController.kt
shubham1g5: Backing out should not log out the user here, app session should stay alive as today - 1. Until the 24 hour session window expires or 2. User clicks logout.

app/src/org/commcare/connect/ConnectAppLaunchController.kt
shubham1g5: this should be resolved by suggestion here as well - https://github.com/dimagi/commcare-android/pull/3756#discussion_r3418321976

app/src/org/commcare/activities/DispatchActivity.java
shubham1g5: I don't have a concrete example here and I am mostly worried hypothetically about race conditions on intent manipulation that can happen depending upon when `onNewIntent` is fired vs where in dispatch activity the code pointer was processing the earlier intent. 


Is there a new block of code that launches dispatch with `FLAG_ACTIVITY_SINGLE_TOP` ?  It might be cleaner to just do a clean launch for dispatch given it's a no UI activity itself.

app/src/org/commcare/connect/ConnectAppLaunchController.kt
conroy-ricketts: Ah I see what you're getting at. I agree that we can probably avoid relying on existing error handling, especially given that these errors are meant to be "retryable". I'm going to look over this again and see what I can do

app/src/org/commcare/connect/ConnectAppLaunchController.kt
conroy-ricketts: Noted! Will tweak this

app/src/org/commcare/connect/ConnectAppLaunchController.kt
conroy-ricketts: Ah yeah, agreed!

conroy-ricketts: This makes the UX so much better!

https://github.com/user-attachments/assets/1f180cd5-86e1-4360-a0c5-9b0301a27012

app/src/org/commcare/connect/ConnectAppLaunchController.kt
conroy-ricketts: I think the UX concern I had is not resolved by the suggestion, but on second thought I think the UX is fine because the user has to log back in to resume the app anyways 👍🏾

app/src/org/commcare/activities/DispatchActivity.java
conroy-ricketts: I removed this now since there's no good reason for it anymore (while addressing the other comments) 👍🏾

app/src/org/commcare/connect/ConnectAppLaunchController.kt
conroy-ricketts: I think the best thing to do for one of these "retryable" errors may be to auto-retry once, and if we see the same error again, then dismiss the login dialog and show the user an error toast message. I'll quickly implement this now, but let me know your thoughts on this approach. I'll upload a quick video to demonstrate what this would look like.

For context, originally, the code would fallback to the legacy error handling for most of the `LoginError`'s (I think the only real exception here was `LoginError.TokenDenied` which got routed to `GlobalErrorUtil.triggerGlobalError()` instead).

shubham1g5: @conroy-ricketts Think for the current scope, it's completely fine to just show user the error and let user retry manually, I don't think we should be trying to retry actively as that can block user for a long time on a slow network here.

conroy-ricketts: Looking good!

https://github.com/user-attachments/assets/3ef45544-423f-4bea-b8e4-0e27cf1ca188

app/src/org/commcare/connect/ConnectAppLaunchController.kt
conroy-ricketts: https://github.com/dimagi/commcare-android/commit/a87d47628618a8065f1bd4329cb8158416f0b88b

app/src/org/commcare/connect/ConnectAppLaunchController.kt
conroy-ricketts: https://github.com/dimagi/commcare-android/commit/a87d47628618a8065f1bd4329cb8158416f0b88b

app/src/org/commcare/connect/ConnectAppLaunchController.kt
conroy-ricketts: https://github.com/dimagi/commcare-android/commit/8e15d67b9ea1d0467cbaac7e637295eaaaa8833f

app/src/org/commcare/activities/DispatchActivity.java
conroy-ricketts: https://github.com/dimagi/commcare-android/commit/c6c6c31aa3eee884b4e5288720a84b2029e51694

app/src/org/commcare/connect/ConnectAppLaunchController.kt
conroy-ricketts: https://github.com/dimagi/commcare-android/commit/d1c573e6987b089c08d3a9914182a5b293be2153

shubham1g5: this looks great, thanks!

app/src/org/commcare/connect/ConnectAppLaunchController.kt
shubham1g5: what I meant here is that even `promptRetry` and `recoverFromSeatFailure`  should log this as those indicate a failed attempt for login as well

app/src/org/commcare/fragments/connect/ConnectDeliveryProgressFragment.java
conroy-ricketts: There have been a lot of changes since this comment and this can be a local variable now

conroy-ricketts: https://github.com/dimagi/commcare-android/commit/81ec3c7852c1cf352fd9d22e8634a0fdce2e4951

app/src/org/commcare/connect/ConnectAppLaunchController.kt
conroy-ricketts: Ah I see, I think we should be good now though because all the failure paths will eventually call the `reportFailure()` function. So, the log will happen for an app seat failure [here](https://github.com/dimagi/commcare-android/blob/17383f3ef1cdddec332a9f2ddd73e7e0cec3a19e/app/src/org/commcare/connect/LaunchOutcomeRouter.kt#L29), and the log for all the retryable errors (which we show the prompt dialog for) will happen [here](https://github.com/dimagi/commcare-android/blob/17383f3ef1cdddec332a9f2ddd73e7e0cec3a19e/app/src/org/commcare/connect/LaunchOutcomeRouter.kt#L34).

So adding the logs to `promptRetry()` and `recoverFromSeatFailure()` would lead to duplicate logs unless we remove the log from `reportFailure()`

shubham1g5: Ahh didn't realise these callbacks are not mutually exclusive, resolving.

shubham1g5 (dismissed):

OrangeAndGreen (dismissed):

conroy-ricketts:
@OrangeAndGreen @shubham1g5 Added a commit for a final lint pass

OrangeAndGreen (approved):

PR #3758, opened 2026-06-10
Fix Nav Drawer Photo Launcher Registration Crash

conroy-ricketts:
## Technical Summary

This PR fixes a flaky BrowserStack CI test failure. 

The nav drawer's photo-edit launcher was being registered during drawer setup, which can run after the activity reaches `STARTED` and violates the `registerForActivityResult` contract. It now registers once in `BaseDrawerActivity.onCreate` and is injected into `BaseDrawerController`, so rebuilding the controller no longer re-registers it.

In CI the illegal re-registration crashed the activity mid-test whenever a home-screen test triggered a UI rebuild after the activity had resumed. Because drawer visibility is gated on a persisted preference, whether a given test hit the crash depended on shard ordering, which is why it surfaced as a flaky cascade of Espresso failures across home-screen tests.

### Safety story

What gives me confidence:
- I ran the change on a device and verified there are no regressions in the PersonalID profile photo update feature (open drawer, take/replace photo, failed-upload warning state).
- The diff is narrow and the registration now follows the documented Android contract (register before `STARTED`); I verified the lifecycle ordering so `BaseDrawerActivity.onCreate` registers the launcher before `onCreateSessionSafe → setupUI` ever builds a controller.

Risks to review:
- `BaseDrawerActivity` is the shared base for drawer-bearing activities, so the registration move affects all of them, not just the home screen.
- The original failure only reproduces on the BrowserStack Espresso suite (needs a device); I could not run that suite locally, so CI is the real confirmation that the cascade of home-screen test failures clears.

### DIFF
--- a/app/src/org/commcare/navdrawer/BaseDrawerActivity.kt
+++ b/app/src/org/commcare/navdrawer/BaseDrawerActivity.kt
@@ -1,8 +1,11 @@
 package org.commcare.navdrawer
 
+import android.content.Intent
 import android.os.Bundle
 import android.view.MenuItem
 import android.view.View
+import androidx.activity.result.ActivityResultLauncher
+import androidx.activity.result.contract.ActivityResultContracts
 import org.commcare.activities.CommCareActivity
 import org.commcare.connect.ConnectActivityCompleteListener
 import org.commcare.connect.ConnectNavHelper.unlockAndGoToConnectJobsList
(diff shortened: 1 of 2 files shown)


conroy-ricketts:
## Suggested Review Order

- `app/src/org/commcare/navdrawer/BaseDrawerController.kt` — shows the removed in-controller `registerForActivityResult` and the new injected launcher + `handlePhotoResult`.
- `app/src/org/commcare/navdrawer/BaseDrawerActivity.kt` — shows where registration moved to: once in `onCreate` (before STARTED), then passed into the controller.

OrangeAndGreen (approved):

PR #3763, opened 2026-06-16
Supporting Subtitle in Action Bar

OrangeAndGreen:
https://dimagi.atlassian.net/browse/CCCT-2513

## Product Description
No user visible changes in this PR (supporting upcoming changes)

## Technical Summary
Supports showing a subtitle in the action bar via two changes:
* Set subtitle text white in action bar layout (to ensure correct coloring)
* Helper methods to set and clear the subtitle (so it doesn't remain when a caller fails to clear it manually).

### Safety story
The subtitle isn't being used anywhere yet so this is a no-opp change for now.
It will be used in the upcoming UI refactors for opp intro, learning progress, and delivery progress.

For now, I tested by manually inserting a call to setActionBarTitle in the opp intro page:
image

### Automated test coverage
None

### DIFF
--- a/app/src/org/commcare/activities/CommonBaseActivity.java
+++ b/app/src/org/commcare/activities/CommonBaseActivity.java
@@ -32,6 +32,26 @@ public class CommonBaseActivity extends AppCompatActivity {
         }
     }
 
+    /**
+     * Sets the action bar title with no subtitle, clearing any subtitle a previous page set so the
+     * bar keeps its single-line appearance.
+     */
+      public void setActionBarTitle(@Nullable CharSequence title) {
+        setActionBarTitle(title, null);
+      }
+
+      /**
+      * Sets the action bar title and an optional second line. Pass a {@code null} subtitle for the
+      * standard single-line bar; a non-null subtitle renders below the title.
+      */
+      public void setActionBarTitle(@Nullable CharSequence title, @Nullable CharSequence subtitle) {
+          ActionBar actionBar = getSupportActionBar();
+          if (actionBar != null) {
+              actionBar.setTitle(title);
+              actionBar.setSubtitle(subtitle);
+          }
+      }
+
     @Override
     public boolean onMenuOpened(int featureId, Menu menu) {
         wasOptionsMenuOpen = true;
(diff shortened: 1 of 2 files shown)


app/src/org/commcare/activities/CommonBaseActivity.java
```diff
@@ -32,6 +32,26 @@ public void onPostCreate(@Nullable Bundle savedInstanceState) {
...
+      * Sets the action bar title and an optional second line. Pass a {@code null} subtitle for the
+      * standard single-line bar; a non-null subtitle renders below the title.
+      */
+      public void setActionBarTitle(@Nullable CharSequence title, @Nullable CharSequence subtitle) {
+          ActionBar actionBar = getSupportActionBar();
+          if (actionBar != null) {
+              actionBar.setTitle(title);
+              actionBar.setSubtitle(subtitle);
+          }
+      }
```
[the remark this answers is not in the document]
OrangeAndGreen: [7d764d1](https://github.com/dimagi/commcare-android/pull/3763/commits/7d764d1cc19c576e1782b487010452c52af70934)

shubham1g5 (approved):

app/src/org/commcare/activities/CommonBaseActivity.java
```diff
@@ -32,6 +32,26 @@ public void onPostCreate(@Nullable Bundle savedInstanceState) {
...
+        setActionBarTitle(title, null);
+    }
+
+    /**
+     * Sets the action bar title and an optional second line. Pass a {@code null} subtitle for the
+     * standard single-line bar; a non-null subtitle renders below the title.
+     */
+    public void setActionBarTitle(@Nullable CharSequence title, @Nullable CharSequence subtitle) {
+        ActionBar actionBar = getSupportActionBar();
+        if (actionBar != null) {
```
conroy-ricketts, line 49: Should we rather assert the action bar is not null (i.e. crash if it is null) because it'd be an illegal state when calling this function?

OrangeAndGreen: I'm going to leave this as is because it's the pattern currently used throughout the code, although I agree it's worth revisiting. I didn't see any places that require the support action bar, although perhaps some should (this one included). But more importantly, I see four places where we aren't null-checking the result of getSupportActionBar.

PR #3765, opened 2026-06-17
CCCT-2441 Silent App Launch For Job Intro And Download Screens

conroy-ricketts:
## Product Description

Completes the rollout of the silent Connect app-launch path to its last two surfaces: starting a job from the opportunity intro screen, and the launch that runs right after an app finishes downloading/installing. These now open the app directly behind the single progress dialog — no `LoginActivity` / `SeatAppActivity` flashing through — and backing out of the app returns to the originating Connect screen rather than the transient setup screen. Backing out past the opportunity list then exits the app instead of re-opening the launched app. After repeated launch failures (two retries), the failure dialog now shows a dismiss-only "try again later" message instead of offering endless retries.

Here's a video of the retry prompt dialog:

https://github.com/user-attachments/assets/f7603d98-9f9c-4385-b9c9-ae780574fada

Here's a video of the download and job intro pages. Note that I show two different cases: 1) the app was pre-installed and thus the download page is skipped, 2) the app was not pre-installed and we show the download page:

https://github.com/user-attachments/assets/2facde91-1226-4251-83f9-48cf5da6513c

## Technical Summary

- Migrates `ConnectJobIntroFragment` and `ConnectDownloadingFragment` onto `ConnectAppLaunchController`, the shared silent-launch orchestration already used by the other Connect surfaces.
- Reproduces the back navigation the old `DispatchActivity` path used to provide, but without `DispatchActivity`: `ConnectActivity` remembers a Connect-initiated launch and exits the page when the user backs out of the opportunity list, and an already-running app is brought forward with `FLAG_ACTIVITY_REORDER_TO_FRONT`.
- Removes the now-dead `IS_LAUNCH_FROM_CONNECT` flag, the auto-login branch in `LoginActivity`, and the `CONNECT_MANAGED_LOGIN` / `redirectToConnectHome` back-out path in `DispatchActivity` — none have remaining writers once these two surfaces stop using the legacy path.
- Hardens the launcher: bails if the fragment view is already gone; soundly detects an existing session for the target app; defers the back-stack pop until Home is in front to avoid a blank-frame flash; and reports the terminal outcome to the calling fragment (`onLaunchSucceeded` / `onLaunchFailed`) while keeping ownership of the failure dialogs.
- `LaunchOutcomeRouter` escalates a retryable failure to a dismiss-only message once `MAX_LAUNCH_ATTEMPTS` (3) is reached.

### Safety story

What gives me confidence:
- I tested on a device: launched apps from both the job-intro **Start** button and the post-download surface, including launching several not-yet-installed apps back-to-back, landing on Home with no login/app-setup screens flashing through.
- I device-tested the back navigation: backing out of a launched app returns to the Connect screen, backing out past the opportunity list exits the app, and re-launching an already-open app reuses its Home without the back stack collapsing.
- I reproduced the back-to-back `NoSuchElementException` crash before the session-shortcut fix and confirmed it no longer occurs after.
- I exercised the new post-2-retries escalated dialog on a device by temporarily forcing the launch to always error.
- The routing/escalation logic is covered by `LaunchOutcomeRouter` and `ConnectAppLauncher` unit tests, and the change is scoped to the Connect launch path.
- The removed flag and branches are dead code — I verified there are no remaining writers/readers once the two surfaces are migrated.

Risks to review:
- Touches the churn-prone `LoginActivity` / `DispatchActivity` by deleting the auto-login and `CONNECT_MANAGED_LOGIN` / `redirectToConnectHome` back-out paths. This is behavior-preserving (those paths were only reachable via the now-removed flag) but worth a careful read.
- The dismiss-only failure dialog and the controller / `ConnectActivity` back-stack handling (the exit-on-back and `REORDER_TO_FRONT` launch) are Android-bound and not unit-tested; they were verified manually on-device.
- The Hausa and Tigrinya translations for the new strings are best-effort and should get a native-speaker review.

### Automated test coverage

`LaunchOutcomeRouterTest` covers the retry → escalation threshold (below the limit prompts retry; at the limit shows the persistent error). The existing `ConnectAppLauncherTest` covers the launcher's seat / login / outcome seams. The fragment-side dialog and lifecycle code is not unit-covered.

[CCCT-2441]: https://dimagi.atlassian.net/browse/CCCT-2441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


conroy-ricketts:
## Suggested Review Order

- `app/src/org/commcare/connect/ConnectAppUtils.kt` — start with the deleted `launchApp` (the legacy path being replaced)
- `app/src/org/commcare/connect/ConnectConstants.java`, `app/src/org/commcare/utils/FirebaseMessagingUtil.java` — removal of the dead `IS_LAUNCH_FROM_CONNECT` / `CONNECT_MANAGED_LOGIN` constants and import
- `app/src/org/commcare/activities/LoginActivity.java` — removal of the now-unreachable `appLaunchedFromConnect` auto-login branch
- `app/src/org/commcare/activities/DispatchActivity.java` — removal of the matching `CONNECT_MANAGED_LOGIN` / `redirectToConnectHome` back-out path
- `app/src/org/commcare/connect/LaunchOutcomeRouter.kt` — the `LaunchActions` seam + new retry-escalation decision
- `app/src/org/commcare/connect/ConnectAppLauncher.kt` — silent launcher; note the sound "already logged in" check
- `app/src/org/commcare/connect/ConnectAppLaunchController.kt` — fragment orchestration: view guard, deferred back-stack pop, persistent-error dialog, failure counting
- `app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.java`, `app/src/org/commcare/fragments/connect/ConnectDownloadingFragment.java` — the two migrated call sites
- `app/unit-tests/src/org/commcare/connect/LaunchOutcomeRouterTest.kt` — escalation-threshold coverage
- `app/res/values/strings.xml` (+ 9 locale files) — new dialog strings and translations
- `docs/commcare/login-engine.md` — updated behavior docs

OrangeAndGreen (changes requested):
[FIXED] The code changes look great but I'm seeing a functionality issue. Here are steps to reproduce:
* Go to the opportunity list and press Resume on an opp where the learn/deliver app needs to be downloaded
* Wait while the Download page appears, the app downloads, and logs in
* Once on App Home, navigate back (i.e. swipe)

The code should navigate to the opportunity list, but instead it navigates to the Download page (showing complete progress)

OrangeAndGreen:
[FIXED] I'm also getting a repeatable crash that seems like it might be related to these changes.

Steps:
* Go to opportunity list, Resume on a job that requires download
* Once that completes, navigate back to opportunity list (through the download page as described by the previous bug)
* Now resume a different opp that requires download
* Wait for the download to complete and login to succeed
* The app crashes loading App Home

Here's the exception:
java.util.NoSuchElementException: No record in table user_key_records for ID 1
                                                                                                    	at org.commcare.models.database.SqlStorage.readBytes(SqlStorage.java:455)
                                                                                                    	at org.commcare.models.database.SqlStorage.read(SqlStorage.java:446)
                                                                                                    	at org.commcare.services.CommCareSessionService.getUserKeyRecord(CommCareSessionService.java:476)
                                                                                                    	at org.commcare.CommCareApplication.getRecordForCurrentUser(CommCareApplication.java:1000)
                                                                                                    	at org.commcare.activities.StandardHomeActivity.preparePinMenu(StandardHomeActivity.java:182)
                                                                                                    	at org.commcare.activities.StandardHomeActivity.onPrepareOptionsMenu(StandardHomeActivity.java:172)
                                                                                                    	at android.app.Activity.onPreparePanel(Activity.java:4480)
                                                                                                    	at androidx.activity.ComponentActivity.onPreparePanel(ComponentActivity.java:511)
                                                                                                    	at androidx.appcompat.view.WindowCallbackWrapper.onPreparePanel(WindowCallbackWrapper.java:99)
                                                                                                    	at androidx.appcompat.app.AppCompatDelegateImpl$AppCompatWindowCallback.onPreparePanel(AppCompatDelegateImpl.java:3097)
                                                                                                    	at androidx.appcompat.view.WindowCallbackWrapper.onPreparePanel(WindowCallbackWrapper.java:99)
                                                                                                    	at androidx.appcompat.app.ToolbarActionBar$ToolbarCallbackWrapper.onPreparePanel(ToolbarActionBar.java:523)
                                                                                                    	at androidx.appcompat.app.ToolbarActionBar.populateOptionsMenu(ToolbarActionBar.java:457)
                                                                                                    	at androidx.appcompat.app.ToolbarActionBar$1.run(ToolbarActionBar.java:57)
                                                                                                    	at android.view.Choreographer$CallbackRecord.run(Choreographer.java:1339)
                                                                                                    	at android.view.Choreographer$CallbackRecord.run(Choreographer.java:1348)
                                                                                                    	at android.view.Choreographer.doCallbacks(Choreographer.java:952)
                                                                                                    	at android.view.Choreographer.doFrame(Choreographer.java:878)
                                                                                                    	at android.view.Choreographer$FrameDisplayEventReceiver.run(Choreographer.java:1322)
                                                                                                    	at android.os.Handler.handleCallback(Handler.java:958)
                                                                                                    	at android.os.Looper.loopOnce(Looper.java:205)
                                                                                                    	at android.os.Looper.loop(Looper.java:294)
                                                                                                    	at android.app.ActivityThread.main(ActivityThread.java:8177)
                                                                                                    	at java.lang.reflect.Method.invoke(Native Method)
                                                                                                    	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:552)
                                                                                                    	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:971)

app/src/org/commcare/activities/DispatchActivity.java
```diff
@@ -36,8 +36,6 @@
 import static org.commcare.activities.LoginActivity.EXTRA_APP_ID;
 import static org.commcare.activities.LoginActivity.EXTRA_FORCE_SINGLE_APP_MODE;
 import static org.commcare.commcaresupportlibrary.CommCareLauncher.SESSION_ENDPOINT_APP_ID;
-import static org.commcare.connect.ConnectAppUtils.IS_LAUNCH_FROM_CONNECT;
```
shubham1g5, line 39: great to see these states cleaned up from here :)

app/src/org/commcare/connect/ConnectAppLaunchController.kt
```diff
@@ -81,6 +81,9 @@ class ConnectAppLaunchController
         ) = launch(LaunchTarget(appId, isLearning))
 
         private fun launch(target: LaunchTarget) {
+            if (fragment.view == null) {
+                return
+            }
```
shubham1g5, line 100: Does this achive something else than `fragment.isAdded()` check ? If not, it would be nice to be consistent here and use the `isAdded()` check itsef.

app/src/org/commcare/connect/ConnectAppLaunchController.kt
```diff
@@ -118,12 +127,16 @@ class ConnectAppLaunchController
...
                     override fun dismissProgress() = dismissLaunchDialog()
 
-                    override fun launchHome() =
+                    override fun launchHome() {
                         activity.startActivity(
                             HomeScreenBaseActivity
                                 .buildHomeLaunchIntent(activity)
                                 .addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP),
                         )
+                        if (target.popLaunchingScreenOnSuccess) {
```
shubham1g5: Should this happen before the call to `startActivity`, given we are adding a lifecycle listener here and doing it after `startActivity` runs the risk of listener getting registered after the lifecycle change.

app/src/org/commcare/connect/ConnectAppLaunchController.kt
```diff
@@ -200,6 +213,20 @@ class ConnectAppLaunchController
             launchDialog = null
         }
 
+        private fun popLaunchingScreenOnceHidden() {
```
shubham1g5: this creates a pre-condition for this class that all callers implement a `NavHostFragment` , think we should validate this at constructor level and throw an IllegalStateEx if there is no`NavHostFragment` .

app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.java
```diff
@@ -149,7 +149,7 @@ public void onSuccess(Boolean success) {
                 String appId = job.getLearnAppInfo().getAppId();
                 boolean appInstalled = AppUtils.isAppInstalled(appId);
                 if (appInstalled) {
-                    new ConnectAppLaunchController(ConnectJobIntroFragment.this).launchApp(appId, true);
+                    new ConnectAppLaunchController(ConnectJobIntroFragment.this).launchApp(appId, true, true);
```
shubham1g5: Reg. back behaviour here, does The Job Intro page do not appear once user goes into the app ? If so, is that because of an underlying opportunity state change that happens when we launch the App from this page ?

app/src/org/commcare/connect/ConnectAppLauncher.kt
```diff
@@ -40,6 +40,25 @@ sealed class LaunchOutcome {
...
+    val instance = CommCareApplication.instance()
+    if (instance.currentApp?.uniqueId != appId) {
+        return false
+    }
+    val sessionUuid = instance.session.userKeyRecordUUID ?: return false
+    return instance
+        .getAppStorage(UserKeyRecord::class.java)
+        .getRecordsForValue(UserKeyRecord.META_SANDBOX_ID, sessionUuid)
+        .isNotEmpty()
+}
```
shubham1g5: this check should strictly happen with the Currently logged into Personal ID user id to establish that it's the current personal ID user that's logged into the app and not just any user.

app/src/org/commcare/connect/ConnectAppLaunchController.kt
```diff
@@ -74,13 +81,19 @@ class ConnectAppLaunchController
...
         private var showingSyncDialog = false
         private var observedLifecycle: Lifecycle? = null
+        private var failedAttempts = 0
 
+        @JvmOverloads
         fun launchApp(
             appId: String,
             isLearning: Boolean,
-        ) = launch(LaunchTarget(appId, isLearning))
+            popLaunchingScreenOnSuccess: Boolean = false,
```
shubham1g5: I don't like that the controller here is manipulating the fragment stack and think it's extending it's responsibilities more than it need to, instead the calling fragment should handle it's own state i.e. it should finish itself on Resume when there is an undelying state change after pressing back that makes that fragment no longer relevant.

shubham1g5 (changes requested):

app/src/org/commcare/connect/ConnectAppLaunchController.kt
Jignesh-dimagi: Yes, I agree with that comment. I think the best approach is to have `ConnectAppLaunchController -> handleLaunchOutcome` inform the calling fragment of the final outcome, and let that fragment manage its own lifecycle. Maybe we can use an abstract method in `BaseConnectFragment` to handle this, as the controller is currently overloaded with lifecycle and fragment management responsibilities.

Jignesh-dimagi (changes requested):

Jignesh-dimagi:
@conroy-ricketts 
I am facing another issue, which may be unrelated to these specific changes and likely from a previous PR. The navigation flow is as follows: Connect Opportunity List -> click on resume  → App Home Screen → press back → Connect Opportunity List (again) → press back → the App Home Screen opens again.

OrangeAndGreen:
Noting the two issues I reported are resolved with the latest code

app/src/org/commcare/connect/ConnectAppLaunchController.kt
conroy-ricketts: I'm on board with that!

app/src/org/commcare/connect/ConnectAppLaunchController.kt
conroy-ricketts: That seems like a good reason to me to reorder this

app/src/org/commcare/connect/ConnectAppLaunchController.kt
conroy-ricketts: I refactored this while addressing the other comments, and the illegal state should be handled in my refactor by the `BaseConnectFragment`

app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.java
conroy-ricketts: If I understand your question correctly, it's not driven by an opp state change. The job intro page does not reappear on a back press because we're explicitly removing the fragment from the back stack after a successful app launch

I tested the existing behavior here on the `master` branch as well, and it seems that the back press navigation was being routed through DispatchActivity so popping the back stack was not necessary

app/src/org/commcare/connect/ConnectAppLauncher.kt
conroy-ricketts: Agreed

app/src/org/commcare/connect/ConnectAppLaunchController.kt
conroy-ricketts: This was intentional because a fragment can be "added" while the view is `null` (e.g. after `onDestroyView()`). So this prevents a potential `IllegalStateException` when we later use `fragment.viewLifecycleOwner`

conroy-ricketts:
> I am facing another issue, which may be unrelated to these specific changes and likely from a previous PR. The navigation flow is as follows: Connect Opportunity List -> click on resume → App Home Screen → press back → Connect Opportunity List (again) → press back → the App Home Screen opens again.

This is a great catch

@Jignesh-dimagi @OrangeAndGreen @shubham1g5 

I'm torn on the best approach here because this does not seem to be a simple fix (please correct me if I'm wrong).

The crux of the issue is that we are finishing the `ConnectActivity`, but `DispatchActivity` is re-launching the CC app.

Thoughts on this?

Unfortunately, I think we may need to re-introduce some sort of "launched from Connect" flag inside `DispatchActivity` so that we can set the `shouldFinish` flag to `true` and exit `DispatchActivity` before it routes to another screen

OrangeAndGreen:
> DispatchActivity is re-launching the CC app

I think if we close the app session (i.e. logout of the app) then DispatchActivity will route to LoginActivity instead of opening StandardHomeActivity again.

app/src/org/commcare/connect/ConnectAppLaunchController.kt
conroy-ricketts: https://github.com/dimagi/commcare-android/commit/7445c7e8a6a08492bad2f43990c22ead1f26ab5a

app/src/org/commcare/connect/ConnectAppLaunchController.kt
conroy-ricketts: https://github.com/dimagi/commcare-android/commit/9da4383e8eb8da05728590bcef72f4ee21169aa9

app/src/org/commcare/connect/ConnectAppLauncher.kt
conroy-ricketts: https://github.com/dimagi/commcare-android/commit/99a8f624b0f7b36d022bacb1832fe63becbe9370

conroy-ricketts:
@OrangeAndGreen

Are we fine with closing the app session regardless of what the user was doing before they landed on the jobs list?

i.e. do we care to preserve this behavior (user logs in to an app → opens job list **_but_** does not launch another app → user backs out → user sees the same app they were already logged in to)?:

https://github.com/user-attachments/assets/6266ca0d-3cb5-410e-9211-116bb3cb933b

Jignesh-dimagi:
> @OrangeAndGreen
> 
> Are we fine with closing the app session regardless of what the user was doing before they landed on the jobs list?
> 
> i.e. do we care to preserve this behavior (user logs in to an app → opens job list **_but_** does not launch another app → user backs out → user sees the same app they were already logged in to)?:
> 
>  Screen_Recording_20260618_155948_CommCare.Debug.mp4

I was unable to view your video, as it stops at 2 seconds for some reason. However, I see two possible solutions here:
- Whenever a user lands on the opportunity list, we should assume they want to log into an app, and the code should close the previous session at this point.
- Pressing the back button on StandHomeActivity should be treated as a logout (for Connect users only), and the session should be closed there. However, this should not apply to traditional CommCare users, as their login needs to be preserved. 

I personally prefer the second option, but I am not sure if it is feasible to distinguish a Connect user while on StandHomeActivity.

app/src/org/commcare/connect/ConnectAppLauncher.kt
shubham1g5: thanks, one other suggestion here - this method seems quite useful outside this class and we can benefit from moving it to a class like `PersonalIDManager`

app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.java
shubham1g5: Think my question here is whether we know if there is a reason we don't want user to get back to the "Intro Page" on pressing back from the App home, I find it weird that user is not returned to the page they launched the app from here. 

Probably a slightly related question -> Once the app is launched and user is on App home, we go back to the Opp List and go into the opportunity again , at this stage do we see the Job Intro page again ?

shubham1g5:
> I think if we close the app session (i.e. logout of the app) then DispatchActivity will route to LoginActivity instead of opening StandardHomeActivity again.

We should not be closing app sessions unless user logs into another app, the reason is that there are several periodic tasks on CC  Appside to sync data, heartbeat, app update etc that depends on user being logged into the app and we would want to keep the session alive as per normal CommCare session timing rules (we log out the user automatically every 24 hours)  for both Connect and CommCare apps. 


> Unfortunately, I think we may need to re-introduce some sort of "launched from Connect" flag inside DispatchActivity so that we can set the shouldFinish flag to true and exit DispatchActivity before it routes to another screen

Think that's a decent solution for time-being and I presume this flow to undergo changes with the upcoming re-design which would be a better time to address this issue.

app/src/org/commcare/connect/ConnectAppLaunchController.kt
shubham1g5: Ahh this implementation differs from what I was trying to suggest. I don't think we need to implement a lifecycle observer at all here, taking example of the Download Fragment, we can do something like - 

````
onResume() {
 if(appInstalled){
  goBack()
}
}
````

This keeps this code simple and let the fragment decide the back behaviour based on underlying state of app rather than being dependent on another process to tell it to go back. 

In general reg. back behaviour, our philosophy should be that user should return to the page they navigated forward from except in cases when the page itself is no longer a valid destination due to an underlying change in App State, in which cases the screen should itself identify that it is no longer a valid destination and either finish itself or return the logic to it's parent screen. 

Thoughts ?

Comment (Jignesh-dimagi) [reply app/src/org/commcare/connect/ConnectAppLaunchController.kt]: The calling fragment (such as ConnectJobIntroFragment or ConnectDownloadingFragment) that attempts to launch the app currently does not receive feedback on whether the launch succeeded or failed. Given that we are going through many design changes, this feedback is essential so the calling fragment can update its UI status, transition states, or display errors properly.
With that in mind, here are the proposed application flows for ConnectAppLaunchController:
On App Launch Success: Launch home activity → Inform calling fragment (success)
On App Launch Failure: Inform calling fragment (failure)
We should let the calling fragment make the final decision (such as navigating back) based on the received success or failure result.

**Comment by conroy-ricketts** [comment]
Thanks for the input everyone!

I think I'll try to find the best way of re-introducing a "launched from Connect" flag to fix this for now

File: app/src/org/commcare/connect/ConnectAppLaunchController.kt (hunk 7 above)
Comment (conroy-ricketts) [reply app/src/org/commcare/connect/ConnectAppLaunchController.kt]: I think I like both of your suggestions, and I'm going to explore these while addressing the other comments

For context, the reason why I was popping the backstack at all was because during dev testing I noticed that the screen would flash on backing out which was bad UX in my opinion

File: app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.java (hunk 5 above)
Comment (conroy-ricketts) [reply app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.java]: > Think my question here is whether we know if there is a reason we don't want user to get back to the "Intro Page" on pressing back from the App home, I find it weird that user is not returned to the page they launched the app from here.

Ah I see, my approach here was to mirror the existing behavior. Testing the code (in a state before any of the login refactor was implemented), I saw that we were skipping the Job Intro page on backing out and taking the user straight back to the Jobs List page

> Probably a slightly related question -> Once the app is launched and user is on App home, we go back to the Opp List and go into the opportunity again , at this stage do we see the Job Intro page again ?

The Job Intro page will not be seen again. In this scenario, when the user goes back to the Jobs List page, the job tile will be in the "in progress (learn)" state and will no longer show as a brand new opportunity. This also mirrors the existing behavior before the login refactor.

So yeah, going back to the original question, there is indeed an underlying opp state change happening here. I misunderstood the question at first

File: app/src/org/commcare/connect/ConnectAppLauncher.kt (hunk 6 above)
Comment (conroy-ricketts) [reply app/src/org/commcare/connect/ConnectAppLauncher.kt]: That's a great idea, going to knock this out before the other stuff

File: app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.java (hunk 5 above)
Comment (OrangeAndGreen) [reply app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.java]: >  if there is a reason we don't want user to get back to the "Intro Page" on pressing back from the App home
Noting that this was all designed to Jon's requirements when he requested back-nav from App Home to the Opportunity List in order to make demos easier. We're probably good to rethink how it works as part of our navigation refactor, but should make sure Jon's still happy with the new behavior (I suspect he will be as long as we're keeping things easy to access and jump around).

File: app/src/org/commcare/connect/ConnectAppLauncher.kt (hunk 6 above)
Comment (conroy-ricketts) [reply app/src/org/commcare/connect/ConnectAppLauncher.kt]: https://github.com/dimagi/commcare-android/commit/94de0312c959d50f0528cec10da820e55a22737f

**Comment by conroy-ricketts** [comment]
@OrangeAndGreen @Jignesh-dimagi @shubham1g5 

I chose to keep the session open and added a session-scoped flag (`sessionLaunchedFromConnect`) to be used inside `DispatchActivity.dispatch()`

This fixes two issues:

1. The user backing out of the jobs list and incorrectly seeing the live CC app home screen (the issue that Jignesh originally pointed out).
2. The user backing out again and again until the entire mobile app closes, and then not seeing the Login screen when they open the mobile app again (another bug I discovered while dev testing).

Moreover, I explored the option of reporting both launch successes and failures to the fragment, and I think it's not necessary because all failures that can occur while launching a CC app is already handled by the launch controller before the user lands on the fragment. So I kept the success callback only and renamed it to `onLaunchSucceeded` to be more explicit.

I kept the "pop backstack" logic as-is for now to prevent the screen from flashing for the user when they back out of the CC app home screen.

Commit - https://github.com/dimagi/commcare-android/commit/392c0b129eac28a66b2646387c0f0102c2dedda3

I think that covers everything remaining that's been said so far. Please let me know if I missed anything or if there are any strong opinions

**Comment by Jignesh-dimagi** [comment]
> Moreover, I explored the option of reporting both launch successes and failures to the fragment, and I think it's not necessary because all failures that can occur while launching a CC app is already handled by the launch controller before the user lands on the fragment. So I kept the success callback only and renamed it to onLaunchSucceeded to be more explicit.

This is the current requirement. However, as we move forward with the latest design changes, we need to inform the calling fragment of both success and failure. Please see the example below, where launch controller error processing is not required, yet the calling fragments still need to update the UI based on the error. Additionally, as previously discussed, we need to reduce the responsibility of the launch controller and avoid handling the fragment lifecycle within it.

OpportunityHome_New Opportunity 1 2


>I kept the "pop backstack" logic as-is for now to prevent the screen from flashing for the user when they back out of the CC app home screen.

I think if launch controller starts the home activity first and then informs the calling fragment, where it will pop back and should not give flashing or UI jerk.

File: app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.java (hunk 5 above)
Comment (shubham1g5) [reply app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.java]: > In this scenario, when the user goes back to the Jobs List page, the job tile will be in the "in progress (learn)" state and will no longer show as a brand new opportunity. 

Got it, And this transition from new to in progress opportunity happens on user downloading the app successfully ? Or just on clicking "Download" ? I am mostly tying to confirm that we do not need to modify the back behaviour when an app downoad was not successful here.

File: app/src/org/commcare/activities/DispatchActivity.java [hunk 8]
```diff
@@ -242,14 +238,14 @@ private void dispatch() {
...
                     handleShortcutLaunch();
-                } else if (redirectToConnectHome) {
-                    redirectToConnectHome = false;
-                    CommCareApplication.instance().closeUserSession();
-                    ConnectNavHelper.INSTANCE.goToConnectJobsList(this);
                 } else if (redirectToConnectOpportunityInfo) {
                     redirectToConnectOpportunityInfo = false;
                     ConnectJobRecord job = ConnectJobHelper.INSTANCE.getJobForSeatedApp(this);
                     ConnectNavHelper.INSTANCE.goToActiveInfoForJob(this, job, true);
+                } else if (CommCareApplication.instance().sessionLaunchedFromConnect) {
```
shubham1g5: Login Screen should not be launched from here as login screen is only a valid screen when there is no active user session. In the example `Connect Opportunity List -> click on resume → App Home Screen → press back → Connect Opportunity List (again) → press back → the App Home Screen opens again. `, given user is starting from Opp List, user should instead exit the app on clicking back from the app list.

I would also prefer passing any state as an intent extra to the DispatchActivity instead of maintaining an exposed live static variable across the lifecycle of user session.

shubham1g5 (changes requested):

app/src/org/commcare/activities/DispatchActivity.java
conroy-ricketts: > given user is starting from Opp List, user should instead exit the app on clicking back from the app list.

Ah, I thought we wanted to avoid that. I think that makes this easier! Tweak coming soon

> login screen is only a valid screen when there is no active user session

Thanks for the context, noted!

app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.java
conroy-ricketts: This transition happens on the user tapping the "Start" button on the Job Intro page.

When that button is tapped, we hit the `/users/start_learn_app/` API endpoint, and we flip the job status to Learn on that API's success.

So, the job status flips _before_ the user sees the Download page

conroy-ricketts: So, the back behavior on a failed download should be good because backing out will land the user on the Jobs List page (with the opp in the Learn state along with the tile's download icon present)

conroy-ricketts: Noting that the Download page would show an error before the user backs out. As an example:

https://github.com/user-attachments/assets/5cc63bea-8d5d-4ae6-ae49-b30705e9d61f

conroy-ricketts:
@Jignesh-dimagi @shubham1g5 @OrangeAndGreen 

This is ready for another round of review. I proceeded with solution A as described in [this](https://docs.google.com/document/d/1jiVEbljnR8abPwJnKzqULTEqbAxU_PAt9sRWkPjTB9k/edit?usp=sharing) doc. I believe the cons are worth it given that the app navigation behavior will change significantly with the Connect Redesign. I dev tested this to make sure the edge cases match the desired behavior. Commit - d49fb11bb01c31c2c9a9a7b158ae6ac34e19c5ad.

> However, as we move forward with the latest design changes, we need to inform the calling fragment of both success and failure.

@Jignesh-dimagi I implemented a minimal version of this for now (to be used later). Moving ownership of failure handling to the fragment (and reducing the responsibility of the launch controller) will be a decent scope increase and I think that's worth a new ticket. Commit - 2353f1cc140d58f0a4b7820cd0b05669ccd48ef3.

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
```diff
@@ -276,4 +279,26 @@ abstract class BaseConnectFragment<B : ViewBinding> :
...
         }
     }
+
+    /**
+     * Pops this fragment off the navigation back stack the next time it is hidden (its [onStop]).
+     * A fragment that has launched another screen on top calls this so the user does not return to
+     * it on back; deferring to onStop keeps the pop from briefly flashing the destination beneath.
+     */
+    protected fun popSelfOnceHidden() {
+        lifecycle.addObserver(
```
Jignesh-dimagi, line 289: @conroy-ricketts Why its required to add the lifecycle just for popping the fragment off the back stack. Just `NavHostFragment.findNavController(this).popBackStack()` should be sufficient.

Jignesh-dimagi:
> @Jignesh-dimagi @shubham1g5 @OrangeAndGreen
> 
> This is ready for another round of review. I proceeded with solution A as described in [this](https://docs.google.com/document/d/1jiVEbljnR8abPwJnKzqULTEqbAxU_PAt9sRWkPjTB9k/edit?usp=sharing) doc. I believe the cons are worth it given that the app navigation behavior will change significantly with the Connect Redesign. I dev tested this to make sure the edge cases match the desired behavior. Commit - [d49fb11](https://github.com/dimagi/commcare-android/commit/d49fb11bb01c31c2c9a9a7b158ae6ac34e19c5ad).
> 
> > However, as we move forward with the latest design changes, we need to inform the calling fragment of both success and failure.
> 
> @Jignesh-dimagi I implemented a minimal version of this for now (to be used later). Moving ownership of failure handling to the fragment (and reducing the responsibility of the launch controller) will be a decent scope increase and I think that's worth a new ticket. Commit - [2353f1c](https://github.com/dimagi/commcare-android/commit/2353f1cc140d58f0a4b7820cd0b05669ccd48ef3).

I think a simple interface between the controller and the fragment could work here, but I will leave it up to you and accept your final call.

RELEASES.md
```diff
@@ -179,6 +179,8 @@ we would like to communicate to QA as part of the release testing
   - Launching via a job's **Start** button (opportunity intro screen) and right after an **app download/install** finishes both open directly behind the single progress dialog (no login/app-setup flash); backing out of the app home returns to the opportunities list, not the intro/download screen.
   - Launch several **not-yet-installed** apps back-to-back (open one, back out, open the next, repeat) and confirm none crash.
   - After repeated launch failures, on the **third consecutive failure** (i.e. after retrying twice) the dialog shows a single **OK** "couldn't open the app after several tries — check your connection and try again later" message instead of another Retry prompt.
+  - After launching an app from Connect and backing out to the opportunities list, press back once more: verify the app exits cleanly rather than re-opening the launched app's home.
```
Jignesh-dimagi: What does the phrase `verify the app exits cleanly` mean? I'm not sure I understand the exact context.

app/src/org/commcare/connect/ConnectAppLaunchController.kt
```diff
@@ -131,15 +131,15 @@ class ConnectAppLaunchController
...
-                        activity.startActivity(
-                            HomeScreenBaseActivity
-                                .buildHomeLaunchIntent(activity)
-                                .addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP),
-                        )
+
+                        val intent = HomeScreenBaseActivity.buildHomeLaunchIntent(activity)
+                        if (alreadyLoggedIn) {
+                            intent.addFlags(Intent.FLAG_ACTIVITY_REORDER_TO_FRONT)
+                        }
```
shubham1g5, line 148: does something goes wrong if we always add `FLAG_ACTIVITY_REORDER_TO_FRONT` irrespective of `alreadyLoggedIn`  ? Asking as `FLAG_ACTIVITY_REORDER_TO_FRONT` should be a no-op if there is no instance of activity in the task stack.

app/src/org/commcare/activities/connect/ConnectActivity.java
```diff
@@ -233,9 +234,23 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
-            super.onBackPressed();
+        if (!backButtonAndActionBarEnabled) {
+            return;
         }
+        if (appLaunchedFromConnect && isAtJobsList()) {
+            finishAffinity();
+            return;
+        }
+        super.onBackPressed();
+    }
```
shubham1g5: I am not yet convinced that this is the best solution here and trying to understand the issue more, putting a few questions to clarify my understanding around this below - 

- Seems like what we are trying to do here is - if user come on Jobs List after launching a home page and clicks back, exit the App ?

- I am curious why does that need to happen only when `appLaunchedFromConnect` is true ?

- Separately Are you able to specify the exact use case this is trying to solve for [from the doc ](https://docs.google.com/document/d/1jiVEbljnR8abPwJnKzqULTEqbAxU_PAt9sRWkPjTB9k/edit?tab=t.0#heading=h.45xudylq4ejp) and what's the behaviour without making these changes for those use cases ?

RELEASES.md
conroy-ricketts: By that I mean exact the entire mobile app, I'll tweak the wording here

app/src/org/commcare/connect/ConnectAppLaunchController.kt
conroy-ricketts: Yes, doing this will break edge case 5 from [the doc](https://docs.google.com/document/d/1jiVEbljnR8abPwJnKzqULTEqbAxU_PAt9sRWkPjTB9k/edit?usp=sharing).

You're right about the no-op, but this is to avoid reusing a previous app's stale Home page due to our app-seating logic (the Connect Job Tile will show the stale information from the previous app)

app/src/org/commcare/activities/connect/ConnectActivity.java
conroy-ricketts: > Seems like what we are trying to do here is - if user come on Jobs List after launching a home page and clicks back, exit the App ?

Yes that's correct. User comes on Jobs List → Launches an app from Jobs List → Clicks back to land on Jobs List again → Clicks back again → Exit the mobile app.

> I am curious why does that need to happen only when appLaunchedFromConnect is true ?

The `appLaunchedFromConnect` flag distinguishes between "the user launched an app from the Jobs List" and "the user is just browsing the Jobs List (i.e. did _not_ launch an app from it)"

> Separately Are you able to specify the exact use case this is trying to solve for from the doc and what's the behaviour without making these changes for those use cases ?

This solves for both edge case 2 and 3 from the doc. If we were to do a simple `super.onBackPressed()` instead,  then this activity would finsh and fall through to the base screen below it (e.g. `LoginActivity` or `DispatchActivity`), which brings us back to the undesired "looping" state mentioned in the doc

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
conroy-ricketts: This is to avoid the screen flashing on the user launching an app. Here's a quick video demonstrating this:

https://github.com/user-attachments/assets/e4ecac19-468a-4622-b7a9-2a0ed2e4141a

Notice how the Job List page flashes in the video when the app is launched.

If we simply use `NavHostFragment.findNavController(this).popBackStack()`, the screen flashes because the immediate pop happens while the `ConnectActivity` is still visible and in the foreground. So, popping the fragment at `onStop` guarantees that it happens while it's already hidden

conroy-ricketts:
> I think a simple interface between the controller and the fragment could work here, but I will leave it up to you and accept your final call.

@Jignesh-dimagi  I'll create a new ticket for this today and make a note of this in the new ticket

RELEASES.md
conroy-ricketts: https://github.com/dimagi/commcare-android/commit/14c488f70488f204e5aee97689290d1be4e51fd4

conroy-ricketts:
>> I think a simple interface between the controller and the fragment could work here, but I will leave it up to you and accept your final call.
>
> @Jignesh-dimagi  I'll create a new ticket for this today and make a note of this in the new ticket

@Jignesh-dimagi  Ticket created here: https://dimagi.atlassian.net/browse/CCCT-2600

cc @OrangeAndGreen @shubham1g5 for visibility

app/src/org/commcare/activities/connect/ConnectActivity.java
shubham1g5: > The appLaunchedFromConnect flag distinguishes between "the user launched an app from the Jobs List" and "the user is just browsing the Jobs List (i.e. did not launch an app from it)"

Yeah I understand what the flag does but not why the `finishAffinity` call is behind this flag i.e. why wouldn't we want to exit if there was no app launched from Opp List ?

Also one suggestion on `isAtJobsList` check. Think instead of `isAtJobsList` we want to check whether the user is at the start destination for a navigation path, so tomorrow if we make some other page the start in nav graph, this behaviour doesn't need to be changed.

app/src/org/commcare/connect/ConnectAppLaunchController.kt
shubham1g5: I see, think instead of trying to wire in `alreadyLoggedIn` in via callbacks, we can just calculate it directly here in `ConnectAppLaunchController.kt` ?

Also trying to understand the behaviour when `alreadyLoggedIn` is `false` here. In that case is the back behaviour different from when `alreadyLoggedIn` is `true` given we only apply this flag in one case and not other ?

conroy-ricketts: > I see, think instead of trying to wire in alreadyLoggedIn in via callbacks, we can just calculate it directly here in ConnectAppLaunchController.kt ?

The goal here is to know if the user was already logged in before the app was launched. If we wait until this point in the code, a new login would have already happened and it would be difficult to distinguish between the user was "already logged in" vs "just now logged in"

> Also trying to understand the behaviour when alreadyLoggedIn is false here. In that case is the back behaviour different from when alreadyLoggedIn is true given we only apply this flag in one case and not other 

The back behavior would be the same in either case

app/src/org/commcare/activities/connect/ConnectActivity.java
conroy-ricketts: > Yeah I understand what the flag does but not why the finishAffinity call is behind this flag i.e. why wouldn't we want to exit if there was no app launched from Opp List ?

Because if no app was launched, we'd want to return the user to the previous screen rather than completely exit the mobile app. 

So, for example, if a user logs into a CommCare app from the Login page directly, then opens the Opp List from the side drawer, and then presses back, then we should return the user back to the CommCare app. I think exiting the mobile app in this scenario would be a bit jarring.

Another good example: Say the user opens the mobile app on their device (assuming PersonalID is already registered) and the first page they see is the Login page. If they open the Opp List page and immediately hit back, we should return to the Login page rather than exit the mobile app IMO.

> Also one suggestion on isAtJobsList check. Think instead of isAtJobsList we want to check whether the user is at the start destination for a navigation path, so tomorrow if we make some other page the start in nav graph, this behaviour doesn't need to be changed.

Agreed!

conroy-ricketts: https://github.com/dimagi/commcare-android/commit/172d3f654e964f96899e36b21a0545369e4e390c

app/src/org/commcare/connect/ConnectAppLaunchController.kt
shubham1g5: > The goal here is to know if the user was already logged in before the app was launched. If we wait until this point in the code, a new login would have already happened and it would be difficult to distinguish between the user was "already logged in" vs "just now logged in"

yeah, I was thinking of calculating it in one of the `launch` methods itself, but keeping this as optional feedback. 

One minor request to close this thread out -  can we add a  small code comment here on why we are adding the intent flag only when `alreadyLoggedIn` is set.

app/src/org/commcare/activities/connect/ConnectActivity.java
shubham1g5: I think whatever back behaviour we keep here needs to be consistent across flows, It sounds more Jarring to me if sometimes back press from Opp List results in exiting the app while sometime not just based on what screen they launched from Opp List and not the history of user navigation. 


Think a mental model we can adapt here is to treat the screens launched on clicking a navigation drawer menu item as Top level screens pressing back from which exits the app i.e. treating each of these screens as the first in the app task stack and then any navigation from these screens should place the launched activity on top of it. _**The drawer is more like changing tabs on the same page**_ and similar behaviour is demonstrated by Gmail/ Gphotos as well which seems like a standard practice. 


Listing out various scenarios based on that mental model - 

1. **Top Level Navigation From drawer:** 

User navigates as `Opp List ->   Messaging(drawer) -> Work History (drawer)` , on clicking back from "Work history" we exit the app instead of tracing back the flow given all items were accessed from drawer. 


2. **Cross Feature flows but not from drawer:** 

User navigates as `Opp List -> Resume ->  App Home" , on clicking abck from "App Home" we should go back to "Opp List" to trace back user navigation history correctly. 


Also FYI @OrangeAndGreen @Jignesh-dimagi

conroy-ricketts: I like that idea, and I think this may simplify all the changes I made recently to satisfy all of the edge cases simultaneously. I'd like to hear thoughts from the team as well (cc @OrangeAndGreen and @Jignesh-dimagi) before making the change and potentially prolonging the code review rounds on this PR even further

OrangeAndGreen: > similar behaviour is demonstrated by Gmail/ Gphotos

I'm not necessarily against this idea, but noting that I'm not observing that as standard behavior when I experiment across other apps. Seems like a more common behavior is for apps to back-nav to a common "home tab" before one more back-nav exits the app.

Example with GMail:
* App opens to Email tab
* User navigates to Messaging tab, then Meet tab
* Back-nav once goes to Email tab (the "home" tab) instead of back to Messaging or exiting
* Back-nav again exits the app

I see the same behavior with the Home, Keypad, and Voicemail tabs in my Phone app, where back-nav always goes back to the Home tab before exiting. Slack and WhatsApp also behave this way for me. Of course, the problem we have with an approach like these apps is that we don't have one clear "Home" tab like they do.

In the Google News app I see more back-nav among the tabs, i.e.:
* Start on Home tab
* Navigate to Following tab, then to Newsstand tab
* Back-nav once returns to Following tab (the previous tab)
* Back-nav second time returns to Home tab
* Back-nav third time exits the app

It seems the trick that this app uses is to remove any previous instance of a tab when it gets added back to the nav stack again. So if I jump back and forth between two tabs other than the Home tab, back-nav doesn't take me back through the multiple visits to each tab. But again, even in the Google News case there's a Home tab that back-nav always goes to before the final back-nav exits the app.

So it seems we should either designate a Home tab that back-nav always uses as the final stop before exiting the app (I would say the Login page before the Connect UX redesign changes), or we use the approach Shubham suggested here.

app/src/org/commcare/connect/ConnectAppLaunchController.kt
conroy-ricketts: Added a short comment here dda8580706e32d105e7bc18ab5a32dd2221d3ed0

app/src/org/commcare/activities/connect/ConnectActivity.java
shubham1g5: Ahh thanks Dave for correcting me here, definitely seeing the behaviour you outlined and was swayed away by a wrong reference that I didn't cross check.

shubham1g5 (approved):

OrangeAndGreen (approved):

Jignesh-dimagi (approved):

PR #3767, opened 2026-06-19
Spec for tasks Persistance and Implementing Messaging Redirect for OCS task

shubham1g5:
https://dimagi.atlassian.net/browse/CCCT-2468


docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
```diff
@@ -0,0 +1,219 @@
...
+
+## Data Layer
+
+### `ParsedConnectTask` (delete)
+`ParsedConnectTask` is removed entirely. All other Connect DB record models (`ConnectJobDeliveryRecord`, `ConnectJobPaymentRecord`, etc.) parse directly from JSON via a `fromJson(JSONObject, ConnectJobRecord)` static factory method — no intermediate transient model. `ConnectTaskRecord` follows the same pattern.
+
+### New `ConnectTaskRecord` (DB-persisted, Java)
+New table `connect_tasks` in the Connect encrypted DB, following the `@Table` / `@Persisting` / `@MetaField` pattern used by `ConnectJobDeliveryRecord`.
+
+Fields:
```
shubham1g5, line 21: @pxwxnvermx are you able to confirm that mobile gets these fields from server in response to `/api/opportunity/{id}/delivery_progress` when working on the corresponding web spec.

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
```diff
@@ -0,0 +1,217 @@
...
+- `isTaskPending(context)` replaces the preference read with `ConnectJobUtils.getPendingTask(context, jobUUID) != null && status == STATUS_DELIVERING`.
+- `shouldShowTasksCompletedMessage(context)` replaces the `KEY_RELEARN_TASKS_COMPLETED_TIME_MS` preference with a DB query:
+  1. Return `false` if any task is still assigned (`getPendingTask(context, jobUUID)` non-null).
+  2. Fetch `getMostRecentlyCompletedTask(context, jobUUID)`.
+  3. Return `true` if that task's `dateModified` is within `RELEARN_TASKS_COMPLETED_MESSAGE_WINDOW_MS` of now.
+
+**`ConnectJobRecord.syncRelearnTasksPrefs` is removed** — task persistence is now handled entirely by `applyToJob()` calling `storeTasks()`. The `applyToJob()` call to `job.syncRelearnTasksPrefs(parsedTasks)` is deleted. No `ParsedConnectTask` list exists anywhere in the new flow.
+
+**`ConnectJobPreferences`** — remove `KEY_RELEARN_TASK_PENDING` and `KEY_RELEARN_TASKS_COMPLETED_TIME_MS` along with their accessor methods. `TIMESTAMP_NOT_SET`, `getLastTaskUpdateTimeMs`, and the payment-confirmation preference methods are unaffected.
+
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3767/commits/279fb77d4855ab18be221346d8afb2fd621ed8bb

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
```diff
@@ -0,0 +1,217 @@
...
+In `DeliveryAppProgressResponseModel.applyToJob()`, only record the timestamp when `storeTasks()` reports an actual change:
+```kotlin
+if (hasTasks) {
+    val tasksChanged = ConnectJobUtils.storeTasks(context, taskRecords, job.jobUUID)
+    if (tasksChanged) {
+        ConnectJobUtils.getJobPreferences(job.jobUUID).setLastTaskUpdateTimeMs(System.currentTimeMillis())
+    }
+}
+```
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3767/commits/279fb77d4855ab18be221346d8afb2fd621ed8bb

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
```diff
@@ -0,0 +1,217 @@
...
+| `slug` | String | Channel ID or task-specific ID; unique |
+| `type` | String | `"messaging"` or `"relearn"` |
+| `dueDate` | Date | Nullable |
+| `dateCreated` | Date | default to now |
+| `dateModified` | Date | default to now |
+
+### `ConnectJobUtils`
+- `storeTasks(context, List<ConnectTaskRecord>, jobUUID): Boolean` — fetches the existing tasks for the job from the DB, diffs them against the incoming list (comparing `taskId`, `taskType`, `status`, and `slug` per task). If anything changed (new task, removed task, or field change), replaces the stored list and returns `true`. Returns `false` if the lists are identical.
+- `getPendingTask(context, jobUUID): ConnectTaskRecord?` — returns the first task with `status = "assigned"` for the job, regardless of type. Callers that need a specific type (e.g. the messaging button) inspect `task.getType()` after the call.
+- `getMostRecentlyCompletedTask(context, jobUUID): ConnectTaskRecord?` — returns the task with `status = "completed"` having the latest `dateModified`, regardless of type.
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3767/commits/279fb77d4855ab18be221346d8afb2fd621ed8bb

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
```diff
@@ -0,0 +1,217 @@
...
+| `description` | String | optional |
+| `status` | String | `"assigned"` or `"completed"` |
+| `slug` | String | Channel ID or task-specific ID; unique |
+| `type` | String | `"messaging"` or `"relearn"` |
+| `dueDate` | Date | Nullable |
+| `dateCreated` | Date | default to now |
+| `dateModified` | Date | default to now |
+
+### `ConnectJobUtils`
+- `storeTasks(context, List<ConnectTaskRecord>, jobUUID): Boolean` — fetches the existing tasks for the job from the DB, diffs them against the incoming list (comparing `taskId`, `taskType`, `status`, and `slug` per task). If anything changed (new task, removed task, or field change), replaces the stored list and returns `true`. Returns `false` if the lists are identical.
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3767/commits/279fb77d4855ab18be221346d8afb2fd621ed8bb

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3767/commits/978527f8ecf8b9d17d7c224045f5f0bf9ced11aa

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
```diff
@@ -0,0 +1,220 @@
...
+
+The check only fires when no earlier health action already kicked off a sync. It short-circuits immediately for jobs not in `STATUS_DELIVERING`, and when no task change has been recorded yet.
+
+## Task Completion Push Notification
+
+### Payload
+
+```json
+{
+  "notifications": [
```
shubham1g5, line 206: @pxwxnvermx FYI that here is the notification payload for notification server should implement to notify user of a task completion.

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
pxwxnvermx: These fields will sent from the Delivery Progress API for Tasks. This currently does not include the slug. Also we don't have a field for dateModified.
I will add the slug to the technical spec doc for OCS conversation Tasks in web.
https://github.com/dimagi/commcare-connect/pull/1312/

| Field | Type |
  |---|---|
  | `assigned_task_id` | UUID (string) |
  | `task_name` | string |
  | `task_description` | string |
  | `status` | enum string |
  | `due_date` | date |
  | `date_created` | datetime |
  | `task_type` | relearn or ocs |
  | `connect_channel_id` | uuid |

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
pxwxnvermx: Thanks. I will add this to the spec.

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: sounds good, ok to skip `dateModified` in the payload. 

Also I defined slug for the `connect_channel_id` field you have above, I would be happy to call it whatever we call on web and keep is consistent be it `connect_channel_id` or `slug`. My assumption was that on web, we in general prefer to define a slug in comparison to the specific id fields.

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
```diff
@@ -0,0 +1,220 @@
...
+`ParsedConnectTask` is removed entirely. All other Connect DB record models (`ConnectJobDeliveryRecord`, `ConnectJobPaymentRecord`, etc.) parse directly from JSON via a `fromJson(JSONObject, ConnectJobRecord)` static factory method — no intermediate transient model. `ConnectTaskRecord` follows the same pattern.
+
+### New `ConnectTaskRecord` (DB-persisted, Java)
+New table `connect_tasks` in the Connect encrypted DB, following the `@Table` / `@Persisting` / `@MetaField` pattern used by `ConnectJobDeliveryRecord`.
+
+Fields:
+
+| Field | Type | Notes |
+|---|---|---|
+| `opportunity_id` | String | Binds the task to Opp |
```
Jignesh-dimagi: Can we name it opportunity_uuid?

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
```diff
@@ -0,0 +1,220 @@
...
+| `taskId` | String | Unique task identifier from server |
+| `name` | String | Required |
+| `description` | String | Optional |
+| `status` | String | `"assigned"` or `"completed"` |
+| `slug` | String | Channel ID or task-specific ID; unique |
```
Jignesh-dimagi: I assume that each task with a type of `messaging` will have its own channel ID. If that is the case:
- The application currently fetches the messaging and channel list using `/messaging/retrieve_notifications/`. Will this API also return the new OCS channel and its related messages?
- Once the OCS task is completed, will the `/messaging/retrieve_notifications/` API remove that channel so the user can no longer interact with it?

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
```diff
@@ -0,0 +1,220 @@
...
+}
+```
+
+### `ConnectMessagingActivity` Refactor
+
+**Extract shared method:**
+```java
+private void navigateToChannel(String channelId) {
+    handleChannelForValidity(channelId);
+}
```
Jignesh-dimagi, line 134: Can we directly try to use [this](https://github.com/dimagi/commcare-android/blob/7597313af80e0ef193a632a31f78d38f821f4670/app/src/org/commcare/activities/connect/ConnectMessagingActivity.java#L128) code in `ConnectMessagingActivity`?

Jignesh-dimagi (changes requested):

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
```diff
@@ -0,0 +1,220 @@
+# Conversation Tasks and CTA to open Messaging from the Home Screen
+
+
+## Overview
+https://dimagi.atlassian.net/browse/CCCT-2468
+
+The Spec tries to ahieve following outcomes - 
+1. Add an "Open Conversation" button to the Connect job tile on the home screen. The button is visible only when a `messaging`-type task is assigned and pending for the active job (in `STATUS_DELIVERING`). Tapping it launches `ConnectMessagingActivity` and navigates directly to the task's channel.
```
OrangeAndGreen, line 8: Curious if there's a strong reason why we're restricting these to jobs with STATUS_DELIVERING? I know right now we only plan to have tasks assigned once a user is in delivery state, but if we wanted to assign tasks during learning is there any reason the app couldn't support it?

I'm trying to think of a situation where a task could be assigned that's intended for delivery but the user isn't in delivery state yet, but that doesn't seem like something that would happen. On the other hand, the idea that we might assign an additional task or start a conversation with a user who is struggling during learning seems like a reasonable future addition that the mobile code could be ready to support.

OrangeAndGreen, line 8: What's the intended behavior related to tasks/conversations once an opportunity has ended or if the user has been suspended? Do we still show the "Open Conversation" button?

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
```diff
@@ -0,0 +1,220 @@
...
+2. Provide DB persistence for tasks and remodel existing code to use the DB 
+3. Trigger Sync on home screen after a change in task given the app usually needs to refresh the case data after a change in task to implement corresponding app side logic.
```
OrangeAndGreen, line 10: For my learning and probably beneficial to other team members as well: I'm curious to understand better why a change in tasks could require an app sync in order to adjust app side logic. Probably out of scope to address in this document, but this could be a good topic for a mobile hour or coworking session.

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
```diff
@@ -0,0 +1,220 @@
...
+- `id`: `open_conversation`
+- `text`: `@string/connect_open_conversation` ("Open Conversation")
+- `drawableStart`: `@drawable/ic_connect_message_large` (existing icon)
+- `visibility`: `gone` by default
+- Style: matches `acb_resume` (rounded lavender background, `connect_blue_color` text, no elevation/state animator, `textAllCaps="false"`)
+- Constrained below `@id/cv_relearn_tasks_pending`, start-aligned to parent
+
+New string resource in `strings.xml`: `personalid_open_conversation`.
+
+### `StandardHomeActivityUIController`
```
OrangeAndGreen, line 98: Wondering whether it also makes sense to show this on the Delivery Progress page?

I'm thinking about a user that has pending tasks preventing them from continuing delivery, but is in a state where they don't currently have the delivery app installed (i.e. just got a new device or reinstalled CommCare).

We'll be forcing the user to download and login to the delivery app, just so we can show them a message saying they can't continue delivery until they go complete the messaging conversation (which doesn't require the delivery app).

If they could access the task directly from the Delivery Progress page, they could avoid downloading the delivery app until they're actually ready to use it.

OrangeAndGreen (dismissed):
Left a few things to consider but all in all this looks good to me

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
conroy-ricketts: Super small thing, but out of curiosity, why use snake case here when the rest of the fields are camel case? Can we change this casing to match?

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
```diff
@@ -0,0 +1,220 @@
...
+| `description` | String | Optional |
+| `status` | String | `"assigned"` or `"completed"` |
+| `slug` | String | Channel ID or task-specific ID; unique |
+| `type` | String | `"messaging"` or `"relearn"` |
+| `dueDate` | Date | Nullable |
+| `dateCreated` | Date | Default to now |
+| `dateModified` | Date | Default to now |
+
+### `ConnectJobUtils`
+- `storeTasks(context, List<ConnectTaskRecord>, jobUUID): Boolean` — fetches the existing tasks for the job from the DB, diffs them against the incoming list (comparing `taskId`, `type`, `status`, and `slug` per task). If anything changed (new task, removed task, or field change), replaces the stored list and returns `true`. Returns `false` if the lists are identical.
```
conroy-ricketts: We'd only need to check if the `status` changed right?

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
```diff
@@ -0,0 +1,220 @@
...
+- `hasPendingTask(context, jobUUID): Boolean` — returns `true` if any task for the job has `status = "assigned"`, regardless of type. Used for `isTaskPending()` and `shouldShowTasksCompletedMessage()` step 1 in `ConnectJobRecord`. **Fallback:** if the DB has no task rows for the job, read `KEY_RELEARN_TASK_PENDING` from `ConnectJobPreferences`; if that pref is `true`, return `true` and immediately clear the pref so the fallback fires only once.
```
conroy-ricketts: `ConnectJobRecord` has a lot of the same logic today - is that going to be refactored to reuse this?

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
```diff
@@ -0,0 +1,220 @@
...
+- `hasPendingTaskOfType(context, jobUUID, type): Boolean` — returns `true` if a task of the given `type` has `status = "assigned"`. Used for the messaging button visibility check. Avoids the problem of a relearn task hiding the messaging CTA when both are pending. No preference fallback — messaging tasks have no prior pref equivalent.
+- `getPendingTaskOfType(context, jobUUID, type): ConnectTaskRecord?` — returns the pending task of the given `type`, or `null`. Used when the actual record is needed (e.g. messaging button needs the `slug` for navigation). `hasPendingTaskOfType` may delegate to this internally. No preference fallback.
```
conroy-ricketts: What if the user has multiple tasks of the same `type` at the same time?

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
```diff
@@ -0,0 +1,220 @@
...
+- `getMostRecentlyCompletedTask(context, jobUUID): ConnectTaskRecord?` — returns the task with `status = "completed"` having the latest `dateModified`, regardless of type. **Fallback:** if the DB has no task rows for the job, read `KEY_RELEARN_TASKS_COMPLETED_TIME_MS` from `ConnectJobPreferences`; if set, synthesise a minimal `ConnectTaskRecord` with `dateModified` equal to that timestamp, return it, and clear the pref.
```
conroy-ricketts: The fallback would always be used given that Server is not sending `dateModified` for OCS tasks right?

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
```diff
@@ -0,0 +1,220 @@
...
+
+### `DeliveryAppProgressResponseModel` / Parser
+- Replace `parsedTasks: List<ParsedConnectTask>` with `tasks: List<ConnectTaskRecord>` (empty list when `assigned_tasks` is absent or empty). Remove the `hasTasks: Boolean` flag.
```
conroy-ricketts: Where does `hasTasks` exist today? Seems that `DeliveryAppProgressResponseModel` does not have it

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
conroy-ricketts, line 1: In this spec can we specify which methods are new (to be added in implementation) vs existing (to be modified)

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
```diff
@@ -0,0 +1,220 @@
...
+- Parser calls `ConnectTaskRecord.fromJson(obj, job)` for each item in `assigned_tasks`, matching how deliveries and payments are parsed.
+- `applyToJob()` always calls `ConnectJobUtils.storeTasks()` unconditionally, even when the list is empty — this ensures stale records are deleted when the server stops returning tasks. The `tasksChanged` return value gates the timestamp update only.
+
+### DB Migration: V.26 → V.27
+- `DatabaseConnectOpenHelper`: bump `CONNECT_DB_VERSION` to 27; register `ConnectTaskRecord` table in `onCreate`.
+- `ConnectDatabaseUpgrader`: add upgrade case for version 26 that creates the `connect_tasks` table via `TableBuilder`.
+
+### Relearn task migration to DB
+
+The relearn task state moves fully from `ConnectJobPreferences` to the DB, so the new `ConnectTaskRecord` is the single source of truth for all task types.
```
conroy-ricketts: > The relearn task state moves fully from ConnectJobPreferences to the DB

This conflicts with what was talked about earlier in the spec. Is this happening at the same time? In a later phase? This is not explicitly clear from reading the spec and should be stated

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
```diff
@@ -0,0 +1,220 @@
...
+
+- `isTaskPending(context)` replaces the preference read with `ConnectJobUtils.hasPendingTask(context, jobUUID) && status == STATUS_DELIVERING`.
+- `shouldShowTasksCompletedMessage(context)` replaces the `KEY_RELEARN_TASKS_COMPLETED_TIME_MS` preference with a DB query:
+  1. Return `false` if any task is still assigned (`hasPendingTask(context, jobUUID)` is true).
+  2. Fetch `getMostRecentlyCompletedTask(context, jobUUID)`.
+  3. Return `true` if that task's `dateModified` is within `RELEARN_TASKS_COMPLETED_MESSAGE_WINDOW_MS` of now.
+
+**`ConnectJobRecord.syncRelearnTasksPrefs` is removed** — task persistence is now handled entirely by `applyToJob()` calling `storeTasks()`. The `applyToJob()` call to `job.syncRelearnTasksPrefs(parsedTasks)` is deleted. No `ParsedConnectTask` list exists anywhere in the new flow.
+
+**`ConnectJobPreferences`** — do **not** remove `KEY_RELEARN_TASK_PENDING`, `KEY_RELEARN_TASKS_COMPLETED_TIME_MS`, or their accessor methods yet. Mark them `@Deprecated` with a note: *"Remove after 2.63 release.
```
conroy-ricketts: Is this meant to say "2.65 release"?

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
```diff
@@ -0,0 +1,220 @@
...
+
+**`ConnectJobPreferences`** — do **not** remove `KEY_RELEARN_TASK_PENDING`, `KEY_RELEARN_TASKS_COMPLETED_TIME_MS`, or their accessor methods yet. Mark them `@Deprecated` with a note: *"Remove after 2.63 release. 
+
+**Call sites** — all callers of the now-renamed `isTaskPending()` and `shouldShowTasksCompletedMessage()` (in `StandardHomeActivityUIController`, `ConnectJobRecord.getCardMessageText`, etc.) must be updated to pass a `Context`.
+
+## UI Layer
+
+### `view_progress_job_card.xml`
+Add a new `AppCompatButton`:
+- `id`: `open_conversation`
```
conroy-ricketts: This convention does not follow the IDs of the other elements currently in the layout. Suggest calling it `acb_open_conversation` (i.e. for the element type) to match

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
```diff
@@ -0,0 +1,220 @@
...
+- `text`: `@string/connect_open_conversation` ("Open Conversation")
```
conroy-ricketts: Could be more specific, how about `@string/connect_open_conversation_button_text`?

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
```diff
@@ -0,0 +1,220 @@
...
+- `drawableStart`: `@drawable/ic_connect_message_large` (existing icon)
+- `visibility`: `gone` by default
+- Style: matches `acb_resume` (rounded lavender background, `connect_blue_color` text, no elevation/state animator, `textAllCaps="false"`)
+- Constrained below `@id/cv_relearn_tasks_pending`, start-aligned to parent
+
+New string resource in `strings.xml`: `personalid_open_conversation`.
```
conroy-ricketts: Flagging the resource name here does not match what was said above

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
```diff
@@ -0,0 +1,220 @@
...
+
+### `StandardHomeActivityUIController`
+- Add field: `AppCompatButton btnOpenConversation`, found in `setupConnectJobTile()` via `viewJobCard.findViewById(R.id.acb_open_conversation)`.
```
conroy-ricketts: Suggest calling it `AppCompatButton acbOpenConversation` to match the earlier suggestion

conroy-ricketts: Also, think the name used in `viewJobCard.findViewById(R.id.acb_open_conversation)` does not match what was said earlier

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
```diff
@@ -0,0 +1,220 @@
...
+  - Button is hidden when no pending messaging task exists or the task's `slug` is empty.
+- Button is hidden in all non-`STATUS_DELIVERING` states.
+
+## Navigation
+
+### `ConnectNavHelper`
+Add:
+```kotlin
+fun goToMessagingWithChannel(context: Context, channelId: String) {
+    val i = Intent(context, ConnectMessagingActivity::class.java)
```
conroy-ricketts: Is it standard practice to use `i` instead of `intent` here?

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
```diff
@@ -0,0 +1,220 @@
...
+- Add field: `AppCompatButton btnOpenConversation`, found in `setupConnectJobTile()` via `viewJobCard.findViewById(R.id.acb_open_conversation)`.
+- In `syncJobCardVisibility(job)`:
+  - If `job.getStatus() == STATUS_DELIVERING`: call `ConnectJobUtils.hasPendingTaskOfType(activity, job.getJobUUID(), "messaging")` to determine visibility.
+  - When visible and the button is tapped: call `ConnectJobUtils.getPendingTaskOfType(activity, job.getJobUUID(), "messaging")` to retrieve the task and pass its `slug` to `ConnectNavHelper.unlockAndGoToMessagingWithChannel(activity, task.getSlug(), listener)`.
+  - Button is hidden when no pending messaging task exists or the task's `slug` is empty.
+- Button is hidden in all non-`STATUS_DELIVERING` states.
+
+## Navigation
+
+### `ConnectNavHelper`
```
conroy-ricketts, line 108: It seems that this section (and the ones after) goes heavy into implementation details. Is this necessary for a spec like this? This would be reviewed later in code review anyways, and these sections could be simplified for the purposes of this initial spec

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: Agree that seems like a reasonable addition but I deliberately wanted to keep it out of scope given the Connect Redesign screens keep it out of scope for Learn Apps and don't really have a place to show tasks at the moment for Learn App, so the idea here is to not to increase scope of re-design and take it as a separate feature when we need to support it in future.

shubham1g5: >  Do we still show the "Open Conversation" button ? 

Yes, don't think there is any harm in doing that if there is a pending task assigned to the user. I would imagine server side to take control here and close any tasks for the user if it's a requirement but definitely not under consideration at the moment.

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: [This wiki should talk in detail about the app side setup for re-learning in general](https://dimagi.atlassian.net/wiki/spaces/connect/pages/3698622469/Guidelines+for+Relearning+in+Connect) , but the crux is on server side implementation for a generated task ->  a generated task updates user case property for the user it's assigned to which is only synced as part of CC App sync/restore. This user case property is used by Apps to hide the delivery forms so that users can't do delivery unless they complete the task upon which the user case property is toggled again and a sync is required on app for it to update locally on phone.

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: > Will this API also return the new OCS channel and its related messages?

No, those will be synced separately as today. This field just links the task with a channel. 

> Once the OCS task is completed, will the /messaging/retrieve_notifications/ API remove that channel so the user can no longer interact with it?

@pxwxnvermx do you know the answer to this ?

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: Ahh didn't realise there is such convention, will correct.

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: Think the only reason we are not adding it to Delivery Progress page at the moment is given the planned re-design for delivery progress page, the changes in this spec are minimum changes we can make to support this feature until the planned redesign which will introduce tasks as a much better CTA on delivery progress page.

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: think this should only be reviewed as pseudo code and think was important to highlight the overall code navigation for the messaging task and where exactly the unlock happens as part of the process along with the planned refactor to existing messaging code. Think it's good for specs to include minor code structures like this when there can be ambiguity but it's sort of a thin line that I do not want to control hard and leave to authors of specs.

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: I don't like that approach given it will add additional overhead in the code to add a dummy notification action. That's a strong anti pattern we should try to avoid as it makes code un-necessary complex when a much simpler alternative exists that still reuses majority of the redirect code.

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: think It's quite common in our code but not sure about being standard practice, I would be happy to rename it to `intent`.

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
Jignesh-dimagi: I am just suggesting that we abstract this code and use it everywhere we need to open the message fragment using a channel ID. This abstraction will automatically handle Firebase data logging, unlocking, and navigating to the message fragment.

shubham1g5: Is that abstraction different from what's said below around `handleRedirectIfAny` ? If so, are you able to outline the method/class structure you are thinking ?

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: https://github.com/dimagi/commcare-android/pull/3767/changes/8c87dbd10440cc8108be03c0a5e8860d935d4677

shubham1g5: https://github.com/dimagi/commcare-android/pull/3767/changes/8c87dbd10440cc8108be03c0a5e8860d935d4677

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: I missed some cleanup here from an earlier approach I was implementing in the spec - https://github.com/dimagi/commcare-android/pull/3767/changes/8c87dbd10440cc8108be03c0a5e8860d935d4677

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: https://github.com/dimagi/commcare-android/pull/3767/changes/8c87dbd10440cc8108be03c0a5e8860d935d4677

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: https://github.com/dimagi/commcare-android/pull/3767/changes/8c87dbd10440cc8108be03c0a5e8860d935d4677

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: https://github.com/dimagi/commcare-android/pull/3767/changes/8c87dbd10440cc8108be03c0a5e8860d935d4677

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: https://github.com/dimagi/commcare-android/pull/3767/changes/8c87dbd10440cc8108be03c0a5e8860d935d4677

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: https://github.com/dimagi/commcare-android/pull/3767/changes/8c87dbd10440cc8108be03c0a5e8860d935d4677

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: Think that's not in scope today, we will just return the first such task found from here for time being and direct user to the corresponding CTA, Once that task has been completed, user can move to the next one in the list.

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: @pxwxnvermx re-checking if web team has a preference here between `connect_channel_id` and `slug` ?

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: https://github.com/dimagi/commcare-android/pull/3767/changes/056cb54f95f882cfbe7b9f1d94edf633b6dbe481 , It's 2.64 i.e. the release this spec  implementation is meant to go out with.

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: Let me know if this edit doesn't make it clearer - https://github.com/dimagi/commcare-android/pull/3767/changes/056cb54f95f882cfbe7b9f1d94edf633b6dbe481

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: added small clarification in https://github.com/dimagi/commcare-android/pull/3767/changes/056cb54f95f882cfbe7b9f1d94edf633b6dbe481

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: Actually think we need to compare all essential fields that can affect a task uniqueness. Mostly, I am trying to acccount for fact if any of the field changes happen on a task as a result of edit on web, we should update the local task to reflect the same. Think the current mechanism here to compare the `date_modified` is another way to do this, which is probably better here. I am going to re-check with web that if web is reliably sending that info today.

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: https://github.com/dimagi/commcare-android/pull/3767/changes/cfffa8c7bd5d1c387c1def92c1af9ab605e00bf5

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: clarified in https://github.com/dimagi/commcare-android/pull/3767/changes/cfffa8c7bd5d1c387c1def92c1af9ab605e00bf5

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: think mobile can still populate `dateModified` locally based on the time we write/update a task in the DB.

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: looks like web doesn't send that info today, I think this will have to compare all task table fields (name,description,status,slug,type,due_date) for a given `task_id`  to decide whether to update that record in the DB.

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
OrangeAndGreen: Okay, there may be some stacking of blue info messages if we support both, or we'll need to decide which takes priority. But I think we're looking at the same real estate for the "opp ended" and "tasks pending" messages that may be worth addressing in the spec.

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
OrangeAndGreen: Ah right, I'm getting mixed up between the pre- and post- redesign items again. Sounds good here.

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: Currently the idea is for the tasks to take the priority (that's already the case in current implementation) irrespective of the Opp status i.e. if a task is present, show the task UI. Again this is fairly short term as re-design will get rid of this card completely.

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
pxwxnvermx: Both of those fields are optionally delivered for Tasks.
Slug are defined for Re-Learn Connect Task.
connect_channel_id is defined for OCS Connect Tasks.

I can combine those fields if needed but keeping them seperate will avoid confusion.

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
pxwxnvermx: The notification API usually removes all notification that a user has acknowledged, so once the user interacts with the notification it will not be sent through the API.

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: Ahh I see, what does "Slug" defines for  Re-Learn Connect Task ?

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: Does it make sense https://github.com/dimagi/commcare-android/pull/3767/changes/dbc4cb3b59d1f6d407719b25a245eba1a517f0ff ?

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
shubham1g5: @Jignesh-dimagi Also the channel remains active even after the task completion as [mentioned in this thread ](https://github.com/dimagi/commcare-connect/pull/1312#discussion_r3449748328)

Jignesh-dimagi:
Looks good to me, no more comments from my side. @shubham1g5

Jignesh-dimagi (dismissed):

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
pxwxnvermx: The "slug" defines the HQ Form question id for "Re-learn" connect tasks.

shubham1g5: interesting, thanks for the info! I will update the spec to use `connect_channel_id` on mobile as well.

shubham1g5: https://github.com/dimagi/commcare-android/pull/3767/changes/9a176411c0a540093de2bd06bfaf88cdb3789651

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
conroy-ricketts: Yup that makes sense to me! Thanks for the clarification

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
conroy-ricketts: Ah that helps a lot, thanks!

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
conroy-ricketts: That works!

docs/superpowers/specs/2026-06-18-connect-messaging-task-button-design.md
conroy-ricketts: Noted! That sounds good to me

conroy-ricketts (approved):
LGTM 🚀

OrangeAndGreen (approved):

PR #3768, opened 2026-06-19
[SAAS-19849] Update audio recording widget UX 

avazirna:
## Product Description

Three recorder UX changes from SAAS-19849:

- **Direct start** — the recorder begins recording immediately when opened (no intermediate "tap to record" step).
- **Direct stop & save** — stopping a recording finalizes and saves it straight away, removing the in-dialog playback/confirm step. Playback now happens in the audio widget itself.
- - **Direct delete** — deleting a recording now takes the user to the initial UI, instead of the intermediate "tap to record" step.

<table>
<tr>
<th>
Before
</th>
<th>
After
</th>
</tr>
<tr>
<td>
<video width="100px" src="https://github.com/user-attachments/assets/23c6ef18-0837-401e-a676-a797e6b00edb"></video>
</td>
<td>
<video width="100px" src="https://github.com/user-attachments/assets/98891e3d-82ac-4479-92b9-9544bc1624b6"></video>
</td>
</tr>
<tr>
<td>
<video width="100px" src="https://github.com/user-attachments/assets/005ca9cf-b71e-4cd9-8bc1-bbfa217413cd"></video>
</td>
<td>

<video width="100px" src="https://github.com/user-attachments/assets/8d6a9a50-1388-47ac-8f9b-de0d641ed87c"></video>
</td>
</tr>
</table>

## Technical Summary

Implements [SAAS-19849](https://dimagi.atlassian.net/browse/SAAS-19849). All changes are in `RecordingFragment.java`:

- **Lifecycle split** — extract resource caching into `initMediaResources()` and move the reload-vs-start decision into `onViewCreated`.
- **Auto-start** — start recording on open when there's no existing recording, gated on `savedInstanceState == null` so a configuration-change recreation doesn't re-trigger a recording.
- **Save on stop** — `stopRecording(boolean shouldSave)`; stopping a non-pausable recording saves and dismisses directly. The flag lets the Save-from-paused path stop without re-triggering save.
- **Dead-code removal** — the post-stop playback transition was the only entry into the fragment's own playback path, so `playAudio`/`pause`/`resume`/`reset`, the `MediaPlayer` field, and the unused `MediaPlayer`/`Uri` imports are now removed (playback lives in `CommCareAudioWidget`).

### Safety story

- These changes only impact the recorder dialog, the scope is to simplify the user experience as described above, so no data-model/storage changes. 
- `AudioRecordingService.stopRecording()` is synchronous (`MediaRecorder.stop()` blocks until the file is finalized), so save-on-stop works with a fully-written file.
- These changes were successfully tested locally and the resulting files verified on HQ;
- This feature locks the screen orientation before recording starts. If the screen is in `portrait` mode, the recording proceeds in `portrait`; same goes for `landscape`;

## Labels and Review

- [ ] Manual QA / RELEASES.md updated as needed
- [ ] Major-change communication via RELEASES.md if needed
- [ ] Risk label set correctly
- [ ] Reviewers appropriate for risk level

[SAAS-19849]: https://dimagi.atlassian.net/browse/SAAS-19849?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


conroy-ricketts (commented):
Out of curiosity, what does the UI look like after a recording is deleted?

.github/workflows/commcare-android-pr-workflow.yml
```diff
@@ -93,6 +93,14 @@ jobs:
...
         run: ./gradlew testCommcareDebug
         working-directory: commcare-android
+      - name: Upload unit test report
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v6
+        with:
+          name: unit-test-report
+          path: commcare-android/app/build/reports/tests/testCommcareDebugUnitTest
+          if-no-files-found: warn
+          retention-days: 14
```
conroy-ricketts, line 103: Was this added by mistake?

avazirna: The build failed a couple of times because of one of the unit tests, `UpdateActivityTest.invalidUpdateTest()`, and I added this to get more info around the error. But after that addition, the builds were successful. I think this is a good addition but not in the scope of this PR, so the plan is to PR this separately.

shubham1g5: think a 3 day retention should be sufficient here, one can always re-trigger the build to get this info again.

shubham1g5 (dismissed):
looks good, would be nice to add safety story around phone rotation on this change.

avazirna:
> looks good, would be nice to add safety story around phone rotation on this change.

We disable screen rotation once the recording starts.

avazirna:
> Out of curiosity, what does the UI look like after a recording is deleted?

It was going back to the re-recording state. I've simplified this as well, i will add a video recording to the PR description.

shubham1g5 (dismissed):
Don't see any concerns but might be worth looking over Code Rabbit comment and refining the notes in `RELEASES.md`

.github/workflows/commcare-android-pr-workflow.yml
conroy-ricketts: Confirming if this will still be scoped out to a separate PR?

conroy-ricketts (commented):
I don't have anything additional to add, just a question

avazirna:
> Confirming if this will still be scoped out to a separate PR?

Yes

avazirna:
> Don't see any concerns but might be worth looking over Code Rabbit comment and refining the notes in `RELEASES.md`

I've decided to disable the button after the first tap - [dc98408](https://github.com/dimagi/commcare-android/pull/3768/commits/dc984080ac2ae69f27b1d19537e2bc4eb9e039ff)

shubham1g5 (approved):

PR #3770, opened 2026-06-22
CCCT-2489 Add Workflow Parameter To PersonalID Continue Analytics

Jignesh-dimagi:
## Technical Summary

Adds a `personalid_flow` parameter to the `personal_id_continue_clicked` analytics event to track the origin of the Continue button press.
- New `PersonalIdWorkflow` enum (`CONFIGURATION` / `USER_PROMPT` / `EDIT_PROFILE`, the last reserved for the future edit-profile feature), threaded through `FirebaseAnalyticsUtil.reportPersonalIDContinueClicked` as a required argument.
- Config-flow fragments pass `CONFIGURATION`; the email screens derive the value from their existing `EmailWorkFlow` (existing-user prompt → `USER_PROMPT`, registration/recovery → `CONFIGURATION`).
- Adds the previously-missing Continue event to `PersonalIdPhoneVerificationFragment` and `PersonalIdPhotoCaptureFragment` so they also report the workflow.

### Safety story

What gives me confidence:
- I ran the unit tests and reviewed the full diff.
- The diff is narrow and analytics-only — it adds a fire-and-forget logging parameter and does not change navigation, auth, or UI behavior.
- The new `fromEmailWorkFlow` mapping is covered by unit tests.

### Automated test coverage

`PersonalIdWorkflowTest` covers the `fromEmailWorkFlow` mapping for all three `EmailWorkFlow` values (existing-user → user_prompt, registration/recovery → configuration).

[CCCT-2489]: https://dimagi.atlassian.net/browse/CCCT-2489?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/org/commcare/fragments/personalId/PersonalIdWorkflow.kt
+++ b/app/src/org/commcare/fragments/personalId/PersonalIdWorkflow.kt
@@ -1,0 +1,29 @@
+package org.commcare.fragments.personalId
+
+/**
+ * PersonalID work flow
+ *
+ *  - [CONFIGURATION]: the PersonalID sign-up / account recovery -  configuration flow
+ *  - [USER_PROMPT]: the user is acting on a prompt surfaced outside the config flow,
+ *    e.g. the email-collection prompt shown to a logged-in user from
+ *    {@code StandardHomeActivity} (EmailOfferHelper.checkEmailCollection).
+ *  - [EDIT_PROFILE]: reserved for the future edit-profile feature.
+ */
+enum class PersonalIdWorkflow {
+    CONFIGURATION,
+    USER_PROMPT,
+    EDIT_PROFILE,
+    ;
+
+    /** Stable lowercase string emitted to analytics ("configuration", "user_prompt", "edit_profile"). */
+    val analyticsValue: String get() = name.lowercase()
+
+    companion object {
+        @JvmStatic
+        fun fromEmailWorkFlow(flow: EmailWorkFlow): PersonalIdWorkflow =
+            when (flow) {
+                EmailWorkFlow.EXISTING_USER -> USER_PROMPT
+                EmailWorkFlow.REGISTRATION, EmailWorkFlow.RECOVERY -> CONFIGURATION
+            }
+    }
+}
--- a/app/src/org/commcare/google/services/analytics/CCAnalyticsParam.java
+++ b/app/src/org/commcare/google/services/analytics/CCAnalyticsParam.java
@@ -43,6 +43,7 @@ public class CCAnalyticsParam {
     static final String PARAM_API_NEW_JOBS = "ccc_api_new_jobs";
     public static final String PERSONAL_ID_CONTINUE_CLICKED_INFO =
             "personal_id_continue_button_clicked_info";
+    public static final String PERSONAL_ID_WORKFLOW = "personalid_flow";
     static final String NOTIFICATION_EVENT_TYPE = "event_type";
     static final String NOTIFICATION_ACTION = "action";
     static final String NOTIFICATION_ID = "notification_id";
--- a/app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdWorkflowTest.kt
+++ b/app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdWorkflowTest.kt
@@ -1,0 +1,32 @@
+package org.commcare.fragments.personalId
+
+import org.junit.Assert.assertEquals
+import org.junit.Test
+
+class PersonalIdWorkflowTest {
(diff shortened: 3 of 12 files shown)


OrangeAndGreen (approved):

app/src/org/commcare/fragments/personalId/PersonalIdWorkflow.kt
```diff
@@ -0,0 +1,29 @@
+package org.commcare.fragments.personalId
+
+/**
+ * PersonalID work flow
+ *
+ *  - [CONFIGURATION]: the PersonalID sign-up / account recovery -  configuration flow
+ *  - [USER_PROMPT]: the user is acting on a prompt surfaced outside the config flow,
+ *    e.g. the email-collection prompt shown to a logged-in user from
+ *    {@code StandardHomeActivity} (EmailOfferHelper.checkEmailCollection).
+ *  - [EDIT_PROFILE]: reserved for the future edit-profile feature.
```
conroy-ricketts, line 10: Small nit: can we take the word "future" out of this just in case we forget to update it later

```suggestion
 *  - [EDIT_PROFILE]: reserved for the edit-profile feature.
```

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdWorkflowTest.kt
```diff
@@ -0,0 +1,32 @@
+package org.commcare.fragments.personalId
+
+import org.junit.Assert.assertEquals
+import org.junit.Test
+
+class PersonalIdWorkflowTest {
```
shubham1g5, line 6: nit: flagging that these tests are not achieving much given it's testing just switch case mapping and tests like this would add to overhead of changing this mapping.

shubham1g5 (approved):

app/src/org/commcare/google/services/analytics/CCAnalyticsParam.java
```diff
@@ -43,6 +43,7 @@ public class CCAnalyticsParam {
     static final String PARAM_API_NEW_JOBS = "ccc_api_new_jobs";
     public static final String PERSONAL_ID_CONTINUE_CLICKED_INFO =
             "personal_id_continue_button_clicked_info";
+    public static final String PERSONAL_ID_WORKFLOW = "personalid_flow";
```
conroy-ricketts, line 46: Nit: `personalid_` is redundant and we could use `workflow` instead

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdWorkflowTest.kt
```diff
@@ -0,0 +1,32 @@
...
+    // --- fromEmailWorkFlow ---
```
conroy-ricketts, line 7: Nit: this comment does not add much value and can be deleted

conroy-ricketts (approved):
Looks great! Just some nits

app/src/org/commcare/fragments/personalId/PersonalIdWorkflow.kt
Jignesh-dimagi: Yeah, we can modify this comment once the edit profile feature becomes available.

app/src/org/commcare/google/services/analytics/CCAnalyticsParam.java
Jignesh-dimagi: There may be different workflows across the app, but since this one is for personalid, we appended it to the name.

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdWorkflowTest.kt
Jignesh-dimagi: Yeah, it has been added for code coverage calculation purposes.

shubham1g5: Emphasizing that I don't think increasing code coverage by addition of meannigless tests should be our target as a team.

Jignesh-dimagi: I agree, but I don't think it is completely meaningless in this case, as the mapping needs to be protected from unintentional modifications across the app. Additionally, if we keep skipping these helper or mapper classes, we might miss testing small pieces of logic. Therefore, there is no harm in adding this test.

PR #3771, opened 2026-06-23
CCCT-2309 Add Unit Tests For Notification History

Jignesh-dimagi:
### Safety story

What gives confidence:
- This is a test-only change — it adds `PushNotificationActivityTest` and touches no production code, so it cannot alter app behavior.
- I ran the suite locally (`./gradlew testCommcareDebugUnitTest --tests PushNotificationActivityTest`): all 11 tests pass.

### Automated test coverage

Adds `PushNotificationActivityTest` (11 cases) covering the notification-history screen end to end through the Activity + ViewModel + adapter, mocking only the DB/API/worker/analytics boundaries: local-cache display, empty state, loading state, API-merge-and-sort, mark-as-read, fetch-error toast, cloud-sync refresh, home navigation, new-notification broadcast reload, and notification click (analytics + target intent, and the no-intent case).

[CCCT-2309]: https://dimagi.atlassian.net/browse/CCCT-2309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/org/commcare/activities/connect/viewmodel/PushNotificationViewModel.kt
+++ b/app/src/org/commcare/activities/connect/viewmodel/PushNotificationViewModel.kt
@@ -2,86 +2,70 @@ package org.commcare.activities.connect.viewmodel
 
 import android.app.Application
 import android.content.Context
-import androidx.annotation.VisibleForTesting
 import androidx.lifecycle.AndroidViewModel
 import androidx.lifecycle.LiveData
 import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.application
 import androidx.lifecycle.viewModelScope
-import kotlinx.coroutines.CoroutineDispatcher
-import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 import org.commcare.android.database.connect.models.PushNotificationRecord
 import org.commcare.connect.database.NotificationRecordDatabaseHelper
 import org.commcare.pn.workermanager.NotificationsSyncWorkerManager
 import org.commcare.utils.PushNotificationApiHelper
 import org.commcare.utils.PushNotificationApiHelper.retrieveLatestPushNotifications
+import org.commcare.utils.coroutines.DispatcherProvider
 
-class PushNotificationViewModel
-    @JvmOverloads
-    constructor(
-        application: Application,
-        private val ioDispatcher: CoroutineDispatcher = dispatcherOverride ?: Dispatchers.IO,
-    ) : AndroidViewModel(application) {
-        private val _fetchApiError =
-            MutableLiveData<String>()
-        val fetchApiError: LiveData<String> =
-            _fetchApiError
+class PushNotificationViewModel(
+    application: Application,
+) : AndroidViewModel(application) {
+    private val _fetchApiError =
+        MutableLiveData<String>()
+    val fetchApiError: LiveData<String> =
+        _fetchApiError
 
-        private val _allNotifications = MutableLiveData<List<PushNotificationRecord>>()
-        val allNotifications: LiveData<List<PushNotificationRecord>> = _allNotifications
-        private val _isLoading = MutableLiveData<Boolean>()
-        val isLoading: LiveData<Boolean> = _isLoading
+    private val _allNotifications = MutableLiveData<List<PushNotificationRecord>>()
+    val allNotifications: LiveData<List<PushNotificationRecord>> = _allNotifications
+    private val _isLoading = MutableLiveData<Boolean>()
+    val isLoading: LiveData<Boolean> = _isLoading
 
-        fun loadNotifications(
-            isRefreshed: Boolean,
-            context: Context,
-        ) {
-            viewModelScope.launch(ioDispatcher) {
-                _isLoading.postValue(true)
-                // Load from DB first
-                if (!isRefreshed) {
-                    val cachedNotifications =
-                        NotificationRecordDatabaseHelper
-                            .getAllNotifications(context)
-                            .orEmpty()
-                            .sortedByDescending { it.createdDate }
-                    if (cachedNotifications.isNotEmpty()) {
-                        _isLoading.postValue(false)
-                    }
-                    _allNotifications.postValue(cachedNotifications)
+    fun loadNotifications(
+        isRefreshed: Boolean,
+        context: Context,
+    ) {
+        viewModelScope.launch(DispatcherProvider.io()) {
+            _isLoading.postValue(true)
+            // Load from DB first
+            if (!isRefreshed) {
+                val cachedNotifications =
+                    NotificationRecordDatabaseHelper
+                        .getAllNotifications(context)
+                        .orEmpty()
+                        .sortedByDescending { it.createdDate }
+                if (cachedNotifications.isNotEmpty()) {
+                    _isLoading.postValue(false)
                 }
-
-                val latestPushNotificationsFromApi = retrieveLatestPushNotifications(context)
-                latestPushNotificationsFromApi
-                    .onSuccess {
-                        val currentNotifications = _allNotifications.value.orEmpty()
-                        NotificationsSyncWorkerManager(
-                            context = application,
-                            showNotification = false,
-                            syncNotification = false,
-                        ).startSyncWorkers(PushNotificationApiHelper.convertPNRecordsToPayload(it))
-                        val updatedNotifications =
-                            (it + currentNotifications)
-                                .distinctBy { it.notificationId }
-                                .sortedByDescending { it.createdDate }
-                        _isLoading.postValue(false)
-                        _allNotifications.postValue(updatedNotifications)
-                    }.onFailure {
-                        _isLoading.postValue(false)
-                        _fetchApiError.postValue(it.message)
-                    }
+                _allNotifications.postValue(cachedNotifications)
             }
-        }
 
-        companion object {
-            // Notification loading runs on Dispatchers.IO, a real background thread pool. Robolectric
-            // only drives the main looper and can't deterministically schedule or wait on that pool,
-            // making such tests flaky. This lets tests swap in a test dispatcher so loading runs
-            // synchronously. The activity creates this ViewModel and loads immediately, so the override
-            // must be set before construction (a post-construction setter would race) - hence a
-            // companion field.
-            @VisibleForTesting
-            internal var dispatcherOverride: CoroutineDispatcher? = null
+            val latestPushNotificationsFromApi = retrieveLatestPushNotifications(context)
+            latestPushNotificationsFromApi
+                .onSuccess {
+                    val currentNotifications = _allNotifications.value.orEmpty()
+                    NotificationsSyncWorkerManager(
+                        context = application,
+                        showNotification = false,
+                        syncNotification = false,
+                    ).startSyncWorkers(PushNotificationApiHelper.convertPNRecordsToPayload(it))
+                    val updatedNotifications =
+                        (it + currentNotifications)
+                            .distinctBy { it.notificationId }
+                            .sortedByDescending { it.createdDate }
+                    _isLoading.postValue(false)
+                    _allNotifications.postValue(updatedNotifications)
+                }.onFailure {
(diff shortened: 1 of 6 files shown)


app/unit-tests/src/org/commcare/activities/PushNotificationActivityTest.kt
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    @Test
+    fun `api notifications are merged with cached ones and sorted newest first`() {
+        dbNotifications = listOf(buildRecord("cached", Date(1000L)))
+        apiResult = {
+            Result.success(
+                listOf(
+                    buildRecord("api", Date(2000L)),
+                    buildRecord("cached", Date(1000L)),
```
shubham1g5: does this API return all notifications or just the new ones ?

app/unit-tests/src/org/commcare/activities/PushNotificationActivityTest.kt
```diff
@@ -0,0 +1,302 @@
...
+        while (!condition()) {
+            if (System.currentTimeMillis() > deadline) {
+                fail("Timed out waiting for $description")
+            }
+            ShadowLooper.idleMainLooper()
+            Thread.sleep(10)
+        }
+    }
+
+    private fun layoutRecyclerView() {
```
shubham1g5: why do we need to do this in tests ?

app/unit-tests/src/org/commcare/activities/PushNotificationActivityTest.kt
```diff
@@ -0,0 +1,302 @@
...
+    private fun waitFor(
+        description: String,
+        condition: () -> Boolean,
+    ) {
+        val deadline = System.currentTimeMillis() + 5000
+        while (!condition()) {
+            if (System.currentTimeMillis() > deadline) {
+                fail("Timed out waiting for $description")
+            }
+            ShadowLooper.idleMainLooper()
```
shubham1g5: think the calls to `ShadowLooper.idleMainLooper()` or `ShadowLooper.runUiThreadTasksIncludingDelayedTasks()`  should be sufficient here instead of implemeting a timed loop.

app/unit-tests/src/org/commcare/activities/PushNotificationActivityTest.kt
```diff
@@ -0,0 +1,302 @@
...
+    fun `clicking a notification reports analytics and opens the target screen`() {
+        val record = buildRecord("a")
+        apiResult = { Result.success(listOf(record)) }
+
+        launchActivity()
+        waitFor("notification list to become visible") { recyclerView.visibility == View.VISIBLE }
+        layoutRecyclerView()
+
+        val targetIntent = Intent(activity, DispatchActivity::class.java)
+        every { FirebaseMessagingUtil.getIntentForPNClick(any(), any()) } returns targetIntent
```
shubham1g5: instead of mocking any intents here, think we would like the tests to verify the resultant redirected screens /activity based on each different type of notification payload.

shubham1g5 (changes requested):

conroy-ricketts (dismissed):
LGTM, nothing additional to add

app/unit-tests/src/org/commcare/activities/PushNotificationActivityTest.kt
Jignesh-dimagi: The API will only return new notifications, but here the code is referencing the notificationId to create the record.

app/unit-tests/src/org/commcare/activities/PushNotificationActivityTest.kt
Jignesh-dimagi: It was added but not required, thanks for flagging. 
https://github.com/dimagi/commcare-android/pull/3771/changes/5653252f8aa4e75b1615cc217f5521c527feaa83

app/unit-tests/src/org/commcare/activities/PushNotificationActivityTest.kt
Jignesh-dimagi: Now only ShadowLooper.idleMainLooper() is used.
The production coroutine runs on the real Dispatchers.IO and Robolectric  runs on main/test thread now.  With the coroutine on Main/test thread, Robolectric's main looper owns all the work, ShadowLooper.idleMainLooper() executes the load, the LiveData posts, and the observers synchronously. So no more wait and delaly.

https://github.com/dimagi/commcare-android/pull/3771/changes/5653252f8aa4e75b1615cc217f5521c527feaa83

app/unit-tests/src/org/commcare/activities/PushNotificationActivityTest.kt
Jignesh-dimagi: Originally, FirebaseMessagingUtilTest should be responsible for verifying that the correct intent to be delivered, so FirebaseMessagingUtil was mocked here. However, I have removed that mock; the activity now calls the real methods of FirebaseMessagingUtil, and test assert against the class name of the started activity instead.

OrangeAndGreen (dismissed):

conroy-ricketts (dismissed):

app/src/org/commcare/activities/connect/viewmodel/PushNotificationViewModel.kt
```diff
@@ -15,57 +17,71 @@ import org.commcare.pn.workermanager.NotificationsSyncWorkerManager
...
-) : AndroidViewModel(application) {
-    private val _fetchApiError =
-        MutableLiveData<String>()
-    val fetchApiError: LiveData<String> =
-        _fetchApiError
+class PushNotificationViewModel
+    @JvmOverloads
+    constructor(
+        application: Application,
+        private val ioDispatcher: CoroutineDispatcher = dispatcherOverride ?: Dispatchers.IO,
```
shubham1g5: I do not really like the approach here around test overrides as I think this will lead to a bunch of `dispatcherOverride` variables at any place we want to write test for and that includes a Coroutine. Curious if you have explored any alternatives here around injecting the test Dispatcher without trying to introduce these class level override variables ?

Jignesh-dimagi: I previously tried using time-based polling instead of changing the dispatchers. The reasoning behind that approach was that PushNotificationActivityTest is the first test that drives an activity whose ViewModel jumps to Dispatchers.IO. Robolectric can't see it (since it's not a looper or AsyncTask thread), leaving real-time polling—hence waitFor with Thread.sleep—as the only remaining tool.
However, while these tests passed locally, they failed consistently on CI due to a race condition, so it isn't a reliable path forward. The only robust solution is to implement a dispatcherOverride mechanism, allowing the production code to use Dispatchers.IO while the test suite utilizes a test dispatcher to load notifications synchronously. The test suite is already injecting the test dispatcher. To answer your question specifically, this approach provides one of the cleanest ways to separate the test coroutines from the real production activity. 
I am open to suggestions and thoughts here.

shubham1g5: How about creating a `DispatcherProvider` class and have methods like `getIODispatcher` that we can override for all tests at a single place ? This will remove the need of adding the `dispatcherOverride` at several places in the non-test code.

Jignesh-dimagi: I'm not sure if that will work here, since this activity test doesn't mock the ViewModel. If you can elaborate a bit more on your idea, I'd be happy to give it a try.

Jignesh-dimagi: We can clean this up by creating a BaseActivityViewModel and moving the dispatcherOverride there. This way, we won't have to define it in every single ViewModel manually.

shubham1g5: Sure, I was talking about a `object DispatcherProvider` (static methods) that we can use in code and can  directly mock in tests rather than mocking the `ViewModel` . This has an advantage over `BaseActivityViewModel` idea that it works for all components and not just view models.

Jignesh-dimagi: Agree, it can be used for all components. Changed [here](https://github.com/dimagi/commcare-android/pull/3771/changes/e6596d036d33c1ed660ecd4658b615f84c65b7d0).

conroy-ricketts (dismissed):

app/src/org/commcare/activities/connect/viewmodel/PushNotificationViewModel.kt
shubham1g5: this looks great, thanks!

app/unit-tests/src/org/commcare/activities/PushNotificationActivityTest.kt
```diff
@@ -0,0 +1,372 @@
...
+        every { DispatcherProvider.io() } returns UnconfinedTestDispatcher()
+
+        // getIntentForPNClick is exercised for real, so cccCheckPassed() must see a logged-in
+        // user. init() is a no-op unless status == NotIntroduced, so setting LoggedIn is enough.
+        val personalIdManager = PersonalIdManager.getInstance()
+        savedPersonalIdStatus = personalIdManager.status
+        personalIdManager.status = PersonalIdManager.PersonalIdStatus.LoggedIn
+
+        mockkObject(NotificationRecordDatabaseHelper)
+        mockkObject(PushNotificationApiHelper)
```
shubham1g5: think using `MockWebServer` is a better approach here as it mocks the pure OkHttp/network layer and not any logic on our side providing better test coverage on our classes.

shubham1g5 (commented):
Left one more suggestion which I think aligns better with the testing strategy we want to adopt and would be great if we can iterate on that.

app/unit-tests/src/org/commcare/activities/PushNotificationActivityTest.kt
Jignesh-dimagi: This seems like a slightly different strategy than what we are currently doing. I see your point about using an end-to-end integration style to include all connecting code, which would definitely reduce dev effort and the total number of tests.

However, for this specific case, I didn't mock the web server because we already cover that logic separately in PushNotificationApiHelperTest, and I wanted to avoid duplicating it in this activity test.

shubham1g5: Think it would be good to set a consistent experience here in tests (It seems inconsistent that we are using MockWebServer in other test classes but mocking differently here) and to be fair we won't be duplicating tests in `PushNotificationApiHelperTest` but will be just providing a different mock in the current test class. I agree though it's redundant coverage for `PushNotificationApiHelperTest` but think that's harmless and sets us up better for implementing standard test practices in the current test class. With that said, I would not block this PR on this feedback but do really think it's a positive change to make on the PR.

shubham1g5 (dismissed):

app/unit-tests/src/org/commcare/activities/PushNotificationActivityTest.kt
Jignesh-dimagi: Alright.  PushNotificationActivityTest now drives data through a MockWebServer instead of mocking PushNotificationApiHelper, so the real helper, response parser, and network plumbing all execute and only the socket is faked. I extracted the server wiring into PersonalIdMockApiServer helper (also adopted by BasePersonalIdConfigurationTest, removing the duplicated setup there), and routed the helper's internal coroutine dispatchers through DispatcherProvider so the loads stay deterministic under Robolectric.
https://github.com/dimagi/commcare-android/pull/3771/changes/4253dd21784bc1c489299fb52788603343bdb353

shubham1g5: Great addition, thanks!

shubham1g5 (approved):

PR #3775, opened 2026-06-25
Add troubleshooting logging for casedb instance crash in entity list

avazirna:
## Product Description

No user-facing changes. Instrumentation only (Crashlytics breadcrumbs / soft-assert logs).

## Technical Summary

Adds diagnostic logging to investigate the crash:

> `RuntimeException: Unable to expand reference instance(casedb)/casedb/case[@case_type = 'menage'][@status = 'open'], no appropriate instance in evaluation context`

The crash occurs in the background `EntityLoaderTask` when the evaluation context built for the load contains no `casedb` instance — i.e. the live session command at load time does not declare it (or is null/diverged from the command captured at activity creation).

This branch is **instrumentation only**; it does not change behavior. What it logs:
- **Entry points into `loadEntities()`** — `refreshView()`, the refresh timer, `deliverLoadResult()` (the deferred location reload), and `onEvalLocationChanged()` — each records the live session command, selectDatum, and `isFinishing()`.
- **`EntityLoaderTask` failure path** — the available instance ids and whether a main instance is present at the moment the load throws (most direct evidence of why `casedb` is missing).
- **Activity lifecycle ordering** — `onPause`/`onStop`/`onDestroy` breadcrumbs, so the crash trail shows whether the load completed after the user navigated away.
- **`HomeScreenBaseActivity`** — caller chain + current command on back / end-of-form navigation.

Leading hypothesis: the `deliverLoadResult()` deferred GPS reload (case lists using `here()`) fires after the user has navigated away — the loader callback is not tied to activity visibility — and reloads **unguarded** against a session that has already advanced/cleared. The logs are intended to confirm this before committing to a fix.

### Safety story

Inherently low risk: every change is a log statement. `CrashUtil.log` is internally gated on `crashlyticsEnabled` and does not throw; new accessors null-check their fields; no control flow, data, or storage is modified. No impact on existing data.

⚠️ All additions are marked temporary for troubleshooting and must be reverted before this reaches a release line (the new `EntitySelectActivity` accessors and `CrashUtil.getTopCallerChain()` are not intended to ship).

### Automated test coverage

None — logging-only change, no behavioral coverage applicable.

### QA Plan

No QA required beyond confirming the app builds and runs; there are no behavioral changes to regress.

## Labels and Review

### DIFF
--- a/app/src/org/commcare/utils/CrashUtil.java
+++ b/app/src/org/commcare/utils/CrashUtil.java
@@ -56,4 +56,28 @@ public class CrashUtil {
             FirebaseCrashlytics.getInstance().log(message);
         }
     }
+
+    // temporary for troubleshooting purposes only
+    public static String getTopCallerChain() {
+        StackTraceElement[] stackTrace = Thread.currentThread().getStackTrace();
+        StringBuilder sb = new StringBuilder();
+        int FRAMES_TO_SKIP = 3;
+        int MAX_CHAIN_LENGTH = 5;
+        int limit = Math.min(FRAMES_TO_SKIP + MAX_CHAIN_LENGTH, stackTrace.length);
+        for (int i = FRAMES_TO_SKIP; i < limit; i++) {
+            sb.append(getSimpleClassName(stackTrace[i].getClassName()))
+                    .append(".")
+                    .append(stackTrace[i].getMethodName())
+                    .append(":")
+                    .append(stackTrace[i].getLineNumber());
+            if (i < limit - 1) {
+                sb.append(" <- ");
+            }
+        }
+        return sb.toString();
+    }
+
+    private static String getSimpleClassName(String fullyqualifiedClassName) {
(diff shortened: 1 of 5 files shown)


app/src/org/commcare/utils/CrashUtil.java
```diff
@@ -56,4 +56,28 @@ public static void log(String message) {
...
+                    .append(".")
+                    .append(stackTrace[i].getMethodName())
+                    .append(":")
+                    .append(stackTrace[i].getLineNumber());
+            if (i < limit - 1) {
+                sb.append(" <- ");
+            }
+        }
+        return sb.toString();
+    }
```
conroy-ricketts, line 78: Curious if there isn't an easier way to print the stack trace?

conroy-ricketts (approved):

app/src/org/commcare/utils/CrashUtil.java
avazirna: the goal is to print the top of the stack without too much detail. I wonder if there is a utility that offers the same functionality.

PR #3776, opened 2026-06-26
Spec: App Seating in Opportunity Home

OrangeAndGreen:
https://dimagi.atlassian.net/browse/CCCT-2522

## Product Description
No code changes

## Technical Summary
Claude tech spec for a refactor to support optional app seating in the new Opportunity Home activity, so that it can be run before the associated learn and deliver apps are downloaded and installed, and then provide "standard app home" functionality once the relevant apps are downloaded, seated, and silently logged in.

The refactor involves extracting several delegates that the existing framework can rely on (without behavioral changes) while the new activity can optionally use.


conroy-ricketts (dismissed):
LGTM, would like another set of eyes before creating new tickets

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
```diff
@@ -0,0 +1,148 @@
...
+## Non-goals
+
+- Modifying `SessionAwareCommCareActivity`. Every other session-gated activity in the app keeps the existing redirect-to-login behavior.
+- Replacing `StandardHomeActivity` for non-Connect users.
+- Designing the state model that picks which UI to show inside Opportunity Home (handled separately).
+- Designing entry points, routing through `DispatchActivity`, or persistence of "last accessed opportunity".
+- Designing the inline silent app-login flow (in-flight work). This spec assumes that capability is available and exposes a session to the activity when it completes.
+- Implementing `WithUIController` on the new activity.
+
+## Approach
```
shubham1g5, line 97: Think one important goal/north-star for this design should be that if in future if we need to add or change  a feature to home screen, we should not need to make the changes in both activities/screens.  

Accordingly, It would be great to define both Inheritance and Composition Approaches in this spec and try to evaluate those two with the app architecutre goal above to decide on what approach makes better sense instead of going by a rule or feeling to favor composition approach (I know I am going against the earlier discussion during co-working hour but just want us to make a more informed decision here)

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
```diff
@@ -0,0 +1,148 @@
...
+Each delegate is a Kotlin class implementing `DefaultLifecycleObserver`, registered on the host's `lifecycle` so it receives `onResume` / `onPause` / `onDestroy` directly. The host activity forwards `onActivityResult` and intent handling to the delegates that need them.
+
+Session-dependent delegates — `SessionExpirationDelegate`, `SyncDelegate`, and `SessionLaunchDelegate` — expose two explicit methods:
+
+```
+attachSession(session: SeatedAppSession)
+detachSession()
+```
+
+When attached, the delegate has a live session and operates as the existing chain does today. When not attached, every public entry point is a clean no-op (callers receive a documented "no session" result; nothing throws `SessionUnavailableException`).
```
shubham1g5, line 120: curious why we decided against the current  ` throw SessionUnavailableException` here ? Probably the answer here is hidden in how and when `attachSession` and `detachSession` is called but is not very clear based on spec.

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
```diff
@@ -0,0 +1,148 @@
...
+| `SyncDelegate` | `SyncCapableCommCareActivity` | Owns `FormAndDataSyncer`, sync state, and the `PullTaskResultReceiver` implementation. Exposes `sendFormsOrSync()` and related entry points. The host activity remains the `CommCareTaskConnector` (see below). |
+| `AppUpdateDelegate` | `HomeScreenBaseActivity` (`AppUpdateController` and drift checks) | App update prompts and drift warnings. **Session-independent** (see below): constructed once and registered on the host lifecycle like `CrashRecoveryDelegate`; only the surfacing of prompts is gated on a session existing. |
+| `SessionLaunchDelegate` | `HomeScreenBaseActivity.doLoginLaunchChecksInOrder`, `SessionNavigator` usage | Form restoration, "Start" → form entry navigation, post-login launch checks, and form-result handling. Must preserve the ordering and early-return semantics of `doLoginLaunchChecksInOrder` (see risks). |
+| `CrashRecoveryDelegate` | `HomeScreenBaseActivity` instance-state and crash-data registration | Persists instance state across recreation. Active regardless of session. |
```
shubham1g5: what does `instance-state` refers to here and how that is only specific to `CrashRecoveryDelegate` ?

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
```diff
@@ -0,0 +1,148 @@
...
+
+### Refactor of existing base classes
+
+To avoid two implementations of the same behaviors, `SyncCapableCommCareActivity` and `HomeScreenBaseActivity` are rebased onto the new delegates. Their public API stays the same; internally they instantiate the delegates, register them with the activity's lifecycle, and forward calls. Because those base classes still extend `SessionAwareCommCareActivity`, they call `attachSession(...)` once in `onCreateSessionSafe` and never detach (they cannot reach a session-less state). `StandardHomeActivity` is unchanged externally.
```
shubham1g5: Think we might still need a common `HomeController` in this approach that takes care of managing all the delegates.

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
```diff
@@ -0,0 +1,148 @@
...
+    override fun onActivityResult(...) {
+        sessionExpirationDelegate.onActivityResult(...)
+        syncDelegate.onActivityResult(...)
+        sessionLaunchDelegate.onActivityResult(...)
+        super.onActivityResult(...)
+    }
+}
+```
+
+`onSessionAvailable(...)` is invoked from two places inside the activity: the completion path for inline silent login (driving the situation 2 → 3 and 4 → 5 transitions), and an `onResume` check that calls it whenever a session exists but no delegate is currently attached. `onSessionLost()` is invoked from two distinct mechanisms, which today reach the chain through different paths:
```
shubham1g5: what are the numerals `2 → 3 and 4 → 5 transitions` referring to here ?

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
```diff
@@ -0,0 +1,148 @@
...
+
+All except app updates are routed through delegates and gated on `attachSession(...)` having been called:
+
+- **Sync** — `SyncDelegate.sendFormsOrSync()`. `SyncDelegate` *is* the `PullTaskResultReceiver`; the host activity stays the `CommCareTaskConnector` passed to `DataPullTask.connect(...)` and its `getReceiver()` returns the delegate (see risks). This keeps the in-activity blocking/spinner UI while moving the result-handling implementation into the delegate.
+- **Form entry** — `SessionLaunchDelegate.startForm(...)`; result handling fans out to `SyncDelegate` (post-form sync prompt) and `SessionExpirationDelegate` (mid-form expiration check), matching the current `HomeScreenBaseActivity` pipeline.
+- **App updates** — `AppUpdateDelegate` is always registered; it self-gates whether to surface a prompt based on whether a session currently exists (re-evaluating `shouldShowInAppUpdate()`), rather than being attached/detached.
+- **Language switching** — app-based, therefore session-gated. The activity exposes language switching only while attached; routed through whichever existing component owns the change (verified during implementation).
```
shubham1g5: How do we plan to avoid duplicating this code and similar code for any other menu items in the 2 home activities ?

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Risks and mitigations
+
+- **Behavior drift between the chain and the new activity.** Mitigated by making the rebased base classes use the same delegates the new activity composes — one source of truth. Forwarding-only changes in the base classes minimize the chance of semantic divergence.
+- **Delegates currently read `CommCareApplication.instance().getCurrentSession()` ambiently.** Required change: thread the session explicitly through `attachSession(...)`. Audit each migrated callsite for ambient reads and replace.
```
shubham1g5, line 318: The `session` object changes a lot on user navigating from App Home to the menu and form screens (through SessionNavigator) and think we would need to ensure that those changes are reflected in the session copies of the delegators or the delegators should directly access it using `CommCareApplication.instance().getCurrentSession()`

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
```diff
@@ -0,0 +1,148 @@
...
+3. **`AppUpdateController` lifecycle — resolved (do not rebind per session).** Reconstructing the controller on each `attachSession` would leak Play Core listeners, orphan in-progress downloads, and re-fetch needlessly. `AppUpdateDelegate` is therefore session-independent: constructed once on the host lifecycle, self-gating only prompt surfacing on a session existing. See Approach.
+
+Remaining verify-during-implementation items:
+
+4. **Preserve `doLoginLaunchChecksInOrder` semantics.** This method is a strict 9-step ordered pipeline (demo-mode early return → update-info form → form restoration → session restoration → post-update sync → pending FCM sync → update prompt → PIN check → drift check). When `SessionLaunchDelegate` and `AppUpdateDelegate` absorb pieces of it, the ordering and early-return behavior must be preserved, or demo users could start receiving prompts they currently skip.
```
shubham1g5: Something this spec can expand on -> "How would we avoid duplication for  the 9-step ordered pipeline in this method ?"

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
OrangeAndGreen: This is referring to Android's saved instance state, the Bundle accessed via onSaveInstanceState and onCreate/onCreateView.

It isn't specific to CrashRecoveryDelegate, but the spec was assigning behavior to that delegate related to the saving/loading of several variables kept in the state bundle. But looking at the flags to be managed, they seem more appropriate to be handled by the SessionLaunchDelegate. Spec updated to make that change and be more clear. [05ea7f6](https://github.com/dimagi/commcare-android/pull/3776/commits/05ea7f6ad475c1223f1bbed1e37ec15bcd800631)

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
OrangeAndGreen: These numbers refer to the list in the Summary section at the top. More specifically, they represent the transition from app-not-ready to app-ready state (i.e. seating app and starting user session) within Opp Home in both learn (2>>3) and deliver (4>>5) scenrarios.

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
OrangeAndGreen: Yeah, I like that idea. Claude convinced me to call it a Coordinator instead of Controller to better distinguish from some similarly named controllers that we already have, so I went with HomeActivityCoordinator. It holds reusable logic that the home activities would otherwise have to duplicate. [f8ad03c](https://github.com/dimagi/commcare-android/pull/3776/commits/f8ad03c57d3b561a55f096c82c3096dc951846f6)

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
OrangeAndGreen: I added the responsibility of handling common logic to the newly introduced HomeActivityCoordinator. That will allow either home activity to call common code when performing any of the relevant actions described here (whether they're hosted in the menu or somewhere else in the UI). [f8ad03c](https://github.com/dimagi/commcare-android/pull/3776/commits/f8ad03c57d3b561a55f096c82c3096dc951846f6)

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
OrangeAndGreen: My understanding is that while the session object changes a lot, it remains the same object. In other words, it's modified but never replaced. The attachSession function passes the session by reference such that each delegate references the original (not a copy) and therefore sees any changes.

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
OrangeAndGreen: I modified the spec to specify that the launch checks will be performed by HomeActivityCoordinator in a common helper that ensures the correct ordering is preserved. [f298b04](https://github.com/dimagi/commcare-android/pull/3776/commits/f298b0408f75c3c78f5eaba747e00b5f322d46ae)

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
```diff
@@ -0,0 +1,186 @@
...
+- Designing the inline silent app-login flow (in-flight work). This spec assumes that capability is available and exposes a session to the activity when it completes.
+- Implementing `WithUIController` on the new activity.
+
+## Approach
+
+`OpportunityHomeActivity` extends `BaseDrawerActivity` — the layer immediately below the session gate that still provides the drawer and the underlying `CommCareActivity` capabilities.
+
+Five behaviors are extracted from the existing inheritance chain into delegates, and a single `HomeActivityCoordinator` owns those delegates and exposes the home screen's capabilities as host-agnostic actions (see [`HomeActivityCoordinator`](#homeactivitycoordinator) below). A host activity holds one coordinator rather than five delegates directly. The five extracted delegates are:
+
+| Delegate | Source today | Responsibilities |
```
Jignesh-dimagi, line 103: As stated above, OpportunityHomeActivity can have the following states also: `Learning state, learn app not installed, seated, or logged in`. How should we handle the login process here? Do we need a dedicated delegate to handle it, or can we re-use ConnectAppLaunchController?

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
```diff
@@ -0,0 +1,186 @@
...
+|---|---|---|
+| `SessionExpirationDelegate` | `SessionAwareHelper.onResumeHelper`, `onActivityResultHelper` | On resume / activity-result, check session expiration. Exposes a listener interface that the host activity registers; reports session-lost events via that listener rather than redirecting to login. |
```
Jignesh-dimagi, line 105: `HomeActivityCoordinator` should be the listener instead of the host activity, as it is the ultimate consumer of these events?

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
```diff
@@ -0,0 +1,186 @@
...
+    override fun onActivityResult(...) {
+        coordinator.onActivityResult(...)  // fans out to delegates that need it
+        super.onActivityResult(...)
+    }
+}
+```
+
+The coordinator's `onCreate(...)` is what registers the five delegates as lifecycle observers and unconditionally restores `SessionLaunchDelegate`'s instance state; the activity no longer enumerates delegates itself. `onSessionAvailable(...)` is invoked from two places inside the activity: the completion path for inline silent login (driving the situation 2 → 3 and 4 → 5 transitions), and an `onResume` check that calls it whenever a session exists but no session is currently attached. `onSessionLost()` is invoked from two distinct mechanisms, which today reach the chain through different paths:
+
+- **Manual logout** — `userTriggeredLogout()` → `CommCareApplication.closeUserSession()` (no broadcast).
```
Jignesh-dimagi, line 231: Will this support user logout? If so, what should the state of OpportunityHomeActivity be? Specifically, do we ever navigate away from this activity once a user has an active opportunity in progress, considering this becomes our app's new home screen?

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
```diff
@@ -0,0 +1,186 @@
...
+
+`CrashRecoveryDelegate` and `AppUpdateDelegate` do not implement `attachSession`/`detachSession`. `CrashRecoveryDelegate` is session-independent by nature. `AppUpdateDelegate` is treated as session-independent because the underlying `AppUpdateController` cannot be safely reconstructed per session: `AppUpdateControllerFactory.create(...)` needs only a `Context` and a callback (not a session), and `register()` attaches an `InstallStateUpdatedListener` to the Google Play Core `AppUpdateManager` and kicks off an async info fetch. Reconstructing on each `attachSession` would leak listeners (duplicate callbacks), orphan any in-progress download from the Play Store state machine, and re-fetch update info needlessly. App-binary updates are not opportunity- or seated-app-scoped, so the delegate is constructed once on the host lifecycle (register on `onResume`, unregister on `onDestroy`); only the *surfacing* of an update prompt is gated on a session existing, re-evaluating `shouldShowInAppUpdate()` at prompt time rather than at construction.
+
+### `HomeActivityCoordinator`
```
Jignesh-dimagi, line 135: I know designing the state model that picks which UI to show inside OpportunityHome is a non-goal for this document (as it is handled separately). However, I am just trying to understand if HomeActivityCoordinator will be the one controlling the UI via HomeActivityHost based on the opportunity state (e.g., learn not downloaded, learn not completed, delivery not downloaded), or if it will use another coordinator or controller for that purpose?

Jignesh-dimagi:
This is really good documentation. I have left a few questions and comments, but they seem to be mostly for my own understanding :).

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
shubham1g5: Think all delegates will optionally need to have their respective methods to save and restore the Instance state they own (instead of assigning this responsibility to a single delegate). Additionally controller class should implement these methods as well and  should call the corresponding delegates methods and also restore it's own state. 

Looking at the current vars - 

````
 outState.putBoolean(WAS_EXTERNAL_KEY, wasExternal);

outState.putBoolean(EXTRA_CONSUMED_KEY, loginExtraWasConsumed);

outState.putBoolean(
                KEY_PENDING_ENDPOINT_NAV_AFTER_SYNC,
                pendingEndpointNavigationAfterSync
        );
````

I anticipate that each of the 3 variables here would probably go to different delegates or live in the conroller class.

File: docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md (hunk 5 above)
Comment (shubham1g5) [reply docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md]: got it, thanks

File: docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md (hunk 7 above)
Comment (shubham1g5) [reply docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md:318]: yeah that makes sense.

File: docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md (hunk 10 above)
Comment (shubham1g5) [inline docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md:105]: the `onActivityResult` should be distribitued across delegates and `HomeActivityCoordinator` similar to the methods around storing/restoring `instance-state`

File: docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md [hunk 13]
```diff
@@ -0,0 +1,186 @@
...
+
+| Delegate | Source today | Responsibilities |
+|---|---|---|
+| `SessionExpirationDelegate` | `SessionAwareHelper.onResumeHelper`, `onActivityResultHelper` | On resume / activity-result, check session expiration. Exposes a listener interface that the host activity registers; reports session-lost events via that listener rather than redirecting to login. |
+| `SyncDelegate` | `SyncCapableCommCareActivity` | Owns `FormAndDataSyncer`, sync state, and the `PullTaskResultReceiver` implementation. Exposes `sendFormsOrSync()` and related entry points. The host activity remains the `CommCareTaskConnector` (see below). |
+| `AppUpdateDelegate` | `HomeScreenBaseActivity` (`AppUpdateController` and drift checks) | App update prompts and drift warnings. **Session-independent** (see below): constructed once and registered on the host lifecycle like `CrashRecoveryDelegate`; only the surfacing of prompts is gated on a session existing. |
```
shubham1g5: drift checks are quite different than app update and think are best separated in different delegates Or we can keep the drift check it in the coordinator given it's a small check.

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
```diff
@@ -0,0 +1,186 @@
...
+| `SessionLaunchDelegate` | `HomeScreenBaseActivity.doLoginLaunchChecksInOrder`, `SessionNavigator` usage, `onSaveInstanceState`/`loadInstanceState` | Form restoration, "Start" → form entry navigation, the step bodies it contributes to the launch pipeline (update-info form, form restoration, session restoration), and form-result handling. The *ordering* and early-return semantics of `doLoginLaunchChecksInOrder` are owned by the coordinator's `runLaunchChecks` (see [`HomeActivityCoordinator`](#homeactivitycoordinator)), not by this delegate. Also owns the launch/nav instance-state keys — `WAS_EXTERNAL_KEY` (`wasExternal`), `EXTRA_CONSUMED_KEY` (`loginExtraWasConsumed`), and `KEY_PENDING_ENDPOINT_NAV_AFTER_SYNC` (`pendingEndpointNavigationAfterSync`) — persisting them across recreation (see note below on save/restore vs. attach). |
```
shubham1g5: A bit confsed what parts of `doLoginLaunchChecksInOrder` is owned by  the `SessionLaunchDelegate`  given the spec states that `The ordering and early-return semantics of doLoginLaunchChecksInOrder are owned by the coordinator's runLaunchChecks (`

shubham1g5 (commented):
- One other thing that would be good to preclude this work is documenting the current behaviour as it is for HomeActivity with unit tests so that if something functional changes as part of the refactor, tests are better equipped to flag that.

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
OrangeAndGreen: Throwing that exception still happens when a delegate encounters an error while a session is attached, but when the delegates shouldn't do anything when there isn't a session detached. This allows them to be wired to lifecycle callbacks, but only active when the owning Activity wants them to be (always active for traditional pages, but optionally for Opp Home).

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
OrangeAndGreen: Ah, I like moving those three variables to the Coordinator class now that we have one. [3e35c00](https://github.com/dimagi/commcare-android/pull/3776/commits/3e35c00acf2b0ba5db923e345a25b97fcb44bd4b)

I also did another check to verify, and there's one more instance-state variable currently managed by a delegate: KEY_LAST_ICON_TRIGGER managed by SyncDelegate. I think that one makes sense to leave where it is since it's only relevant when a session is attached and is specific to sync functionality (computing sync state in onCreateSessionSafe). I added some text clarifying the usage of the 4 variables and their different ownership. [889f187](https://github.com/dimagi/commcare-android/pull/3776/commits/889f187587f3bfef6652dfa27de969b45f549e18)

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
OrangeAndGreen: Yes, the plan is to mostly re-use ConncetAppLaunchController for this, although we'll also use ConnectAppUtils.downloadApp to download the app if necessary. I added a couple commits to help clarify this part: [acace20](https://github.com/dimagi/commcare-android/pull/3776/commits/acace20514afe3f2c27f17af5f88f3875cd21213) and [49643ff](https://github.com/dimagi/commcare-android/pull/3776/commits/49643ff0cbd312d07446c6376c555380c46e6520)

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
OrangeAndGreen: That's correct, I added some text clarifying that the Coordinator will register for life cycle callbacks. [ab8d11c](https://github.com/dimagi/commcare-android/pull/3776/commits/ab8d11ca6360f7c643cc2ba07e04120d9bf47ba5)

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
OrangeAndGreen: I think in the latest design we aren't providing a manual logout option for Connect apps. Sessions can expire due to inactivity (in which case the page will automatically start a new session when the user returns), or they can be closed when the user changes to a different app (i.e. reviews learning, goes to a different opp, or loads a traditional app).

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
OrangeAndGreen: Good point to bring up, and it seems worth addressing a bit more in this spec even if we leave some of the details pending for decisions at implementation time.

I introduced an OpportunityHomeStateController class to handle state resolution for Opp Home, including fragment selection as well as auto-login actions and error handling (i.e. login fails due to bad network). There's also a new section addressing some of the plans for how that logic will work. [06f45f0](https://github.com/dimagi/commcare-android/pull/3776/commits/06f45f00194515a6989d3aadfcfca4bfa996d56c)

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
OrangeAndGreen: I think this has been cleared up in other changes but let me know if additional clarification should be added.

The spec now indicates that onActivityResult will be forwarded to the Coordinator, which then calls delegates that it knows are relevant (rather than generically calling all delegates).

Instance-state functionality is handled directly by delegates that require it, and doesn't require pass-through from the Activity or Coordinator.

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
OrangeAndGreen: Ah, nice callout. I modified the spec to have the Coordinator handle the logic for drift checks since it's pretty simple and already mostly contained by DriftHelper. [282d17f](https://github.com/dimagi/commcare-android/pull/3776/commits/282d17f9ba9804c97a7c3aa55f142152d28f3206)

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
OrangeAndGreen: I think this is better clarified now after several changes to the spec, so didn't make any additional changes.

The idea now is that the logic that used to live in doLoginLaunchChecksInOrder now lives in the Coordinator's runLaunchChecks function, which calls into delegates in the correct order for the specific checks to be performed. SessionLaunchDelegate owns three of the checks but other delegates own the others.

OrangeAndGreen:
> One other thing that would be good to preclude this work is documenting the current behaviour as it is for HomeActivity with unit tests so that if something functional changes as part of the refactor, tests are better equipped to flag that.

Good idea and something Clayton emphasized as well. I had Claude do an analysis and indeed the existing unit testing for home screen functionality is pretty sparse. The existing testing primarily goes through the page incidentally while testing other functionality. [7f2c2fe](https://github.com/dimagi/commcare-android/pull/3776/commits/7f2c2fe65c4bb8e095496d947cc7b9b596200ce5)

I added a section at the end of the doc detailing the existing state of testing and recommended pre-refactor tests to add. I'm thinking we should create a ticket to add that testing in so we can get it prioritized for next sprint before taking on the main refactoring work. @shubham1g5 Agreed?

conroy-ricketts (dismissed):

shubham1g5:
@OrangeAndGreen Hey, I am afraid that this spec has expanded a lot in verbosity and I am finding it quite impractical to go over it. Would you be willing to take a pass on it to remove any un-needed details, and try to concise it and focus on changes rather than existing behaviour. I think the spec is very heavy as well on code references and it would be nice to  try to simplify the text and make it more read-friendly. 

I understand that this might just be the nature of this spec as well and if you disagree with my comment, I can take a look as it is on it.

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
OrangeAndGreen: Forgot to address this one earlier! I added another section at the top comparing the composition vs inheritance approaches and advantages to each: [5974246](https://github.com/dimagi/commcare-android/pull/3776/commits/597424699350942ec291fc3da7bfd656062a923e)

docs/superpowers/specs/2026-06-22-opportunity-home-composition-design.md
```diff
@@ -0,0 +1,373 @@
...
+- Implementing `WithUIController` on the new activity.
+
+## Design approach: inheritance vs. composition
+
+The behavior this activity needs — sync, form entry, app updates, launch checks — lives today in an inheritance chain (`SessionAwareCommCareActivity` → `SyncCapableCommCareActivity` → `HomeScreenBaseActivity` → `StandardHomeActivity`), where the session gate is the common ancestor `SessionAwareCommCareActivity` and the behaviors layer on top of it. Two approaches can give `OpportunityHomeActivity` those behaviors: an inheritance approach that keeps the existing chain, and the composition approach the rest of the spec adopts.
+
+They are not variations on one axis. They divide on a single question — **must the activity seat an app and establish a session without leaving the activity** (the defining requirement in the Summary)? The inheritance approach answers no; composition answers yes.
+
+### Inheritance — two activities split by session state
```
shubham1g5, line 43: what's the specific thing that compostion approach achieves that negates the need of the 2 activities split like this here ? Assuming it's just the session management, think the best approach here would be to retain the inheritance and just implement the session management part with the composition approach so that we don't need the 2 activity split in inheritance and achieve the desired outcome with minimum changes here. Thoughts ?

OrangeAndGreen: I'm not sure I understand what you have in mind, but maybe there's something here you understand better than I do.

Or maybe there's a misunderstanding in the intended approach? With the suggested composition approach, the existing inheritance chain is preserved for traditional CommCare code (i.e. StandardHomeActivity still inherits from HomeScreenBaseActivity, SessionAwareActivity, etc.). But since that existing chain requires an active session at activity creation and we don't want to modify that behavior since various classes in that inheritance chain are relied on by multiple traditional activities, the new Opp Home needs to use the refactored session management via composition without relying on the traditional inheritance chain.

Does that match your understanding? If so, does it make sense to leave the inheritance chain as it is for existing code, while extracting the session management stuff so that it can be used by both that inheritance chain (mandatory) and the new functionality (optional)?

Otherwise, it seems we are stuck with two possibilities:
1. Modify the existing inheritance chain to make the active session optional, and be careful not to break any existing functionality on activities that rely on the inheritance chain
2. Use two activities to host Opp Home, one that requires an active session and implements the inheritance chain, and the other that shows the UI when there isn't a session

Put another way, I don't see how to have the inheritance chain preserve current behavior (requiring active session) for all existing pages that use it, while also supporting optional sessions for the new Opp Home. That's what the composition approach solves.

shubham1g5: > . But since that existing chain requires an active session at activity creation and we don't want to modify that behavior since various classes in that inheritance chain are relied on by multiple traditional activities

 I was thinking composition should get rid of that dependence given it will be implementing the `no session` as a result from the delegate methods to stop-gate any session dependent functionality in the activity along with callbacks like `onSessionAvailable(session)`/`onSessionLost` and we can simply mimik the behaviour in `SessionAwareCommCareActivity` using those callbacks (they both feel very similar except the action taken on session lost). Also think it would be very easy to modify `SessionAwareCommCareActivity` to pass in a parameter like `enforceSession` that can control whether to exit on no session, if needed. 


That said, think we should make sure one of the 2 outcomes below - 

1. No`HomeScreenBaseActivity` exists in the final structure at all and all home functionality is controlled by  `HomeActivityCoordinator`. The reasoning here is existence of `HomeScreenBaseActivity` signifies inheritance for implementing home functionality and it seems quite confusing to me if a new Home screen implementation doesn't need to extend from `HomeScreenBaseActivity`. 

2. Retain `HomeScreenBaseActivity` and make sure all new implementations of Home screen extends from it.  (Preferential as it would simplify the composition given we won't need to export everything into the delegates right away)

OrangeAndGreen: I think this may just be a difference in how much we're willing to modify traditional code with this work vs attempting to leave that code untouched. Looking at the two options you listed with respect to how much we want to change traditional code:
1. I prefer this option, and I think removing HomeScreenBaseActivity (as well as SyncCapableCommCareActivity) would be pretty easy to do as part of this work. They both become very thin after their current code is extracted out, and the only real reason for keeping them was to avoid modifying StandardHomeActivity.
2. I think this would be a more feasible route if we can make some modifications to the existing code to support optional gating (like the enforceSession flag you mentioned), but I'm worried that may be more complicated than it seems and could open up new issues in the traditional code if not handled well. Claude pointed out concerns with this as well:
> More importantly, the gate is not the only session assumption in the chain. Per your own §Problem and §Risks, SyncCapableCommCareActivity/HomeScreenBaseActivity read getCurrentSession() ambiently for sync, form entry, updates, and launch checks. Flipping enforceSession=false removes the top gate but leaves those ambient reads throwing SessionUnavailableException in the no-session state. To make an inherited activity tolerate no-session you'd have to thread the session through the entire chain — which is exactly the delegate work, now done inside the classes "various traditional activities rely on." So "very easy to modify" isn't accurate; it's the same work with a larger blast radius.

Are you okay with the first option? I can modify the spec to include removing HomeScreenBaseActivity and SyncCapableCommCareActivity.

shubham1g5: Think I am not understanding the concerns here as such, but happy to go with first option if that's easiest.

conroy-ricketts:
Hi @OrangeAndGreen, just confirming if this is either waiting for another round of review or can be closed?

OrangeAndGreen:
> Hi @OrangeAndGreen, just confirming if this is either waiting for another round of review or can be closed?

I think we're good to close this one, we're moving on to implementation tickets. @shubham1g5 Agreed?

shubham1g5:
yeah no objections on my end .

PR #3779, opened 2026-06-29
CCCT-2515 Add Reusable Semi-Circle Progress Bar

OrangeAndGreen:
## Technical Summary

Adds a reusable 180° semi-circle progress indicator. It is split into two pieces:

- `SemiCircleArcView` — a focused custom `View` that draws only the track + progress arcs (exposes a `progress` fraction, arc colors, and stroke width).
- `SemiCircleProgressBar` — a `ConstraintLayout` composite that overlays two centered `TextView`s (a "X of Y" value line and a description line) on the arc, so text centering/rendering is handled by the layout rather than canvas math.

The four colors (`progressColor`, `trackColor`, `valueTextColor`, `descriptionTextColor`) are mapped in one place via the `Widget.CommCare.SemiCircleProgressBar` style; `current`, `max`, and `descriptionText` are the per-instance content. The value line uses a localized format string (`semi_circle_progress_value_format`) translated across all supported languages.

Note that the active value is clipped between 0 and the max, i.e.:
* Inputs -1 and 40 result in "0 of 40" and the bar showing 0% filled
* Inputs 50 and 40 result in "40 of 40" and the bar showing 100% filled

Also, the max value is floored at 0 (to handle negative inputs) and will always show an empty progress bar when drawing with max=0. Since the current value is clipped between (0, max), the text will always read "0 of 0" when max=0.

This PR adds the component only — it is not yet wired into any production screen.

Demo of the new component (temporarily added to Learning Progress and hard-coded with some sample data):
Screenshot_20260701_160119

### Safety story

What gives me confidence:
- I ran the app and visually confirmed the component renders correctly (arc fill, centered text) on a temporary demo placement on the Learning Progress page — see the screenshot above.
- The diff is purely additive: new view classes, a new layout, and new resource entries (style, attrs, dimens, string + translations). No existing code paths or layouts are modified.
- Compiles cleanly and passes ktlint; resources link successfully.

[CCCT-2515]: https://dimagi.atlassian.net/browse/CCCT-2515?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


OrangeAndGreen:
## Suggested Review Order

- `app/src/org/commcare/views/connect/SemiCircleArcView.kt` — core arc drawing (measurement + the two `drawArc` calls); start here
- `app/src/org/commcare/views/connect/SemiCircleProgressBar.kt` — composite that wraps the arc and routes content/colors to its children
- `app/res/layout/view_semi_circle_progress_bar.xml` — layout the composite inflates (arc + centered text block)
- `app/res/values/attrs.xml` — the `SemiCircleProgressBar` styleable attributes
- `app/res/values/styles.xml` — `Widget.CommCare.SemiCircleProgressBar` style mapping the four colors
- `app/res/values/dimens.xml` — default stroke width and text sizes
- `app/res/values/strings.xml` — the `semi_circle_progress_value_format` value string
- `app/res/values-*/strings.xml` — translations of the value string (skim last)

app/src/org/commcare/views/connect/SemiCircleArcView.kt
```diff
@@ -0,0 +1,109 @@
...
+
+        override fun onMeasure(
+            widthMeasureSpec: Int,
+            heightMeasureSpec: Int,
+        ) {
+            val width =
+                when (MeasureSpec.getMode(widthMeasureSpec)) {
+                    MeasureSpec.EXACTLY, MeasureSpec.AT_MOST -> MeasureSpec.getSize(widthMeasureSpec)
+                    else -> resources.getDimension(R.dimen.semi_circle_progress_default_width).toInt()
+                }
```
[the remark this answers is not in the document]
OrangeAndGreen: [4a6bb3e](https://github.com/dimagi/commcare-android/pull/3779/commits/4a6bb3ef0b90e8cfa2c55ef44639bc7b00319f47)

conroy-ricketts:
Out of curiosity, was the design change intentional (i.e. using X of Y rather than X/Y)?

app/res/values/strings.xml
```diff
@@ -529,6 +529,7 @@
     <string name="connect_progress_delivery">Delivery</string>
     <string name="connect_progress_title">Delivery Progress</string>
     <string name="connect_progress_status">You have completed %d of %d visits.</string>
+    <string name="semi_circle_progress_value_format" cc:translatable="true">%1$d of %2$d</string>
```
conroy-ricketts, line 534: Can we name this `connect_progress_text` instead

And confirming that this needs to be `cc:translatable`?

app/res/layout/view_semi_circle_progress_bar.xml
```diff
@@ -0,0 +1,45 @@
...
+        app:layout_constraintTop_toTopOf="@id/semi_circle_arc"
+        app:layout_constraintVertical_bias="0.78">
+
+        <TextView
+            android:id="@+id/semi_circle_value_text"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:textSize="@dimen/semi_circle_progress_value_text_size"
+            android:textStyle="bold"
+            tools:text="25 of 40" />
```
conroy-ricketts, line 34: Was this tested in the Android Studio preview window? Trying to gauge how valuable `tools:text` is in this layout. We could maybe go without it

Image

app/res/values/attrs.xml
```diff
@@ -144,6 +144,19 @@
         <attr name="strokeWidth" format="dimension" />
     </declare-styleable>
 
+    <declare-styleable name="SemiCircleProgressBar">
```
conroy-ricketts, line 147: Curious how the team feels about prefixing these with `Connect`

app/res/values/dimens.xml
```diff
@@ -141,4 +141,9 @@
     <dimen name="connect_info_card_corner_radius">15dp</dimen>
     <dimen name="connect_info_card_elevation">4dp</dimen>
     <dimen name="connect_info_card_min_height">130dp</dimen>
+
+    <dimen name="semi_circle_progress_default_width">240dp</dimen>
+    <dimen name="semi_circle_progress_stroke_width">14dp</dimen>
+    <dimen name="semi_circle_progress_value_text_size">28sp</dimen>
+    <dimen name="semi_circle_progress_description_text_size">16sp</dimen>
```
conroy-ricketts, line 150: Similar comment as above, are these worth prefixing with `connect` like we do strings currently?

conroy-ricketts: Noting that there is a contrast with how it was defined for the Connect info card

OrangeAndGreen:
> Out of curiosity, was the design change intentional (i.e. using X of Y rather than X/Y)?

@conroy-ricketts Confirmed, the design changed slightly in the [final Figma](https://www.figma.com/design/4QLInPUv5wIfjfDGfrFESW/Opportunity-Home-Design-_-Ishwari?node-id=4298-31383&t=ssDOYiuMdg5isU9n-4) so I went with the latest from there

app/res/layout/view_semi_circle_progress_bar.xml
shubham1g5: Think it's harmless and generally provides a quick direction to developers looking at layout preview in Android Studio on whether the layout is behaving as expected.

app/res/values/attrs.xml
shubham1g5: IMO, we should keep the UI widgets open for usage across CommCare and Connect development.

app/res/values/strings.xml
shubham1g5: `cc:translatable` is applicable for all string that can be shown after an App Login , so think it's good to keep this as `translatable` but think this text needs to be defined per usage rather than being widget specific. That is for the widget we should just expose a Text View and let the screen level UI code set the exact string. This provides us better felxibility on managing translations but also to change the format of String or have different text formats at different places.

shubham1g5 (commented):
No other feedback from my side.

app/res/values/strings.xml
conroy-ricketts: That makes sense, agreed!

app/res/layout/view_semi_circle_progress_bar.xml
```diff
@@ -0,0 +1,45 @@
...
+
+        <TextView
+            android:id="@+id/semi_circle_description_text"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:textSize="@dimen/semi_circle_progress_description_text_size"
+            tools:text="total visits completed" />
```
Jignesh-dimagi, line 42: According to the final [Figma design](https://www.figma.com/design/4QLInPUv5wIfjfDGfrFESW/Opportunity-Home-Design-_-Ishwari?node-id=4298-31422&t=IHLhrGZDvNJxh432-0), it looks like semi_circle_description_text should start from the bottom of the SemiCircleArcView, with semi_circle_value_text positioned right above it. Currently, the view is centered, causing both text elements to display exactly in the middle of the SemiCircleArcView.

app/res/values/strings.xml
Jignesh-dimagi, line 534: Should this also include another string for total visits completed, which semi_circle_description_text can then use as its default value?

Jignesh-dimagi:
Looks awesome! Just left a few minor comments on the thread.

app/res/values/strings.xml
OrangeAndGreen: @shubham1g5 Are you suggesting to change the top row of text to make it customizable by the caller in addition to the "description" (second line, already customizable)?

My thought with having the first line driven by the "current" and "max" progress values is that it ensures consistency between the bar and the text, particularly with the data conditioning around negative values, values greater than max, or max <= 0.

app/res/layout/view_semi_circle_progress_bar.xml
OrangeAndGreen: Yeah I think this is to be expected at this level. The component doesn't render a useful preview since it requires the associated code in addition to the XML to fully render.

But when the component is hosted in another view everything renders correctly in Android Studio's Design preview. In testing I saw that in action while temporarily adding the view to the Learning Progress page.

app/res/values/attrs.xml
OrangeAndGreen: Seconding that, I think naming items to distinguish them for Connect made more sense early on but now we should think about future reuse and name things accordingly. This semi-circle progress bar could easily be used elsewhere in the app in the future

app/res/values/dimens.xml
OrangeAndGreen: As discussed in another comment, leaving the Connect-specific naming out for this one.

Your point about the Connect info card has me thinking though, in hindsight maybe that component could have been named more generically, like simply InfoHalfWidthCard instead of ConnectInfoHalfCard. Doesn't seem worth going back to change now and perhaps the layout for those cards is specific enough that they'll only be used in Connect screens, but generally this is something to keep in mind when implementing several more UI components that are coming with this UX redesign work.

app/res/layout/view_semi_circle_progress_bar.xml
OrangeAndGreen: Ah, great catch! Fixed, and I'll update the screenshot in the description as well. [4b5b2c7](https://github.com/dimagi/commcare-android/pull/3779/commits/4b5b2c7912f02278adcd93170db9bd56c263ba5f)

app/res/layout/view_semi_circle_progress_bar.xml
OrangeAndGreen: Screenshot showing the rendered preview when the progress bar is added to Learning Progress. Fun fact: you can change the values for current and max in the XML and watch the preview update.
image

app/res/values/strings.xml
OrangeAndGreen: Eventually yes, but since that text is specified by the owner and nothing is using this component yet I didn't add the string. It will be added in the Delivery Progress UI refactor work when we start using the new progress bar.

shubham1g5: I am mostly thinking about if in someplace else, we might want to show a different format for the same widget like instead of `5 of 10` , there might be places where we want to show a different formats like - 

`5 out of 10`
`5 / 10` 


But re-thinking, I agree with you that it's probably better for platform to control this and not have it as a customisable string i.e. if folks changes the translation for this one place, all usage of such widgets change their translations.

shubham1g5 (dismissed):

Jignesh-dimagi (dismissed):

app/res/layout/view_semi_circle_progress_bar.xml
conroy-ricketts: Aaah I see, at the time I wasn't thinking of the preview in other fragments/creens, and that's a cool tip!

app/res/values/attrs.xml
conroy-ricketts: Yeah that makes total sense to me

conroy-ricketts (dismissed):

shubham1g5 (approved):

shubham1g5 (approved):

shubham1g5:
@OrangeAndGreen Seems like this PR changed one of the screens `app/res/layout/fragment_connect_learning_progress.xml` , are you able to confirm ?

OrangeAndGreen:
> Seems like this PR changed one of the screens app/res/layout/fragment_connect_learning_progress.xml , are you able to confirm?

@shubham1g5 Argh, indeed! Here's a quick PR to remove that demo code, it shouldn't have been checked in: https://github.com/dimagi/commcare-android/pull/3817

PR #3781, opened 2026-06-30
Add Common Edge-Cases Reference Doc

conroy-ricketts:
## Technical Summary

Adds `docs/common-edge-cases.md` — a planning/spec checklist of recurring edge-case categories, seeded from real defects on this codebase. `AGENTS.md` gains a short section pointing to it so the checklist is consulted when planning or spec'ing new features. It's a living document, meant to grow as new recurring cases surface.

### Safety story

Docs-only change.

What gives confidence:
- No code or behavior is touched; the new file sits under `docs/` and is referenced by no build or runtime path.
- The `AGENTS.md` change is a single additive section.

Risks to review:
- None to runtime. The useful review here is on the *content* — whether the checklist categories are accurate and genuinely helpful during planning — rather than on code correctness.


docs/common-edge-cases.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Connectivity & device state
+
+- CommCare runs in **low-connectivity** settings: handle slow/failed network, partial sync, and
+  offline entry explicitly.
+- Consider low storage, low memory, and interrupted background work.
+
+## Localization
+
+- New user-facing strings must be added to **all supported locale files**, not just the default.
```
OrangeAndGreen: Curious how others feel, but I think I'd rather see this point directly in the CLAUDE.md. My thinking is that the points here are intended to be used at design/spec time, whereas this is an implementation point. Also, the points here focus on being edge cases, whereas this point really addresses common implementation advice

OrangeAndGreen (commented):
All in all I think this is in the right direction. But the initial list feels more like general design/planning guidance than specific edge-cases, and it seems like with points like these the list is likely to grow quite large over time. That being said, maybe that's okay? It isn't how I initially pictured the list but might be a better solution, and the important thing is whether it makes Claude work better for us.

Some thoughts to that point:
1. Having Claude run through a long list of design/planning considerations could cloud up context
2. Some of these points feel like the kind of general guidance that Claude is already good at considering on its own, so may not be adding much value
3. A long list of items feels difficult for humans to manage well, and if we let Claude manage the list on its own I worry about possible conflicts in the list, eventually leading to the doc just confusing Claude rather than helping it

conroy-ricketts:
@OrangeAndGreen Actually, I totally agree. I'm going to refine this, including dropping the point about strings because that should already be handled by Claude pretty well without directing it

docs/common-edge-cases.md
```diff
@@ -0,0 +1,51 @@
...
+- **`FLAG_ACTIVITY_REORDER_TO_FRONT` on a shared Activity class.** Home (`StandardHomeActivity` /
+  `RootMenuHomeActivity`) is one class shared across all apps, so an unconditional reorder can
+  surface a *different* app's stale Home. Only reorder when you know this app's instance is the one
+  in the task. ("No-op if no instance exists" is not true for shared classes.) Verify cross-app task
+  scenarios on-device.
+- **Pop the back stack on `onStop`, not immediately.** Popping a fragment / back-stack entry while
+  the host Activity is still foregrounded briefly shows the popped screen (a visible flash). Defer
+  the pop to an `onStop` lifecycle callback.
+
+## Session & authentication
```
shubham1g5: Think we should aim here to provide common guidelines for someone to consider when writing a spec with AI.

I think the whole document can be just a list of common things we often forget to include in a spec can just drill down to high level points like below without going into specifics as they will differ per spec/work item - 

- Specify back navigation for UI specs on new features
- Specify any user session and auth management as part of spec 
- Specify activity lifecycle management and what state we need to retain through configuration changes
- For UI changes, think about whether we need to support different form factors like phone and tablet etc vs just designing for a single form factor.

conroy-ricketts: I like the idea of providing guidelines for a human to consider, but also, more importantly, I'd want AI to catch these things on its own. So, my aim is to accomplish both at the same time.

Also flagging that there may be a conflict with Dave's suggestion (if I understand correctly).

I do agree with making these more high-level though!

conroy-ricketts: https://github.com/dimagi/commcare-android/commit/c808eb9113cef4372efd9275ae4644137a7c20d6

OrangeAndGreen (approved):

shubham1g5 (approved):

PR #3782, opened 2026-06-30
CCCT-2597 Add Responsive Rectangle Reticle Overlay View

Jignesh-dimagi:
## Technical Summary

Adds `RectangleOverlayView`, a self-contained custom `View` for the upcoming MUAC camera capture flow. It draws a centered rectangular reticle — a dim scrim around a clear center plus a high-contrast white-on-dark border — sized responsively as a uniform inset (`0.08 × width`) recomputed on size change.

### Safety story

What gives confidence:
- I verified rendering via a temporary preview harness on both an emulator and a physical Motorola device (Android 16): the reticle draws as a centered rectangle with equal margins, a correct dim scrim, and a border legible on both light and dark backgrounds.

This is just for reference to show how it looks.
Screenshot_20260630_201240

- The diff is narrow and additive — one new `View` plus two color/dimension resources; no existing code paths are touched.

[CCCT-2597]: https://dimagi.atlassian.net/browse/CCCT-2597?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/org/commcare/views/RectangleOverlayView.kt
+++ b/app/src/org/commcare/views/RectangleOverlayView.kt
@@ -6,7 +6,7 @@ import android.graphics.Paint
 import android.graphics.RectF
 import android.util.AttributeSet
 import android.view.View
-import androidx.core.content.ContextCompat
+import androidx.core.content.withStyledAttributes
 import org.commcare.dalvik.R
 
 /**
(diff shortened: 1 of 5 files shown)


Jignesh-dimagi:
## Suggested Review Order

- `app/src/org/commcare/views/RectangleOverlayView.kt` — the entire feature: sizing, scrim, and border drawing logic.
- `app/res/values/colors.xml` — adds `black_80`; the outline color the view references.
- `app/res/values/dimens.xml` — adds the reticle stroke/outline widths the view references.

OrangeAndGreen (dismissed):

conroy-ricketts (dismissed):

shubham1g5:
> Sized responsively as a uniform inset (0.08 × width) recomputed on size change.

@Jignesh-dimagi Was the idea behind providing a inset `0.08 × width` is so that the inset changes with the screen width ? I think the value `0.08` here needs to be exported to the xml for the view itself (i.e  it should not be a fixed value and different views should be able to set different values here)

app/src/org/commcare/views/RectangleOverlayView.kt
```diff
@@ -0,0 +1,80 @@
...
+        override fun onSizeChanged(
+            w: Int,
+            h: Int,
+            oldw: Int,
+            oldh: Int,
+        ) {
+            super.onSizeChanged(w, h, oldw, oldh)
+            if (w > 0 && h > 0) {
+                val margin = RETICLE_MARGIN_RATIO * w
+                reticleRect = RectF(margin, margin, w - margin, h - margin)
```
shubham1g5: The margin applied uniformly to all four sides of the RectF. On a square view this is perfectly symmetric. But on a portrait camera preview (e.g. 9:16 aspect), the horizontal insets will be 8% of width (~small), while the vertical insets will also be 8% of width — which is a proportionally smaller fraction of the height. Thus the rectangle overlay will not be centered in a visually symmetric way.

                                                                                                                                                                         
Suggestion: Switch to minOf(w, h) for true visual symmetry.

shubham1g5 (changes requested):

app/src/org/commcare/views/RectangleOverlayView.kt
Jignesh-dimagi: We assumed the activity would only be in portrait mode, so this scenario shouldn't occur. However, I’ve still implemented the suggested changes for better protection. https://github.com/dimagi/commcare-android/pull/3782/changes/00c134c0e0308e1c9078d160146a8b07238eb68c

Jignesh-dimagi:
> > Sized responsively as a uniform inset (0.08 × width) recomputed on size change.
> 
> @Jignesh-dimagi Was the idea behind providing a inset `0.08 × width` is so that the inset changes with the screen width ? I think the value `0.08` here needs to be exported to the xml for the view itself (i.e it should not be a fixed value and different views should be able to set different values here)

Although configuring the reticle was [out of scope](https://docs.google.com/document/d/1n0g37HhqfSxrmKoYi-4aBd3PLG4vYvotm-OlpblL1ew/edit?tab=t.0) in the design doc, I followed the feedback here and made both its size and other properties configurable.

app/src/org/commcare/views/RectangleOverlayView.kt
```diff
@@ -75,6 +81,6 @@ class RectangleOverlayView
         }
 
         companion object {
-            const val RETICLE_MARGIN_RATIO = 0.08f
+            const val DEFAULT_RETICLE_MARGIN_RATIO = 0.08f
```
shubham1g5, line 84: can we increase the default here to be > 0.4f , my sense is the use cases of this feature will most probably like a smaller box given the things we want to center on are usually small.

shubham1g5:
> Although configuring the reticle was [out of scope](https://docs.google.com/document/d/1n0g37HhqfSxrmKoYi-4aBd3PLG4vYvotm-OlpblL1ew/edit?tab=t.0) in the design doc

Clarifying here that the out of scope element is ability to configure the size from HQ and not the widget itself.

shubham1g5 (approved):
left one non-blocking suggestion

Jignesh-dimagi:
> > Although configuring the reticle was [out of scope](https://docs.google.com/document/d/1n0g37HhqfSxrmKoYi-4aBd3PLG4vYvotm-OlpblL1ew/edit?tab=t.0) in the design doc
> 
> Clarifying here that the out of scope element is ability to configure the size from HQ and not the widget itself.

That’s correct. I was hesitant to invest time into it initially, since it will likely never be reconfigured once it's coded.

app/src/org/commcare/views/RectangleOverlayView.kt
Jignesh-dimagi: My only concern was for smaller screen sizes. I am keeping the default at 0.08f, but we can test different sizes against actual camera preview use cases. I will post some comparison images in that PR.

conroy-ricketts (approved):

OrangeAndGreen (approved):

PR #3784, opened 2026-07-01
CCCT-2459 PersonalId Profile Screen Phase Four

conroy-ricketts:
## Product Description

Phase 4 of the Manage Profile feature. 

Adds the "Forget PersonalID Account" action to the Manage Profile screen. 

The screen's drawer entry point is still hidden until Phase 6, so this is not yet reachable by users.

https://github.com/user-attachments/assets/d0498d1b-9ad8-48f4-b29e-815493d442be

Note: I will demo this to the wider team once we complete the Edit screen.

## Technical Summary

- The host `PersonalIdProfileActivity` owns the terminal transition (`forgetPersonalIdAccount()`).
- The confirmation uses the lifecycle-managed `DialogController` path (`showAlertDialog`), matching sibling PersonalID fragments.
- Drops the now-unused analytics `reason` argument from `reportPersonalIdAccountForgotten()`.

### Safety story

What gives confidence:
- I exercised the forget flow in the app (temporarily making the screen reachable): the confirmation dialog, the wipe, and the restart to the dispatch screen all behaved as expected.
- The diff is narrow, and the feature is dark-launched — the drawer entry point stays `View.GONE` until Phase 6, so there is no user-facing exposure in this release.
- `assembleCommcareDebug`, ktlint, and the existing profile unit tests pass locally.

Risks to review:
- Forget is destructive and irreversible (full local session wipe: DB, tokens, prefs). The confirmation dialog is the only guard.
- The other three `forgetUser()` callers (login menu, setup menu, global DB error) now emit the same analytics event with no way to distinguish the trigger — intentional, but worth confirming analytics consumers don't rely on the old `reason` param.
- No automated test covers the new forget wiring; confidence there rests on manual exercise and the narrow diff.

[CCCT-2459]: https://dimagi.atlassian.net/browse/CCCT-2459?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/org/commcare/connect/PersonalIdManager.java
+++ b/app/src/org/commcare/connect/PersonalIdManager.java
@@ -172,9 +172,9 @@ public class PersonalIdManager {
         }
     }
 
-    public void forgetUser(String reason) {
+    public void forgetUser() {
         if (ConnectDatabaseHelper.dbExists()) {
-            FirebaseAnalyticsUtil.reportPersonalIdAccountForgotten(reason);
+            FirebaseAnalyticsUtil.reportPersonalIdAccountForgotten();
         }
         ConnectUserDatabaseUtil.forgetUser();
         personalIdSatus = PersonalIdStatus.NotIntroduced;
--- a/app/src/org/commcare/connect/PersonalIdManager.java
+++ b/app/src/org/commcare/connect/PersonalIdManager.java
@@ -172,9 +172,9 @@ public class PersonalIdManager {
         }
     }
 
-    public void forgetUser() {
+    public void forgetUser(String reason) {
         if (ConnectDatabaseHelper.dbExists()) {
-            FirebaseAnalyticsUtil.reportPersonalIdAccountForgotten();
+            FirebaseAnalyticsUtil.reportPersonalIdAccountForgotten(reason);
         }
         ConnectUserDatabaseUtil.forgetUser();
         personalIdSatus = PersonalIdStatus.NotIntroduced;
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -71,6 +71,8 @@ These are published publicly on Playstore, Github Releases and CommCare Forums
 - Fixed the back arrow on the camera capture screen so it correctly returns to the previous screen
 - Fixed an issue where Connect messages opened from a notification could be viewed without completing the unlock prompt
 - Fixed PersonalID app linking so a failed fingerprint scan no longer skips you past the login screen; you can retry the unlock and linking completes once it succeeds
+- Fixed a crash caused by tapping a payment unit row more than once on the Connect delivery progress screen
+- Fixed a crash that could occur when leaving a Connect opportunity's intro screen before its Start Learning request finished
 
 
 ### QA Notes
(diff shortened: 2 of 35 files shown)


conroy-ricketts:
## Suggested Review Order

- `app/src/org/commcare/google/services/analytics/FirebaseAnalyticsUtil.java` — the forget event loses its `reason` param; start here to see the analytics contract change.
- `app/src/org/commcare/google/services/analytics/AnalyticsParamValue.java` — the three now-dead param constants removed.
- `app/src/org/commcare/connect/PersonalIdManager.java` — `forgetUser()` drops the `reason` argument.
- `app/src/org/commcare/activities/LoginActivity.java`, `CommCareSetupActivity.java`, `connect/database/ConnectDatabaseHelper.java` — the three call sites updated to the no-arg signature.
- `app/src/org/commcare/personalId/profile/PersonalIdProfileActivity.kt` — new `forgetPersonalIdAccount()` owning the wipe + restart to Dispatch.
- `app/src/org/commcare/personalId/profile/PersonalIdProfileFragment.kt` — wires the button to the confirmation dialog delegating to the activity.

app/src/org/commcare/connect/PersonalIdManager.java
```diff
@@ -172,9 +172,9 @@ public void handleFinishedActivity(CommCareActivity<?> activity, int resultCode)
         }
     }
 
-    public void forgetUser(String reason) {
```
shubham1g5, line 175: think `reason` should still be retained to differential between user triggered vs auto-forget due to a global error.

shubham1g5 (changes requested):

app/src/org/commcare/connect/PersonalIdManager.java
conroy-ricketts: That makes sense to me

Dropping the `reason` argument was decided and agreed on during one of our grooming meetings, so I'm curious to hear thoughts from the wider team before adding it back in 

cc @OrangeAndGreen @Jignesh-dimagi

Jignesh-dimagi: @conroy-ricketts 
A user getting removed from using 'Connect' is a major event since we're essentially losing them at that point. It would be really valuable to track the root cause behind this. Apologies if we previously decided to remove this tracking, but it might be worth reconsidering.

Jignesh-dimagi:
I don't have any additional comments. Looks good.

app/src/org/commcare/connect/PersonalIdManager.java
conroy-ricketts: Cool, wanted to make sure we're aligned on this. Tweak coming very soon!

conroy-ricketts: https://github.com/dimagi/commcare-android/commit/70d61e599e785c8616dfabaf99045a489d6355ab

shubham1g5 (approved):

OrangeAndGreen (approved):

PR #3786, opened 2026-07-02
CI-766 Fix Connect Delivery Progress And Job Intro Crashes

conroy-ricketts:
## Product Description

Fixes two Connect crashes reported on the Readers project:

- Tapping a payment-unit row on the delivery progress screen more than once in quick succession (e.g. a two-finger tap, or taps queued behind a busy main thread) crashed the app.
- Leaving a brand new opportunity's intro screen while its Start Learning request was still in flight crashed the app when the response arrived.

## Technical Summary

The ticket only reported that the app crashes for Readers project users starting June 28, so I queried Firebase crash data in BigQuery — crashes since June 25 (a day before the Readers project started just to be safe), filtered to the project's app/domain — to surface likely culprits; this PR fixes two of them. I highly suspect that the crash shown in the video on the ticket is traditional CommCare. But, the two Connect related crashes I saw were quick, easy wins, so I decided to go ahead and raise a PR for those 2 even though they may not be what's causing the crash in the video on the ticket.

Both fixes guard race conditions:

- Delivery progress row taps only navigate if the NavController is still on the delivery progress destination, and the controller is resolved through the fragment instead of the (possibly already detached) row view.
- The Start Learning API callbacks return early when the fragment is detached or its view is destroyed, instead of touching the activity and view.

For easy reference, here are links to the 2 crashes in Firebase that this PR fixes:
- [Crash 1](https://console.firebase.google.com/project/commcare-a57e4/crashlytics/app/android:org.commcare.dalvik/issues/a0a29bf69fbe21450c8b93a43931f146?time=90d&types=crash&sessionEventKey=6A46A9D201B300011DEF2278FB4198BE_2236443423138947080)
- [Crash 2](https://console.firebase.google.com/project/commcare-a57e4/crashlytics/app/android:org.commcare.dalvik/issues/b4fccd14c4005d2afcab1c54b0277c07?time=90d&types=crash&sessionEventKey=6A4608FF003B0001558E1CDFBBBC3AF6_2236353040060775135)

### Safety story

What gives confidence:

- I reproduced both crashes on a physical device (using temporary debug code to force the race timing) and verified the same steps no longer crash with these fixes applied.
- Both changes are narrow guards on existing code paths that only take effect in the raced conditions.

Risks to review:

- Neither flow has automated test coverage; verification was manual.
- When a guard triggers, the tap or callback is dropped silently (no navigation or error UI). The screen is gone or has already navigated in those cases, but it is a behavior change in those races.

[CI-766]: https://dimagi.atlassian.net/browse/CI-766?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -25,6 +25,8 @@ along with the public release notes above
 
 - Fix bug causing form widgets to lose focus when a Combobox dropdown on the same screen is re-validated.
 - Fixed a crash that could occur when navigating back from a case list that uses the GPS `here()` function
+- Fixed a crash caused by tapping a payment unit row more than once on the Connect delivery progress screen
+- Fixed a crash that could occur when leaving a Connect opportunity's intro screen before its Start Learning request finished
 
 ### QA Notes
 <!--
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -25,8 +25,6 @@ along with the public release notes above
 
 - Fix bug causing form widgets to lose focus when a Combobox dropdown on the same screen is re-validated.
 - Fixed a crash that could occur when navigating back from a case list that uses the GPS `here()` function
-- Fixed a crash caused by tapping a payment unit row more than once on the Connect delivery progress screen
-- Fixed a crash that could occur when leaving a Connect opportunity's intro screen before its Start Learning request finished
 
 ### QA Notes
 <!--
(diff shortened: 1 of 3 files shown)


conroy-ricketts:
## Suggested Review Order

- `app/src/org/commcare/fragments/connect/ConnectDeliveryProgressDeliveryFragment.java` — the crash from the ticket's video; destination guard + fragment-based NavController lookup
- `app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.java` — lifecycle guards on the Start Learning API callbacks
- `RELEASES.md` — release and QA notes

shubham1g5:
> so I queried Firebase crash data in BigQuery — crashes since June 25 (a day before the Readers project started just to be safe), filtered to the project's app/domain — to surface likely culprits; this PR fixes two of them.

I am only seeing 2 crashes in firebase in last week for this project - 

Screenshot 2026-07-02 at 9 10 34 PM


Are you seeing a different outcome in BigQuery ?

shubham1g5 (approved):

conroy-ricketts:
> Are you seeing a different outcome in BigQuery ?

@shubham1g5 

Yeah in BigQuery I saw those same 2 crashes you saw, plus an additional 2 crashes. I confirmed that all 4 crashes have the same domain value and/or app name value for the Readers project for at least some of the reported events

PR #3787, opened 2026-07-02
Audio quality profiles behind an Dev gated App Preference

ctsims:
## Product Description
Adds an app-specific preference to control the quality of Audio Recordings in the app, including a new "Balanced" recording profile which decreases voice distortion.

## Technical Summary
[Motivation and change scope / spec](https://docs.google.com/document/d/1dsozUHyEIMFrfg3zXFM6MIAkl-1bp9Y16KkNFSgOATs/edit?tab=t.g3ve57aqbc5o#bookmark=id.jipsfzcxkt0g)

Our previous audio recording configuration was hyper-focused on minimizing file size at the cost of fairly significant distortion of voice waveforms. We want to use those recordings for processing, so this change lets apps pick a recording mode which accepts more background noise and slightly higher file sizes.

New recording settings are gated by the new app preference (`cc-audio-quality-profile` set to `balanced`):
  * Output audio files receive an `m4a` extension (Note: Technically this is not correct for the Opus encoded files, but shifting the extension adds significant complexity, and audio tools are extremely tolerant for containers and encoders). 
  * Sets an explicit 24k bitrate for all modes
  * Shifts to the `VOICE_RECOGNITION` audio source, which maintains voice waveforms more cleanly than the processing heavy `VOICE_COMMUNICATIONS` source.
  * Uses a preference order of encoders: Opus, HE_AAC, then AAC based on availability.

Existing audio recording settings (or apps where the new preference is set to 'default' or 'smallest') will be unaffected by the change.

### Safety story
I confirmed all codepaths by manually testing on a real Android device and simulating the lack of availability of individual codecs or Android versions. 

I confirmed that the existing recording formats aren't affected by the new preference, and remain `.mp3` files with their existing encoding.

This change affects a very limited set of users (people using audio recording) and very limited codepaths, making it appropriate and acceptable for a point release.

In a minor release, we should move the setting to a public one and add this to the HQ available app preferences.

### Automated test coverage
This code is very close to the device, so there is poor existing coverage.

## Labels and Review

### DIFF
--- a/app/res/values/strings.xml
+++ b/app/res/values/strings.xml
@@ -527,6 +527,16 @@
         <item>Unsent</item>
         <item>Saved</item>
     </string-array>
+    <string-array name="pref_audio_quality_profile_vals">
+        <item>smallest</item>
+        <item>balanced</item>
+        <item>default</item>
+    </string-array>
+    <string-array name="pref_audio_quality_profile_labels">
+        <item>Smallest</item>
+        <item>Balanced</item>
+        <item>System Default</item>
+    </string-array>
     <string-array name="pref_password_show_options">
         <item>always_hidden</item>
         <item>default_show</item>
--- a/app/res/xml/preferences_developer.xml
+++ b/app/res/xml/preferences_developer.xml
@@ -93,6 +93,13 @@
         android:entryValues="@array/pref_load_form_payload_as_vals"
         android:key="cc-form-payload-status"
         android:title="Load form payload status"/>
+    <androidx.preference.ListPreference
+        android:defaultValue="default"
+        android:enabled="true"
+        android:entries="@array/pref_audio_quality_profile_labels"
+        android:entryValues="@array/pref_audio_quality_profile_vals"
+        android:key="cc-audio-quality-profile"
+        android:title="Default Audio Quality"/>
     <org.commcare.preferences.FilePreference
         android:defaultValue=""
         android:enabled="true"
--- a/app/src/org/commcare/views/widgets/AudioRecordingHelper.java
+++ b/app/src/org/commcare/views/widgets/AudioRecordingHelper.java
@@ -2,26 +2,55 @@ package org.commcare.views.widgets;
 
 import android.media.MediaCodecInfo;
 import android.media.MediaCodecList;
+import android.media.MediaFormat;
 import android.media.MediaRecorder;
 import android.os.Build;
 
+import org.commcare.preferences.DeveloperPreferences;
+import org.commcare.preferences.PrefValues;
 import org.commcare.util.LogTypes;
 import org.javarosa.core.services.Logger;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
 
 import static android.media.MediaFormat.MIMETYPE_AUDIO_AAC;
 
 public class AudioRecordingHelper {
     private static final int HEAAC_SAMPLE_RATE = 44100;
     private static final int AMRNB_SAMPLE_RATE = 8000;
+    private static final int OPUS_SAMPLE_RATE = 48000;
+    private static final int BALANCED_BIT_RATE = 24000;
 
-    public MediaRecorder setupRecorder(String fileName) {
+    static final class EncoderProfile {
+        final int encoder, samplerate, bitrate, container;
+        public EncoderProfile(int encoder, int samplerate) {
+            this(encoder, samplerate, -1);
+        }
+        public EncoderProfile(int encoder, int samplerate, int bitrate) {
+            this(encoder, samplerate, bitrate, MediaRecorder.OutputFormat.MPEG_4);
+        }
+        public EncoderProfile(int encoder, int samplerate, int bitrate, int container) {
+            this.encoder = encoder;
+            this.samplerate = samplerate;
+            this.bitrate = bitrate;
+            this.container = container;
+        }
+    }
+
+
+    public MediaRecorder setupRecorder(String fileName, String profile) {
         MediaRecorder recorder = new MediaRecorder();
 
-        boolean isHeAacSupported = isHeAacEncoderSupported();
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+        EncoderProfile encoder = filterBestSupportedEncoder(profile);
+
+        if (PrefValues.AUDIO_QUALITY_SMALLEST.equals(profile) && Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
             recorder.setAudioSource(MediaRecorder.AudioSource.VOICE_COMMUNICATION);
+        } else if (PrefValues.AUDIO_QUALITY_BALANCED.equals(profile)) {
+            recorder.setAudioSource(MediaRecorder.AudioSource.VOICE_RECOGNITION);
         } else {
             recorder.setAudioSource(MediaRecorder.AudioSource.MIC);
         }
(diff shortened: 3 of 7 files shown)


app/src/org/commcare/views/widgets/AudioRecordingHelper.java
```diff
@@ -52,7 +83,28 @@ public MediaRecorder setupRecorder(String fileName) {
...
+
+        List<EncoderProfile> encoderPreference = new ArrayList<>();
+
+        if (PrefValues.AUDIO_QUALITY_BALANCED.equals(profileKey)) {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+                encoderPreference.add(new EncoderProfile(MediaRecorder.AudioEncoder.OPUS,
+                        OPUS_SAMPLE_RATE, BALANCED_BIT_RATE, MediaRecorder.OutputFormat.OGG));
+            }
+            encoderPreference.add(new EncoderProfile(MediaRecorder.AudioEncoder.HE_AAC, HEAAC_SAMPLE_RATE, BALANCED_BIT_RATE));
+            encoderPreference.add(new EncoderProfile(MediaRecorder.AudioEncoder.AAC, HEAAC_SAMPLE_RATE, BALANCED_BIT_RATE));
```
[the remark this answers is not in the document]
ctsims: Noting that this was a known limitation as outlined in the PR, and is no different than the situation today (we are already misusing file extensions, as none of our containers/encodings are mp3 files right now). Will rely on human feedback to decide if that stated risks are acceptable.

app/res/xml/preferences_developer.xml
```diff
@@ -93,6 +93,13 @@
         android:entryValues="@array/pref_load_form_payload_as_vals"
         android:key="cc-form-payload-status"
         android:title="Load form payload status"/>
+    <androidx.preference.ListPreference
```
shubham1g5, line 96: nit: Is rationale of introducing this as a dev setting is to ease testing this for internal team ? I would assume this setting be controlled mostly by HQ app settings in real projects.

app/res/values/strings.xml
```diff
@@ -527,6 +527,16 @@
         <item>Unsent</item>
         <item>Saved</item>
     </string-array>
+    <string-array name="pref_audio_quality_profile_vals">
```
shubham1g5, line 530: nit: Why do we want to add a explicit `default` given internally it's same as `smallest` today ? I would think future migration of default profile from "smallest" to "balanced" can be managed with similar complexity without explicity providing it as an option in setting.

app/src/org/commcare/views/widgets/RecordingFragment.java
```diff
@@ -116,7 +119,13 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
     }
 
     private void initAudioFile() {
-        fileName = CommCareApplication.instance().getAndroidFsTemp() + new Date().getTime() + FILE_EXT;
+        String extension;
+        if (DeveloperPreferences.getAudioQualityProfile().equals(PrefValues.AUDIO_QUALITY_BALANCED)) {
+            extension = FILE_EXT_BALANCED;
```
shubham1g5, line 124: checking if the extension doesn't need to be determined by the encoding profile we ended up using based on `AudioRecordingHelper.setupRecorder` ?

shubham1g5 (commented):
Little worried on extensions being determined by the preference rather than encoder, rest are non-blocking comments.

app/res/xml/preferences_developer.xml
ctsims: Just following the standard for other settings that are unlikely to be used locally. Agreed that my expectation is that this will be set on the HQ side for the project. I believe that all (or most) hq app settings have a corresponding way to control locally, with how exposed they are being a matter of practicality.

The goal here is to make local testing of formats possible if a specific device isn't working well, since audio recording is very device specific, which wouldn't be an expected user facing action, but would be impossible to do device-by-device remotely.

app/res/values/strings.xml
ctsims: Intended to be a hedge for rollout once this makes it to HQ. 

I think the current 'smallest' behavior is a little problematic and unlikely to be the best standard moving forward, but also don't' want to change functionality underneath people. 

If we made smallest the default without an explicit fallback, when we roll this out on the HQ side it'll start setting 'smallest' as the default, locking that in for everyone making new builds. Then if we wanted to change what we thought the best option was, those devices would still be on smallest.

Agreed that the pattern is a little new/awkward, though, and not married to it at all if we think it's too complex.

app/src/org/commcare/views/widgets/RecordingFragment.java
ctsims: So the answer to your question is 'yes, it really should be' but I didn't make that change here for three reasons.

1 - That change will make the scope/blast radius of this much larger, because it will require changing the recorder to have a non-deterministic file. Since the recorder has kind of a funny pattern where it expects the file to be where it requested upfront, that seems like a lot more risk to introduce in a hotfix.
2 - Our current file extension usage is also totally wrong. We record either AAC/mp4 audio or we record AMR audio, neither of which is correct for a `.mp3` file, so we're not really regressing anything by not carefully tuning the extension.
3 - every platform I tested works with the new files in `.m4a`, but some don't work with `.mp3`. I would have changed everything to `.m4a` (which does more accurately reflect our other audio files), but I didn't think it was worth the risk of changing prior behavior out from under users in a hotfix.

app/res/xml/preferences_developer.xml
shubham1g5: > I believe that all (or most) hq app settings have a corresponding way to control locally, with how exposed they are being a matter of practicality.

Think this is not true today. 

> The goal here is to make local testing of formats possible if a specific device isn't working well, since audio recording is very device specific, which wouldn't be an expected user facing action, but would be impossible to do device-by-device remotely.

that makes sense, Feels like it should probably therefore lie in Advanced setting as a long term plan for this setting (as you mentioned in PR as well)

app/res/values/strings.xml
shubham1g5: Think it might be simpler to Keep the options as "Smallest (default)" and "Balanced" but happy to park that for future and think can be combined with the move to make this setting more user facing

app/src/org/commcare/views/widgets/RecordingFragment.java
shubham1g5: Makes sense, Agree it's not worth to fix this in hotfixing given our existing code is itself not geat here.

shubham1g5 (approved):
It would be nice to note remaining action items from the PR in the scope doc so that they doesn't get buried with the PR.

avazirna (approved):

app/src/org/commcare/views/widgets/RecordingFragment.java
avazirna: > 2 - Our current file extension usage is also totally wrong. We record either AAC/mp4 audio or we record AMR audio, neither of which is correct for a .mp3 file, so we're not really regressing anything by not carefully tuning the extension.

I recently came across this issue and made a note to address it. I've created a ticket for it and will update it to account for these changes.

app/res/xml/preferences_developer.xml
ctsims: > Think this is not true today.

Ah, interesting. I thought the default pattern was for configurable settings to go into [Developer Options](https://dimagi.atlassian.net/wiki/spaces/GS/pages/2144315875/CommCare+Developer+Options+-+Internal) at the very least, but generally the application settings. I think this one should still have a local config for the device reason, but will consider more strongly next time around whether to expose anything locally.

app/res/values/strings.xml
ctsims: Yeah, I think good to work through when public, since it's possible we'll want to also be making these options available in other contexts too (like per-question).

ctsims:
> It would be nice to note remaining action items from the PR in the scope doc so that they doesn't get buried with the PR.

Agreed, this makes sense to me. Updated the [Phase 2 Spec](https://docs.google.com/document/d/1dsozUHyEIMFrfg3zXFM6MIAkl-1bp9Y16KkNFSgOATs/edit?tab=t.6caomet780h1) with next steps, since I think there'll be other mobile work in Phase 2 as well.

PR #3788, opened 2026-07-03
CCCT-2598 Add CameraOverlayActivity For Rectangle Reticle Capture

Jignesh-dimagi:
## Technical Summary

Adds `CameraOverlayActivity`, a new `BaseCameraActivity` subclass (base extracted in CCCT-2596) for the MUAC-style capture flow:

- Back camera + CameraX `ImageCapture` at `CAPTURE_MODE_MAXIMIZE_QUALITY`; writes the full-resolution frame to a caller-supplied output URI via a content-resolver `OutputStream` (no crop).
- A static `RectangleOverlayView` reticle (added in CCCT-2597) as a framing guide only — it is never drawn into the saved image.
- Manual shutter with in-flight progress and a re-entry guard; capture errors are non-fatal and retryable (toast + re-enable shutter) instead of closing the screen.
- Migrates the shared camera error/permission strings from `android_translatable_strings.txt` to `strings.xml` (`cc:translatable`), collapsing `BaseCameraActivity.logErrorAndExit` to a single resolved-`String` overload; `MicroImageActivity` updated to match.

In accordance with the [design document decision](https://docs.google.com/document/d/1n0g37HhqfSxrmKoYi-4aBd3PLG4vYvotm-OlpblL1ew/edit?tab=t.0#heading=h.uaeddr5838vv) to use independent scaling factors for height and width, the implementation has been updated to adapt to this change. The scaling factor includes a minimum threshold to guarantee that the target object is safely captured inside the reticle.

Please find the screenshots attached below. Additionally, since the landscape mode layout isn't complex, I was able to implement it without creating a separate landscape XML layout file.

Height / width adjusted as per comment:

Portrait - 5% increase in height and 5% decrease in width

Landscape - 5% increase in height only

### Safety story

What gives confidence:
- I ran the overlay screen on two emulators (including 1280×720) and a real Motorola API 36 device: the reticle renders correctly, the shutter captures, and the in-flight progress shows.
- 8 unit tests cover the activity's behavior (see below).

### Automated test coverage

`CameraOverlayActivityTest` (Robolectric, 8 tests): the `getIntent` factory contract, capturing-state UI toggling (shutter enabled/alpha + progress visibility), non-fatal error handling, and the `captureImage` branches — delegates to `ImageCapture.takePicture`, re-entry guard while a capture is in flight, no-op when the camera isn't ready, and the fail-fast NPE when the launch intent omits the output URI.

[CCCT-2598]: https://dimagi.atlassian.net/browse/CCCT-2598?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Jignesh-dimagi:
## Suggested Review Order

- `app/src/org/commcare/activities/camera/BaseCameraActivity.kt` — shared base; the `logErrorAndExit` signature change underpins the rest
- `app/src/org/commcare/activities/camera/CameraOverlayActivity.kt` — the new activity, core of the PR
- `app/res/layout/camera_overlay_activity.xml` — layout it inflates (view ids, reticle, shutter, progress)
- `app/res/drawable/bg_camera_shutter_button.xml` — shutter drawable referenced by the layout
- `app/src/org/commcare/activities/camera/MicroImageActivity.java` — sibling subclass updated for the string/`logErrorAndExit` change
- `app/res/values/strings.xml` — migrated/new user-facing strings
- `app/assets/locales/android_translatable_strings.txt` — dead/migrated keys removed
- `app/AndroidManifest.xml` — registers the new activity
- `app/unit-tests/src/org/commcare/activities/camera/CameraOverlayActivityTest.kt` — tests documenting the behavior

conroy-ricketts:
I like the height of `reticleMarginRatio = 0.3` (so it doesn't overlap with the shutter button) and the width of `reticleMarginRatio = 0.2` (so it doesn't appear too small)

shubham1g5:
Also think we need to define different height and width ratios here. The box get too narrow for 0.3. and 0.4, maybe we should define minmum width/height of the box as well so that any aspect ratio doesn't make the box dimension smaller than a min threshold

app/src/org/commcare/activities/camera/CameraOverlayActivity.kt
```diff
@@ -0,0 +1,153 @@
...
+    override fun getTargetResolution(): Size = PREVIEW_TARGET_RESOLUTION
+
+    override fun buildCaptureUseCase(
+        targetResolution: Size?,
+        targetRotation: Int,
+    ): UseCase {
+        val capture =
+            ImageCapture
+                .Builder()
+                .setCaptureMode(ImageCapture.CAPTURE_MODE_MAXIMIZE_QUALITY)
```
conroy-ricketts, line 55: Have we considered using `CAPTURE_MODE_MINIMIZE_LATENCY` here?

Curious about thoughts on that

app/src/org/commcare/activities/camera/CameraOverlayActivity.kt
```diff
@@ -0,0 +1,153 @@
...
+     * Reports a non-fatal capture failure: logs it, toasts the user, and re-enables the shutter so
+     * the capture can be retried without leaving the screen.
+     */
+    @VisibleForTesting
+    internal fun handleCaptureError(
+        logMessage: String,
+        e: Throwable?,
+    ) {
+        if (e == null) {
+            Logger.log(LogTypes.TYPE_EXCEPTION, logMessage)
```
conroy-ricketts: Thoughts on creating a new custom exception for logging if it's null, and using `Logger.exception()` for all cases?

app/src/org/commcare/activities/camera/CameraOverlayActivity.kt
```diff
@@ -0,0 +1,153 @@
...
+    private fun OutputStream.closeQuietly() {
+        try {
+            close()
+        } catch (e: IOException) {
+            Logger.exception("Failed to close output stream after image capture", e)
+        }
+    }
+
+    private fun intentOutputUri(): Uri? =
+        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.TIRAMISU) {
```
conroy-ricketts: Nit: can we import `android.os.Build`

app/unit-tests/src/org/commcare/activities/camera/CameraOverlayActivityTest.kt
```diff
@@ -0,0 +1,141 @@
...
+
+    @Test
+    fun `getIntent targets CameraOverlayActivity and carries the output uri`() {
+        val uri = Uri.parse("content://org.commcare.dalvik.fileprovider/external/output.jpg")
+
+        val intent =
+            CameraOverlayActivity.getIntent(ApplicationProvider.getApplicationContext(), uri)
+
+        assertEquals(CameraOverlayActivity::class.java.name, intent.component?.className)
+        assertEquals(uri, intent.getParcelableExtra<Uri>(CameraOverlayActivity.OUTPUT_FILE_URI_EXTRA))
```
conroy-ricketts: Looks like `getParcelableExtra` is deprecated

app/unit-tests/src/org/commcare/activities/camera/CameraOverlayActivityTest.kt
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    @Test
+    fun `getIntent targets CameraOverlayActivity and carries the output uri`() {
+        val uri = Uri.parse("content://org.commcare.dalvik.fileprovider/external/output.jpg")
+
+        val intent =
+            CameraOverlayActivity.getIntent(ApplicationProvider.getApplicationContext(), uri)
+
+        assertEquals(CameraOverlayActivity::class.java.name, intent.component?.className)
```
conroy-ricketts: Should this use `!!`?

app/res/values/strings.xml
```diff
@@ -637,6 +637,12 @@
     </string-array>
 
     <string name="micro_image_activity_title">Take Profile Photo</string>
+    <string name="image_capture_activity_title" cc:translatable="true">Take Photo</string>
```
OrangeAndGreen, line 641: Thinking we should add the translations for these strings since they're system-level now and not app-based.

OrangeAndGreen (changes requested):

shubham1g5 (dismissed):
(wrong comment - ignore)
(approved by mistake - I am yet pending to review this)

Jignesh-dimagi:
> I like the height of `reticleMarginRatio = 0.3` (so it doesn't overlap with the shutter button) and the width of `reticleMarginRatio = 0.2` (so it doesn't appear too small)

Good callout, but because the reticle size is dynamic based on screen dimensions, we can't fully guarantee it won't overlap with the shutter button on every device.

app/src/org/commcare/activities/camera/CameraOverlayActivity.kt
Jignesh-dimagi: While MINIMIZE_LATENCY ensures the photo is captured and returned faster, it can result in noisier image quality. The core requirement of the MUAC flow is legibility; the saved photo must be clear enough to read the numbers on the measuring tape. Because this is a single, deliberate manual capture rather than a burst-shot scenario, image quality must take priority over shutter speed.

conroy-ricketts: Got it, thanks for the clarification!

app/res/layout/camera_overlay_activity.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="utf-8"?>
+<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
shubham1g5, line 2: given this is used on CC side, it need to support landscape version as well, as such are you able to verify if we need to tweak this for landscape vs if the current layout is acceptable in landscape and if so, provide screenshots/video for the landscape in the  PR description.

Jignesh-dimagi:
> Also think we need to define different height and width ratios here. The box get too narrow for 0.3. and 0.4, maybe we should define minmum width/height of the box as well so that any aspect ratio doesn't make the box dimension smaller than a min threshold

@shubham1g5 @OrangeAndGreen @conroy-ricketts I have added [this section](https://docs.google.com/document/d/1n0g37HhqfSxrmKoYi-4aBd3PLG4vYvotm-OlpblL1ew/edit?tab=t.0#heading=h.5a8r32e3vr5m) directly to the design doc. Please feel free to review it and provide your feedback there.

app/res/values/strings.xml
Jignesh-dimagi: Yeah, agree. https://github.com/dimagi/commcare-android/pull/3788/changes/04aeec7cf859c2aac8ec136b51d64a51e46d8de9

app/AndroidManifest.xml
```diff
@@ -572,6 +572,9 @@
         <activity
             android:name="org.commcare.activities.camera.MicroImageActivity"
             android:screenOrientation="portrait"/>
+        <activity
+            android:name="org.commcare.activities.camera.CameraOverlayActivity"
+            android:screenOrientation="portrait"/>
```
shubham1g5: This needs to support both landscape and portrait given it's a CC side feature.

app/src/org/commcare/activities/camera/CameraOverlayActivity.kt
```diff
@@ -0,0 +1,148 @@
...
+
+    override fun getContentLayout(): Int = R.layout.camera_overlay_activity
+
+    override fun getTitleRes(): Int = R.string.image_capture_activity_title
+
+    override fun getCameraView(): PreviewView = findViewById(R.id.view_finder)
+
+    override fun getCameraSelector(): CameraSelector = CameraSelector.DEFAULT_BACK_CAMERA
+
+    override fun getTargetResolution(): Size = PREVIEW_TARGET_RESOLUTION
```
shubham1g5: Can we assign `null` instead given it's not used in this screen ?

app/res/values/strings.xml
shubham1g5, line 641: all strings that support `cc:translatable="true"` needs to use `StringUtils.getStringRobust` to get String in the corresponding code references as the default `Context.getString` doesn't take the app translations into account.

app/src/org/commcare/activities/camera/CameraOverlayActivity.kt
```diff
@@ -0,0 +1,148 @@
...
+     * Reports a non-fatal capture failure: logs it, toasts the user, and re-enables the shutter so
+     * the capture can be retried without leaving the screen.
+     */
+    @VisibleForTesting
+    internal fun handleCaptureError(
+        logMessage: String,
+        e: Throwable?,
+    ) {
+        if (e == null) {
+            Logger.log(LogTypes.TYPE_EXCEPTION, logMessage)
```
shubham1g5: all errors needs to use `Logger.exception` for it to be logged as a non-fatal in Crashlytics, ` Logger.log` is used for supporting debug statements for an error.

app/src/org/commcare/activities/camera/CameraOverlayActivity.kt
```diff
@@ -0,0 +1,148 @@
...
+        isCapturing = capturing
+        findViewById<ImageView>(R.id.camera_shutter_button).apply {
+            isEnabled = !capturing
+            alpha = if (capturing) DISABLED_SHUTTER_ALPHA else 1f
+        }
+        findViewById<View>(R.id.capture_progress).visibility =
+            if (capturing) View.VISIBLE else View.GONE
+    }
+
+    private fun OutputStream.closeQuietly() {
```
shubham1g5: this would be a good method to put in a new `StreamExtensions.kt`

app/unit-tests/src/org/commcare/activities/camera/CameraOverlayActivityTest.kt
```diff
@@ -0,0 +1,145 @@
...
+        )
+    }
+
+    @Test
+    fun `entering capturing state disables the shutter and shows progress`() {
+        val activity = buildActivity()
+        val shutter = activity.findViewById<ImageView>(R.id.camera_shutter_button)
+        val progress = activity.findViewById<View>(R.id.capture_progress)
+
+        activity.handleCapturingState(true)
```
shubham1g5: this should not be called directly but we should just interact with activity UI and simulate a click on shutter

app/unit-tests/src/org/commcare/activities/camera/CameraOverlayActivityTest.kt
```diff
@@ -0,0 +1,145 @@
...
+
+    @Test
+    fun `captureImage enters capturing state and delegates to ImageCapture`() {
+        val activity = buildActivity()
+        val shutter = activity.findViewById<ImageView>(R.id.camera_shutter_button)
+        val progress = activity.findViewById<View>(R.id.capture_progress)
+        val imageCapture = mock<ImageCapture>()
+        activity.imageCapture = imageCapture
+
+        activity.captureImage()
```
shubham1g5: we should simulate shutter click instead.

shubham1g5 (changes requested):

app/unit-tests/src/org/commcare/activities/camera/CameraOverlayActivityTest.kt
```diff
@@ -0,0 +1,141 @@
...
+
+        verify(imageCapture).takePicture(any(), any(), any())
+        assertFalse("Shutter should be disabled while the capture is in flight", shutter.isEnabled)
+        assertEquals(View.VISIBLE, progress.visibility)
+    }
```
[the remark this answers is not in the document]
shubham1g5: +1 on verifying the complete flow here , also we should be able to mock the the Image Capture APIs to simulate success vs error here without having to call the `handleCaptureError` directly.

app/src/org/commcare/activities/camera/CameraOverlayActivity.kt
shubham1g5: >  it can result in noisier image quality

Curious if you are seeing that in practice ? Given we have been using that mode already with Personal ID photo capture and have not heard of this issue. 

Also what's the impact on image size here in the 2 modes and do we have more info on what intermal camera mechanics do the 2 mode controls ?

app/unit-tests/src/org/commcare/activities/camera/CameraOverlayActivityTest.kt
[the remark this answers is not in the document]
Jignesh-dimagi: I have added the said tests now.

app/unit-tests/src/org/commcare/activities/camera/CameraOverlayActivityTest.kt
Jignesh-dimagi: Done https://github.com/dimagi/commcare-android/pull/3788/changes/d6fb703f236f08d34884e8891079ca9e00f8c6ba

app/unit-tests/src/org/commcare/activities/camera/CameraOverlayActivityTest.kt
Jignesh-dimagi: Done https://github.com/dimagi/commcare-android/pull/3788/changes/d6fb703f236f08d34884e8891079ca9e00f8c6ba

app/src/org/commcare/activities/camera/CameraOverlayActivity.kt
Jignesh-dimagi: Done https://github.com/dimagi/commcare-android/pull/3788/changes/d6fb703f236f08d34884e8891079ca9e00f8c6ba

app/src/org/commcare/activities/camera/CameraOverlayActivity.kt
Jignesh-dimagi: Done https://github.com/dimagi/commcare-android/pull/3788/changes/d6fb703f236f08d34884e8891079ca9e00f8c6ba

app/res/values/strings.xml
Jignesh-dimagi: Thanks for flagging it. Done https://github.com/dimagi/commcare-android/pull/3788/changes/d6fb703f236f08d34884e8891079ca9e00f8c6ba

app/src/org/commcare/activities/camera/CameraOverlayActivity.kt
Jignesh-dimagi: Done https://github.com/dimagi/commcare-android/pull/3788/changes/d6fb703f236f08d34884e8891079ca9e00f8c6ba

app/AndroidManifest.xml
Jignesh-dimagi: Done https://github.com/dimagi/commcare-android/pull/3788/changes/d6fb703f236f08d34884e8891079ca9e00f8c6ba

app/res/layout/camera_overlay_activity.xml
Jignesh-dimagi: Since the landscape layout isn't complex, I was able to implement it without creating a separate landscape XML layout file. I have attached screenshots of the implementation in the PR description and look forward to your thoughts.

app/unit-tests/src/org/commcare/activities/camera/CameraOverlayActivityTest.kt
Jignesh-dimagi: Done https://github.com/dimagi/commcare-android/commit/d6fb703f236f08d34884e8891079ca9e00f8c6ba

app/unit-tests/src/org/commcare/activities/camera/CameraOverlayActivityTest.kt
Jignesh-dimagi: Done https://github.com/dimagi/commcare-android/commit/d6fb703f236f08d34884e8891079ca9e00f8c6ba

app/src/org/commcare/activities/camera/CameraOverlayActivity.kt
Jignesh-dimagi: Because the primary requirement here is maximizing image clarity over other performance metrics, the implementation explicitly uses the CAPTURE_MODE_MAXIMIZE_QUALITY flag and ignores alternative flags.

app/res/layout/camera_overlay_activity.xml
shubham1g5: Think the layout looks good here, a few comments on the overlay - 

In portrait -> think we want the width a little smaller and height a little larger
In landscape -> the width seems great but we should increase the height a bit

Jignesh-dimagi: I actually used Gemini for this, providing it with all the context alongside our desired image capture samples. It generated these scaling numbers, but I am completely open to feedback or suggestions.

shubham1g5: Actually, reg. the landscape layout, the shutter should move to the right hand side of the layout in the landscape mode , i.e. what I am noticing on most camera apps in my phone (it might just be that these apps are locking to portrait and if so, think we want just the overlay to adjust and not the shutter position) . I highly suggest looking at this behaviour in other camera apps.

app/unit-tests/src/org/commcare/activities/camera/CameraOverlayActivityTest.kt
shubham1g5: can we remove the ` @VisibleForTesting` annotations now from the activity

app/res/layout/camera_overlay_activity.xml
shubham1g5: > I actually used Gemini for this, providing it with all the context alongside our desired image capture samples. It generated these scaling numbers, but I am completely open to feedback or suggestions.

Still strongly think we should make those size tweaks for overlays to be better positioned to capture a generic object.

app/src/org/commcare/activities/camera/CameraOverlayActivity.kt
Jignesh-dimagi: > do we have more info on what intermal camera mechanics do the 2 mode controls ?

Claude/Gemini: The CAPTURE_MODE_* constants are a CameraX abstraction — they don't apply to the stock camera app at all. The default/system camera app (the one that handles MediaStore.ACTION_IMAGE_CAPTURE) is a separate app (Google Camera, Samsung Camera, etc.) built directly on camera2 + the vendor HAL (Hardware Abstraction Layer). It doesn't consume these CameraX enum values, so there's no CAPTURE_MODE it uses.

> Also what's the impact on image size here in the 2 modes

I ran small-scale tests comparing the two modes, focusing on the same object. While my manual framing wasn't identical each time, I observed a 10% to 15% reduction in file size when using MINIMIZE_LATENCY. Theoretically, this mode can produce images 30% to 50% smaller than MAXIMIZE_QUALITY, but actual savings vary by device. Visually, I couldn't detect a quality difference, likely due to testing on a higher-end phone. However, we can conclude that MINIMIZE_LATENCY does consistently result in smaller file sizes.

How the modes handle picture quality (From Claude):
MINIMIZE_LATENCY: 3A (Auto-Focus, Auto-Exposure, and Auto-White-Balance) convergence is not guaranteed. It can capture a frame before these loops finish running. How aggressively it short-circuits 3A depends on the vendor's HAL; some devices still produce sharp frames, while lower-end hardware might hand back soft-focus or underexposed shots.

MAXIMIZE_QUALITY: While not entirely device-independent, it raises the quality floor and reduces variance across different hardware. It forces 3A convergence and requests HIGH_QUALITY processing everywhere, making it less likely to produce a bad frame on an unverified field device.

There is also Zero Shutter Lag (CAPTURE_MODE_ZERO_SHUTTER_LAG) mode: It’s highly restricted (incompatible with flash/ImageAnalysis, falls back to latency mode if unsupported) and meant for action shots. It doesn't add value for our static reticle use case.

If CommCare runs across a wide, heterogeneous, may be low-end devices,  MAXIMIZE_QUALITY seems better option. Thoughts?

app/res/layout/camera_overlay_activity.xml
Jignesh-dimagi: > In portrait -> think we want the width a little smaller and height a little larger
> In landscape -> the width seems great but we should increase the height a bit

I have increased the width and height as suggested and updated the screenshots in the PR description.


> Actually, reg. the landscape layout, the shutter should move to the right hand side of the layout in the landscape mode , i.e. what I am noticing on most camera apps in my phone (it might just be that these apps are locking to portrait and if so, think we want just the overlay to adjust and not the shutter position) . I highly suggest looking at this behaviour in other camera apps.

I shifted the shutter to the side to fix the overlap with the reticle in landscape mode. I decided to stick with this layout because matching the system camera app's exact behavior would require a moderate amount of effort and introduce unresolved design questions (as default camera app doesn't have such custom reticle).

app/res/values/strings.xml
```diff
@@ -638,6 +638,12 @@
...
+    <string name="image_capture_failed" cc:translatable="true">Failed to save photo, please try again</string>
+    <string name="camera_start_failed" cc:translatable="true">Camera failed to start</string>
+    <string name="camera_permission_denied" cc:translatable="true">Go to settings and enable camera permission to use this feature</string>
+    <string name="microimage_face_detection_mode_failed" cc:translatable="true">Face detection mode failed, switching to manual mode</string>
+    <string name="microimage_cropping_failed" cc:translatable="true">Image cropping failed, make sure the target is inside the capture area</string>
```
conroy-ricketts, line 646: Just to confirm something here - do these still need `cc:translatable="true"` even though we defined translations for them?

app/src/org/commcare/activities/camera/BaseCameraActivity.kt
```diff
@@ -46,7 +45,11 @@ abstract class BaseCameraActivity :
             if (isGranted) {
                 startCamera()
             } else {
-                logErrorAndExit("Error acquiring camera permission", "microimage.camera.permission.denied", null)
+                logErrorAndExit(
+                    "Error acquiring camera permission",
+                    StringUtils.getStringRobust(this, R.string.camera_permission_denied),
```
conroy-ricketts, line 50: Confirming that `getStringRobust` is still necessary with the new translations that were added?

app/src/org/commcare/utils/TypedArrayExtensions.kt
```diff
@@ -0,0 +1,8 @@
+@file:JvmName("TypedArrayExtensions") // This sets the generated class name
+
+package org.commcare.utils
+
+import android.content.res.TypedArray
+
+/** Returns the float at [index] if the attribute is present, otherwise null. */
+fun TypedArray.optionalFraction(index: Int): Float? = if (hasValue(index)) getFloat(index, 0f) else null
```
OrangeAndGreen: "optionalFraction" seems like an odd name for this, I'm not sure how it relates to fractions. Thinking getOptionalFloat might be better?

OrangeAndGreen (dismissed):

app/res/values/strings.xml
shubham1g5: That's right as these strings are used inside the app so App translations should be able to override and provide translations they want here through HQ App config.

app/src/org/commcare/activities/camera/BaseCameraActivity.kt
shubham1g5: yes, as cc:translatable="true" doesn't have any effect otherwise.

app/src/org/commcare/activities/camera/CameraOverlayActivity.kt
shubham1g5: > , I observed a 10% to 15% reduction in file size when using MINIMIZE_LATENCY.

Seems like the size variance is not a lot even on a higher end device where I will expect it to be more prominent because of higher pixels camera availability, so seems good to opt for `MAXIMIZE_QUALITY` here.

app/src/org/commcare/utils/TypedArrayExtensions.kt
shubham1g5: +1 for `getOptionalFloat`

shubham1g5 (dismissed):
Think this is good to be merged for now and we can make the size changes based on product feedback in another PR if you prefer.

app/res/values/strings.xml
conroy-ricketts: Got it, thanks!

conroy-ricketts (dismissed):
🚀

Jignesh-dimagi:
> Think this is good to be merged for now and we can make the size changes based on product feedback in another PR if you prefer.

Yeah, I'd prefer that since we're still waiting for feedback.

shubham1g5 (approved):

PR #3796, opened 2026-07-08
CCCT-2476 Log Manage Profile Actions To Firebase

conroy-ricketts:
## Technical Summary
The three Manage Profile actions are reported as one event keyed by an action parameter rather than separate events, so they can be filtered together in Firebase. The name-updated action fires only when the name actually changed, so an email-only save is not double-counted.

### Safety story
- I ran the build on my test device and used breakpoints to confirm every meaningful Manage Profile action reaches Firebase, alongside the pre-existing screen-open and email-verification events.
- Change is analytics-only: no user-facing behavior, data model, or control flow is altered.

### Automated test coverage
- Fragment tests assert the name-updated action is logged on a successful name save, is not logged when only the email changes, and that the discard action is logged when the discard dialog is confirmed.

[CCCT-2476]: https://dimagi.atlassian.net/browse/CCCT-2476?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/org/commcare/personalId/profile/PersonalIdProfileEditFragment.kt
+++ b/app/src/org/commcare/personalId/profile/PersonalIdProfileEditFragment.kt
@@ -206,6 +206,10 @@ class PersonalIdProfileEditFragment : BasePersonalIdProfileFragment() {
             sessionData = null,
             tracker = viewModel.emailOtpTracker,
             onSuccess = {
+                FirebaseAnalyticsUtil.reportPersonalIdProfileAction(
+                    AnalyticsParamValue.MANAGE_PROFILE_ACTION_EMAIL_UPDATE_INITIATED,
+                    AnalyticsParamValue.MANAGE_PROFILE_OUTCOME_SUCCESS,
+                )
                 _binding ?: return@sendEmailOtp
                 findNavController().navigate(
                     PersonalIdProfileEditFragmentDirections.actionProfileEditToEmailVerification(
@@ -216,6 +220,10 @@ class PersonalIdProfileEditFragment : BasePersonalIdProfileFragment() {
                 )
             },
             onFailure = { errorCode, throwable ->
+                FirebaseAnalyticsUtil.reportPersonalIdProfileAction(
+                    AnalyticsParamValue.MANAGE_PROFILE_ACTION_EMAIL_UPDATE_INITIATED,
+                    AnalyticsParamValue.MANAGE_PROFILE_OUTCOME_FAILURE,
+                )
                 _binding ?: return@sendEmailOtp
                 onSaveFailed(errorCode, throwable)
             },
--- a/app/src/org/commcare/personalId/profile/PersonalIdProfileEditFragment.kt
+++ b/app/src/org/commcare/personalId/profile/PersonalIdProfileEditFragment.kt
@@ -39,10 +39,15 @@ class PersonalIdProfileEditFragment : BasePersonalIdProfileFragment() {
                     _binding?.let { loadUserPhoto(it.profileHeader.profileUserImage, photoBase64) }
                     FirebaseAnalyticsUtil.reportPersonalIdProfileAction(
                         AnalyticsParamValue.MANAGE_PROFILE_ACTION_PHOTO_UPDATED,
+                        AnalyticsParamValue.MANAGE_PROFILE_OUTCOME_SUCCESS,
                     )
                 },
                 onFailure = { _, _ ->
-                    // No-op for the Profile screen. The app sidebar shows a warning icon in the other flow.
+                    // No UI feedback on the Profile screen; the app sidebar shows a warning icon in the other flow.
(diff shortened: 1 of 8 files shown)


conroy-ricketts:
## Suggested Review Order

- `app/src/org/commcare/google/services/analytics/CCAnalyticsEvent.java` — new event name
- `app/src/org/commcare/google/services/analytics/CCAnalyticsParam.java` — new param key
- `app/src/org/commcare/google/services/analytics/AnalyticsParamValue.java` — the three action values
- `app/src/org/commcare/google/services/analytics/FirebaseAnalyticsUtil.java` — reporting helper tying the above together
- `app/src/org/commcare/personalId/profile/PersonalIdProfileEditFragment.kt` — call sites for photo, name, and discard actions
- `app/unit-tests/src/org/commcare/personalId/profile/BasePersonalIdProfileTest.kt` — static mock setup for the analytics util
- `app/unit-tests/src/org/commcare/personalId/profile/PersonalIdProfileEditFragmentTest.kt` — assertions for each logged action

app/src/org/commcare/personalId/profile/PersonalIdProfileEditFragment.kt
```diff
@@ -35,6 +37,9 @@ class PersonalIdProfileEditFragment : BasePersonalIdProfileFragment() {
                 onSuccess = { photoBase64 ->
                     viewModel.onPhotoUpdated(photoBase64)
                     _binding?.let { loadUserPhoto(it.profileHeader.profileUserImage, photoBase64) }
+                    FirebaseAnalyticsUtil.reportPersonalIdProfileAction(
+                        AnalyticsParamValue.MANAGE_PROFILE_ACTION_PHOTO_UPDATED,
```
shubham1g5, line 41: nit: Think this should be logged from the drawer photo update as well for completeness though maybe we want to pass an additional parameter to signify that it's been updated from drawer instead.

app/src/org/commcare/personalId/profile/PersonalIdProfileEditFragment.kt
```diff
@@ -156,6 +165,14 @@ class PersonalIdProfileEditFragment : BasePersonalIdProfileFragment() {
         }.updateProfile(requireActivity(), user.userId, user.password, viewModel.currentName, null, null)
     }
 
+    private fun reportSavedProfileDetails() {
+        if (viewModel.isNameModified()) {
+            FirebaseAnalyticsUtil.reportPersonalIdProfileAction(
+                AnalyticsParamValue.MANAGE_PROFILE_ACTION_NAME_UPDATED,
+            )
```
shubham1g5, line 180: nit: how about changing this to a 2 param event with the first param being a generalised `MANAGE_PROFILE_ACTION_PROFILE_UPDATED` and second param being what got updated (`photo`, `name` etc)

shubham1g5 (dismissed):

Jignesh-dimagi:
@conroy-ricketts The changes look good. I think we should also add logging for the following events:

- Whenever a user clicks 'Cancel' or 'Accept' on the email OTP confirmation dialog, it should be reported to Firebase. We can use reportUserPromptEvent here.

- We currently report how many users successfully update their profiles. It would be valuable to also track how many users attempt the update, alongside the number of unsuccessful attempts.

- We currently report when a user updates their photo or name, but we should also log whenever a user's email is updated.

conroy-ricketts:
@Jignesh-dimagi 

> Whenever a user clicks 'Cancel' or 'Accept' on the email OTP confirmation dialog, it should be reported to Firebase. We can use reportUserPromptEvent here.

Good idea!

> We currently report how many users successfully update their profiles. It would be valuable to also track how many users attempt the update, alongside the number of unsuccessful attempts.

Love the idea of tracking failed attempts, but I'm curious if there's a lot of value in explicitly tracking a number here? Feel like tracking the number for OTPs, for example, like we currently do makes sense because sending OTPs involves relatively more retries, but not sure if that has much value on the Profile screen itself 

> We currently report when a user updates their photo or name, but we should also log whenever a user's email is updated.

Thought about this and originally argued against it because we already have a lot of logs for the Email verification screen, but there may be some value in adding another log on the profile screen. I'll look into this again.

conroy-ricketts:
@Jignesh-dimagi 

I looked into the point about logging the email change some more. I don't think there's a simple place to log an email update with an outcome on the Profile Edit screen specifically, but it may still be valuable to log a distinct `"email_update_initiated"`. I'm going to add that shortly

app/src/org/commcare/personalId/profile/PersonalIdProfileEditFragment.kt
conroy-ricketts: I'm really torn on this one, so chose to skip

shubham1g5 (approved):

PR #3799, opened 2026-07-09
Fix GPS Polling Lifecycle Bugs And Track Accuracy Degradation

shubham1g5:
## Product Description

No User facing changes

https://dimagi.atlassian.net/browse/CCCT-2605

[inspired from the doc](https://docs.google.com/document/d/1Oy68CXCo9KQgBTHuLDrCtkNJp-ViC5c3xqzPq9_sJgw/edit?tab=t.4mrw48u8hau2#heading=h.8hvu5tslm97u). 

## Technical Summary

The PR try to ahieve 2 specific goals - 

1. Handle `PollSensorController` location updates better in conjunction with the Form entry by 

a) moving the call to `stopLocationPolling()` from `onPause(isFinishing=true)` to `onDestroy(!isChangingConfigurations)` so that GPS is always released when the activity is truly destroyed. 

b) Adds an `actions.isEmpty()` guard inside `startLocationPolling()`'s `Handler.post()` lambda. Prevents an orphaned `LocationController` when `stopLocationPolling()` runs between the background-thread call to `startLocationPolling()` and its deferred main-thread execution.

2. Tracks in analytics whenever auto-capture receives a location with accuracy worst than last location, this is to see how common it is for GPS location to get worst in subsequent tries. It uses an exiting catch all `common_commcare_event` to protect against 200 events quota, also this is only a temporary event and we will remove this in 2.64 release.

### Safety story

**Confidence:**
- I tested automated location capture locally and confirmed it still works end-to-end.
- The diff is narrow — three focused changes to `PollSensorController` and one lifecycle hook change in `FormEntryActivity`.
- The race condition fix is provably safe: both `stopLocationPolling()` and the `Handler.post()` lambda execute on the main thread, so the `actions.isEmpty()` check always sees consistent post-stop state.

**Risks:**
- The OS-kill-while-backgrounded path (the primary motivation for the `onDestroy` change) is hard to reproduce reliably and was not directly exercised.

Review commit by commit.

### DIFF
--- a/app/src/org/commcare/activities/FormEntryActivity.java
+++ b/app/src/org/commcare/activities/FormEntryActivity.java
@@ -988,10 +988,6 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
         attemptToUnregisterBroadcastReceiver(mLocationServiceIssueReceiver);
 
         saveInlineVideoState();
-
-        if (isFinishing()) {
-            PollSensorController.INSTANCE.stopLocationPolling();
-        }
         TextToSpeechConverter.INSTANCE.stop();
 
         attemptToUnregisterBroadcastReceiver(pendingSyncAlertBroadcastReceiver);
@@ -1375,6 +1371,7 @@ public class FormEntryActivity extends SaveSessionCommCareActivity<FormEntryActi
 
         if (!isChangingConfigurations()) {
             isFormEntryActive = false;
+            PollSensorController.INSTANCE.stopLocationPolling();
         }
 
         TextToSpeechConverter.INSTANCE.shutDown();
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -2,6 +2,12 @@
 This file is meant as an easy way for us to collate notes and change logs across releases. 
 -->
 
+## CommCare 2.63.2
+
+### QA Notes
+
+- Make sure there are no regressions in automated location capture in a CC form.  
+
 ## CommCare 2.63.1
 
 ### Release Notes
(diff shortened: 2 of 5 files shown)


app/src/org/commcare/activities/FormEntryActivity.java
```diff
@@ -1375,6 +1371,7 @@ protected void onDestroy() {
 
         if (!isChangingConfigurations()) {
             isFormEntryActive = false;
+            PollSensorController.INSTANCE.stopLocationPolling();
```
avazirna, line 1374: Is there any evidence that `stopLocationPolling()` was not always called from `onPause()`?

avazirna (approved):

app/src/org/commcare/activities/FormEntryActivity.java
shubham1g5: yeah that's easily reproducible if you just background the `FormEntryActivity` using App home button, `isFinishing` will be `false` and `stopLocationPolling` won't be called.

shubham1g5: And think that was intentional bdw so that the location capture doesn't stop when we background the app, just that putting it in `onDestory` achieves both objectives much more cleanly -> 1. Location capture being alive in background and 2. Making sure that we always call `stopLocationPolling` when activity gets killed anyhow.

avazirna: Yes, I was referring during activity shutdown. The reasoning sounds right to me.

conroy-ricketts (approved):

ctsims (approved):
Looks good to me with the change to the accuracy reporting interval. 

noting one thing: This does let accuracy drift upwards incrementally, so in theory we could be recording something like "Best accuracy," but that re-introduces the lifecycle challenge (we'd need to clear the accuracy or reset it back when we trigger the threshold), so good to start here looking for big jumps and add more instrumentation later if we don't see it.

PR #3805, opened 2026-07-13
CCCT-2536 Add ConnectSuccessFailureCard Dismissible Status Card

Jignesh-dimagi:
## Product Description
A reusable dismissible status card for Connect showing a short message in a success or failure style. Not yet placed on any screen, so no user-visible change ships in this PR.

## Technical Summary
- Colors and card chrome live in a style read as the `defStyleRes` fallback. 
- The card is code-driven—hidden by default and revealed via the `show(mode, message)` method. This method also scopes the optional dismiss handler, which triggers whenever the user presses the close (X) button.
- It defaults to `GONE` in `init` only when the XML declares no explicit `android:visibility`, so declarative visibility is still honored.

```
Usage
val card = findViewById<ConnectSuccessFailureCard>(R.id.status_card)
  // or ViewBinding: val card = binding.statusCard
  
  // without callback
  card.show(ConnectSuccessFailureCard.Mode.SUCCESS, R.string.tasks_completed)
  card.show(ConnectSuccessFailureCard.Mode.FAILURE, context.getString(R.string.tasks_completed))
  
  // with dismiss callback
  card.show(ConnectSuccessFailureCard.Mode.FAILURE, R.string.download_failed) {
      viewModel.onStatusDismissed()
  }

```
The sandbox activity was created solely to test the card.
Success Failure

### Safety story
- I exercised both success and failure modes and the ✕ dismiss on a device/emulator via a throwaway sandbox activity (reverted, not part of this PR).
- The 10 Robolectric tests pass.

### Automated test coverage
Robolectric tests cover per-mode styling, the hidden-by-default and XML-visibility-honoring rules, `show()` (mode/message/visibility and string-resource overload), and the dismiss-plus-callback behavior.

[CCCT-2536]: https://dimagi.atlassian.net/browse/CCCT-2536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/org/commcare/views/connect/ConnectSuccessFailureCard.kt
+++ b/app/src/org/commcare/views/connect/ConnectSuccessFailureCard.kt
@@ -80,15 +80,18 @@ class ConnectSuccessFailureCard
                 onDismiss?.invoke()
             }
 
+            context.withStyledAttributes(attrs, intArrayOf(android.R.attr.visibility), defStyleAttr) {
+                if (!hasValue(0)) {
+                    visibility = GONE
+                }
+            }
+
             context.withStyledAttributes(
                 attrs,
                 R.styleable.ConnectSuccessFailureCard,
                 defStyleAttr,
                 R.style.Widget_CommCare_ConnectSuccessFailureCard,
             ) {
-                if (!hasValue(R.styleable.ConnectSuccessFailureCard_android_visibility)) {
-                    visibility = GONE
-                }
                 radius = getDimension(R.styleable.ConnectSuccessFailureCard_cardCornerRadius, radius)
                 cardElevation = getDimension(R.styleable.ConnectSuccessFailureCard_cardElevation, cardElevation)
                 successBackgroundColor = getColor(R.styleable.ConnectSuccessFailureCard_successBackgroundColor, 0)
--- a/app/src/org/commcare/views/connect/ConnectSuccessFailureCard.kt
+++ b/app/src/org/commcare/views/connect/ConnectSuccessFailureCard.kt
@@ -1,0 +1,123 @@
+package org.commcare.views.connect
+
+import android.content.Context
+import android.content.res.ColorStateList
+import android.util.AttributeSet
+import android.view.LayoutInflater
+import androidx.annotation.StringRes
+import androidx.cardview.widget.CardView
+import androidx.core.content.withStyledAttributes
+import androidx.core.widget.ImageViewCompat
+import org.commcare.dalvik.R
+import org.commcare.dalvik.databinding.ViewConnectSuccessFailureCardBinding
+
+/**
+ * Reusable dismissible card that shows a short status message in a [Mode.SUCCESS] or
+ * [Mode.FAILURE] style. Mode drives the background, leading icon, and tint; the message
+ * text is supplied by the caller. The card is hidden by default (unless the XML sets an
+ * explicit `android:visibility`) and revealed via [show]; tapping the close icon hides it
+ * again and invokes the handler passed to [show].
+ */
+class ConnectSuccessFailureCard
+    @JvmOverloads
+    constructor(
+        context: Context,
+        attrs: AttributeSet? = null,
+        defStyleAttr: Int = 0,
+    ) : CardView(context, attrs, defStyleAttr) {
+        enum class Mode { SUCCESS, FAILURE }
+
+        private val binding =
+            ViewConnectSuccessFailureCardBinding.inflate(LayoutInflater.from(context), this, true)
+
+        var mode: Mode? = null
+            set(value) {
+                field = value
+                value?.let { applyMode(it) }
+            }
+
+        var messageText: CharSequence?
+            get() = binding.successFailureCardText.text
+            set(value) {
+                binding.successFailureCardText.text = value
+            }
+
+        private var onDismiss: (() -> Unit)? = null
+
+        /**
+         * Configures the card for [mode] with [message], attaches an optional [onDismiss]
+         * handler for this showing, and makes the card visible.
+         */
+        @JvmOverloads
+        fun show(
+            mode: Mode,
+            message: CharSequence,
+            onDismiss: (() -> Unit)? = null,
+        ) {
+            this.onDismiss = onDismiss
+            this.mode = mode
+            messageText = message
+            visibility = VISIBLE
+        }
+
+        @JvmOverloads
+        fun show(
+            mode: Mode,
+            @StringRes messageRes: Int,
+            onDismiss: (() -> Unit)? = null,
+        ) = show(mode, context.getString(messageRes), onDismiss)
+
+        private var successBackgroundColor = 0
+        private var successAccentColor = 0
+        private var failureBackgroundColor = 0
+        private var failureAccentColor = 0
+
+        init {
+            useCompatPadding = false
+            if (attrs?.getAttributeValue(ANDROID_NAMESPACE, "visibility") == null) {
+                visibility = GONE
+            }
+
+            binding.successFailureCardClose.setOnClickListener {
+                visibility = GONE
+                onDismiss?.invoke()
+            }
+
+            context.withStyledAttributes(
+                attrs,
+                R.styleable.ConnectSuccessFailureCard,
+                defStyleAttr,
+                R.style.Widget_CommCare_ConnectSuccessFailureCard,
+            ) {
+                radius = getDimension(R.styleable.ConnectSuccessFailureCard_cardCornerRadius, radius)
+                cardElevation = getDimension(R.styleable.ConnectSuccessFailureCard_cardElevation, cardElevation)
+                successBackgroundColor = getColor(R.styleable.ConnectSuccessFailureCard_successBackgroundColor, 0)
+                successAccentColor = getColor(R.styleable.ConnectSuccessFailureCard_successAccentColor, 0)
+                failureBackgroundColor = getColor(R.styleable.ConnectSuccessFailureCard_failureBackgroundColor, 0)
+                failureAccentColor = getColor(R.styleable.ConnectSuccessFailureCard_failureAccentColor, 0)
+                if (hasValue(R.styleable.ConnectSuccessFailureCard_mode)) {
+                    mode = Mode.values()[getInt(R.styleable.ConnectSuccessFailureCard_mode, 0)]
+                }
+                getString(R.styleable.ConnectSuccessFailureCard_messageText)?.let { messageText = it }
+            }
+        }
+
+        private fun applyMode(mode: Mode) {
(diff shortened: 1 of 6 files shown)


Jignesh-dimagi:
## Suggested Review Order

- `app/res/values/attrs.xml` — the styleable defines the component's configuration contract (mode, message, colors, chrome).
- `app/res/values/colors.xml` — new `pale_coral` base color (failure background).
- `app/res/values/styles.xml` — `Widget.CommCare.ConnectSuccessFailureCard` supplies the defaults the view reads.
- `app/res/layout/view_connect_success_failure_card.xml` — the card's icon/text/close layout.
- `app/src/org/commcare/views/connect/ConnectSuccessFailureCard.kt` — the component tying the above together.
- `app/unit-tests/.../ConnectSuccessFailureCardTest.kt` — tests documenting the expected behavior.

app/src/org/commcare/views/connect/ConnectSuccessFailureCard.kt
```diff
@@ -0,0 +1,123 @@
...
+        private var successBackgroundColor = 0
+        private var successAccentColor = 0
+        private var failureBackgroundColor = 0
+        private var failureAccentColor = 0
+
+        init {
+            useCompatPadding = false
+            if (attrs?.getAttributeValue(ANDROID_NAMESPACE, "visibility") == null) {
+                visibility = GONE
+            }
```
shubham1g5: AttributeSet.getAttributeValue only reads attributes written inline on the XML tag and this would fail to inherit the `visibility` tag from any styles defined for this view, a simple fix I believe would be to use ` context.obtainStyledAttributes()`.

app/unit-tests/src/org/commcare/views/connect/ConnectSuccessFailureCardTest.kt
```diff
@@ -0,0 +1,149 @@
...
+        card.show(ConnectSuccessFailureCard.Mode.FAILURE, "done")
+
+        assertEquals(context.getColor(R.color.pale_coral), card.cardBackgroundColor.defaultColor)
+        assertEquals(failureColor(), card.message().currentTextColor)
+        assertEquals(failureColor(), card.icon().tintColor())
+        assertEquals(failureColor(), card.close().tintColor())
+    }
+
+    @Test
+    fun `card is hidden by default`() {
```
shubham1g5: this should be tested by `stays hidden when xml omits visibility` and can be removed.

app/src/org/commcare/views/connect/ConnectSuccessFailureCard.kt
```diff
@@ -0,0 +1,123 @@
...
+
+            binding.successFailureCardClose.setOnClickListener {
+                visibility = GONE
+                onDismiss?.invoke()
```
shubham1g5, line 80: Add `nDismiss = null` so that another call like to `card.show()` doesn't fire the old `onDismiss` again.

app/src/org/commcare/views/connect/ConnectSuccessFailureCard.kt
```diff
@@ -0,0 +1,123 @@
...
+                R.styleable.ConnectSuccessFailureCard,
+                defStyleAttr,
+                R.style.Widget_CommCare_ConnectSuccessFailureCard,
+            ) {
+                radius = getDimension(R.styleable.ConnectSuccessFailureCard_cardCornerRadius, radius)
+                cardElevation = getDimension(R.styleable.ConnectSuccessFailureCard_cardElevation, cardElevation)
+                successBackgroundColor = getColor(R.styleable.ConnectSuccessFailureCard_successBackgroundColor, 0)
+                successAccentColor = getColor(R.styleable.ConnectSuccessFailureCard_successAccentColor, 0)
+                failureBackgroundColor = getColor(R.styleable.ConnectSuccessFailureCard_failureBackgroundColor, 0)
+                failureAccentColor = getColor(R.styleable.ConnectSuccessFailureCard_failureAccentColor, 0)
```
shubham1g5, line 100: can we provide better defaults instead of `0` here so that individual components only define any customization and not the default colors.

app/src/org/commcare/views/connect/ConnectSuccessFailureCard.kt
```diff
@@ -0,0 +1,123 @@
...
+                if (hasValue(R.styleable.ConnectSuccessFailureCard_mode)) {
+                    mode = Mode.values()[getInt(R.styleable.ConnectSuccessFailureCard_mode, 0)]
```
shubham1g5: `Enum.values()` seems to be deprecated and hence should not be used in any new code.

app/src/org/commcare/views/connect/ConnectSuccessFailureCard.kt
```diff
@@ -0,0 +1,123 @@
...
+                }
+                getString(R.styleable.ConnectSuccessFailureCard_messageText)?.let { messageText = it }
+            }
+        }
+
+        private fun applyMode(mode: Mode) {
```
shubham1g5: A better design pattern here is to abstract the style as a separate Object - 

````
 private data class ModeStyle(                                                                                                                             
      val backgroundColor: Int,                                                                                                                             
      val iconRes: Int,                                                                                                                                     
      val accentColor: Int,                                                                                                                                 
  )
````

This would make it real easy to add a new `warning` mode to this view in future.

shubham1g5 (changes requested):

app/src/org/commcare/views/connect/ConnectSuccessFailureCard.kt
Jignesh-dimagi: I have changed here https://github.com/dimagi/commcare-android/pull/3805/changes/ea47ac28f42e592f6fc85d1803e50a739a03e017.
Note: Visibility is per-instance runtime state, not appearance. I would not prefer putting visibility in style as it's not appearance.

app/unit-tests/src/org/commcare/views/connect/ConnectSuccessFailureCardTest.kt
Jignesh-dimagi: Done https://github.com/dimagi/commcare-android/pull/3805/changes/ea47ac28f42e592f6fc85d1803e50a739a03e017

app/src/org/commcare/views/connect/ConnectSuccessFailureCard.kt
Jignesh-dimagi: Every show() call reassigns this.onDismiss — to the handler we pass, or to null when we omit it. So after a dismiss, the next show() overwrites the old handler before the card is visible again.

app/src/org/commcare/views/connect/ConnectSuccessFailureCard.kt
Jignesh-dimagi: The defaults already come from the style, not from 0. The withStyledAttributes call passes R.style.Widget_CommCare_ConnectSuccessFailureCard as the `defStyleRes`, and that style defines all four colors. Falling back to 0 would only happen if the style were missing a mode's color — that'd be a programming error, since each mode's default colors are expected to be defined in the style.

app/src/org/commcare/views/connect/ConnectSuccessFailureCard.kt
Jignesh-dimagi: Yeah, agree. Changed https://github.com/dimagi/commcare-android/pull/3805/changes/ea47ac28f42e592f6fc85d1803e50a739a03e017

app/src/org/commcare/views/connect/ConnectSuccessFailureCard.kt
Jignesh-dimagi: Replaced the Triple with a ModeStyle data class for clarity, and applyMode now takes a ModeStyle. Changed https://github.com/dimagi/commcare-android/pull/3805/changes/ea47ac28f42e592f6fc85d1803e50a739a03e017

app/src/org/commcare/views/connect/ConnectSuccessFailureCard.kt
shubham1g5: I am confused, why did we add the `  <attr name="android:visibility" />` given we are not supporting for styles ?

app/src/org/commcare/views/connect/ConnectSuccessFailureCard.kt
shubham1g5: yeah makes sense.

app/src/org/commcare/views/connect/ConnectSuccessFailureCard.kt
shubham1g5: > That'd be a programming error, since each mode's default colors are expected to be defined in the style.

My sense is that the standard practice here is for Styles to override the default behaviour here with use of these attributes but the widget should be usable without a style providing these values. Conside for eg. a `TextView` , it defines `black` (ish)  as the default text color and let users override it by setting a style.

Jignesh-dimagi: Yes, I agree with your point for default behaviour, but that's exactly how TextView works. TextView doesn't hardcode a color literal in its constructor; instead, its default text color resolves from the theme's default text view style or textAppearance (via the theme's textViewStyle).

Our widget applies that same pattern: it bundles its own default style via defStyleRes (passing Widget.CommCare.ConnectSuccessFailureCard into obtainStyledAttributes), allowing it to render correctly without any consumer-supplied style. If the defaults need to be overridden, developers can still do that directly in the XML. This makes the widget entirely usable on its own.

That said, I'm still open to your thoughts and can add the hardcoded defaults if you prefer.

app/src/org/commcare/views/connect/ConnectSuccessFailureCard.kt
Jignesh-dimagi: Yeah, I don't favor putting visibility in the style block either. I falsely assumed it was required, but I have removed it now. https://github.com/dimagi/commcare-android/pull/3805/changes/a4a33bc485794057fc2583d23bfc1c605916abd3

app/src/org/commcare/views/connect/ConnectSuccessFailureCard.kt
shubham1g5: Makes sense, thanks for looking into this..

shubham1g5 (dismissed):

OrangeAndGreen (dismissed):

shubham1g5 (approved):

PR #3812, opened 2026-07-15
CCCT-2542 Add Connect Progress Card View

OrangeAndGreen:
## Technical Summary

Adds `ConnectProgressCard`, a reusable Connect view intended for both the Learning Progress and Delivery Progress screens. It composes existing sub-views — an optional semi-circle indicator, a horizontal progress bar (label, count, caption), and an independently controllable info-message banner with an optional CTA — with appearance driven entirely by content properties rather than a state enum. Not yet wired into any live screen; this PR is the building block only.

Independently-controllable pieces:
* Title visibility
* Semi-circle progress visibility
* Disabled (grayed out) state
* Blue message appendage at bottom of card
* CTA visibility in blue message

## Screenshots
Learning progress:
image
Delivery progress:
image

Disabled state with message and CTA:
Disabled

### Safety story

What gives confidence:
- I added demo cards to the Learning Progress page to test the visible functionality (see screenshots above)
- I ran the `ConnectProgressCardTest` Robolectric suite locally and it passed, and rendered the card in a layout preview/harness to check appearance.
- The sync removal + rename is a self-contained, mostly-deletion change verified by the trimmed test suite; the class has no live-screen consumers, so there is effectively no blast radius for existing users.
- The rest of the diff is additive and self-contained: a new view class, its layout/attr/style resources, and its test. No existing files change behavior.

Risks to review:
- Because the card is not yet integrated into a real screen, its behavior in context (layout under real data, interaction with surrounding views) has only been exercised in a preview harness, not in the actual Learning/Delivery flows.

### Automated test coverage

`ConnectProgressCardTest` covers property-to-view bindings, `progressCurrent` clamping to `progressMax` (including setting current before max), the `contentEnabled` gray-out of the progress text and semi-circle, optional-content visibility (title, label, caption, count), the info-message/CTA banner and its click listener, attribute inflation, and per-instance color overrides.

[CCCT-2542]: https://dimagi.atlassian.net/browse/CCCT-2542?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


OrangeAndGreen:
## Suggested Review Order

- `app/res/values/attrs.xml` — declares the styleable attributes the view exposes; orients the rest
- `app/res/layout/view_connect_progress_and_status_card.xml` — the layout structure the view inflates and binds to
- `app/src/org/commcare/views/connect/ConnectProgressAndStatusCard.kt` — core logic tying attributes and layout together
- `app/unit-tests/.../ConnectProgressAndStatusCardTest.kt` — documents intended property/appearance behavior
- `app/res/drawable/bg_connect_progress_info_message.xml`, `app/res/drawable/bg_connect_sync_badge.xml`, `app/res/values/colors.xml`, `app/res/values/strings.xml` (+ translations) — supporting resources, read last

app/src/org/commcare/views/connect/ConnectProgressAndStatusCard.kt
```diff
@@ -0,0 +1,193 @@
...
+            binding.progressCardSyncIcon.setColorFilter(ContextCompat.getColor(context, iconColorRes))
+            binding.progressCardSyncIcon.backgroundTintList =
+                ColorStateList.valueOf(ContextCompat.getColor(context, badgeColorRes))
+        }
+
+        private fun applyContentColors() {
+            val primary =
+                ContextCompat.getColor(context, if (contentEnabled) R.color.connect_text_color else R.color.connect_grey)
+            val accent =
+                ContextCompat.getColor(context, if (contentEnabled) R.color.connect_dark_blue_color else R.color.connect_grey)
```
Jignesh-dimagi: This component has hardcoded colors inside the code, making it difficult to maintain—especially when changes are required. If we implement a color -> style -> component strategy, it will provide a default style. If needed, consumers can easily override it in their XML files. 

For example, we can define the content enabled/disabled colors inside the style block like this:
```
<style name="Widget.CommCare.ConnectProgressAndStatusCard" parent="">
    <item name="contentEnabledColor">@color/connect_text_color</item>
    <item name="contentDisabledColor">@color/connect_grey</item>
</style>
```


By referencing contentDisabledColor inside the class instead of hardcoding it, any future theme changes can be managed directly within this default style rather than hunting for colors in the codebase. Following this same pattern, all other component properties can also be defined as defaults within this style.

Jignesh-dimagi:
Currently, consumers have to manually bind each individual property of the card to configure it correctly:
```kotline
card.titleText = data.title
card.progressLabel = data.label
card.progressMax = data.total
card.progressCurrent = data.completed
// and so on...
```

This imperatively coupled approach leaves room for developers to accidentally miss properties, which can easily lead to inconsistent UI states or bugs.

Can we encapsulate this into a single, cohesive UI State model instead? We could introduce a State data class and expose a single bind method on the component:

``` kotlin
data class State(
    val title: CharSequence? = null,
    val progressLabel: CharSequence? = null,
    val progressCurrent: Int = 0,
    val progressMax: Int = 0,
    val progressCaption: CharSequence? = null,
    val contentEnabled: Boolean = true,
    val semiCircle: SemiCircle? = null,
    val sync: Sync? = null,
    val info: Info? = null,
) {
    data class SemiCircle(
        val current: Int,
        val max: Int,
        val description: CharSequence? = null,
    )

    data class Sync(
        val status: SyncStatus,
        val text: CharSequence,
        val subtext: CharSequence? = null,
    )

    class Info(
        val message: CharSequence,
        val ctaText: CharSequence? = null,
        val onCtaClick: (() -> Unit)? = null,
    )
}
```
This allows the consumer to bind the entire view state safely in a single atomic update:

``` kotlin
card.bind(
    ConnectProgressAndStatusCard.State(
        title = data.title,
        progressCurrent = data.completed,
        progressMax = data.total,
        contentEnabled = data.isActive,
        semiCircle = if (data.showSemiCircle) {
            State.SemiCircle(data.completed, data.total, data.semiCircleLabel)
        } else null,
        sync = data.syncMessage?.let {
            State.Sync(
                status = if (data.hasSyncProblem) SyncStatus.WARNING else SyncStatus.UP_TO_DATE,
                text = it,
                subtext = data.lastSynced,
            )
        },
        info = data.banner?.let {
            State.Info(it, data.ctaLabel, onCtaClick = { openMoreInfo() })
        },
    ),
)


```

Jignesh-dimagi (changes requested):

shubham1g5:
Based on screenshots, think this PR is implementing [outdated/under-iteration designs for task and sync cards](https://www.figma.com/design/4QLInPUv5wIfjfDGfrFESW/Opportunity-Home-Design-_-Ishwari?node-id=5288-29074&t=SXJIAoIo0NNKbuam-0#1840823118), both seems to be undergoing changes in latest iterations. Curious if it's not clear from the [Solutions Summary](https://docs.google.com/document/d/1rntc16FW2Jfr6CbzNcZ9RDOxIRrmfcWNU6_d5WL1rA8/edit?tab=t.xdhnuice1358#heading=h.8ghlb2mbk1zz) that these cards were going through iterations, [product is supposed to mark](https://docs.google.com/document/d/1rntc16FW2Jfr6CbzNcZ9RDOxIRrmfcWNU6_d5WL1rA8/edit?tab=t.xdhnuice1358#heading=h.8ghlb2mbk1zz) things getting iterated clearly on this doc and if it's not marked correctly think we should be loud and flag it to product on slack that we are repeatedly implementing Outdated designs due to lack of clarity.

OrangeAndGreen:
> Based on screenshots, think this PR is implementing [outdated/under-iteration designs for task and sync cards](https://www.figma.com/design/4QLInPUv5wIfjfDGfrFESW/Opportunity-Home-Design-_-Ishwari?node-id=5288-29074&t=SXJIAoIo0NNKbuam-0#1840823118), both seems to be undergoing changes in latest iterations. Curious if it's not clear from the [Solutions Summary](https://docs.google.com/document/d/1rntc16FW2Jfr6CbzNcZ9RDOxIRrmfcWNU6_d5WL1rA8/edit?tab=t.xdhnuice1358#heading=h.8ghlb2mbk1zz) that these cards were going through iterations.

@shubham1g5 Ah, the tasks message part is my mistake, I knew that was being discussed and left it out of the initial work, then when testing the implementation saw it missing, forgot it was intentional, and added it in. But for the sync part of the card, I don't see any indication in the Solutions Summary section about that being in flux, and based on your link it seems the only good source for that discussion right now is in the Playground section in Figma (really just a side effect of moving the tasks message around). So maybe there's some room for improvement in communicating which parts aren't finalized.

Should we move this ticket to blocked until the design is finalized?

shubham1g5:
> So maybe there's some room for improvement in communicating which parts aren't finalized.

Got it, are you able to flag it to Kim so that she is aware of these gaps. 


> Should we move this ticket to blocked until the design is finalized?

We should just remove the tasks and syncs card out of scope of this ticket and tackle them as other tickets as we need this work to be merged sooner than later to be able to make progress on the rest of the phase 1 work.

app/res/layout/view_connect_progress_card.xml
```diff
@@ -0,0 +1,142 @@
...
+        android:layout_height="wrap_content"
+        app:cardBackgroundColor="@color/white"
+        app:cardCornerRadius="@dimen/connect_info_card_corner_radius"
+        app:cardElevation="@dimen/connect_info_card_elevation">
+
+        <LinearLayout
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:orientation="vertical"
+            android:padding="16dp">
```
shubham1g5: lets not use any hardcoded dimensions values for new designs given we have a great chance to use standardize the dimensions across connect screen. We should also make the same changes for `SemiCircleProgressBar` and `LinearProgressBar` used in this view.

app/src/org/commcare/views/connect/ConnectProgressCard.kt
```diff
@@ -0,0 +1,184 @@
...
+                binding.progressCardSemiCircle.max = value
+            }
+
+        var semiCircleDescription: CharSequence?
+            get() = binding.progressCardSemiCircle.descriptionText
+            set(value) {
+                binding.progressCardSemiCircle.descriptionText = value
+            }
+
+        var progressLabel: CharSequence?
```
shubham1g5: nit: it would be nice to name linear and semi-circle progress bar properties clearly here by appending prefixes like `semiCircleProgressBarX` and `linearProgressBarX` as it's not very clear at the moment what progress we are talking about when we say `progressLabel`

app/res/values/strings.xml
```diff
@@ -533,6 +533,7 @@
     <string name="connect_progress_title">Delivery Progress</string>
     <string name="connect_progress_status">You have completed %d of %d visits.</string>
     <string name="semi_circle_progress_value_format" cc:translatable="true">%1$d of %2$d</string>
+    <string name="connect_progress_count_format" cc:translatable="true">%1$d of %2$d</string>
```
shubham1g5, line 536: What's the value in definining these strings across locales if they are identical across locales ?

shubham1g5 (changes requested):

shubham1g5:
would be nice to clean up commit history on this PR as well

app/res/layout/view_connect_progress_card.xml
```diff
@@ -0,0 +1,142 @@
...
+
+            <TextView
+                android:id="@+id/progress_card_title"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:textColor="@color/connect_text_color"
+                android:textSize="16sp"
```
conroy-ricketts: Question about the redesign as a whole: should we consider utilizing/adding text sizes in `dimens.xml` (e.g. `text_small`, `text_medium`, etc.)? That may make it easier for us to adjust text styles for users as needed in the future

app/res/values/styles.xml
```diff
@@ -309,4 +309,10 @@
         <item name="descriptionTextColor">@color/connect_text_color</item>
     </style>
 
+    <style name="Widget.CommCare.ConnectProgressCard" parent="">
```
conroy-ricketts, line 312: Curiosity question: what's the intention behind explicitly defining an empty parent here?

app/src/org/commcare/views/connect/ConnectProgressCard.kt
```diff
@@ -0,0 +1,184 @@
...
+            set(value) {
+                field = value.coerceAtLeast(0)
+                updateProgressDisplay()
+            }
+
+        var progressMax: Int = 0
+            set(value) {
+                field = value.coerceAtLeast(0)
+                updateProgressDisplay()
+            }
```
conroy-ricketts: Should we rather crash or log errors instead of coercing minimum values here? Getting the sense that invalid values for this card would be an indication of a programmatic error and thus may indicate a bigger issue for the user

app/src/org/commcare/views/connect/ConnectProgressCard.kt
```diff
@@ -0,0 +1,184 @@
...
+                progressMax = getInt(R.styleable.ConnectProgressCard_progressMax, progressMax)
+                progressCurrent = getInt(R.styleable.ConnectProgressCard_progressCurrent, progressCurrent)
+                progressCaption = getString(R.styleable.ConnectProgressCard_progressCaption)
+                infoMessage = getString(R.styleable.ConnectProgressCard_infoMessage)
+                infoCtaText = getString(R.styleable.ConnectProgressCard_infoCtaText)
+                contentEnabled = getBoolean(R.styleable.ConnectProgressCard_contentEnabled, true)
+            }
+        }
+
+        private fun dp(value: Int) = (value * resources.displayMetrics.density).toInt()
```
conroy-ricketts: Can we use the utility function instead to avoid code duplication: `ViewUtil.dpToPx()`

app/src/org/commcare/views/connect/ConnectProgressCard.kt
```diff
@@ -0,0 +1,184 @@
...
+            binding.progressCardBarCaption.setTextColor(accent)
+            binding.progressCardLinearBar.setProgressColor(accent)
+            binding.progressCardSemiCircle.progressColor = accent
+            binding.progressCardSemiCircle.valueTextColor = accent
+            binding.progressCardSemiCircle.descriptionTextColor = primary
+        }
+
+        private fun updateProgressDisplay() {
+            val fraction = if (progressMax > 0) progressCurrent.toFloat() / progressMax else 0f
+            binding.progressCardLinearBar.setProgress(fraction * 100f)
```
conroy-ricketts: Noticed that we have very similar logic in other parts of the codebase. Can we extract this to either a Kotlin extension or a utility function?

For example:
```kotlin
fun calculateProgress(current: Int, max: Int): Float {
    return if (max > 0) current.toFloat() / max else 0f
}
```
or:
```kotlin
fun Int.getProgressFraction(max: Int): Float = if (max > 0) this.toFloat() / max else 0f
```

Then, we could refactor our codebase to reuse this [here](https://github.com/dimagi/commcare-android/blob/042cafd3adf22c399a24596c0af42fa09cd4c99b/app/src/org/commcare/views/connect/SemiCircleProgressBar.kt#L125) and [here](https://github.com/dimagi/commcare-android/blob/042cafd3adf22c399a24596c0af42fa09cd4c99b/app/src/org/commcare/adapters/ConnectProgressJobSummaryAdapter.java#L43) as well. There might be other places where we can reuse a function like this

app/unit-tests/src/org/commcare/views/connect/ConnectProgressCardTest.kt
```diff
@@ -0,0 +1,274 @@
...
+    @Test
+    fun `infoCtaText shows the CTA and the listener fires on click`() {
+        val card = newCard()
+        val cta = card.findViewById<View>(R.id.progress_card_info_cta)
+
+        assertEquals(View.GONE, cta.visibility)
+
+        var clicked = false
+        card.infoCtaText = "Learn more"
+        card.setOnInfoCtaClickListener(View.OnClickListener { clicked = true })
```
conroy-ricketts: nit: `View.OnClickListener` is redundant

```suggestion
        card.setOnInfoCtaClickListener { clicked = true }
```

app/unit-tests/src/org/commcare/views/connect/ConnectProgressCardTest.kt
```diff
@@ -0,0 +1,274 @@
...
+
+        assertEquals(View.VISIBLE, cta.visibility)
+        cta.performClick()
+        org.junit.Assert.assertTrue(clicked)
+    }
+
+    @Test
+    fun `attributes inflate into properties`() {
+        val attrs =
+            org.robolectric.Robolectric
```
conroy-ricketts: nit: could import `Robolectric`

conroy-ricketts: same for line 259

conroy-ricketts (changes requested):
Looks fantastic! Just had a few comments

app/res/layout/view_connect_progress_card.xml
shubham1g5: definitely, think we want to abstract the complete set of dimensions in `dimens.xml` with the redesign work as the designers here have worked hard to standadize those across the figma designs and their work would be lost without us doing the same.

app/res/values/styles.xml
OrangeAndGreen: I learned a bit here as well. When a style use dot notation like this one, parent styles are automatically inherited from the parent component (i.e. Widget.CommCare here). Setting parent="" disables the behavior so nothing is inherited (preventing possible surprises later if something in the parent widget changes, even though in this case there isn't one right now).

app/src/org/commcare/views/connect/ConnectProgressAndStatusCard.kt
OrangeAndGreen: I mapped the colors to styles (commits were rebased so no link available)

app/res/layout/view_connect_progress_card.xml
OrangeAndGreen: I started adding some new dimensions that are intended to be reused across Connect pages and used them in this PR (including usage in some of the other newly created components). No commit to link since I rebased afterward.

app/res/values/strings.xml
OrangeAndGreen: It's not the same for each language, the word "of" gets translated

app/src/org/commcare/views/connect/ConnectProgressCard.kt
OrangeAndGreen: I think this is clearer now after some other changes involving component state. progressLabel is gone now.

app/src/org/commcare/views/connect/ConnectProgressCard.kt
OrangeAndGreen: I decided to just log it as I think it may happen under some reasonable scenarios, i.e. not setting max to force the bar to always be empty.

app/src/org/commcare/views/connect/ConnectProgressCard.kt
OrangeAndGreen: I created the helper and plugged it in (including the places you linked).

app/unit-tests/src/org/commcare/views/connect/ConnectProgressCardTest.kt
OrangeAndGreen: This went away, obsolete now

app/unit-tests/src/org/commcare/views/connect/ConnectProgressCardTest.kt
OrangeAndGreen: Done

app/src/org/commcare/views/connect/ConnectProgressCard.kt
OrangeAndGreen: This went away.

app/res/layout/view_connect_progress_card.xml
OrangeAndGreen: Done

app/res/values/dimens.xml
```diff
@@ -140,12 +140,37 @@
...
+    <dimen name="connect_text_title">16sp</dimen>
+    <dimen name="connect_text_metric">22sp</dimen>
+    <dimen name="connect_text_value">28sp</dimen>
+    <!-- Spacing / padding -->
+    <dimen name="connect_space_xs">4dp</dimen>
+    <dimen name="connect_space_sm">8dp</dimen>
+    <dimen name="connect_space_md">12dp</dimen>
+    <dimen name="connect_space_lg">16dp</dimen>
+    <!-- Corner radii -->
+    <dimen name="connect_radius_sm">5dp</dimen>
```
shubham1g5, line 173: confirming that the radius values are as per figma ? (slightly surprised that we are using odd values for dimensions)

shubham1g5 (dismissed):

app/res/values/dimens.xml
OrangeAndGreen: Ah, indeed, and linking the Figma MCP allowed Claude to find a couple other small inconsistencies. Addressed here: [ca0462c](https://github.com/dimagi/commcare-android/pull/3812/commits/ca0462cbfd2d4825094fe7b294af95ea5bce1cfa)

shubham1g5 (approved):

shubham1g5:
@Jignesh-dimagi @conroy-ricketts It's blocked on your review at the moment, it'd be great to prioritize review on this as it blocks some other tickets in current Sprint.

Jignesh-dimagi (approved):

conroy-ricketts (approved):

PR #3801, opened 2026-07-09
CCCT-2516 Add Connect CTA Bar View

OrangeAndGreen:
## Product Description

Adds `ConnectCtaBar`, a reusable bottom action bar for Connect screens. It shows a title/subtitle on the left and, on the right, either a CTA button (Main state) or a circular percent indicator (In-progress state). An optional non-dismissible banner can appear above the bar. This PR only introduces the component; it is not yet wired into any screen.

### Screenshots

**Main display**
CTA1_Main

**In-progress (progress circle) with popup message**

CTA1_ProgressAndMessage

## Technical Summary

The view extends `LinearLayout` and inflates its layout via a `<merge>` binding. State is driven by public properties (`titleText`, `subtitleText`, `buttonText`, `infoMessage`, `progress`), also settable via XML attributes. Setting `progress` swaps the button for the circular indicator; a null/empty value hides the corresponding element. The progress ring reuses the existing `CircleProgressBar`.

### Safety story

I previewed the layout in Android Studio to check the Main / In-progress / banner states against the Figma, and ran the view in the app to confirm rendering and state switching. The change is purely additive — a new view, its layout, and resources — and is not referenced by any existing screen, so it cannot affect current behavior or existing data. Risk to review: no production code path exercises the component yet, so integration behavior (sizing, click handling within a real screen) is only validated once it is wired in.

### Automated test coverage

`ConnectCtaBarTest` (Robolectric) covers title/subtitle binding and visibility, button text and click dispatch, the null-vs-set `progress` toggle between button and percent cluster, banner visibility for set/empty `infoMessage`, and inflation from XML attributes.

[CCCT-2516]: https://dimagi.atlassian.net/browse/CCCT-2516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3815, opened 2026-07-21
CCCT-2517 Refactor Opportunity Intro Screen To New Design

Jignesh-dimagi:
### [CCCT-2517](https://dimagi.atlassian.net/browse/CCCT-2517)

https://github.com/user-attachments/assets/b67fd44e-8328-45d7-843e-21896077843f

## Product Description

Redesigns the Connect opportunity introduction screen: a "New Opportunity" badge, an Expiry / Maximum Earnings header row, a tappable Learn Modules card that opens a modules bottom sheet, three delivery stat cards, and a sticky footer that starts learning.

## Technical Summary

The four cards reuse the existing `ConnectInfoCard` component; the Learn Modules card opens a new `BottomSheetDialogFragment` nav destination. 

### Safety story

- I ran it on a device and exercised the flow, and ran the unit tests.
- Scope is limited to the opportunity intro screen and one new bottom sheet; the four cards reuse an existing shared component.

### Automated test coverage

Robolectric tests drive the real views: the intro screen's rendered card values, the Learn Modules card click opening the sheet, the Start Learning click navigating to downloading, and the bottom sheet's per-module rows.

[CCCT-2517]: https://dimagi.atlassian.net/browse/CCCT-2517?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Jignesh-dimagi:
## Suggested Review Order

- `app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.java` — core screen logic + the Start-Learning guard change (key review point)
- `app/src/org/commcare/fragments/connect/ConnectLearnModulesBottomSheet.kt` — new modules bottom sheet; reads the active job
- `app/src/org/commcare/adapters/ConnectLearnModuleAdapter.kt` — module row adapter
- `app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java` — new symbol money helper + max-earnings source
- `app/src/org/commcare/connect/ConnectDateUtils.kt` — new short-date formatter
- `app/res/layout/fragment_connect_job_intro.xml` — redesigned screen layout
- `app/res/layout/fragment_connect_learn_modules_sheet.xml`, `app/res/layout/item_connect_learn_module.xml` — sheet + row layouts
- `app/res/navigation/nav_graph_connect.xml` — bottom sheet destination + action
- `app/res/values/strings.xml`, `colors.xml`, `styles.xml` (+ `values-*/strings.xml`) — new strings, badge color, sheet animation
- `app/unit-tests/.../ConnectJobIntroFragmentTest.kt`, `ConnectLearnModulesBottomSheetTest.kt`, `BaseConnectJobIntroTest.kt` — Robolectric coverage of render, clicks, and navigation

app/res/layout/fragment_connect_learn_modules_sheet.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:background="@color/connect_background_color"
+    android:orientation="vertical"
+    android:paddingHorizontal="16dp"
```
shubham1g5: As part of new layouts, we should try to  expose all dimensions value to `dimen.xml` and try to standardize things like padding/margin , components width/height etc as part of the process.

app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.java
```diff
@@ -122,7 +109,7 @@ private void startLearning() {
             @Override
             public void onFailure(@NonNull PersonalIdOrConnectApiErrorCodes errorCode, @Nullable Throwable t) {
                 reportApiCall(false);
-                if (!isAdded() || getView() == null) {
```
shubham1g5, line 125: are you able to outline reasons behind removing the `getView` check ?

app/src/org/commcare/fragments/connect/ConnectLearnModulesBottomSheet.kt
```diff
@@ -0,0 +1,71 @@
...
+        _binding = FragmentConnectLearnModulesSheetBinding.inflate(inflater, container, false)
+        return binding.root
+    }
+
+    override fun onViewCreated(
+        view: View,
+        savedInstanceState: Bundle?,
+    ) {
+        super.onViewCreated(view, savedInstanceState)
+        expandToFullHeight(view)
```
shubham1g5: Looking at [the designs here](https://docs.google.com/document/d/1rntc16FW2Jfr6CbzNcZ9RDOxIRrmfcWNU6_d5WL1rA8/edit?tab=t.xdhnuice1358#heading=h.6xomuhyo8has), bottom sheet only expands to a fixed height and takes the full height only when user scrolls up

RELEASES.md
```diff
@@ -94,6 +94,7 @@ These are published publicly on Playstore, Github Releases and CommCare Forums
 - [SMS Invite Links Open App] Clicking a Connect invite link in an SMS message opens the app and navigates to the opportunity
 - Launching an app from a Connect opportunity now opens it directly with a single loading dialog, instead of briefly flashing the login and app-setup screens
 - Image capture questions support a new `rectangle-overlay` appearance that shows a rectangular framing guide in the camera preview, helping users consistently frame the subject (e.g. a MUAC arm + tape).
+- Redesigned the Connect opportunity introduction screen.
```
shubham1g5: no need to add anything to this file for connect redesign items.

app/src/org/commcare/connect/ConnectDateUtils.kt
```diff
@@ -21,6 +21,16 @@ object ConnectDateUtils {
         return formatter.format(date)
     }
 
+    @JvmStatic
+    fun formatShortDate(date: Date): String {
```
shubham1g5: can be simplified by passing `format` as an argument to `formatDate`

app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.java
shubham1g5: would be nice to also convert this fragment to Kotlin given we are re-writing majority of it.

app/unit-tests/src/org/commcare/fragments/connect/ConnectJobIntroFragmentTest.kt
```diff
@@ -0,0 +1,213 @@
...
+    private fun launch(): ConnectJobIntroFragment = navigateToIntroFragment()
+
+    private fun cardValue(
+        fragment: ConnectJobIntroFragment,
+        cardId: Int,
+    ): String =
+        fragment
+            .requireView()
+            .findViewById<ConnectInfoCard>(cardId)
+            .findViewById<TextView>(R.id.info_card_value_text)
```
shubham1g5: nit: can be simplified to `findViewById<ConnectInfoCard>(cardId).valueText`

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
```diff
@@ -602,6 +603,19 @@ public String getMoneyString(int value) {
         return String.format(Locale.getDefault(), "%d%s", value, currency);
     }
 
+    public String getMoneyStringWithSymbol(int value) {
```
shubham1g5: this doesn't seem apt for a model class, think we should create a util like `ConnectMoneyUtils` to extract these sort of methods to.

shubham1g5 (changes requested):
Design implementation looks great overall except that the bottom bar design seems a bit off (height and button style)

shubham1g5:
This needs to be PR'ed to `feature/connect_redesign`

app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.java
Jignesh-dimagi: The getView() == null clause was added in [#3786](https://github.com/dimagi/commcare-android/pull/3786). On device it was returning null in the onSuccess path even though the fragment was still added and interactive, so the guard hit an early return and tapping Start never navigated to ConnectDownloadingFragment. This reproduces on master too (with the current design's Start button). 
I kept !isAdded() for the added-state safety that #3786 intended, and dropped only the getView() == null clause to unblock navigation. The root cause of getView() returning null here is still unclear and pre-existing — worth a separate look, but out of scope for this ticket.

app/res/layout/fragment_connect_learn_modules_sheet.xml
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3815/changes/dea63e147a6d8c1b379973036402c2c68846f5b1)

app/src/org/commcare/fragments/connect/ConnectLearnModulesBottomSheet.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3815/changes/dea63e147a6d8c1b379973036402c2c68846f5b1). Updated the video also.

RELEASES.md
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3815/changes/dea63e147a6d8c1b379973036402c2c68846f5b1).

app/src/org/commcare/connect/ConnectDateUtils.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3815/changes/dea63e147a6d8c1b379973036402c2c68846f5b1).

app/unit-tests/src/org/commcare/fragments/connect/ConnectJobIntroFragmentTest.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3815/changes/dea63e147a6d8c1b379973036402c2c68846f5b1).

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3815/changes/dea63e147a6d8c1b379973036402c2c68846f5b1).

Jignesh-dimagi:
> Design implementation looks great overall except that the bottom bar design seems a bit off (height and button style)

Yup! Updated the bottom bar design and added the new video to the description.

app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.java
Jignesh-dimagi: Happy to do that, https://github.com/dimagi/commcare-android/pull/3815/changes/9e56f12c1d8ee80b4c28d7afae027878571c5931

app/res/layout/fragment_connect_learn_modules_sheet.xml
shubham1g5: Values like ` <dimen name="dp10">10dp</dimen>` defeats the purpose of exporting things to `dimen.xml` as the idea here is to be able to standarize dimensions across components.

app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.java
shubham1g5: I see, @conroy-ricketts think  https://github.com/dimagi/commcare-android/pull/3786 would need to be reverted based on this before we code freeze.

app/src/org/commcare/fragments/connect/ConnectLearnModulesBottomSheet.kt
```diff
@@ -0,0 +1,73 @@
...
+        view.post {
+            val dialog = dialog as? BottomSheetDialog ?: return@post
+            val bottomSheet =
+                dialog.findViewById<FrameLayout>(
+                    com.google.android.material.R.id.design_bottom_sheet,
+                ) ?: return@post
+            bottomSheet.layoutParams =
+                bottomSheet.layoutParams.apply {
+                    height = ViewGroup.LayoutParams.MATCH_PARENT
+                }
```
shubham1g5: we should avoid accessing internal Android components, there is no guarantee that `com.google.android.material.R.id.design_bottom_sheet` will remain same across Android OS versions.

shubham1g5 (changes requested):

app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.java
conroy-ricketts: [AI - Claude Opus 4.8] Root cause found: the unqualified `getView()` in this guard does **not** resolve to `Fragment.getView()` — inside the anonymous `ConnectApiHandler` it binds to `BaseApiHandler`'s Kotlin `val view` getter, which is always `null` here (the handler is constructed with no `view`). So the guard's `getView() == null` was permanently true and the callback always early-returned, which is why Start silently did nothing. `isAdded()` was unaffected (no such method on the handler), which is why only the view half misbehaved. Confirmed on-device with logging.

Dropping the clause is correct. Companion fixes:
- #3818 — `master` (the Java version still has the broken clause, so Start is fully blocked there)
- #3819 — this branch (guards on the fragment's own view via `this@ConnectJobIntroFragment.view` and moves `hideError()` below the guard, restoring CI-766's destroyed-view protection)

app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.kt
```diff
@@ -0,0 +1,167 @@
...
+    ): View {
+        val view = super.onCreateView(inflater, container, savedInstanceState)
+        requireActivity().setTitle(R.string.connect_job_info_view_opportunity)
+
+        (requireActivity() as AppCompatActivity)
+            .supportActionBar
+            ?.setHomeAsUpIndicator(R.drawable.ic_connect_close)
+
+        binding.tvJobTitle.text = job.title
+        binding.tvJobDescription.text = job.description
```
conroy-ricketts: Should this rather be the job's short description? Noticed that the example in the design doc is a brief description rather than a long one

app/res/layout/fragment_connect_job_intro.xml
```diff
@@ -1,234 +1,307 @@
...
+                    <LinearLayout
+                        android:layout_width="0dp"
+                        android:layout_height="wrap_content"
+                        android:layout_weight="1"
+                        android:orientation="vertical">
+
+                        <TextView
+                            android:layout_width="wrap_content"
+                            android:layout_height="wrap_content"
+                            android:text="@string/connect_opportunity_expiry_label"
```
conroy-ricketts, line 92: Should this read "Expiry" or "Expires on"? 

Might be worth confirming the intention with Product because both are displayed on the design doc

conroy-ricketts: Actually, there's quite a few contradictions with the design, e.g.:

- "Learn" vs "Learn App Details"
- "Delivery" vs "Delivery App Details"
- "Expires on" vs "Expiry"
- "Complete training to begin visits" vs "Finish training to start visits"
- "Complete visits and submit forms to receive payments" vs "Give care, submit forms to earn payments"
- "Total Visits" vs "Maximum Visits"
- "Download Learn" vs "Download Learn App"

Which is correct?

I also added a comment on the design doc for Product's input

app/res/layout/fragment_connect_job_intro.xml
conroy-ricketts, line 1: Is there another ticket for the downloading state?:

Image

app/src/org/commcare/connect/ConnectMoneyUtils.kt
```diff
@@ -0,0 +1,24 @@
...
+            if (currency.isNullOrEmpty()) {
+                ""
+            } else {
+                try {
+                    Currency.getInstance(currency.uppercase(Locale.ROOT)).symbol
+                } catch (_: IllegalArgumentException) {
+                    currency
+                }
+            }
+        return String.format(Locale.getDefault(), "%s%d", symbol, value)
```
conroy-ricketts: This will sometimes return the number directly appended to the symbol's acronym if there is no symbol, which will appear odd for some users. Can we add a space between the symbol and the value if no symbol was found? For example, in this screenshot the currency should read as either "AMD 500" or "500 AMD"

Image

app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.kt
```diff
@@ -0,0 +1,167 @@
...
+
+        populateLearnCard()
+        populateDeliveryCards()
+
+        return view
+    }
+
+    override fun onDestroyView() {
+        super.onDestroyView()
+        (requireActivity() as AppCompatActivity).supportActionBar?.setHomeAsUpIndicator(0)
```
conroy-ricketts: Is there ever a case where the action bar would be null here?

app/src/org/commcare/fragments/connect/ConnectLearnModulesBottomSheet.kt
```diff
@@ -0,0 +1,73 @@
...
+            ConnectLearnModuleAdapter(job.learnAppInfo.learnModules)
+    }
+
+    private fun configureBottomSheet(view: View) {
+        view.post {
+            val dialog = dialog as? BottomSheetDialog ?: return@post
+            val bottomSheet =
+                dialog.findViewById<FrameLayout>(
+                    com.google.android.material.R.id.design_bottom_sheet,
+                ) ?: return@post
```
conroy-ricketts: Confirming that these null checks are necessary?

app/src/org/commcare/adapters/ConnectLearnModuleAdapter.kt
```diff
@@ -0,0 +1,44 @@
...
+    override fun onBindViewHolder(
+        holder: ModuleViewHolder,
+        position: Int,
+    ) {
+        val module = modules[position]
+        val context = holder.itemView.context
+        with(holder.binding) {
+            tvModuleName.text = "${position + 1}. ${module.name}"
+            tvModuleEstimate.text =
+                context.getString(R.string.connect_opportunity_module_estimated_time, module.timeEstimate)
```
conroy-ricketts: To polish this, can we make this singular if it's just 1 hour (i.e. "1 hour" vs "1 hours")

app/res/layout/fragment_connect_job_intro.xml
conroy-ricketts, line 1: Multiple Linear Layouts are being nested here. Can we simplify this to use a Constraint Layout instead for better performance and readability?

app/res/layout/fragment_connect_learn_modules_sheet.xml
conroy-ricketts, line 1: Can we tweak this new bottom sheet to have rounded corners at the top to match both the design doc and other bottom sheets in the app currently for consistent design language

Image
Image

app/res/anim/connect_bottom_sheet_slide_in.xml
conroy-ricketts, line 1: Confirming that these animation are necessary? They weren't needed for other bottom sheets in our app like job details in the jobs list for example

app/res/values/dimens.xml
```diff
@@ -123,6 +123,11 @@
     <dimen name="audio_button_size">55dp</dimen>
 
     <dimen name="dp6">6dp</dimen>
+    <dimen name="dp10">10dp</dimen>
+    <dimen name="dp12">12dp</dimen>
+    <dimen name="dp20">20dp</dimen>
+    <dimen name="dp22">22dp</dimen>
+    <dimen name="dp32">32dp</dimen>
```
conroy-ricketts: This seems like an extra step?

What's the context here?

app/res/values/styles.xml
```diff
@@ -318,4 +318,13 @@
...
     </style>
 
+    <style name="ConnectBottomSheetAnimation" parent="Animation.MaterialComponents.BottomSheetDialog">
+        <item name="android:windowEnterAnimation">@anim/connect_bottom_sheet_slide_in</item>
+        <item name="android:windowExitAnimation">@anim/connect_bottom_sheet_slide_out</item>
+    </style>
+
+    <style name="ConnectBottomSheetDialogTheme" parent="Theme.MaterialComponents.Light.BottomSheetDialog">
+        <item name="android:windowAnimationStyle">@style/ConnectBottomSheetAnimation</item>
+    </style>
```
conroy-ricketts, line 328: Similar comment as earlier - are these animations necessary? They weren't needed for our other bottom sheets

conroy-ricketts (changes requested):
Looks great so far! I haven't had a chance to look over the unit tests yet - will get back to this soon

app/res/layout/fragment_connect_job_intro.xml
OrangeAndGreen: Confirming the download stuff happens in other work. The UI part was done in [this PR](https://github.com/dimagi/commcare-android/pull/3801), and the actual enabling of download on the page happens in a future ticket.

app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.kt
```diff
@@ -0,0 +1,167 @@
...
+                        getString(R.string.error),
+                        error,
+                        false,
+                        R.string.ok,
+                    )
+                }
+            }
+
+            override fun onSuccess(success: Boolean) {
+                hideError()
```
OrangeAndGreen: I think this should happen after the check for isAdded

app/res/layout/fragment_connect_job_intro.xml
```diff
@@ -1,234 +1,307 @@
...
+                android:background="@color/white"
                 android:orientation="vertical"
-                android:padding="16dp">
+                android:padding="@dimen/standard_spacer_double">
 
-                <TextView
+                <LinearLayout
                     android:layout_width="wrap_content"
                     android:layout_height="wrap_content"
-                    android:text="@string/connect_job_learn_details"
```
OrangeAndGreen, line 86: Looks like some strings (like this one) are no longer used because of these changes, it would be nice to clean those up

OrangeAndGreen (commented):
Added a couple things but also I think Conroy brought up several good points.

app/res/layout/fragment_connect_job_intro.xml
Jignesh-dimagi: This was implemented based on the [Figma design](https://www.figma.com/design/4QLInPUv5wIfjfDGfrFESW/Opportunity-Home-Design-_-Ishwari?node-id=5257-18339&t=bK4bdG255sOV8rqY-0) (section: OpportunityHome_New Opportunity) and the screenshot attached to the ticket. As shown in the screenshot right below your comment in the design doc, it shares this exact design. I’ll keep it as-is since it appears to be correct, but I’m open to making changes if needed.

app/res/layout/fragment_connect_job_intro.xml
Jignesh-dimagi: Given the large number of UI components placed at a single level, I'm not sure if using ConstraintLayout is the best approach here.

Claude also replied:
The main perf benefit of ConstraintLayout comes from removing deep nesting and nested weights — this layout has neither (the weights are all single-level), so the gain here would be marginal, and the long vertical stack (badge → title → cards) actually reads cleanly as a LinearLayout.

app/res/anim/connect_bottom_sheet_slide_in.xml
Jignesh-dimagi: Yep, it's included in the new design! As shown in [Figma design](https://www.figma.com/design/4QLInPUv5wIfjfDGfrFESW/Opportunity-Home-Design-_-Ishwari?node-id=5257-15830&p=f&t=64ZJVdA91G64na04-0) at OpportunityHome_NewOpportunity_BottomSheet page, it initially displays as a half bottom sheet and expands as you scroll up. This transition is animated to prevent UI jerking, and the same smooth behavior applies when scrolling back down.

app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.kt
Jignesh-dimagi: Not sure what is your exact ask here but it's non-null for this screen today, but that depends on the host's layout, so the fragment guards it with ?. rather than assuming. Keeping the null-safe call as  AppCompatActivity.getSupportActionBar() is officially @Nullable.

app/res/values/styles.xml
Jignesh-dimagi: Yep for bottom sheet animation, as mentioned [above](https://github.com/dimagi/commcare-android/pull/3815#discussion_r3627657793)

app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.kt
shubham1g5: @Jignesh-dimagi that is a defensive programming approach and we would rather crash here in such scenarios given we know that the host is `ConnectActivity` (we already hard couple with `ConnectActivity in other fragments) and it hosts a action bar today. 

This approach around null-safety would be expectable if we know that the fragment is loaded by multiple hosts with one of the host not having a action bar but given we know that the fragment has a single host with action bar, and action bar being `null` would be a programmatic error or  edge case, think we should not accept `null` values here and remove the `?` and instead use `!!`

app/res/layout/fragment_connect_job_intro.xml
shubham1g5: @Jignesh-dimagi  can you refer to the [Design Doc](https://docs.google.com/document/d/1rntc16FW2Jfr6CbzNcZ9RDOxIRrmfcWNU6_d5WL1rA8/edit?tab=t.xdhnuice1358#heading=h.8ghlb2mbk1zz) and make sure the strings are per the latest designs. 

The design doc is source of truth for us and when there is discrepancy between figma and doc.

app/res/layout/fragment_connect_learn_modules_sheet.xml
Jignesh-dimagi: I have standardize now https://github.com/dimagi/commcare-android/pull/3815/changes/3a65d58d4a361d3f71489f129a814e51bbf8066b

app/src/org/commcare/fragments/connect/ConnectLearnModulesBottomSheet.kt
Jignesh-dimagi: Makes sense, changed https://github.com/dimagi/commcare-android/pull/3815/changes/3a65d58d4a361d3f71489f129a814e51bbf8066b

app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.kt
Jignesh-dimagi: yeah, changed https://github.com/dimagi/commcare-android/pull/3815/changes/3a65d58d4a361d3f71489f129a814e51bbf8066b

app/res/layout/fragment_connect_job_intro.xml
Jignesh-dimagi: I think the confusion here was that even the design doc had two versions. To be safe, I went with the design that was consistent across the design doc, Figma, and the Jira ticket. Kim has since replied in the doc with the correct version date, so it's clear now! I've updated the implementation accordingly. https://github.com/dimagi/commcare-android/commit/3a65d58d4a361d3f71489f129a814e51bbf8066b

app/src/org/commcare/connect/ConnectMoneyUtils.kt
Jignesh-dimagi: I added a space for non-symbol characters, but kept it without a space if it's a symbol. https://github.com/dimagi/commcare-android/commit/3a65d58d4a361d3f71489f129a814e51bbf8066b

app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.kt
Jignesh-dimagi: Makes sense! I don't have any references showing this will be hosted by another activity in future implementations, so I've updated it. I also applied the change inside onCreate as well. https://github.com/dimagi/commcare-android/commit/3a65d58d4a361d3f71489f129a814e51bbf8066b

app/src/org/commcare/adapters/ConnectLearnModuleAdapter.kt
Jignesh-dimagi: I've added singular and plural support for hours, payment units, and learn modules.

Note: I initially omitted these because Claude suggested avoiding them, citing that it isn't standard practice in our codebase. https://github.com/dimagi/commcare-android/commit/3a65d58d4a361d3f71489f129a814e51bbf8066b

app/res/layout/fragment_connect_learn_modules_sheet.xml
Jignesh-dimagi: Done https://github.com/dimagi/commcare-android/commit/3a65d58d4a361d3f71489f129a814e51bbf8066b

app/res/values/dimens.xml
Jignesh-dimagi: It's removed now. I had added it because I misinterpreted Shubham's comment.

app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.kt
Jignesh-dimagi: Yeah I agree, I commented same [here](https://github.com/dimagi/commcare-android/pull/3819/changes#r3628865771).

Jignesh-dimagi:
Note: Updated ConnectInfoCard and its style per Kim's feedback [here](https://dimagi.slack.com/archives/C09H7D1V1NK/p1784650024633149?thread_ts=1784642290.749409&cid=C09H7D1V1NK).

app/res/layout/fragment_connect_job_intro.xml
Jignesh-dimagi: Cleaned up all unused strings resulting from these changes.

app/src/org/commcare/fragments/connect/ConnectLearnModulesBottomSheet.kt
Jignesh-dimagi: Updated to use parent instead, but the null check is still necessary. If the sheet is dismissed or detached before the posted runnable executes, view.parent will be null.

app/res/layout/fragment_connect_job_intro.xml
conroy-ricketts: Ah thanks for the context! @Jignesh-dimagi can we reuse that UI component here as well to avoid code duplication

conroy-ricketts: Downgrading this to a nit

app/res/anim/connect_bottom_sheet_slide_in.xml
conroy-ricketts: Aaah I see, cool!

app/res/layout/fragment_connect_job_intro.xml
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3815/changes/0f8d0cdae5ba2b818b60202b788aebc477e6ce57).

app/res/layout/fragment_connect_job_intro.xml
```diff
@@ -1,234 +1,261 @@
...
+                    app:titleText="@string/connect_opportunity_max_earnings_card_label" />
+            </LinearLayout>
+        </LinearLayout>
+    </androidx.core.widget.NestedScrollView>
+
+    <org.commcare.views.connect.ConnectCtaBar
+        android:id="@+id/connect_intro_cta_bar"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_margin="10dp"
```
conroy-ricketts: Suggest removing the margin here - it's creating unnecessary white space around the bar

Image

app/unit-tests/src/org/commcare/fragments/connect/BaseConnectJobIntroTest.kt
```diff
@@ -0,0 +1,192 @@
...
+        every { repository.getOpportunities(any(), any()) } returns emptyFlow()
+
+        mockkStatic(FirebaseAnalyticsUtil::class)
+        every { FirebaseAnalyticsUtil.getNavControllerPageChangeLoggingListener() } returns
+            object : NavController.OnDestinationChangedListener {
+                override fun onDestinationChanged(
+                    controller: NavController,
+                    destination: NavDestination,
+                    arguments: Bundle?,
+                ) = Unit
```
conroy-ricketts, line 62: nit: this can be simplified to a lambda for improved readability

app/unit-tests/src/org/commcare/fragments/connect/BaseConnectJobIntroTest.kt
```diff
@@ -0,0 +1,192 @@
...
+                .start()
+                .resume()
+                .get()
+
+        navHostFragment =
+            activity.supportFragmentManager
+                .findFragmentById(R.id.nav_host_fragment_connect) as NavHostFragment
+
+        job = buildJob()
+        activity.setActiveJob(job)
```
conroy-ricketts, line 80: nit: can use kotlin property syntax

```suggestion
        activity.activeJob = job
```

app/unit-tests/src/org/commcare/fragments/connect/BaseConnectJobIntroTest.kt
```diff
@@ -0,0 +1,192 @@
...
+        activity.runOnUiThread {
+            navController.navigate(
+                R.id.action_connect_jobs_list_fragment_to_connect_job_intro_fragment,
+            )
+        }
+        ShadowLooper.idleMainLooper()
+        return navHostFragment.childFragmentManager.primaryNavigationFragment as ConnectJobIntroFragment
+    }
+
+    protected fun showBottomSheet(fragment: androidx.fragment.app.DialogFragment) {
```
conroy-ricketts, line 103: nit: can import `DialogFragment`

app/unit-tests/src/org/commcare/fragments/connect/ConnectLearnModulesBottomSheetTest.kt
```diff
@@ -0,0 +1,79 @@
...
+@RunWith(AndroidJUnit4::class)
+class ConnectLearnModulesBottomSheetTest : BaseConnectJobIntroTest() {
+    private fun showSheet(): ConnectLearnModulesBottomSheet {
+        val sheet = ConnectLearnModulesBottomSheet()
+        showBottomSheet(sheet)
+        return sheet
+    }
+
+    private fun recyclerView(sheet: ConnectLearnModulesBottomSheet): RecyclerView {
+        val rv = sheet.requireView().findViewById<RecyclerView>(R.id.rv_modules)
```
conroy-ricketts, line 30: nit: suggest either renaming `rv` to `recyclerView` or we can use kotlin to avoid the need for a variable entirely 

```kotlin
    return sheet.requireView().findViewById<RecyclerView>(R.id.rv_modules).apply {
        measure(0, 0)
        layout(0, 0, 1000, 4000)
    }
```

conroy-ricketts (commented):
Just had one more comment that should be addressed for the UI - the rest are just nits for the unit tests

app/res/layout/fragment_connect_job_intro.xml
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3815/changes/4c147da79764ccbd99cad49bb62ae5c1ee08dc65)

conroy-ricketts (approved):
🔥

app/src/org/commcare/connect/ConnectMoneyUtils.kt
```diff
@@ -0,0 +1,28 @@
...
+            return value.toString()
+        }
+        val code = currency.uppercase(Locale.ROOT)
+        val symbol =
+            try {
+                Currency.getInstance(code).symbol
+            } catch (_: IllegalArgumentException) {
+                code
+            }
+        return if (symbol == code) {
```
OrangeAndGreen, line 22: Nit: Not a big deal but the logic here feels a bit difficult to read. Is it possible for symbol==code without hitting the IllegalArgumentException, i.e. for Currency.getInstance(code).symbol == code?

If not it might be more straight-forward to move the String.format return calls into the try-catch, i.e.
try { return String.format("%s%d", Currency.getInstance(code).symbol, value) }
catch (...) { return String.format("%d %s", value, code) }

OrangeAndGreen (approved):

app/unit-tests/src/org/commcare/fragments/connect/BaseConnectJobIntroTest.kt
```diff
@@ -0,0 +1,192 @@
...
+
+        mockkStatic(MessageManager::class)
+        every { MessageManager.retrieveMessages(any(), any()) } returns Unit
+
+        // The jobs-list start destination fetches opportunities on boot; stub the repository so no
+        // real network call is made (which crashes background coroutines under Robolectric).
+        mockkObject(ConnectRepository.Companion)
+        val repository = mockk<ConnectRepository>(relaxed = true)
+        every { ConnectRepository.getInstance(any()) } returns repository
+        every { repository.getOpportunities(any(), any()) } returns emptyFlow()
```
shubham1g5, line 53: do we need to mock `ConnectRepository` for these tests ?

app/src/org/commcare/connect/ConnectMoneyUtils.kt
shubham1g5: +1

app/unit-tests/src/org/commcare/fragments/connect/ConnectJobIntroFragmentTest.kt
```diff
@@ -0,0 +1,221 @@
...
+        )
+    }
+
+    @Test
+    fun `tapping the footer button starts learning and navigates to downloading`() {
+        mockkStatic(ConnectUserDatabaseUtil::class)
+        every { ConnectUserDatabaseUtil.getUser(any()) } returns mockk<ConnectUserRecord>()
+        mockkStatic(ConnectJobUtils::class)
+        every { ConnectJobUtils.upsertJob(any()) } returns Unit
+        mockkStatic(AppUtils::class)
```
shubham1g5: DB layer should be working as it is in tests without needing to mock.

shubham1g5 (changes requested):
confirming that all layout dimenstions are taken from Figma values now ?

app/src/org/commcare/connect/ConnectMoneyUtils.kt
Jignesh-dimagi: Good point but has reason for this implementation.

Currency.getInstance(code) only throws when the code isn't a real currency code. .symbol never throws — it looks up the currency's symbol in the device's locale data. If that data has no symbol for the currency, it just returns the code back. So there are three cases:
 
- USD → symbol is $ → we show $100
- KES (Kenyan Shilling) → no symbol available, so symbol comes back as KES → we show 100 KES
- XYZ (not a real code) → exception → we show 100 XYZ
 
The symbol == code check is what sends the middle case (valid code, but no symbol) to the 100 KES format.

With your version, KES wouldn't throw, so it would go into the %s%d branch and print KES100 — losing the space and putting the code in front. That's the case the current if is there to handle.

app/unit-tests/src/org/commcare/fragments/connect/BaseConnectJobIntroTest.kt
Jignesh-dimagi: Yes, we need it. Our base test boots a real ConnectActivity, and its start screen is the jobs list. As soon as that screen loads, it calls getOpportunities() through ConnectRepository, which triggers a real network fetch on a background coroutine. So we're mocking the repository to return an empty flow.

Jignesh-dimagi:
> Design implementation looks great overall except that the bottom bar design seems a bit off (height and button style)

Yes, I have updated the individual components and overall UI to match Figma for padding, font sizes, and card designs. I've also uploaded an updated video in the description. https://github.com/dimagi/commcare-android/pull/3815/changes/0549560e55caba658fe1f3eea874aae96379969e

app/unit-tests/src/org/commcare/fragments/connect/ConnectJobIntroFragmentTest.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3815/changes/0549560e55caba658fe1f3eea874aae96379969e)

app/src/org/commcare/connect/ConnectMoneyUtils.kt
shubham1g5: makes sense.

shubham1g5 (approved):

PR #3818, opened 2026-07-21
CCCT-2517 Fix Start Learning Navigation

conroy-ricketts:
## Product Description

Tapping Start on a new Connect opportunity's introduction screen now proceeds to learning (downloading the learn app, or launching it if already installed) instead of silently doing nothing.

## Technical Summary

The guard in the start-learning callbacks called an unqualified `getView()`, which resolved to `BaseApiHandler`'s Kotlin `view` property (always null here) rather than `Fragment.getView()`, so every callback early-returned. Qualifying it to the fragment restores the guard, and `hideError()` moved below it. `ConnectJobUtils.upsertJob` now sources the application context itself, so persistence no longer depends on the fragment's lifecycle.

### Safety story

What gives confidence:
- Built on a device and confirmed tapping Start navigates to the downloading screen.
- Scoped to the start-learning callbacks and the shared `upsertJob` helper.

Risks to review:
- No automated coverage; verified manually.
- Re-enables navigation/launch/error paths that had been silently dead since CI-766.
- `upsertJob` no longer takes a `Context` param; its other two callers were updated.

[CCCT-2517]: https://dimagi.atlassian.net/browse/CCCT-2517?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.java
+++ b/app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.java
@@ -1,5 +1,6 @@
 package org.commcare.fragments.connect;
 
+import android.content.Context;
 import android.os.Bundle;
 import android.text.TextUtils;
 import android.view.LayoutInflater;
@@ -116,6 +117,7 @@ public class ConnectJobIntroFragment extends ConnectJobFragment<FragmentConnectJ
     }
 
     private void startLearning() {
+        Context appContext = requireContext().getApplicationContext();
         ConnectUserRecord user = ConnectUserDatabaseUtil.getUser(getContext());
 
         new ConnectApiHandler<Boolean>() {
@@ -143,7 +145,7 @@ public class ConnectJobIntroFragment extends ConnectJobFragment<FragmentConnectJ
                 reportApiCall(success);
 
                 job.setStatus(ConnectJobRecord.STATUS_LEARNING);
-                ConnectJobUtils.upsertJob(getContext(), job);
+                ConnectJobUtils.upsertJob(appContext, job);
 
                 if (!isAdded() || ConnectJobIntroFragment.this.getView() == null) {
                     return;
--- a/app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.java
+++ b/app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.java
@@ -1,6 +1,5 @@
 package org.commcare.fragments.connect;
 
-import android.content.Context;
 import android.os.Bundle;
 import android.text.TextUtils;
 import android.view.LayoutInflater;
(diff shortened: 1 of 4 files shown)


app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.java
```diff
@@ -140,16 +140,17 @@ public void onFailure(@NonNull PersonalIdOrConnectApiErrorCodes errorCode, @Null
...
             public void onSuccess(Boolean success) {
-                hideError();
                 reportApiCall(success);
 
                 job.setStatus(ConnectJobRecord.STATUS_LEARNING);
                 ConnectJobUtils.upsertJob(getContext(), job);
 
-                if (!isAdded() || getView() == null) {
+                if (!isAdded() || ConnectJobIntroFragment.this.getView() == null) {
                     return;
```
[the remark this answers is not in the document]
conroy-ricketts: [AI - Claude Opus 4.8] Good catch — valid. `getContext()` returns null once the fragment detaches (the same back-out-while-in-flight race), and it flows into `JobStoreManager`/`getConnectDbOpenHelper(context)`. Captured `requireContext().getApplicationContext()` up front and use it for `upsertJob()`, so the status persists regardless of fragment lifecycle while the UI guard still gates the view work. Fixed in 4593cd469.

app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.java
```diff
@@ -140,16 +142,17 @@ public void onFailure(@NonNull PersonalIdOrConnectApiErrorCodes errorCode, @Null
 
             @Override
             public void onSuccess(Boolean success) {
-                hideError();
                 reportApiCall(success);
 
                 job.setStatus(ConnectJobRecord.STATUS_LEARNING);
-                ConnectJobUtils.upsertJob(getContext(), job);
+                ConnectJobUtils.upsertJob(appContext, job);
```
shubham1g5: we can rather just get the appContext directly inside `upsertJob`

shubham1g5 (dismissed):

OrangeAndGreen (dismissed):

app/src/org/commcare/fragments/connect/ConnectJobIntroFragment.java
conroy-ricketts: https://github.com/dimagi/commcare-android/commit/121c4bc3e16ea6d89e0d201e472f3f268fd0d0b4

shubham1g5 (approved):

PR #3822, opened 2026-07-22
CCCT-2678 Add Connect Sync Status Card

OrangeAndGreen:
## Technical Summary

The former `ConnectProgressAndStatusCard` combined progress and sync status in one card; the redesign splits them into two. This PR adds the standalone sync-status card (the sync row is being removed from the progress card in the CCCT-2542 PR).

- New `ConnectSyncStatusCard`: a `CardView` subclass following the existing `ConnectInfoCard` pattern (ViewBinding, style-driven colors with `ContextCompat` fallbacks).
- Appearance is content-property driven (no state enum): `statusText`, optional `statusSubtext`, and a `warning` boolean that toggles the badge between OK (green check) and warning (amber refresh).
- Badge/icon colors are exposed via a `declare-styleable` + default `Widget.CommCare.ConnectSyncStatusCard` style so consumers can override per-instance.

Sample cards showing both states per July 28 update [here](https://docs.google.com/document/d/1rntc16FW2Jfr6CbzNcZ9RDOxIRrmfcWNU6_d5WL1rA8/edit?tab=t.xdhnuice1358#heading=h.x6enzildo19r):
image

### Safety story

What gives me confidence:
- I added demo cards to the Learning Progress page to test the views (see above)
- I ran the 9 Robolectric unit tests locally (pass under JDK 17), covering the card's binding, show/hide, and both badge appearances.
- The diff is purely additive — all new files, no existing code paths modified.
- The component has no screen consumers yet (adoption is out of scope), so it cannot affect any production flow.
- It mirrors the established `ConnectInfoCard` / `ConnectSuccessFailureCard` conventions, including badge icon tinting via `ImageViewCompat`.

Risks to review:
- Not yet wired into any layout, so it is unexercised inside a real screen.
- Visual details (spacing, badge sizing, colors) are asserted only by unit tests, not rendered in the running app.

### Automated test coverage

9 Robolectric tests in `ConnectSyncStatusCardTest`: `statusText` binding; `statusSubtext` show/hide; default OK badge (check icon + light-green); `warning` switch (refresh icon + light-amber); per-instance style color override; attribute inflation.

[CCCT-2678]: https://dimagi.atlassian.net/browse/CCCT-2678?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


OrangeAndGreen:
## Suggested Review Order

- `app/res/values/attrs.xml` — new styleable declaring the card's attributes
- `app/res/values/styles.xml` — default style supplying the badge/icon colors
- `app/res/values/colors.xml` — new light-amber color for the warning badge
- `app/res/drawable/bg_connect_sync_badge.xml` — circular badge background
- `app/res/layout/view_connect_sync_status_card.xml` — the card's view hierarchy and ids
- `app/src/org/commcare/views/connect/ConnectSyncStatusCard.kt` — component logic tying the above together
- `app/unit-tests/src/org/commcare/views/connect/ConnectSyncStatusCardTest.kt` — behavior verification

app/res/values/styles.xml
```diff
@@ -318,4 +318,11 @@
         <item name="failureAccentColor">@color/connect_red</item>
     </style>
 
+    <style name="Widget.CommCare.ConnectSyncStatusCard" parent="">
```
shubham1g5: think this needs to have a parent style corresponding to the card view like `Widget.MaterialComponents.CardView" or `Widget.AppCompat.CardView`

app/src/org/commcare/views/connect/ConnectSyncStatusCard.kt
```diff
@@ -0,0 +1,112 @@
...
+
+        /** Render [state], replacing whatever was previously shown. */
+        fun bind(state: State) {
+            this.state = state
+            binding.syncCardText.text = state.statusText
+            bindOptionalText(binding.syncCardSubtext, state.statusSubtext)
+            applyAppearance(state.warning)
+        }
+
+        private fun bindOptionalText(
```
shubham1g5: This is getting duplicated across views and think it would be beneficial to extract this in a `TextViewExtensions.kt` so that we can directly call `binding.syncCardSubtext.bindOptional(state.statusSubtext).` on these.

app/src/org/commcare/views/connect/ConnectSyncStatusCard.kt
```diff
@@ -0,0 +1,112 @@
...
+            value: CharSequence?,
+        ) {
+            view.text = value
+            view.visibility = if (value.isNullOrEmpty()) GONE else VISIBLE
+        }
+
+        private fun applyAppearance(warning: Boolean) {
+            val badgeColor = if (warning) syncWarningBadgeColor else syncOkBadgeColor
+            val iconColor = if (warning) syncWarningIconColor else syncOkIconColor
+            val iconRes = if (warning) R.drawable.ic_connect_directory_sync else R.drawable.check_update
```
shubham1g5: we should avoid if/else based on states when there is a strong possibility of more states to be added like `error` in this case, think the right pattern to follow in these cases is to abstract the state as - 

````
  sealed class Appearance {                                                                                                                                            
      object Ok : Appearance()                                                                                                                                         
      object Warning : Appearance()                                                                                                                                    
  }
````

and then using a `when(appearance)` block to set the properties here.

File: app/res/layout/view_connect_sync_status_card.xml [hunk 4]
```diff
@@ -0,0 +1,51 @@
...
+    <ImageView
+        android:id="@+id/sync_card_icon"
+        android:layout_width="@dimen/connect_sync_card_badge_size"
+        android:layout_height="@dimen/connect_sync_card_badge_size"
+        android:background="@drawable/bg_connect_sync_badge"
+        android:backgroundTint="@color/connect_light_green"
+        android:contentDescription="@null"
+        android:padding="@dimen/connect_sync_card_badge_icon_inset"
+        android:src="@drawable/check_update"
+        app:tint="@color/connect_green" />
```
shubham1g5: nit:  Would setting backgroundTint and app:tint  here takes any effect given these values are overriden in `applyAppearance` as - ` ImageViewCompat.setImageTintList` ?

app/src/org/commcare/views/connect/ConnectSyncStatusCard.kt
```diff
@@ -0,0 +1,112 @@
...
+                attrs,
+                R.styleable.ConnectSyncStatusCard,
+                defStyleAttr,
+                R.style.Widget_CommCare_ConnectSyncStatusCard,
+            ) {
+                okBadge = getColor(R.styleable.ConnectSyncStatusCard_syncOkBadgeColor, okBadge)
+                okIcon = getColor(R.styleable.ConnectSyncStatusCard_syncOkIconColor, okIcon)
+                warningBadge = getColor(R.styleable.ConnectSyncStatusCard_syncWarningBadgeColor, warningBadge)
+                warningIcon = getColor(R.styleable.ConnectSyncStatusCard_syncWarningIconColor, warningIcon)
+                initialState = readStateFromAttributes(this)
```
shubham1g5: Think we should replace all color defaults as `0`  ([more context here](https://github.com/dimagi/commcare-android/pull/3805#discussion_r3569360123))

app/src/org/commcare/views/connect/ConnectSyncStatusCard.kt
shubham1g5: nit: We can can further abstract `iconRes` in the style attrs as well and have this card as a more generalised component that can be potentially used for other simlar components (with ok, warning, error) states as well.

shubham1g5 (changes requested):

app/res/values/styles.xml
OrangeAndGreen: Done — set \`parent="CardView"\` so the style inherits the framework card defaults instead of \`parent=""\`. Addressed in f7fe06764.

app/src/org/commcare/views/connect/ConnectSyncStatusCard.kt
OrangeAndGreen: Extracted into \`views/extensions/TextViewExtensions.kt\` as \`TextView.bindOptional(text)\` and adopted it here plus across \`ConnectCtaBar\`, \`ConnectInfoCard\`, \`ConnectInfoHalfCard\` and \`SemiCircleProgressBar\`, removing the duplicated helpers. Addressed in 1069f379c.

app/src/org/commcare/views/connect/ConnectSyncStatusCard.kt
OrangeAndGreen: Replaced the if/else with a sealed \`Appearance\` type (Ok/Warning) resolved via an exhaustive \`when(appearance)\`, backed by an \`AppearanceStyle\` holder — so adding an \`error\` state later is compile-checked. Addressed in eed4d7c9f.

app/res/layout/view_connect_sync_status_card.xml
OrangeAndGreen: Good catch — they were overridden at runtime in \`applyAppearance\` (\`ImageViewCompat.setImageTintList\` / \`backgroundTintList\`), so they had no effect. Moved them (and \`src\`) to \`tools:\` preview-only attrs and dropped the now-unused \`xmlns:app\`. Addressed in 1455b00b9.

app/src/org/commcare/views/connect/ConnectSyncStatusCard.kt
OrangeAndGreen: Switched all color defaults to \`0\`, resolving the actual values from the widget style via \`defStyleRes\` (matching the \`ConnectSuccessFailureCard\` convention). Addressed in 881eb7d70.

app/src/org/commcare/views/connect/ConnectSyncStatusCard.kt
OrangeAndGreen: Added \`syncOkIcon\`/\`syncWarningIcon\` reference attrs read via \`getResourceId\`, so the icons are style-driven like the colors. Addressed in b67792707.

shubham1g5 (dismissed):

app/res/values/styles.xml
shubham1g5: `CardView` is a class name and not a style right ?

Think the parent should be `Widget.MaterialComponents.CardView` here.

OrangeAndGreen: That style defaults to androidx.cardview which I think is what we want (see [here](https://github.com/dimagi/commcare-android/blob/6e1006d1127788a233f11b8f965f9433306f06a5/app/src/org/commcare/views/connect/ConnectInfoCard.kt#L7))

conroy-ricketts (dismissed):

shubham1g5 (approved):

app/res/values/styles.xml
shubham1g5: Ahh not really, that's a view and we need parent to be a style.

shubham1g5 (changes requested):
Actually blocking on https://github.com/dimagi/commcare-android/pull/3822#discussion_r3656109020

Jignesh-dimagi:
@OrangeAndGreen @shubham1g5 
We're currently getting feedback on individual components only after full layouts are posted to the demo channel. To avoid this delay, can we start sharing individual components on the channel earlier to get quicker feedback?

app/res/values/styles.xml
OrangeAndGreen: Hmm, I think in this case it's a style and not a view. When I mouse over it I see @style/CardView:
image

If I change it as you suggested I see the different style, i.e.:
image

The difference is that we're using the androidx version of the CardView, not Material.

OrangeAndGreen: Ah, and now after reworking the card to be clickable for the latest design changes we're using a Material CardView after all! The parent style is set accordingly

OrangeAndGreen:
I reworked this to reflect design changes, see the July 28 update [here](https://docs.google.com/document/d/1rntc16FW2Jfr6CbzNcZ9RDOxIRrmfcWNU6_d5WL1rA8/edit?tab=t.xdhnuice1358#heading=h.x6enzildo19r) and linked Figma components. The screenshot in the description is also updated.

app/res/values/styles.xml
shubham1g5: that makes sense, thanks for clarifying the 2 styles.

app/src/org/commcare/views/connect/ConnectSyncStatusCard.kt
```diff
@@ -0,0 +1,147 @@
...
+import android.content.res.ColorStateList
+import android.content.res.TypedArray
+import android.util.AttributeSet
+import android.view.LayoutInflater
+import androidx.annotation.ColorInt
+import androidx.annotation.DrawableRes
+import androidx.core.content.ContextCompat
+import androidx.core.content.withStyledAttributes
+import androidx.core.widget.ImageViewCompat
+import com.google.android.material.card.MaterialCardView
```
shubham1g5, line 13: seems like we now are inconsitent between using `com.google.android.material.card.MaterialCardView` vs  `androidx.cardview.widget.CardView`. Curious when should we go for one vs other here and whether we should move other cards to `com.google.android.material.card.MaterialCardView`  as well ?

shubham1g5 (approved):

Jignesh-dimagi:
Looks great! I don't have any additional comments.

Jignesh-dimagi (approved):

conroy-ricketts (approved):

app/src/org/commcare/views/connect/ConnectSyncStatusCard.kt
OrangeAndGreen: Yeah that's a good point. In this case Claude decided to switch to the more decorated component (Material) when we changed the card to be clickable, as apparently we get some improved visual effects (ripple, animations, etc.) that way.

I think we should probably prefer the Material version going forward for the improvements it adds including more states and theming.

PR #3830, opened 2026-07-25
CCCT-2643 Add Unit Tests For Existing Home Screen

OrangeAndGreen:
## Technical Summary
Characterization (behavior-pinning) tests for the existing home screen, added ahead of the Opportunity Home / optional-app-seating refactor so the refactor is caught if it changes current behavior. Coverage centers on the refactor-risk surface: drawer navigation + app switching, the Connect job tile and hidden-button rules, launch-intent construction, menu visibility/routing, the grid buttons, sync-button network branching, session navigation, and pin/result flows.

Tests run on the existing Robolectric harness. Home is taken to a laid-out window, so the suites read and click the views the screen actually renders rather than calling activity methods: `clickHomeButton(home, label)` clicks the same `R.id.card` the adapter bound the production listener to. Connect state is seated by writing to the real Connect DB, and the Connect job-progress fetch runs against a `MockWebServer`.

No production code is changed.

### Safety story
What gives me confidence:
- Test-only change — no production files touched.
- Full app unit-test suite green locally: 141 test classes, 954 tests, 0 failures.
- The suites assert observable outputs (rendered views, outbound intents, HTTP requests, DB state, toasts, notification text) and were verified against the production source while being written, so they pin real current behavior rather than tautologies.

Risks to review:
- Reading and clicking the button grid depends on Robolectric laying out the RecyclerView. It does, but a Robolectric upgrade that changes layout behaviour would take these rows with it.
- Coverage is deliberately incomplete in two places, both noted in the suites: `GET_CASE` session navigation needs a case-list app rather than the single-form `form_nav_tests` fixture, and `RootMenuHomeActivity` is out of scope.

### Automated test coverage
Adds Robolectric suites under `org.commcare.activities`, plus three shared test utilities. These will fail if the corresponding behaviors change during the Opportunity Home refactor.

Shared:
- `BaseHomeScreenActivityTest` — installs a test app + user and builds home, either stopped at `onCreate` or laid out and clickable. The only thing stubbed is `ConnectivityStatus` (the device's radios); everything else, Connect DB included, runs for real.
- `ConnectTestUtils` — seats a Connect job and signs in a PersonalId user by writing the rows a Connect sync would leave, so production lookups resolve them.
- `ActivityAssertions` — assertions on the activities an activity launched.
- `ConnectMockApiServer` — points `ConnectNetworkClient` at a local `MockWebServer`.

Suites:
- `HomeDrawerNavigationTest` — app switching via `COMMCARE_APPS`, including the no-op cases.
- `HomeConnectDrawerNavigationTest` — Opportunities / Messaging / Work History under both managed and legacy login.
- `HomeConnectTileTest` — the Connect job tile: hidden-button rules, job-card visibility and title, and the message card's text and icon.
- `HomeConnectJobProgressTest` — clicking the Connect button opens the job status screen; clicking sync fetches delivery progress over HTTP and writes the parsed result back to the job, and the cases where it must not fetch at all.
- `StandardHomePresentationTest` — the behavior getters this host answers for its base activity, and the full set of buttons its grid renders.
- `HomeMenuTest` — options-menu item visibility for normal and demo users, and the routing for every item `onOptionsItemSelected` handles.
- `HomeButtonsTest` — each grid button, driven by clicking it: where it goes, the sync button's offline toast and notification text, and the sub-text strip.
- `HomeSessionNavigationTest` — home's end of the session-navigation loop: the callout to the root menu, the stale-result guard, and the advance into form entry.
- `HomeLoginLaunchChecksTest` — which login-launch check claims the launch and the boolean it returns, plus `buildHomeLaunchIntent`.
- `HomeActivityResultTest` — the PIN arms of `onActivityResultSessionSafe`.
- `HomeInstanceStateTest` — the launch/nav instance state that must survive `recreate()`.

[CCCT-2643]: https://dimagi.atlassian.net/browse/CCCT-2643?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


OrangeAndGreen:
## Suggested Review Order

- `app/src/org/commcare/activities/StandardHomeActivityUIController.java` — the only production change; two `@VisibleForTesting` widenings, no behavior change
- `app/unit-tests/src/org/commcare/activities/HomeScreenActivityTest.kt` — shared base (app install/login + Connect static mocks + helpers) every suite builds on
- `app/unit-tests/src/org/commcare/activities/HomeDrawerNavigationTest.kt` — highest refactor-risk: app-seating + drawer nav routing
- `app/unit-tests/src/org/commcare/activities/HomeConnectTileTest.kt` — Connect job tile visibility + hidden-button rules
- `app/unit-tests/src/org/commcare/activities/HomeIntentBuilderTest.kt` — home launch-intent construction
- `app/unit-tests/src/org/commcare/activities/StandardHomeActivityBehaviorTest.kt` — pinned flag getters + Connect delegate behavior
- `app/unit-tests/src/org/commcare/activities/HomeMenuTest.kt` — options-menu visibility + item routing
- `app/unit-tests/src/org/commcare/activities/HomeSyncButtonTest.kt` — sync-button network branching
- `app/unit-tests/src/org/commcare/activities/HomeLoginLaunchChecksTest.kt` — login-launch normal (no-op) path
- `app/unit-tests/src/org/commcare/activities/HomeActivityResultTest.kt` — onActivityResult pin-flow routing

app/src/org/commcare/activities/StandardHomeActivityUIController.java
```diff
@@ -162,7 +163,8 @@ private void setOpenConversationButtonVisibility(ConnectJobRecord job) {
         }
     }
 
-    private void updateConnectJobMessage() {
+    @VisibleForTesting
+    void updateConnectJobMessage() {
```
shubham1g5: we should not make any internal activity methods as public here but instead write the tests via user-UI interaction. For eg. we can simulate the same user action that leads to the call to  `updateConnectJobMessage` or interact with UI components to get their visibilty.

app/unit-tests/src/org/commcare/activities/HomeScreenActivityTest.kt
```diff
@@ -0,0 +1,262 @@
...
+        (CommCareTestApplication.instance() as CommCareTestApplication).initWorkManager()
+        TestAppInstaller.installAppAndLogin(TEST_APP_PATH, TEST_USER, TEST_PASSWORD)
+
+        // Keep Connect DB lookups deterministic and offline.
+        mockkStatic(
+            ConnectDatabaseHelper::class,
+            ConnectJobUtils::class,
+            ConnectMessagingDatabaseHelper::class,
+            ConnectAppDatabaseUtil::class,
+        )
```
shubham1g5: We don't really need to mock this to keep DB lookups offline/determininstic but instead just manipulate the DB directly by storing/deleting data. The pro here is that the test coverage would be more wholesome and cover even the DB functionality.

app/unit-tests/src/org/commcare/activities/HomeScreenActivityTest.kt
```diff
@@ -0,0 +1,262 @@
...
+        every { ConnectDatabaseHelper.isDbBroken() } returns false
```
shubham1g5: think this hides failures in test cases where a functionality breaks the DB somehow and we should avoid mocking it. 

Can we revise the list of things being mocked below as well and only mock what is absolutely necessary to mock in these tests. We would ideally like to test the whole functionality on UI level tests by mocking a least possible set of classes and mocks should be used to simulate external behaviours like network failing.

app/unit-tests/src/org/commcare/activities/HomeScreenActivityTest.kt
```diff
@@ -0,0 +1,262 @@
...
+        var cls: Class<*>? = start
+        while (cls != null) {
+            try {
+                return cls.getDeclaredField(name).apply { isAccessible = true }
+            } catch (e: NoSuchFieldException) {
+                cls = cls.superclass
+            }
+        }
+        throw NoSuchFieldException("$name not found on ${start.name} or its superclasses")
+    }
```
shubham1g5: Think these should go to a `ReflectionUtils` class so that we can reuse these but in general these methods should be avoided for anything that is testable by interacting with UI components.

app/unit-tests/src/org/commcare/android/tests/personalid/PersonalIdTestSandbox.kt
```diff
@@ -0,0 +1,24 @@
...
+ * config, which earns the test a separate classloader and keeps its churn contained.
+ *
+ * This is deliberately a *separate* marker from `org.commcare.activities.HomeTestSandbox` rather
+ * than a shared one: both suites do heavy Connect instrumentation, and giving each its own sandbox
+ * keeps them from polluting each other as well as the default sandbox.
+ */
+class PersonalIdTestSandbox
+
+@Implements(PersonalIdTestSandbox::class)
+class ShadowPersonalIdTestSandbox
```
shubham1g5: would the tests break without these empty shadows ? I am unclear why they are needed - is it just performance vs test stability or something else ?

app/unit-tests/src/org/commcare/activities/HomeActionsTest.kt
```diff
@@ -0,0 +1,269 @@
...
+        // showCommCareUpdateMenu starts false and is only flipped by the in-app update check.
+        val (_, menu) = homeWithMenu()
+
+        assertFalse(menu.findItem(R.id.action_update_commcare).isVisible)
+    }
+
+    @Test
+    fun `commcare update item shown once an update is available`() {
+        val home = buildHome()
+        home.showCommCareUpdateMenu = true
```
shubham1g5: Instead we should be mocking the `FlexibleAppUpdateController appUpdateController` in tests otherwise we are loosing test coverage on how `showCommCareUpdateMenu` gets set .

shubham1g5 (changes requested):
Have not been look into complete PR but did a general review based on the test harness and left some comments around the test approach we should follow for ui level tests, some of them might not be applicable based on implementation level details but think it's a good list of comments to review. 

Also flagging that the PR is really heavy on AI contextual comments.

app/src/org/commcare/activities/StandardHomeActivityUIController.java
OrangeAndGreen: [AI] Done: https://github.com/dimagi/commcare-android/pull/3830/commits/f139ec890

- Both methods are private again; the PR now changes no production code.
- Tests read the rendered button grid and the message card instead.
- Same pattern reverted in `DriftHelper` (two preference-key constants): https://github.com/dimagi/commcare-android/pull/3830/commits/885d8d318

app/unit-tests/src/org/commcare/activities/HomeScreenActivityTest.kt
OrangeAndGreen: [AI] Done: https://github.com/dimagi/commcare-android/pull/3830/commits/809606067

- Connect statics are no longer mocked; the fixture writes real Connect DB rows and home reads them back through the production lookups.
- Caught two things the mocks hid: a `STATUS_LEARNING` job NPEs when its learn app record is missing, and the tile rows were asserting a stubbed `getCardMessageText()` instead of real job state.
- Gotcha for future Connect tests: `getUser()` returns null unless `dbExists()` finds a DB file, so the fixture creates one alongside the in-memory DB.

app/unit-tests/src/org/commcare/activities/HomeScreenActivityTest.kt
OrangeAndGreen: [AI] Done: https://github.com/dimagi/commcare-android/pull/3830/commits/b37177a6b

- `ConnectNavHelper` stubs removed; the tests assert the activity each destination launches and the extras carried to the job-status screen.
- Four mocks left, all external: `ConnectivityStatus` (radios), `PersonalIdUnlocker` (biometrics), `ConnectJobHelper.updateDeliveryProgress` (spied, everything else runs for real), `AppUpdateControllerFactory` (Play Store).

app/unit-tests/src/org/commcare/activities/HomeScreenActivityTest.kt
OrangeAndGreen: [AI] Partly done: https://github.com/dimagi/commcare-android/pull/3830/commits/570575746

- Moved to `org.commcare.android.util.ReflectionUtils`.
- `HomeLoginLaunchChecksTest` no longer needs reflection — it calls `showPendingAlertDialog()` and asserts the real dialog fragment.
- Pushing back on `HomeInstanceStateTest`: the four instance-state flags it pins are private, have no accessor and reach no view or intent, so there is no UI to interact with. Reflection retires when CCCT-2679/2683 move them onto `SavedStateProvider`. Happy to drop the suite instead if you prefer.

app/unit-tests/src/org/commcare/android/tests/personalid/PersonalIdTestSandbox.kt
OrangeAndGreen: [AI] Done: https://github.com/dimagi/commcare-android/pull/3830/commits/91da1dc91

- Both markers deleted. Not performance: they forced those suites onto their own Robolectric classloader, because MockK inline instrumentation from static-mocking the Connect classes corrupted Play-Services state for later tests.
- Removing the Connect static mocks removed the cause; full app unit-test suite green without them.

app/unit-tests/src/org/commcare/activities/HomeActionsTest.kt
OrangeAndGreen: [AI] Done: https://github.com/dimagi/commcare-android/pull/3830/commits/f145b0b89

- Agreed: setting the field skipped `handleAppUpdate()` entirely.
- The test stubs `AppUpdateControllerFactory` to return a fake `FlexibleAppUpdateController` and runs the callback home registers.
- Both arms pinned: over `MAX_CC_UPDATE_CANCELLATION` offers the menu item, under it calls `startUpdate()`.

OrangeAndGreen:
[AI] Round 1 addressed, one commit per thread. Test commentary trimmed too: https://github.com/dimagi/commcare-android/pull/3830/commits/126b8705c

- No production changes left: both `StandardHomeActivityUIController` widenings and the two `DriftHelper` constants reverted.
- No DB mocking: Connect state is written to the real Connect DB and read back through production lookups.
- Four mocks total, all external — network status, biometric unlock, job-progress fetch, Play Store updates (down from 15+).
- Both sandbox marker classes deleted.
- One pushback, in the `ReflectionUtils` thread: `HomeInstanceStateTest` keeps field reflection, as the flags it pins have no accessor and no UI surface.
- Full app unit-test suite green: 139 test classes, 0 failures.

app/unit-tests/src/org/commcare/activities/HomeScreenActivityTest.kt
```diff
@@ -0,0 +1,202 @@
...
+ * other dependency, Connect DB included, runs for real. An empty Connect DB is what makes a direct
+ * subclass boot with no job seated. Tests that assert on Connect behaviour extend
+ * [HomeConnectTestBase], which seats that state in the DB.
+ *
+ * Subclasses run under [AndroidJUnit4] with [CommCareTestApplication]; add `@Config(sdk = [...])`
+ * on individual tests only when a specific Android level matters.
+ */
+@Config(application = CommCareTestApplication::class)
+@RunWith(AndroidJUnit4::class)
+abstract class HomeScreenActivityTest {
```
shubham1g5: nit: rename to `BaseHomeScreenActivityTest` to align with it being a base class for home tests.

app/unit-tests/src/org/commcare/activities/HomeScreenActivityTest.kt
```diff
@@ -0,0 +1,202 @@
...
+        injectFakeSyncer(controller.get())
+        return controller
+    }
+
+    /** Swap in a syncer that records instead of hitting the network. */
+    protected fun injectFakeSyncer(home: StandardHomeActivity) = home.setFormAndDataSyncer(FormAndDataSyncerFake())
+
+    protected fun shadowOf(activity: StandardHomeActivity): ShadowActivity = Shadows.shadowOf(activity)
+
+    protected fun uiController(home: StandardHomeActivity): StandardHomeActivityUIController =
```
shubham1g5: unused ?

app/unit-tests/src/org/commcare/activities/HomeScreenActivityTest.kt
```diff
@@ -0,0 +1,202 @@
...
+    /**
+     * The labels of the buttons the home grid renders, read off the bound card views. Position 0 is
+     * the banner header rather than a button, so it is skipped.
+     */
+    protected fun homeButtonLabels(home: StandardHomeActivity): List<String> {
+        val grid = home.findViewById<RecyclerView>(R.id.home_gridview_buttons)
+        val adapter = grid.adapter!!
+        return (1 until adapter.itemCount).map { position ->
+            val holder = adapter.createViewHolder(grid, adapter.getItemViewType(position))
+            adapter.bindViewHolder(holder, position)
```
shubham1g5: Is the objective of this method is to get labels for all buttons ? If so why do we need to call `createViewHolder` and `bindViewHolder` as that'd be called already in the Home Activity production code ?

app/unit-tests/src/org/commcare/activities/HomeActionsTest.kt
```diff
@@ -0,0 +1,293 @@
...
+        every { ConnectivityStatus.isAirplaneModeOn(any()) } returns true
+        val home = buildHome()
+
+        home.syncButtonPressed()
+
+        assertEquals(
+            Localization.get("notification.sync.airplane.action"),
+            ShadowToast.getTextOfLatestToast(),
+        )
+        assertTrue("a Sync_AirplaneMode notification should be pending", notificationsArePending())
```
shubham1g5: Should we also add exact matching against notification title/message here ?

app/unit-tests/src/org/commcare/activities/HomeActionsTest.kt
```diff
@@ -0,0 +1,293 @@
...
+    fun `sub text press starts no activity when no messages pending`() {
+        val home = buildHome()
+
+        home.syncSubTextPressed()
+
+        assertNull(shadowOf(home).nextStartedActivity)
+    }
+
+    @Test
+    fun `sub text press launches the notifications view when messages are pending`() {
```
shubham1g5: Not related to the PR but don't think we have considered this workflow in Connect Redesign and think we should surface it to the Product to see if they want to retain this behaviour.

app/unit-tests/src/org/commcare/activities/HomeActionsTest.kt
```diff
@@ -0,0 +1,293 @@
...
+ * (demo and non-demo), item-selection routing, and the sync button's offline handling and sub-text.
+ *
+ * Driven through [StandardHomeActivity] only; [RootMenuHomeActivity] overrides
+ * `onOptionsItemSelected` and needs its own rows. Routing is pinned for the two menu items that
+ * leave an assertable intent — the rest open dialogs or preference fragments.
+ *
+ * The Connect job-progress fetch that `syncButtonPressed()` also triggers is pinned in
+ * [HomeConnectJobProgressTest].
+ */
+class HomeActionsTest : HomeScreenActivityTest() {
```
shubham1g5: We seem to be missing other actions here corresponding to the buttons/menus on the Home Screen, for example - we should check for behaviour on clicking all option menu items plus the home buttons other than sync present on this screen.

app/unit-tests/src/org/commcare/activities/HomeConnectTestBase.kt
```diff
@@ -0,0 +1,132 @@
...
+import java.util.concurrent.TimeUnit
+
+/**
+ * Shared fixtures for the home screen's PersonalId/Connect tests (`HomeConnect*`).
+ *
+ * Seating a job writes the same rows into the real Connect DB that a Connect sync would leave
+ * behind — a signed-in PersonalId user, the job, and the app record linking it to the seated
+ * CommCare app — so the home screen resolves it through the production lookups rather than a stub.
+ */
+abstract class HomeConnectTestBase : HomeScreenActivityTest() {
```
shubham1g5: would be nice to abstract it into a utility class instead like `ConnectTestUtils` instead given all these methods would be helpful outside home screen context (Similar to `TestAppInstaller`)

app/unit-tests/src/org/commcare/activities/HomeConnectDrawerNavigationTest.kt
```diff
@@ -0,0 +1,111 @@
...
+    fun `work history legacy login unlocks before work history`() {
+        val home = buildHome(personalIdManagedLogin = false)
+
+        home.handleDrawerItemClick(NavItemType.WORK_HISTORY, null)
+
+        verify(exactly = 1) { PersonalIdUnlocker.unlock(home, any(), any()) }
+        assertStarted(home, PersonalIdWorkHistoryActivity::class.java)
+    }
+
+    private fun assertStarted(
```
shubham1g5: nit: seems like a good candidate to be abstracted out in a `ActivityAssertions.kt` or any other util class

app/unit-tests/src/org/commcare/activities/HomeConnectJobProgressTest.kt
```diff
@@ -0,0 +1,106 @@
...
+ * Characterization pins for the home screen's Connect job-progress paths: opening a job's status
+ * screen, and when `fetchJobProgressOverNetwork()` actually hits the network — both when called
+ * directly and when reached through the sync button.
+ *
+ * The sync button's traditional behaviour — the toast and notification raised when a sync can't be
+ * attempted — lives in [HomeActionsTest].
+ */
+class HomeConnectJobProgressTest : HomeConnectTestBase() {
+    @Before
+    fun stubProgressFetch() {
```
shubham1g5: we usually stub network calls using `MockWebServer`, we should stick to the same unless there is a good reason for us to mock the internal app code.

app/unit-tests/src/org/commcare/activities/HomeConnectJobProgressTest.kt
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    // ---- Opportunity status ----
+
+    @Test
+    fun `view opportunity status navigates to job info when a job is seated`() {
+        seatJob(connectJob())
+        val home = buildHome()
+
+        home.userPressedOpportunityStatus()
```
shubham1g5: instead we should directly do a click with Espresso APIs here - `onView(withText()).perform(click())`

app/unit-tests/src/org/commcare/activities/HomeConnectJobProgressTest.kt
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    // ---- fetchJobProgressOverNetwork ----
+
+    @Test
+    fun `fetch job progress over network updates progress for a delivering job`() {
+        seatJob(connectJob(status = ConnectJobRecord.STATUS_DELIVERING))
+        val home = buildHome()
+
+        home.fetchJobProgressOverNetwork()
```
shubham1g5: We should click the relevant action button for this instead unless there is an explicit reason to interact with a code component here.

app/unit-tests/src/org/commcare/activities/HomeConnectJobProgressTest.kt
```diff
@@ -0,0 +1,106 @@
...
+
+    // ---- Reached via the sync button ----
+
+    @Test
+    fun `sync with no network in airplane mode does not fetch job progress`() {
+        every { ConnectivityStatus.isNetworkAvailable(any()) } returns false
+        every { ConnectivityStatus.isAirplaneModeOn(any()) } returns true
+        val home = buildHome()
+
+        home.syncButtonPressed()
```
shubham1g5: same comment as above on using Espresso Apis to interact with views.

app/unit-tests/src/org/commcare/activities/HomeActivityResultTest.kt
```diff
@@ -0,0 +1,58 @@
...
+ * Characterization pins for the PIN-related arms of `onActivityResultSessionSafe`:
+ * `AUTHENTICATION_FOR_PIN` routing into PIN creation, and the two `CREATE_PIN` outcomes the user
+ * sees as a toast.
+ *
+ * Scope: `onActivityResultSessionSafe` dispatches twelve request codes plus an early `RESULT_RESTART`
+ * branch, and only the PIN codes are pinned here. The rest — `PREFERENCES_ACTIVITY`,
+ * `ADVANCED_ACTIONS_ACTIVITY`, `GET_INCOMPLETE_FORM`, `GET_COMMAND`, `GET_CASE`, `MODEL_RESULT`,
+ * `MAKE_REMOTE_POST`, `GET_REMOTE_DATA`, `IN_APP_UPDATE_REQUEST_CODE` — all drive session navigation
+ * through `startNextSessionStepSafe()` and need a seated multi-step session to be meaningful; they
+ * belong with the session-navigation slices rather than here. The third `CREATE_PIN` arm
```
shubham1g5: seems like the session navigation related tests are missing currently from the PR and also seems to be the most important ones to check for in context of Home activity. Are they covered some-where else in our test suite ?

app/unit-tests/src/org/commcare/activities/HomeLoginLaunchChecksTest.kt
```diff
@@ -0,0 +1,447 @@
...
+    /**
+     * The boolean `doLoginLaunchChecksInOrder` returned, as stored on the activity. Private with no
+     * accessor; the walk up from [StandardHomeActivity] finds this field rather than the same-named
+     * one further up on `SessionAwareCommCareActivity`.
+     */
+    private fun claimedLaunch(home: StandardHomeActivity): Boolean = ReflectionUtils.readField(home, "redirectedInOnCreate") as Boolean
+
+    /**
+     * The dialog a launch check raised during `onCreate`, or null if none was raised. Asking the
+     * activity to show what it stashed is what `onResume` does, so the dialog becomes a real
```
shubham1g5: can we call `activity.onResume()` instead directly if we want to test for a change caused by that method rather than replicating `onResume` code here ?

shubham1g5 (changes requested):
As some general feedback over PR, we should not be invoking the activity level methods written by us directly from the test classes but should interact with the view itself. Espresso provides APIs to do this and we can also directly use `activity.findViewById()` as convenient.

app/unit-tests/src/org/commcare/activities/HomeScreenActivityTest.kt
OrangeAndGreen: [AI] Done: https://github.com/dimagi/commcare-android/pull/3830/commits/f959fcfa5

app/unit-tests/src/org/commcare/activities/HomeScreenActivityTest.kt
OrangeAndGreen: [AI] Done: unused, deleted — https://github.com/dimagi/commcare-android/pull/3830/commits/f959fcfa5

app/unit-tests/src/org/commcare/activities/HomeScreenActivityTest.kt
OrangeAndGreen: [AI] Done: https://github.com/dimagi/commcare-android/pull/3830/commits/f959fcfa5

- Right, it should not have bound anything. `buildHome()` stopped at `onCreate`, so the `RecyclerView` had no children to read.
- The base now offers `buildVisibleHome()` (`start().resume().visible()`); Robolectric lays the grid out, so `homeButtonLabels()` walks the rendered cards.
- Side effects: the `OnGlobalLayoutListener` in `setupGridView()` now fires, and the click-based rows became possible.

app/unit-tests/src/org/commcare/activities/HomeActionsTest.kt
OrangeAndGreen: [AI] Done: https://github.com/dimagi/commcare-android/pull/3830/commits/f959fcfa5

- Rows assert the notification's exact title, detail and category, and that it is the only one reported.
- Category is worth pinning: `syncButtonPressed()` clears notifications by that category when the network is back.
- The rows moved to `HomeButtonsTest` and are driven by clicking the rendered sync card.

app/unit-tests/src/org/commcare/activities/HomeActionsTest.kt
OrangeAndGreen: [AI] Done: https://github.com/dimagi/commcare-android/pull/3830/commits/f959fcfa5

- Split into `HomeMenuTest` (a routing row for every item `onOptionsItemSelected` handles) and `HomeButtonsTest` (every rendered card, clicked).
- Menu rows: update, saved forms, settings vs advanced, set pin, change language, about, update commcare.
- Button rows: start, saved and incomplete forms, log out, sync, and the sync sub-text strip.
- Training and report are not rendered by the test app's profile, so they stay in `StandardHomePresentationTest`; Connect is clicked in `HomeConnectJobProgressTest`.

app/unit-tests/src/org/commcare/activities/HomeConnectTestBase.kt
OrangeAndGreen: [AI] Done: https://github.com/dimagi/commcare-android/pull/3830/commits/7a07d9679

- Now `org.commcare.android.util.ConnectTestUtils`, alongside `TestAppInstaller`; `HomeConnectTestBase` deleted.
- `seatedAppId()` was duplicated, so it moved to `TestAppInstaller`.
- `signInToPersonalId()` now stores a valid Connect token, which keeps the mock-web-server tests off the PersonalId token endpoint.

app/unit-tests/src/org/commcare/activities/HomeConnectDrawerNavigationTest.kt
OrangeAndGreen: [AI] Done: https://github.com/dimagi/commcare-android/pull/3830/commits/e33f689f5

- Now `org.commcare.android.util.ActivityAssertions`: `assertStarted`, `assertStartedForResult`, `assertStartedNothing`, `assertOnly`, plus two queue-draining readers.
- Worth having in one place: a `startActivityForResult` launch lands in both Robolectric queues, so a follow-up "nothing else was started" check silently passes — two of the new session-navigation rows were green for that reason.

app/unit-tests/src/org/commcare/activities/HomeConnectJobProgressTest.kt
OrangeAndGreen: [AI] Done: https://github.com/dimagi/commcare-android/pull/3830/commits/f5b1eaab6

- New `ConnectMockApiServer` points `ConnectNetworkClient` at a `MockWebServer`; the `ConnectJobHelper` stub is gone.
- Request, parser and the DB write in `applyToJob` all run for real, so rows assert the endpoint hit and the job state left behind — plus a 500 case.
- Two Robolectric notes are documented in the harness: waiting means take-request then idle the main looper, and the seated user needs a valid Connect token.

app/unit-tests/src/org/commcare/activities/HomeConnectJobProgressTest.kt
OrangeAndGreen: [AI] Discussion — these are clicks now, but on the rendered view rather than through `onView`: https://github.com/dimagi/commcare-android/pull/3830/commits/f5b1eaab6

- `onView(...).perform(click())` finds the view but refuses to click it: under Robolectric `getGlobalVisibleRect()` reports no rect, so click's 90%-visible constraint can never pass.
- No `onView` usage exists anywhere in `app/unit-tests` today; existing suites click via `performClick()`.
- So the base offers `clickHomeButton(home, label)`, clicking the same `R.id.card` the adapter bound the production listener to.
- Happy to revisit if `onView` works in a Robolectric suite I have missed.

app/unit-tests/src/org/commcare/activities/HomeConnectJobProgressTest.kt
OrangeAndGreen: [AI] Done: https://github.com/dimagi/commcare-android/pull/3830/commits/f5b1eaab6

- Now `clickHomeButton(home, Localization.get("home.connect"))` on the rendered card; no direct `userPressedOpportunityStatus()` call. Mechanism in the Espresso thread above.
- One row dropped as a result: the NPE pin for "no job seated" — `shouldShowJobStatus()` hides the Connect button, so the throw is unreachable through the UI. `HomeConnectTileTest` pins the absence.

app/unit-tests/src/org/commcare/activities/HomeConnectJobProgressTest.kt
OrangeAndGreen: [AI] Done: https://github.com/dimagi/commcare-android/pull/3830/commits/f5b1eaab6

- Sync rows click the rendered card via `clickHomeButton`; the sub-text strip via `clickHomeButtonSubText` (`R.id.card_subtext`, a separate listener).
- Why `performClick()` rather than `onView`: see the thread above.

app/unit-tests/src/org/commcare/activities/HomeActivityResultTest.kt
OrangeAndGreen: [AI] Partly covered elsewhere; the gap is now filled: https://github.com/dimagi/commcare-android/pull/3830/commits/a9d0a0c36

- `EndOfFormTest`, `FormRecordProcessingTest` and `XFormUpdateInfoTest` already feed form results back through home's `onActivityResult`.
- `ActivityLaunchUtils.launchEntitySelectActivity` covers the entity-select launch.
- New `HomeSessionNavigationTest` pins home's own end: the `GET_COMMAND` callout and advance into form entry, `RESULT_CANCELED` resetting the session, and the stale-callout guard (previously uncovered).
- Open question: `GET_CASE` needs a case-list app seated, as `form_nav_tests` has no case datum. Want it in this PR?

app/unit-tests/src/org/commcare/activities/HomeLoginLaunchChecksTest.kt
OrangeAndGreen: [AI] Done: https://github.com/dimagi/commcare-android/pull/3830/commits/f6dd6e879

- `controller.resume()` now, so `onResumeFragments()` raises the dialog exactly as production does.
- One catch: `onResumeSessionSafe()` ends in `resetNavigationFlags()`, so reading `redirectedInOnCreate` after a resume would pass for free.
- Each test therefore uses a `LaunchCheckRun` that snapshots create-time state on construction and exposes `alertDialogOnResume()` to be called last.

app/unit-tests/src/org/commcare/activities/HomeActionsTest.kt
OrangeAndGreen: [AI] Discussion — no code change; the pin just moved to `HomeButtonsTest`.

- Current behaviour: the sync card's sub-text is a second tap target that opens `MessageActivity` when CommCare notifications are pending, silent otherwise.
- Want me to raise it with Product, or would you rather take it?

OrangeAndGreen:
[AI] Round 2 addressed, one commit per thread.

- Everything goes through the rendered UI: home is taken to a laid-out window, so the grid binds its own cards and buttons are clicked on the production listener. `onView` does not work under Robolectric — details in that thread.
- Connect job progress is no longer mocked: `ConnectMockApiServer` points `ConnectNetworkClient` at a `MockWebServer`, so request, parser and DB write run for real.
- Action coverage complete for this screen: `HomeMenuTest` (every item `onOptionsItemSelected` handles) and `HomeButtonsTest` (every rendered button).
- `HomeSessionNavigationTest` adds home's own callout, stale-result guard and advance into form entry.
- Extracted `ConnectTestUtils`, `ActivityAssertions`, `ConnectMockApiServer`; base renamed `BaseHomeScreenActivityTest`, `HomeConnectTestBase` deleted.
- Two rows changed shape instead of moving as asked: the no-job NPE pin is dropped (unreachable through the UI), and the launch-check test resumes but snapshots create-time state first.
- Open question: the `GET_CASE` leg needs a second app seated — want it in this PR?
- Suite green: 141 test classes, 954 tests, 0 failures. PR description updated.

shubham1g5:
@OrangeAndGreen Flagging that the AI text on the PR comments here is brutal to read and review and would be great to avoid for ease of review unless necessary.

OrangeAndGreen:
> @OrangeAndGreen Flagging that the AI text on the PR comments here is brutal to read and review and would be great to avoid for ease of review unless necessary.

Ah, okay, I decided to give it a try and see how well Claude did but I agree that the write-ups are long, I'll see if I can get Claude to be more concise and otherwise revert to writing responses by hand. Apologies!

OrangeAndGreen:
@shubham1g5 Each response now starts with a quick indication of whether the recommendation was taken (along with a link to the commit if there is one), and then some simple bullets describing what was done.

For any response that says "Done" you can skip the additional write-up if you don't feel the need for extra info.

app/unit-tests/src/org/commcare/activities/HomeActionsTest.kt
shubham1g5: @OrangeAndGreen I can't tell whether it's claude talking to reviewer on your behalf or Claude talking to you as author , but confirming it'd be great if you can raise it to the Product.

shubham1g5 (approved):

PR #3834, opened 2026-07-28
CCCT-2679 Add HomeActivityHost And HomeActivityCoordinator

OrangeAndGreen:
## Technical Summary

First slice of the home-activity composition refactor (home-composition design spec, CCCT-2522), which moves session-dependent home behavior out of the `SyncCapableCommCareActivity` → `HomeScreenBaseActivity` inheritance chain and into composable units, so `OpportunityHomeActivity` can later gain and lose a session without leaving the screen. This slice stands up the substrate only — no delegates, no action facade, no launch pipeline.

**What lands**

- **`HomeActivityHost`** (new, `org.commcare.home`) — the capability surface a home activity offers the coordinator: `SavedStateRegistryOwner`, plus `hostContext`, `startActivityForResult`, `showAlertDialog`, `rebuildOptionsMenu`, `refreshHostUi`, `isDemoUser()`, `areAppActionsAvailable()` and `refreshActionSurface()`.
- **`HomeActivityCoordinator`** (new) — registers itself on the host lifecycle during `init`, owns the three launch/nav flags (`wasExternal`, `loginExtraWasConsumed`, `pendingEndpointNavigationAfterSync`), and persists them through the host's `SavedStateRegistry`. It also carries an `onActivityResult(...)` fan-out seam, empty until delegates arrive, so later slices only add targets rather than re-wiring the host.
- **`HomeScreenBaseActivity`** implements `HomeActivityHost`, constructs the coordinator in a field initializer, forwards `onActivityResult` into it, and drops its own `loadInstanceState`/`onSaveInstanceState` handling of those three keys. Seven read/write sites reroute to the coordinator; the persisted key names are unchanged.

`isDemoUser()` changes from `protected static` to `public` instance — a static cannot implement an interface method in Java. One production call site updated (`StandardHomeActivityUIController`), plus one test assertion.

NOTE:
The branch also merges `master` and repairs the #3830 characterization tests against the drawer app-list removal that landed there: `HomeDrawerNavigationTest` now pins the switch-app confirmation dialog instead of the old record-id click contract, and `StandardHomePresentationTest` drops `seated app is highlighted`.

### Safety story

**What gives me confidence:**

- The characterization suite from #3830 covers the changed paths and is green. `./gradlew testCommcareDebugUnitTest --tests "org.commcare.activities.*"` runs **109 tests, 0 failures** locally; the home-screen subset of that package goes 77 → 87 tests across this branch.
- Dev tested basic App Home functionality, including logging into an app, syncing, and delivering forms.
- Dev tested launching the app from an external command to the app home screen, rotating the screen back and forth between portrait and landscape during both login and app home, then completing a form and verifying that the app exits completely when finished (intended behavior when launched externally).
  - Ran the following command from a terminal while phone was connected via USB
  - `adb shell am start -a org.commcare.dalvik.action.CommCareSession -n org.commcare.dalvik.debug/org.commcare.activities.DispatchActivity --es ccodk_session_endpoint_id "cc_app_home"`
- The only behavior change intended here is *where* three booleans are persisted. Seven call sites, mechanically rerouted; the persisted key names are unchanged from the old `HomeScreenBaseActivity` constants.
- The one red test on CI is `IntentCalloutTests > testIntentCalloutNoData` — a form-entry test that touches nothing in this diff and has been flaky before. No home-screen test fails on CI.
- Each commit compiles and leaves the suite green, so the slice bisects cleanly.

**Risks to review:**

- **`rebuildOptionsMenu()` name collision — fixed, was the one real bug.** The host member originally declared `rebuildOptionsMenu()`, colliding with `CommCareActivity.rebuildOptionsMenu()`. Implementing it on `HomeScreenBaseActivity` overrode `SyncCapableCommCareActivity`'s version without chaining to `super`, silently diverting all three pre-existing callers (`triggerSyncIconRefresh`, the change-language dialog, `StandardHomeActivityUIController`'s layout listener). The member is now `refreshActionSurface()`, the override is gone, and those callers take their `master` path again.
- The coordinator is constructed in a field initializer and passes `this` out during construction to register a lifecycle observer. This is safe because `ComponentActivity` initializes its `LifecycleRegistry` in a superclass field initializer, and the class already does the same for `FirebaseMessagingDataSyncer` — but it is worth a reviewer's eye.
- Six `HomeActivityHost` members have no consumer yet (`hostContext`, `startActivityForResult`, `showAlertDialog`, `refreshHostUi`, `areAppActionsAvailable`, `refreshActionSurface`); only `isDemoUser()` is called today. Declared up front per the ticket; wired up in later slices.

### Automated test coverage

- `HomeActivityCoordinatorTest` (7 tests) — activity-free coverage via a `FakeHomeActivityHost` that drives `performRestore`, `ON_CREATE` and `performSave` independently: self-registration on the host lifecycle, all-false defaults, the full save/restore round trip, per-key independence, and the no-op activity-result fan-out. Two tests pin the lifecycle-ordering constraint described above — a write made before `ON_CREATE` is not clobbered by the restore pass, and state set before `ON_CREATE` is still saved. Either would fail under the spec's original mechanism.
- `HomeInstanceStateRecreationTest` (7 tests) — activity-level `recreate()` round trip on `StandardHomeActivity`, covering each key alone, all three together, the all-false baseline, that an external launch actually writes to the coordinator, and that post-login launch checks do not re-run after recreation.
- These replace `HomeInstanceStateTest` (2 tests), the reflection-based characterization pin from #3830. Now that the flags live behind coordinator accessors, the `ReflectionUtils` scaffolding that test needed retires with it.

[CCCT-2679]: https://dimagi.atlassian.net/browse/CCCT-2679?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


OrangeAndGreen:
## Suggested Review Order

The commits tell the story cleanly and each one compiles with the suite green, so reviewing commit-by-commit is the easiest path. If reading by file instead:

- `app/src/org/commcare/activities/home/HomeActivityHost.kt` — the new contract; everything else consumes it
- `app/src/org/commcare/activities/home/HomeActivityCoordinator.kt` — the substrate; read `ensureRestored()` closely, it is where the lifecycle-ordering deviation lives
- `app/unit-tests/src/org/commcare/activities/home/FakeHomeActivityHost.kt` — mirrors `ComponentActivity`'s restore/dispatch/save order, which is what makes the ordering testable
- `app/unit-tests/src/org/commcare/activities/home/HomeActivityCoordinatorTest.kt` — the coordinator's contract, including the two ordering tests
- `app/src/org/commcare/activities/HomeScreenBaseActivity.java` — the only production file with real churn: host implementation, coordinator field, seven rerouted sites, three deletions
- `app/src/org/commcare/activities/StandardHomeActivityUIController.java` — one line, follows from `isDemoUser()` becoming an instance method
- `app/unit-tests/src/org/commcare/activities/HomeScreenActivityTest.kt` — harness change only; extracts `buildHomeController` so tests can drive `recreate()`
- `app/unit-tests/src/org/commcare/activities/HomeInstanceStateRecreationTest.kt` — the acceptance tests; read last, they assert the end state

app/unit-tests/src/org/commcare/activities/HomeInstanceStateRecreationTest.kt
```diff
@@ -0,0 +1,113 @@
...
+            }
+        assertTrue(controller.get().coordinator.loginExtraWasConsumed)
+
+        controller.recreate()
+        ShadowLooper.idleMainLooper()
+
+        val recreated = controller.get()
+        assertTrue(recreated.coordinator.loginExtraWasConsumed)
+        assertFalse(recreated.isFinishing)
+    }
```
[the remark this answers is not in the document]
OrangeAndGreen: [237a521](https://github.com/dimagi/commcare-android/pull/3834/commits/237a52161685f56c5ceb92f2a320797b889e2209)

app/src/org/commcare/activities/home/HomeActivityCoordinator.kt
```diff
@@ -0,0 +1,110 @@
+package org.commcare.activities.home
```
shubham1g5: thoughts on moving these classes to a new package ->  `org.commcare.home`  ?

app/src/org/commcare/home/HomeActivityCoordinator.kt
```diff
@@ -0,0 +1,110 @@
...
+    private var endpointNavPendingAfterSync = false
+
+    /** Activity was launched by an external app, so form submission may redispatch back to it. */
+    var wasExternal: Boolean
+        get() {
+            ensureRestored()
+            return externalLaunch
+        }
+        set(value) {
+            ensureRestored()
```
shubham1g5, line 32: why do we need to call this with a setter ?

app/src/org/commcare/activities/HomeScreenBaseActivity.java
```diff
@@ -225,25 +233,15 @@ private void updateLastSuccessfulCommCareVersion() {
...
-            pendingEndpointNavigationAfterSync = savedInstanceState.getBoolean(
-                    KEY_PENDING_ENDPOINT_NAV_AFTER_SYNC
-            );
-        }
-    }
-
     /**
      * Set state that signifies activity was launch from external app
      */
     private void processFromExternalLaunch(Bundle savedInstanceState) {
```
shubham1g5, line 239: Shoud we not move the whole method to co-ordinator given the end result of this work is not to have the `HomeScreenBaseActivity` at all ?

app/src/org/commcare/activities/home/HomeActivityHost.kt
```diff
@@ -0,0 +1,47 @@
...
+    )
+
+    /** Show [dialog] on the host, so dialogs survive the host's own dismissal bookkeeping. */
+    fun showAlertDialog(dialog: CommCareAlertDialog)
+
+    /**
+     * Re-render whatever surface the host uses to present coordinator actions. `StandardHomeActivity`
+     * invalidates its options menu; other activities may refresh the options in other ways
+     */
+    fun rebuildOptionsMenu()
```
shubham1g5: I don't see this being called anywhere right now, is that slotted for an upcoming PR ?

shubham1g5: Ahh I see this is already defined method in `CommCareActivity`, given that do we need it in the host ? Also think the current implementation introduces a regression/behaviour change given we are no longer calling this method in base implementations of `HomeScreenBaseActivity` which would have otherwise got called on `master`

app/src/org/commcare/activities/home/HomeActivityCoordinator.kt
OrangeAndGreen: [AI] Done: https://github.com/dimagi/commcare-android/pull/3834/commits/881423e6d

app/src/org/commcare/home/HomeActivityCoordinator.kt
OrangeAndGreen: [AI] Discussion — it is load-bearing, so I left it in place.

- Without it a write that lands before `ON_CREATE` gets clobbered: the first later *getter* would run `ensureRestored()` and overwrite the value with the restored one. Calling it first in the setter makes restore happen, then the write wins.
- This is a real path, not hypothetical: `onCreateSessionSafe` -> `processFromExternalLaunch` sets `wasExternal` and `pendingEndpointNavigationAfterSync` during `Activity.onCreate`, before `ON_CREATE` is dispatched.
- `ensureRestored()` is also what registers the saved-state provider, so a set with no intervening get would otherwise persist nothing.
- `HomeActivityCoordinatorTest` pins both halves — `write before ON_CREATE survives the restore pass` and `state set before ON_CREATE is still saved`.

app/src/org/commcare/activities/HomeScreenBaseActivity.java
OrangeAndGreen: [AI] Discussion — agreed on the destination, but I would like to keep it for the launch-pipeline slice.

- The method needs `sessionNavigator.startNextSessionStep()`, `triggerSync(false)`, `processSessionEndpoint()` -> `validateIntentForSessionEndpoint()`, and `redirectedInOnCreate`, which `onResumeSessionSafe` also reads. Moving it now means pulling the whole session-navigation surface onto `HomeActivityHost` in this PR.
- That surface is the next slice in the CCCT-2522 spec, so this would either duplicate that work or leave a half-migrated method behind.
- Happy to fold it in here instead if you would rather have one bigger PR than two.

app/src/org/commcare/activities/home/HomeActivityHost.kt
OrangeAndGreen: [AI] Yes — nothing consumes it yet; it is for the action facade in a later slice, same as `hostContext`, `startActivityForResult`, `showAlertDialog` and `areAppActionsAvailable`. Only `isDemoUser()` has a caller today. See the thread below for the collision you spotted.

OrangeAndGreen: First, fixed the regression: https://github.com/dimagi/commcare-android/pull/3834/commits/aa54c7b9d

As part of the fix I renamed the function to refreshActionSurface since it isn't only necessarily updating a standard menu (it could be updating actions located in other parts of the UI).

[AI]
- Confirmed, and you are right that the collision was the cause. The interface member is now `refreshActionSurface()`, so `HomeScreenBaseActivity` no longer overrides `SyncCapableCommCareActivity.rebuildOptionsMenu()` and all three pre-existing callers go back to their `master` path.
- What the regression was: on `StandardHomeActivity` it dropped `clearCurrentAnimation()` and the `getCurrentApp() != null` guard; on `RootMenuHomeActivity` the options menu stopped invalidating at all, since `refreshCCUpdateOption()` there only refreshes drawer items behind `usingNavDrawer()`.
- Callers left alone deliberately: `triggerSyncIconRefresh`, the change-language dialog and `StandardHomeActivityUIController`'s layout listener all still call `rebuildOptionsMenu()`.
- Kept on the interface rather than dropped, since the facade needs it next slice — and the new name is the more honest one, given the surface is the drawer on `RootMenuHomeActivity`.

Re: it not being called yet:
Correct, nothing consumes it yet; it is for the action facade in upcoming work, same as hostContext, startActivityForResult, showAlertDialog and areAppActionsAvailable. Only isDemoUser() has a caller today.

conroy-ricketts (dismissed):
LGTM, but would love another set of eyes before merge

app/src/org/commcare/activities/home/HomeActivityHost.kt
shubham1g5: Think the refactor introduces another regression now given `refreshCCUpdateOption` would never be called now while it was being called earlier by activity call chain for `rebuildOptionsMenu`. 

Also is there a conceptual difference between `refreshActionSurface` and `rebuildOptionsMenu` ? Can we not use the `rebuildOptionsMenu` from activity callchain or putting in other words why do we want this to be in `HomeActivityHost.kt` at all ?

app/src/org/commcare/home/HomeActivityCoordinator.kt
```diff
@@ -0,0 +1,110 @@
...
+            loginExtraConsumed = state.getBoolean(KEY_LOGIN_EXTRA_CONSUMED)
+            endpointNavPendingAfterSync = state.getBoolean(KEY_PENDING_ENDPOINT_NAV_AFTER_SYNC)
+        }
+        registry.registerSavedStateProvider(PROVIDER_KEY) {
+            Bundle().apply {
+                putBoolean(KEY_WAS_EXTERNAL, externalLaunch)
+                putBoolean(KEY_LOGIN_EXTRA_CONSUMED, loginExtraConsumed)
+                putBoolean(KEY_PENDING_ENDPOINT_NAV_AFTER_SYNC, endpointNavPendingAfterSync)
+            }
+        }
```
Jignesh-dimagi, line 101: Just curious about the decision to use SavedStateRegistry over a standard Bundle here. HomeScreenBaseActivity still contains quite a bit of logic that relies directly on bundle data. Could we potentially move that entire state-handling logic from the Activity into this Coordinator instead?

app/src/org/commcare/home/HomeActivityCoordinator.kt
OrangeAndGreen: Human rewrite:
All access to the coordinator needs to be protected to ensure the one-time restore operation has run first, and relying on onCreate doesn't work.
* Interactions can happen before onCreate and could otherwise later be clobbered when the restore happens
* This problem actually occurs in practice: onCreateSessionSafe sets wasExternal before the onCreate gets called
* Also, ensureRestored sets up the state persistence (via SavedStateRegistry), which guarantees the set variable survives state changes.
* Unit tests cover both of the above behaviors

app/src/org/commcare/home/HomeActivityCoordinator.kt
OrangeAndGreen: (Note: I wrote this by hand then had Claude tweak it slightly and add formatting)

Yes — all of this moves out of the Activity, though not all in this ticket: some pieces land in delegates that come with upcoming tickets.

On the SavedStateRegistry specifically:
  * **No known host.** The Coordinator is hosted by `StandardHomeActivity` today and by the Opp Home host later (`ConnectActivity` or a new `OpportunityHomeActivity`). With a forwarded Bundle, every host needs matching save/load plumbing, and every new delegate means another edit to the host to thread its state through. With the registry each delegate owns its own key and the host stays out of it entirely — same pattern for the upcoming delegates, and it keeps the Coordinator testable without a concrete Activity.

  * **Timing.** The registry becomes readable before `ON_CREATE` reaches a lifecycle observer, and the home screen reads these flags in that gap:

    ```
    performRestore()          ← registry readable (first line of ComponentActivity.onCreate)
    super.onCreate() returns
    onCreateSessionSafe()     ← coordinator flags read here (processFromLoginLaunch / processFromExternalLaunch)
    onCreate() returns
    ON_CREATE dispatched      ← DefaultLifecycleObserver.onCreate(owner) — too late
    ```

    That's why `ensureRestored()` is lazy instead of restoring in `onCreate(owner)`: restoring at `ON_CREATE` would hand out `false` during the launch pipeline and only load the real values afterward, so a recreation would silently re-run the login checks.

app/src/org/commcare/activities/home/HomeActivityHost.kt
OrangeAndGreen: [Human written, fact-checked by Claude]

> Think the refactor introduces another regression now given `refreshCCUpdateOption` would never be called now while it was being called earlier by activity call chain for `rebuildOptionsMenu`.

I don't think that's correct. Looking at master I see that function being called only in two places by `HomeScreenBaseActivity.handleAppUpdate` (in the `AVAILABLE` and `DOWNLOADING` cases), and those calls are unchanged in this PR. Are you sure that function is supposed to be called as part of `rebuildOptionsMenu`?

I think it actually goes the other way. `refreshCCUpdateOption` invalidates the menu so it gets rebuilt (but that's only the case in `StandardHomeActivity`, not `RootMenuHomeActivity`).


> Also is there a conceptual difference between `refreshActionSurface` and `rebuildOptionsMenu` ? Can we not use the `rebuildOptionsMenu` from activity callchain or putting in other words why do we want this to be in `HomeActivityHost.kt` at all ?

The main difference is that what gets refreshed may or may not involve refreshing the actual options menu. In `StandardHomeActivity` that's the case, but in `RootMenuHomeActivity` (today's example but this will apply to Opp Home later) it's the drawer that needs to be updated.

In the end I decided to rename `refreshCCUpdateOption` (named based on what triggered the call) to `refreshActionSurface` (named based on what the call does): [8f3064e](https://github.com/dimagi/commcare-android/pull/3834/commits/8f3064e3e7ee672e7ce158a289a5b5f282f1330b)

Now `HomeScreenBaseActivity` doesn't have to define an abstract function (the former `refreshCCUpdateOption`), and instead it comes from the inherited `HomeActivityHost` interface (now `refreshActionSurface`). Eventually the implementing classes will inherit this interface directly (when `HomeScreenBaseActivity` is deleted in [slice 8](https://dimagi.atlassian.net/browse/CCCT-2686)).

shubham1g5: You are right, think the [comit diff ](https://github.com/dimagi/commcare-android/commit/aa54c7b9d) was a misleading here as it shows `refreshCCUpdateOption` called under `rebuildOptionsMenu`. .

shubham1g5 (approved):

app/src/org/commcare/home/HomeActivityCoordinator.kt
Jignesh-dimagi: Ok got it. Makes sense.

Jignesh-dimagi (approved):

PR #3836, opened 2026-07-28
QA-8597 / QA-8594 Fix PersonalID Email OTP Error Messages

conroy-ricketts:
## Product Description

- **QA-8597** — verifying an email already tied to another account crashed the app; it now shows "This email is already linked to another account."
- **QA-8594** — a wrong verification code showed "You are not authorized to make this request."; it now shows the existing wrong-passcode message.

Both apply to signup, recovery, and the Manage Profile email edit.

Link to related issue in Firebase [here](https://console.firebase.google.com/project/commcare-a57e4/crashlytics/app/android:org.commcare.dalvik/issues/5542c95495917ac7bd7e145d9e0cc79f?time=90d&types=crash&versions=CommCare%202.64%20Beta%20(490711);2.64%20(490711);2.64%20(490000);2.64%20(4900001);2.64%20(1)&sessionEventKey=6A68A14B017F000159F937913773EEDC_2246002934789993000).

Screenshot_20260728-120618_CommCare Debug
Screenshot_20260728-120538_CommCare Debug

## Technical Summary

The email OTP calls carry no `PersonalIdSessionData`, so they never consulted `handleErrorCodeIfPresent` and failures were classified by HTTP status alone: `401 INCORRECT_OTP` became `FAILED_AUTH_ERROR`, and `400 EMAIL_ALREADY_IN_USE` became `BAD_REQUEST_ERROR`, which has no case in `PersonalIdOrConnectApiErrorHandler` and hit the `default` branch's `throw`. Status isn't enough here — Server also returns 401 for `LOCKED_ACCOUNT`, `INVALID_TOKEN` and `TOKEN_EXPIRED` — so `sessionData` is now optional and both calls route through a callback that reads `error_code`. `EMAIL_ALREADY_IN_USE` was the only code missing from the map.

### Safety story

I verified/reproduced both issues, including the crash, and confirmed the correct error messages now show in both cases.

### Automated test coverage

Four MockWebServer cases on the verification screen: duplicate email and wrong code assert their own messages, while `401 LOCKED_ACCOUNT` and `400 MISSING_DATA` assert they aren't misreported as either. The first reproduced the Firebase stack before the fix. 848 unit tests pass.

[QA-8597]: https://dimagi.atlassian.net/browse/QA-8597?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[QA-8594]: https://dimagi.atlassian.net/browse/QA-8594?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


conroy-ricketts:
## Suggested Review Order

[AI] Read the net diff rather than commit-by-commit — the first commit's approach (inferring the error from the bare 400) is replaced by the second, so `EmailHelper.kt` cancels out and isn't in the net diff at all.

- `app/src/org/commcare/connect/network/base/BaseApiHandler.kt` — the one new error code, which the rest hangs off
- `app/src/org/commcare/connect/network/connectId/PersonalIdApiHandler.java` — the fix: `sessionData` made optional, `EMAIL_ALREADY_IN_USE` added to the existing map, and the new callback that routes both email OTP calls through it
- `app/src/org/commcare/connect/network/PersonalIdOrConnectApiErrorHandler.java` — codes to strings; note the three generic cases that keep newly-precise codes off the throwing `default`
- `app/res/values/strings.xml` — English wording; the nine `values-*` files are its translations
- `app/unit-tests/.../PersonalIdEmailVerificationFragmentTest.kt` — worth reading closely: the `LOCKED_ACCOUNT` and `MISSING_DATA` cases are what pin the behavior, since both share a status code with the errors they must not be confused for

app/src/org/commcare/fragments/personalId/EmailHelper.kt
```diff
@@ -125,21 +125,35 @@ object EmailHelper {
...
+     * rejects a verify with 400 when the code was correct but the email already belongs to another
+     * active account — its remaining 400 branches need an empty email or OTP, which this screen
+     * cannot submit. A wrong code returns 401.
+     */
+    private fun resolveVerifyFailure(failureCode: PersonalIdOrConnectApiErrorCodes) =
+        if (failureCode == PersonalIdOrConnectApiErrorCodes.BAD_REQUEST_ERROR) {
+            PersonalIdOrConnectApiErrorCodes.EMAIL_ALREADY_IN_USE_ERROR
+        } else {
+            failureCode
+        }
```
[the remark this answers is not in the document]
conroy-ricketts: [AI] The first half is an accurate reading: every 400 on this endpoint becomes `EMAIL_ALREADY_IN_USE_ERROR`. That collapse is deliberate — the PR description and the KDoc on `resolveVerifyFailure` both record it as the cost of this approach.

The suggested test isn't implementable as written, though. `resolveVerifyFailure` receives only the already-mapped `PersonalIdOrConnectApiErrorCodes`; the response body is consumed upstream in `BaseApiCallback.processFailure` and never propagated, and the accompanying `Throwable` carries only `"Unknown failure with 400 for /users/verify_email_otp"`. So there is no path on which a generic 400 "remains `BAD_REQUEST_ERROR`" — a test asserting that distinction would fail by construction. Making it pass would require plumbing `error_code` down to the callback, which is the shared-network-layer change this PR deliberately avoids: the email OTP calls carry no `PersonalIdSessionData` and so never reach `PersonalIdApiHandler`'s existing `error_code` switch.

Why the collapse holds today, per connect-id's `verify_email_otp`: its only other 400 branches are `MISSING_DATA` (empty email or OTP) and `INVALID_DATA` (no OTP device for that user + email). This screen enforces a non-empty email and a 6-digit code, and the device row is created by the preceding `send_email_otp`, so neither is reachable from here. A wrong code returns 401, not 400.

The reachable version of the concern is future drift — if connect-id adds a 400 branch, this silently mislabels it. A characterization test asserting that a 400 carrying a *different* `error_code` still renders the email-in-use message would pin the collapse as intentional, so a later change breaks a test rather than shipping a wrong message. I can add that if reviewers want it.

app/src/org/commcare/connect/network/PersonalIdOrConnectApiErrorHandler.java
```diff
@@ -65,6 +65,17 @@ public static String handle(
                 return context.getString(R.string.personalid_incorrect_otp);
             case JSON_PARSING_ERROR:
                 return context.getString(R.string.personalid_network_response_parsing_error);
+            case EMAIL_ALREADY_IN_USE_ERROR:
+                return context.getString(R.string.personalid_email_already_in_use);
+            // Reachable from the email OTP endpoints now that their error_code is read: a malformed
+            // request, or a configuration session that expired mid-signup. Nothing actionable to tell
+            // the user beyond retrying, but they must not reach the throwing default below.
+            case BAD_REQUEST_ERROR:
+            case MISSING_DATA_ERROR:
```
shubham1g5: would this break error handling here - https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/fragments/personalId/PersonalIdPhoneFragment.java#L546

If the answer is yes, we should verify the safety story for codes added here as well.

shubham1g5 (commented):
PR seems high around AI comments verbosity and would be useful to trim any contextual AI commits.

app/src/org/commcare/connect/network/PersonalIdOrConnectApiErrorHandler.java
conroy-ricketts: Ah yeah, this may affect that slightly. I'll revisit this

conroy-ricketts: Yeah, these extra cases diverted from our fail-fast approach which also affected the error handling for the start configuration API call (i.e. showing a generic error message rather than crashing in some cases). I tweaked these changes so that it only affects email OTP verification instead

conroy-ricketts: https://github.com/dimagi/commcare-android/commit/17acc011e8a8ee5b314c1a9a24e2e115d30857d4

shubham1g5 (approved):

app/src/org/commcare/connect/network/connectId/PersonalIdApiHandler.java
```diff
@@ -198,6 +209,41 @@ private boolean handleErrorCodeIfPresent(
...
+                        || !handleErrorCodeIfPresent(errorCodes.getFirst(), errorCodes.getSecond(), null)) {
+                    super.processFailure(responseCode, url, errorBody, t);
+                }
+            }
+        };
+    }
+
+    private static boolean isEmailOtpErrorCode(String errorCode) {
+        return "INCORRECT_OTP".equals(errorCode) || "EMAIL_ALREADY_IN_USE".equals(errorCode);
+    }
```
Jignesh-dimagi: Similarly here, handleErrorCodeIfPresent should be sufficient to handle this error_code instead of introducing a new method

app/src/org/commcare/connect/network/connectId/PersonalIdApiHandler.java
```diff
@@ -198,6 +209,41 @@ private boolean handleErrorCodeIfPresent(
...
+                Pair<String, String> errorCodes = getErrorCodes(errorBody);
+
+                // Deliberately narrow: every other code keeps its existing status-based handling.
+                if (!isEmailOtpErrorCode(errorCodes.getFirst())
+                        || !handleErrorCodeIfPresent(errorCodes.getFirst(), errorCodes.getSecond(), null)) {
+                    super.processFailure(responseCode, url, errorBody, t);
+                }
+            }
+        };
+    }
```
Jignesh-dimagi: My main concern is that if each API creates its own custom callback, we start losing the benefit of centralized success and error handling across the codebase. 

verifyEmailOtp and sendEmailOtp should use existing PersonalIdApiHandler's createCallback and for that I think you should create NoParsingPersonalIdApiResponseParser extending PersonalIdApiResponseParser. verifyEmailOtp and sendEmailOtp should call this with sessionData as null. 

```
ApiPersonalId.sendEmailOtp(
                activity,
                email,
                personalIdConfigurationToken,
                user,
                createCallback((PersonalIdSessionData)null, new NoParsingPersonalIdResponseParser())
        );
```

That NoParsingPersonalIdApiResponseParser should have logic `responseCode >= 200 && responseCode < 300)`.

Your thoughts?

Jignesh-dimagi (changes requested):

app/src/org/commcare/connect/network/connectId/PersonalIdApiHandler.java
conroy-ricketts: Agreed!

No new parser was needed though because `createCallback(sessionData, null)` is a supported path, and with a null parser there's no body parsing to do which suits these endpoints since they return an empty `200`. So I used `createCallback((PersonalIdSessionData) null, null)`

I tested this and verified it works for the error paths and the success path

conroy-ricketts: 7209d63b1e6c47a42b3adadd7694a902beaf9d8d

app/src/org/commcare/connect/network/connectId/PersonalIdApiHandler.java
conroy-ricketts: 7209d63b1e6c47a42b3adadd7694a902beaf9d8d

app/src/org/commcare/connect/network/connectId/PersonalIdApiHandler.java
Jignesh-dimagi: I would have preferred using NoParsingPersonalIdResponseParser just to ensure the response code is validated and to stay consistent with our existing parsing structure, but that's completely optional.

Jignesh-dimagi (approved):

PR #3843, opened 2026-07-31
Tech Spec: Backup code and account recovery

shubham1g5:
[Jira](https://dimagi.atlassian.net/browse/CCCT-2677)
[Design Doc](https://docs.google.com/document/d/1EC8d9TnQMRAJbAAIGcVaaHE9VLc-stoYZjirMyuaGjw/edit?tab=t.0)

Reviewer Notes:

- Please review the design doc first
- Review spec directly (no commits structuring) 
- Can skip the section on Implementation Details (I have tried separating out the architectural changes from implementation details that Claude will need to take this spec forward) 
- Review in Android Studio for better formatting with markdown preview as the doc contains a flow chart.

### DIFF
--- a/docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
+++ b/docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
@@ -312,11 +312,14 @@ object PersonalIdReminderHelper {
 
 ### 1. Changing email (user has existing email)
 
-`PersonalIdProfileEditFragment.onSaveClicked()` when `isEmailModified()` and `user.email != null`:
-store new email in `PersonalIdProfileActivityViewModel.pendingEmail`, then navigate to
-`personalid_confirm_backup_code`.
+`PersonalIdSendEmailOtpFragment` replaces the existing `showEmailOtpConfirmationDialog` in this flow.
 
-On backup code confirmed → `PersonalIdSendEmailOtpFragment(email=pendingEmail, EXISTING_USER)`
+`PersonalIdProfileEditFragment.onSaveClicked()` when `isEmailModified()` and `user.email != null`:
+1. Store new email in `PersonalIdProfileActivityViewModel.pendingEmail`.
+2. If name is also modified, call `saveProfileDetails` first (name save must complete before navigating).
+3. Navigate to `personalid_confirm_backup_code`.
+
+On backup code confirmed → `PersonalIdSendEmailOtpFragment(email=pendingEmail, masked=true, EXISTING_USER)`
 → `PersonalIdEmailVerificationFragment(EXISTING_USER)` → `PersonalIdUnlocker.unlock(ALWAYS)`
 → persist new email locally, pop to profile with success toast.
 
@@ -329,7 +332,7 @@ nav action for the email-change exit path.
 store new email in `PersonalIdProfileActivityViewModel.pendingEmail`, then navigate to
 the Phone OTP verification screen(`PersonalIdProfilePhoneVerificationFragment`) (sends otp to `ConnectUserRecord.primaryPhone`)
 
-On phone OTP verified → `PersonalIdSendEmailOtpFragment(email=pendingEmail, EXISTING_USER)`
+On phone OTP verified → `PersonalIdSendEmailOtpFragment(email=pendingEmail, masked=false, EXISTING_USER)`
 → `PersonalIdEmailVerificationFragment(EXISTING_USER)` → `PersonalIdUnlocker.unlock(ALWAYS)`
 → persist new email locally, pop to profile with success toast.
 
--- a/docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
+++ b/docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
@@ -37,29 +37,34 @@ Manage Profile
  │                                                    ▼
  │                                           ◄ Profile (success)
  │
- └─[Save — email change]─────────────────────────────────────────────────┐
-                                                                         ▼
-                                              [has email]           [no email]
-                                                   │                     │
-                                                   ▼                     ▼
-                                        Confirm backup code    Send Phone OTP screen
-                                         │           │                   │
-                                [Forgot?]│  [correct]│                   ▼
-                                         ▼           │           Phone OTP entry
-                              (→ same as Change      │                   │
-                               backup code           └─────────┬─────────┘
-                               forgot flow)                    │
-                                                               ▼
-                                                   Send email OTP screen
-                                                               │
-                                                               ▼
-                                                       Email OTP entry
-                                                               │
-                                                               ▼
-                                                   Biometric / PIN unlock
-                                                               │
-                                                               ▼
-                                                    ◄ Profile (success)
+ └─[Save — email change]────────────────────────────────────────────────────┐
+                                                                            ▼
+                                                            Confirm backup code
+                                                           │                │
+                                                  [Forgot?]│       [correct]│
+                                                           │                │
+                                               [has email] │  [no email]    │
+                                                      │         │           │
+                                                      ▼         ▼           │
+                                          (→ same as    Send Phone OTP      │
+                                           Change        screen             │
+                                           backup code        │             │
+                                           forgot flow)       ▼             │
+                                                        Phone OTP entry     │
+                                                                   │        │
+                                                                   └────────┘
+                                                                        │
+                                                                        ▼
+                                                            Send email OTP screen
+                                                                        │
+                                                                        ▼
+                                                                Email OTP entry
+                                                                        │
+                                                                        ▼
+                                                            Biometric / PIN unlock
+                                                                        │
+                                                                        ▼
+                                                             ◄ Profile (success)
 ```
 
 **Account Configuration Flow (`nav_graph_personalid.xml`) — additions only:**
@@ -89,7 +94,8 @@ BasePersonalIdBackupCodeFragment
 │   auth      : PersonalID session token
 │   nav graph : nav_graph_personalid
 └── PersonalIdProfileBackupCodeFragment 
-    workflows : CONFIRM_BACKUP_CODE, SET_NEW_CODE
+    workflows : CONFIRM_BACKUP_CODE_CHANGE_CODE, CONFIRM_BACKUP_CODE_CHANGE_EMAIL,
+                CONFIRM_BACKUP_CODE_ADD_EMAIL, SET_NEW_CODE
     auth      : ProvidedAuth(userId, password)
     nav graph : nav_graph_personalid_profile
 ```
@@ -111,12 +117,14 @@ abstract fun setupHeader()                // implementations own full header UI 
 
 #### `BackupCodeWorkflow` enum
 
-| Value | API call | API Auth | Biometric/Pin Unlock on Save |
-|---|---|---|------------------------------|
-| `REGISTRATION` | none (stored in session) | — | no                           |
-| `CONFIRM_RECOVERY` | `users/recover/confirm_backup_code` | session token | no                           |
-| `CONFIRM_BACKUP_CODE` | none (local validation against `ConnectUserRecord.password`) | — | no                           |
-| `SET_NEW_CODE` | `/users/set_backup_code` | user credentials or session token | yes                          |
+| Value | API call | API Auth | Biometric/Pin Unlock on Save | Forgot? routing |
+|---|---|---|---|---|
+| `REGISTRATION` | none (stored in session) | — | no | n/a |
+| `CONFIRM_RECOVERY` | `users/recover/confirm_backup_code` | session token | no | n/a |
+| `CONFIRM_BACKUP_CODE_CHANGE_CODE` | none (local validation against `ConnectUserRecord.password`) | — | no | no email → Manage Profile toast; has email → email OTP recovery |
+| `CONFIRM_BACKUP_CODE_CHANGE_EMAIL` | none (local validation against `ConnectUserRecord.password`) | — | no | → email OTP recovery (has email guaranteed) |
+| `CONFIRM_BACKUP_CODE_ADD_EMAIL` | none (local validation against `ConnectUserRecord.password`) | — | no | → Send Phone OTP screen |
+| `SET_NEW_CODE` | `/users/set_backup_code` | user credentials or session token | yes | n/a |
 
 ### API Changes
 
@@ -192,9 +200,11 @@ to save time during spec review.
 **Flow:**
 
 1. Tap "Change backup code" → `personalid_confirm_backup_code`
-   (`PersonalIdProfileBackupCodeFragment(CONFIRM_BACKUP_CODE)`)
+   (`PersonalIdProfileBackupCodeFragment(CONFIRM_BACKUP_CODE_CHANGE_CODE)`)
    - One field, no API call — local validation only (compare against `ConnectUserRecord.password`)
-   - "Forgot backup code?" link visible only if `user.email != null`
+   - "Forgot backup code?" link always visible. If `user.email == null`, tapping it redirects
+     immediately to Manage Profile with toast "Please add email to recover your backup code".
+     If `user.email != null`, navigates to email OTP recovery.
 
 2. On submit → `PersonalIdProfileBackupCodeFragment(SET_NEW_CODE)`
    - Two fields (new + confirm)
@@ -317,27 +327,27 @@ object PersonalIdReminderHelper {
 `PersonalIdProfileEditFragment.onSaveClicked()` when `isEmailModified()` and `user.email != null`:
 1. Store new email in `PersonalIdProfileActivityViewModel.pendingEmail`.
 2. If name is also modified, call `saveProfileDetails` first (name save must complete before navigating).
-3. Navigate to `personalid_confirm_backup_code`.
+3. Navigate to `personalid_confirm_backup_code` with `CONFIRM_BACKUP_CODE_CHANGE_EMAIL`.
 
 On backup code confirmed → `PersonalIdSendEmailOtpFragment(email=pendingEmail, masked=true, EXISTING_USER)`
 → `PersonalIdEmailVerificationFragment(EXISTING_USER)` → `PersonalIdUnlocker.unlock(ALWAYS)`
 → persist new email locally, pop to profile with success toast.
 
-The `PersonalIdProfileBackupCodeFragment` is reused from Area 1 with a separate
-nav action for the email-change exit path.
+On Forgot? → email OTP recovery (email guaranteed; same forgot flow as `CONFIRM_BACKUP_CODE_CHANGE_CODE` with email).
 
 ### 2. Adding email (no existing email, already signed in)
 
 `PersonalIdProfileEditFragment.onSaveClicked()` when `isEmailModified()` and `user.email == null`:
-store new email in `PersonalIdProfileActivityViewModel.pendingEmail`, then navigate to
-the Phone OTP verification screen(`PersonalIdProfilePhoneVerificationFragment`) (sends otp to `ConnectUserRecord.primaryPhone`)
+1. Store new email in `PersonalIdProfileActivityViewModel.pendingEmail`.
+2. Navigate to `personalid_confirm_backup_code` with `CONFIRM_BACKUP_CODE_ADD_EMAIL`.
 
-On phone OTP verified → `PersonalIdSendEmailOtpFragment(email=pendingEmail, masked=false, EXISTING_USER)`
+On backup code confirmed → `PersonalIdSendEmailOtpFragment(email=pendingEmail, masked=false, EXISTING_USER)`
 → `PersonalIdEmailVerificationFragment(EXISTING_USER)` → `PersonalIdUnlocker.unlock(ALWAYS)`
 → persist new email locally, pop to profile with success toast.
 
-Note: On manage profile screen under "Email" row, if `user.email == null`, the row should show a hint message -
-"A verification code will be sent to your registered phone number to add an email address." 
+On Forgot? → `PersonalIdProfilePhoneVerificationFragment` (sends OTP to `ConnectUserRecord.primaryPhone`)
+→ on phone OTP verified → `PersonalIdSendEmailOtpFragment(email=pendingEmail, masked=false, EXISTING_USER)`
+→ (same path as above).
 
 ### 3. Skip email dialog during signup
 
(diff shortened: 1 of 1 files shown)


docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
```diff
@@ -0,0 +1,367 @@
...
+
+### `PersonalIdSendEmailOtpFragment`
+
+Reusable screen that shows the email address an OTP will be sent to, with a "Send code" button.
+Nav args: `email: String`, `masked: Boolean`, `workflow: EmailWorkFlow`.
+
+- `masked = true`: displays `email` as `x***y@gmail.com` (existing account email)
+- `masked = false`: displays `email` as-is (newly entered email)
+
+On "Send code" → calls `send_email_otp` → navigates to `PersonalIdEmailVerificationFragment`.
```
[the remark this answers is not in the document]
shubham1g5: Skipping as the `PersonalIdEmailVerificationFragment` is exisiting functionality with the required features mentioned here already in place.

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
```diff
@@ -0,0 +1,367 @@
...
+```
+BasePersonalIdBackupCodeFragment
+├── PersonalIdBackupCodeFragment 
+│   workflows : REGISTRATION, CONFIRM_RECOVERY
+│   auth      : PersonalID session token
+│   nav graph : nav_graph_personalid
+└── PersonalIdProfileBackupCodeFragment 
+    workflows : CONFIRM_BACKUP_CODE, SET_NEW_CODE
+    auth      : ProvidedAuth(userId, password)
+    nav graph : nav_graph_personalid_profile
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3843/changes/197ac6dbdf9a904e8e5760c333de382b1968cd93

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
```diff
@@ -0,0 +1,367 @@
...
+- `masked = false`: displays `email` as-is (newly entered email)
+
+On "Send code" → calls `send_email_otp` → navigates to `PersonalIdEmailVerificationFragment`.
+
+### `PersonalIdProfileActivityViewModel`
+
+Single activity-scoped ViewModel on `PersonalIdProfileActivity`. Consolidates profile display and cross-fragment workflow state:
+- `profileDisplayModel: LiveData<PersonalIdProfileDisplayModel>` — replaces the existing fragment-scoped `PersonalIdProfileViewModel`
+- `currentCode: String?` — entered at `CONFIRM_BACKUP_CODE` step, used as auth for the final API call
+- `pendingEmail: String?` — set when email change is gated behind backup code confirmation
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3843/changes/197ac6dbdf9a904e8e5760c333de382b1968cd93

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
```diff
@@ -0,0 +1,367 @@
...
+                                                   Send email OTP screen
+                                                               │
+                                                               ▼
+                                                       Email OTP entry
+                                                               │
+                                                               ▼
+                                                   Biometric / PIN unlock
+                                                               │
+                                                               ▼
+                                                    ◄ Profile (success)
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3843/changes/197ac6dbdf9a904e8e5760c333de382b1968cd93

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
```diff
@@ -0,0 +1,366 @@
...
+ │                                                    ▼
+ │                                          Set new backup code
+ │                                                    │ Save
+ │                                                    ▼
+ │                                          Biometric / PIN unlock
+ │                                                    │
+ │                                                    ▼
+ │                                           ◄ Profile (success)
+ │
+ └─[Save — email change]─────────────────────────────────────────────────┐
```
OrangeAndGreen: Just making sure, is this equivalent to "Change email", i.e. to go along with "Change backup code" above? Thinking it might just be weird Claude wording...

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
```diff
@@ -0,0 +1,366 @@
...
+
+## Navigation Overview
+
+New screens and flows added across both nav graphs.
+
+**Profile graph (`nav_graph_personalid_profile.xml`):**
+
+```
+Manage Profile
+ ├─[Change backup code]──────────────────────────────────────────────────┐
```
OrangeAndGreen, line 17: I'm wondering if it makes more sense to do the biometric/PIN unlock at the beginning of these workflows instead of at the end? It seems like that should be a quick gate for the intended user to pass before starting the more involved steps (OTPs, logic branches, etc.), whereas for someone who isn't going to pass the unlock anyway it would prevent them from wasting their time (and any money we might spend sending phone OTPs, like in the change-email workflow when there's no existing email).

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
```diff
@@ -0,0 +1,366 @@
...
+ │                                                    │ Save
+ │                                                    ▼
+ │                                          Biometric / PIN unlock
+ │                                                    │
+ │                                                    ▼
+ │                                           ◄ Profile (success)
+ │
+ └─[Save — email change]─────────────────────────────────────────────────┐
+                                                                         ▼
+                                              [has email]           [no email]
```
OrangeAndGreen: I'm wondering why we're using the backup code as a gate in the has-email flow, but phone OTP as a gate in the no-email flow? Couldn't we use backup code in both cases?

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
```diff
@@ -0,0 +1,366 @@
...
+                                                   │                     │
+                                                   ▼                     ▼
+                                        Confirm backup code    Send Phone OTP screen
+                                         │           │                   │
+                                [Forgot?]│  [correct]│                   ▼
+                                         ▼           │           Phone OTP entry
+                              (→ same as Change      │                   │
```
OrangeAndGreen: "Same as"... just clarifying this means proceed to the email OTP screen, and if they can verify their old email they can set a new one?

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
```diff
@@ -0,0 +1,366 @@
...
+
+```kotlin
+abstract fun getPhoneNumber(): String
+abstract fun createOtpManager(): OtpManager
+abstract fun onOtpVerified()
+abstract fun recordFailedAttempt()
+abstract fun getOtpAttemptCount(): Int
+```
+
+### `PersonalIdSendEmailOtpFragment`
```
conroy-ricketts, line 173: Is the intention here to delete the current "Verify your new email" dialog? We can be explicit about deleting the old dialog in favor of reusing this Fragment instead to avoid essentially duplicating UI, and the implications of that

conroy-ricketts (commented):
Thanks for clearly separating out the implementation reference for Claude! That made this a lot easier to review and is a good model for specs going forward

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
shubham1g5: Currently user gets asked to verify biometric as soon as they click on "manage profile" screen and also when they open the app for first time. So I was thinking of this as more of a final check to verify that the phone is in hands of the device owner as close to the exact moment they are making this change.

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
shubham1g5: It's the email change workflow in the Manage Profile that already exists today (Save refers to the save on mange profile) but needs modifications as outlined here to be more secure.

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
shubham1g5: [The design doc explains it here](https://docs.google.com/document/d/1EC8d9TnQMRAJbAAIGcVaaHE9VLc-stoYZjirMyuaGjw/edit?tab=t.0#heading=h.44bwazd4yq4), let me know if you are not clear even after giving that a read.

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
```diff
@@ -0,0 +1,366 @@
...
+
+#### New: `POST /users/set_backup_code`
+
+We currently set the backup code only during registration using the `complete_profile` endpoint.
+This new endpoint allows users to change their backup code after registration or during account recovery.
+
+| Field | Value                                                                                       |
+|---|---------------------------------------------------------------------------------------------|
+| Auth | `ProvidedAuth(userId, password)` (change flow) or `TokenAuth(sessionToken)` (recovery flow) |
+| Request | `{ "recovery_pin": "<new_code>" }`                                                          |
```
Jignesh-dimagi, line 139: Should we validate the old back up code on the API side instead of checking it locally? Moving that check to the server would be much safer from a security perspective.

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
shubham1g5: right.

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
shubham1g5: Good catch, didn't realise we arleady have an dialog that this screen can replace. Made it explicit with implementation details here - https://github.com/dimagi/commcare-android/pull/3843/changes/1f074d84ad47064678e6060dbea9ee6981bac704

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
```diff
@@ -0,0 +1,369 @@
...
+
+Reminder cadence:
+- First: 1h after registration
+- Second: ~3d after registration (scheduled from current time when first fires)
+- Subsequent: every 7d rolling
+
+### Hook
+
+All CommCare launching activities (e.g. `ConnectActivity`, `LoginActivity`, `StandardHomeActivity`) check for due reminders on resume. 
+If due, show the reminder dialog.
```
Jignesh-dimagi, line 290: My main concern here is that we keep triggering these dialog prompts directly inside the Activity. This could lead to multiple dialogs popping up at the same time—for example, if a user doesn't have an email set, they get prompted to add one, but they might simultaneously trigger a prompt to update their backup code.

We should consider establishing a priority queue or hierarchy for these dialogs so only one is presented to the user at a time.

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
```diff
@@ -0,0 +1,369 @@
...
+
+`onResume()` calls `PersonalIdReminderHelper.isDue()`. If due, show
```
Jignesh-dimagi, line 292: Since this is triggered in onResume, it will keep popping up every time the activity resumes—even if the user dismissed it the first time. Should we add a cooldown or delay (e.g., waiting at least 4 hours or 1 day) before re-triggering this dialog?"

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
Jignesh-dimagi: Small concern here: since we're calling the SET_NEW_CODE API before biometric validation, what happens if the user fails or cancels the biometric check?
The server will already have saved the new backup code, but the app might treat the operation as failed, leading to a state mismatch.

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
shubham1g5: this will use the preference `KEY_BACKUP_CODE_REMINDER_NEXT_DUE` to know about the next time we should remind as per the `Scheduling` section above.

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
shubham1g5: Agree. can you remind me if we have the email pop-up already on launch screens ?

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
shubham1g5: @Jignesh-dimagi given we already use the ``ProvidedAuth(userId, password)` for API calls while logged in, a local comparison here is essentially the same as server comparison because the subsequent API call (for eg. to set a new back up code will fail otherwise) 

We are essentially doing 2 checks - 

- user entered code is same as `password`
- API call is successful using `password`

which essentially confirms that 

- user entered code is same as server stored password

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
Jignesh-dimagi: Yeah, it's getting triggered from [here](https://github.com/dimagi/commcare-android/blob/e1c8ba80ab43114c48d6eda3dd73e5cc724ce194/app/src/org/commcare/activities/StandardHomeActivity.java#L62).

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
Jignesh-dimagi: Understood, that makes sense.

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
shubham1g5: Ahh the flow chart makes it hard to understand but the idea is to do the Biometric check on the set new backup code page itself before making the API call (it's not a separate page but the auth pop up on the same page)

OrangeAndGreen: The biometric check at the end feels odd to me:
1. If they passed biometric just before deciding to change their email or backup code (when entering the profile page), then having them do it again shortly later seems unnecessary
2. This doesn't seem consistent with the other places where we do unlocks: registering/recovering PersonalID, accessing and sending messages, etc. In those cases one unlock at the beginning of the workflow is sufficient.

That being said, I'm okay with keeping this as is if we feel the extra security is needed here.

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
OrangeAndGreen: Ah, right. Something still feels not-ideal to me, but I'm not sure I have a better suggestion so maybe just best to go with this. I'm thinking maybe we need to separate this into two cases:
1. User has no email set but remembers their backup code
   We can have the user verify their backup code, then set an email address
2. User has no email set and doesn't remember their backup code
   Now we need to use OTP, but we should have the user set both an email and a new backup code

So I'd propose that we ask for the backup code first, and if the user can't remember it then we fallback to OTP and add a "change backup code" step. That would give users a chance to provide their backup code instead of going straight to phone OTP, which is the mechanism we'd prefer anyway and also prevents sending a costly SMS message. Granted this also makes the workflow a bit more complex...

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
shubham1g5: I was thinking of an attack scenario when user leaves his phone un-attended after entering the manage profile phone and a co-worker noticing this try to utilize the opportunity to manipulate the account factors that can directly lead to their account being hacked. Think it's good in general to try to be as secure as possible when dealing with such sensitive account operations.

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
shubham1g5: That resonates with me, made the change here - https://github.com/dimagi/commcare-android/pull/3843/changes/419d992455bd3c0149812cdf1bf76027ff47a50d

OrangeAndGreen (dismissed):

conroy-ricketts (dismissed):

Jignesh-dimagi (dismissed):

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
```diff
@@ -0,0 +1,370 @@
...
+### API Changes
+
+#### New: `POST /users/set_backup_code`
+
+We currently set the backup code only during registration using the `complete_profile` endpoint.
+This new endpoint allows users to change their backup code after registration or during account recovery.
+
+| Field | Value                                                                                       |
+|---|---------------------------------------------------------------------------------------------|
+| Auth | `ProvidedAuth(userId, password)` (change flow) or `TokenAuth(sessionToken)` (recovery flow) |
```
calellowitz: Is `ProvidedAuth` basic auth?

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
```diff
@@ -0,0 +1,370 @@
...
+| `REGISTRATION` | none (stored in session) | — | no | n/a |
+| `CONFIRM_RECOVERY` | `users/recover/confirm_backup_code` | session token | no | n/a |
+| `CONFIRM_BACKUP_CODE_CHANGE_CODE` | none (local validation against `ConnectUserRecord.password`) | — | no | no email → Manage Profile toast; has email → email OTP recovery |
+| `CONFIRM_BACKUP_CODE_CHANGE_EMAIL` | none (local validation against `ConnectUserRecord.password`) | — | no | → email OTP recovery (has email guaranteed) |
+| `CONFIRM_BACKUP_CODE_ADD_EMAIL` | none (local validation against `ConnectUserRecord.password`) | — | no | → Send Phone OTP screen |
+| `SET_NEW_CODE` | `/users/set_backup_code` | user credentials or session token | yes | n/a |
+
+### API Changes
+
+#### New: `POST /users/set_backup_code`
```
calellowitz: Existing workflows require confirmation of newly set backup codes to ensure the user entered the backup code they intended (and know what they put in). Are we intentionally removing that step?

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
```diff
@@ -0,0 +1,370 @@
...
+
+We currently set the backup code only during registration using the `complete_profile` endpoint.
+This new endpoint allows users to change their backup code after registration or during account recovery.
```
calellowitz: There is already view to set backup codes after registration `set_recovery_pin`, that seems to work very similarly to how you are describing. Are you able to reuse that one?

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
```diff
@@ -0,0 +1,370 @@
...
+ │                                            [email?]             Set new backup code
+ │                                         no │    │ yes                      │
+ │                                             ▼    ▼                         ▼
+ │                                   ◄ Profile   Send email OTP screen  Biometric / PIN unlock
+ │                                   (toast)          │                       │
+ │                                                    ▼                       ▼
+ │                                           Email OTP entry          ◄ Profile (success)
+ │                                                    │
+ │                                                    ▼
+ │                                          Set new backup code
```
calellowitz, line 32: why do we allow setting a new backup code before biometric/PIN unlock? That allows them to change the profile info, before they are fully authenticated. Setting a backup code isn't really part of the authentication process in this flow, but is happening part way through. If we moved it to the end, after the success, we would only need to allow logged in users to change the code. Rather than users part way through a recovery and would therefore only need to support one authentication method (and one view that already exists).

calellowitz (commented):
One real concern about the flow, and a few smaller questions

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
shubham1g5: yes

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
shubham1g5: we are not, the [UI on this page](https://claude.ai/design/p/fea1ba1c-f472-4fb0-a24d-68dc24af4d3e?file=Change+Backup+Code.dc.html) presents users with 2 input fields for user to verify the backup code before we make the API call.

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
shubham1g5: ohh nice! We can definitely reuse it if it already exists today (FYI @OrangeAndGreen )

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
shubham1g5: Ahh the flow diagram makes it a bit confusing but the  Biometric unlock is the pop on the same page that happens before we make the API call. So the order is more like this - 

1. User enters and confirms backup code on "Set New Backup Code Page"
2. We show the biometric unlock pop on the same page. If failed, the only option user have is to press back from here. 
3. If success, we make the API call to set the backup code 

> Rather than users part way through a recovery and would therefore only need to support one authentication method (and one view that already exists).

We do need to support setting a new back up code during Account recovery /configuration flow otherwise how will we let users signing on a device but who have forgotten their backup code recover their accounts ?

calellowitz: > how will we let users signing on a device but who have forgotten their backup code recover their accounts 

We will set it after they successfully recover their account. If they have recovered their account by other means, they will have a password for basic auth. If they have not finished recovering, it is not safe to allow them to modify the authentication methods for the account.

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
OrangeAndGreen: Confirmed! Cal and I discussed yesterday and it looks like the existing call does exactly what we need. So much for my first web sprint ticket! :-)

docs/superpowers/specs/2026-07-28-personalid-backup-code-management-design.md
shubham1g5: revised spec - https://github.com/dimagi/commcare-android/pull/3843/commits/f5545409d76cc6efe20eb3882c9be99266ee33d3 and https://github.com/dimagi/commcare-android/pull/3843/changes/eeeef491299f2a2c07f4c77b54b49ef64564e554

PR #3844, opened 2026-07-31
CCCT-2520 Start-Of-App Routing And Back Navigation Spec

conroy-ricketts:
## Technical Summary

Spec for the Connect redesign's start-of-app routing and app-wide back navigation. The plan is to create new tickets once it is approved.

### Safety story

What gives me confidence:
- This PR is for a spec only — it changes no code and will not be merged.
- The spec was iterated over several review rounds with input from stakeholders.

[CCCT-2520]: https://dimagi.atlassian.net/browse/CCCT-2520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CCCT-2522]: https://dimagi.atlassian.net/browse/CCCT-2522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,201 @@
...
+**Design doc:** [Mobile App Redesign — UX](https://docs.google.com/document/d/1rntc16FW2Jfr6CbzNcZ9RDOxIRrmfcWNU6_d5WL1rA8/edit?tab=t.qf5nvzne52l2)
+
+## Summary
+
+Two coupled navigation changes for the Connect UI redesign:
+
+1. **Start-of-app routing** — a run-once resolver that lands each user on the correct surface for their configuration and session state, with no login/setup screens flashing through.
+2. **App-wide back navigation** — one consistent, flag-free back model spanning the Intro page, the top-level navigation drawer, and the app-home surfaces.
+
+The spec defines a **north-star** target (a two-tier navigation shell with a single startup router) and a **pragmatic interim** that reaches the target behavior within today's activity topology without disrupting traditional CommCare users.
```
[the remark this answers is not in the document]
conroy-ricketts: [AI] (Claude Opus 4.8) Good call — clarified in the Summary that the flag-free back model is the **north-star** target; the interim still relies on `appLaunchedFromConnect` / `finishAffinity` / `REORDER_TO_FRONT` (already stated in the Technical gap analysis, now cross-referenced from the Summary). Addressed in 7d79af65c.

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,201 @@
...
+
+**The Connect path collapses Tier 2 in place.** Per [#3776](https://github.com/dimagi/commcare-android/pull/3776), `OpportunityHomeActivity` is itself a drawer-host surface that **attaches an app session in place** (no separate home activity, no for-result hop). So the for-result seam describes the *traditional/PersonalID* path; the *Connect* path gains app-home capability within the Opp Home surface itself. #3776's `HomeActivityCoordinator` + delegates + `attachSession`/`detachSession` decouple app-home behavior from the session gate and from any one activity — which is precisely the mechanism that makes the single-`NavHost` north-star reachable later. This work is a stepping stone toward the Shell, not a detour from it.
+
+**The startup router** is a run-once resolver (not a resident activity) that replaces `DispatchActivity`'s routing role. `DispatchActivity` remains the manifest launcher in the interim (it still owns DB-bad-state, recovery, and external/session-endpoint launches), but the landing decision moves into the resolver and stops re-evaluating in `onResume()` for redesign paths.
+
```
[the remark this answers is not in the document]
conroy-ricketts: [AI] (Claude Opus 4.8) Agreed. Added a new **first** precedence step: explicit intent-driven launches (external `ACTION_VIEW` installs via `CommCareSetupActivity`, `KEY_REQUIRE_REFRESH` verifications via `CommCareVerificationActivity`, deep links, push) are resolved before configuration-based routing, and remain `DispatchActivity`'s responsibility in the interim. Addressed in 7d79af65c.

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,201 @@
...
+| Last session context | **new persistence — see below** |
+
+**Precedence (first match wins):**
+
+1. **Active session alive → resume its home, no login shown.** Session on an opportunity-linked app → `OpportunityHomeActivity` for that opportunity; otherwise → `StandardHomeActivity` / `RootMenuHomeActivity`.
+2. **No active session → resolve by configuration** (table below). Entering a Connect/PersonalID surface may require a **PersonalID unlock** (biometric/PIN via [`PersonalIdUnlocker`](https://github.com/dimagi/commcare-android/blob/dc7697645fefd99de4e234be569bd8447fb6e0ba/app/src/org/commcare/personalId/PersonalIdUnlocker.kt)); on cancel or failure the prompt dismisses and the underlying screen is left unchanged (both route through the same `connectActivityComplete(false)` path today), rather than falling back to Intro. Traditional CommCare users never hit this gate — they have no PersonalID. The router only *lands* on `OpportunityHomeActivity`; [#3776](https://github.com/dimagi/commcare-android/pull/3776)'s controller performs the in-place silent app-login on resume.
```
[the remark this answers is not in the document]
conroy-ricketts: [AI] (Claude Opus 4.8) Addressed. The active-session step now resumes by the persisted login provenance (the "last session context" marker) rather than linkage alone, with an explicit tie-break: if provenance disagrees with the seated app's current `evaluateAppState`, provenance wins (linkage may be stale). Addressed in 7d79af65c.

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,201 @@
...
+| PersonalID status | [`PersonalIdManager.isloggedIn()`](https://github.com/dimagi/commcare-android/blob/dc7697645fefd99de4e234be569bd8447fb6e0ba/app/src/org/commcare/connect/PersonalIdManager.java#L149) |
+| Connect access + opportunities | [`ConnectUserDatabaseUtil.hasConnectAccess()`](https://github.com/dimagi/commcare-android/blob/dc7697645fefd99de4e234be569bd8447fb6e0ba/app/src/org/commcare/connect/database/ConnectUserDatabaseUtil.java#L44) + opportunity records |
+| Installed apps present | [`MultipleAppsUtil.usableAppsPresent()`](https://github.com/dimagi/commcare-android/blob/dc7697645fefd99de4e234be569bd8447fb6e0ba/app/src/org/commcare/utils/MultipleAppsUtil.java#L50) || Last accessed opportunity | **new persistence — see below** |
+| Last session context | **new persistence — see below** |
```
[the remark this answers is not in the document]
conroy-ricketts: [AI] (Claude Opus 4.8) Fixed — split the merged row so "Installed apps present" and "Last accessed opportunity" are separate two-cell rows. Addressed in 7d79af65c.

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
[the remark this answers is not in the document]
conroy-ricketts: [AI] (Claude Opus 4.8) Good catch — "dismiss, screen unchanged" only applies to a **warm** entry. Clarified: on a **cold start** (nothing beneath), unlock cancel/failure now **exits the app** rather than leaving a partially-initialized home or falling back to Intro/`DispatchActivity`. Addressed in 7d79af65c.

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,201 @@
...
+- **Last session context** — a marker (`manual` / `PersonalId-non-opportunity` / `PersonalId-on-opportunity-X`) that lets the **"Both Connect and non-Connect usage"** scenario (above) route deterministically when no session is live, instead of guessing. This replaces the design doc's vague "last track used."
+- **Terminal-state acknowledgment** — a per-opportunity flag recording that the user has already seen an opportunity's terminal state since it transitioned. **Rule:** if the last-accessed opportunity is now terminal — expired, cancelled, **or completed** — land on its Opp Home **once** (the first open since the transition, so the user sees the ended/completed state), set the flag, and on later launches fall back to the Opp List. This is based on the *locally-known* status: if connectivity hasn't yet surfaced the transition, routing proceeds to Opp Home as normal and the terminal state appears after the next sync.
+
+These are stored in a new dedicated shared-preferences store.
```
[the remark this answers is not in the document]
conroy-ricketts: [AI] (Claude Opus 4.8) Agreed. The dedicated prefs store is now specified as **scoped to the active PersonalID account** — namespaced per account, cleared on `forgetUser()` / account switch, never inherited across accounts, and pre-existing unscoped values discarded on first run. Addressed in 7d79af65c.

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,201 @@
...
+
+## App-wide back-navigation model
+
+**Governing principle: Back is temporal and flag-free.** Back reverses the last navigation step and is never conditioned on a hidden state flag (`appLaunchedFromConnect` is removed in the north-star). Five rules make it deterministic:
+
+1. **Start destination per host.** Back from the start destination finishes the host, revealing whatever launched it — or exiting if the host is the task root. This is the correct-stack replacement for the exit flag.
+2. **Top-level sections don't stack.** Navigating among the drawer's main sections pops to the start destination; Back from a section → start destination.
+3. **Detail screens stack temporally.** Pushed screens return to their opener on Back.
+4. **Login validity.** The Intro entry and the app-login bottom sheet are valid only with no active session for the target app; they are never revealed beneath a live or just-closed session. Session transitions (logout, expiry) route **explicitly** — they are not "Back."
+5. **Notification deep links synthesize a logical parent stack** so Back from a deep-linked destination reaches its parent, not the app exit.
```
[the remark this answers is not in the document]
conroy-ricketts: [AI] (Claude Opus 4.8) Added an explicit session-expiry rule (foreground-visible even when resumed from the background): **Standard Home** expiry → **CommCare Apps list**, not relying on `DispatchActivity` re-dispatch. One deliberate deviation from the suggestion: **Opp Home** expiry does *not* drop to Opp List — per #3776 it detaches and **silently re-establishes on the next resume**, which is the intended seamless behavior. Addressed in 7d79af65c.

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,201 @@
...
+
+**Logout vs forget-PersonalID (distinct concepts):**
+- **App-session logout** (all personas) ends only that app session; the PersonalID account stays registered. It routes to the launcher's parent list: traditional/PersonalID → CommCare Apps list; Connect → Opp List. For Opp Home, logout must **navigate away** to Opp List as it ends the session, since Opp Home's auto-login would otherwise immediately re-establish it.
+- **Forget PersonalID** de-registers the account (re-registration required) and always returns to Intro (via [`forgetUser()`](https://github.com/dimagi/commcare-android/blob/dc7697645fefd99de4e234be569bd8447fb6e0ba/app/src/org/commcare/connect/PersonalIdManager.java#L175) → `DispatchActivity` with `CLEAR_TASK`). It lives in Profile and is out of scope beyond noting the distinction. Note it does **not** currently close an active CommCare app session — it clears PersonalID data only; whether it should is an open question.
```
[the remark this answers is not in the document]
conroy-ricketts: [AI] (Claude Opus 4.8) Agreed and specified: since forget is only reachable by a PersonalID user (whose active session is PersonalID-backed), de-registration must also `closeUserSession()` before routing to Intro so no PersonalID-less session dangles. The spec now notes `forgetUser()` does not do this today and requires it. Addressed in 7d79af65c.

conroy-ricketts:
Hi @ctsims @avazirna, this spec touches traditional CommCare so I requested your reviews as well. I also have a document [here](https://docs.google.com/document/d/1rntc16FW2Jfr6CbzNcZ9RDOxIRrmfcWNU6_d5WL1rA8/edit?tab=t.ongpgraabwyz) that covers these changes at a high-level and may be worth a read before diving deep into the spec

shubham1g5:
@conroy-ricketts Flagging that this spec is overly verbose and think we should be able to simplify it greatly by deviating from text heavy approach to more use visualization techniques to explain the flow and just listing he new changes we need to make. The current text is also heavy on AI jargon and I would encourage re-writing most of this spec in a simplified manner (avoid use of AI if needed) 

Also the user-facing requirements from the spec (eg. desired user flows ) needs to be defined first i[n the design doc](https://docs.google.com/document/d/1rntc16FW2Jfr6CbzNcZ9RDOxIRrmfcWNU6_d5WL1rA8/edit?tab=t.ongpgraabwyz#heading=h.hactq68ebjby)  for a product review and this doc should just explain the tech challeges and solution approach for achieving the desired flow.

conroy-ricketts:
@shubham1g5 

Agreed, that's a fair point. I added new diagrams to both the tech spec here and the design doc. Also, I trimmed down the tech spec and separated out the section that does not need to be reviewed now (i.e. can be deferred to later code review). The diagrams actually came out pretty good in GitHub's rich diff preview.

I believe what's left now in both the design doc and the tech spec is only the information/context necessary for review, with user flows laid out in the design doc.

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,112 @@
...
+
+**Where the app opens**
+
+```mermaid
+flowchart TD
+    Open([User opens the app]) --> Active{Session still active?}
+    Active -- Yes --> Resume[Reopen the last screen they were on]
+    Active -- No --> Type{What has the user set up?}
+    Type -- Nothing yet --> Intro[Intro screen]
+    Type -- CommCare apps only --> Apps[CommCare Apps list]
```
Jignesh-dimagi: When you mention the 'CommCare Apps list', do you mean a new screen entirely, or are you referring to the list (with dropdown) on the current login page?

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+**How the Back button behaves**
+
+```mermaid
+flowchart TD
+  OppHome[Opportunity Home] -->|Back| OppList[Opportunity List]
+  AppHome[CommCare app Home] -->|Back| AppsList[CommCare Apps list]
+  OppList -->|Back| ExitA([Exit app])
+  AppsList -->|Back| ExitB([Exit app])
```
Jignesh-dimagi: When the user presses the back button (Opportunity Home or CommCare app Home), should we log them out of the CommCare app? Also, what should happen when they reopen the app afterward?

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,112 @@
...
+
+- `DispatchActivity` (the app's launcher) re-runs its routing every time it returns to the foreground, so backing into it re-dispatches and loops.
+- It only ever decides *login vs. CommCare home* — there is no first-class path to a Connect home.
+- Navigation is spread across four activities (`DispatchActivity` / `LoginActivity` / `ConnectActivity` / `StandardHomeActivity`), and Back is patched per-case with activity flags. That combination is the direct source of the loops, stale screens, stack growth, and inconsistent exits the team already hit while building the Connect launch flow (#3765).
+
+## The approach
+
+Two changes deliver the target behavior.
+
+**1. A one-time startup router.** A single resolver runs *once* at launch, decides the landing screen, and hands off — taking over the routing role `DispatchActivity` plays today. Running once (instead of on every foreground) is what removes the re-dispatch loop, and it adds the Connect-home branch the current tree lacks. Its outcomes are the "Where the app opens" diagram.
```
Jignesh-dimagi, line 73: Quick question regarding the routing architecture:

Will this be implemented as a dedicated RoutingActivity, or is it meant to be a stateless helper class/router?

Is this router going to be the LAUNCHER entry point triggered every time the app opens, or is the destination/home activity preference stored separately (e.g., in SharedPreferences) to bypass routing on subsequent launches?

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,112 @@
...
+
+**2. A two-tier structure with stack-based Back.** The identity / Connect / list screens are consolidated into one navigation surface (the "shell"); the CommCare app runtime stays a separately launched screen. Back is then driven by the real screen stack — the user's home screen is the task root (Back there exits) and every other screen sits above it — so the per-case flags disappear. Connect opportunity screens become app-capable *in place* (via #3776), so opening an opportunity doesn't launch a separate home screen and the stack can't grow. This produces the "How the Back button behaves" diagram.
+
+## Why this covers the tricky cases
+
+The stack-based model resolves the edge cases from [#3765's edge-cases doc](https://docs.google.com/document/d/1jiVEbljnR8abPwJnKzqULTEqbAxU_PAt9sRWkPjTB9k/edit?usp=sharing) without special-casing:
+
+- Backing out of a list exits because the list *is* the task root — the same rule whether or not an app was launched (no `appLaunchedFromConnect`-style flag).
+- Switching apps, or re-launching a running app, reuses the existing home rather than leaving a stale one to Back into.
```
Jignesh-dimagi, line 82: When a user switches the active CommCare app via the drawer menu, will it update and save that app as the last selected Connect Opportunity? Also, will this change immediately reflect on the Opportunity home screen?

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
conroy-ricketts: A new screen entirely. It's this page in the redesign:

image

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
conroy-ricketts: No, Back/Exit closes the UI but does not log the user out - the session stays alive. On reopen, the active session resumes them straight to their Home. Logout is a separate explicit action

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
conroy-ricketts: A stateless resolver/helper, not a dedicated activity.  It runs once per cold start, reading the persisted values (e.g. the last session context and the last accessed opportunity). It does not cache a destination to skip routing.

In the interim, `DispatchAcitivty` will still exist and invoke the new resolver/helper.

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
conroy-ricketts: Switching an app via the drawer menu will update the "last-used app track", not the last accessed opportunity. The opportunity home will not be changed. So, changing the last accessed opportunity, rather, would mean opening one from the Opportunities List page

conroy-ricketts:
Hi @dimagi/connect-mobile-devs @avazirna @ctsims,

Bumping this for review - it has been updated to align with the design spec in [the Google doc](https://docs.google.com/document/d/1rntc16FW2Jfr6CbzNcZ9RDOxIRrmfcWNU6_d5WL1rA8/edit?tab=t.ongpgraabwyz)

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,151 @@
...
+
+```mermaid
+flowchart TD
+    AppHome[App Home] -->|Back| Login[Login Page]
+    Login -->|Back| Exit([Exit app])
+```
+
+## Why today's navigation can't deliver this
+
+- `DispatchActivity` (the app's launcher) re-runs its routing every time it returns to the foreground, so backing into it re-dispatches and loops.
```
shubham1g5: This seems a bit misleading to me as if that's a problem today, I suppose exiting from the app would have been impossible.

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,151 @@
...
+- It only ever decides *login vs. CommCare home* — there is no first-class path to a Connect home.
+- Navigation is spread across four activities (`DispatchActivity` / `LoginActivity` / `ConnectActivity` / `StandardHomeActivity`), and Back is patched per-case with activity flags. That combination is the direct source of the loops, stale screens, stack growth, and inconsistent exits the team already hit while building the Connect launch flow (#3765).
+
+## The approach
+
+Two changes deliver the target behavior.
+
+**1. A one-time startup router.** A single resolver runs *once* at launch, decides the landing screen, and hands off — taking over the routing role `DispatchActivity` plays today. Running once (instead of on every foreground) is what removes the re-dispatch loop, and it adds the Connect-home branch the current tree lacks. Its outcomes are the "Where the app opens" diagram.
```
shubham1g5: I think `DispatchActivity` is definitely meant to play the role specified here and think It would be good to avoid having 2 classes just for start up routing. Curious if there is a reason we think `DispatchActivity` can't be modified to suit our purposes or what are the challenges in doing so ?

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
conroy-ricketts: Agreed, will fix the wording here

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
conroy-ricketts: Think I should update the wording here. The idea is that we want to modify `DispatchActivity`, not replace it. I'll tweak this wording very soon - let me know if that helps resolve the question

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Back stack: north-star vs. interim
+
+- **North-star:** a single `NavHost` shell; **Back** is pure temporal (start-destination-exit), **Up** uses the nav graph's hierarchy (with `NavDeepLinkBuilder`-synthesized parents for deep links); no flags.
+- **Interim (Solution A on today's activities):** `ConnectActivity` stays the for-result parent (edge-cases doc's Option D); keep `appLaunchedFromConnect` / `finishAffinity` / `REORDER_TO_FRONT` until the shell lands. `DispatchActivity` stays the **single router** — modified to decide once per cold start and to add the Connect-home branch — and continues to own DB-bad-state, recovery, and external/session-endpoint launches; no second routing class is introduced.
+
+### Back vs. Up (implementation)
+
+- **Back** = temporal: the natural activity/fragment back stack, exiting from the task root; no flags.
+- **Up** = hierarchical: a top-bar arrow wired via `NavigationUI.setupActionBarWithNavController` (already used by the newer PersonalID/Connect screens), with `NavDeepLinkBuilder`-synthesized parents for deep links.
```
Jignesh-dimagi: Up uses the nav graph's hierarchy with NavDeepLinkBuilder - synthesized parents, but the root is chosen at runtime: E.g. ConnectActivity.java calls [graph.setStartDestination(getStartDestinationId(startArgs))](https://github.com/dimagi/commcare-android/blob/8ddc9359cda0cf5b76ccad9d8771d6483538b6a8/app/src/org/commcare/activities/connect/ConnectActivity.java#L94), and nav_graph_connect.xml declares no startDestination at all. 

A Connect user with several opportunities taps a payment notification for opportunity X. What does Up from that payment screen synthesize down to? 

NavDeepLinkBuilder inflates its own graph from XML, so it can't see graph.setStartDestination(...) — and nav_graph_connect.xml declares no startDestination for it to fall back on. Does this mean declaring deep links plus a static start destination in the graph, and if so does the router still get to choose the root?

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,164 @@
...
+
+- **Back** is *temporal*: it retraces the exact screens the user moved through and exits the app from the screen the session opened to. No hidden flags.
+- **Up** is *hierarchical*: a top-bar arrow that moves up the app's structure to the area's home; it never exits. For a directly-launched or deep-linked screen, Up follows a synthesized parent path.
+
+Where they differ: tabs (Back → the previous tab; Up → the page's Home), side-menu sections (Back → the previous section; Up → home), and a directly-opened screen (Back → exit; Up → its parent, e.g. Opportunity Home → Opportunity List).
+
+```mermaid
+flowchart LR
+    OH[Opportunity Home opened at launch] -->|Back| EX([Exit app])
+    OH -->|Up| OL[Opportunity List]
```
Jignesh-dimagi: Opportunity Home --Up--> Opportunity List, but Opportunity Home should carry a drawer menu button, not an Up arrow ?

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Why this covers the tricky cases
+
+The stack-based model resolves the edge cases from [#3765's edge-cases doc](https://docs.google.com/document/d/1jiVEbljnR8abPwJnKzqULTEqbAxU_PAt9sRWkPjTB9k/edit?usp=sharing) without special-casing:
+
+- Backing out of the launch screen exits, because that screen *is* the task root — the same rule whether the user launched there or navigated there (no `appLaunchedFromConnect`-style flag).
+- Switching apps, or re-launching a running app, reuses the existing home rather than leaving a stale one to Back into.
+- A notification deep-link opens directly to its target: **Back** exits (it's the task root), while **Up** follows a synthesized parent path (Android's deep-link guidance) so the user can still move up into the app.
```
Jignesh-dimagi: When opening ConnectMessageFragment directly from a push notification, pressing Up takes the user back to the channel list. Should the channel list display an Up arrow or a side drawer menu icon in this notification flow? Additionally, should the back stack propagate all the way up to the Opportunity Home screen?

conroy-ricketts:
Converting this to a draft until the Google doc is finalized and approved

conroy-ricketts:
Hey @Jignesh-dimagi, both the design doc and the tech spec have went through significant revision since you last reviewed, so I'm resolving your comments for now. Please let me know if I missed anything!

conroy-ricketts:
Hi @shubham1g5 @Jignesh-dimagi @OrangeAndGreen,

This is ready for a fresh look!

I'm also hoping that y'all can let me know what you think about the AI verbosity on this (i.e. is it too much, just right, or too little?) 

Note that only the first half of the spec (before the implementation notes) is necessary for review at this moment.

Thanks!

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,180 @@
...
+The unlock gate belongs to the router, so no entry point can bypass it — `ConnectUnlockFragment` stops being the startup host. Entries from inside the app, after the session has already expired, keep today's per-call helpers; north-star collapses both into the one gate.
+
+The fallback screen has to explain why the user is on it. For a PersonalID-assisted user the recovery action there is the PersonalID option, not username/password.
+
+## The seeded back path
+
+| Entry | Task stack, bottom → top |
+|---|---|
+| Cold start, Connect user with a current opportunity | Opp List → Opp Home |
+| Cold start, Connect user with no current opportunity | Opp List |
```
shubham1g5: Confirming that this doesn't mean user would go back to Opp List when clicking back from Opp Home given user never sees the screen in both these workflows, in 1. User move directly to Opp Home and in 2. to CC login/Intro page.

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,180 @@
...
+| Cold start, traditional CommCare or PersonalID user | Login page (CommCare Apps list, once it ships) → app Home |
```
shubham1g5: Pressing back from App Home should exit as that's the first page user sees and user should only go back to login page when there is no active user session .

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,180 @@
...
+- **Why the user was on the Opportunity List** — passing through it on the way out vs. going there deliberately to switch opportunities.
+
+## Interim vs. north-star
+
+- **North-star:** a single `NavHost` shell; the seeded path is the nav back stack.
+- **Interim:** the seeded path ships now as a **task stack** on today's activities. This *replaces* rather than coexists with `appLaunchedFromConnect` / `finishAffinity`. `REORDER_TO_FRONT` stays for reusing a running home. `DispatchActivity` stays the single router and continues to own the corrupted-database and recovery paths, plus launches that come from outside the app.
+
+## Feature flag
+
+A new flag in [`PersonalIdFeatureFlagChecker`](https://github.com/dimagi/commcare-android/blob/dc7697645fefd99de4e234be569bd8447fb6e0ba/app/src/org/commcare/personalId/PersonalIdFeatureFlagChecker.kt), merged to `master` switched off, and turned on with the redesign. While it is off, today's routing and today's sidebar rules both stay in force and no back path is built.
```
shubham1g5: Think we should not need a feature flag on this given there is no new functionality here, just the app navigation which would be really difficult to  turn on/off behind a feature flag

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,180 @@
...
+
+Cost to be aware of: each seeded activity is really created, so a Connect cold start instantiates Opportunity List underneath Opportunity Home. Keep the list's `onCreate` cheap (defer network and heavy binding to `onStart`/`onResume`, which won't run while it is beneath another activity). This cost is interim-only — in the shell, seeding is just pushing nav back-stack entries, with no activity creation.
+
+**Sidebar sections replacing each other.** Launching a section from the drawer finishes the section activity currently on top (if any) before starting the new one, rather than relying on intent flags — `FLAG_ACTIVITY_CLEAR_TOP` clears *ancestors*, not siblings, so it will not pop Messaging when launching Work History. In the shell this becomes `popUpTo(<workspace destination>)` followed by `navigate`, and the depth bound falls out of the graph rather than needing enforcement.
```
shubham1g5, line 110: Trying to imagine how it would work, suppose if we have a navigation - 

Opp Home -> More tab -> Click on OCS TASK CTA -> Messaging List -> Messaging chat 

Now if user press back, think user should move back to Opp Home instead of exiting from Messaging List. Thoughts ?

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,180 @@
...
+
+### Slot enforcement
+
+Interim: `BaseDrawerActivity` gains a declared slot per screen (sidebar / back / none) and configures the toggle and `setDisplayHomeAsUpEnabled` from it, replacing the per-activity `onOptionsItemSelected` handling. North-star: `AppBarConfiguration(topLevelDestinationIds)` with `NavigationUI.setupActionBarWithNavController`, listing the sidebar screens as top-level.
+
+### Unlock mechanics
+
+[`PersonalIdUnlocker`](https://github.com/dimagi/commcare-android/blob/dc7697645fefd99de4e234be569bd8447fb6e0ba/app/src/org/commcare/personalId/PersonalIdUnlocker.kt): cancel and failure both resolve through `connectActivityComplete(false)`. `unlock` takes an `UnlockPolicy`; `SESSION_WITH_TIME_THRESHOLD` skips the prompt when unlocked within the last 10 minutes. `lastUnlockTime` is in-memory, so it does not survive process death — a background kill means a fresh prompt.
+
+Helpers that stay in place for sessions that have already expired: `ConnectNavHelper.unlockAndGoToConnectJobsList` / `unlockAndGoToMessaging` / `unlockAndGoToWorkHistory`, called from `BaseDrawerActivity`.
```
shubham1g5, line 120: This is how it is today with no change right ?

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,180 @@
...
+
+At cold start, show a splash / branded base behind the prompt — Android 12+ `SplashScreen` kept on screen via `setKeepOnScreenCondition`, or a splash-themed `windowBackground`, with `androidx.core:core-splashscreen` for pre-12 — so the biometric/PIN dialog isn't over a blank window.
```
shubham1g5, line 122: very excited to see CommCare open with a Splash screen :) 

> so the biometric/PIN dialog isn't over a blank window.

Does this mean we will ask user to unlock even before they get into a Connect/PersonalID screen, that seems right to me and just confirming that's the plan.

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,180 @@
...
+
+### Sidebar code changes
+
+- `ConnectActivity` and `ConnectMessagingActivity` extend `NavigationHostCommCareActivity`; `PersonalIdWorkHistoryActivity` extends `CommCareActivity`. None extend `BaseDrawerActivity`, which is why those screens have no drawer. Move all three onto `BaseDrawerActivity` and override `shouldShowDrawer()` to return `true`, since the interim ships on today's activities. The shell absorbs them later, at which point the drawer comes from the shell activity instead.
+- Retire `NavDrawerHelper.drawerShownBefore()` / `setDrawerShown()` and the `shouldShowDrawerAfterCheck(requirePersonalIDLogin)` gate. Today `CommCareSetupActivity:315` passes `false` and sets the stored flag, and that stored flag then causes the checks in `LoginActivity:1069` and `StandardHomeActivity:354` to return early without ever testing for PersonalID. `shouldShowDrawer()` defaults to `false`, so only those three activities opt in at present.
```
shubham1g5: `drawerShownBefore` is required since the logic to show drawer is 1. A personalID user OR 2. the drawer shown before on same app sandbox

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,180 @@
...
+- Keep `checkDeviceCompability()` (`SDK_INT >= P`).
+- `RootMenuHomeActivity` still uses the legacy `HomeNavDrawerController`, pinned `LOCK_MODE_LOCKED_CLOSED`. Replace it rather than leaving two drawer implementations in the tree.
```
shubham1g5: we plan to remove it as part of phase 2 work and don't need to worry about it.

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,180 @@
...
+- Drawer contents come from `BaseDrawerController.refreshDrawerContent`; the signed-out branch (`setSignedInState(false)` → `configureErrorState()`) renders the sign-up prompt, whose button calls `PersonalIdManager.launchPersonalId(activity, PERSONAL_ID_SIGN_UP_LAUNCH)`.
+
+### Nothing to remove for `Up`
+
+`NavUtils` / `navigateUpFromSameTask` are unused across the app. `CommCareActivity.onOptionsItemSelected:261` routes `android.R.id.home` to `onBackPressed()`. The single `android.support.PARENT_ACTIVITY` entry in the manifest (`ReportProblemActivity`) is inert because nothing calls `navigateUp`. `FormEntryActivity:651` raises the quit prompt from the arrow — leave as-is.
+
+### Tabs
+
+`ConnectDeliveryProgressFragment` hosts the tabs as a `TabLayout` + `ViewPager2`. Pass the wanted tab as a destination argument / intent extra and apply it with `setCurrentItem(index, false)` before the first frame; default to Dashboard when absent.
```
shubham1g5: this might be a little outdated now, `ConnectDeliveryProgressFragment` is not the host but it's called `ConnectDeliveryHomeFragment` now and `ConnectDeliveryProgressFragment` is destined to be cleaned up soon.

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,180 @@
...
+
+### Persistence
+
+A dedicated shared-preferences store, **scoped to the active PersonalID account** (namespaced per account, cleared on `forgetUser()` / account switch, never inherited): last-accessed opportunity (`jobUUID`), last session context (`manual` / `PersonalId-non-opportunity` / `PersonalId-on-opportunity-X`), and a per-opportunity terminal-state acknowledgment flag (drives reopen-once-then-fall-back-to-list for ended opportunities).
```
shubham1g5: think this already exists today as `PersonalIdUserPreferences`

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
conroy-ricketts: _**Note: longer reply, but I think the reasoning is important**_

Yeah, users will see the Opp List when they tap back from Opp Home.

One note on the row you've commented on: that's the "no current opportunity" case, where Opp List *is* the landing screen and back exits in one press. The extra step is the row above, where they have a current opportunity and land straight into it.

The reason the list stays underneath is that it's also a sidebar destination. Open the sidebar → Opportunities → tap the wrong opportunity → back puts you on the list so you can pick again. If Opp Home were the bottom, that same back press exits the app.

And I don't think we can have both: either back from Opp Home always exits (so a mis-tap drops you out), or it sometimes returns to the list (so back depends on how you got there, which is the state-dependent behaviour we're trying to remove). The trade-off is one extra press on the way out versus an accidental exit, which often means unlocking PersonalID again.

This came up with Dave on the design doc and we went a few rounds on it. He landed on trying it this way. We also committed to measuring how often people pass through the list on the way out versus going there deliberately to switch, so we can revisit with data if it turns out to be friction.

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
conroy-ricketts: The reason it's in the path is because the Login page has the app picker dropdown. That's what the design doc landed on in the near-term section ("back behaves against that Login Page as the home").

That said, I'm torn. Today, Back from App Home exits, so as written this changes Back behaviour for traditional CommCare users, and I'd rather not change their experience if we can avoid it.

Once the Login page becomes the CC Apps list, is "Back from App Home exits" still what we want? 

If Back → CC Apps list is right at that point, the choice is between exiting now and changing later, or being consistent from the start.

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
conroy-ricketts: That sounds right to me

conroy-ricketts: https://github.com/dimagi/commcare-android/pull/3844/commits/0008ebd25

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
conroy-ricketts: Correct, the idea is that the user would go back to Opp Home in that case

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
conroy-ricketts: Yup!

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
conroy-ricketts: Yes, correct!

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
conroy-ricketts: This reverts a decision made in the design doc. Just confirming that we no longer want to show the sidebar on every home screen for all users?

The original reason for this was that the sidebar is the only place in the app an established traditional CommCare  user can discover PersonalID in the interim until the new CommCare App List page is implemented. Gating it on already being a PersonalID user makes that circular. 

Please let me know if I'm missing something here

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
conroy-ricketts: Good catch, thanks!

conroy-ricketts: https://github.com/dimagi/commcare-android/pull/3844/commits/9f8cf77f2

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
conroy-ricketts: Thanks! Will update

conroy-ricketts: https://github.com/dimagi/commcare-android/pull/3844/commits/5dc5ae525

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
conroy-ricketts: Great point, this should explicitly state we are adding to it, not creating a new object.

conroy-ricketts: https://github.com/dimagi/commcare-android/pull/3844/commits/f8706c942

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
shubham1g5: > This reverts a decision made in the design doc. Just confirming that we no longer want to show the sidebar on every home screen for all users?

I didn't realise we talked about this on desigc doc so think there is some confusion somewhere but I am unable to relate the 2 decisions here. 

IMO, we want to show the side bar on all the home pages only if user is signed in as a Personal ID account or if the drawer has been shown before on that app sandox. We strictly do not want to show drawer for non-personal ID users on all pages. 


> the sidebar is the only place in the app an established traditional CommCare user can discover PersonalID in the interim until the new CommCare App List page is implemented.

We still have the "Sign up" option on the 3 dot menu on Login Page.

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
shubham1g5: >  that's the "no current opportunity" case, where Opp List is the landing screen and back exits in one press.

In this case, Opp List page remains hidden and user would never land on Opp List but the CC Login or App Home as per the starting router logic. 

> The reason the list stays underneath is that it's also a sidebar destination. Open the sidebar → Opportunities → tap the wrong opportunity → back puts you on the list so you can pick again. 

Opp home is a sidebar destination as well so having user go back to Opp List is not justified here given user never came to this page during forward nav. 

>  If Opp Home were the bottom, that same back press exits the app.

Think that's exactly what we want. Do you have a different read from the design doc on this ? Think we should go back to design doc on this if so. 

A primary goal of this work is to remove the Opp List page from user workflows given almost all our users don't have more than a single opprotunity today and it feels very weird to me if user sees a page they don't see in their natural app navigation suddently coming on pressing back and is a major bad smell in this design.

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
shubham1g5: > Once the Login page becomes the CC Apps list, is "Back from App Home exits" still what we want?

Yes. I see the Login and App Home on the same heirarchy with different states (signed in and signed out). Pressing back to login seems like pressing back to logout which will cause session disruptions and defeats the purpose of us keeping sessions alive deliberately for 24 hours since the user login.

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
conroy-ricketts: @shubham1g5 

Ah, that's the confusion I'm having. I didn't realize we wanted to remove the Opp List page, nor that the Opp Home would be the sidebar destination and not the Opp List. I think that changes the proposal here significantly. 

Yeah, we may want to go back to the design doc on this one - it assumed the Opp List was still a major piece of the workflow, and the app nav design spec does not describe this behavior.

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
conroy-ricketts: That makes sense to me. I'll revisit this. This also conflicts with the design spec and we may want to revisit/update that too

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
conroy-ricketts: That sounds good to me, I'll come back to this! I almost forgot about that 3 dot menu

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
conroy-ricketts: @shubham1g5 @OrangeAndGreen @Jignesh-dimagi 

Before I make further edits on these specs, I need to confirm some details:

If Opp Home will become a sidebar destination, how will the app sidebar accommodate it? For example, will it be a new dedicated button?

Relatedly, are we still keeping the "Opportunities" button?

shubham1g5: That's a good question and seems like a gap in redesign. Think it would be great to ask this to product/Ishwari on the design doc. My sense is it's alright to move user to the Opportunity List page on clicking Opportunities button in side drawer.

shubham1g5: >  I didn't realize we wanted to remove the Opp List page,

Also to be clear, the page is still there but won't part of the launch navigation for user as in most cases user would land on Opp Home page directly as per the [app startup routing](https://docs.google.com/document/d/1rntc16FW2Jfr6CbzNcZ9RDOxIRrmfcWNU6_d5WL1rA8/edit?tab=t.ongpgraabwyz#heading=h.7vv9z3mwchqs) logic.

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
shubham1g5: > This also conflicts with the design spec

Aplogies if I am oversee-ing something here but I am struggling to see where this information is surfaced on the design doc.

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,160 @@
...
+**Which home an active session resumes to** depends on how that session was created. A session signed in through PersonalID, on an app tied to an opportunity, resumes Opportunity Home; a session signed in with a username and password, or on an app with no opportunity, resumes the CommCare app home.
+
+The unlock gate belongs to the router, so no entry point can bypass it — `ConnectUnlockFragment` stops being the startup host. Entries from inside the app, after the session has already expired, keep today's per-call helpers; north-star collapses both into the one gate.
+
+## The back path
+
+| Entry | Task stack, bottom → top                                 |
+|---|----------------------------------------------------------|
+| Cold start, Connect user with a current opportunity | Opp Home                                                 |
+| Cold start, Connect user with no current opportunity | Opp List                                                 |
```
shubham1g5: A "Connect user with no current opportunity" is essentially not a connect user but a CommCare user and Opp List is not a valid navigation option for such users.

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
```diff
@@ -0,0 +1,160 @@
...
+| Cold start, traditional CommCare or PersonalID user | App Home (Login page when there is no session to resume) |
+| Notification into a chat | Opp Home → Messaging channel list → Chat                 |
+| Sidebar section opened from inside an opportunity | Opp Home → Section                                       |
+| A second sidebar section opened after the first | Opp Home → New section (the first section is closed)     |
```
shubham1g5, line 47: +1, this seems like a really good approach to maintain Home heirarchy!

shubham1g5 (dismissed):

docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md
conroy-ricketts: Apologies for the ambiguity in the wording here, this case refers to a user who does have opportunities, but has yet to open one

shubham1g5: makes sense, Think the specific case then would be , "User has not opened any opps and have more than one opp"

conroy-ricketts: Yup yup! I'll tweak this

conroy-ricketts: https://github.com/dimagi/commcare-android/pull/3844/commits/a2e4e4727

conroy-ricketts:
**[AI - Claude Opus 5]**

Thanks everyone for the reviews. I'm closing this PR and moving to implementation. The work has been split into six tickets under [CCCT-2688](https://dimagi.atlassian.net/browse/CCCT-2688), each linking back to both specs:

- [CCCT-2823](https://dimagi.atlassian.net/browse/CCCT-2823) — Route each user to the right screen when the app opens
- [CCCT-2824](https://dimagi.atlassian.net/browse/CCCT-2824) — Make back exit the app from the screen it opened on
- [CCCT-2825](https://dimagi.atlassian.net/browse/CCCT-2825) — Add the sidebar to the PersonalID screens
- [CCCT-2826](https://dimagi.atlassian.net/browse/CCCT-2826) — Give notifications a back path into the app
- [CCCT-2827](https://dimagi.atlassian.net/browse/CCCT-2827) — Log which back control the user pressed
- [CCCT-2830](https://dimagi.atlassian.net/browse/CCCT-2830) — Add the sidebar to the Connect screens

The final tech spec lives at [this permalink](https://github.com/dimagi/commcare-android/blob/a2e4e4727/docs/superpowers/specs/2026-07-23-app-navigation-changes-design.md).

[CCCT-2688]: https://dimagi.atlassian.net/browse/CCCT-2688?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CCCT-2823]: https://dimagi.atlassian.net/browse/CCCT-2823?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CCCT-2824]: https://dimagi.atlassian.net/browse/CCCT-2824?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CCCT-2825]: https://dimagi.atlassian.net/browse/CCCT-2825?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CCCT-2826]: https://dimagi.atlassian.net/browse/CCCT-2826?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CCCT-2827]: https://dimagi.atlassian.net/browse/CCCT-2827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


[CCCT-2830]: https://dimagi.atlassian.net/browse/CCCT-2830?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3845, opened 2026-08-03
CCCT-2547 Redesign Learn Complete Screen And Claim Opportunity Inline

Jignesh-dimagi:
### [CCCT-2547](https://dimagi.atlassian.net/browse/CCCT-2547)

Screenshot_20260803_125953

Screenshot_20260805_185453

## Product Description

The learning-complete state of the Connect learn screen now follows the new design: a Learn summary card with a collapsible certificate, the delivery figures inline, and a pinned footer whose button claims the opportunity directly.

## Technical Summary

The fragment's root became a `FrameLayout` holding two mutually exclusive children so the existing in-progress markup stays untouched while the new screen owns its own scroll container and pinned footer. Claim logic moved into `ConnectJobClaimController`; `ConnectInfoCard` is shared, so its spacing change intentionally applies to the opportunity intro screen as well, keeping the two screens identical.

### Safety story

What gives confidence:

- I ran the screen on a device/emulator and checked the certificate, delivery cards and footer against the design.
- The change is confined to one state of one screen; the in-progress state's markup and IDs are unchanged.
- 19 new Robolectric tests cover the new view and the fragment's state selection and claim cycle.

### Automated test coverage

`ConnectLearnCompleteViewTest` and `ConnectLearningProgressFragmentTest` (19 tests) lock down which state renders before any repository emission — the fix for the wrong state flashing on load — and drive the footer button through real clicks to the claim call, the local status write read back from the database, and navigation, including the already-claimed and claim-failure paths.

[CCCT-2547]: https://dimagi.atlassian.net/browse/CCCT-2547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Jignesh-dimagi:
## Suggested Review Order

- `app/src/org/commcare/fragments/connect/ConnectDeliveryDetailsFragment.java` + `app/res/layout/fragment_connect_delivery_details.xml` — the deleted screen, for what is being replaced
- `app/src/org/commcare/connect/ConnectJobClaimController.kt` — claim logic extracted out of it
- `app/src/org/commcare/views/connect/ConnectLearnCompleteView.kt` + `view_connect_learn_complete.xml` + `view_connect_learn_certificate.xml` — the new screen
- `app/src/org/commcare/fragments/connect/ConnectLearningProgressFragment.java` + `fragment_connect_learning_progress.xml` — state switch and CTA wiring
- `app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java` + `app/res/navigation/nav_graph_connect.xml` — the rerouted entry point
- `app/src/org/commcare/views/connect/ConnectInfoCard.kt` + `view_connect_info_card.xml` + `ConnectCtaBar.kt` — shared component changes, these also affect the opportunity intro screen
- `app/unit-tests/…` — the three test files
- `app/res/values*/strings.xml`, `dimens.xml` — mechanical, read last

app/src/org/commcare/connect/ConnectJobClaimController.kt
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    private fun proceed(
+        job: ConnectJobRecord,
+        deliveryAppInstalled: Boolean,
+        onClaimed: OnClaimedListener,
+    ) {
+        job.status = ConnectJobRecord.STATUS_DELIVERING
+        ConnectJobUtils.upsertJob(job)
+        CommCareApplication.instance().closeUserSession()
```
OrangeAndGreen: I don't think this is necessary here anymore. Claude agrees, stating that relevant teardown happens correctly in other places now. Worth testing to be sure that's really the case though

OrangeAndGreen (commented):
One thing to consider and also flagging some notes from CodeRabbit that seem worth addressing, but otherwise looks good

app/src/org/commcare/views/connect/ConnectInfoCard.kt
```diff
@@ -72,6 +77,24 @@ class ConnectInfoCard
...
+                verticalBias = if (hasSubtitle) 0f else CENTERED_BIAS
+            }
+        }
+
+        private fun subtitleReservedHeight(): Int {
+            val subtitle = binding.infoCardSubtitleText
+            val unspecified = MeasureSpec.makeMeasureSpec(0, MeasureSpec.UNSPECIFIED)
+            subtitle.measure(unspecified, unspecified)
+            return subtitle.measuredHeight + (subtitle.layoutParams as MarginLayoutParams).topMargin
+        }
```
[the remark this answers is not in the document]
Jignesh-dimagi: Nothing is missing. MarginLayoutParams is a nested class of ViewGroup, and ConnectInfoCard extends CardView → FrameLayout → ViewGroup, so Kotlin resolves it by simple name through the supertype.

app/src/org/commcare/connect/ConnectJobClaimController.kt
Jignesh-dimagi: I kept this here mainly for two reasons:

- ConnectDeliveryDetailsFragment.java (which has now been deleted) contained this exact teardown logic. Since the scope of this ticket is strictly UI changes, I avoided altering the underlying behavior.
 - The Claude recommended keeping it to avoid unintended side effects.
 
 Open to suggestions here.

app/res/values-lt/strings.xml
```diff
@@ -363,6 +363,9 @@
         <item quantity="many">%d mokėjimo vieneto</item>
         <item quantity="other">%d mokėjimo vienetų</item>
     </plurals>
+    <string name="connect_learn_cert_score">Rezultatas: %s</string>
+    <string name="connect_learn_complete_footer_start_visits">Pradėti vizitus</string>
+    <string name="connect_learn_claim_failed">Atsisiuntimas nepavyko. Bandykite dar kartą.</string>
```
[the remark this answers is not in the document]
Jignesh-dimagi: Added translation for Lithuanian and Norwegian [here](https://github.com/dimagi/commcare-android/pull/3845/changes/3a9c0604f369cb374240a6ac99c66b9df62742e7)

app/res/values-es/strings.xml
```diff
@@ -256,25 +256,25 @@
...
     <string name="connect_sms_invite_opportunity_not_found">Oportunidad no encontrada</string>
     <string name="connect_learn_need_assessment">Completa la evaluación para terminar la formación.</string>
     <string name="connect_learn_failed">Lo sentimos, no obtuviste la calificación aprobatoria en tu evaluación. Inténtalo de nuevo.\nTu calificación: %d\nCalificación aprobatoria: %d</string>
     <string name="connect_learn_complete">Completar</string>
     <string name="connect_learn_certificate">Certificado</string>
-    <string name="connect_learn_completed">Completado el: %s</string>
+    <string name="connect_learn_completed">Completado el %s</string>
+    <string name="connect_learn_cert_score">Puntuación: %s</string>
+    <string name="connect_learn_complete_footer_start_visits">Comenzar visitas</string>
+    <string name="connect_learn_claim_failed">La descarga falló. Inténtalo de nuevo.</string>
```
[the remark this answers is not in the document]
Jignesh-dimagi: Updated the string key name to connect_learn_download_failed. Additionally, whenever the app fails to claim, it is treated as a download failure from the user's perspective.

app/src/org/commcare/connect/ConnectJobClaimController.kt
```diff
@@ -0,0 +1,82 @@
...
+        onClaimed.onClaimed(deliveryAppInstalled)
+    }
+
+    private fun errorMessage(
+        errorCode: PersonalIdOrConnectApiErrorCodes,
+        t: Throwable?,
+    ): String =
+        if (errorCode == PersonalIdOrConnectApiErrorCodes.BAD_REQUEST_ERROR) {
+            context.getString(R.string.recovery_unable_to_claim_opportunity)
```
shubham1g5, line 76: - think we should show this for all error codes, except network error where we should trim the `please contact an administrator` 

- We should also replace the  `please contact an administrator`  with ` please contact support` as that's more inline with other user messages. 

- can we have a screenshot on PR showin this error on UI. I am not clear where is this message getting displayed.

app/src/org/commcare/views/connect/ConnectInfoCard.kt
```diff
@@ -73,9 +79,31 @@ class ConnectInfoCard
             }
         }
 
+        private fun applySubtitleLayout(hasSubtitle: Boolean) {
+            binding.infoCardSubtitleText.visibility = if (hasSubtitle) VISIBLE else GONE
```
shubham1g5: think this is already happening in `bindOptional`

app/unit-tests/src/org/commcare/fragments/connect/ConnectLearningProgressFragmentTest.kt
```diff
@@ -0,0 +1,279 @@
...
+
+    /**
+     * Captures the claim callback without invoking it, so success is delivered after the click
+     * returns as a real async response would be.
+     */
+    private fun captureClaimCallback(): CapturingSlot<IApiCallback> {
+        val slot = slot<IApiCallback>()
+        mockkStatic(ApiConnect::class)
+        every { ApiConnect.claimJob(any(), any(), any(), capture(slot)) } returns Unit
+        return slot
```
shubham1g5: this should instead just mock the neworking layer using `mockServer`  (we also soon plan to get rid of `ApiConnect` as part of networking refactor)

app/unit-tests/src/org/commcare/fragments/connect/ConnectLearningProgressFragmentTest.kt
```diff
@@ -0,0 +1,279 @@
...
+
+        val failureCard =
+            fragment.requireView().findViewById<ConnectSuccessFailureCard>(
+                R.id.learn_complete_failure_card,
+            )
+        assertEquals(View.VISIBLE, failureCard.visibility)
+        assertEquals(
+            activity.getString(R.string.connect_learn_download_failed),
+            failureCard.messageText.toString(),
+        )
```
shubham1g5, line 207: Does a failed claim ends up in "Learn Download failed" error ? Think we should just show an error on the lines of `claim failed` instead

RELEASES.md
```diff
@@ -2,6 +2,18 @@
 This file is meant as an easy way for us to collate notes and change logs across releases. 
 -->
 
+## CommCare 2.65
```
shubham1g5: please remove the changes here as we have discussed before.

app/src/org/commcare/views/connect/ConnectLearnCompleteView.kt
```diff
@@ -0,0 +1,179 @@
...
+                ConnectMoneyUtils.moneyStringWithSymbol(job.currency, job.totalBudget)
+            binding.cardMaxEarnings.subtitleText =
+                resources.getQuantityString(
+                    R.plurals.connect_opportunity_payment_units,
+                    job.paymentUnits.size,
+                    job.paymentUnits.size,
+                )
+        }
+
+        private fun bindCta(
```
shubham1g5: what CTA is getting referred to here ? I don't see one in screenshot in the learn completed section.

shubham1g5 (changes requested):

app/src/org/commcare/connect/ConnectJobClaimController.kt
shubham1g5: Voting for removing it as there doesn't seem to be an explicit reason to keep it here and given we are rewriting a lot of logic here which would need to go through a robust QA before we publish.

app/src/org/commcare/connect/ConnectJobClaimController.kt
Jignesh-dimagi: >think we should show this for all error codes, except network error where we should trim the please contact an administrator
We should also replace the please contact an administrator with  please contact support as that's more inline with other user messages.

Whenever a user taps the 'Download App' CTA and it fails, exposing low-level errors can be confusing. To keep the UX clear and consistent, I’ve generalized all failures during this step to display a simple 'Download failed' error state. Regardless of the underlying cause, any failure during this flow is presented as a download failure from the user's perspective.

> can we have a screenshot on PR showin this error on UI. I am not clear where is this message getting displayed.

I have already attached the error screenshot in the description of this ticket.

app/src/org/commcare/views/connect/ConnectLearnCompleteView.kt
Jignesh-dimagi: Binds the bottom bar CTA button with its corresponding title and subtitle strings.

app/src/org/commcare/connect/ConnectJobClaimController.kt
```diff
@@ -0,0 +1,82 @@
...
+class ConnectJobClaimController(
+    private val context: Context,
+) {
+    fun interface OnClaimedListener {
+        fun onClaimed(deliveryAppInstalled: Boolean)
+    }
+
+    fun interface OnErrorListener {
+        fun onError(message: String)
+    }
```
conroy-ricketts, line 29: Nit: we could use Kotlin to our benefit here. Instead of defining custom interfaces, Kotlin's builtin function types may be used. This reduces boilerplate code and improves readability when we call the functions later

For example:
```kotlin
fun claimIfNeededAndProceed(
    job: ConnectJobRecord,
    onClaimed: (deliveryAppInstalled: Boolean) -> Unit,
    onError: (message: String) -> Unit,
)
```

app/src/org/commcare/views/connect/ConnectLearnCompleteView.kt
conroy-ricketts: Nit: can rename this function to be more specific, for example `bindCTABottomBar()`, to make its purpose more clear

app/src/org/commcare/views/connect/ConnectLearnCompleteView.kt
```diff
@@ -0,0 +1,179 @@
...
+                        if (deliveryAppInstalled) {
+                            R.string.connect_delivery_go
+                        } else {
+                            R.string.connect_opportunity_footer_download_app
+                        },
+                    )
+                subtitleText =
+                    if (deliveryAppInstalled) {
+                        null
+                    } else {
```
conroy-ricketts: Nit: for readability can we use just one bigger if statement for the `deliveryAppInstalled` state check and set both texts at the same time

app/src/org/commcare/views/connect/ConnectLearnCompleteView.kt
```diff
@@ -0,0 +1,179 @@
...
+                setOnCtaClickListener(onCtaClick)
+            }
+        }
+
+        private fun toggleCertificate() {
+            val expand = binding.certificate.root.visibility != VISIBLE
+            binding.certificate.root.visibility = if (expand) VISIBLE else GONE
+            binding.certificateChevron.rotation = if (expand) EXPANDED_CHEVRON_ROTATION else 0f
+            binding.certificateContainer.updatePadding(
+                bottom = if (expand) certificateBottomPadding else 0,
```
conroy-ricketts: Nit: Similar comment here, we could use just one if statement to check for `expand`

app/src/org/commcare/fragments/connect/ConnectLearningProgressFragment.java
```diff
@@ -101,54 +103,70 @@ private void observeLearningProgress() {
...
-
-        if (!hideProgress) {
-            getBinding().connectLearningProgressBar.setProgress(learningProgressPercent);
-            getBinding().connectLearningProgressText.setText(
-                    String.format(Locale.getDefault(), "%d%%", learningProgressPercent)
+        boolean learningComplete = job.getLearningPercentComplete(false) >= 100;
+        boolean showLearningComplete = learningComplete && job.passedAssessment();
+
+        getBinding().progressContainer.setVisibility(showLearningComplete ? View.GONE : View.VISIBLE);
+        getBinding().learnCompleteView.setVisibility(showLearningComplete ? View.VISIBLE : View.GONE);
```
conroy-ricketts: Nit: Suggest moving these `setVisibility()` calls inside `if (showLearningComplete)` below

app/res/layout/fragment_connect_learning_progress.xml
```diff
@@ -1,7 +1,12 @@
 <layout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools">
+<FrameLayout
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
+
```
conroy-ricketts, line 7: Nit: the spacing here is odd - thought this was a random empty FrameLayout at first and future readers might have the same confusion. We can indent its children elements

app/src/org/commcare/views/connect/ConnectInfoCard.kt
```diff
@@ -73,9 +79,31 @@ class ConnectInfoCard
...
+
+            val reserved = if (hasSubtitle) 0 else subtitleReservedHeight()
+            binding.infoCardTextContainer.setPadding(0, reserved / 2, 0, reserved - reserved / 2)
+
+            binding.infoCardValueText.updateLayoutParams<ConstraintLayout.LayoutParams> {
+                verticalBias = if (hasSubtitle) 0f else CENTERED_BIAS
```
conroy-ricketts: Nit: can use a single if statement here too for `hasSubtitle`

conroy-ricketts (commented):
The new UI looks fantastic!

I had nothing additional to add besides one major comment here, and a few nits throughout the changes:

One thing I noticed throughout all of the new UI - it seems that we are using color resources directly. Can we derive the colors from styles/themes instead?

Think we should aim for all colors to be easily managed from our styles/themes improving long-term maintainability, and the redesign is a great opportunity for us to do this

app/src/org/commcare/connect/ConnectJobClaimController.kt
shubham1g5: > Regardless of the underlying cause, any failure during this flow is presented as a download failure from the user's perspective.

The approach to generalise error makes sense to me but do think that we need to separate out the "Download" failure from "Claim failure" here as we would like user to explicitly know that they are unable to claim opportunity here (happens frequently due to opportunity budgets maxed out) rather than an error downloading the application after claim.

shubham1g5: Also why try to evaluate an `errorMessage` here if we do not plan to use it, given we are generalizing all errors here as `'Download failed'

Jignesh-dimagi:
> One thing I noticed throughout all of the new UI - it seems that we are using color resources directly. Can we derive the colors from styles/themes instead?
> 
> Think we should aim for all colors to be easily managed from our styles/themes improving long-term maintainability, and the redesign is a great opportunity for us to do this

We definitely follow that pattern when building individual components, allowing them to be customized via styles. However, for full page layouts composed of multiple components with distinct placements, driving everything through styles becomes overly complex and hard to maintain or identify by name.

That said, I'm still open to suggestions if anyone has a cleaner approach in mind!

app/src/org/commcare/connect/ConnectJobClaimController.kt
Jignesh-dimagi: Removed [here](https://github.com/dimagi/commcare-android/pull/3845/changes/27db871f607a2fda82ca9d2f4733ab076a75e701). Confirmed the teardown is redundant.

app/src/org/commcare/connect/ConnectJobClaimController.kt
Jignesh-dimagi: Yeah, I agree. Failure card now renders what errorMessage() produces. Also, corrected the string to `Unable to claim the opportunity. Please try again or contact support.` Changed [here](https://github.com/dimagi/commcare-android/pull/3845/changes/27db871f607a2fda82ca9d2f4733ab076a75e701)

app/src/org/commcare/views/connect/ConnectInfoCard.kt
Jignesh-dimagi: Thanks, changed [here](https://github.com/dimagi/commcare-android/pull/3845/changes/27db871f607a2fda82ca9d2f4733ab076a75e701).

app/unit-tests/src/org/commcare/fragments/connect/ConnectLearningProgressFragmentTest.kt
Jignesh-dimagi: Done — ApiConnect and IApiCallback are out of this test entirely. Added ConnectMockApiServer, which points ConnectApiClient at a MockWebServer; PersonalIdMockApiServer couldn't be reused as-is because it targets a different Retrofit client. Changed [here](https://github.com/dimagi/commcare-android/pull/3845/changes/27db871f607a2fda82ca9d2f4733ab076a75e701)

app/unit-tests/src/org/commcare/fragments/connect/ConnectLearningProgressFragmentTest.kt
Jignesh-dimagi: Fixed [here](https://github.com/dimagi/commcare-android/pull/3845/changes/27db871f607a2fda82ca9d2f4733ab076a75e701).

app/src/org/commcare/views/connect/ConnectLearnCompleteView.kt
Jignesh-dimagi: Renamed to bindCtaBar() [here](https://github.com/dimagi/commcare-android/pull/3845/changes/27db871f607a2fda82ca9d2f4733ab076a75e701)

app/src/org/commcare/connect/ConnectJobClaimController.kt
Jignesh-dimagi: I actually asked Claude to make that change before submitting the PR, but it brought up an interesting point in its response:

Agreed in principle, but the sole caller here is Java  ConnectLearningProgressFragment.onDeliveryCtaClicked(). With function types the parameters become Function1<Boolean, Unit>, so every lambda body needs an explicit return Unit.INSTANCE;, including at the early returns:

  deliveryAppInstalled -> {
      if (!hasLiveView(this)) {
          return Unit.INSTANCE;   // was: return;
      }
      Navigation.findNavController(view).navigate(...);
      return Unit.INSTANCE;       // new
  },
  
  That's four extra return Unit.INSTANCE; lines plus a kotlin.Unit import in the fragment, and the Boolean boxes — net more boilerplate at the call site than the two fun interface declarations save.

fun interface is what Kotlin provides for this interop case, and it's already the pattern we use elsewhere for Kotlin callbacks consumed from Java: ConnectAppLauncher.OutcomeCallback, LoginController.ResultCallback, LoginProgressListener, and BaseConnectFragment.DataStateConsumer — which this same fragment uses via observeDataState.

Happy to switch to function types once ConnectLearningProgressFragment is converted to Kotlin, in [this ticket](https://dimagi.atlassian.net/browse/CCCT-2518); that conversion felt out of scope for this ticket.

app/src/org/commcare/views/connect/ConnectLearnCompleteView.kt
Jignesh-dimagi: Changed [here](https://github.com/dimagi/commcare-android/pull/3845/changes/27db871f607a2fda82ca9d2f4733ab076a75e701)

app/src/org/commcare/views/connect/ConnectLearnCompleteView.kt
Jignesh-dimagi: Changed [here](https://github.com/dimagi/commcare-android/pull/3845/changes/27db871f607a2fda82ca9d2f4733ab076a75e701)

app/src/org/commcare/fragments/connect/ConnectLearningProgressFragment.java
Jignesh-dimagi: Changed [here](https://github.com/dimagi/commcare-android/pull/3845/changes/27db871f607a2fda82ca9d2f4733ab076a75e701)

app/res/layout/fragment_connect_learning_progress.xml
Jignesh-dimagi: Fair point on the readability, but I've left it. Indenting it properly means shifting the whole subtree, about 215 whitespace-only lines on top of a functional diff of 15 added / 121 removed, which felt like the wrong trade mid-review.

app/src/org/commcare/views/connect/ConnectInfoCard.kt
Jignesh-dimagi: Changed [here](https://github.com/dimagi/commcare-android/pull/3845/changes/27db871f607a2fda82ca9d2f4733ab076a75e701)

RELEASES.md
Jignesh-dimagi: Removed [here](https://github.com/dimagi/commcare-android/pull/3845/changes/b4dc20e13a1cd755d262ffd9e6e4d952e818a269)

app/src/org/commcare/connect/ConnectJobClaimController.kt
shubham1g5: are we able to add a screenshot with this error, I am worried about how we are fitting such long texts on UI

app/src/org/commcare/connect/ConnectJobClaimController.kt
conroy-ricketts: Aaah interesting!

conroy-ricketts:
> We definitely follow that pattern when building individual components, allowing them to be customized via styles. However, for full page layouts composed of multiple components with distinct placements, driving everything through styles becomes overly complex and hard to maintain or identify by name.

> That said, I'm still open to suggestions if anyone has a cleaner approach in mind!

@Jignesh-dimagi

Doing this for individual components but not the layouts they'll be used in feels like it's defeating the purpose, and having a style/theming system doesn't necessarily have to be complex nor hard to maintain. It will be an expensive effort to convert the UI to using themes/styles instead later on when the redesign is all said and done if we ever want to take that route in the future.

For example, we could have a larger `ConnectTheme` for colors that are consistent across the entire redesign, and we can define a couple smaller themes or styles if (or as) necessary

app/src/org/commcare/connect/ConnectJobClaimController.kt
Jignesh-dimagi: That component will handle this automatically by wrapping the text. I've updated the screenshot in the PR description.

Jignesh-dimagi:
> Doing this for individual components but not the layouts they'll be used in feels like it's defeating the purpose, and having a style/theming system doesn't necessarily have to be complex nor hard to maintain. It will be an expensive effort to convert the UI to using themes/styles instead later on when the redesign is all said and done if we ever want to take that route in the future.
> 
> For example, we could have a larger `ConnectTheme` for colors that are consistent across the entire redesign, and we can define a couple smaller themes or styles if (or as) necessary

We are already defining design-specific colors such as connect_text_color, connect_subtext_color, and connect_dark_blue_color etc.

Because this layout uses multiple color variants, standardizing them under generic naming conventions might not be practical. Using semantic names means that if we need to update colors in the future, we would only need to change the values of connect_text_color, connect_subtext_color, etc.

Before moving forward with this, I’d like to get everyone else's opinion. @shubham1g5 @OrangeAndGreen.

conroy-ricketts:
> We are already defining design-specific colors such as connect_text_color, connect_subtext_color, and connect_dark_blue_color etc.

> Because this layout uses multiple color variants, standardizing them under generic naming conventions might not be practical. Using semantic names means that if we need to update colors in the future, we would only need to change the values of connect_text_color, connect_subtext_color, etc.

Ah, this is part of the reason why it's recommended to only define base colors inside `colors.xml`. We're currently using a mix of base colors and colors like `connect_text_color` throughout the fragments/layouts

> Before moving forward with this, I’d like to get everyone else's opinion. 

Sounds good, happy to hear other thoughts

app/src/org/commcare/connect/ConnectJobClaimController.kt
shubham1g5: thanks!

shubham1g5:
@conroy-ricketts I generally agree with what you are thinking here although are you able to give an explicit example of implementation you would like to see here. There are multiple ways here to abstract out the different atrributes from individual layouts to styles and it would be nice for us to align on the implementation approach first here. 

That said, I think we should not block this particular PR on these changes  (as other tickets are blocked on this PR) but we can start implementing and blocking other PRs as soon as we finalize the approach and can trace back the changes to this PR once we have a settled approach. Thoughts ?

shubham1g5 (approved):

OrangeAndGreen (approved):
I agree with putting a bit more thought into how we want to clean up and standardize the color naming and usage (generally on board with moving to more themes), but also okay with merging this PR in the meantime and revisiting it along with other pages when we're ready

conroy-ricketts (approved):
Agreed, let's proceed with these changes and I'll follow up async on Slack

PR #3846, opened 2026-08-04
QA-8611 QA-8610 Update Delivery Progress Tabs In Place On Sync

conroy-ricketts:
## Product Description

Payment details on the Delivery Progress screen now update as soon as a sync completes — a rolled-back payment disappears and a confirmed payment switches to Revert without having to leave the screen and reopen it.

## Technical Summary

The 2.64 offline-first conversion re-reads the job from the DB on every emission, so the parent rebound its own `job` while the tabs and `ConnectActivity` kept the instance they captured — and since the parser mutates whichever instance was passed into the network call, a tab saw fresh data only while it still held that one. Publishing each reload back to `ConnectActivity` restores the pre-2.64 single shared record; `refresh()` iterates the attached child fragments because a ViewPager2 state restore re-adopts the previously added instances rather than the adapter's.

Deliberately left alone: the Progress tab's per-payment-unit report cards still only rebuild from that tab's own sync button. That behaves the same way in 2.63, so it is not part of this regression.

### Safety story

What gives confidence:

- I reproduced both tickets on a device and confirmed each is fixed, including the back-navigation case where the Payment tab previously kept showing pre-sync data after returning from the deliveries list.
- Scope is limited to how Connect job fragments obtain the job record — no data-layer, schema, or network changes.
- The Connect view-model and repository unit tests pass.

Risks to review:

- There is no automated coverage for this path; all verification was manual.
- The Learn progress screen shares the same one-line change, so its record now also propagates to the activity.
- `ConnectActivity.getActiveJob()` is written from more call sites than before. Every writer is the fragment currently displaying that job, but it is worth a second look.

[QA-8611]: https://dimagi.atlassian.net/browse/QA-8611?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[QA-8610]: https://dimagi.atlassian.net/browse/QA-8610?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/org/commcare/fragments/connect/ConnectJobFragment.java
+++ b/app/src/org/commcare/fragments/connect/ConnectJobFragment.java
@@ -17,7 +17,16 @@ public abstract class ConnectJobFragment<T extends ViewBinding> extends BaseConn
     @Override
     public void onCreate(@Nullable Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-        job = ((ConnectActivity)requireActivity()).getActiveJob();
+        reloadActiveJob();
+    }
+
+    protected void setActiveJob(ConnectJobRecord updatedJob) {
+        job = updatedJob;
+        ((ConnectActivity) requireActivity()).setActiveJob(updatedJob);
+    }
+
+    protected void reloadActiveJob() {
+        job = ((ConnectActivity) requireActivity()).getActiveJob();
         Objects.requireNonNull(job);
     }
 
--- a/app/src/org/commcare/fragments/connect/ConnectDeliveryProgressDeliveryFragment.java
+++ b/app/src/org/commcare/fragments/connect/ConnectDeliveryProgressDeliveryFragment.java
@@ -54,6 +54,8 @@ public class ConnectDeliveryProgressDeliveryFragment extends ConnectJobFragment<
     }
 
     public void updateProgressSummary() {
+        reloadActiveJob();
+
         int completed = job.getCompletedVisits();
         int total = job.getMaxVisits();
         int percent = total > 0 ? (100 * completed / total) : 100;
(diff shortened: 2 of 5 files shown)


conroy-ricketts:
## Suggested Review Order

- `ConnectJobFragment.java` — the two new accessors every other change routes through
- `ConnectDeliveryProgressFragment.java` — publishes reloaded records; `refresh()` now iterates attached child fragments
- `ConnectResultsSummaryListFragment.java` — the Payment tab from both tickets; reloads on view creation and on refresh
- `ConnectDeliveryProgressDeliveryFragment.java` — same reload for the Progress tab
- `ConnectLearningProgressFragment.java` — same publish pattern, unrelated screen

app/src/org/commcare/fragments/connect/ConnectJobFragment.java
```diff
@@ -17,7 +17,16 @@ public abstract class ConnectJobFragment<T extends ViewBinding> extends BaseConn
...
+
+    protected void setActiveJob(ConnectJobRecord updatedJob) {
+        job = updatedJob;
+        ((ConnectActivity) requireActivity()).setActiveJob(updatedJob);
+    }
+
+    protected void reloadActiveJob() {
+        job = ((ConnectActivity) requireActivity()).getActiveJob();
         Objects.requireNonNull(job);
     }
```
shubham1g5, line 31: do we need this given the `setActiveJob` itself reset the job to the `updatedJob` ?

shubham1g5 (approved):
Nice catch! Left one comment but nothing blocking.

app/src/org/commcare/fragments/connect/ConnectJobFragment.java
conroy-ricketts: Yeah, without over-explaining, removing it would break the fix by leaving the data for the job stale

shubham1g5: Sounds good

OrangeAndGreen (approved):

PR #3847, opened 2026-08-05
QA-8617 Dispatch Activity Results To The AndroidX Registry

conroy-ricketts:
## Product Description

Backing out of the profile-photo camera opened from the app home screen's nav drawer now returns to the home screen, instead of pushing the user forward into session navigation.

This also fixes another issue: Photos captured from that drawer also now actually upload — previously the result never reached the photo uploader.

## Technical Summary

`SessionAwareCommCareActivity.onActivityResult` never called `super`, which is the only place `ComponentActivity` hands results to the AndroidX result registry. So `registerForActivityResult` callbacks were dead on every session-aware activity, and the registry's request code fell out the bottom of `HomeScreenBaseActivity`'s `switch` into `startNextSessionStepSafe()` — the stray forward navigation QA saw. The dispatch went into `SessionAwareHelper` rather than the activity override so the session-expiration check keeps its existing precedence over the callback, and it uses `dispatchResult` directly rather than `super` because only `dispatchResult` reports whether the result was consumed.

### Safety story

What gives confidence:

- I manually verified both paths on device: backing out of the camera from the home-screen drawer stays on the home screen, and a successful capture uploads and updates the drawer photo.
- No collision is possible between the two code spaces: the registry allocates request codes in `[0x00010000, MAX_INT]`, and every request code in the repo is ≤ 8000 (session-aware constants are 0–10). Legacy results still fall through to `onActivityResultSessionSafe` unchanged.
- The only registry launcher hosted by a session-aware activity is the drawer photo updater. The PersonalID and Connect fragment launchers sit under `NavigationHostCommCareActivity`, outside this path, and `LoginActivity` / `CommCareSetupActivity` already called `super`, so their drawer photo flow was already working.

Risks to review:

- This sits in a base class on the result path of every session-aware activity, so scope of impact is wide even though the behavioral delta is narrow.
- Results the registry consumes no longer reach `onActivityResultSessionSafe`. On the home screen that also means `sessionNavigationProceedingAfterOnResume` is no longer set, so `onResumeSessionSafe` refreshes the home screen rather than suppressing the refresh — intended here, but it is a behavior change.
- Fragments that use the legacy `Fragment.startActivityForResult` get codes encoded as `((index+1) << 16) | code`, landing in the same numeric range as registry codes. Those results already never reached fragments in these activities and still do not; a false hit would require the registry to have randomly drawn that exact value while pending.

### Automated test coverage

None added — verified manually instead, at the author's request. The touched path has no existing unit coverage.

[QA-8617]: https://dimagi.atlassian.net/browse/QA-8617?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/org/commcare/activities/SessionAwareHelper.java
+++ b/app/src/org/commcare/activities/SessionAwareHelper.java
@@ -45,8 +45,12 @@ public class SessionAwareHelper {
                                                  int requestCode, int resultCode, Intent intent) {
         boolean redirectedToLogin =
                 SessionRegistrationHelper.handleSessionExpiration(a);
-        if (!redirectedToLogin) {
-            sessionAware.onActivityResultSessionSafe(requestCode, resultCode, intent);
+        if (redirectedToLogin) {
+            return;
         }
+        if (a.getActivityResultRegistry().dispatchResult(requestCode, resultCode, intent)) {
+            return;
+        }
+        sessionAware.onActivityResultSessionSafe(requestCode, resultCode, intent);
     }
 }
--- a/docs/personalid/photo_management.md
+++ b/docs/personalid/photo_management.md
@@ -39,6 +39,9 @@ endpoint, stored locally, and reflected in the drawer header.
     (e.g. the drawer's warning icon).
   * Re-reads the `ConnectUserRecord` from the database at upload time rather than caching
     it, so it always reflects the currently signed-in user.
+  * Relies on the host activity letting the AndroidX activity-result registry see results;
+    `SessionAwareHelper` dispatches to the registry before falling back to CommCare's legacy
+    request-code handling, which is what makes the flow work on the app home screen.
 
 * **`BaseDrawerController`** (`navdrawer/BaseDrawerController.kt`)
   * Wires the drawer's user image tap to the updater's `initiatePhotoUpdate()`.


conroy-ricketts:
## Suggested Review Order

- `app/src/org/commcare/activities/SessionAwareHelper.java` — the fix; the ordering of expiration check → registry dispatch → legacy handling is the whole review
- `docs/personalid/photo_management.md` — records the host requirement the photo flow now depends on

shubham1g5:
@conroy-ricketts 

I think `onActivityResult` in `SessionAwareCommCareActivity` should be making the super call as the first thing instead. Have you tried that solution here ? That seems to me more in line of standard patterns to follow for onActivityResult call chain.

conroy-ricketts:
> I think onActivityResult in SessionAwareCommCareActivity should be making the super call as the first thing instead. Have you tried that solution here ? That seems to me more in line of standard patterns to follow for onActivityResult call chain.

@shubham1g5 

Doing this would only fix half the problem (i.e. fixes photo upload but not the navigation issue)

Think the real issue here is that `HomeScreenBaseActivity` currently treats any unrecognized request code as "continue session navigation", and I think properly fixing that is more effort than we would like for a 2.64 QA fix. Let me know if I should create a new backlog ticket for this. Properly fixing it would mean changes to code that affects the main navigation flows of traditional CommCare apps and would require regression testing

shubham1g5:
@conroy-ricketts I feel like fixing that would be a simple addition to [existing switch construct](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/activities/HomeScreenBaseActivity.java#L659), and the change should be isolated enough if we make the change based on the photo capture request code. Let me know if I am missing something here.

conroy-ricketts:
@shubham1g5

Think doing this would require us to move the photo update flow off the modern API and onto the legacy (and now deprecated) `startActivityForResult()` method instead.

Currently, the photo update flow uses `registerForActivityResult()` [here](https://github.com/dimagi/commcare-android/blob/ebb051c9698f837d015618556dd81be2f1ff714a/app/src/org/commcare/personalId/photo/PersonalIdPhotoUpdater.kt#L33).

Please correct me if I'm wrong, but I believe that `registerForActivityResult` allocates a random request code at registration time, differing per process. It'd be difficult to add a case for it to that switch construct in a non-destructive way.

OrangeAndGreen (approved):
Interesting discussion on this one but I think the proposed solution is the correct one based on reading both of your comments and some discussion with Claude to confirm.

shubham1g5 (approved):

shubham1g5:
> Let me know if I should create a new backlog ticket for this

This would be great, @OrangeAndGreen Thoughts on adding it to the home activity refactor work as a new ticket ?

conroy-ricketts:
Merging this for now. Please tag me if I should create a new ticket for that

PR #3848, opened 2026-08-05
CCCT-2302 Unit Tests And Kotlin Conversion For The Backup Code Screen

conroy-ricketts:
## Technical Summary

Adds unit tests to the backup code screen and converts it to Kotlin.

### Robolectric SDK pins removed

`NumericCodeView` used try-with-resources on `TypedArray`, whose `close()` is API 31+. D8 backports that call, so shipped builds are fine on every supported API level (verified on an API 30 emulator) — but Robolectric doesn't run D8, so any test inflating this view had to be pinned to API 32, costing coverage on the API levels most users run. Switching to `recycle()` removes the constraint. The pin is also dropped from `PersonalIdEmailVerificationFragmentTest`, which carried the same workaround for the same reason.

## Open Questions

Two pre-existing issues surfaced while writing the tests. **Neither is fixed here**, and both are worth a decision before the backup code change management work builds on this screen:

- The recovery path never writes the entered code onto the session data, so `PersonalIdRecoveryCompleter` stores a `ConnectUserRecord` with a null `pin`. A test documents the current behavior — is that intended?
- `not_me_button` is `visibility="gone"` in the layout and the fragment never shows it (deliberately hidden in 0cb8f34f9), so its click listener is unreachable in production. I kept the listener and exercise it directly from a test. Should we rather delete the code for this button?

### Safety story

What gives confidence:

- I reviewed the diff and manually verified the backup code screen still works after the conversion.
- The conversion is line-for-line. No caller, layout, or navigation contract changed, and the class keeps its name and package.
- Both modes of the screen are now covered by tests that drive the real fragment against a mock server.

Risks to review:

- Account recovery runs through this screen, so a regression here blocks a user from recovering their account on a new device.
- The tests are new, so they encode post-conversion behavior — they would not catch a conversion mistake that I mirrored into a test. The manual pass over the screen is the mitigation.

### Automated test coverage

33 new Robolectric tests split across two classes, one per screen mode.

[CCCT-2302]: https://dimagi.atlassian.net/browse/CCCT-2302?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


conroy-ricketts:
## Suggested Review Order

- `app/src/org/commcare/fragments/personalId/PersonalIdBackupCodeFragment.kt` — the conversion; diff it against the deleted `.java` in the same commit
- `app/unit-tests/src/org/commcare/fragments/personalId/BasePersonalIdBackupCodeFragmentTest.kt` — harness the two test classes share: session-data factory, launch, view accessors
- `app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdBackupCodeFragmentTest.kt` — set-code mode: validation, mismatch, submission routing
- `app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdBackupCodeFragmentRecoveryTest.kt` — recovery mode, including the API responses and the null-`pin` note from the description

app/src/org/commcare/fragments/personalId/PersonalIdBackupCodeFragment.kt
```diff
@@ -0,0 +1,300 @@
...
+            binding.welcomeBackLayout.visibility = View.GONE
+        }
+    }
+
+    private fun setUserNameAndPhoto() {
+        binding.welcomeBack.text = getString(R.string.personalid_welcome_back_msg, personalIdSessionData.userName)
+        val photoBase64 = personalIdSessionData.photoBase64
+        if (!photoBase64.isNullOrEmpty()) {
+            binding.userPhoto.setImageBitmap(MediaUtil.decodeBase64EncodedBitmap(photoBase64))
+        }
```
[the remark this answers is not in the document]
conroy-ricketts: **[AI — Claude Opus 5]**

Not taking this one. The suggestion isn't baseless — `photoBase64` is loaded through Glide in `BaseDrawerController.loadUserPhoto()` and `BasePersonalIdProfileFragment.loadUserPhoto()` — but there are two reasons against it here:

1. **Scope.** This PR is a behavior-preserving Java→Kotlin conversion plus tests. Swapping the image-loading mechanism is a production behavior change the ticket doesn't call for.
2. **Regression risk.** `MediaUtil.decodeBase64EncodedBitmap` strips a `data:image` prefix *and* decodes bare base64. Glide's String loader only handles the `data:` form. Camera captures carry the `data:image/webp;base64,` prefix (`MicroImageActivity.BASE_64_IMAGE_PREFIX`), but the photo on this screen arrives from the server in the session data, and I can't confirm it is prefixed. If it isn't, the swap silently renders the error placeholder on exactly the screen this PR covers.

Unifying photo loading across these screens is worth doing deliberately, with the server payload format confirmed first.

On the test: the `BitmapDrawable` check is not incidental. The layout's `android:src` is a vector placeholder, so a `BitmapDrawable` can only come from `setImageBitmap()`, and the assertion goes on to compare the shadow's source bytes against `Base64.decode(TEST_PHOTO_BASE64)`. That ties the rendered image to the session payload rather than to a rendering implementation detail.

app/src/org/commcare/fragments/personalId/PersonalIdBackupCodeFragment.kt
```diff
@@ -0,0 +1,300 @@
...
+
+        binding.confirmCodeView.setCodeCompleteListener { submitIfEnabled() }
+
+        val enterKeyListener = NumericCodeView.OnEnterKeyPressedListener { submitIfEnabled() }
+        binding.backupCodeView.setOnEnterKeyPressedListener(enterKeyListener)
+        binding.confirmCodeView.setOnEnterKeyPressedListener(enterKeyListener)
+
+        binding.connectBackupCodeButton.setOnClickListener { handleBackupCodeSubmission() }
+        binding.notMeButton.setOnClickListener { handleNotMeButtonPressed() }
```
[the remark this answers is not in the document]
conroy-ricketts: **[AI — Claude Opus 5]**

Half taken.

**Not un-hiding the control.** `not_me_button` was hidden deliberately in 0cb8f34f9 ("hide this is not me"), which removed the `setVisibility(VISIBLE)` call from the fragment and added `android:visibility="gone"` to the layout in the same commit. Making it visible again would revert a product decision, not fix a bug — and it would put an untested "this isn't me" path back in front of users mid-recovery. That call belongs to the team, which is why the PR description raises it as an open question rather than acting on it.

**Taken:** the test now asserts the control is `GONE` before driving its listener (1c642fb02), so the fact that it bypasses the production visibility constraint is explicit in the test rather than implied. Asserting the control is *visible*, as suggested, would fail against current behavior.

app/src/org/commcare/fragments/personalId/PersonalIdBackupCodeFragment.kt
```diff
@@ -0,0 +1,300 @@
...
+        binding.connectBackupCodeErrorMessage.visibility = if (showMismatch) View.VISIBLE else View.GONE
+        binding.connectBackupCodeErrorMessage.text = errorText
+        enableContinueButton(isCodeComplete && isCodeConfirmed)
+    }
+
+    private fun handleNotMeButtonPressed() {
+        personalIdSessionData.accountExists = false
+        clearBackupCodeFields()
+        configureUiByMode()
+    }
```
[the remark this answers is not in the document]
conroy-ricketts: **[AI — Claude Opus 5]**

Correct analysis, deliberately not fixed here.

`configureUiByMode()` does update `titleId` without the activity title following it, and the Java original had the same gap — this is a faithful conversion of it, not something introduced by this PR.

It is also unreachable in production: the only caller of `handleNotMeButtonPressed()` is `not_me_button`, which is permanently `GONE` (see the sibling thread). Only the test reaches it, via `performClick()`.

So the one-line fix would change behavior on a dead path in a PR whose value is being behavior-preserving. It belongs with the decision on whether to bring the not-me control back — if that happens, this needs to ship alongside it, otherwise the screen keeps the confirm-code title after switching to set-code mode. Flagging it here so it isn't lost.

app/src/org/commcare/fragments/personalId/PersonalIdBackupCodeFragment.kt
```diff
@@ -0,0 +1,300 @@
...
+            ) {
+                if (handleCommonSignupFailures(failureCode)) {
+                    return
+                }
+                showError(PersonalIdOrConnectApiErrorHandler.handle(requireActivity(), failureCode, t))
+                if (failureCode.shouldAllowRetry()) {
+                    enableContinueButton(true)
+                }
+            }
+        }.confirmBackupCode(activity, backupCode, personalIdSessionData)
```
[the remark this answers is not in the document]
conroy-ricketts: **[AI — Claude Opus 5]**

Right about the behavior, intentionally left alone. This is the open question the PR description raises, and the reasoning for not fixing it here:

- **It predates this PR.** The Java original never assigned `backupCode` on the recovery path either. Fixing it inside a behavior-preserving conversion would hide a real data-behavior change inside a "no functional change" diff.
- **It has consequences worth a decision.** `ConnectUserRecord.pin` feeds device unlock, so starting to populate it on recovery changes what a recovered account can do on the device. That is a product call, and it lands squarely in the backup code change management work this ticket was sequenced ahead of.

The test asserts `assertNull(storedUser.pin)` against current behavior on purpose. When the team decides to populate it, that assertion flips to `assertEquals(TEST_BACKUP_CODE, storedUser.pin)` and the test turns into the proof the fix worked — which is the reason for writing this coverage before the change management work starts.

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdBackupCodeFragmentTest.kt
```diff
@@ -0,0 +1,227 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.annotation.Config
+
+/**
+ * Tests [PersonalIdBackupCodeFragment] in account registration mode.
+ *
+ * Pinned to API 32 because NumericCodeView uses try-with-resources on TypedArray, and
+ * TypedArray.close() was added in API 31. Robolectric 4.8.2 supports up to API 32 (S_V2).
+ */
```
shubham1g5: would not it then create an issue in production given we are supporting Personal ID signups on Android 9+ devices ?

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdBackupCodeFragmentTest.kt
```diff
@@ -0,0 +1,227 @@
...
+    @Test
+    fun `no email offer date is recorded when the email toggle is inactive`() {
+        enterBackupCode(TEST_BACKUP_CODE)
+        enterConfirmCode(TEST_BACKUP_CODE)
+
+        assertNull(PersonalIdUserPreferences.getLastEmailOfferDate())
+    }
+
+    @Test
+    fun `matching codes navigate to the email screen when the email toggle is active`() {
```
shubham1g5, line 141: we should also add a test for the UX when email toggle is inactive.

conroy-ricketts: I was thinking that this case is already covered well by these tests (although, I do realize that the behavior is implicit):
- [test 1](https://github.com/dimagi/commcare-android/pull/3848/changes#diff-85b7f3fd3deb30f48609ba293ef3f305f1794a39498547e7d1cacf01751b7eecR128)
- [test 2](https://github.com/dimagi/commcare-android/pull/3848/changes#diff-85b7f3fd3deb30f48609ba293ef3f305f1794a39498547e7d1cacf01751b7eecR137)
- [test 3](https://github.com/dimagi/commcare-android/pull/3848/changes#diff-4d7787a346c8681e2fe9202f7df63469184d58703c753c18fa690cebb397110aR150)

Let me know if we should rather rename those tests, make assertions, or add code comments so that it's explicit that the toggle is inactive in those test cases

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdBackupCodeFragmentTest.kt
conroy-ricketts: Ah, I could have been more clear in the code comment.

There is no user-facing issue exposed here - I verified this by testing the PersonalID signup flow on an emulator running API 30.

This issue only affects our unit tests.

Flagging that the unit tests we created for the email screens have this same exact issue because of the `NumericCodeView`: [see here](https://github.com/dimagi/commcare-android/blob/e1c8ba80ab43114c48d6eda3dd73e5cc724ce194/app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdEmailVerificationFragmentTest.kt#L35)

I think a much better solution here would be to do a small tweak to `NumericCodeView` so that our unit tests don't have this issue at all in the first place. I'll push a commit soon, let me know what you think

conroy-ricketts: 967881816030e3ee28f0f4df58e2f4f90a8ba46e

shubham1g5: The fix seems great to me although still confused why it was not affecting the live app and only robolectric (though fine to ignore given the fix eliminates this conversation altogether)

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdBackupCodeFragmentTest.kt
shubham1g5: that makes sense, I didn't review the tests in great detail.

shubham1g5 (approved):

PR #3850, opened 2026-08-06
Fall Back To PersonalID SMS (Twilio) When Firebase Cannot Send The OTP

Jignesh-dimagi:
## Product Description

When Firebase cannot deliver the phone verification code during PersonalID sign-up, the screen now falls back to sending it via PersonalID (Twilio) automatically, with no need for the user to tap Resend Code. Previously the fallback only fired for a narrow set of Firebase errors, so most real delivery failures left the user stuck on the code entry screen with no way forward.

## Technical Summary

The fix inverts the classification: everything except `INVALID_CREDENTIAL` is non-recoverable. This makes falling back the default for unrecognised errors instead of dead-ending, which matters because the default branch is where new or unmapped Firebase errors arrive.

### What Firebase actually throws

Verified against the status-code mapper in firebase-auth 23.2.0.

| Firebase failure | SDK exception | `OtpErrorType` | Fell back before | Falls back now |
|---|---|---|---|---|
| `ERROR_QUOTA_EXCEEDED` (17052) — project SMS quota exhausted | `FirebaseTooManyRequestsException` | `TOO_MANY_REQUESTS` | No | **Yes** |
| Device blocked for unusual activity (17010) | `FirebaseTooManyRequestsException` | `TOO_MANY_REQUESTS` | No | **Yes** |
| Any code absent from the SDK's table (default branch) — e.g. billing disabled, SMS region blocked | plain `FirebaseException` | `GENERIC_ERROR` | No | **Yes** |
| Internal error (17499) | plain `FirebaseException` | `GENERIC_ERROR` | No | **Yes** |
| Network failure (17020) | `FirebaseNetworkException` | `GENERIC_ERROR` | No | **Yes** |
| Play Services unavailable (17063 / 17080) | `FirebaseApiNotAvailableException` | `GENERIC_ERROR` | No | **Yes** |
| reCAPTCHA cannot be shown | `FirebaseAuthMissingActivityForRecaptchaException` | `MISSING_ACTIVITY` | No | **Yes** |
| `ERROR_APP_NOT_AUTHORIZED` (17028), `ERROR_INVALID_CERT_HASH` (17064), `ERROR_MISSING_CLIENT_IDENTIFIER` (17093) | `FirebaseAuthException` | `VERIFICATION_FAILED` | Yes | Yes |
| Bad phone number, or wrong/expired code | `FirebaseAuthInvalidCredentialsException` | `INVALID_CREDENTIAL` | No | No — intentional |

### Why `INVALID_CREDENTIAL` stays recoverable

It is the one bucket the user can act on themselves. These are the codes Firebase routes into it that this app's phone flow can reach:

| Error | Code | SDK description |
|---|---|---|
| `ERROR_INVALID_VERIFICATION_CODE` | 17044 | The verification code from SMS/TOTP is invalid. |
| `ERROR_MISSING_VERIFICATION_CODE` | 17043 | The verification code from SMS/TOTP is empty. |
| `ERROR_SESSION_EXPIRED` | 17051 | The sms code has expired. Please re-send the verification code to try again. |
| `ERROR_INVALID_PHONE_NUMBER` | 17042 | The format of the phone number provided is incorrect. |
| `ERROR_MISSING_PHONE_NUMBER` | 17041 | To send verification codes, provide a phone number for the recipient. |
| `ERROR_INVALID_VERIFICATION_ID` | 17046 | The verification ID used to create the phone auth credential is invalid. |
| `ERROR_MISSING_VERIFICATION_ID` | 17045 | The Phone Auth Credential was created with an empty verification ID. |
| `ERROR_RETRY_PHONE_AUTH` | 17049 | An error occurred during authentication using the PhoneAuthCredential. Please retry authentication. |
| `ERROR_INVALID_CREDENTIAL` | 17004 | The supplied auth credential is incorrect, malformed or has expired. |
| `ERROR_REJECTED_CREDENTIAL` | 17075 | The request contains malformed or mismatching credentials. |

The remaining codes Firebase routes into this exception belong to email/password, custom-token, multi-factor and passkey flows this app doesn't use.

The `instanceof` ordering in `errorTypeFrom` is load-bearing: `FirebaseAuthInvalidCredentialsException` extends `FirebaseAuthException`, so the subclass check must stay ahead of the parent check or a mistyped code would be classified non-recoverable and silently trigger a resend instead of showing "incorrect OTP".

### Request cycle — `smsMethod = "firebase"`, `otp_fallback` true

| Step | `otpAttempts` in | Manager chosen | Outcome | `otpAttempts` out |
|---|---|---|---|---|
| Initial request | 0 | Firebase | non-recoverable failure | 1 |
| Auto-switch | 1 | PersonalID | Twilio SMS sent | 2 |
| Resend #1 | 2 | Firebase | fails again | 3 |
| Auto-switch | 3 | PersonalID | Twilio SMS sent | 4 |
| Resend #2 | 4 | Firebase | fails again | 5 |
| Auto-switch | 5 | PersonalID | Twilio SMS sent | 6 |

Resend deliberately retries Firebase first and falls back on failure, so each press costs one failed Firebase round-trip and advances `otpAttempts` by two. When `otp_fallback` is false there is nothing to switch to, so the error surfaces and `otpAttempts` advances by one.

### Request cycle — `smsMethod = "personal_id"`

| Step | `otpAttempts` in | Manager chosen | Outcome | `otpAttempts` out |
|---|---|---|---|---|
| Initial request | 0 | PersonalID | Twilio SMS sent | 1 |
| Resend #1 | 1 | PersonalID | Twilio SMS sent | 2 |
| Resend #2 | 2 | PersonalID | Twilio SMS sent | 3 |

Firebase is never used for these users. Both branches of `setupOtpManager` resolve to `PersonalIdAuthService`, because the 3-arg `OtpManager` constructor re-derives the method from session data. The auto-switch never fires, and `PersonalIdAuthService` never calls `onFailure`, so the classification change has no effect on this configuration.

### Safety story

*What gives confidence:*
- Tested this on my device by triggering frequent Firebase OTP requests until hitting the rate limit. The fallback kicked in as expected—subsequent OTPs were delivered via Twilio, and the flow completed successfully.
- That manual test covered `FirebaseTooManyRequestsException` → `TOO_MANY_REQUESTS`, one of the types this change newly treats as non-recoverable, so it exercised a path that did not previously fall back.
- The classification change is a single expression, and the fallback path it feeds was already in place.
- New tests cover the exception classification, the fallback decision, and both `smsMethod` configurations; several fail against the pre-fix code.
- Users configured for `personal_id` are provably unaffected — that path never reaches the changed code.

### Automated test coverage

23 tests across four classes: Firebase exception → `OtpErrorType` (`FirebaseAuthServiceTest`), the recoverable/non-recoverable split (`OtpErrorTypeTest`), auth-service resolution per `smsMethod` (`OtpManagerTest`), and the fallback decision plus method tracking on the fragment (`PersonalIdPhoneVerificationFragmentTest`). Verified by reverting the production changes: 4 of the 7 fragment tests and both `OtpErrorTypeTest` cases fail without them. Full suite: 456 tests, 25 failures, all pre-existing and identical on a pristine tree.

### QA Plan

Covered by the QA Notes added to `RELEASES.md` under 2.63.5.


Jignesh-dimagi:
## Suggested Review Order

- `app/src/org/commcare/utils/OtpErrorType.java` — the classification flip; the whole fix follows from this one expression
- `app/src/org/commcare/utils/FirebaseAuthService.java` — which Firebase exception becomes which error type; ordering of the `instanceof` checks matters
- `app/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragment.java` — consumes the classification; gates the auto-switch and tracks the resolved method
- `app/unit-tests/src/org/commcare/utils/FirebaseAuthServiceTest.kt` — pins the exception-to-error-type table above
- `app/unit-tests/src/org/commcare/utils/OtpErrorTypeTest.kt` — pins the recoverable/non-recoverable split
- `app/unit-tests/src/org/commcare/utils/OtpManagerTest.kt` — which auth service each constructor resolves to, per `smsMethod`
- `app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragmentTest.kt` — the fallback decision and method tracking on the screen
- `app/unit-tests/src/org/commcare/android/util/FirebaseTestUtils.kt` — shared Robolectric Firebase setup; no production behavior

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragmentTest.kt
```diff
@@ -0,0 +1,162 @@
...
+                .getDeclaredField("lastOtpMethod")
+        field.isAccessible = true
+        return field.get(verificationFragment) as String?
+    }
+
+    /**
+     * The callback the fragment hands to OtpManager. Invoking it is what FirebaseAuthService does
+     * from handleFirebaseException; there is no view interaction that can produce a Firebase error.
+     */
+    private fun otpCallback(): OtpVerificationCallback {
```
shubham1g5: we really should not mimick callbacks in test and instead mock the layer that is responsible for triggering callbacks to provide a good test coverage as otherwise we are assuming that the callback trigeering code in the particular fragment is bug free.

app/unit-tests/src/org/commcare/utils/OtpErrorTypeTest.kt
```diff
@@ -0,0 +1,27 @@
+package org.commcare.utils
+
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Test
+
+class OtpErrorTypeTest {
```
shubham1g5: nit: these just seeems to repeat the functionality already tested in `FirebaseAuthServiceTest` and think can be deleted instead (I don't think only one of these test classes can pass standalone and other fail)

shubham1g5 (changes requested):
Left one request on test harnessing, otherwise looks good to go.

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragmentTest.kt
Jignesh-dimagi: Agree. 
Done https://github.com/dimagi/commcare-android/pull/3850/changes/3b2c4dc0ecf8627ecb989a8d5c8d21c4dbf02773

I actually tried having Claude test the full loop instead of invoking the individual callback, but it kept pointing back to shadowing PhoneAuthProvider. I originally reverted that approach thinking it was overkill, but since you both flagged the same thing, I went ahead and re-introduced ShadowPhoneAuthProvider to close the loop!

Claude also doesn't feel happy while reverting: ` The reviewer is right, and worth being upfront: that's exactly the approach I built two steps ago — a Robolectric shadow of ShadowPhoneAuthProvider — which you had me revert as too much work` :)

app/unit-tests/src/org/commcare/utils/OtpErrorTypeTest.kt
Jignesh-dimagi: Delete [here](https://github.com/dimagi/commcare-android/pull/3850/changes/3b2c4dc0ecf8627ecb989a8d5c8d21c4dbf02773)

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragmentTest.kt
```diff
@@ -0,0 +1,208 @@
...
+ * Runs above the project-wide sdk=23 default because the OTP screen inflates NumericCodeView,
+ * which closes a TypedArray with try-with-resources; TypedArray only became AutoCloseable at API 31.
+ */
+@Config(
+    application = CommCareTestApplication::class,
+    sdk = [31],
+    shadows = [ShadowPhoneAuthProvider::class],
+)
+@RunWith(AndroidJUnit4::class)
+class PersonalIdPhoneVerificationFragmentTest : BasePersonalIdPhoneFragmentTest() {
```
shubham1g5, line 59: it would be great to have a success test (otp verified) in this class but not blocking on it.

shubham1g5 (approved):

app/unit-tests/src/org/commcare/fragments/personalId/PersonalIdPhoneVerificationFragmentTest.kt
Jignesh-dimagi: Added [here](https://github.com/dimagi/commcare-android/pull/3850/changes/3aa15fe80df038589207ae5767afb6ff8a1db49a)

shubham1g5 (approved):

PR #3851, opened 2026-08-06
Display Date and Time widget components horizontally in landscape

avazirna:
## Product Description

This PR changes the layout of the DateTime widget in landscape mode: previously the date and time pickers would be rendered vertically regardless of the orientation, so even when the device was rotated to landscape. They should now be laid out correctly — stacked vertically in portrait and side by side in landscape.

<table>
<tr>
<th>
Porttrait
</th>
<th>
landscape
</th>
</tr>
<tr>
<td>

</td>
<td>

</td>
</tr>
</table>

## Technical Summary
The initial approach was to change the orientation of the parent layout,`QuestionWidget`, from a `vertical` to a `horizontal` `LinearLayout`. However, this layout also contains  the question label, help, and hint views. So changing its orientation was resulting in all components to be laid out in a single row. And since the question label spans the full width, it occupied the entire row, leaving the date and time pickers with no available width.
The solution was to Wrap the two pickers in their own dedicated `LinearLayout` container, added to the `QuestionWidget` root as a single child, so the root's own vertical orientation is left untouched.
- Toggle only that container's orientation based on `Configuration.orientation` 

### Safety story
- Successfully tested locally
- The change is confined to `DateTimeWidget`'s view construction/layout; there are no data-model, storage, or answer-handling changes (answer read/write paths are untouched).

## Labels and Review

- [ ] Do we need to enhance the manual QA test coverage ? If yes, [RELEASES.md](../RELEASES.md) is updated accordingly
- [ ] Does the PR introduce any major changes worth communicating ? If yes, [RELEASES.md](../RELEASES.md) is updated accordingly

### DIFF
--- a/app/src/org/commcare/views/widgets/DateTimeWidget.java
+++ b/app/src/org/commcare/views/widgets/DateTimeWidget.java
@@ -1,8 +1,10 @@
 package org.commcare.views.widgets;
 
 import android.content.Context;
+import android.content.res.Configuration;
 import android.view.Gravity;
 import android.view.LayoutInflater;
+import android.view.View;
 import android.view.inputmethod.InputMethodManager;
 import android.widget.CalendarView;
 import android.widget.DatePicker;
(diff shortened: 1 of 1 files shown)


shubham1g5 (approved):

shubham1g5:
Same comment as the other PR here to co-ordinate with QA on this for the release.

PR #3852, opened 2026-08-06
QA-8614 Restore Active Job When Connect Activity Is Recreated

conroy-ricketts:
## Product Description

Returning to the Learn or Delivery progress screen after Android has torn down and rebuilt the Connect activity now restores the screen the user was on, instead of crashing out to Connect home.

## Technical Summary

The active job lives only in memory on `ConnectActivity`, so it was gone whenever the activity was rebuilt from saved state. It is now saved as an opportunity UUID and reloaded at the top of `onCreate`, before `super.onCreate()` — the job screens read the active job from their own `onCreate`, which runs while the fragment back stack is being restored inside that `super` call, so the reload has to happen ahead of it.

### Safety story

What gives confidence:

- I reproduced the crash on a physical device on the Learn progress screen, then confirmed the same steps no longer crash with this change: the screen and its back stack are restored intact, with the correct job, and back navigation still returns to the jobs list.
- The change is confined to saving and restoring one value on a single activity; no data model, migration, or network behavior is touched.

Risks to review:

- Restoration is exercised on every rebuild of `ConnectActivity`, so it is worth a look at the notification deep-link entry points, which set the active job from intent extras rather than saved state.
- If the saved opportunity no longer resolves to a record, the job screens still fail fast as they do today; this change does not alter that behavior.

[QA-8614]: https://dimagi.atlassian.net/browse/QA-8614?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/org/commcare/activities/connect/ConnectActivity.java
+++ b/app/src/org/commcare/activities/connect/ConnectActivity.java
@@ -62,6 +62,13 @@ public class ConnectActivity extends NavigationHostCommCareActivity<ConnectActiv
 
     @Override
     protected void onCreate(@Nullable Bundle savedInstanceState) {
+        if (savedInstanceState != null) {
+            String savedJobUuid = savedInstanceState.getString(OPPORTUNITY_UUID);
+            if (!Strings.isNullOrEmpty(savedJobUuid)) {
+                job = ConnectJobUtils.getCompositeJob(this, savedJobUuid);
+            }
+        }
+
         super.onCreate(savedInstanceState);
 
         PersonalIdManager personalIdManager = PersonalIdManager.getInstance();
@@ -102,6 +109,14 @@ public class ConnectActivity extends NavigationHostCommCareActivity<ConnectActiv
     public void onPostCreate(@Nullable Bundle savedInstanceState) {
         super.onPostCreate(savedInstanceState);
         getSupportActionBar().setTitle(getString(R.string.connect_title));
+    }
+
+    @Override
+    protected void onSaveInstanceState(Bundle outState) {
+        super.onSaveInstanceState(outState);
+        if (job != null) {
+            outState.putString(OPPORTUNITY_UUID, job.getJobUUID());
+        }
     }
 
     private int getStartDestinationId(Bundle startArgs) {
--- a/app/src/org/commcare/activities/connect/ConnectActivity.java
+++ b/app/src/org/commcare/activities/connect/ConnectActivity.java
@@ -132,7 +132,7 @@ public class ConnectActivity extends NavigationHostCommCareActivity<ConnectActiv
     private void initStateFromExtras() {
         redirectionAction = getIntent().getStringExtra(REDIRECT_ACTION);
         opportunityUuid = getIntent().getStringExtra(OPPORTUNITY_UUID);
-        if (!TextUtils.isEmpty(opportunityUuid)) {
+        if (job == null && !TextUtils.isEmpty(opportunityUuid)) {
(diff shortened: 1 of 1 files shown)


app/src/org/commcare/activities/connect/ConnectActivity.java
```diff
@@ -62,6 +62,13 @@ public class ConnectActivity extends NavigationHostCommCareActivity<ConnectActiv
 
     @Override
     protected void onCreate(@Nullable Bundle savedInstanceState) {
+        if (savedInstanceState != null) {
+            String savedJobUuid = savedInstanceState.getString(OPPORTUNITY_UUID);
+            if (!Strings.isNullOrEmpty(savedJobUuid)) {
+                job = ConnectJobUtils.getCompositeJob(this, savedJobUuid);
+            }
+        }
+
```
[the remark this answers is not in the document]
conroy-ricketts: **[AI - Claude Opus 5]**

Confirmed, and fixed in 5892eafcd.

The scenario is reachable: `ConnectUnlockFragment.setFragmentRedirection()` falls back to `connect_jobs_list_fragment` whenever the redirect action is empty or unrecognized — including the opp-invite-failure path, which clears it deliberately. So an activity launched from a notification carrying `OPPORTUNITY_UUID` for job A can land the user on the jobs list, where they select job B. On recreation the restore sets `job` to B, the fragments rebuild against B, and then `initStateFromExtras()` overwrites the activity's job with A from the retained intent. The visible fragment holds its own reference so the screen stays correct, but `getActiveJob()` returns A until the fragment's data observer calls `setActiveJob(B)` — long enough for the job detail bottom sheet to show the wrong job.

Fixed by letting restored state win over the intent extras:

```java
if (job == null && !TextUtils.isEmpty(opportunityUuid)) {
    job = ConnectJobUtils.getCompositeJob(this, opportunityUuid);
}
```

Fresh creation is unchanged, since `job` is null there and the intent still wins.

On the redirect start args: those turned out not to bite. `NavController.setGraph` ignores `startDestinationArgs` when there is restored navigation state, so on recreation `handleInfoRedirect`/`handleSecureRedirect` only cost some wasted work — the restored back stack is what actually gets used. I left `redirectionAction` and `opportunityUuid` being read unconditionally rather than gating all of `initStateFromExtras()` on fresh creation, since those fields are consumed elsewhere and blanking them has wider reach than this fix warrants.

app/src/org/commcare/activities/connect/ConnectActivity.java
```diff
@@ -62,6 +62,13 @@ public class ConnectActivity extends NavigationHostCommCareActivity<ConnectActiv
 
     @Override
     protected void onCreate(@Nullable Bundle savedInstanceState) {
+        if (savedInstanceState != null) {
+            String savedJobUuid = savedInstanceState.getString(OPPORTUNITY_UUID);
+            if (savedJobUuid != null) {
+                job = ConnectJobUtils.getCompositeJob(this, savedJobUuid);
```
shubham1g5, line 68: nit: we can abstract the call into a `loadJob` method

shubham1g5 (approved):
nice fix.

app/src/org/commcare/activities/connect/ConnectActivity.java
conroy-ricketts: chose to skip

PR #3853, opened 2026-08-07
CCCT-2655 Add Missing Connect Endpoints To Coroutine Network Stack

shubham1g5:
## Technical Summary

Part 1 of a two-part migration away from the callback-based `ConnectApiHandler`/`ApiConnect`/`ConnectApiClient` stack. Adds three missing endpoints (`startLearnApp`, `claimJob`, `confirmPayments`) to the existing coroutine-based `ConnectNetworkClient` → `ConnectRepository` chain, without touching call sites or removing old code. Part 2 will migrate call sites and delete the old stack.

### Safety story

**What gives confidence:**
- Purely additive — no existing code paths are modified. The new methods are not called from anywhere yet.
- New unit tests cover all three methods in both `ConnectNetworkClient` and `ConnectRepository`.
- The new methods follow the same `executeApiCall` pattern already used and tested by existing endpoints.

**Risks to review:**
- I reviewed the diff only; no device testing was done, as the changes are no-op in production until Part 2 wires them up.

### Automated test coverage

- `ConnectNetworkClientTest`: tests for `startLearnApp`, `claimJob`, `confirmPayments` (success + failure paths)
- `ConnectRepositoryTest`: tests for `startLearning`, `claimJob`, `confirmPayments` (success + failure paths)

[CCCT-2655]: https://dimagi.atlassian.net/browse/CCCT-2655?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/org/commcare/connect/network/connect/models/ConnectRequestModels.kt
+++ b/app/src/org/commcare/connect/network/connect/models/ConnectRequestModels.kt
@@ -1,0 +1,12 @@
+package org.commcare.connect.network.connect.models
+
+import com.google.gson.annotations.SerializedName
+
+data class ConfirmPaymentsRequest(
+    @SerializedName("payments") val payments: List<PaymentConfirmationBody>,
+)
+
+data class PaymentConfirmationBody(
+    @SerializedName("id") val id: String,
+    @SerializedName("confirmed") val confirmed: String,
+)
--- a/app/src/org/commcare/connect/repository/ConnectRepository.kt
+++ b/app/src/org/commcare/connect/repository/ConnectRepository.kt
@@ -95,20 +95,13 @@ class ConnectRepository
                 mapToEmit = { _ -> getCompositeJob(CommCareApplication.instance(), job.jobUUID) },
             )
 
-        fun startLearning(
-            user: ConnectUserRecord,
-            jobUUID: String,
-        ): Flow<DataState<Unit>> = networkOnlyFlow { networkClient.startLearnApp(user, jobUUID) }
+        fun startLearning(jobUUID: String): Flow<DataState<Unit>> =
+            networkOnlyFlow { networkClient.startLearnApp(getConnectUser(), jobUUID) }
 
-        fun claimJob(
-            user: ConnectUserRecord,
-            jobUUID: String,
-        ): Flow<DataState<Unit>> = networkOnlyFlow { networkClient.claimJob(user, jobUUID) }
+        fun claimJob(jobUUID: String): Flow<DataState<Unit>> = networkOnlyFlow { networkClient.claimJob(getConnectUser(), jobUUID) }
 
-        fun confirmPayments(
-            user: ConnectUserRecord,
-            paymentConfirmations: List<ConnectPaymentConfirmationModel>,
-        ): Flow<DataState<Unit>> = networkOnlyFlow { networkClient.confirmPayments(user, paymentConfirmations) }
+        fun confirmPayments(paymentConfirmations: List<ConnectPaymentConfirmationModel>): Flow<DataState<Unit>> =
+            networkOnlyFlow { networkClient.confirmPayments(getConnectUser(), paymentConfirmations) }
 
         /**
          * Emits Cached first,then Loading, then Success or Error after network call.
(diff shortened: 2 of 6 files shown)


app/unit-tests/src/org/commcare/connect/network/connect/ConnectNetworkClientTest.kt
```diff
@@ -214,4 +214,95 @@ class ConnectNetworkClientTest {
...
+        runBlocking {
+            val errorBody = "".toResponseBody("application/json".toMediaType())
+            coEvery { mockApiService.confirmPayments(any(), any(), any()) } returns Response.error(500, errorBody)
+            val result = client.confirmPayments(mockUser, emptyList())
+            assertTrue(result.isFailure)
+            assertEquals(
+                PersonalIdOrConnectApiErrorCodes.SERVER_ERROR,
+                (result.exceptionOrNull() as ConnectApiException).errorCode,
+            )
+        }
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3853/changes/8f3fe91559846209ffc911d7f978a4663070664a

app/src/org/commcare/connect/repository/ConnectRepository.kt
```diff
@@ -131,6 +148,14 @@ class ConnectRepository
...
             }.flowOn(Dispatchers.IO)
 
+        private fun <T> networkOnlyFlow(networkCall: suspend () -> Result<T>): Flow<DataState<T>> =
+            flow {
+                emit(DataState.Loading)
+                networkCall()
+                    .onSuccess { emit(DataState.Success(it)) }
+                    .onFailure { emit(DataState.Error.from(it)) }
+            }.flowOn(Dispatchers.IO)
+
```
Jignesh-dimagi, line 151: Right now, this call might runs within viewModelScope. If the ViewModel is destroyed before the network request completes, the coroutine will be cancelled. Since this operation needs a guaranteed execution (even if the user navigates away or the screen is destroyed), should we shift this to an application-level CoroutineScope instead?

app/src/org/commcare/connect/repository/ConnectRepository.kt
```diff
@@ -93,6 +95,21 @@ class ConnectRepository
                 mapToEmit = { _ -> getCompositeJob(CommCareApplication.instance(), job.jobUUID) },
             )
 
+        fun startLearning(
+            user: ConnectUserRecord,
```
Jignesh-dimagi: nit: For consistency with our other API calls, should we fetch ConnectUserRecord directly from the local DB (with a null-check) inside this repository layer, rather than passing it in as a parameter?

app/src/org/commcare/connect/network/connect/models/ConnectRequestModels.kt
```diff
@@ -0,0 +1,12 @@
...
+import com.google.gson.annotations.SerializedName
+
+data class ConfirmPaymentsRequest(
+    @SerializedName("payments") val payments: List<PaymentConfirmationBody>,
+)
+
+data class PaymentConfirmationBody(
+    @SerializedName("id") val id: String,
+    @SerializedName("confirmed") val confirmed: String,
+)
```
Jignesh-dimagi, line 12: Rather than keeping all request models in a single file, should we split them into individual files grouped by feature/functionality? This could improve code readability.

shubham1g5: Think I don't have strong thoughts here other than it might be an overkill at this stage, I do prefer it as it is unless it grows up to need a split (We similary also have the API calls for these requests in a single file)

app/src/org/commcare/connect/repository/ConnectRepository.kt
shubham1g5: > Since this operation needs a guaranteed execution (even if the user navigates away or the screen is destroyed),

I don't think these calls need guaranteed execution as these are synchronous calls unlike the `offlineFirstFlow` and it might be preferrable to cancel these if user decides to move away from the screen. That said, this work is refactor only and aims to not introduce any UX changes so the aim with this work is to have the same behaviour as earlier without having 2 parallel network stacks.

Jignesh-dimagi: Makes sense. I thought we were aligning this with our other API patterns to guarantee the call succeeds, but we can definitely address that refactoring in a follow-up ticket if required.

Jignesh-dimagi (dismissed):

app/src/org/commcare/connect/repository/ConnectRepository.kt
shubham1g5: https://github.com/dimagi/commcare-android/pull/3853/changes/3a978d1466a808dbbc63581f5e1afb0560065521

Jignesh-dimagi (approved):

PR #3863, opened 2026-08-11
QA-8627 Don't Write Connect Data After Signing Out Of PersonalID (1/2: gate the write sites)

OrangeAndGreen:
### [QA-8627](https://dimagi.atlassian.net/browse/QA-8627)

First of two PRs. This one gates the Connect write sites on the sign-in status; [#3860](https://github.com/dimagi/commcare-android/pull/3860) stacks the storage-layer hardening on top.

## Product Description

Signing out of PersonalID could leave the nav drawer showing "Logged out of PersonalID — A problem occurred, please configure your PersonalID account again", instead of the sign-in and register options. The error persisted for 24 hours or until dismissed. Sign-out now leaves the drawer in the correct signed-out state.

## Technical Summary

The trigger is a stale write: an in-flight Connect request (notification retrieval, release toggles) lands after signing out has deleted the Connect DB and its passphrase. `ConnectDatabaseHelper.getHandle()` then can't open the DB, flags `dbBroken` and raises `PERSONALID_GENERIC_ERROR`, which the exception handler persists as a `GlobalErrorRecord` for the drawer to display.

Connect storage is only valid while signed in, so the callers violating that contract now check before they write:

- `PushNotificationApiHelper.callPushNotificationApi` resumes with an empty list instead of storing retrieved notifications and acknowledging them.
- `ConnectReleaseTogglesParser` still parses the response and returns it, it just doesn't persist it.
- `MessagingChannelsKeySyncWorker` read Connect storage with no check at all, and now bails out.

Two supporting changes are what make those checks mean anything:

- `forgetUser` flips the status to `NotIntroduced` **before** deleting the DB rather than after. Previously a callback landing between the delete and the status change would pass its check and hit a deleted DB. Nothing between the two reads the DB, but the ordering is load-bearing.
- `personalIdSatus` is now `volatile`, since every one of these checks runs on a worker thread while sign-out writes the field from the UI thread.

### Safety story

What gives me confidence:

- The change is additive at each call site — early returns on a state that previously produced the bug. No behavior change while signed in.
- New tests cover the release-toggle guard in both directions, and the existing `PushNotificationActivityTest` suite passes against the new gate.
- The sign-out flow was exercised on device and the drawer shows the sign-in/register state as expected.

Risks to review:

- **A status check is check-then-act, so this narrows the window rather than closing it.** A caller already past its check when sign-out lands still reaches the deleted DB and still produces the banner. That interleaving is what the follow-up PR addresses; this PR covers the ordinary case, which is the whole in-flight duration of a request rather than the microseconds inside `forgetUser`.
- **The race is not directly reproduced by any test.** It needs a response to land inside a narrow window after sign-out. QA's repeated sign-out while notifications are actively arriving is the real check here.
- **Responses arriving while registering or recovering are now dropped too**, not just after sign-out. The Connect DB does exist in that state, so this discards data that could have been stored. Deliberate — the simpler `isloggedIn()` gate was preferred over tracking that case — but it is a real behavior change. `PushNotificationActivityTest`'s click-gating test had to be adjusted for it: it previously loaded notifications while registering, and now loads signed in and drops the login only for the click it actually covers.
- **This is a caller-by-caller fix, so coverage is only as complete as the list of callers.** Guarded here: notification retrieval, release toggles, messaging-channel key sync. The FCM notification raise in `NotificationsSyncWorker` is guarded in the follow-up. **Not guarded anywhere:** `ConnectOpportunitiesParser` (reachable from the same `NotificationsSyncWorker` via `syncOpportunities`) and the remaining `connectId` response parsers. Those can still land post-sign-out. `ConnectOpportunitiesParser` was left out because its tests assert `storeJobs` is always called and would all need updating; worth deciding whether to widen this or track it separately. Note that the follow-up PR makes the consequence non-fatal for all of them, guarded or not.

### Automated test coverage

`ConnectReleaseTogglesParserTest` gains two cases: toggles are stored while signed in, and are parsed but not persisted once signed out — the stale-write path this PR fixes.

The race window itself is not covered; the failure needs a response to arrive inside a narrow window after sign-out, which the existing test harness cannot schedule. The follow-up PR covers the interleaving with a latch-based concurrency test.

[QA-8627]: https://dimagi.atlassian.net/browse/QA-8627?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -14,6 +14,8 @@ This file is meant as an easy way for us to collate notes and change logs across
 
 - On a device where PersonalID sign-up previously got stuck due to a Firebase OTP error, a pre-invited user will now receive the OTP via Twilio automatically and can complete the sign-up flow.
 - Confirm that entering an incorrect verification code still displays the "incorrect code" error and does not silently trigger a new OTP.
+- After signing out of PersonalID, the nav drawer should offer sign-in/register rather than a "Logged out of PersonalID" error asking you to configure the account again. Worth repeating a few times, and while push notifications are actively arriving, since the original failure depended on timing.
+- Confirm signing back in still works and that notifications and messaging behave normally afterward.
 
 ## CommCare 2.63.4
 
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -14,9 +14,6 @@ This file is meant as an easy way for us to collate notes and change logs across
 
 - On a device where PersonalID sign-up previously got stuck due to a Firebase OTP error, a pre-invited user will now receive the OTP via Twilio automatically and can complete the sign-up flow.
 - Confirm that entering an incorrect verification code still displays the "incorrect code" error and does not silently trigger a new OTP.
-- After signing out of PersonalID, the nav drawer should offer sign-in/register rather than a "Logged out of PersonalID" error asking you to configure the account again. Worth repeating a few times, and while push notifications are actively arriving, since the original failure depended on timing.
-- Confirm signing back in still works and that notifications and messaging behave normally afterward.
-- The periodic notification retrieval, Connect heartbeat, and release toggle workers no longer restart from scratch on every biometric unlock, so they keep their existing schedule instead of firing immediately. Confirm notifications, heartbeat, and release toggles still update on a fresh sign-in and continue working across unlocks.
 
 ## CommCare 2.63.4
 
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -16,6 +16,7 @@ This file is meant as an easy way for us to collate notes and change logs across
 - Confirm that entering an incorrect verification code still displays the "incorrect code" error and does not silently trigger a new OTP.
 - After signing out of PersonalID, the nav drawer should offer sign-in/register rather than a "Logged out of PersonalID" error asking you to configure the account again. Worth repeating a few times, and while push notifications are actively arriving, since the original failure depended on timing.
 - Confirm signing back in still works and that notifications and messaging behave normally afterward.
+- The periodic notification retrieval, Connect heartbeat, and release toggle workers no longer restart from scratch on every biometric unlock, so they keep their existing schedule instead of firing immediately. Confirm notifications, heartbeat, and release toggles still update on a fresh sign-in and continue working across unlocks.
 
 ## CommCare 2.63.4
 
--- a/app/src/org/commcare/connect/PersonalIdManager.java
+++ b/app/src/org/commcare/connect/PersonalIdManager.java
@@ -138,9 +138,12 @@ public class PersonalIdManager {
                     TimeUnit.MILLISECONDS
             ).build();
 
+            // UPDATE rather than REPLACE: REPLACE cancels and re-enqueues, which makes the worker
+            // eligible to run immediately on every unlock. An in-flight run started that way can
+            // land after a subsequent sign-out and write to a deleted Connect DB.
             WorkManager.getInstance(CommCareApplication.instance()).enqueueUniquePeriodicWork(
                     CONNECT_HEARTBEAT_REQUEST_NAME,
-                    ExistingPeriodicWorkPolicy.REPLACE,
+                    ExistingPeriodicWorkPolicy.UPDATE,
                     heartbeatRequest
             );
         }
--- a/app/src/org/commcare/connect/PersonalIdManager.java
+++ b/app/src/org/commcare/connect/PersonalIdManager.java
@@ -85,7 +85,7 @@ public class PersonalIdManager {
     private BiometricManager biometricManager;
 
     private static volatile PersonalIdManager manager = null;
-    private PersonalIdStatus personalIdSatus = PersonalIdStatus.NotIntroduced;
+    private volatile PersonalIdStatus personalIdSatus = PersonalIdStatus.NotIntroduced;
     private Context parentActivity;
     private int failedPinAttempts = 0;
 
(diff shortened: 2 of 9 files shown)


Jignesh-dimagi:
@OrangeAndGreen 

Trigger on Unlock: Every successful biometric [unlock enqueues the Notification and Connect Toggle workers](https://github.com/dimagi/commcare-android/blob/192d1a9656d04ba7d8b954f034004f773cbd57e6/app/src/org/commcare/personalId/PersonalIdUnlocker.kt#L100).

The PeriodicWorkRequest is built with no initial delay, making it eligible to execute immediately upon enqueueing during the biometric unlock step.

When the user navigates to "Manage Profile" (after biometric unlock) and triggers "Forgot Personal ID", the app attempts to cancel these workers. However, NotificationsSyncWorker is a CoroutineWorker executing blocking I/O calls (withContext(Dispatchers.IO)). Because coroutine cancellation is cooperative, the blocking network/DB operations do not observe cancellation immediately and complete in-flight tasks anyway.

Reproducible: If the user presses "Forgot Personal ID" immediately after biometric unlock (in-flight worker task will cause this issue).

Non-Reproducible: If the user pauses long enough, on the PersonalId Profile screen, to allow in-flight calls to finish before pressing "Forgot Personal ID".

I think changing the ExistingPeriodicWorkPolicy should work here.

```
WorkManager.getInstance(context).enqueueUniquePeriodicWork(
                PERIODIC_NOTIFICATION_REQUEST_NAME,
                ExistingPeriodicWorkPolicy.REPLACE, --> cancel existing,  re-enqueues every time and immediate call
                retrievalRequest,
            )
```

to 

```
WorkManager.getInstance(context).enqueueUniquePeriodicWork(
                PERIODIC_NOTIFICATION_REQUEST_NAME,
                ExistingPeriodicWorkPolicy.UPDATE, --> don't cancel but call as per the scheduled time only
                retrievalRequest,
            )
```

Also, needs to be corrected for Connect's Heartbeat

CC @shubham1g5

shubham1g5:
@Jignesh-dimagi Agree that seems like a good change to make on this PR. Seems like we should also modify `ConnectReleaseTogglesWorker.Companion.schedulePeriodicFetch` to use `Update` instead of `keep` to be consistent here.

OrangeAndGreen:
Done in [e58f3ce](https://github.com/dimagi/commcare-android/pull/3863/commits/e58f3ce982f6b7ada3e53b1a58f38e1b493a1ee0)

- `NotificationsSyncWorkerManager.schedulePeriodicPushNotificationRetrieval` — `REPLACE` → `UPDATE`
- `PersonalIdManager.scheduleHeartbeat` (Connect heartbeat) — `REPLACE` → `UPDATE`
- `ConnectReleaseTogglesWorker.schedulePeriodicFetch` — `KEEP` → `UPDATE`

Added a QA note to `RELEASES.md` covering the scheduling change.

conroy-ricketts:
@OrangeAndGreen 

With these changes, I'm now seeing a new issue where forgetting PersonalID incorrectly drops me on a CC app home page.

Will share a video with you privately to avoid leaking my email address

RELEASES.md
```diff
@@ -14,6 +14,9 @@ This file is meant as an easy way for us to collate notes and change logs across
 
 - On a device where PersonalID sign-up previously got stuck due to a Firebase OTP error, a pre-invited user will now receive the OTP via Twilio automatically and can complete the sign-up flow.
 - Confirm that entering an incorrect verification code still displays the "incorrect code" error and does not silently trigger a new OTP.
+- After signing out of PersonalID, the nav drawer should offer sign-in/register rather than a "Logged out of PersonalID" error asking you to configure the account again. Worth repeating a few times, and while push notifications are actively arriving, since the original failure depended on timing.
+- Confirm signing back in still works and that notifications and messaging behave normally afterward.
+- The periodic notification retrieval, Connect heartbeat, and release toggle workers no longer restart from scratch on every biometric unlock, so they keep their existing schedule instead of firing immediately. Confirm notifications, heartbeat, and release toggles still update on a fresh sign-in and continue working across unlocks.
```
conroy-ricketts: Maybe we can remove these notes because QA has already tested the release and the QA ticket already provides a test case for them to verify the ticket. We can tag Nitin in the QA ticket directly if additional details are necessary

OrangeAndGreen:
> @OrangeAndGreen
> 
> With these changes, I'm now seeing a new issue where forgetting PersonalID incorrectly drops me on a CC app home page.
> 
> Will share a video with you privately to avoid leaking my email address

Ah, good find! I think this surfaces a new issue that we'll want to address, although I'm not sure whether to include it in this PR. All in all, I think the issue is that forgetting PersonalID never currently forces any app navigation but sometimes it should.

In the case of App Home, we should probably logout of the app and navigate to Login.
In the case of any other Connect page (Opp List, Messaging, Work History, etc.), we should just navigate back to Login (or Setup if no apps installed yet).

But automatically logging out of the app (the App Home case) could be risky, for instance if the user still has un-synced forms. I wonder if we should consider disabling the Forget button while in an app? Or we could leave it enabled and show the user a message telling them to logout of their app first if they try to forget their account while on App Home (thinking the disabled button could otherwise be confusing). @shubham1g5 I'm thinking you probably have good insight here.

Thinking a bit more, although maybe this gets fancier than we need... If we're on App Home for a non-Connect app at the time the user forgets PersonalID, we don't need to logout of the app.

Finally, flagging that this is a separate issue unrelated to and currently unaffected by this PR, so we could consider merging these changes and tackling the additional issue in a new ticket.

conroy-ricketts (approved):
Thanks for the context, and agreed, this fix is time-sensitive

conroy-ricketts:
We may want to provide some of that context in the ticket as well when this gets merged so that QA is not caught off guard

RELEASES.md
OrangeAndGreen: Yeah, makes sense: [c9adc91](https://github.com/dimagi/commcare-android/pull/3863/commits/c9adc918783b5490de240d7ce1d29d1522067075)

Jignesh-dimagi (approved):

shubham1g5:
> , for instance if the user still has un-synced forms. I wonder if we should consider disabling the Forget button while in an app? 

Ahh think that's alright, an app can have un-synced forms even where user is logged out. And technically it'd be possible for user to recover that data if they sign back into PersonalID account on the same app installation. It'd be nice for us to warn user when they click Forget but think that's most we should do here.

PR #3864, opened 2026-08-11
QA-8628 Remove The App List From The Nav Drawer

conroy-ricketts:
## Product Description

The side navigation drawer no longer lists the installed CommCare apps. Tapping **CommCare Apps** now asks for confirmation and, on confirm, signs the user out of the current app and returns them to the login screen, where the existing app picker handles choosing another app. On the login and setup screens, where there is nowhere to navigate, the row just dismisses the drawer.

This replaces #3857, which fixed the reported bug by repairing the in-drawer switch. Removing the list resolves the same ticket by removing the entry point.

https://github.com/user-attachments/assets/516e1e66-a4ac-4d1b-bded-c9c4b2738803

## Technical Summary

Reaching the login screen means ending the session, so the row goes through `DispatchActivity` rather than Home's usual `setResult` + `finish`: that result is only read when Dispatch started Home for result, which is not true of a Connect-launched Home nor of one brought forward with `FLAG_ACTIVITY_REORDER_TO_FRONT`. No `FLAG_ACTIVITY_CLEAR_TOP`, so `ConnectActivity` stays in the stack and backing out of the login screen still lands on the opportunity list.

The sync-lock guard is now shared with `userTriggeredLogout` instead of duplicated.

With the app list gone, nothing produced `EXTRA_APP_ID` as an activity result, so `DispatchActivity.redirectToLoginAppId` and its `launchLoginScreen` fallback were dead and are removed. The apps entry was also the only user of the drawer's parent/child machinery, so `NavDrawerItem` collapses to one data class and `NavDrawerAdapter` to a single view type.

### Safety story

What gives confidence:
- Tested manually on a physical device and confirmed to work as expected.
- Traditional CommCare users are unaffected: Home only builds this drawer when PersonalID is logged in (`shouldShowDrawerAfterCheck(true)`) and `refreshDrawerContent` hides the list otherwise, so their app switching was already the login screen's picker and is unchanged.
- Multi-app users keep a route to every app — `LoginActivityUIController.refreshView` shows the app spinner whenever two or more usable apps exist.
- Largely a deletion: 128 insertions against 290 deletions, and the removed adapter paths had no test coverage to lose.
- `ExternalLaunchTests`, `PersonalIdDrawerVisibilityTest`, `RecoveryMeasuresTest`, `SessionEndpointNotificationTest`, `DemoUserRestoreTest` and `SmsInviteLinkFlowTest` pass.

Risks to review:
- `NavDrawerAdapter` and `nav_drawer_list_item.xml` are shared by every drawer surface — Home, Login and Setup — so the row rendering is worth a look on each, signed in and signed out.
- Switching apps now costs a confirmation and a full re-login rather than a tap, which is a real workflow change for anyone who used the list often.
- Four new strings are translated in nine locales; `values-am` and `values-ne` carry only `arrays.xml` and were skipped, so those locales fall back to English for the dialog.

### Automated test coverage

None added. Existing Robolectric coverage over `DispatchActivity` and the drawer was run for regressions.

[QA-8628]: https://dimagi.atlassian.net/browse/QA-8628?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


conroy-ricketts:
## Suggested Review Order

- `app/src/org/commcare/navdrawer/BaseDrawerController.kt` — the app list and seated-app highlight come out of the drawer content here
- `app/src/org/commcare/activities/StandardHomeActivity.java` — the new row behavior: confirm, end session, hand off to Dispatch
- `app/src/org/commcare/navdrawer/BaseDrawerActivity.kt` — the base branch every other drawer surface gets
- `app/src/org/commcare/activities/HomeScreenBaseActivity.java` — sync guard extracted so logout and this path share it
- `app/src/org/commcare/activities/DispatchActivity.java` — dead app-id plumbing removed
- `app/src/org/commcare/navdrawer/NavDrawerItem.kt`, `NavDrawerAdapter.kt` and the layout changes — mechanical fallout of losing child rows

app/src/org/commcare/activities/HomeScreenBaseActivity.java
```diff
@@ -608,17 +608,24 @@ protected void goToFormArchive(boolean incomplete, FormRecord record) {
...
                     Localization.get("background.sync.logout.attempt.during.sync"),
                     Toast.LENGTH_LONG
             ).show();
-            return;
+            return true;
         }
-        CommCareApplication.instance().closeUserSession();
-        setResult(RESULT_OK);
-        finish();
+        return false;
```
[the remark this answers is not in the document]
conroy-ricketts: **[AI] Claude Opus 5 (1M context)**

The race is real, but it predates this PR and the suggested fix would be worse than the defect. Not changing it here.

**On the race.** Confirmed: `sessionAliveLock` has two protocols in this codebase. The automatic paths acquire it properly — `DataPullTask.doTaskBackground` does `tryLock()` and aborts the sync when logout holds it, and the scheduled session expiry in `CommCareSessionService` does the same. The user-initiated paths only *observe* it with `isLocked()` and never acquire it, as do `HomeButtons` and `FirebaseMessagingDataSyncer`, and `CommCareApplication.closeUserSession()` never touches it. So the check-then-act window is genuine.

**On this PR's role.** That window is verbatim what `userTriggeredLogout()` already did; this change extracts the check into `isBlockedByActiveSync()` and adds a second caller with identical behavior. The protocol is unchanged. Worth stating plainly, though: a drawer row is more discoverable than the logout menu item, so exposure does go up somewhat even if the mechanism doesn't.

**On the proposed fix.** "Acquire and hold `sessionAliveLock` through `closeUserSession()`" read literally means a blocking `lock()` — and both call sites run on the UI thread from click handlers. A data pull can run for minutes, so that converts a rare race into a reliable ANR. That is almost certainly why the original author chose an advisory check plus a toast.

**What would actually work**, for whoever picks this up: `tryLock()` rather than `lock()`, wrapping `closeUserSession()` with the release in a `finally`. Same fail-fast toast, no blocking, and it matches what the two existing lock holders already do. No deadlock risk — `cancelTasks()` does not join, and the lock is reentrant.

Deliberately out of scope here: this is a release-branch change whose QA plan covers a nav drawer row, and the fix lands in the logout path every CommCare user hits — including the traditional users this PR otherwise leaves entirely untouched. Verifying it means staging a long background sync and logging out underneath it, which is a QA pass this ticket won't get.

app/res/values/strings.xml
```diff
@@ -795,6 +795,10 @@
     <string name="nav_drawer_not_signed_in_to_personal_id">You are not signed in to PersonalID</string>
     <string name="nav_drawer_opportunities">Opportunities</string>
     <string name="nav_drawer_commcare_apps">CommCare Apps</string>
+    <string name="nav_drawer_switch_app_dialog_title">Sign out of this app?</string>
```
shubham1g5: can we say "Log out" instead of "Sign out" for consistency.

app/src/org/commcare/activities/StandardHomeActivity.java
```diff
@@ -288,6 +277,36 @@ protected void handleDrawerItemClick(@NonNull BaseDrawerController.NavItemType i
...
+    private void returnToLogin() {
+        if (isBlockedByActiveSync()) {
+            return;
+        }
+        CommCareApplication.instance().closeUserSession();
+        Intent intent = new Intent(this, DispatchActivity.class);
+        intent.putExtra(LoginActivity.USER_TRIGGERED_LOGOUT, true);
+        startActivity(intent);
+        finish();
+    }
```
shubham1g5, line 310: Thoughts on calling `userTriggeredLogout()` directly from here ? I am finding it hard to distinguish how and why should the 2 logouts (here and by clicking log out button)  be different.

app/src/org/commcare/activities/DispatchActivity.java
```diff
@@ -109,6 +109,12 @@ protected void onCreate(Bundle savedInstanceState) {
             waitingForActivityResultFromLogin = savedInstanceState.getBoolean(
                     KEY_WAITING_FOR_ACTIVITY_RESULT
             );
+            userTriggeredLogout = savedInstanceState.getBoolean(KEY_USER_TRIGGERED_LOGOUT);
+        } else {
+            userTriggeredLogout = getIntent().getBooleanExtra(
+                    LoginActivity.USER_TRIGGERED_LOGOUT,
+                    false
+            );
```
shubham1g5, line 117: Curious what's triggering these changes ? Mostly surprised why we need them given we do not need it for normal App logout ?

shubham1g5 (changes requested):

app/src/org/commcare/activities/StandardHomeActivity.java
```diff
@@ -288,6 +277,36 @@ protected void handleDrawerItemClick(@NonNull BaseDrawerController.NavItemType i
...
+
+    private void returnToLogin() {
+        if (isBlockedByActiveSync()) {
+            return;
+        }
+        CommCareApplication.instance().closeUserSession();
+        Intent intent = new Intent(this, DispatchActivity.class);
+        intent.putExtra(LoginActivity.USER_TRIGGERED_LOGOUT, true);
+        startActivity(intent);
+        finish();
```
Jignesh-dimagi, line 309: Since we are already explicitly closing the user session, is it necessary to propagate USER_TRIGGERED_LOGOUT all the way up to LoginActivity?

LoginActivity should handle session termination automatically without needing a specific event code, especially since we do not want to trigger an auto-login retry.

app/src/org/commcare/activities/StandardHomeActivity.java
conroy-ricketts: The idea here is that a logout from the "Log out" button (on the CC app home page) should return to the Opp List page rather than the Login page for Connect users.

I wanted to avoid modifying `userTriggeredLogout()` because it sets the result and finishes the activity, leading the user to whatever started the Home page. So for a normal login, that routes the user through `DispatchActivity`, landing on the Login page. For a Connect-launched login, nothing reads the result, so the user is taken back to `ConnectActivity`, landing on the Opp List page. 

Long story short - this is the behavior that we want to avoid for this QA ticket. I'm going to add a short comment to the function regarding this

app/src/org/commcare/activities/DispatchActivity.java
conroy-ricketts: We don't need it for normal logout because it returns a result to `DispatchActivity` that ultimately sets `userTriggeredLogout` to `true`, so the flag never has to "travel" between activities.

So, these changes are here because we are rather starting a fresh `DispatchActivity` rather than returning a result

app/src/org/commcare/activities/StandardHomeActivity.java
conroy-ricketts: Yeah, `LoginActivity` should not need anything to handle a closed session, but it seems that if we do not pass this flag, then `LoginActivity` will not stop the auto-login retry. 

Auto-logins are gated on the `USER_TRIGGERED_LOGOUT` flag [here](https://github.com/dimagi/commcare-android/blob/d53f4091630d3237a2cde59c4813adac2a875af0/app/src/org/commcare/activities/LoginActivity.java#L466).

Rather than potentially auto-logging the user into a CC app, we would like to have them land on the Login page so they are able to choose a different app from there

app/res/values/strings.xml
conroy-ricketts: Yes! And I'll update the video in the PR too

app/src/org/commcare/activities/StandardHomeActivity.java
conroy-ricketts: e425c4f34825a02e9f378c4728dc7f45cc5a4ae3

app/res/values/strings.xml
conroy-ricketts: e425c4f34825a02e9f378c4728dc7f45cc5a4ae3

app/src/org/commcare/activities/StandardHomeActivity.java
Jignesh-dimagi: Yeah, that makes sense. To avoid code duplication in this activity, could we pass an nullable targetActivity to HomeScreenBaseActivity.userTriggeredLogout()? If a target activity is provided, we start it; otherwise, we just call finish() with RESULT_OK.

app/src/org/commcare/activities/StandardHomeActivity.java
Jignesh-dimagi: Yeah, I'm seeing that same behavior on the master branch as well.

shubham1g5 (approved):

shubham1g5:
Merging this as we are trying to finalize QA today, any followups to this can be made separately.

PR #3867, opened 2026-08-13
Handle Connect DB Not Found Error

shubham1g5:
## Product Description

https://dimagi.atlassian.net/browse/QA-8627

## Technical Summary

Decided to do an alternate implementation on https://github.com/dimagi/commcare-android/pull/3860 to save time with PR reviews. 

The main change here is that we are explicitly checking for DB file existance before trying to access the DB and throwing a dedicated exception for different callers to helper. This error message doesn't really end up showing in my testing given the calls that were happening at the time of Forget were all background calls with no mechanism to show error on UI. 

### Safety story

- Tested the Forget flow a few times and could not reproduce the issue anymore. 
- Think the biggest risk here is if there is a code bug that ends up deleting the DB, we won't crash hard but ask user to sign in again. 

## Labels and Review

### DIFF
--- a/app/res/values-no/strings.xml
+++ b/app/res/values-no/strings.xml
@@ -227,7 +227,7 @@
     <string name="personalid_biometric_error_lockout_permanent">For mange mislykkede forsøk. Lås opp enheten og prøv igjen.</string>
     <string name="personalid_biometric_error_timeout">Biometrisk operasjon ble tidsavbrutt. Kontakt kundestøtte hvis problemet vedvarer.</string>
     <string name="personalid_biometric_error_cannot_configure">Vi kunne ikke konfigurere biometri på enheten din. Kontakt kundestøtte hvis problemet vedvarer.</string>
-    <string name="personalid_no_connect_database_error">Ingen PersonalID-konto er til stede, vennligst logg inn på nytt til PersonalID-kontoen din og prøv igjen.</string>
+    <string name="personalid_no_connect_database_error">Ingen PersonalID-konto er til stede. Logg inn på PersonalID-kontoen din på nytt, og prøv igjen.</string>
     <string name="personalid_work_history_no_records_available">Ingen arbeidslogger er tilgjengelige for øyeblikket. Klikk på Oppdater, eller kom tilbake senere.</string>
     <string name="pn_no_notification_available">Ingen varsler tilgjengelig</string>
     <string name="connect_no_jobs">Du har ikke blitt invitert til noen muligheter ennå. Når du har blitt invitert til en mulighet, vil den vises i denne listen.</string>
--- a/app/res/values-no/strings.xml
+++ b/app/res/values-no/strings.xml
@@ -227,6 +227,7 @@
     <string name="personalid_biometric_error_lockout_permanent">For mange mislykkede forsøk. Lås opp enheten og prøv igjen.</string>
     <string name="personalid_biometric_error_timeout">Biometrisk operasjon ble tidsavbrutt. Kontakt kundestøtte hvis problemet vedvarer.</string>
     <string name="personalid_biometric_error_cannot_configure">Vi kunne ikke konfigurere biometri på enheten din. Kontakt kundestøtte hvis problemet vedvarer.</string>
+    <string name="personalid_no_connect_database_error">Ingen PersonalID-konto er til stede, vennligst logg inn på nytt til PersonalID-kontoen din og prøv igjen.</string>
     <string name="personalid_work_history_no_records_available">Ingen arbeidslogger er tilgjengelige for øyeblikket. Klikk på Oppdater, eller kom tilbake senere.</string>
     <string name="pn_no_notification_available">Ingen varsler tilgjengelig</string>
     <string name="connect_no_jobs">Du har ikke blitt invitert til noen muligheter ennå. Når du har blitt invitert til en mulighet, vil den vises i denne listen.</string>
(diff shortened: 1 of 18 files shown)


app/res/values-no/strings.xml
```diff
@@ -227,6 +227,7 @@
     <string name="personalid_biometric_error_lockout_permanent">For mange mislykkede forsøk. Lås opp enheten og prøv igjen.</string>
     <string name="personalid_biometric_error_timeout">Biometrisk operasjon ble tidsavbrutt. Kontakt kundestøtte hvis problemet vedvarer.</string>
     <string name="personalid_biometric_error_cannot_configure">Vi kunne ikke konfigurere biometri på enheten din. Kontakt kundestøtte hvis problemet vedvarer.</string>
+    <string name="personalid_no_connect_database_error">Ingen PersonalID-konto er til stede, vennligst logg inn på nytt til PersonalID-kontoen din og prøv igjen.</string>
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-android/pull/3867/commits/17841daa8b4737bb2302b19cd7e96a5679576408

app/src/org/commcare/connect/database/ConnectDatabaseHelper.java
```diff
@@ -53,10 +53,16 @@ public IDatabase getHandle() {
...
                             connectDatabase = CommCareApplication.instance().getConnectDbOpenHelper(context);
                         } catch (Exception e) {
-                            //Flag the DB as broken if we hit an error opening it (usually means corrupted or bad encryption)
-                            dbBroken = true;
-                            Logger.exception("Error opening Connect DB", e);
-                            GlobalErrorUtil.triggerGlobalError(GlobalErrors.PERSONALID_GENERIC_ERROR);
+                            if (!dbExists()) {
+                                ConnectDatabaseNotFoundException dbNotFound = new ConnectDatabaseNotFoundException();
+                                Logger.exception("Error opening Connect DB", dbNotFound);
+                                throw dbNotFound;
```
Jignesh-dimagi, line 59: Currently, ConnectDatabaseNotFoundException is only handled at the BaseApi layer under the assumption that it originates from API calls only. However, if it gets thrown during DB operations ([like in this case](https://github.com/dimagi/commcare-android/blob/62ea4630955842e0f748463f41ae0891ca2f38e8/app/src/org/commcare/connect/database/ConnectJobUtils.java#L48)), it will have not exception catching and crash the app. Should we add this exception at [CommCareExceptionHandler](https://github.com/dimagi/commcare-android/blob/62ea4630955842e0f748463f41ae0891ca2f38e8/app/src/org/commcare/utils/CommCareExceptionHandler.java#L30).

Jignesh-dimagi (changes requested):

app/src/org/commcare/connect/database/ConnectDatabaseHelper.java
shubham1g5: Think that would make sense to me, but I am not clear around what would we do after catching it ? Should we still handle it as a global error or something else ?

In general, I would not expect this to happen for any Synchronous calls and think API calls are probably the largest surface area this issue would happen. So think it might still be ok crashing in other scenarios with the assumption that we don't expect those crashes to surface in other places, although I would definitely be interested in seeing this bubble in Crashlytics to decide if we need to handle it at other places in the code and with what priority.

shubham1g5: Actually I think I would not want us to handle it as Global error as that would be equal to hiding this issue under global error umbrella (which was the earlier behaviour here as well and think why we didn't knew about this issue sooner).

Jignesh-dimagi: I agree with your reasoning here. I think we can move ahead with this.

Jignesh-dimagi (dismissed):

app/src/org/commcare/connect/database/ConnectDatabaseHelper.java
OrangeAndGreen: I'm a little confused as to what the difference would be since a global error crashes the app anyway. Is it just a matter of whether we show an error message when the app restarts after crashing?

shubham1g5: I didn't realise that global error always crashes the app. I think majorly I would like to see this crash appearing in Crashlytics as a standalone crash rather than being burried into some other error factor and I am not sure if that's the behaviour with Global errors.

OrangeAndGreen: Confirmed, a LoginInvalidatedException gets thrown from [here](https://github.com/dimagi/commcare-android/blob/d53f4091630d3237a2cde59c4813adac2a875af0/app/src/org/commcare/utils/GlobalErrorUtil.java#L20) and crashes the app. I put together a query [here](https://console.cloud.google.com/bigquery?ws=!1m7!1m6!12m5!1m3!1scommcare-a57e4!2sus-central1!3s16e8acb5-c5b4-4593-b22b-6207f484fcda!2e1) showing recent occurrences of four possible reasons that are still being observed in the recent past (the ...DB_UPGRADE_ERROR case hasn't occurred recently).

My take is that almost all of these errors (excluding ...LOGIN_FROM_DIFFERENT_DEVICE) are ones that should never happen and we should treat with equal importance as bugs. But given that the excluded case is one we know will keep happening for the foreseeable future, maybe it's worth separating out LoginInvalidatedExceptions that are user-caused vs. bugs so we can have better visibility of "the bad ones".

PR #3869, opened 2026-08-14
CCCT-2518 Redesign The Learn In Progress Screen

Jignesh-dimagi:
### [CCCT-2518](https://dimagi.atlassian.net/browse/CCCT-2518)

Normal Learn In-Progress:

https://github.com/user-attachments/assets/1e4c5137-0fb1-40bd-a22b-eb5b2a9f27cc

Learn Assessment Failure State

https://github.com/user-attachments/assets/48fc975f-3f89-4995-9f87-daf07468579d

Regardless of the sequence completed, the UI dynamically updates to highlight and prompt the user with the next pending module that needs to be completed:

https://github.com/user-attachments/assets/f8a67364-9eb8-4e15-bc03-afb19e74abf0

## Product Description

The learning screen for an in-progress opportunity now follows the redesign: an opportunity header, a module progress card, and a "Continue Learning" card naming the module to do next. A failed assessment shows a retry banner tucked behind the progress card.

## Technical Summary

A worker can finish modules in any order, so the screen now matches completion records to modules by the server's module id instead of counting how many are done — finishing module 2 first used to make the screen point back at module 2. That id was never stored, which is why there is a schema change; modules already saved on a device have no id until the next opportunities sync, and fall back to the old counting behaviour in the meantime.
The edge-to-edge changes for Android 15.

### Safety story

*What gives confidence*

- I ran the change on a real device and exercised the learn-in-progress screens.
- The new view is covered by Robolectric tests for all three learning states, including the out-of-order and unsynced-id paths.
- The screen is reached through an existing navigation destination; no new entry points or API calls.

*Risks to review*

- Two things disappear from this screen with no replacement, both deliberate: the way in to the job detail bottom sheet, and the "relearn tasks pending" warning.

### Automated test coverage

13 new Robolectric tests for `ConnectLearnProgressView` cover the three learning states plus out-of-order completion, the missing-id fallback and the no-module-records fallback; two of them pin the same input to different expected modules, so a regression back to count-based indexing cannot pass silently. The fragment test gains coverage of the progress CTA routing to the learn download.

[CCCT-2518]: https://dimagi.atlassian.net/browse/CCCT-2518?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Jignesh-dimagi:
## Suggested Review Order

The four commits tell the story in order, so commit-by-commit works well here. By file:

- `ConnectLearnModuleSummaryRecord.java` — new server module id field and its JSON key
- `ConnectLearnModuleSummaryRecordV28.kt` — pre-change record shape, read only by the migration
- `ConnectDatabaseUpgrader.java` / `ConnectDatabaseSchemaManager.kt` — the 28 → 29 migration
- `ConnectJobRecord.java` — `modulesRemaining()`, the predicate the new screen keys off
- `ConnectProgressCard.kt` / `view_connect_progress_card.xml` — padding removed, elevation now applied only while the banner shows
- `ConnectLearnProgressView.kt` — the state logic; `continueModule` is the part worth reading closely
- `view_connect_learn_progress.xml` — the redesigned layout
- `ConnectLearningProgressFragment.kt` — Kotlin conversion; now only picks a view and navigates
- `fragment_connect_learning_progress.xml` — old layout replaced by the two views
- `ConnectLearnProgressViewTest.kt` — one test per state, plus the out-of-order and fallback cases
- `ConnectLearningProgressFragmentTest.kt` — CTA lookups scoped per view, new progress-CTA navigation test
- strings, dimens, drawable — mechanical

conroy-ricketts:
Hi @Jignesh-dimagi, can we remove the requests for review and tag @dimagi/connect-mobile-devs instead? 

When tagging the team (i.e. @dimagi/connect-mobile-devs) for review, GitHub will automatically choose 2 people for review round-robin style. The hope is that this will help reduce the code review burden on the team

app/src/org/commcare/android/database/connect/models/ConnectLearnModuleSummaryRecord.java
```diff
@@ -17,6 +17,9 @@ public class ConnectLearnModuleSummaryRecord extends Persisted implements Serial
      */
     public static final String STORAGE_KEY = "connect_learn_modules";
 
+    public static final String META_MODULE_ID = "module_id";
+    /** Key the server sends the module id under; distinct from the local storage column. */
+    private static final String META_SERVER_ID = "id";
```
shubham1g5: Only DB fields should be defined as `META_` under this class.

app/src/org/commcare/activities/connect/ConnectActivity.java
```diff
@@ -108,6 +110,12 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
     public void onPostCreate(@Nullable Bundle savedInstanceState) {
         super.onPostCreate(savedInstanceState);
         getSupportActionBar().setTitle(getString(R.string.connect_title));
+
+        // The bottom inset is left to BaseConnectFragment, which knows whether the screen's bottom
+        // bar spans the navigation bar or has to stay clear of it.
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.VANILLA_ICE_CREAM) {
+            AndroidUtil.attachWindowInsetsListener(this, android.R.id.content, false);
+        }
```
shubham1g5: this already happens in `CommonBaseActivity` (base class for this activity)

shubham1g5: nit: think a cleaner way to do this would be to abstract the insets code in `CommCareBaseActivity` as `applyInsets()` and then override it here with this so that we don't end up applying insets twice (Assuming it causes view to redraw or re-measure and might be expensive)

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
```diff
@@ -117,6 +120,13 @@ abstract class BaseConnectFragment<B : ViewBinding> :
         )
         rootFrame.addView(verticalContainer)
 
+        // Only from Android 15, where edge-to-edge is enforced and ConnectActivity hands the bottom
+        // inset down so a bar flush with the screen edge can span the navigation bar. Everything else
+        // keeps clear of it. Earlier releases lay the window out inside the system bars already.
```
shubham1g5: Curious why are we applying insets at 2 places - activity and fragment, for eg. why not move the whole logic to fragment for both bottom and window insets so that code around insets is not distributed at multiple places and we do not end up re-applying insets repeatedly on the same screen.

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
```diff
@@ -145,6 +155,75 @@ abstract class BaseConnectFragment<B : ViewBinding> :
...
+    /**
+     * Bars sitting flush with the screen edge, which absorb the inset into their own padding. A bar
+     * laid out with a bottom margin already floats clear of the navigation bar, so it is left alone
+     * and the inset is reserved above the content instead.
+     *
+     * Hidden bars are included deliberately. Insets arrive on attach, not when a view's visibility
+     * changes, so skipping them would leave a bar shown later — the learn-complete state replacing
+     * the in-progress one — drawing its button under the navigation bar.
+     */
+    private fun flushBottomBars(view: View): List<BottomInsetAbsorber> =
```
shubham1g5: can we call it `getBottomBars` , `flushBottomBars` seems like we are deleting these .

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
```diff
@@ -145,6 +155,75 @@ abstract class BaseConnectFragment<B : ViewBinding> :
...
+            }
+
+            view is ViewGroup -> {
+                (0 until view.childCount).flatMap { flushBottomBars(view.getChildAt(it)) }
+            }
+
+            else -> {
+                emptyList()
+            }
+        }
```
shubham1g5: To consider: I think instead of making it generic and doing a whole view heirarchy pass, it would be better that if we directly define a hardcoded method at each fragment level `getViewsForBottomInsets()` that return a hardcoded list of views (think it's just CTA bar at the moment) . 

````
getViewsForBottomInsets(): List<Views> {
return listOf(ctaBar)
}
````

File: app/src/org/commcare/fragments/base/BaseConnectFragment.kt [hunk 6]
```diff
@@ -145,6 +155,75 @@ abstract class BaseConnectFragment<B : ViewBinding> :
...
+        ViewCompat.setOnApplyWindowInsetsListener(container) { view, insets ->
+            if (!isNestedConnectFragment()) {
+                val bars =
+                    insets
+                        .getInsets(
+                            WindowInsetsCompat.Type.systemBars() or
+                                WindowInsetsCompat.Type.displayCutout(),
+                        ).bottom
+                val ime = insets.getInsets(WindowInsetsCompat.Type.ime()).bottom
+                val flushBars = flushBottomBars(content)
```
shubham1g5: think this should be called `bottomInsetAbsorbers` to be consistent.

shubham1g5 (changes requested):
Think we are mixing 2 different code components here - Insets and Learn Progress implementation and it would have been great for these to be 2 seaprate PRs.

app/src/org/commcare/android/database/connect/models/ConnectLearnModuleSummaryRecord.java
Jignesh-dimagi: Done

Jignesh-dimagi:
> Think we are mixing 2 different code components here - Insets and Learn Progress implementation and it would have been great for these to be 2 seaprate PRs.

I've removed the edge-to-edge changes (including the commit history) from this PR. I originally included them so I could showcase the Learn In-Progress feature in the tech-demos channel, but I've decoupled them now to keep this PR focused.

I'll be opening a separate PR specifically for the Connect edge-to-edge implementation shortly!

app/src/org/commcare/views/connect/ConnectLearnProgressView.kt
```diff
@@ -0,0 +1,207 @@
...
+        /**
+         * Once every module is done the card names the most recently completed one; while modules
+         * remain it names the first unfinished one. Both are resolved through the server ids on the
+         * completion records, since modules can be completed in any order.
+         *
+         * Modules stored before that id was persisted all carry 0, which matches no completion and
+         * would pin the card to the first module. Those fall back to the completed count — the
+         * previous, order-assuming behaviour — until the next sync repopulates the ids.
+         */
+        private fun continueModule(
```
shubham1g5: Can we name methods with a verb rather than a noun for eg. `findContinueModule`  (the issue is present at several places in the PR and do not align with how we name methods in code today)

Also this method should not be inside a view given it's business logic for the app. 

Same comments for `lastCompletedModule` as well

app/src/org/commcare/views/connect/ConnectLearnProgressView.kt
```diff
@@ -0,0 +1,207 @@
...
+            job: ConnectJobRecord,
+            modules: List<ConnectLearnModuleSummaryRecord>,
+        ): ConnectLearnModuleSummaryRecord {
+            if (!job.modulesRemaining()) {
+                return lastCompletedModule(job, modules) ?: modules.last()
+            }
+
+            if (modules.any { it.moduleId == ConnectLearnModuleSummaryRecord.UNKNOWN_MODULE_ID }) {
```
shubham1g5: Do we see modules without id in the payload today or it's just a hypothetical precaution ? 

Also think it's better to not show the next module at all in these cases rather than showing a random module which doesn't adhere the intended logic for showing the next module.

app/src/org/commcare/fragments/connect/ConnectLearningProgressFragment.java
```diff
@@ -1,340 +0,0 @@
...
-
-public class ConnectLearningProgressFragment extends ConnectJobFragment<FragmentConnectLearningProgressBinding>
-        implements RefreshableFragment {
-
-    private boolean showAppLaunch = true;
-    private ConnectLearningProgressViewModel viewModel;
-
-    public static ConnectLearningProgressFragment newInstance(boolean showAppLaunch) {
-        ConnectLearningProgressFragment fragment = new ConnectLearningProgressFragment();
-        fragment.showAppLaunch = showAppLaunch;
```
shubham1g5, line 48: I don't see this variable in Kotlin implementation, is it not required anymore ? If so, should we remove it from fragment callers as well ?

app/src/org/commcare/android/database/connect/models/ConnectLearnModuleSummaryRecord.java
```diff
@@ -26,6 +27,18 @@ public class ConnectLearnModuleSummaryRecord extends Persisted implements Serial
...
+     * Key the server sends the module id under, distinct from the {@link #META_MODULE_ID} column it
+     * is stored in. Not a database field, so it is not a {@code META_} constant.
+     */
+    private static final String SERVER_ID = "id";
+
+    /**
+     * Stands in for a module whose server id is unavailable: it predates the id being persisted, or
+     * the payload omitted it. Consumers fall back to the completed-module count.
+     */
+    public static final int UNKNOWN_MODULE_ID = 0;
```
shubham1g5, line 40: Are we sure that server can never assign `0` as a module id ? Otherwise we might have a collision here.

app/src/org/commcare/views/connect/ConnectLearnProgressView.kt
```diff
@@ -0,0 +1,207 @@
...
+                ),
+            )
+        }
+
+        /**
+         * The "Continue Learning" card points at a module — the next unfinished one while modules
+         * remain, or the last completed one after a failed attempt — and otherwise at the
+         * assessment, which is also the fallback when no module record is available to name.
+         */
+        private fun bindContinueCard(job: ConnectJobRecord) {
```
shubham1g5, line 86: nit: It'd be nice to create a state object to manage methods like this better - 

````
sealed interface LearnState {                                                                                                                     
      object ModulesInProgress                                                                                                    
      object AssessmentPending 
      object AssessmentFailed                                                                                                          
  }
````

**Review by shubham1g5 — CHANGES_REQUESTED** [review]

File: app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java [hunk 12]
```diff
@@ -534,6 +534,13 @@ public int getDeliveryProgressPercentage() {
         return total > 0 ? (100 * completed / total) : 100;
     }
 
+    /**
+     * Whether any learning module is still unfinished.
+     */
+    public boolean modulesRemaining() {
```
OrangeAndGreen: Nit: The name modulesRemaining sounds like it would return a number at first glance. Also, I noticed in over half of the usages of this function we negate the result. So what if instead we call the function completedAllModules and flip the logic?

app/src/org/commcare/android/database/connect/models/ConnectLearnModuleSummaryRecord.java
```diff
@@ -67,6 +89,9 @@ public static ConnectLearnModuleSummaryRecord fromJson(JSONObject json, int modu
         info.name = json.getString(META_NAME);
         info.description = json.getString(META_DESCRIPTION);
         info.timeEstimate = json.getInt(META_ESTIMATE);
+        // Optional: a module the server sends without an id is still usable, so it degrades to the
```
OrangeAndGreen: Is this an issue we've observed happening? I would expect the server to always include the module ID, but I could be wrong

OrangeAndGreen (dismissed):
Left a couple small comments but otherwise things look pretty good to me pending conflict resolution and a couple of Shubham's comments

app/src/org/commcare/activities/connect/ConnectActivity.java
Jignesh-dimagi: This will be taken care in Android 15's edge to edge case [task](https://dimagi.atlassian.net/browse/CCCT-2778).

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
Jignesh-dimagi: This will be taken care in Android 15's edge to edge case [task](https://dimagi.atlassian.net/browse/CCCT-2778).

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
Jignesh-dimagi: This will be taken care in Android 15's edge to edge case [task](https://dimagi.atlassian.net/browse/CCCT-2778).

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
Jignesh-dimagi: This will be taken care in Android 15's edge to edge case [task](https://dimagi.atlassian.net/browse/CCCT-2778).

app/src/org/commcare/fragments/base/BaseConnectFragment.kt
Jignesh-dimagi: This will be taken care in Android 15's edge to edge case [task](https://dimagi.atlassian.net/browse/CCCT-2778).

app/src/org/commcare/views/connect/ConnectLearnProgressView.kt
Jignesh-dimagi: Since workers can complete modules out of order, the UI now matches completion records using the server's module_id. Previously, finishing Module 2 out of order would cause the screen to incorrectly point back to Module 2 as the next available module.

To resolve this, we now fetch real module_id values from the server and include the database migration in this PR. If a user views the Learn screen before the next sync pulls down those real IDs (which happens during the Opportunity sync), those modules will temporarily have a module_id of 0 (UNKNOWN_MODULE_ID). This change gracefully handles that fallback logic for sorting the modules.

app/src/org/commcare/fragments/connect/ConnectLearningProgressFragment.java
Jignesh-dimagi: Unused code with no implementation; safely cleaned up.

app/src/org/commcare/android/database/connect/models/ConnectLearnModuleSummaryRecord.java
Jignesh-dimagi: The server always assigns valid module ID values. I've updated the KDoc to reflect this and clarify the behavior.

app/src/org/commcare/views/connect/ConnectLearnProgressView.kt
Jignesh-dimagi: Good suggestion! However, I don't think we need a dedicated UI state type here. That fallback logic lives entirely within the domain logic for determining the current active learn module and is never exposed to the view level.

app/src/org/commcare/views/connect/ConnectLearnProgressView.kt
Jignesh-dimagi: The primary reason for creating this custom UI component was to encapsulate its UI-specific presentation logic and make it testable in isolation. That said, I completely agree that any database operations or broader business logic shouldn't live here. What are your thoughts on keeping only the presentation state inside this class.

app/src/org/commcare/android/database/connect/models/ConnectLearnModuleSummaryRecord.java
Jignesh-dimagi: The server always returns a valid module ID here, so I've changed this field to be mandatory.

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
shubham1g5: +1 for renaming this to signify a boolean method like `ifModulesRemining` or similar.

app/src/org/commcare/views/connect/ConnectLearnProgressView.kt
shubham1g5: I understand the logc here but still think it's better  to not show the next module at all if we can't realistically determine what to show here.

Jignesh-dimagi: Now, We've updated the logic to hide the 'Next Module' UI if the Learn module lacks a valid module_id.

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
Jignesh-dimagi: Done

app/src/org/commcare/connect/repository/ConnectRequestManager.kt
```diff
@@ -38,18 +38,22 @@ object ConnectRequestManager {
...
                 deferred.cancel(e)
                 throw e
             } catch (e: Exception) {
+                inFlightRequests.remove(url, deferred)
                 deferred.completeExceptionally(e)
                 throw e
-            } finally {
-                inFlightRequests.remove(url)
             }
         }
```
shubham1g5: a `finally` call is more resilient with failures and crashes, what's causing these changes ?

app/src/org/commcare/views/connect/ConnectLearnProgressView.kt
Jignesh-dimagi: Moved application logic out of the view layer [here](https://github.com/dimagi/commcare-android/pull/3869/changes/fb116640995c5a6e7c6bc40953de462d443bde35).

shubham1g5 (approved):

PR #3878, opened 2026-08-24
CCCT-2723 Document reCAPTCHA SMS Defense Rollout And Add Toggle Script

Jignesh-dimagi:
## Technical Summary

`scripts/recaptcha_sms_defense.py` reads and toggles reCAPTCHA SMS Defense on the
Firebase project through the Identity Toolkit admin API, showing a diff, asking for
confirmation and backing up the config before any write. The doc records the
CCCT-2723 prep-up findings — key and threshold semantics, monitoring signals, cost,
and the staged test plan.

Nothing here enables anything; the rollout is still blocked on an unresolved
question about the Android site key.

### Safety story

What gives confidence:

- No app code, dependencies or resources are touched — a doc and a standalone
  operator script, neither of which ships to devices.
- The script is stdlib-only, so there is no new dependency to audit.
- I ran `status` against the live project and it returned the expected
  configuration; `--dry-run` builds the real PATCH payload and stops before sending.
- Every write path prints a diff, requires typed confirmation, and backs up the
  whole config before sending.

Risks to review:

- The PATCH path has never been executed. `firebaseauth.configs.update` is
  unconfirmed for our account, so the first real write is also the first test of it.
- The script writes production Firebase Auth config. A bad write affects phone
  sign-in for all four packages sharing the project, since `recaptchaConfig` is
  project-level with no per-app scoping.
- The Logs Explorer filter in the monitoring section is a starting point, not
  verified against real log entries.

[CCCT-2723]: https://dimagi.atlassian.net/browse/CCCT-2723?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/personalid/recaptcha_sms_defense.md
+++ b/docs/personalid/recaptcha_sms_defense.md
@@ -171,6 +171,10 @@ things are easy to conflate, though:
 Mobile keys are always score-based, so they are invisible to users. Our existing web key is
 CHECKBOX type and cannot be reused for Android.
 
+**Prefer the console for creation.** `gcloud` has no flag for SMS Toll Fraud Defense, so a
+gcloud-created key always needs a follow-up console visit — see
+[Enable SMS defense on the key](#enable-sms-defense-on-the-key).
+
 ### Create an Android key with gcloud
 
 One key can cover several packages
@@ -215,12 +219,13 @@ integration with reCAPTCHA"* — Identity Platform manages that one itself.
 Neither method takes SHA-256 as an input. Both ask only for a display name and package names, so an
 Android key is bound to the package name alone.
 
-> [!CAUTION]
-> ### Confirm with support before creating a reCAPTCHA key
+> [!NOTE]
+> ### Resolved: support's SHA-256 instruction
 >
-> The CCCT-2723 comment *"Email Response from Firebase"* instructs us to *"Enter your Package Name
-> and your SHA-256 fingerprint (matching what is in your Firebase Project Settings)"* **during
-> reCAPTCHA key creation**, which matches neither flow above.
+> The CCCT-2723 comment *"Email Response from Firebase"* instructed us to enter a package name and
+> SHA-256 during key creation, matching neither flow above. Resolved 2026-08-27: SHA-256 belongs to
+> the Firebase Android app (Project Settings → Your apps) and was already registered. The step
+> actually missing was [Enable SMS defense on the key](#enable-sms-defense-on-the-key).
 
 ## Register the key with Identity Platform
 
@@ -304,11 +309,43 @@ Expected output *after* registering:
   done while enforcement is still off — the safe order, since raising enforcement with no key
   registered leaves Identity Platform nothing to produce assessments with.
 
+## Enable SMS defense on the key
+
+**Console only**, and required. Neither `gcloud` nor the v1 Key resource exposes this setting, and
+`useSmsTollFraudProtection: true` does not imply it. Two layers must both be on:
+
+| Layer | Setting | Visibility |
+| --- | --- | --- |
+| Identity Platform | `useSmsTollFraudProtection` | API, set by [Toggle script](#toggle-script) |
+| reCAPTCHA Enterprise | SMS defense enablement | console only, manual |
+
+Google Cloud console → **reCAPTCHA** (or Fraud Defense) → select the key → configuration panel →
+enable **SMS Toll Fraud Defense**. It mirrors to **Settings → SMS defense → Configure**, so either
+surface confirms it.
+
+With the Identity Platform flag on and this off, assessments still run and return a bot score but no
+toll-fraud verdict: SMS Defense is inert while `status` reports it enabled. On
+`<firebase_project_id>` that state persisted for a day, producing `INVALID_APP_CREDENTIAL` on
+`SendVerificationCode`, a fallback to legacy reCAPTCHA v2, and continued error 39.
+
+### Verify
+
+No API exposes the setting, so check functionally: `sms_tf_risk_scores` records a point per assessed
+request once it is on, and nothing while it is off — see
+[Raw metric queries](#raw-metric-queries). On `<firebase_project_id>` the first score point landed
+the same second the toggle was enabled, so there is no propagation delay to wait out.
+
 ## Toggle script
 
 `scripts/recaptcha_sms_defense.py` turns reCAPTCHA SMS Defense on and off, and reports the current
 setting. Before changing anything it shows what will change and asks for confirmation, and it saves
 a backup of the old config so the change can be undone.
+
+> [!IMPORTANT]
+> `status` reads the Identity Platform layer only. It cannot see the console-only reCAPTCHA-side
+> enablement, so it will report `useSmsTollFraudProtection: true` even when no toll-fraud assessment
+> is happening. Treat it as necessary but not sufficient, and confirm with `sms_tf_risk_scores` —
+> see [Enable SMS defense on the key](#enable-sms-defense-on-the-key).
 
 ### reCAPTCHA ON and OFF with different modes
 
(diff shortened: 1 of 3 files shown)


Jignesh-dimagi:
## Suggested Review Order

- `docs/personalid/recaptcha_sms_defense.md` — read first; establishes the key, threshold and monitoring semantics the script relies on
- `scripts/recaptcha_sms_defense.py` — the toggle itself

shubham1g5:
The plan looks good to me here. Putting next steps on Jira ticket itself.

docs/personalid/recaptcha_sms_defense.md
```diff
@@ -0,0 +1,405 @@
...
+## Test plan
+
+The staged rollout from CCCT-2723. Every window runs during late EST / early IST, when Connect
+traffic is lowest.
+
+### 1. Prep up
+
+| Ticket item | Status |
+| --- | --- |
+| Review the reCAPTCHA docs, change what is needed without enabling it, flag blockers | Done. Main blocker is the site key: `recaptchaKeys` is empty, and support's instruction to supply a SHA-256 during key creation matches neither creation flow — see [Creating an Android key never asks for a SHA-256 fingerprint](#creating-an-android-key-never-asks-for-a-sha-256-fingerprint) |
```
shubham1g5, line 613: does this mean that you have now configured the `recaptchaKeys` corrrectly ? Otherwise it should not be marked as done.

Jignesh-dimagi: This needs to be confirmed with the Firebase support team. Creating an Android reCAPTCHA key (site key) only requires the package name—it never prompts for a SHA-256 fingerprint. However, according to an [email from Firebase support](https://dimagi.atlassian.net/browse/CCCT-2723?focusedCommentId=485022), it should require a SHA-256 fingerprint, which isn't happening. Please clarify this with them before moving forward.

Note: Nothing has been implemented yet since we are still in the planning phase. 'Done' in this context simply means accounted for in the plan.

shubham1g5: @Jignesh-dimagi  will followup with them, can we complete all prep up steps other than that though.

shubham1g5: Actually I can see the sha-256 already added for our app here - https://console.firebase.google.com/u/0/project/commcare-a57e4/appcheck/apps

Screenshot 2026-08-25 at 6 12 29 PM

docs/personalid/recaptcha_sms_defense.md
```diff
@@ -0,0 +1,405 @@
...
+./scripts/recaptcha_sms_defense.py off --project <firebase_project_id>
+```
+
+Note the metrics do not exist until this step begins, so an empty chart in the first minutes is
+expected rather than a fault.
+
+### 3. Enforce
+
+> [!IMPORTANT]
+> ### Open question: is enforce mode needed at all?
```
conroy-ricketts, line 652: I'd vote for `AUDIT` mode

conroy-ricketts (commented):
LGTM

Just to confirm, this does not need to be merged right?

docs/personalid/recaptcha_sms_defense.md
Jignesh-dimagi: I checked that, but I'm still confused by the steps provided by the Firebase support team. According to them, the site key generation process requires a SHA-256 fingerprint. Here is what they said:

'''
Click Configure site keys and add your Android app. Enter your Package Name and your SHA-256 fingerprint (matching what is in your Firebase Project Settings). This will generate the unique key your app needs to communicate with reCAPTCHA.
'''

I know the legacy reCAPTCHA console used to ask for SHA-1/SHA-256 fingerprints alongside package names when generating keys. If possible, let's wait for their response to clarify this. Otherwise, we can move ahead using latest reCAPTCHA Enterprise.

Jignesh-dimagi:
> LGTM
> 
> Just to confirm, this does not need to be merged right?

This PR includes a script to toggle reCAPTCHA on and off, which could be useful in the future. I vote to merge this so we have it readily available in case of an emergency.

docs/personalid/recaptcha_sms_defense.md
Jignesh-dimagi: @shubham1g5 

> @Jignesh-dimagi will followup with them, can we complete all prep up steps other than that though.

The first step to moving forward is creating the Site key, as nothing can proceed without it. Since the steps provided by Firebase support do not match the current setup interface, we have two options:

- Wait for Support: Wait for a reply from the Firebase support team to get clarification before proceeding.

- Move Ahead: Proceed immediately with creating the site key. I am confident that support provided outdated instructions from the legacy reCAPTCHA console, whereas we are using reCAPTCHA Enterprise. We can continue moving forward while still waiting for their response.

I am fine with either approach.

conroy-ricketts:
> This PR includes a script to toggle reCAPTCHA on and off, which could be useful in the future. I vote to merge this so we have it readily available in case of an emergency.

@Jignesh-dimagi 

That makes sense to me!

In that case, what are your thoughts on generalizing the plan doc to not mention the ticket specifically given that open source devs will not be able to access the ticket for context?

docs/personalid/recaptcha_sms_defense.md
shubham1g5: yeah I think it's best to trust the current documentation over their support and we can move ahead on this. I have copied you on a email with Firebase support as well where you can ask further questions on this if required.

docs/personalid/recaptcha_sms_defense_results.md
```diff
@@ -0,0 +1,273 @@
...
+```
+
+Swap `token_count` for `verdict_count` to get the second table. Both are counters: the totals are the
+sum of `points[].value.int64Value`, and the state is a metric *label* (`token_state` /
+`verdict_state`), so each state arrives as its own time series. Bucket on
+`points[].interval.endTime` for the per-day split.
+
+### Token Result
+
+| Day | Total | `valid` | `missing` | `invalid` | Success |
```
shubham1g5: do we know what's causing missing tokens ?

docs/personalid/recaptcha_sms_defense_results.md
```diff
@@ -0,0 +1,273 @@
...
+| `TOO_MANY_ATTEMPTS_TRY_LATER` | 3 | 1% |
+
+## Before Enabling ENFORCE Mode
+
+**Enabling `ENFORCE` while CommCare LTS is absent from the reCAPTCHA key would stop OTP delivery for
+those users entirely.** `org.commcare.lts` is not in the key's `allowedPackageNames` and
+`allowAllPackageNames` is `false`, so the LTS app cannot mint a token at all — every LTS phone-auth
+request necessarily lands in `missing`. Today that is survivable because `AUDIT` falls back to a
+silent push and then a visual reCAPTCHA, and the OTP still arrives. Under `ENFORCE` there is no
+fallback: the request is blocked and the client gets `onVerificationFailed` immediately.
```
shubham1g5, line 215: >  because `AUDIT` falls back to a silent push and then a visual reCAPTCHA,

As per the diagram at top, reCAPTCHA only appears when the Integrity check fails, while here we seem to say that it also appears when the token is missing ? 


Doesn't that also mean that today all LTS users are seeing reCAPTCHA already before getting otp ? If so, are you able to verify that with a LTS installation on your device.

Jignesh-dimagi: I verified that reCAPTCHA isn't triggering because Play Services is active on my Android device. Firebase silently falls back to Play Integrity or FCM push verification to pass Stage 1 (App Verification).

shubham1g5: So would not turning on the `ENFORCE` mode cause errors only when the play integerity fails and not always as mentioned in table below for CommCare LTS ?

shubham1g5: Separately are you able to see reCaptcha triggered on CommCare LTS if you install the app directly and not from play store (that should lead to integrity check fail ?)

Jignesh-dimagi: >Separately are you able to see reCaptcha triggered on CommCare LTS if you install the app directly and not from play store (that should lead to integrity check fail ?)

Whenever I sideloaded CommCare LTS, it opened reCAPTCHA, but the process was automated—it redirected to the browser, verified automatically, and returned to the app to send the OTP.

>So would not turning on the ENFORCE mode cause errors only when the play integerity fails and not always as mentioned in table below for CommCare LTS ?

That is still unconfirmed. Claude suggests that CommCare LTS requests will fail in any case, whereas Gemini's logic indicates that ENFORCE mode will fail only if Play Integrity fails. We have emailed Firebase support to clarify the behavior.

PR #3879, opened 2026-08-24
CCCT-2774 Update Connect Action Bar Icons To New Designs

OrangeAndGreen:
## Product Description

The refresh and notifications icons in the PersonalID/Connect action bar are updated to the new designs. The refresh icon changes from a cloud-with-arrow to two arrows in a circle, and the notification bell takes on the new outline shape. The unread badge is now part of the bell glyph itself rather than a separate red circle.

The order of the two menu options was also reversed to match the Figma design.

Screenshot:
image

## Technical Summary

The new icons are Material Symbols Outlined at weight 300, 24px. I confirmed the family by converting the Figma export's coordinates back to the 960 grid and matching them against Google's published `_wght300_24px.svg` sources, then generated the drawables from those canonical paths rather than from the Figma clipboard output.

Two things worth knowing before reading the diff:

- `ic_backup` is renamed to `ic_sync`, since "backup" no longer describes the artwork. `@color/ic_backup` went with it — the deleted drawable was its only consumer, and all three menus that show the icon already tint it white. The menu item *IDs* (`notification_cloud_sync`, `cloud_sync`) still say "cloud"; they're wired into Java/Kotlin so I left them alone.
- Only the unread bell appears in the Figma designs, and its source component has been removed from the "[OLD] CommCare Connect Components" library. I used Material Symbols `notifications` at the same weight as the read-state counterpart — worth a designer confirm.

### Safety story

What gives confidence:

- I ran the app and visually checked the Connect action bar icons, and confirmed the sync icon still renders on the work history and push notification screens.
- The diff is resource-only — no Java or Kotlin changes, so there is no behavior or data impact.
- `assembleCommcareDebug` passes and all three vector drawables compile into the APK, which means AAPT accepted the path data.
- I verified `@color/ic_backup` had no other consumer before removing it.

Risks to review:

- The unread badge changes color. It was a separate red `#E44434` circle; per Figma it is now white and part of the monochrome glyph. This is a deliberate design change, but it is a visible departure from what ships today and is the thing most likely to draw a "was this intended?" question.
- The read-state bell is inferred rather than taken from Figma (see Technical Summary). If the designer intended something else, that icon is wrong.
- `ic_bell` is also used untinted in the nav drawer footer, which sits on `@color/connect_blue_color`. I did not exercise the nav drawer with an unread notification, so the unread bell in that surface is unverified.
- `menu_work_history.xml` was pulled in because it referenced the same drawable. That screen is slightly outside the ticket's stated scope.

[CCCT-2774]: https://dimagi.atlassian.net/browse/CCCT-2774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


OrangeAndGreen:
## Suggested Review Order

- `app/res/drawable/ic_backup.xml` — the deleted icon; read it first for the before/after contrast
- `app/res/drawable/ic_sync.xml` — its replacement, the new two-arrows-in-a-circle glyph
- `app/res/drawable/ic_new_notification_bell.xml` — unread bell; the badge color change lives here
- `app/res/drawable/ic_bell.xml` — read-state bell, the one icon inferred rather than taken from Figma
- `app/res/menu/menu_connect.xml` — main Connect action bar, where both icons appear together
- `app/res/menu/menu_notification.xml`, `app/res/menu/menu_work_history.xml` — the other two sync-icon consumers
- `app/res/values/colors.xml` — removal of the now-orphaned `ic_backup` color
- `RELEASES.md` — release note only; no QA notes for this change

Jignesh-dimagi:
Looks good! Just a minor detail: the icon order in the implementation differs slightly from the design specs. The design places Sync then Notification (left-to-right), whereas the implementation currently has Notification then Sync. I'm not sure if the exact ordering is mandatory. I also noticed the spacing between them is slightly off compared to Figma.

Jignesh-dimagi (dismissed):

OrangeAndGreen:
> Just a minor detail: the icon order in the implementation differs slightly from the design specs

Ah definitely, good call. Easy enough to just fix it now so they're in the right order. [fd452de](https://github.com/dimagi/commcare-android/pull/3879/commits/fd452de5ba218e7833fbfd76e9403cb67b8f6c78)

I also added a screenshot in the description showing the icons in their new order.

Jignesh-dimagi (approved):

PR #3880, opened 2026-08-24
SAAS-20177 - Support rotation in the audio recording dialog

avazirna:
## Product Description

This PR adds support rotation to the audio recording dialog. Previously it locked the screen for the duration of a recording, so a user who turned the device mid-recording either could not, or lost the recording if the dialog was rebuilt for any other reason.

Rotating now keeps the recording running and the elapsed time correct, and the dialog picks the session back up in whatever state it was in — recording or paused. Notification Pause/Resume/Save presses that are triggered while the screen is turning are replayed once the dialog comes back rather than being dropped. 

<table>
<tr>
<td>
Device
</td>
</tr>
<tr>
<td>
<video width="200px" src="https://github.com/user-attachments/assets/75fd57d3-e58d-4bf4-8013-2ebb45d37662"></video>
</td>
</tr>
<tr>
<td>
Tablet
</td>
</tr>
<tr>
<td>

<video width="200px" src="https://github.com/user-attachments/assets/67fbc9a5-1478-48b5-9749-332dbbc4d08b"></video>
</td>
</tr>
</table>

## Technical Summary

Ticket: [SAAS-20177](https://dimagi.atlassian.net/browse/SAAS-20177)

Rotation was previously *prevented* rather than supported, and the code underneath was not recreation-safe. The design moves the authoritative state out of the fragment:
- **`AudioRecordingService` owns the session.** A `RecordingState` enum, the filename, and a `getChronometerBase()` that returns exactly what `Chronometer.setBase()` wants with paused stretches excluded. The fragment renders from the service on every bind, so a fresh recording and a rebuilt view take the same path through `restoreViewFromRecordingState()`.
- **The fragment rebinds rather than restarts.** `onDestroyView` splits on `isChangingConfigurations()` — a configuration change unbinds and leaves the recording running; any other teardown ends the session. `onViewCreated` branches on a persisted `recordingSessionStarted` flag instead of file existence.
- **Notification actions survive the rebuild window.** The service queues an action that arrives while nothing is bound and replays it when the new view binds, after the view state has been restored so the replay sees accurate state.
- **Outcome delivery moved to the fragment result API**, keyed by form index. A listener handed to the dialog cannot survive the widget being rebuilt; `setFragmentResultListener` also holds a result set before the listener registers, which matters because widgets are built only after the form finishes loading. Keying by index replaces a shared `"Recorder"` fragment tag that would have crossed wires between two audio questions on one screen.
- **Orientation handling removed entirely.** The old `enableScreenRotation()` restored `SENSOR`, not the `UNSPECIFIED` that `FormEntryActivityUIController.checkForOrientationRequirements()` expects, so the activity was left in the wrong mode after every recording.

### Safety story

- Tested locally on both a phone and a tablet, with rotation triggered repeatedly at various points. The behavior was successful in all cases.
- Existing recordings are unaffected by these changes: the widget's answer is still a file path — only the delivery mechanism changes (how that path reaches the widget), not what's stored.

### Automated test coverage
Added `AudioRecordingServiceTest.kt` with 8 tests covering the service's state machine and elapsed-time calculations, the areas of this change most likely to fail silently. Tests cover:
- elapsed time across an uninterrupted recording, a pause, a resume, and a stop
- stopping while paused does not re-freeze the base and count the pause
- repeated stops are ignored
- a second start intent does not restart a session already under way
- the filename is reported to a rebinding UI

## Labels and Review

[SAAS-20177]: https://dimagi.atlassian.net/browse/SAAS-20177?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/res/layout-land/recording_fragment.xml
+++ b/app/res/layout-land/recording_fragment.xml
@@ -1,0 +1,144 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+  Landscape counterpart of layout/recording_fragment.xml. Same ids, laid out as two columns so the
+  record control and the text/actions sit side by side rather than stacking past the bottom of a
+  short screen.
+-->
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="fill_parent"
+    android:layout_height="wrap_content"
+    android:orientation="vertical"
+    android:gravity="center_horizontal"
+    android:paddingStart="10dp"
+    android:paddingLeft="10dp"
+    android:paddingTop="10dp"
+    android:paddingEnd="10dp"
+    android:paddingRight="10dp"
+    android:paddingBottom="10dp"
+    tools:viewBindingIgnore="true">
+
+    <RelativeLayout
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content">
+
+        <TextView
+            android:id="@+id/recording_header"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:layout_alignParentStart="true"
+            android:layout_alignParentLeft="true"
+            android:layout_marginStart="10dp"
+            android:layout_marginEnd="10dp"
+            android:layout_toStartOf="@+id/discardrecording"
+            android:layout_toLeftOf="@+id/discardrecording"
+            android:textColor="@color/grey"
+            android:textSize="@dimen/text_medium" />
+
+        <ImageButton
+            android:id="@+id/discardrecording"
+            android:layout_width="50dp"
+            android:layout_height="50dp"
+            android:layout_alignParentEnd="true"
+            android:layout_alignParentRight="true"
+            android:background="@drawable/icon_close_darkwarm" />
+
+    </RelativeLayout>
+    <LinearLayout
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:orientation="horizontal">
+
+        <FrameLayout
+            android:id="@+id/recording_layout"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_gravity="center_vertical"
+            android:paddingHorizontal="10dp">
+
+            <ImageButton
+                android:id="@+id/startrecording"
+                android:layout_width="@dimen/audio_recording_button_size"
+                android:layout_height="@dimen/audio_recording_button_size"
+                android:layout_gravity="center"
+                android:background="@drawable/record_start" />
+
+            <ProgressBar
+                android:id="@+id/demo_mpc"
+                android:layout_width="@dimen/audio_recording_progress_size"
+                android:layout_height="@dimen/audio_recording_progress_size"
+                android:layout_gravity="center"
+                android:indeterminateDrawable="@drawable/progress_bar"
+                android:indeterminateDuration="2000"
+                android:visibility="invisible" />
+
+        </FrameLayout>
+
+        <LinearLayout
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_gravity="center_vertical"
+            android:gravity="center_horizontal"
+            android:orientation="vertical"
+            android:paddingHorizontal="10dp">
+
+            <Chronometer
+                android:id="@+id/recording_time"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:textSize="@dimen/audio_recording_timer_text_size"
+                android:visibility="gone" />
+
+            <TextView
+                android:id="@+id/recording_animation_text"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:layout_marginBottom="@dimen/standard_spacer_double"
+                android:drawableStart="@drawable/recording_dot"
+                android:drawablePadding="8dp"
+                android:text="@string/recording_in_progress"
+                android:textColor="@color/audio_recording_primary_color"
+                android:textSize="@dimen/text_large"
+                android:textStyle="bold"
+                android:visibility="invisible" />
+
+            <LinearLayout
+                android:id="@+id/recording_action_container"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:gravity="center"
+                android:visibility="gone"
+                android:orientation="horizontal">
+
+                <com.google.android.material.button.MaterialButton
+                    android:id="@+id/negative_action_button"
+                    android:layout_width="wrap_content"
+                    android:layout_height="wrap_content"
+                    android:layout_marginStart="15dp"
+                    android:layout_marginLeft="15dp"
+                    android:layout_marginEnd="15dp"
+                    android:layout_marginRight="15dp"
+                    android:layout_marginBottom="10dp"
+                    android:gravity="center"
+                    android:padding="10dp"
+                    style="@style/NegativeButtonStyle"
+                    tools:text="Finish"
+                    android:visibility="gone" />
+
+                <com.google.android.material.button.MaterialButton
+                    android:id="@+id/positive_action_button"
+                    android:layout_width="wrap_content"
+                    android:layout_height="wrap_content"
+                    android:layout_marginStart="15dp"
+                    android:layout_marginLeft="15dp"
+                    android:layout_marginEnd="15dp"
+                    android:layout_marginRight="15dp"
+                    android:layout_marginBottom="10dp"
+                    android:gravity="center"
+                    android:padding="10dp"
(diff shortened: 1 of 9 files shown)


app/src/org/commcare/views/widgets/AudioRecordingService.java
```diff
@@ -78,23 +121,25 @@ public int onStartCommand(Intent intent, int flags, int startId) {
...
-        String fileName = intent.getExtras().getString(RECORDING_FILENAME_EXTRA_KEY);
+        fileName = intent.getExtras().getString(RECORDING_FILENAME_EXTRA_KEY);
         pauseSupported = intent.getBooleanExtra(PAUSE_SUPPORTED_EXTRA_KEY, false);
         if (recorder == null) {
             recorder = audioRecordingHelper.setupRecorder(fileName,
                     DeveloperPreferences.getAudioQualityProfile());
         }
         recorder.start();
+        chronometerBase = SystemClock.elapsedRealtime();
+        state = RecordingState.RECORDING;
```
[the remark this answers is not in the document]
avazirna: The service is `START_NOT_STICKY`, so it's not expected for the system to restart the service if killed.

app/src/org/commcare/views/widgets/RecordingFragment.java
```diff
@@ -125,21 +134,43 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
...
         }
 
-        if (fileName == null) {
-            initAudioFile();
+        if (recordingSessionStarted) {
+            // Recreated mid-session; the service kept recording, so rebind and restore from it rather
+            // than looking at the audio file, which exists but is still being written to.
+            bindAudioRecordingService();
+            return;
         }
```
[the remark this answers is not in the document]
avazirna: triggering `stopRecording()` automatically saves the recording, so the dialog is not expected to be recreated

app/src/org/commcare/views/widgets/AudioRecordingService.java
```diff
@@ -28,11 +29,22 @@
...
  *
  * @author avazirna
  **/
 public class AudioRecordingService extends Service {
+
+    /**
+     * The lifecycle of a single recording session, from the service's point of view.
+     */
+    public enum RecordingState {
+        IDLE,
```
Jignesh-dimagi: Since this service is only initialized when recording starts, it realistically operates in only three states. The IDLE state is never actually reached, so we might want to simplify the state machine and remove it.

app/src/org/commcare/views/widgets/RecordingFragment.java
```diff
@@ -125,21 +134,43 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
...
     public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceState) {
         super.onViewCreated(view, savedInstanceState);
-        Bundle args = getArguments();
-        if (args != null) {
-            fileName = args.getString(AUDIO_FILE_PATH_ARG_KEY);
+        if (savedInstanceState != null) {
+            fileName = savedInstanceState.getString(FILE_NAME_STATE_KEY);
+            recordingSessionStarted = savedInstanceState.getBoolean(SESSION_STARTED_STATE_KEY);
+            savedRecordingExists = savedInstanceState.getBoolean(SAVED_RECORDING_EXISTS_STATE_KEY);
+            pausedByUser = savedInstanceState.getBoolean(PAUSED_BY_USER_STATE_KEY, true);
```
Jignesh-dimagi, line 142: This should be cleared in onDestroyView, provided it isn't a configuration change. Otherwise, it could cause issues if the user switches between apps while recording audio: the fragment may be recreated using saved instance state, but the current onDestroyView logic has already called finishAndRelease on the audio service.

app/res/layout-land/recording_fragment.xml
```diff
@@ -0,0 +1,144 @@
...
+    android:layout_height="wrap_content"
+    android:orientation="vertical"
+    android:gravity="center_horizontal"
+    android:paddingStart="10dp"
+    android:paddingLeft="10dp"
+    android:paddingTop="10dp"
+    android:paddingEnd="10dp"
+    android:paddingRight="10dp"
+    android:paddingBottom="10dp"
+    tools:viewBindingIgnore="true">
```
Jignesh-dimagi, line 11: We should wrap this layout in a ScrollView to ensure all UI components remain accessible across different screen sizes

avazirna: @Jignesh-dimagi I've decided to set dimensions for smaller screens instead - [73c95da](https://github.com/dimagi/commcare-android/pull/3880/commits/73c95da2caf9eb9f01dce403b55ea2bcf0751035)

app/src/org/commcare/views/widgets/RecordingFragment.java
avazirna: @Jignesh-dimagi good point. We are now calling `finishAndRelease` only on dismissal or when the host activity is finished, any other scenarios we try to rebind to the service during recreation. And in case rebind fails, any existing file is discarded - [ef50a4e](https://github.com/dimagi/commcare-android/pull/3880/commits/ef50a4e6870f766493686a8eacf8aa9150e14dd0)

app/src/org/commcare/views/widgets/RecordingFragment.java
```diff
@@ -244,8 +279,42 @@ private void startRecording() {
...
+     * @return whether there was a running recording to attach to
+     */
+    private boolean rebindToRunningRecording() {
+        Intent serviceIntent = new Intent(requireActivity(), AudioRecordingService.class);
+        boolean bound = requireActivity().bindService(serviceIntent,
+                getAudioRecordingServiceConnection(), 0);
+        // Unbinding is required even when the bind fails, so record that we asked.
+        audioRecordingServiceBounded = true;
+        return bound;
+    }
```
Jignesh-dimagi, line 302: Can we reuse bindAudioRecordingService here to avoid code duplication?

app/src/org/commcare/views/widgets/RecordingFragment.java
```diff
@@ -537,10 +605,24 @@ private boolean hasRecordingGoneSilent(List<AudioRecordingConfiguration> configs
...
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
             unregisterAudioRecordingConfigurationChangeCallback();
         }
+        // A config change, or the system reclaiming a stopped activity, rebuilds this view and the
+        // recording has to outlive it. Only a dismissal or a finishing activity ends the session.
+        boolean endingSession = isRemoving() || requireActivity().isFinishing();
+        if (endingSession && audioRecordingService != null) {
+            audioRecordingService.finishAndRelease();
+        }
+        unbindAudioRecordingService();
```
Jignesh-dimagi, line 615: Could we combine the finish, release, and unbind operations into one clear method?

app/res/layout-land/recording_fragment.xml
Jignesh-dimagi: How can we make sure all views fit on smaller screen sizes? I think wrapping this in a ScrollView could help. Thoughts?

app/src/org/commcare/views/widgets/RecordingFragment.java
avazirna: Might not be that clear, `unbindAudioRecordingService` is called during rotation while finishAndRelease() is not, only when we finish the recording. Also `finishAndRelease()` calls `recorder.release()` so it needs access to the `MediaRecorder` instance which lives in the service, while unbind needs to access the activity, which is accessible through the fragment.

avazirna:
@Jignesh-dimagi it seems that I can't comment there, about:

> How can we make sure all views fit on smaller screen sizes? I think wrapping this in a ScrollView could help. Thoughts?

I tested this on the emulator with the smallest screen I could find in my environment (320dp × 534dp). Here’s the result:


The calculation I did based on the layout is that we need 10dp+4dp of padding, a 40dp dismiss icon (taller than the text), and a 190dp progress bar. This comes to a total of 244dp, which is fine for a 320dp-wide device. 
But as a safety precaution, I'm happy to wrap the layout in a ScrollView - [9fffe4b](https://github.com/dimagi/commcare-android/pull/3880/commits/9fffe4b30002b6dd31f5ac6fff3cee5af78cc6c2)

app/src/org/commcare/views/widgets/RecordingFragment.java
avazirna: [20b968f](https://github.com/dimagi/commcare-android/pull/3880/commits/20b968f8d808dda9cd80d14636a9e61e4f8d9d14)

Jignesh-dimagi (approved):

PR #3882, opened 2026-08-25
Lint the PR's own diff, not the release branch's divergence from master

OrangeAndGreen:
Point `DEFAULT_BRANCH` at `github.base_ref` so the diff base is the branch the PR actually targets.


shubham1g5 (dismissed):
Great to see this fixed! 


Comment on verbosity: we can remove the comment and the complete PR description here as both of them do not provide extra context other than code diff and can be omitted without sacrificing correctness or context.

shubham1g5 (approved):

PR #3883, opened 2026-08-26
CCCT-2535 Show Payment Unit Cards On The Delivery Visits Tab

Jignesh-dimagi:
### [CCCT-2535](https://dimagi.atlassian.net/browse/CCCT-2535)

https://github.com/user-attachments/assets/fb6c5ddd-5df0-4782-a0d8-b4f4f75e171c

Screenshot_20260826_112428

Please ignore the bottom spacing for the CTA bar—it will be fixed in a follow-up [task](https://dimagi.atlassian.net/browse/CCCT-2778).

## Product Description

The Visits tab of a delivery opportunity was empty. It now lists a card per payment
unit with that unit's approved visits, amount earned and visits remaining, and tapping
a card opens the visit list filtered to that unit.

## Technical Summary

Restores the previous card layout design.

The identity handed to the detail screen is the payment unit's UUID rather than its
name — two payment units can carry the same name by mistake, a UUID cannot — so that
screen now filters deliveries on `slugUUID`, which is also what the dashboard's own
per-unit counts key on.

### Safety story

What gives confidence:

- I tested this by running it on a device.
- I ran the new test class and the existing `ConnectDeliveryHomeFragmentTest` suite; both
  pass. I also confirmed the new tests fail when the unit filter and the approved-only
  filter are broken, so they are not vacuous.
- Scope is one tab plus one filter predicate. No schema, migration or storage changes.

### Automated test coverage

`ConnectDeliveryVisitsFragmentTest` adds nine Robolectric tests driving the tab end to
end — the job seeded through the real Connect database, deliveries arriving over the real
networking stack from a mock server. They lock down the per-unit figures, that only
approved visits count, each phrasing of the remaining-visits line, and the tap-through
that opens the detail screen listing that unit's visits and no others.

[CCCT-2535]: https://dimagi.atlassian.net/browse/CCCT-2535?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/res/layout/connect_delivery_progress_item.xml
+++ b/app/res/layout/connect_delivery_progress_item.xml
@@ -5,25 +5,25 @@
     android:id="@+id/rootView"
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
-    android:layout_margin="10dp"
+    android:layout_margin="@dimen/connect_space_md"
     android:clickable="true"
     android:focusable="true"
     android:foreground="?android:attr/selectableItemBackground"
-    app:cardBackgroundColor="@color/white"
-    app:cardCornerRadius="15dp"
+    app:cardBackgroundColor="?attr/colorSurface"
+    app:cardCornerRadius="@dimen/connect_radius_lg"
     app:cardElevation="10dp">
 
     <androidx.constraintlayout.widget.ConstraintLayout
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
-        android:padding="16dp">
+        android:padding="@dimen/connect_space_lg">
 
         <TextView
             android:id="@+id/tvDeliveryTitle"
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
-            android:textColor="@color/black"
-            android:textSize="14sp"
+            android:textColor="?attr/connectOnSurfaceEmphasis"
+            android:textSize="@dimen/connect_text_body"
             android:textStyle="bold"
             app:layout_constraintStart_toStartOf="parent"
             app:layout_constraintTop_toTopOf="parent"
@@ -33,8 +33,8 @@
             android:id="@+id/imgApproved"
             android:layout_width="13dp"
             android:layout_height="13dp"
-            android:layout_marginStart="5dp"
-            android:layout_marginTop="10dp"
+            android:layout_marginStart="@dimen/connect_space_xs"
+            android:layout_marginTop="@dimen/connect_space_sm"
             android:contentDescription="@null"
             android:src="@drawable/ic_connect_thumb_up"
             app:layout_constraintStart_toStartOf="parent"
@@ -44,7 +44,7 @@
             android:id="@+id/approvedStats"
             android:layout_width="0dp"
             android:layout_height="wrap_content"
-            android:layout_marginStart="6dp"
+            android:layout_marginStart="@dimen/connect_space_sm"
             android:orientation="vertical"
             app:layout_constraintEnd_toStartOf="@id/centerGuideline"
             app:layout_constraintStart_toEndOf="@id/imgApproved"
@@ -54,17 +54,17 @@
                 android:layout_width="wrap_content"
                 android:layout_height="wrap_content"
                 android:text="@string/connect_results_summary_approved"
-                android:textColor="@color/connect_grey"
-                android:textSize="11sp"
+                android:textColor="?attr/connectOnSurfaceMuted"
+                android:textSize="@dimen/connect_text_caption"
                 android:textStyle="bold" />
 
             <TextView
                 android:id="@+id/tvApproved"
                 android:layout_width="wrap_content"
                 android:layout_height="wrap_content"
-                android:layout_marginTop="3dp"
-                android:textColor="@color/connect_green"
-                android:textSize="12sp"
+                android:layout_marginTop="@dimen/connect_space_xs"
+                android:textColor="?attr/connectStatusPositive"
+                android:textSize="@dimen/connect_text_caption"
                 android:textStyle="bold"
                 tools:text="10" />
 
@@ -72,9 +72,9 @@
                 android:id="@+id/tvDeliveryTotalAmount"
                 android:layout_width="wrap_content"
                 android:layout_height="wrap_content"
-                android:layout_marginTop="3dp"
-                android:textColor="@color/connect_green"
-                android:textSize="12sp"
+                android:layout_marginTop="@dimen/connect_space_xs"
+                android:textColor="?attr/connectStatusPositive"
+                android:textSize="@dimen/connect_text_caption"
                 android:textStyle="bold"
                 tools:text="10,000 MWK" />
         </LinearLayout>
@@ -83,7 +83,7 @@
             android:id="@+id/imgRemaining"
             android:layout_width="13dp"
             android:layout_height="13dp"
-            android:layout_marginStart="5dp"
+            android:layout_marginStart="@dimen/connect_space_xs"
             android:contentDescription="@null"
             android:src="@drawable/ic_connect_delivery_check_circle"
             app:layout_constraintStart_toStartOf="@id/centerGuideline"
@@ -93,7 +93,7 @@
             android:id="@+id/remainingStats"
             android:layout_width="0dp"
             android:layout_height="wrap_content"
-            android:layout_marginStart="6dp"
+            android:layout_marginStart="@dimen/connect_space_sm"
             android:orientation="vertical"
             app:layout_constraintEnd_toEndOf="parent"
             app:layout_constraintStart_toEndOf="@id/imgRemaining"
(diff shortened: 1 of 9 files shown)


Jignesh-dimagi:
## Suggested Review Order

- `app/src/org/commcare/models/connect/ConnectDeliveryDetails.kt` — the model the cards read
- `app/res/layout/connect_delivery_progress_item.xml` — the restored card design
- `app/src/org/commcare/adapters/ConnectDeliveryProgressReportAdapter.kt` — binds the model to the card
- `app/src/org/commcare/fragments/connect/ConnectDeliveryVisitsFragment.kt` — builds the per-unit figures and handles the tap
- `app/src/org/commcare/fragments/connect/ConnectDeliveryVisitsDetailFragment.java` — filters by UUID instead of name
- `app/res/navigation/nav_graph_connect.xml` — the action the tap navigates through
- `app/unit-tests/src/org/commcare/fragments/connect/ConnectDeliveryVisitsFragmentTest.kt` — end-to-end coverage of the above

app/res/layout/connect_delivery_progress_item.xml
conroy-ricketts, line 1: Can we derive all colors in this file from the theme instead as per the team's [theming convention ](https://docs.google.com/document/d/1rntc16FW2Jfr6CbzNcZ9RDOxIRrmfcWNU6_d5WL1rA8/edit?disco=AAACF45TCxg&tab=t.n6fzzoxreuvd)

It's understandable that we are reusing the old Payment Unit cards, but we should aim for all of Connect colors to be derived from a theme/style as per the reasons outlined in the convention doc

conroy-ricketts, line 1: Suggest reusing the values from `dimens.xml` for the font sizes and spacing in this layout (i.e. the `sp` and `dp` values) so that the Connect UI shares common values

app/src/org/commcare/adapters/ConnectDeliveryProgressReportAdapter.kt
```diff
@@ -0,0 +1,88 @@
...
+import android.content.Context
+import android.view.LayoutInflater
+import android.view.ViewGroup
+import androidx.core.content.ContextCompat
+import androidx.recyclerview.widget.RecyclerView
+import org.commcare.dalvik.R
+import org.commcare.dalvik.databinding.ConnectDeliveryProgressItemBinding
+import org.commcare.models.connect.ConnectDeliveryDetails
+
+class ConnectDeliveryProgressReportAdapter(
```
conroy-ricketts: **_Nit:_**

Can we rename this to be specific to either visits or payment units? It's not immediately obvious what a "progress report" is. 

How about `ConnectDeliveryPaymentUnitAdapter` or `ConnectDeliveryVisitsAdapter`?

app/src/org/commcare/adapters/ConnectDeliveryProgressReportAdapter.kt
```diff
@@ -0,0 +1,88 @@
...
+    private val onDeliveryClicked: (ConnectDeliveryDetails) -> Unit,
```
conroy-ricketts: _**Nit:**_

Similar concern here: thoughts on calling this either `onDeliveryVisitClicked` or `onDeliveryPaymentUnitClicked`? It's not obvious what "delivery" means here.

Also, similar concern for both the `deliveryProgressList` variable and the `ProgressBarViewHolder` class below this.

app/src/org/commcare/adapters/ConnectDeliveryProgressReportAdapter.kt
```diff
@@ -0,0 +1,88 @@
...
+
+    override fun onBindViewHolder(
+        holder: ProgressBarViewHolder,
+        position: Int,
+    ) {
+        val details = deliveryProgressList[position]
+        val context = holder.itemView.context
+        with(holder.binding) {
+            linearProgressBar.setProgress(details.approvedPercentage.toFloat())
+            linearProgressBar.setProgressColor(ContextCompat.getColor(context, R.color.connect_green))
```
conroy-ricketts: We should get this color from a theme or style

app/src/org/commcare/adapters/ConnectDeliveryPaymentUnitAdapter.kt
```diff
@@ -0,0 +1,88 @@
...
+            }
+        }
+    }
+
+    override fun getItemCount(): Int = deliveryProgressList.size
+
+    fun updateData(newData: List<ConnectDeliveryDetails>) {
+        deliveryProgressList.clear()
+        deliveryProgressList.addAll(newData)
+        notifyDataSetChanged()
```
conroy-ricketts, line 88: _**Nit:**_

Is it much trouble to use methods like `notifyItemChanged()` instead? That would be more efficient/performant which is a bit more useful for users on low-end devices.

app/unit-tests/src/org/commcare/fragments/connect/ConnectDeliveryVisitsFragmentTest.kt
```diff
@@ -0,0 +1,494 @@
...
+            )
+
+        assertEquals(25f, cards[0].progress(), 0.01f)
+        assertEquals("a unit with no deliveries sits at zero", 0f, cards[1].progress(), 0.01f)
+    }
+
+    @Test
+    fun `a unit that has met its target reports that all visits are done`() {
+        val cards =
+            cardsOn(launch(progressResponse(deliveries = approvedFor(unit = 1, count = perUnitTotal))))
```
conroy-ricketts: It's not obvious what these function calls are doing which hurts readability.

Can we rename the helper functions in this test class to be more clear on their purpose?

As an example, `cardsOn()` → `getPaymentUnitCards()`. But, we should look into renaming the rest as well, not just that one (i.e. `launch()`, `progressResponse()`, `approvedFor()`, etc.).

app/unit-tests/src/org/commcare/fragments/connect/ConnectDeliveryVisitsFragmentTest.kt
```diff
@@ -0,0 +1,494 @@
...
+                    }
+                return MockResponse().setResponseCode(200).setBody(body)
+            }
+        }
+
+    /**
+     * Waits for the delivery-progress response to be applied — the sync timestamp is written right
+     * after the database write — then drains the main looper so the resulting UI update lands.
+     */
+    private fun awaitDeliverySync() {
```
conroy-ricketts, line 394: Question: Why not use fake data instead? Shouldn't we avoid using a real database in tests?

app/src/org/commcare/fragments/connect/ConnectDeliveryVisitsDetailFragment.java
conroy-ricketts, line 1: Have these changes been demoed in Slack yet? We should do that before merging

conroy-ricketts (changes requested):
The UI looks fantastic! The Delivery page is really coming together

app/unit-tests/src/org/commcare/fragments/connect/ConnectDeliveryVisitsFragmentTest.kt
conroy-ricketts: Just realized my comment here may go against the recent testing guidelines. Feel free to disregard this particular comment

app/res/layout/connect_delivery_progress_item.xml
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3883/changes/e59c3e4e4434ae53fad4de75bfb831adbdc3b800)

Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3883/changes/e59c3e4e4434ae53fad4de75bfb831adbdc3b800)

app/src/org/commcare/adapters/ConnectDeliveryProgressReportAdapter.kt
Jignesh-dimagi: https://github.com/dimagi/commcare-android/pull/3883/changes/e59c3e4e4434ae53fad4de75bfb831adbdc3b800

app/src/org/commcare/adapters/ConnectDeliveryProgressReportAdapter.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3883/changes/1ffdaf53f761e28dd1d5b04d07420df844edb073)

app/src/org/commcare/adapters/ConnectDeliveryProgressReportAdapter.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3883/changes/1ffdaf53f761e28dd1d5b04d07420df844edb073)

app/unit-tests/src/org/commcare/fragments/connect/ConnectDeliveryVisitsFragmentTest.kt
Jignesh-dimagi: Done [here](https://github.com/dimagi/commcare-android/pull/3883/changes/1ffdaf53f761e28dd1d5b04d07420df844edb073)

app/src/org/commcare/fragments/connect/ConnectDeliveryVisitsDetailFragment.java
Jignesh-dimagi: Since the current app UI already matches this design, I don't think there's anything to demo right now. Plus, this is an intermediate state that will be updated very soon.

app/src/org/commcare/adapters/ConnectDeliveryPaymentUnitAdapter.kt
Jignesh-dimagi: Keeping notifyDataSetChanged() here deliberately. updateData isn't an incremental update — it's called from updateView(), which recomputes the entire list from the freshly reloaded job on every tab refresh. A sync can change the approved count, amount earned, remaining days and progress percentage across several payment units at once, so in practice the changed set is usually "most or all rows" rather than one.

conroy-ricketts (approved):
🚀

PR #3885, opened 2026-08-26
SAAS-20146 - Support device rotation during Wi-Fi Direct transfers

avazirna:
## Product Description

This PR adds support to device rotation to Wi-Fi Direct form transfer. It now rotates, and rotating mid-transfer no longer loses the session. Previously rotation was prevented rather than supported. It now works while choosing a mode, discovering peers, hosting, transfer and unzip, and the screen comes back with the same mode, peer list, connection status, device row and in-flight transfer.
<table>
<tr>
<td>
Sender
</td>
<td>
Receiver
</td>
</tr>
<tr>
<td>
<video width="200px" src="https://github.com/user-attachments/assets/085e36ae-ed08-4ca1-be4d-2e7b869814be"></video>
</td>
<td>
<video width="200px" src="https://github.com/user-attachments/assets/cbcf17af-a56f-46ae-a5e5-eeb17e3984ab"></video>
</td>
</tr>
</table>

## Technical Summary

Ticket: https://dimagi.atlassian.net/browse/SAAS-20146

The changes were:

1. **Remove the static state** — under a recreation, `static mActivity` / `mStatusText` / `receiveZipDirectory` were become correctness bugs, since the old instance's callbacks write into whichever instance last overwrote the static.
2. **Move the file server into a ViewModel** — the receive socket was owned by a non-retained XML fragment and bound port 8988. Rotating while hosting orphaned the task with the port held, so re-entering receive mode hit `BindException` until the process restarted. The self-relaunching `AsyncTask` becomes an accept loop whose socket is a field, which is what makes stopping it possible at all.
3. **Split entering a mode from rendering it** — `beSender`/`beReceiver`/`beSubmitter` each mixed rendering with side effects, so restoring a mode by replaying one would also remove the P2P group, clear the peer list and re-run the unzip. Now `enter*` (side effects, from the dialog only) and `render*` (view state, every create).
4. **One channel per session** — `initialize()` used to run on every create and the old `Channel` was never closed, leaking one per rotation.
5. **Restore the rest of the screen** — peer list, this device's name/status, and the status line, all of which were written only from broadcast callbacks that a recreated activity is never sent again.
6. **One data-loss fix beyond rotation**  - the receive loop keeps accepting while the activity is stopped, and a single-value `LiveData` was holding only the latest. So two transfers with the receiver backgrounded dropped the first, considering that the delivered zip's unzip deletes the whole receive directory which takes the dropped one with it. Received zips are now published as a queue the observer acknowledges entry by entry.

Design decisions worth a reviewer's attention:

- **Session state lives in ViewModels, not the saved-instance bundle.**  - the P2P group and the receive socket both die with the process, so a mode restored from a bundle after process death would be in an inconsistent state. A ViewModel dies with them, which makes "no state" a reliable signal to ask for a mode again rather than fake a restore. 
- **Re-derive from the system service in preference to carrying state across.** - the group and peer list are owned by the framework, not the activity, so we are not saving them to use in a restore. One exception: this device's own name and status are retained, because `requestDeviceInfo` is API 29 and the min SDK is 23. That row is display-only and the broadcast corrects it on the next real change.
- **The channel is built against the application context.** A `Channel` retains the `Context` it was made with, so hoisting one built from the activity would have swapped a leaked channel per rotation for a pinned dead activity.

### Safety story

- **Blast radius is one screen**, reached from Advanced Actions. No shared code paths are touched; nothing outside `CommCareWiFiDirectActivity` and its four fragments changes behaviour. 
- These changes were successfully tested locally on two devices, one running Android 16 and the other Android 15. The tests covered:
  - Rotation mid-transfer on the sender, and mid-unzip on the receiver
  - Multiple rotations during a single transfer
  - Sending two batches back to back with the receiver backgrounded

## Labels and Review

### DIFF
--- a/app/src/org/commcare/fragments/FileServerViewModel.kt
+++ b/app/src/org/commcare/fragments/FileServerViewModel.kt
@@ -131,7 +131,7 @@ class FileServerViewModel : ViewModel() {
     }
 
     companion object {
-        private val TAG = LogTypes.TYPE_WIFI_DIRECT
+        private const val TAG = LogTypes.TYPE_WIFI_DIRECT
         private const val PORT = 8988
     }
 }
--- a/app/src/org/commcare/fragments/FileServerViewModel.kt
+++ b/app/src/org/commcare/fragments/FileServerViewModel.kt
@@ -1,0 +1,125 @@
+package org.commcare.fragments
+
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import org.commcare.activities.CommCareWiFiDirectActivity
+import org.commcare.util.LogTypes
+import org.javarosa.core.services.Logger
+import java.io.Closeable
+import java.io.File
+import java.io.FileOutputStream
+import java.io.IOException
+import java.net.ServerSocket
+import java.net.Socket
+
+class FileServerViewModel : ViewModel() {
+    private val _statusText = MutableLiveData<String>()
+    val statusText: LiveData<String> = _statusText
+
+    private val _receivedZipPath = MutableLiveData<String>()
+    val receivedZipPath: LiveData<String> = _receivedZipPath
+
+    @Volatile
+    private var running = false
+
+    @Volatile
+    private var serverSocket: ServerSocket? = null
+
+    val isRunning: Boolean
+        get() = running
+
+    fun startServer(receiveZipDirectory: String) {
+        if (running) {
+            Logger.log(TAG, "File server already running, keeping the existing socket")
+            return
+        }
+        Logger.log(TAG, "File Server starting...")
+        running = true
+        _statusText.value = "Starting server"
+        viewModelScope.launch(Dispatchers.IO) { serve(receiveZipDirectory) }
+    }
+
+    fun stopServer() {
+        running = false
+        serverSocket.closeQuietly()
+    }
+
+    private fun serve(receiveZipDirectory: String) {
+        val socket =
+            try {
+                ServerSocket(PORT)
+            } catch (e: IOException) {
+                Logger.exception("Wi-fi direct file server could not bind port $PORT", e)
+                _statusText.postValue("Could not start the file server, the port is in use.")
+                running = false
+                return
+            }
+        serverSocket = socket
+
+        try {
+            while (running) {
+                _statusText.postValue("Ready to accept new file transfer.")
+                val client =
+                    try {
+                        socket.accept()
+                    } catch (e: IOException) {
+                        if (running) {
+                            Logger.exception("Wi-fi direct file server stopped accepting connections", e)
+                            _statusText.postValue("File server stopped.")
+                        }
+                        break
+                    }
+                Logger.log(TAG, "Ready in wi-fi direct file server receive loop")
+                receive(client, receiveZipDirectory)
+            }
+        } finally {
+            running = false
+            socket.closeQuietly()
+            serverSocket = null
+        }
+    }
+
+    private fun receive(
+        client: Socket,
+        receiveZipDirectory: String,
+    ) {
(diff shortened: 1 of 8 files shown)


app/res/values-w720dp/dimens.xml
```diff
@@ -14,4 +14,5 @@
     <dimen name="nav_drawer_width">320dp</dimen>
     <dimen name="side_nav_horizontal_padding">16dp</dimen>
     <dimen name="side_nav_sign_out_top_margin">64dp</dimen>
+    <dimen name="wifi_direct_button_row_side_margin">180dp</dimen>
```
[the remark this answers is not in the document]
avazirna: Good point. I will actually remove this margin and size buttons according to available width

app/src/org/commcare/fragments/FileServerViewModel.kt
```diff
@@ -0,0 +1,152 @@
...
+        try {
+            val f = File(receiveZipDirectory + System.currentTimeMillis() + ".zip")
+            f.parentFile?.mkdirs()
+            f.createNewFile()
+
+            Logger.log(TAG, "server: copying files $f")
+            CommCareWiFiDirectActivity.copyFile(client.getInputStream(), FileOutputStream(f))
+
+            _statusText.postValue("copied files: ${f.absolutePath}")
+            _receivedZipPath.postValue(f.absolutePath)
```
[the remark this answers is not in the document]
avazirna: I've addressed this in [0948b88](https://github.com/dimagi/commcare-android/pull/3885/commits/0948b8860709b90c85387544169b6e2cccabb5f3)

app/src/org/commcare/fragments/FileServerViewModel.kt
```diff
@@ -0,0 +1,137 @@
...
+    companion object {
+        private val TAG = LogTypes.TYPE_WIFI_DIRECT
+        private const val PORT = 8988
+    }
+}
+
+private fun Closeable?.closeQuietly() {
+    try {
+        this?.close()
+    } catch (e: IOException) {
```
conroy-ricketts: Should we do something with the exception here?

app/src/org/commcare/fragments/FileServerViewModel.kt
```diff
@@ -0,0 +1,137 @@
...
+        }
+    }
+
+    override fun onCleared() {
+        super.onCleared()
+        stopServer()
+    }
+
+    companion object {
+        private val TAG = LogTypes.TYPE_WIFI_DIRECT
```
conroy-ricketts: _**Nit:**_

We could make this a `const`

app/src/org/commcare/fragments/WiFiDirectManagementFragment.java
```diff
@@ -138,6 +130,17 @@ public void startReceiver(WifiP2pManager mManager, Channel mChannel) {
...
     }
 
+    public void refreshConnectionInfo() {
+        if (mManager == null || mChannel == null) {
+            return;
+        }
+        try {
+            mManager.requestConnectionInfo(mChannel, this);
+        } catch (SecurityException e) {
+            Logger.log(TAG, "Cannot read Wi-fi direct connection info without permission");
```
conroy-ricketts: Seems like this is something worth logging and tracking as non-fatal to Firebase?

app/src/org/commcare/fragments/FileServerViewModel.kt
avazirna: I actually found a shared one. Using that instead - [808223b](https://github.com/dimagi/commcare-android/pull/3885/commits/808223b22097a2da9a61a6c17a8f2bb70ffcb1f4)

app/src/org/commcare/fragments/WiFiDirectManagementFragment.java
avazirna: [33b83f5](https://github.com/dimagi/commcare-android/pull/3885/commits/33b83f58d07ee034185f93c31a7b27419b86bd26)

conroy-ricketts (approved):

PR #3887, opened 2026-08-27
CI-909 Preload A Placeholder Photo On The QA Automation Build

OrangeAndGreen:
## Technical Summary

Adds the ability for the special QA build to avoid using the camera to capture a user photo during PersonalID account registration.
On entering the Photo page (final page in the configuration workflow when registering a new account) when `IS_QA_AUTOMATION` is true, the default placeholder image (already in the codebase, for accounts with no user photo) is rendered and converted to a base64-encoded string that can be sent to the server in `complete_profile`.
After setting the image, the "Save Photo" button is enabled so the user (or automation script) can proceed.

Demo video, note the Save Photo button is enabled immediately when entering the page (0:28), and account creation succeeds when the Save Photo button is pressed:

https://github.com/user-attachments/assets/fc22395b-008e-4f5e-a281-a1d97d89d83b

### Safety story

What gives confidence:

- Manually enabled the `IS_QA_AUTOMATION` flag in code and tested the workflow (see demo video)
- Verified the Photo page still works like before when the flag is false

[CI-909]: https://dimagi.atlassian.net/browse/CI-909?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -14,6 +14,8 @@ This file is meant as an easy way for us to collate notes and change logs across
 
 - On a device where PersonalID sign-up previously got stuck due to a Firebase OTP error, a pre-invited user will now receive the OTP via Twilio automatically and can complete the sign-up flow.
 - Confirm that entering an incorrect verification code still displays the "incorrect code" error and does not silently trigger a new OTP.
+- On the QA automation build, the photo step of PersonalID sign-up should open with a placeholder photo already shown and Save Photo enabled, and saving it should create the account without the camera ever opening.
+- On a normal build, the photo step should be unchanged: Save Photo stays disabled until a photo is actually taken.
 
 ## CommCare 2.63.4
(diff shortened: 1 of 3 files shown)


OrangeAndGreen:
## Suggested Review Order

- `app/src/org/commcare/utils/QaAutomationPlaceholderPhoto.kt` — the new helper; read it first to see what the placeholder actually is and how it is encoded
- `app/src/org/commcare/fragments/personalId/PersonalIdPhotoCaptureFragment.java` — the single call site and the build-type seam that gates it
- `RELEASES.md` — QA notes, including the regression check on normal builds

shubham1g5 (approved):

PR #3888, opened 2026-08-28
CI-911 Require All Learn Modules Before Claiming A Job

OrangeAndGreen:
## Product Description

Fixes a bug where the user could potentially proceed to claim an opportunity (i.e. start delivery) before they were ready to do so (not all learn modules submitted yet). For this scenario to occur the following had to happen:
* Learn app designed such that the assessment is visible before all learn modules have been submitted
* User submits and passes the assessment before submitting all learn modules (i.e. they skip some modules)

At that point, some logic points in the mobile code would allow the user to proceed to the Delivery Details page and attempt to claim the job, even though the server would not allow that until all learn modules had been submitted.

The changes in this PR fix those logic points, so mobile won't advance the user in the workflow until the assessment has been passed AND al learn modules have been submitted.

An extra thing addressed by this fix: If the user gets into this special state, the progress bar will now still be visible in the Learning Progress page (before it was hidden once `passedAssessment` became true).

## Demos
Note Sample Job 7 reports 66% done with 1 of 2 learn modules submitted and the assessment passed (2/3 = 66%)

Old behavior (v2.64), incorrectly navigates to Delivery Details:

https://github.com/user-attachments/assets/2d924651-5999-4872-af29-62fd8e58d7bc

New behavior, navigates to Learning Progress and shows progress bar with 66% complete:

https://github.com/user-attachments/assets/d885f9c5-49bc-425a-b0a8-bd7909ea9888

## Technical Summary

* Added `ConnectJobRecord.isLearningComplete()` as the single point of truth (SPOT)
* Swapped the new function in (in place of `passedAssessment`) at several places in the code
* Replaced a couple places with identical logic to use the SPOT helper instead
* Added bug fix and QA notes in RELEASES.md

## Safety story

- Small code change, affects logic in very specific and easily-verifiable places
- Tested the change by completing assessment before submitting all modules in a test app and verifying behavior.

[CI-911]: https://dimagi.atlassian.net/browse/CI-911?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
+++ b/app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
@@ -542,7 +542,7 @@ public class ConnectJobRecord extends Persisted implements Serializable {
     }
 
     /**
-     * Whether the user has finished everything requiresd before claiming the job:
+     * Whether the user has finished everything required before claiming the job:
      * every learn module submitted, and a passing assessment score
      */
     public boolean isLearningComplete() {
--- a/app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
+++ b/app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
@@ -541,6 +541,14 @@ public class ConnectJobRecord extends Persisted implements Serializable {
                 || getAssessmentScore() >= getLearnAppInfo().getPassingScore();
     }
 
+    /**
+     * Whether the user has finished everything requiresd before claiming the job:
+     * every learn module submitted, and a passing assessment score
+     */
+    public boolean isLearningComplete() {
+        return getLearningPercentComplete(false) >= 100 && passedAssessment();
+    }
+
     public int getAssessmentScore() {
         int mostRecentFailingScore = 0;
         int firstPassingScore = -1;
@@ -644,7 +652,7 @@ public class ConnectJobRecord extends Persisted implements Serializable {
     }
 
     public boolean readyToTransitionToDelivery() {
-        return status == STATUS_LEARNING && passedAssessment();
+        return status == STATUS_LEARNING && isLearningComplete();
     }
 
     @Nullable
--- a/app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
+++ b/app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
@@ -377,7 +377,7 @@ public class ConnectJobRecord extends Persisted implements Serializable {
      */
     public int getLearningPercentComplete(boolean includeAssessmentModule) {
         int totalModules = numLearningModules;
-        int modulesCompleted = learningModulesCompleted;
+        int modulesCompleted = Math.min(learningModulesCompleted, numLearningModules);
 
         if (includeAssessmentModule) {
             // Add 1 to the calculation to represent the assessment module.
@@ -546,7 +546,7 @@ public class ConnectJobRecord extends Persisted implements Serializable {
      * every learn module submitted, and a passing assessment score
      */
     public boolean isLearningComplete() {
-        return getLearningPercentComplete(false) >= 100 && passedAssessment();
+        return getLearningPercentComplete(true) >= 100;
     }
 
     public int getAssessmentScore() {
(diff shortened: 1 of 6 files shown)


OrangeAndGreen:
## Suggested Review Order

- `app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java` — defines `isLearningComplete()`; read the doc comment first, everything else consumes it
- `app/src/org/commcare/fragments/connect/ConnectJobsListsFragment.java` — the reachable bypass to the claim screen, the actual CI-911 repro
- `app/src/org/commcare/fragments/connect/ConnectLearningProgressFragment.java` — two behavior-neutral swaps plus the progress-bar visibility change
- `app/src/org/commcare/fragments/connect/ConnectJobDetailBottomSheetDialogFragment.java` — cosmetic milestone indicator, no logic impact
- `RELEASES.md` — release and QA notes, in their own commits

app/src/org/commcare/fragments/connect/ConnectLearningProgressFragment.java
```diff
@@ -130,12 +130,13 @@ private void updateProgressViews(int learningProgressPercent, boolean hideProgre
...
-    private void updateCertificateView(boolean learningComplete, boolean passedAssessment) {
+    private void updateCertificateView() {
+        boolean showCertificate = job.isLearningComplete();
         getBinding().connectLearningCertificateContainer.setVisibility(
-                learningComplete && passedAssessment ? View.VISIBLE : View.GONE
+                showCertificate ? View.VISIBLE : View.GONE
         );
 
-        if (learningComplete && passedAssessment) {
+        if (showCertificate) {
```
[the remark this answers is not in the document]
OrangeAndGreen: Since the certificate is specific to passing the assessment, I think we're okay with showing that date on the certificate even if the user completes the rest of the learn modules later.

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
```diff
@@ -541,6 +541,14 @@ public boolean passedAssessment() {
                 || getAssessmentScore() >= getLearnAppInfo().getPassingScore();
     }
 
+    /**
+     * Whether the user has finished everything requiresd before claiming the job:
```
conroy-ricketts: Typo here: "requiresd"

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
```diff
@@ -541,6 +541,14 @@ public boolean passedAssessment() {
...
+     * every learn module submitted, and a passing assessment score
+     */
+    public boolean isLearningComplete() {
+        return getLearningPercentComplete(false) >= 100 && passedAssessment();
```
conroy-ricketts: Would `getLearningPercentComplete(true) >= 100` also work here? (to simplify this logic)

conroy-ricketts (approved):
Looks great, nothing blocking

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
OrangeAndGreen: Ah indeed, thanks! [5c8d76e](https://github.com/dimagi/commcare-android/pull/3888/commits/5c8d76eb7de03d25a72b70a6018a209c52ac285b)

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
OrangeAndGreen: Yeah I think so, although I added one additional guard in getLearningPercentComplete for the following:
* Learn app has 5 modules
* User completes 4 modules
* Learn app is changed to only have 3 modules

Before, getLearningPercentComplete would have returned 100% even though the assessment wasn't passed yet:
(4 completed modules + 0 passed assessment) / (3 total modules + 1 required assessment) = 100%

I capped num completed so it can't be greater than the total, ensuring that the additional +1 in the denominator for the assessment has to be matched in the numerator with a +1 for the passed assessment in order for the function to return 100%.

One explanation that was investigated as part of this interrupt ticket was that the learn app might have been changed to define **more** modules, and while the case above would be the app being changed to define **fewer** modules I think it's worth being ready for the behavior and handling it properly on mobile side if it happens. [b13c666](https://github.com/dimagi/commcare-android/pull/3888/commits/b13c6662068ef23d8d84d0252dc96726d059aae5)

conroy-ricketts: Interesting! Thanks for the context

conroy-ricketts (approved):

PR #3889, opened 2026-08-28
CCCT-2534 Add Outstanding Tasks And Revisit Learning To The More Tab

conroy-ricketts:
## Product Description

The More tab now lists outstanding tasks by expiry, opens a task's conversation, and shows the learning certificate.

https://github.com/user-attachments/assets/8af1cb70-9fc8-46aa-aa01-1a3db83f9125

Here's a video showing the OCS task card. It's a fake one that I forced in the code just so that I can show the UI. If it was real, tapping it would take me directly to the OCS message, not the entire channel list

https://github.com/user-attachments/assets/0fb5a8c6-6338-4f8d-b1e7-ae9437f07ddf

This is what they look like if they have no expiration date

Screenshot_20260828-101905_CommCare Debug

**_Edit:_** Here's an updated video of the More tab. I moved the pending task count badge to the correct side of the tab (directly to the right of "More"), and I also added some UI for when we do the 2nd data fetch for the learn app data

https://github.com/user-attachments/assets/ed2e8359-0aae-4e8a-8703-3ed55bf28c4f

## Technical Summary

Learning records are fetched only after delivery progress lands — both save the same stored opportunity, so running them together lets one overwrite the other's fields. The design's Optional task category is omitted: the server carries no blocking flag.

### Safety story

**What gives confidence**
- I exercised the More tab on a device, including tapping through to a fake OCS conversation.
- Full unit test suite passes.

**Risks to review**
- Learning fetch is silent — a failure leaves the certificate hidden with no error.
- A task missing its channel id crashes by design.

### Automated test coverage

Robolectric covers task ordering, badge count, empty state, and Revisit Learning; unit tests cover the learning completion date.

[CCCT-2534]: https://dimagi.atlassian.net/browse/CCCT-2534?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


conroy-ricketts:
## Suggested Review Order

Reading commit-by-commit tells the cleanest story — the seven commits are ordered as below.

1. `PushNotificationAdapter.kt` — pure ktlint reformat, split out so it doesn't obscure the rest.
2. `attrs.xml`, `themes.xml`, `colors.xml`, `dimens.xml`, `strings.xml` (+ translations), drawables — new theme roles, tokens, and renamed icons.
3. `ContextExtensions.kt`, `ImageViewExtensions.kt` — shared theme-color and tint helpers used throughout.
4. `ConnectTaskCard.kt`, `view_connect_task_card.xml` — the reusable card the tab is built from.
5. `ConnectJobRecord.java`, `ConnectCertificateBinder.kt`, `ConnectLearnCompleteView.kt`, `ConnectLearningProgressFragment.kt` — learning completion date moved onto the opportunity.
6. `ConnectTaskRecord.kt`, `ConnectTaskUtils.kt` — JSON-null parsing fix and the pending-task query.
7. `ConnectDeliveryMoreFragment.kt`, `ConnectDeliveryHomeFragment.kt`, `ConnectDeliveryHomeViewModel.kt`, `fragment_connect_delivery_more.xml` — the tab itself and the fetch ordering.
8. Tests — `ConnectDeliveryMoreFragmentTest.kt` documents the intended behavior best.

app/src/org/commcare/connect/database/ConnectTaskUtils.kt
```diff
@@ -75,6 +75,19 @@ object ConnectTaskUtils {
...
+     * without a due date sorts last; the API always sends one, so that is defensive only.
+     */
+    @JvmStatic
+    fun getPendingTasksForJob(
+        context: Context,
+        jobUUID: String,
+    ): List<ConnectTaskRecord> =
+        getTasksForJob(context, jobUUID, null)
+            .filter { it.status == ConnectTaskRecord.STATUS_ASSIGNED }
+            .sortedWith(compareBy(nullsLast()) { it.dueDate })
```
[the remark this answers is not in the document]
conroy-ricketts: [AI] Not acting on this, for three reasons:

1. The divergence predates this PR. `getPendingTaskOfMode` already has no legacy fallback, so `getValidPendingOcsTask` — which drives the "Open Conversation" button — already returns null in exactly this state. The new query inherits the established semantics rather than introducing a new inconsistency.

2. The window is one sync wide. `storeTasks` writes task rows *and* rewrites the legacy preference on every successful sync, and delivery progress syncs on the host fragment's `onResume`. Reaching the divergent state requires a device upgraded from a pre-task-records build whose first sync has not succeeded.

3. The suggested remedy is not implementable. A boolean preference carries no title, expiry, or channel id, so a synthesized pending task would render as a blank card and tapping it would fail the channel-id precondition. The synthetic-record approach used by `getMostRecentlyCompletedTask` only works because its caller reads a single field.

The legacy fallback is already slated for removal (see the KDoc on `hasPendingTask`), which closes this at the source.

app/src/org/commcare/fragments/connect/ConnectDeliveryMoreFragment.kt
```diff
@@ -1,16 +1,185 @@
...
+    ) = ConnectTaskCard(requireContext()).apply {
+        bind(
+            ConnectTaskCard.State(
+                title = task.name,
+                iconRes =
+                    if (task.isOCSConversation) {
+                        R.drawable.ic_chat_bubble_outline
+                    } else {
+                        R.drawable.ic_connect_learn_app
+                    },
```
Jignesh-dimagi, line 99: ConnectTaskCard should be a self-reliant component that accepts a workflow type (e.g., CONVERSATION or LEARN). This keeps the layout modular and makes it easy to support new card types in the future

app/src/org/commcare/fragments/connect/ConnectDeliveryMoreFragment.kt
```diff
@@ -1,16 +1,185 @@
...
+    /**
+     * Every outstanding task blocks delivery today and the API carries no blocking flag, so they all
+     * render as mandatory and the optional group stays hidden until that flag exists.
+     */
+    private fun bindTasks() {
+        val tasks = ConnectTaskUtils.getPendingTasksForJob(requireContext(), job.jobUUID)
+
+        binding.deliveryTasksEmpty.isVisible = tasks.isEmpty()
+        bindTaskGroup(binding.deliveryTasksMandatoryGroup, binding.deliveryTasksMandatoryList, tasks)
+        binding.deliveryTasksOptionalGroup.isVisible = false
```
Jignesh-dimagi: If I understand correctly, we don't currently have optional tasks in our system, so this view's visibility is set to GONE—is that correct?

app/src/org/commcare/fragments/connect/ConnectDeliveryMoreFragment.kt
```diff
@@ -1,16 +1,185 @@
...
+    /**
+     * Learning is always complete by the delivery stage, so the card is always offered. Its date and
+     * certificate wait on the learn records, which only reach a device that ran the learn sync.
+     */
+    private fun bindRevisitLearning() {
+        val learnCompletionDate = job.learningCompletionDate
+        val hasCompletionDate = learnCompletionDate != null
+        binding.revisitLearningTitle.text = job.title
+
+        if (hasCompletionDate) {
```
Jignesh-dimagi, line 150: Should this always have a completion date, given that the user is already on a delivery?

app/src/org/commcare/fragments/connect/ConnectDeliveryMoreFragment.kt
```diff
@@ -1,16 +1,185 @@
...
+
+    private fun toggleCertificate() {
+        certificateExpanded = !certificateExpanded
+        binding.revisitLearningCertificate.root.isVisible = certificateExpanded
+    }
+
+    /** Launching is the host's job, so the install check and download screen stay in one place. */
+    private fun launchApp(isLearning: Boolean) {
+        (parentFragment as? ConnectDeliveryHomeFragment)?.launchApp(isLearning)
+    }
```
Jignesh-dimagi, line 193: Tightly casting parentFragment as ConnectDeliveryHomeFragment might break if this fragment is reused elsewhere in the future. Let's define an openlaunchApp() method in BaseConnectFragment instead, which ConnectDeliveryHomeFragment can override as needed.

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
```diff
@@ -546,6 +547,32 @@ public boolean attemptedAssessment() {
         return assessments != null && !assessments.isEmpty();
     }
 
+    /**
+     * The date learning finished: the latest assessment date, or the latest completed module if the
+     * user never took an assessment. Null when this device holds no dated record of either, which is
+     * the case until it has run a learn sync.
+     */
+    @Nullable
+    public Date getLearningCompletionDate() {
```
Jignesh-dimagi: getLearningCompletionDate — the name doesn't match what this does. It returns the last learn update date: the latest assessment the user sat (passed or failed), falling back to the latest completed module. Could we rename to something like 'getLastLearnUpdateDate' or 'getLatestLearningActivityDate'?

Also, there seems bugs, attemptedAssessment() only checks that the list is non-empty, so a single assessment record with an unparseable date returns null even when good module dates exist; the else if never runs.

RELEASES.md
```diff
@@ -2,6 +2,18 @@
...
+
+### Release Notes
+
+<!--
+These are published publicly on Playstore, Github Releases and CommCare Forums
+-->
+
+#### What's New
+
+- The More tab of a delivery opportunity now lists outstanding tasks with their expiry dates, and gives access to the learn app and the learning certificate.
```
Jignesh-dimagi: Let's remove this for now, as we'll add a note once the full redesign is complete.

app/src/org/commcare/fragments/connect/ConnectLearningProgressFragment.kt
```diff
@@ -75,15 +74,17 @@ class ConnectLearningProgressFragment :
...
+            learnCompletionDate != null && job.getLearningPercentComplete(false) >= 100 && job.passedAssessment()
 
         binding.learnProgressView.visibility = if (showLearningComplete) View.GONE else View.VISIBLE
         binding.learnCompleteView.visibility = if (showLearningComplete) View.VISIBLE else View.GONE
 
         if (showLearningComplete) {
             binding.learnCompleteView.bind(
                 job,
-                latestCompletionDate(),
+                learnCompletionDate,
```
Jignesh-dimagi: So on a fresh install, with a job already at STATUS_DELIVERING: percent complete is 100, passedAssessment() is true, but the learnCompletionDate is null — and a worker who is actively delivering gets shown the in-progress module view, which may send them back into the learn app. The old latestCompletionDate() couldn't return null (it fell back to Date()), so the gate was payload-only and internally consistent.

Jignesh-dimagi (changes requested):

app/src/org/commcare/fragments/connect/ConnectDeliveryMoreFragment.kt
conroy-ricketts: I've updated this PR to not implement the optional task UI at all after the discussion on the demo I put in Slack

RELEASES.md
conroy-ricketts: That sounds right to me!

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
conroy-ricketts: Agreed on the function rename!

The bug you're describing was actually intentional/deliberate. The idea was that if an assessment exists, then learning ended at the assessment. However, a regular learn module date isn't an approximation of that, but rather an earlier and therefore wrong date. And, this value gets printed on a certificate. I figured silently stamping a certificate with a date we know is too early is worse than showing no date.

What are your thoughts on that?

app/src/org/commcare/fragments/connect/ConnectLearningProgressFragment.kt
conroy-ricketts: Not sure if I completely understand the ask here, but please let me know if this reply [here](https://github.com/dimagi/commcare-android/pull/3889#discussion_r3897767674) covers it

app/src/org/commcare/fragments/connect/ConnectDeliveryMoreFragment.kt
conroy-ricketts: The reasoning here is that, although learning is complete when a job is in delivery, the date lives in learn records that the user's device may not have synced yet.

app/src/org/commcare/fragments/connect/ConnectDeliveryMoreFragment.kt
conroy-ricketts: How do you feel about leaving the cast for now given that each call site would navigate to the download page through its own generated Directions, so a shared implementation needs a per-subclass hook?

I can make a note to revisit this in [CCCT-2521](https://dimagi.atlassian.net/browse/CCCT-2521) when we remove that navigation to the download page, and make the `launchApp()` function a plain protected method on the `ConnectJobFragment`, where the job already lives, and the cast goes away.

app/src/org/commcare/fragments/connect/ConnectDeliveryMoreFragment.kt
conroy-ricketts: Is this a strong opinion?

I figured that `iconRes` is generic (i.e. a new card type needs no change to `ConnectTaskCard`), and `ConnectTaskCard` as it is currently is very reusable. A workflow enum would require a new case and branch inside the card for every type, which I feel like makes it less extensible.

app/src/org/commcare/fragments/connect/ConnectDeliveryMoreFragment.kt
conroy-ricketts: https://github.com/dimagi/commcare-android/commit/b4f28e5e4

RELEASES.md
conroy-ricketts: https://github.com/dimagi/commcare-android/commit/d366b930f

app/src/org/commcare/android/database/connect/models/ConnectJobRecord.java
conroy-ricketts: https://github.com/dimagi/commcare-android/commit/438f45c9f

app/src/org/commcare/fragments/connect/ConnectDeliveryMoreFragment.kt
Jignesh-dimagi: I don't have a strong opinion, but I feel this decision should ideally be handled internally by the component rather than by the caller. That said, I'm okay with keeping it as is or going with your preference here.

app/src/org/commcare/fragments/connect/ConnectDeliveryMoreFragment.kt
Jignesh-dimagi: I'm fine including this under CCCT-2521. Let's make sure to leave a note on that ticket to address and remove the hardcoded casting.

app/src/org/commcare/fragments/connect/ConnectLearningProgressFragment.kt
Jignesh-dimagi: As per your [comment](https://github.com/dimagi/commcare-android/pull/3889#discussion_r3897915298), if the user hasn't synced yet, hasCompletionDate might not exist in the local state. Relying on hasCompletionDate here could introduce a bug—it would incorrectly display the 'Learn Progress' view instead of the 'Learn Complete' view for un-synced completed states.

Jignesh-dimagi:
I also noticed a slight design mismatch with the tab badge: the [Figma design](https://www.figma.com/design/4QLInPUv5wIfjfDGfrFESW/Opportunity-Home-Design-_-Ishwari?node-id=6001-24058&t=G19mnaLyORqaGFhp-1) places the number badge beside the tab label, whereas the current implementation renders it centered above the text. Could we adjust the badge layout to align with the design spec?

app/src/org/commcare/fragments/connect/ConnectDeliveryMoreFragment.kt
Jignesh-dimagi: Yeah ok. Instead of hiding the entire certificate due to our current program state error, could we render the certificate UI in full and simply leave the completion date field blank?

app/src/org/commcare/fragments/connect/ConnectDeliveryMoreFragment.kt
conroy-ricketts: Sounds good, I'll keep it as-is for now and, if needed, we can revisit later

app/src/org/commcare/fragments/connect/ConnectDeliveryMoreFragment.kt
conroy-ricketts: Sounds like a plan! I've added a note on the ticket [here](https://dimagi.atlassian.net/browse/CCCT-2521?focusedCommentId=488347)

app/src/org/commcare/fragments/connect/ConnectLearningProgressFragment.kt
conroy-ricketts: Ah I see, I'll update this!

app/src/org/commcare/fragments/connect/ConnectDeliveryMoreFragment.kt
conroy-ricketts: I wouldn't necessarily call it an error, because the app will have to wait for the learn progress sync to finish so that we can show the information on the "Revisit Learning" card.

If we want to leave the completion date field blank until that sync finishes, I think we would also have to show blank fields for other values on the certificate (e.g. the score would be hidden as well) which wouldn't look so good.

I'm going to take this over to Slack for input from the Product team and tag you!

conroy-ricketts: I proceeded with what Kim agreed was the best approach here. I added a new commit

app/src/org/commcare/fragments/connect/ConnectLearningProgressFragment.kt
conroy-ricketts: I've reverted this logic to match the same behavior on `master` before we started work on the "More" tab

app/src/org/commcare/fragments/connect/ConnectDeliveryMoreFragment.kt
conroy-ricketts: https://github.com/dimagi/commcare-android/commit/22b1e3552

app/src/org/commcare/fragments/connect/ConnectLearningProgressFragment.kt
conroy-ricketts: https://github.com/dimagi/commcare-android/commit/aa6cdc30e

conroy-ricketts:
@Jignesh-dimagi 

> I also noticed a slight design mismatch with the tab badge: the [Figma design](https://www.figma.com/design/4QLInPUv5wIfjfDGfrFESW/Opportunity-Home-Design-_-Ishwari?node-id=6001-24058&t=G19mnaLyORqaGFhp-1) places the number badge beside the tab label, whereas the current implementation renders it centered above the text. Could we adjust the badge layout to align with the design spec?

Great eye! Updated here: 3d196a7ff9a2a3c9ff791c0dc8eb1673a362315e

app/src/org/commcare/fragments/connect/ConnectDeliveryMoreFragment.kt
```diff
@@ -1,16 +1,233 @@
...
+
+                override fun onLost(network: Network) = rebindOnMainThread()
+
+                private fun rebindOnMainThread() {
+                    view?.post { if (isAdded) bindRevisitLearning() }
+                }
+            }
+        manager.registerDefaultNetworkCallback(callback)
+        connectivityCallback = callback
+    }
```
Jignesh-dimagi: ConnectDeliveryHomeFragment already reloads learn data whenever network availability changes. Having ConnectDeliveryMoreFragment also listen for network changes is redundant, as the request in ConnectDeliveryHomeFragment will already be in flight and may not reflect updated data yet. Since ConnectDeliveryMoreFragment already observes the learn data stream, it should simply update its UI by invoking bindRevisitLearning whenever the state transitions to success.

app/src/org/commcare/fragments/connect/ConnectDeliveryHomeFragment.kt
```diff
@@ -138,6 +188,52 @@ class ConnectDeliveryHomeFragment :
...
+        if (job.latestLearningActivityDate == null) {
+            viewModel.loadLearningProgress(job)
+        }
+    }
+
+    private fun retryLearningFetchOnReconnect() {
+        val manager = requireContext().getSystemService(ConnectivityManager::class.java) ?: return
+        val callback =
+            object : ConnectivityManager.NetworkCallback() {
+                override fun onCapabilitiesChanged(
```
Jignesh-dimagi, line 209: onCapabilitiesChanged can be called multiple times whenever network-related metadata is updated. This will result in firing multiple requests simultaneously. I think we need something like the code below to prevent these duplicate calls:

```
val hasInternet = networkCapabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)
val isValidated = networkCapabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED)
val isOnline = hasInternet && isValidated
  if (isOnline != lastIsValidated) {
         lastIsValidated = isOnline
         // Update UI here
  }
```

app/res/layout/view_connect_tab_label.xml
```diff
@@ -0,0 +1,34 @@
...
+    <TextView
+        android:id="@android:id/text1"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:ellipsize="end"
+        android:maxLines="1"
+        android:textAppearance="@style/TabTextStyle"
+        android:textColor="?attr/colorOnPrimary"
+        android:textSize="@dimen/connect_text_tab"
+        tools:text="More" />
```
Jignesh-dimagi, line 18: Is this actually being used? Looks like tab.setText handles setting the title directly, so we might be able to remove this.

app/src/org/commcare/fragments/connect/ConnectLearningProgressFragment.kt
```diff
@@ -144,12 +144,6 @@ class ConnectLearningProgressFragment :
                 false,
             )
 
-    private fun latestCompletionDate(): Date {
```
Jignesh-dimagi, line 147: I think we should revert this change as well. If the data isn't synced yet, this approach could incorrectly display the current date instead of the actual completion date. Reverting it will ensure we reliably fallback to showing the last learning update date.

app/src/org/commcare/fragments/connect/ConnectDeliveryHomeFragment.kt
```diff
@@ -138,6 +188,52 @@ class ConnectDeliveryHomeFragment :
...
+                    network: Network,
+                    capabilities: NetworkCapabilities,
+                ) {
+                    val validated = capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED)
+                    if (validated && !networkValidated) {
+                        view?.post {
+                            if (isAdded && job.latestLearningActivityDate == null) {
+                                refresh(false)
+                            }
```
Jignesh-dimagi: We should call loadLearningProgressIfMissing() here if our goal is specifically to fetch learning progress. If the intention is to fetch both delivery and learning data, relying solely on checking job.latestLearningActivityDate seems incorrect.

Jignesh-dimagi (changes requested):

app/res/layout/view_connect_tab_label.xml
conroy-ricketts: It's required because we are using a custom view to move the position of the pending task count badge. Removing it while keeping `tab.setText()` will cause all the tabs to render blank text

app/src/org/commcare/fragments/connect/ConnectDeliveryHomeFragment.kt
conroy-ricketts: Totally agreed that we need some sort of guard here. But curious why the current guard does not already achieve the outcome we want here?

```
val validated = capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED)
if (validated && !networkValidated) {
    view?.post {
        if (isAdded && job.latestLearningActivityDate == null) {
            // Update UI here
        }
    }
}
networkValidated = validated
```

 Regardless, think it's fine and harmless to add the `hasInternet` flag to the same guard. I'll make that tweak very soon

conroy-ricketts: https://github.com/dimagi/commcare-android/commit/88ab5b698

app/src/org/commcare/fragments/connect/ConnectDeliveryHomeFragment.kt
conroy-ricketts: That's a good point. I'm going to extract this into a helper with a short comment!

Long story short, we do want to fetch both the delivery and learning data again in this case because it solves a bug that I ran into earlier in implementation. Please let me know if I should provide more context on that issue

conroy-ricketts: https://github.com/dimagi/commcare-android/commit/2f2a041b0

app/src/org/commcare/fragments/connect/ConnectDeliveryMoreFragment.kt
conroy-ricketts: Great point, I will simplify this and move the logic to the home fragment instead! I think this allows us to clean up the code a little too

conroy-ricketts: https://github.com/dimagi/commcare-android/commit/fa69f40da

app/src/org/commcare/fragments/connect/ConnectLearningProgressFragment.kt
conroy-ricketts: That was my original concern as well. On `master`, before any of these changes were made, we were incorrectly displaying the current date instead of the actual completion date in some cases. See [here](https://github.com/dimagi/commcare-android/blob/3db15bfc638c0829fd020d0d8f79ff725c026164/app/src/org/commcare/fragments/connect/ConnectLearningProgressFragment.kt#L150). Reverting this back to `master` will have the same issue

Think our options here are either 1) keep falling back to the current date or 2) fall back to a `null` value and either display an empty value on the learn certificate or hide the certificate. Thoughts on this?

Think for the sake of this PR, I'm leaning towards keeping the fall back to the current date to match the same behavior for learn certificates on the `master` branch for now

conroy-ricketts: For what it's worth, I think the fall back here would very rarely happen for our users

app/src/org/commcare/fragments/connect/ConnectDeliveryHomeFragment.kt
```diff
@@ -112,21 +131,55 @@ class ConnectDeliveryHomeFragment :
...
+
+        badge.isVisible = pendingTasks > 0
+        badge.text = pendingTasks.toString()
+    }
+
+    /**
+     * The tabs read the opportunity back off the activity, and the repository hands back a fresh
+     * instance each sync, so the refreshed job has to be published there and not just kept here.
+     */
     private fun observeDeliveryProgress() {
```
Jignesh-dimagi: Nit: Please rename to 'observeDeliveryAndLearningProgress'

app/src/org/commcare/fragments/connect/ConnectLearningProgressFragment.kt
Jignesh-dimagi: Nit: I'd still prefer falling back to the original code here, if possible

Jignesh-dimagi (dismissed):

app/src/org/commcare/fragments/connect/ConnectDeliveryHomeFragment.kt
conroy-ricketts: Totally agreed!

conroy-ricketts: https://github.com/dimagi/commcare-android/commit/462a05efb

app/src/org/commcare/fragments/connect/ConnectLearningProgressFragment.kt
conroy-ricketts: Chose to skip on this. Adding a little more context here before resolving this thread:

Reverting wouldn't change the date behavior in question. The old code fell back to today's date in the same case. Moreover, a revert here would restore a crash: it compared dates without filtering out ones that failed to parse, so any job with two or more learn records where one date is bad would take down the screen. A single bad date was safe; two or more were not.

Jignesh-dimagi (approved):

